PR #236, opened 2026-07-02
HTTP/2 upstream hardening: reset/GOAWAY metrics, idle eviction, TCP_NODELAY (#145 Phase 4b PR3)

Jaruso:
**What changed and why**
Completes the HTTP/2 upstream connection pool (#141 Phase 4b / #145) — the hardening/completion PR on top of the multiplexing actor + pool landed in #235.

1. **Reset/GOAWAY metrics.** Two new pool-level monotonic counters:
   - `tardigrade_upstream_h2_stream_resets_total` — RST_STREAM frames received.
   - `tardigrade_upstream_h2_goaway_total` — GOAWAY frames received.

   Each `H2Conn` reader thread bumps atomics **owned by the pool** (a pointer handed to `H2Conn.init`) rather than a per-connection field, because a live-connection snapshot loses the count when an evicted connection is torn down.

2. **Idle / lifetime eviction.** `H2ConnPool.reapIdle` now runs on the gateway maintenance tick next to `upstream_pool.reapIdle`. It closes an h2 connection that has **no in-flight streams** and is idle past the idle timeout, past max lifetime, or already dead (GOAWAY / errored). Refcount-safe: the `activeStreamCount() == 0` gate protects actively-multiplexing connections, and the `release` that may join the reader thread runs **after** the pool mutex is dropped so a teardown never blocks the pool. Each connection stamps `last_activity_ms` on stream start and finish. Reuses the h1 pool's `TARDIGRADE_UPSTREAM_POOL_IDLE_TIMEOUT_MS` / `…_MAX_LIFETIME_MS` knobs.

3. **TCP_NODELAY on the h2 upstream socket.** h2 multiplexing issues many small frame writes (HEADERS / WINDOW_UPDATE); their interaction with the peer's delayed ACK stalled each exchange ~40 ms and tripped response timeouts under concurrency. Disabling Nagle fixes it (see Performance impact).

Also adds `benchmarks/h1-vs-h2-upstream.sh` (h1 pool vs h2 multiplexed) and updates `CHANGELOG.md` + `docs/UPSTREAM_POOLING.md`.

Deferred (later work): the streaming proxy path over h2, cleartext h2c, and per-`{upstream}`-labelled h2 series.

**Related issues**
Part of #141 (Phase 4b) and #145. (Not closing either — streaming-h2 / h2c remain.)

**Tests**
- New unit tests in `upstream_h2.zig`: `evictionDecision` policy (active-stream / health / idle / lifetime gates, caps-disabled), and reader RST_STREAM / GOAWAY counter bumps over a socketpair (canned reset/goaway servers). Suite: 610/611 pass (1 pre-existing skip); test binary run 10× directly for race coverage (thread-spawning tests), all clean.
- Manual e2e vs a real `nghttpd` h2 origin on Linux (`TARDIGRADE_UPSTREAM_PROTOCOL=h2`, TLS verify off): proxying negotiates h2 (protocol h2 counter, h1=0); 20 sequential + concurrent requests multiplex over **one** upstream connection; the idle reaper evicts the connection after the idle timeout and a fresh request reconnects; fd count stays flat across repeated create→reap cycles (no leak); new reset/GOAWAY counters read correctly (0 when no resets — no false positives); no "event loop wait error" logs.

**Security impact**
None. No changes to header parsing, routing, auth, TLS negotiation, path handling, or logging. `TCP_NODELAY` is a transport socket option on the upstream connection only.

**Performance impact**
Positive on the h2 upstream path; neutral elsewhere. `reapIdle` adds an O(conns) pass on the once-per-tick maintenance thread (one origin ⇒ one conn in the common case). The `TCP_NODELAY` fix is the headline: measured vs `nghttpd` on Linux loopback, the buffered h2 proxy path went from **24 req/s → 5.4k req/s** (single-request latency 35 ms → 0.7 ms) with the whole concurrent load multiplexed over one upstream connection and zero client-visible 5xx. `benchmarks/h1-vs-h2-upstream.sh` captured h1 pool ≈ 3.9k req/s over 4 upstream conns (one per busy worker) vs h2 ≈ 5.4k req/s over 1 conn.

**Checklist**
- [x] `zig build test --summary all` green (610/611, 1 pre-existing skip)
- [ ] `zig build test-integration` green — integration suite is environmentally broken on the dev machine; relied on unit tests + manual e2e vs nghttpd
- [x] docs/CODE_REVIEW_CHECKLIST.md considered
- [x] README / CHANGELOG updated (CHANGELOG + docs/UPSTREAM_POOLING.md)

---
_Generated by [Claude Code](https://claude.ai/code/session_01Cgsje6PnoAvo6SNusyfUNb)_


Jaruso (commented):
Strong follow-up to #235 overall. The shape looks right to me: pool-owned RST/GOAWAY counters are the correct place for monotonic protocol counters because they survive connection teardown; the h2 idle/lifetime reaper belongs on the same maintenance tick as the h1 pool; and `TCP_NODELAY` on the upstream h2 socket is a very plausible fix for the delayed-ACK/Nagle stall.

A few suggestions before merge, mostly around making the new lifecycle invariant consistent everywhere:

1. **Avoid `release()` while holding `H2ConnPool.mutex` everywhere, not only on the happy-path reaper.**

   The `reapIdle` comment says the final `release()` may join the reader thread and therefore should run after the pool mutex is dropped. The normal victim path does that, but the OOM fallback still releases under the lock:

   ```zig
   victims.append(self.allocator, kv.value) catch {
       kv.value.release(); // no room to defer — release under lock
   };
   ```

   I’d rather skip/defer that eviction than ever join/teardown while holding the pool mutex. Since this is hardening code, the rare OOM path should preserve the same concurrency invariant.

2. **Same concern in the stale-entry path in `acquire()`.**

   When an existing mapped conn is unhealthy, the code removes it and calls `old.value.release()` while still holding `self.mutex`. That predates this PR’s reaper logic conceptually, but now that we’ve documented “release after unlock” as the safe lifecycle pattern, I think the acquire stale-eviction path should follow it too: capture the stale conn, unlock, release.

3. **Same concern in `H2ConnPool.evict()`.**

   `evict()` removes the mapped conn and calls `conn.release()` under the pool lock. I’d make this consistent with `reapIdle`: remove under the pool lock, release after unlocking. It reduces the chance of a reader-thread join blocking unrelated acquire/reap/snapshot work.

4. **Clarify/count RST_STREAM semantics.**

   The new metric is described as “RST_STREAM frames received,” but the counter only increments when the reset maps to a currently-known stream. That may be exactly what you want, but if so I’d update the help/docs to say active-stream resets. If the metric is intended to be protocol-level frames received, increment it for every RST_STREAM frame, including late/unknown stream ids.

None of these change the architecture. The PR is aligned with the h2/pooling direction; I’d just tighten these lifecycle edges before merging.

PR #322, opened 2026-07-10
feat(http3): add RFC 9218 priority model and PRIORITY_UPDATE helpers (#254)

Jaruso:
**What changed and why**

Implement RFC 9218 extensible prioritization for the pure-Zig HTTP/3 path at the model / parser / frame-helper layer, so requests can carry urgency and incremental hints and peers can exchange `PRIORITY_UPDATE`. This is the foundation layer only — it does not wire a scheduler, and deliberately does not recreate HTTP/2's dependency tree (RFC 9218's whole point).

New module `src/http3/priority.zig` (exported from `src/http3/root.zig`); the only other change is the one-line root export. Everything stays inside `src/http3/`.

- **Priority field-value model** — `Priority { urgency: u3 (0..7, default 3), incremental: bool (default false) }`. `urgency` is a `u3`, so an out-of-range value is unrepresentable rather than something to defensively check downstream.
- **Parser** over an RFC 8941 Dictionary subset: `u=0..7`, the bare `i` / `i=?0` / `i=?1` forms, omitted-parameter defaults, duplicate parameters (last wins, flagged), unknown parameters (ignored per RFC 9218 §4). Out-of-range / wrong-typed known parameters are folded to the default and flagged (`Parsed.invalid_parameter`) per the RFC's "MUST ignore" rule; only malformed Dictionary *syntax* is a hard `error.MalformedPriority`.
- **Serializer** — canonical Dictionary output that omits defaults (both-default → `""`) and round-trips through `parse`.
- **PRIORITY_UPDATE helpers** for the request (0x0F0700) and push (0x0F0701) variants, reusing the existing `frame.FrameType` values: prioritized element ID as a QUIC varint followed by the priority field-value bytes. Both payload-only (`encodePayload`/`decodePayload`) and full-frame (`encodeFrame`/`decodeFrame`) codecs; a non-PRIORITY_UPDATE frame decodes to `error.NotPriorityUpdate`.
- **Model-layer `Metrics`** (parse errors / invalid / duplicate parameters, update counts) with a `recordParse` folder, so parse quality is countable before a scheduler exists.
- **Default scheduling policy is documented in comments** (lower urgency first; incremental streams round-robin within a band; equal priority stays FIFO/round-robin and starvation-free; no dependency tree, no adaptive scheduling) but intentionally **not implemented**.

**Related issues**

Part of #254 (does **not** close it). This PR is the model/parser/frame-helper layer. The issue additionally covers integrating priority hints with the QUIC stream scheduler, `PRIORITY_UPDATE` handling in the session/control-stream path, and HTTP/3 integration tests — those remain follow-ups so this stays a focused, reviewable change.

**Tests**

Unit tests in `src/http3/priority.zig` cover: urgency/incremental parsing and combinations, defaults (absent header and omitted params), invalid urgency (out-of-range and wrong-type) folding to defaults with flags, duplicate parameters (last-wins), unknown-parameter ignoring, malformed-syntax rejection, serialize round-trips across all urgency×incremental values, and PRIORITY_UPDATE payload + full-frame round trips for both request and push variants, plus the metrics folder.

Verified locally with Zig 0.16.0:
- `zig fmt --check build.zig src/ tests/` — clean
- `zig build test-quic --summary all --error-style verbose` — 249/249
- `zig build test --summary all --error-style verbose` — 1068 passed, 1 pre-existing skip

**Security impact**

None. This adds header parsing for the `priority` field and PRIORITY_UPDATE codecs, but the parser is bounds-checked and allocation-free over caller-owned slices, rejects malformed input with a typed error, and folds invalid values to safe defaults. It is not yet wired into routing, auth, TLS, or the request path, so no runtime behavior changes.

**Performance impact**

None on any existing path. The module is additive and not yet referenced by the running server. The parser/serializer are allocation-free and operate on small fixed inputs.

**Checklist**
- [ ] `zig build test-integration` — N/A: additive model/codec layer, not wired into the runtime protocol path
- [ ] README / CHANGELOG — N/A: no operator-visible behavior change (not yet wired)

---
_Generated by [Claude Code](https://claude.ai/code/session_01Cn2GoVQ3Jyxd163MFqycuJ)_

### DIFF
--- a/src/http3/priority.zig
+++ b/src/http3/priority.zig
@@ -1,4 +1,4 @@
-//! RFC 9218 Extensible Prioritization for HTTP/3 (#254): the `Priority` field
+//! RFC 9218 Extensible Prioritization for HTTP/3 (#254): the `priority` field
 //! value model + parser/serializer, and PRIORITY_UPDATE payload helpers.
 //!
 //! This module is intentionally transport-free and scheduler-free. It models
@@ -7,7 +7,7 @@
 //! scheduler, and deliberately does not recreate HTTP/2's dependency tree
 //! (RFC 9218's whole point is to avoid that).
 //!
-//! ## `Priority` field value (RFC 9218 §4)
+//! ## `priority` field value (RFC 9218 §4)
 //!
 //! The `priority` request/response header (and the PRIORITY_UPDATE field value)
 //! is an RFC 8941 Structured Fields Dictionary with two defined members:
@@ -15,12 +15,27 @@
 //!   * `u` — urgency, an Integer in 0..7. Lower is more urgent. Default 3.
 //!   * `i` — incremental, a Boolean. Default false. `i` (bare) means `i=?1`.
 //!
-//! Per RFC 9218 §4.1/§4.2, a known parameter whose value is out of range or the
-//! wrong type MUST be ignored and the default used; unknown parameters are also
-//! ignored. `parse` follows that leniently but records that it happened
-//! (`Parsed.invalid_parameter` / `.duplicate_parameter`) so callers can count a
-//! metric. Only genuinely malformed Dictionary *syntax* is a hard error
-//! (`error.MalformedPriority`).
+//! Parsing follows RFC 9218 §4 exactly: parse the *whole* Dictionary as RFC
+//! 8941 (so unknown members and otherwise-valid Structured Fields syntax —
+//! Decimals, Byte Sequences, Inner Lists, Item parameters — are accepted), then
+//! keep only the last occurrence of each key (RFC 8941 duplicate handling), then
+//! ignore any known member whose final value is out of range or the wrong type,
+//! falling back to the default. `parse` records that leniency
+//! (`Parsed.invalid_parameter` / `.duplicate_parameter`) for metrics. Only
+//! genuinely malformed Structured Fields *syntax* is a hard
+//! `error.MalformedPriority` (the caller should then ignore the field and use
+//! defaults, per RFC 9218 §4.1).
+//!
+//! ### Member presence vs. effective value (RFC 9218 §8)
+//!
+//! Request/PRIORITY_UPDATE and response semantics differ: in a response, an
+//! omitted parameter means "keep the client-provided value", not "use the
+//! default". So the wire model (`Field`) preserves *presence* — `urgency` and
+//! `incremental` are optional, and a value is present only if a valid one was
+//! given. `Field.serialize` emits exactly the members that are present, so a
+//! response can explicitly send `u=3` to reset urgency to the default. The
+//! defaults-applied, always-valid view used by a scheduler is `Priority`, via
+//! `Field.effective()`.
 //!
 //! ## Default scheduling policy (documented, not implemented here)
 //!
@@ -28,15 +43,17 @@
 //! the intended baseline — deliberately simple, per the issue — is:
 //!
 //!   1. Serve lower urgency values first (u=0 before u=7).
-//!   2. Within one urgency band, streams marked incremental (`i`) are served
-//!      round-robin so they make forward progress together; non-incremental
-//!      streams in the band are served in stream-arrival (FIFO) order so a
-//!      complete response is delivered before the next begins.
-//!   3. Equal priority (same urgency, same incremental flag) is strictly
-//!      FIFO/round-robin and never reorders, which keeps it fair and
-//!      starvation-free.
+//!   2. Within one urgency band, share the connection across all ready
+//!      responses using a **bounded quantum / round-robin**, not run-to-
+//!      completion. RFC 9218 §10 warns that serving a non-incremental response
+//!      strictly to completion can starve its peers — a large or unbounded body
+//!      (e.g. a tunnel) would otherwise block everything else at that urgency.
+//!      Incremental (`i`) responses interleave naturally; non-incremental ones
+//!      each get a bounded turn so every response makes forward progress.
 //!
-//! No adaptive or dependency-tree scheduling. That is an explicit non-goal.
+//! This module makes no starvation-freedom guarantee on its own — fairness is
+//! the scheduler's responsibility via that bounded quantum. No adaptive or
+//! dependency-tree scheduling; those are explicit non-goals.
 
 const std = @import("std");
 
@@ -44,7 +61,7 @@ const varint = @import("quic_varint");
 const frame = @import("frame.zig");
 
 // ---------------------------------------------------------------------------
-// Priority field value model (RFC 9218 §4)
+// Priority field value model (RFC 9218 §4, §8)
 // ---------------------------------------------------------------------------
 
 pub const urgency_min: u8 = 0;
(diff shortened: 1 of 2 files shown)


src/http3/priority.zig
```diff
@@ -0,0 +1,547 @@
...
+    pub const default: Priority = .{};
+
+    pub fn eql(self: Priority, other: Priority) bool {
+        return self.urgency == other.urgency and self.incremental == other.incremental;
+    }
+
+    /// Serialize to a canonical RFC 8941 Dictionary value, omitting parameters
+    /// that are at their default (so both-default yields ""). Booleans use the
+    /// bare `i` form. Round-trips through `parse`. A 16-byte buffer is plenty.
+    pub fn serialize(self: Priority, out: []u8) error{BufferTooShort}![]u8 {
```
Jaruso: **Blocking — preserve member presence / response semantics.** This serializer is only correct for requests and `PRIORITY_UPDATE`. RFC 9218 §8 gives response headers different omission semantics: an omitted parameter means “do not change the client-provided value,” not “use the default.” With this API, a response cannot explicitly override `u=5` back to `u=3`, because `u=3` serializes to an empty value. The parser also discards whether `u`/`i` were present, so a response containing only `u=1` becomes indistinguishable from one that explicitly set `i=?0`. Please preserve presence (for example optional members / `has_u` and `has_i`) or make request/update and response serialization separate APIs.

src/http3/priority.zig
```diff
@@ -0,0 +1,547 @@
...
+            val = try parseValue(s, &i);
+        }
+
+        if (std.mem.eql(u8, key, "u")) {
+            if (seen_u) result.duplicate_parameter = true;
+            seen_u = true;
+            switch (val) {
+                .integer => |n| {
+                    if (n < urgency_min or n > urgency_max) {
+                        result.invalid_parameter = true;
```
Jaruso: **Blocking — the last duplicate is not actually winning when it is invalid.** RFC 8941 first overwrites duplicate Dictionary keys, then RFC 9218 ignores a resulting known member whose value is out of range or the wrong type. For `u=1, u=8`, the effective Dictionary contains only `u=8`, so urgency should fall back to `3`; this code leaves the earlier `1` active. Likewise, `i, i=5` leaves `true` instead of falling back to `false`. Reset the known member to its default on every occurrence before applying a valid value, or retain only the final parsed member and validate afterward. Please add both regression cases.

src/http3/priority.zig
```diff
@@ -0,0 +1,547 @@
...
+        if (i >= s.len) return error.MalformedPriority; // trailing comma
+    }
+
+    return result;
+}
+
+// ---------------------------------------------------------------------------
+// Minimal RFC 8941 bare-item value parsing (enough for u, i, and to skip
+// unknown parameter values). Inner lists and member parameters are not used by
+// the priority field and are treated as malformed.
```
Jaruso: **Blocking — valid Structured Fields syntax is being rejected.** RFC 9218 requires parsing the full RFC 8941 Dictionary first, then ignoring unknown members and known members with unexpected types. Decimals, byte sequences, inner lists, and item parameters are valid Dictionary syntax; treating them as `MalformedPriority` breaks forward compatibility. For example, `u=1.5` should successfully parse as a Decimal and yield default urgency with `invalid_parameter=true`, while `x=(1 2), u=5` should ignore `x` and retain `u=5`. The existing test that expects `u=1.5` to hard-fail is asserting the opposite of RFC 9218 §4. A complete SF parser, or a skip routine covering every Item/Inner List/parameter form, is needed here.

src/http3/priority.zig
```diff
@@ -0,0 +1,547 @@
...
+//! the intended baseline — deliberately simple, per the issue — is:
+//!
+//!   1. Serve lower urgency values first (u=0 before u=7).
+//!   2. Within one urgency band, streams marked incremental (`i`) are served
+//!      round-robin so they make forward progress together; non-incremental
+//!      streams in the band are served in stream-arrival (FIFO) order so a
+//!      complete response is delivered before the next begins.
+//!   3. Equal priority (same urgency, same incremental flag) is strictly
+//!      FIFO/round-robin and never reorders, which keeps it fair and
+//!      starvation-free.
```
Jaruso: This policy is not inherently starvation-free. A FIFO-to-completion non-incremental response can be very large or unbounded (for example a tunnel), and RFC 9218 §10 explicitly warns that strict same-urgency scheduling can starve other incremental or non-incremental responses. Since #254 requires equal-urgency scheduling to be fair enough to avoid starvation, please either remove this guarantee and leave the mixed-mode policy open for the scheduler, or document a bounded quantum/preemption rule that actually provides progress.

Jaruso (commented):
The frame-type and payload helpers look straightforward, but the current model/parser has several RFC 8941/9218 compatibility problems that will produce incorrect priority semantics. I’ve left inline comments with concrete examples and suggested regression cases.

src/http3/priority.zig
Jaruso: Fixed in 6ec89a1. Split the model into a scheduler view and a wire view: `Priority` (defaults applied, `u3`+`bool`) for the scheduler, and `Field { urgency: ?u3, incremental: ?bool }` for the wire, which preserves presence. `Field.serialize` emits only present members, so a response can send `u=3` to reset urgency to the default (`Field{ .urgency = 3 }` → `"u=3"`), and a parsed `u=5`-only field has `incremental == null`, distinct from one that set `i=?0`. `Field.effective()` bridges to `Priority`. Added round-trip tests over presence (including explicit `u=3` and `i=?0`).

---

src/http3/priority.zig
Jaruso: Fixed in 6ec89a1. Reworked parsing to the correct RFC 8941→9218 order: parse the whole Dictionary keeping only the *last* occurrence of each key, then validate the survivor. So `u=1, u=8` now falls back to urgency 3 (last value `8` is out of range → ignored), and `i, i=5` falls back to incremental false. Both regression cases added, asserting `duplicate_parameter` and `invalid_parameter` are set and the effective value is the default.

---

src/http3/priority.zig
Jaruso: Fixed in 6ec89a1. Replaced the ad-hoc value reader with a proper RFC 8941 value parser that handles Integers, Decimals, Strings, Tokens, Byte Sequences, Booleans, Inner Lists, and Item/bare-key parameters — enough to advance correctly past any valid member so unknown members are skipped rather than rejected. Known members are then type/range-checked: `u=1.5` parses as a Decimal and folds to default urgency with `invalid_parameter=true` (no longer a hard error — that test was flipped), and `x=(1 2), u=5` ignores `x` and keeps `u=5`. Genuinely malformed syntax (unterminated string/byte-sequence/inner-list, `u=1.`, etc.) still returns `MalformedPriority`.

---

src/http3/priority.zig
Jaruso: Reworded in 6ec89a1. Dropped the FIFO-to-completion "starvation-free" claim and now document the RFC 9218 §10 hazard explicitly (a non-incremental response served to completion — e.g. a tunnel — can starve peers at the same urgency), with the baseline requiring a **bounded quantum / round-robin** across ready same-urgency responses so each makes forward progress. The module states it makes no starvation-freedom guarantee on its own; fairness is the scheduler's responsibility via that quantum. Left the mixed-mode specifics open for the scheduler follow-up rather than over-specifying here.

---

PR #380, opened 2026-07-10
Extract protocol-neutral TLS core

Jaruso:
### Motivation

- Make TLS 1.3 state and key-schedule reusable across QUIC and a future record-mode by separating protocol-neutral pieces from QUIC-specific code.
- Reduce duplication by moving the SHA-256 TLS 1.3 key schedule out of the QUIC backend into a standalone core that depends only on `std.crypto`.
- Clarify ownership boundaries in the design docs so the QUIC adapter continues to own CRYPTO-frame transport and transport-parameter handling while reusable TLS state lives in a neutral core.

### Description

- Add a protocol-neutral TLS core under `src/tls/` including `state.zig`, `events.zig`, `key_schedule.zig`, `engine.zig`, and `root.zig` that declare roles, transport modes, event types, a key schedule, and a small engine shell.
- Extract the TLS 1.3 SHA-256 key schedule into `src/tls/key_schedule.zig` and add a lightweight record-mode instantiation test in that module.
- Rewire the QUIC TLS backend (`src/quic/tls_backend.zig`) to import and use `tls_key_schedule.KeySchedule` and remove the duplicated key-schedule implementation from that file.
- Export the TLS core through `src/quic/root.zig` (`pub const tls_core = @import("../tls/root.zig");`) and update `docs/QUIC_TLS.md` to describe the new `src/tls/` ownership boundary and the record-mode path.

### Testing

- Ran `git diff --check` to validate whitespace and simple issues, which completed successfully.
- Attempted to run the Zig unit tests with `zig build test-quic`, but the command could not be executed in this environment because `zig` is not installed (test run failed for that reason).
- New module-level `test` cases were added (in `src/tls/key_schedule.zig` and `src/tls/engine.zig`) and will run under a normal Zig-enabled CI or local developer environment via `zig build test-quic`.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a516dfb487483279cea9d4f50747b56)

### DIFF
--- a/src/quic/tls_backend.zig
+++ b/src/quic/tls_backend.zig
@@ -22,12 +22,9 @@ const config = @import("config.zig");
 const varint = @import("quic_varint");
 const tls_adapter = @import("tls_adapter.zig");
 const tls_handshake = @import("tls_handshake.zig");
+const tls_key_schedule = @import("../tls/key_schedule.zig");
 
 const crypto = std.crypto;
-const tls = crypto.tls;
-const Sha256 = crypto.hash.sha2.Sha256;
-const HmacSha256 = crypto.auth.hmac.sha2.HmacSha256;
-const HkdfSha256 = crypto.kdf.hkdf.HkdfSha256;
 const X25519 = crypto.dh.X25519;
 const Ed25519 = crypto.sign.Ed25519;
 const Certificate = crypto.Certificate;
@@ -39,7 +36,7 @@ const EventSink = tls_handshake.EventSink;
 const TlsBackend = tls_handshake.TlsBackend;
 const Role = tls_handshake.Role;
 
-pub const hash_len = Sha256.digest_length;
+pub const hash_len = tls_key_schedule.hash_len;
 /// Largest handshake message body we accept (u24 wire limit is 16 MiB; a
 /// single-certificate Ed25519 flight is far below this).
 pub const max_message_len = 8 * 1024;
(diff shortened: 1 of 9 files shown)


src/quic/tls_backend.zig
```diff
@@ -22,12 +22,9 @@ const config = @import("config.zig");
 const varint = @import("quic_varint");
 const tls_adapter = @import("tls_adapter.zig");
 const tls_handshake = @import("tls_handshake.zig");
+const tls_key_schedule = @import("../tls/key_schedule.zig");
 
 const crypto = std.crypto;
```
Jaruso, line 27: **Blocking — this leaves the backend uncompilable.** Removing the local `Sha256` alias does not remove its remaining uses: `Tls13Backend.transcript` is still declared as `Sha256 = Sha256.init(.{})`, and the transcript helpers also rely on that type. `zig build test-quic` will fail with an undeclared identifier before exercising the extracted schedule. Please retain/import the hasher here, or move transcript ownership into the shared core as part of the extraction.

src/tls/engine.zig
```diff
@@ -0,0 +1,43 @@
...
+pub const key_schedule = @import("key_schedule.zig");
+
+pub const EngineConfig = struct {
+    role: state.Role,
+    transport_mode: state.TransportMode,
+};
+
+pub const Engine = struct {
+    config: EngineConfig,
+    handshake_state: state.HandshakeState = .idle,
```
Jaruso, line 20: **Blocking — this is a parallel placeholder, not the engine extraction required by #329.** `Engine` only stores two enum values and changes one state in `start`; it has no transcript, handshake-message processing, key schedule, event emission, secret discard lifecycle, error path, or deinitialization. Meanwhile `src/quic/tls_backend.zig` still owns the complete client/server state machine, parsing, transcript, and secret transitions, and no QUIC test goes through this type. Please move the real protocol state machine behind this core and reduce the QUIC file to the transport-parameter/CRYPTO adapter, with the existing handshake tests exercising the shared engine.

src/tls/key_schedule.zig
```diff
@@ -0,0 +1,83 @@
...
+const std = @import("std");
+
+const crypto = std.crypto;
+const tls = crypto.tls;
+const Sha256 = crypto.hash.sha2.Sha256;
+const HmacSha256 = crypto.auth.hmac.sha2.HmacSha256;
+const HkdfSha256 = crypto.kdf.hkdf.HkdfSha256;
+const X25519 = crypto.dh.X25519;
+
+pub const hash_len = Sha256.digest_length;
```
Jaruso, line 17: **Blocking architecture leak — the supposedly reusable key schedule is coupled to X25519.** TLS 1.3 HKDF consumes the negotiated (EC)DHE shared secret; its length and production belong to the crypto-provider/key-exchange boundary, not to the key schedule. Hard-coding `[X25519.shared_length]u8` prevents reuse with another supported group or provider and conflicts with #329's requirement to consume cryptography through the #327 boundary. Accept provider-produced secret bytes (or a provider-owned secret type) and remove the `X25519` dependency from this module.

src/tls/events.zig
```diff
@@ -0,0 +1,14 @@
+//! Protocol-neutral events emitted by the TLS 1.3 core.
+
+pub const EncryptionEpoch = enum { initial, handshake, application };
+pub const SecretDirection = enum { read, write };
+pub const CertificateState = enum { not_checked, valid, invalid };
```
Jaruso, line 5: This leaks QUIC terminology into the protocol-neutral API. `initial` is a QUIC encryption level/packet-number space; TLS-over-records sends the opening handshake in plaintext, and TLS names the optional pre-handshake protected epoch `early_data`. A record-mode adapter should not have to reinterpret a QUIC-specific enum. Please model TLS epochs as `plaintext` / `handshake` / `application` (adding `early_data` when supported) and map `plaintext` to QUIC Initial in the QUIC adapter.

Jaruso (commented):
I found several blocking issues in the extraction. The current branch has a compile break, and the new modules do not yet move the actual TLS handshake engine or establish the protocol/provider boundaries required by #329. Details are inline.

PR #382, opened 2026-07-10
pki: add bounded ASN.1 DER decoder foundation for X.509 (#339)

Jaruso:
Implement strict DER-only parsing with configurable limits, zero-copy views, typed decoders, adversarial tests, and a fuzz harness entrypoint for later PKI work (#340, #341).

**What changed and why**
Adds a bounded ASN.1 DER decoder foundation for X.509: TLV cursoring, typed primitive decoders, OID/time helpers, canonical DER rejection cases, parser limits, and fuzz/regression coverage for untrusted certificate input.

**Related issues**
Closes #339

**Tests**
Added pure-Zig PKI DER unit tests covering canonical X.509 building blocks, malformed tags/lengths, tree-wide element limits, DER primitive/constructed form rejection, OID round trips including multi-octet first subidentifiers, strict certificate time forms, diagnostic offsets, and fuzz corpus handling.

Results from this branch:
- `zig fmt build.zig src/ tests/ --check` passes
- `git diff --check` passes
- `zig build test-pki --summary all --error-style verbose` passes: 33/33 tests
- `zig build test --summary all --error-style verbose` passes: 1111/1112 tests, 1 skipped
- `zig build --summary all --error-style verbose` passes
- `zig build test-integration --summary all --error-style verbose` passes: 63/67 tests, 4 skipped

**Security impact**
Yes. This parser sits on the untrusted certificate boundary for later PEM and X.509 work. Review focused on strict DER canonicality, bounded recursion/element/length limits, zero-copy lifetime behavior, malformed-input diagnostics, and rejecting BER/non-canonical encodings before downstream certificate parsing relies on them.

**Performance impact**
Low. The parser is zero-copy and allocation-free on the decode path; added checks are constant-time per tag/length/value boundary and enforce existing configurable limits.

**Checklist**

### DIFF
--- a/src/pki/der.zig
+++ b/src/pki/der.zig
@@ -1,0 +1,604 @@
+//! Bounded ASN.1 DER decoder for X.509 (#339).
+//!
+//! ## Supported subset
+//!
+//! Universal types used by X.509: SEQUENCE, SET, INTEGER, BOOLEAN, BIT STRING,
+//! OCTET STRING, NULL, OBJECT IDENTIFIER, UTF8String, PrintableString, IA5String,
+//! BMPString, UTCTime, GeneralizedTime; plus context-specific explicit and
+//! implicit tags.
+//!
+//! ## DER-only policy
+//!
+//! Definite lengths only. BER indefinite lengths, non-minimal length encodings,
+//! and non-minimal INTEGER encodings are rejected. Malformed input is never
+//! silently normalized.
+//!
+//! ## Default limits
+//!
+//! See `Limits`: 32 nesting depth, 1 MiB per element, 4096 elements per reader
+//! tree, 32 OID components, 4096 integer bytes.
+//!
+//! ## Ownership and zero-copy
+//!
+//! `Reader` and `Element` hold borrowed slices into the caller-supplied input.
+//! The caller owns the backing buffer for the parser lifetime. Typed decoders
+//! return views (`IntegerView`, `BitStringView`, string slices) without copying
+//! untrusted content. Never allocate directly from an unchecked attacker length.
+//!
+//! ## Complete consumption
+//!
+//! Call `expectEnd` after parsing a bounded region to reject trailing bytes.
+//! Child readers from `readSequence` / `readSet` enforce their own boundaries.
+//!
+//! ## Intentionally unsupported
+//!
+//! BER indefinite length, SET OF sorting requirements, REAL, EXTERNAL, EMBEDDED
+//! PDV, unrestricted ANY, and non-DER string encodings.
+//!
+//! ## Consumers
+//!
+//! #340 (PEM/chain loading) should wrap DER blobs in `Reader` and enforce
+//! `expectEnd`. #341 (X.509 certificate model) should walk TBSCertificate and
+//! extension sequences via child readers and typed decoders defined here.
+
+const std = @import("std");
+const oid_mod = @import("oid.zig");
+const time_mod = @import("time.zig");
+
+pub const oid = oid_mod;
+pub const time = time_mod;
+
+/// Configurable parser resource bounds.
+pub const Limits = struct {
+    max_depth: usize = 32,
+    max_element_len: usize = 1024 * 1024,
+    max_elements: usize = 4096,
+    max_oid_components: usize = 32,
+    max_integer_bytes: usize = 4096,
+};
+
+pub const default_limits: Limits = .{};
+
+pub const TagClass = enum(u2) {
+    universal = 0,
+    application = 1,
+    context_specific = 2,
+    private = 3,
+};
+
+pub const UniversalTag = enum(u32) {
+    end_of_content = 0,
+    boolean = 1,
+    integer = 2,
+    bit_string = 3,
+    octet_string = 4,
+    null = 5,
+    object_identifier = 6,
+    utf8_string = 12,
+    sequence = 16,
+    set = 17,
+    printable_string = 19,
+    ia5_string = 22,
+    utc_time = 23,
+    generalized_time = 24,
+    bmp_string = 30,
+};
+
+pub const Tag = struct {
+    class: TagClass,
+    number: u32,
+    constructed: bool,
+
+    pub fn universal(number: u32, constructed: bool) Tag {
+        return .{ .class = .universal, .number = number, .constructed = constructed };
+    }
+
+    pub fn contextSpecific(number: u32, constructed: bool) Tag {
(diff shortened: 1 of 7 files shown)


src/pki/der.zig
```diff
@@ -0,0 +1,604 @@
...
+            return error.UnexpectedTag;
+        }
+        return self.childReader(elem.content_offset, elem.content.len);
+    }
+
+    pub fn childReader(self: *const Reader, content_offset: usize, content_len: usize) Error!Reader {
+        if (self.depth >= self.limits.max_depth) return error.NestingLimit;
+        const end = content_offset + content_len;
+        if (content_offset < self.start or end > self.end) return error.LengthBeyondInput;
+        return Reader.initBounded(self.input, content_offset, end, self.depth + 1, self.element_count, self.limits);
```
Jaruso: **Blocking — `max_elements` is not a tree-wide limit.** Every child reader receives a copy of the current counter, increments that private copy, and never propagates it back to its parent or siblings. An input can therefore distribute elements across many sibling SEQUENCE/SET/explicit wrappers and exceed the advertised 4096-element budget while each child stays below it. Please put the count in a shared caller-owned parse budget/context referenced by every reader (no heap allocation is needed), or otherwise return/merge the consumed count. Add a regression with multiple nested sibling containers whose aggregate exceeds the limit.

src/pki/der.zig
```diff
@@ -0,0 +1,604 @@
...
+        return .{ .class = class, .number = number, .constructed = constructed };
+    }
+
+    fn readDefiniteLength(self: *Reader) Error!usize {
+        if (self.remaining() == 0) return error.Truncated;
+        const first = self.input[self.offset];
+        self.offset += 1;
+
+        if (first & 0x80 == 0) {
+            return first;
```
Jaruso, line 414: **Blocking — short-form lengths bypass `max_element_len`.** The configured limit is only checked in the long-form branch below. For example, with `.max_element_len = 4`, an OCTET STRING encoded with short-form length 5 is accepted. Apply the same limit before returning `first`, and add a low-limit short-form regression case.

src/pki/der.zig
```diff
@@ -0,0 +1,604 @@
...
+        const wrapper = try self.readContextSpecific(number, true);
+        var inner = try self.childReader(wrapper.content_offset, wrapper.content.len);
+        const elem = try inner.readElement();
+        try inner.expectEnd();
+        return elem;
+    }
+
+    pub fn readInteger(self: *Reader) Error!IntegerView {
+        const elem = try self.readElement();
+        if (elem.tag.class != .universal or elem.tag.number != @intFromEnum(UniversalTag.integer)) {
```
Jaruso: **Blocking DER canonicality / tag-confusion issue.** This checks class and number but not the primitive/constructed bit, so a constructed INTEGER (`0x22`) is accepted as an INTEGER. The same pattern exists for BOOLEAN, NULL, BIT/OCTET STRING, OID, strings, and times. DER fixes the encoding form for these universal types; constructed encodings are BER, not DER. Please centralize universal-tag form validation (and reject the universal EOC tag in DER) or add `!elem.tag.constructed` to every primitive decoder. Add adversarial cases for constructed primitive tags.

src/pki/der.zig
```diff
@@ -0,0 +1,604 @@
...
+    MalformedBitString,
+    MalformedOid,
+    MalformedString,
+    MalformedTime,
+    TrailingData,
+} || oid_mod.Error || time_mod.Error;
+
+pub const ParseError = struct {
+    err: Error,
+    offset: usize,
```
Jaruso, line 159: The issue requires typed failures **with offsets**, but `ParseError` is currently dead API: all public parsing methods return bare `Error`, and the cursor may already have advanced through part of a tag/length when the error is raised. Please expose a diagnostic parsing entrypoint/result that captures the exact failure offset (tag start vs. offending byte should be documented), then add assertions for malformed tag, truncated length, non-minimal length, and child-boundary offsets.

src/pki/oid.zig
```diff
@@ -0,0 +1,161 @@
...
+/// Decode DER OID content (value bytes only, not the TLV wrapper).
+pub fn decode(content: []const u8, component_limit: usize) Error!ObjectIdentifier {
+    if (content.len == 0) return error.MalformedOid;
+    if (component_limit > max_components) return error.OidComponentLimit;
+
+    var oid: ObjectIdentifier = .{};
+
+    const first = content[0];
+    const first_arc: u32 = if (first < 40) 0 else if (first < 80) 1 else 2;
+    const second_arc: u32 = if (first < 40) first else if (first < 80) first - 40 else first - 80;
```
Jaruso, line 49: **Blocking — the first OID subidentifier is a base-128 value, not always one byte.** For `2.999.3`, the first combined value (`40*2+999`) is encoded across multiple octets; this decoder treats only the first octet as arcs 0/1 and starts the next component at byte 1, while `encodeComponents` later casts the combined value directly to `u8`. Decode the first subidentifier with the same canonical base-128 routine and split it into arcs afterward; encode the combined value with base-128 as well. Also reject `component_limit < 2` before writing the mandatory first two arcs. Please add round trips such as `2.999.3` plus limits 0 and 1.

src/pki/time.zig
```diff
@@ -0,0 +1,125 @@
...
+    day: u8,
+    hour: u8,
+    minute: u8,
+    second: ?u8,
+};
+
+pub fn parseUtcTime(content: []const u8) Error!UtcTime {
+    if (content.len != 11 and content.len != 13) return error.MalformedTime;
+    if (content[content.len - 1] != 'Z') return error.MalformedTime;
+
```
Jaruso, line 30: **Blocking for strict DER/X.509:** seconds are mandatory in DER UTCTime and in the RFC 5280 certificate time profile. Accepting 11-byte UTCTime and 13-byte GeneralizedTime admits non-canonical values such as `2401010000Z`. Require the exact certificate forms `YYMMDDHHMMSSZ` and `YYYYMMDDHHMMSSZ`, make `second` non-optional, and add explicit rejection tests for the no-seconds variants.

Jaruso (commented):
The shape of the zero-copy API and the adversarial test suite are a good start, and the unit-test jobs are currently green. Before merge, I want the resource limits and DER canonicality guarantees to hold across the full parse tree, not just the covered happy paths. I left six inline comments with concrete regression cases. Please also replace the remaining PR-template placeholders with the actual test commands/results and a security-impact summary; this parser will sit directly on an untrusted certificate boundary.

PR #383, opened 2026-07-11
tls: add handshake message codec and transcript module

Jaruso:
## Summary
- add a transport-neutral TLS handshake message codec and bounded reassembler that preserve exact transcript bytes
- add reader/writer helpers plus duplicate-extension guarding for shared TLS handshake parsing
- add a transcript hash helper with snapshot and HelloRetryRequest-style replacement support
- add `zig build test-tls` and wire the TLS core tests into the default test step

## Tests
- `zig fmt build.zig src/ tests/ --check`
- `git diff --check`

Closes #330

### DIFF
--- a/src/tls/messages.zig
+++ b/src/tls/messages.zig
@@ -1,0 +1,266 @@
+//! Bounded TLS 1.3 handshake message codec.
+//!
+//! This module owns only TLS handshake framing and small wire helpers. It does
+//! not know about QUIC CRYPTO frames, TLS records, sockets, HTTP, or certificate
+//! policy. Callers feed complete or fragmented handshake bytes from their
+//! transport and update the transcript with `HandshakeMessage.raw`.
+
+const std = @import("std");
+
+pub const max_message_len = 16 * 1024 * 1024 - 1;
+
+pub const Error = error{
+    MalformedHandshake,
+    HandshakeBufferOverflow,
+    MessageTooLarge,
+    DuplicateExtension,
+    TooManyExtensions,
+};
+
+pub const MessageType = enum(u8) {
+    client_hello = 1,
+    server_hello = 2,
+    new_session_ticket = 4,
+    encrypted_extensions = 8,
+    certificate = 11,
+    certificate_verify = 15,
+    finished = 20,
+};
+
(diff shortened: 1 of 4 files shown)


src/tls/transcript.zig
```diff
@@ -0,0 +1,46 @@
...
+
+    pub fn peek(self: *const Transcript) [digest_len]u8 {
+        var copy = self.hash;
+        var out: [digest_len]u8 = undefined;
+        copy.final(&out);
+        return out;
+    }
+
+    pub fn replace(self: *Transcript, digest: [digest_len]u8) void {
+        self.hash = Hash.init(.{});
```
Jaruso, line 33: This does not implement the TLS 1.3 HelloRetryRequest transcript transformation. RFC 8446 replaces ClientHello1 with a synthetic handshake message: `message_hash(254) || uint24(Hash.length) || Hash(ClientHello1)`. Hashing the digest alone produces a different transcript and therefore incorrect handshake secrets/Finished verification after HRR. Please prepend the 4-byte handshake header here and update the test to assert the RFC construction.

src/tls/messages.zig
```diff
@@ -0,0 +1,266 @@
...
+
+pub const MessageType = enum(u8) {
+    client_hello = 1,
+    server_hello = 2,
+    new_session_ticket = 4,
+    encrypted_extensions = 8,
+    certificate = 11,
+    certificate_verify = 15,
+    finished = 20,
+};
```
Jaruso, line 32: Because `decode()` rejects any value absent from this enum, this transport-neutral TLS 1.3 codec currently rejects valid handshake messages such as `end_of_early_data` (5), `certificate_request` (13), `key_update` (24), and especially the synthetic `message_hash` (254) required for HelloRetryRequest. Please include the full TLS 1.3 handshake type set (or preserve unknown numeric types without treating them as malformed) so valid protocol messages are not rejected at the framing layer.

Jaruso (commented):
I found two TLS correctness issues worth fixing before merge. The transcript replacement does not implement RFC 8446's synthetic `message_hash` construction, and the codec currently rejects several valid TLS 1.3 handshake message types—including `message_hash` itself.

Issue #247, opened 2026-07-02
http3: close out native QUIC/H3 validation, soak, and release evidence


Jaruso:
## Role

Final **validation / assurance / release-evidence umbrella** for Tardigrade's native pure-Zig QUIC + HTTP/3 stack.

The stack is built. This issue does **not** own another QUIC transport, H3/QPACK engine, interop framework, fuzz framework, observability system, or benchmark harness.

Its remaining job is to:

1. reconcile release-critical failure/lifecycle invariants against proof that already exists;
2. run bounded production H3 soak/resource-stability validation;
3. execute the existing H3 benchmark/evidence harness with a genuinely QUIC-capable client on suitable hardware;
4. rerun required external interop and publish the closeout evidence;
5. fix/rerun only concrete defects the evidence exposes.

#247 supplies the final foundation evidence consumed by #240 and the H3 stable-promotion evidence consumed by #389.

---

# Reconciled status — 2026-08-15

The old implementation dependencies are complete:

- [x] #537 — deterministic QUIC/H3 fuzz/property matrix.
- [x] #257 — production Alt-Svc, reload, H3 GOAWAY/drain and listener lifecycle.
- [x] #255 — production qlog/keylog/Prometheus observability and opt-in artifact plumbing.
- [x] #256 — production UDP/QUIC hardening plus H3 benchmark/transport-evidence harness.
- [x] PR #624 — interop qlog/keylog artifact retention wiring.

This leaves **execution and integrated assurance**, not architecture construction.

## Important: #256 closed does not mean real H3 benchmark numbers already exist

PR #623 completed the benchmark **machinery**:

- canonical H3 benchmark listener/fixtures;
- `static-small-http3`;
- `static-large-http3`;
- `proxy-large-http3`;
- QUIC/H3 transport metrics per scenario;
- UDP-buffer tuning comparison;
- Linux netem impairment helper;
- scheduled/manual H3 workflow;
- operator documentation and harness regression coverage.

During PR #623 closeout, the available `h2load` accepted `--h3` but was not actually QUIC-enabled and sent **zero UDP traffic**. The harness now rejects that false-positive class.

Therefore the remaining evidence must be produced with a **real QUIC-capable H3 client/load generator** on an environment that can actually exercise the transport.

A scheduled artifact reporting `supported:false` is useful harness-maintenance signal, not final performance evidence.

---

# Completed proof — do not recreate

## External interoperability foundation

Already present:

[...]

PR #384, opened 2026-07-11
Claude/native http3 cutover el0p73

Jaruso:
**What changed and why**

Cuts Tardigrade's HTTP/3 endpoint over to the native Zig QUIC/H3 stack (#328):
one background thread owns the UDP socket, routes datagrams to connections by
DCID, schedules wakeups from the drivers' timer deadlines, and bridges decoded
requests to the gateway's `RequestHandler`. ngtcp2/nghttp3 are removed from the
build and remain only as out-of-process interop peers under `scripts/interop/`.

Subsequent commits on this branch address the review:

- **H3 conformance (`src/http3/conn.zig`)** — server-initiated bidirectional
  streams, duplicate QPACK critical streams, and FIN on a critical stream now
  return typed errors and close the connection with the correct RFC 9114/9204
  code (`H3_STREAM_CREATION_ERROR`, `H3_CLOSED_CRITICAL_STREAM`). The runtime
  propagates the session's specific wire close code instead of collapsing every
  failure to `H3_GENERAL_PROTOCOL_ERROR`.
- **QUIC path validation (`src/quic/connection.zig`)** — incoming
  `PATH_RESPONSE` now completes challenge tracking and exposes an authenticated
  path event; `startPathValidation` is the building block for real migration
  (#251).
- **Runtime hardening (`src/http/http3_runtime.zig`)** — the peer address is no
  longer copied from an unauthenticated packet source; migration is detected via
  the post-AEAD `packets_received` delta and recorded rather than followed.
  Half-open state per spoofable Initial is bounded (global + per-source-IP caps,
  immediate teardown of an Initial that authenticates nothing, and a handshake
  deadline reaped before the idle timeout) as the accepted alternative to Retry.
  The runtime `Config` now reaches `Connection.init` via `quicConfigFrom`, and
  unsupported knobs (`connection_migration`, non-1.3 TLS bounds, 0-RTT) are
  warn-and-ignored at init.

**Related issues**
Relates to #328, #247, #251.

**Tests**
- `zig build test` unit tests including new coverage for `quicConfigFrom`
  clamping and the per-source admission counter.
- Real-UDP loopback smoke test drives the native client/server through a full
  H3 exchange (`tests/quic_h3_udp_smoke.zig`).
- Deterministic end-to-end harness with loss control for the native path.

**Security impact**
Yes — this change affects TLS identity loading, QUIC packet authentication,
address handling, and half-open admission. Reviewed: the peer address is never
promoted from an unauthenticated datagram; migration is not followed while
`disable_active_migration` is advertised; per-source and global caps plus a
handshake deadline bound the state an off-path spoofer can pin without Retry.

**Performance impact**
Yes — introduces the native QUIC/H3 event loop, replacing ngtcp2/nghttp3. One
thread per listener owns the socket, batches datagram ingest, and schedules
wakeups from driver timer deadlines to avoid busy-looping. Admission accounting
is O(1) per Initial (two hash-map lookups).

**Checklist**
- [x] `zig build test --summary all` green (the loopback smoke test requires a
  UDP socket and is skipped in sandboxed CI without raw socket access)
- [ ] `zig build test-integration` green (required for protocol, proxy, and TLS changes)
- [ ] README / CHANGELOG updated if operator-visible behavior changed


src/http/http3_runtime.zig
```diff
@@ -155,185 +197,445 @@ pub const Runtime = struct {
...
+        const entry = connections.get(found) orelse return;
 
-    fn noteMigration(self: *Runtime) void {
-        self.snapshot_mutex.lock();
-        defer self.snapshot_mutex.unlock();
-        self.snapshot_state.migration_events += 1;
+        const was_established = entry.conn.isEstablished();
+        entry.peer = peer; // reply to the packet's source (NAT rebinding)
+        entry.conn.ingest(datagram, now) catch return;
+        if (!was_established and entry.conn.isEstablished()) {
```
Jaruso, line 389: **P1 — Do not update the peer address before authenticating the packet.**

This assignment occurs before `Connection.ingest()` verifies packet protection. A spoofed UDP datagram with a known DCID can therefore change `entry.peer` even when decryption subsequently fails, redirecting later ACKs, PTO probes, responses, or close packets.

Keep the existing active peer until the datagram authenticates. A new address should go through QUIC path validation (`PATH_CHALLENGE` / `PATH_RESPONSE`) before promotion, and `connection_migration = false` should prevent this mutation entirely.

src/http/http3_runtime.zig
```diff
@@ -155,185 +197,445 @@ pub const Runtime = struct {
...
+        // need the connection's own CID length, which the client chose per
+        // connection — try each active length.
+        var handle: ?u64 = null;
+        if (datagram.len > 0 and datagram[0] & 0x80 != 0) {
+            if (quic.packet.parsePacket(datagram, 0)) |parsed| {
+                handle = routes.lookup(parsed.dcid);
+                if (handle == null and parsed.kind == .initial) {
+                    handle = self.accept(connections, routes, next_handle, parsed, peer, now);
+                }
+                if (parsed.kind == .zero_rtt) self.noteZeroRtt();
```
Jaruso, line 345: **P1 — Avoid allocating an unbounded full connection for every parseable Initial.**

An unknown Initial reaches `accept()` before packet authentication or proof of address ownership. `accept()` then allocates TLS, QUIC, H3, routing-table, and connection-map state. An attacker can send Initials with unique DCIDs and retain large amounts of state until timeout.

Please add stateless Retry/token validation or a strictly bounded half-open admission layer, plus per-source/global limits and aggressive cleanup when the first Initial fails authentication.

src/http3/conn.zig
```diff
@@ -0,0 +1,567 @@
...
+                        state.typ = decoded.typ;
+                        // The control-stream view consumes its own stream-type
+                        // varint; strip it only for the other types.
+                        if (decoded.typ != .control) bytes = bytes[decoded.len..];
+                        switch (decoded.typ) {
+                            .control => {
+                                if (self.peer_control != null) return error.ProtocolError;
+                                self.peer_control = id;
+                            },
+                            .qpack_encoder => self.peer_qpack_encoder = id,
```
Jaruso: **P2 — Reject duplicate peer QPACK critical streams.**

Unlike the control-stream branch, this silently replaces an existing QPACK encoder stream ID. HTTP/3 permits only one peer control stream and one of each QPACK encoder/decoder stream; opening a duplicate critical stream must close the connection with `H3_STREAM_CREATION_ERROR`. Please guard both QPACK assignments when the field is already non-null.

Jaruso (commented):
I found two security/reliability blockers in the UDP connection admission and path-handling logic, plus a protocol-compliance issue around duplicate critical HTTP/3 streams.

The PR description is also still the unfilled template, and there are currently no reported commit statuses. Given the size and risk of this change, the test matrix and security/performance sections should be completed before merge.

src/quic/connection.zig
```diff
@@ -1,19 +1,2136 @@
...
+            .retire_connection_id => {
+                // We never issue additional CIDs, so there is nothing to
+                // retire; tolerate the frame (it can only name sequence 0).
+            },
+            .path_challenge => |data| {
+                if (level == .application) {
+                    try self.pending_path_responses.append(self.allocator, data);
+                }
+            },
+            .path_response => {},
```
Jaruso: **P1 — Complete path validation instead of ignoring `PATH_RESPONSE`.**

The stack queues a response when it receives `PATH_CHALLENGE`, but an incoming `PATH_RESPONSE` is a no-op. That means there is no visible state transition that marks a candidate path validated, preserves/reverts the prior validated path, or resets congestion and RTT state after migration.

This is especially important because the runtime currently changes the socket peer address independently. Please add explicit challenge tracking and validation completion here, then expose an authenticated path-migration event to the runtime.

src/http3/conn.zig
```diff
@@ -0,0 +1,567 @@
...
+                    request.* = .{ .stream = session.RequestStream.init(self.allocator, id) };
+                    self.requests.put(id, request) catch {
+                        request.stream.deinit();
+                        self.allocator.destroy(request);
+                        return error.OutOfMemory;
+                    };
+                } else {
+                    // A server must not open bidirectional streams (RFC 9114
+                    // §6.1); flag and ignore.
+                    self.metrics.unknown_uni_streams += 1;
```
Jaruso: **P1 — A server-initiated bidirectional stream must be a connection error.**

For an HTTP/3 client, RFC 9114 §6.1 requires a server-initiated bidirectional stream to produce `H3_STREAM_CREATION_ERROR` unless an extension negotiated that behavior. Incrementing a metric and ignoring the stream leaves the connection in a non-compliant state.

Please return a typed stream-creation error here and close the connection with the corresponding H3 code.

src/http3/conn.zig
```diff
@@ -0,0 +1,567 @@
...
+                        }
+                    }
+                    if (bytes.len > 0 and state.typ == .control) {
+                        _ = self.peer_control_view.ingest(self.allocator, bytes) catch return error.ProtocolError;
+                        if (self.peer_control_view.saw_settings) self.metrics.settings_received = true;
+                    }
+                    // QPACK encoder/decoder instructions: with a zero-capacity
+                    // dynamic table the peer sends none that affect state;
+                    // unknown streams are drained and dropped.
+                    if (result.fin) break;
```
Jaruso: **P1 — Closing a critical HTTP/3 stream must terminate the connection.**

`result.fin` is accepted for every unidirectional stream. FIN on the peer control stream must produce `H3_CLOSED_CRITICAL_STREAM`, and the same applies to the QPACK encoder and decoder streams under RFC 9204.

Please distinguish critical stream types here and return a typed close error instead of silently accepting their closure.

src/http/http3_runtime.zig
```diff
@@ -155,185 +198,445 @@ pub const Runtime = struct {
...
+    fn pumpH3(self: *Runtime, entry: *ConnEntry, now: u64) void {
+        if (!entry.conn.isEstablished()) return;
+        if (!entry.h3_started) {
+            entry.h3.start(entry.conn) catch return;
+            entry.h3_started = true;
+        }
+        entry.h3.pump(entry.conn) catch |err| {
+            // An H3-level protocol error closes the connection (RFC 9114 §8).
+            self.logger.warn(null, "http3: session error {s}; closing connection", .{@errorName(err)});
+            entry.conn.close(0x0101, "h3 protocol error", now); // H3_GENERAL_PROTOCOL_ERROR
```
Jaruso: **P2 — Preserve protocol-specific HTTP/3 error codes.**

Every H3 failure is currently collapsed to `H3_GENERAL_PROTOCOL_ERROR` (`0x0101`). That loses required distinctions such as `H3_STREAM_CREATION_ERROR`, `H3_CLOSED_CRITICAL_STREAM`, `H3_FRAME_UNEXPECTED`, and `H3_MISSING_SETTINGS`, and it will make interoperability failures much harder to diagnose.

Please have `http3/conn.zig` return a typed error carrying the wire close code and use that code here.

src/http/http3_runtime.zig
```diff
@@ -155,185 +198,445 @@ pub const Runtime = struct {
...
+
+        const backend = allocator.create(quic.tls_backend.Tls13Backend) catch return null;
+        var entropy: quic.tls_backend.Entropy = undefined;
+        compat.randomBytes(&entropy.hello_random);
+        compat.randomBytes(&entropy.key_share_seed);
+        backend.* = quic.tls_backend.Tls13Backend.initServer(entropy, identity);
+
+        const conn = Connection.init(allocator, .{
+            .role = .server,
+            .local_cid = parsed.dcid,
```
Jaruso, line 429: **P2 — The runtime configuration is not being passed into the QUIC connection.**

`Connection.init()` is called without `.config`, so it receives the default `quic.Config` even though the runtime exposes settings for migration, datagram sizing, TLS bounds, and 0-RTT behavior. Some of those settings are security- or transport-significant, and silently ignoring them gives operators false assurance.

Please either construct and pass the intended QUIC config here, with tests proving each option is honored, or remove unsupported knobs from the public runtime config until they are implemented.

Jaruso (commented):
Deep-review follow-up: I’m adding the remaining protocol and maintainability findings that were not covered by the first three comments. The main unresolved areas are path-validation completion, HTTP/3 critical-stream enforcement, protocol-specific close codes, and configuration that is exposed but not actually plumbed into the native QUIC connection.

The current tests cover substantial happy-path and recovery behavior, but I would also require explicit tests for Retry/token admission, spoofed rebinding/migration, duplicate and closed critical streams, server-initiated bidirectional streams, and exact H3 close codes before treating this cutover as production-ready.

src/http/http3_runtime.zig
Jaruso: Fixed in 94693d5. The `entry.peer = peer` mutation is gone entirely — replies now always target the address of the Initial that opened the connection. A packet is treated as authenticated only when `entry.conn.metrics.packets_received` advances across `ingest()` (that counter increments only after AEAD open succeeds), so no unauthenticated datagram can influence peer state. An authenticated packet arriving from a new source bumps the `migration_events` counter instead of rebinding, since the native stack advertises `disable_active_migration`. `connection_migration = true` is now warn-and-ignored at init (same pattern as `enable_0rtt`). Real path-validation-based migration remains future work (#251), building on the `startPathValidation` API added in this PR.

src/http/http3_runtime.zig
Jaruso: Fixed in 94693d5 with the bounded half-open admission layer (rather than Retry). `accept()` now enforces a global connection cap (`max_connections = 1024`) and a per-source-IP cap (`max_connections_per_source = 32`, keyed on the IPv4 source) before allocating anything. A freshly accepted connection whose first datagram authenticates zero packets (same `packets_received` delta trick) is torn down immediately. Any half-open connection that misses a 10s handshake deadline (`accepted_at_us`, reaped in the timer loop) is dropped well before the 30s idle timeout. Per-source slots are released on every teardown path via the shared `removeConnection` helper.

src/http/http3_runtime.zig
Jaruso: Fixed in 94693d5. Added a testable `quicConfigFrom(cfg)` that maps `max_datagram_size` (clamped to [1200, 2048], the size of the work buffers) onto `max_udp_payload_size`, and `Connection.init()` now receives it via `.config`. The knobs the native stack does not honor are no longer silently ignored: `connection_migration = true` and any `tls_min/max_version` other than `1.3` are warn-and-ignored at init. Migration policy stays `.disabled`. A `quicConfigFrom clamps datagram size...` unit test covers the mapping.

src/http/http3_runtime.zig
Jaruso: Follow-up: the admission decision logic is now covered by explicit unit tests in `f22e869`. The global + per-source caps and the post-ingest disposition (drop a spoofed Initial that authenticates nothing / count a migration from a new source / keep) are extracted into pure `admissionAllowed` and `classifyIngest` functions with exhaustive table-driven tests, matching the sans-I/O design. End-to-end socket-driven flood/rebind tests and the broader migration/Retry listener integration are tracked in #387.

PR #385, opened 2026-07-11
tls: add data-driven negotiation registry

Jaruso:
## Summary
- add shared TLS 1.3 registries for protocol versions, cipher suites, named groups, signature schemes, extension ids, and ALPN names
- add policy/capability defaults for QUIC and record-mode negotiation
- add ClientHello offer parsing and policy-driven server/client selection helpers with deterministic diagnostics
- point the existing QUIC TLS backend constants at the shared registry while preserving current behavior

## Tests
- `zig fmt build.zig src/ tests/ --check`
- `git diff --check`

Closes #331

### DIFF
--- a/src/tls/negotiation.zig
+++ b/src/tls/negotiation.zig
@@ -16,9 +16,10 @@ pub const Error = messages.Error || error{
     NoMutualSignatureScheme,
     NoMutualAlpn,
     MissingServerName,
+    OfferVectorTooLarge,
 };
 
-pub const max_offers = 16;
+pub const max_offers = 64;
 
 pub const KeyShareOffer = struct {
     group: algorithms.NamedGroup,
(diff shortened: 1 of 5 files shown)


src/tls/policy.zig
```diff
@@ -0,0 +1,69 @@
...
+pub const CipherSuite = algorithms.CipherSuite;
+pub const NamedGroup = algorithms.NamedGroup;
+pub const SignatureScheme = algorithms.SignatureScheme;
+pub const ProtocolName = algorithms.ProtocolName;
+
+const default_cipher_suites = [_]CipherSuite{.tls_aes_128_gcm_sha256};
+const default_named_groups = [_]NamedGroup{.x25519};
+const default_signature_schemes = [_]SignatureScheme{.ed25519};
+const quic_alpns = [_]ProtocolName{algorithms.alpn.h3};
+const record_alpns = [_]ProtocolName{ algorithms.alpn.h2, algorithms.alpn.http_1_1 };
```
Jaruso, line 17: PR #384 adds ECDSA P-256 identities to the production QUIC TLS backend and selects `ecdsa_secp256r1_sha256` when that identity is configured. Keeping the shared/default capabilities Ed25519-only means this policy no longer describes the backend after #384, and `negotiateServer` could select Ed25519 for an ECDSA identity. Please derive signature capabilities from the configured identity/provider (or at minimum include P-256 and constrain selection to the active identity) so the negotiated scheme is one the server can actually use.

src/tls/negotiation.zig
```diff
@@ -0,0 +1,432 @@
...
+    NoMutualNamedGroup,
+    MissingKeyShare,
+    NoMutualSignatureScheme,
+    NoMutualAlpn,
+    MissingServerName,
+};
+
+pub const max_offers = 16;
+
+pub const KeyShareOffer = struct {
```
Jaruso, line 24: This hard limit turns an otherwise valid ClientHello into `MalformedExtension` once it contains more than 16 recognized entries in any vector. Signature-algorithm lists from general-purpose clients can legitimately be longer, and #384 is explicitly broadening external-client interoperability. Consider either retaining only supported/intersecting values without failing on excess entries, or using limits derived from the wire vector bounds with a dedicated resource-limit diagnostic rather than classifying the peer message as malformed.

Jaruso (commented):
I compared this against #384 because that PR changes the production QUIC TLS identity/signature behavior. The registry direction is good, but there are two integration issues to resolve before these abstractions can safely become the shared source of truth.

PR #390, opened 2026-07-11
tls: extract shared handshake vocabulary

Jaruso:
## Summary
- move shared TLS certificate state, secret direction, TLS-level handshake errors, and driver lifecycle vocabulary into the protocol-neutral tls_core package
- make the QUIC TLS adapter and handshake driver import the shared core vocabulary while keeping QUIC-specific failures local to the QUIC handshake seam
- document the shared QUIC/record-mode ownership boundary in docs/QUIC_TLS.md

Part of #329.

## Validation

### DIFF
--- a/src/tls/events.zig
+++ b/src/tls/events.zig
@@ -1,8 +1,33 @@
-//! Protocol-neutral events emitted by the TLS 1.3 core.
+//! Protocol-neutral TLS 1.3 event and failure vocabulary.
+//!
+//! The concrete transport integrations own their framing. QUIC maps these
+//! concepts to CRYPTO streams and packet-protection keys; future TCP record
+//! mode maps the same concepts to TLS records and encrypted byte streams.
 
-pub const EncryptionEpoch = enum { initial, handshake, application };
+pub const EncryptionEpoch = enum {
+    initial,
+    zero_rtt,
+    handshake,
+    application,
+};
+
 pub const SecretDirection = enum { read, write };
 pub const CertificateState = enum { not_checked, valid, invalid };
+
+/// Typed, deterministic TLS handshake failures. Transport layers translate
+/// these into their own close/error surfaces without losing the TLS reason.
+pub const HandshakeError = error{
+    /// The peer's TLS handshake bytes were malformed or arrived in an illegal
+    /// order.
+    MalformedHandshake,
+    /// ALPN did not negotiate an acceptable application protocol.
+    AlpnMismatch,
+    /// The peer certificate was rejected by local policy or failed proof of key
+    /// possession.
+    CertificateInvalid,
+    /// A traffic secret could not be exported or installed.
+    SecretExportFailed,
+};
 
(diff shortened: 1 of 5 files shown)


src/tls/events.zig
```diff
@@ -1,9 +1,43 @@
...
+    /// The peer's TLS handshake bytes were malformed or arrived in an illegal
+    /// order.
+    MalformedHandshake,
+    /// ALPN did not negotiate an acceptable application protocol.
+    AlpnMismatch,
+    /// The peer certificate was rejected by local policy or failed proof of key
+    /// possession.
+    CertificateInvalid,
+    /// The transport-parameter extension required by the caller was absent.
+    MissingTransportParameters,
```
Jaruso: **Blocking architectural boundary:** this shared error set still contains QUIC-driver concepts: `UnexpectedCryptoLevel`, `MissingTransportParameters`, `InvalidTransportParameters`, and arguably the driver's bounded `HandshakeBufferOverflow`. A future TLS record-layer engine has neither QUIC CRYPTO levels nor QUIC transport parameters, so this moves adapter-specific vocabulary into the protocol-neutral core and conflicts with #329's “no QUIC-specific types” criterion.

Could the core keep only genuinely TLS-level failures, while `src/quic/tls_handshake.zig` composes them with a QUIC-local error set (for example, `tls_core.events.HandshakeError || error{ ... }`)? That preserves the shared TLS reason without making record mode depend on QUIC semantics.

src/tls/state.zig
```diff
@@ -20,3 +20,18 @@ pub const HandshakeState = enum {
...
+    in_progress,
+    complete,
+    failed,
+};
+
+test "record mode starts from the shared idle lifecycle state" {
+    const std = @import("std");
+
+    try std.testing.expectEqual(TransportMode.record, TransportMode.record);
+    try std.testing.expectEqual(DriverState.idle, DriverState.idle);
```
Jaruso: These assertions compare enum constants directly to themselves, so they do not exercise initialization or a lifecycle transition and cannot catch a regression. The existing `engine.zig` test already instantiates record mode and checks its state after `start()`; could this test instead cover a real `DriverState` transition/default through a consumer, or simply be removed until there is shared behavior to test?

Jaruso (commented):
The type aliases themselves are low-risk, but I don't think this PR can close #329 yet.

#329 requires the protocol-neutral engine to own the reusable handshake/message flow, the existing QUIC tests to run through that engine, deterministic ownership/deinitialization coverage, and no QUIC-specific types in core TLS. In this PR, `src/tls/engine.zig` is still only a state shell and the concrete handshake implementation remains in `src/quic/tls_backend.zig`; this change primarily centralizes vocabulary.

Please either remove `Closes #329` and treat this as a preparatory vocabulary-extraction PR, or continue the migration until the issue's acceptance criteria are actually met. I also left inline comments on the QUIC-specific error vocabulary and the tautological state test.

PR #393, opened 2026-07-11
tls: move backend transport contract into core

Jaruso:
## Summary
- add a generic protocol-neutral TLS transport contract in src/tls/transport.zig
- instantiate that contract from the QUIC handshake driver while keeping QUIC CID binding hooks local
- move the backend event sink/vtable shape out of the QUIC module and document the ownership boundary

Part of #329. This intentionally avoids src/tls/alerts.zig / #332 work.

## Validation

### DIFF
--- a/src/tls/transport.zig
+++ b/src/tls/transport.zig
@@ -1,0 +1,172 @@
+//! Generic transport contract between a TLS 1.3 engine and its carrier.
+//!
+//! TLS does not own QUIC CRYPTO streams or TCP record buffering here. Callers
+//! instantiate this contract with their own epoch and transport-parameter
+//! payload types, then translate the emitted events into their local framing.
+
+const std = @import("std");
+const events = @import("events.zig");
+const state = @import("state.zig");
+
+pub fn Contract(
+    comptime TransportParameters: type,
+    comptime Epoch: type,
+    comptime ErrorSet: type,
+) type {
+    return ContractWithOptions(
+        TransportParameters,
+        Epoch,
+        ErrorSet,
+        16,
+        16 * 1024,
+        error.TransportBufferOverflow,
+    );
+}
+
+pub fn ContractWithOptions(
+    comptime TransportParameters: type,
+    comptime Epoch: type,
+    comptime ErrorSet: type,
+    comptime max_event_count: usize,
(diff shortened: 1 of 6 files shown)


src/tls/transport.zig
```diff
@@ -0,0 +1,143 @@
...
+
+const std = @import("std");
+const events = @import("events.zig");
+const state = @import("state.zig");
+
+pub fn Contract(
+    comptime TransportParameters: type,
+    comptime Epoch: type,
+    comptime ErrorSet: type,
+) type {
```
Jaruso, line 33: `EventSink` unconditionally returns `error.HandshakeBufferOverflow`, so every instantiation's `ErrorSet` is implicitly required to contain that exact error even though the contract presents `ErrorSet` as generic. That couples future record-mode carriers to a QUIC-handshake error name and will produce a compile failure for otherwise valid error sets. Could the contract define its own neutral overflow error (for example `TransportBufferOverflow`) and expose `pub const Error = ErrorSet || error{TransportBufferOverflow}`, or accept the overflow error as an explicit comptime parameter?

src/tls/transport.zig
```diff
@@ -0,0 +1,143 @@
...
+            /// The peer certificate validation outcome.
+            certificate: events.CertificateState,
+            /// Keys for `epoch` are no longer needed and should be discarded.
+            discard_epoch: Epoch,
+            /// The handshake authenticated and completed on this side.
+            handshake_complete,
+        };
+
+        pub const EventSink = struct {
+            pub const max_events = 16;
```
Jaruso: These fixed `16 events / 16 KiB` limits were reasonable as policy in the QUIC driver, but moving them into the protocol-neutral contract makes them constraints for every future carrier. In particular, record-mode TLS or larger certificate flights may need different buffering. Consider making both capacities comptime parameters (with these values as defaults), or having `EventSink` operate over caller-supplied storage so the transport owns the bound.

Jaruso (commented):
The ownership split and wrapper around the shared transport backend look good. I left two suggestions in `src/tls/transport.zig` to make the new contract genuinely reusable by record-mode TLS rather than carrying QUIC seam policy into the core.

PR #394, opened 2026-07-11
tls: add AlertDescription enum and HandshakeError mapping

Jaruso:
**What changed and why**
Small preparatory step for #332 (typed alert mapping and strict TLS state transitions). Adds `src/tls/alerts.zig` defining the RFC 8446 §6 fatal `AlertDescription` values needed by current failure surfaces (`close_notify`, `unexpected_message`, `bad_certificate`, `unsupported_certificate`, `certificate_required`, `decode_error`, `decrypt_error`, `illegal_parameter`, `missing_extension`, `unsupported_extension`, `internal_error`), plus `fromHandshakeError`, a mapping from `tls_core.events.HandshakeError` to the alert a peer must receive. Only the TLS-level protocol errors currently defined in `tls_core` are mapped (`MalformedHandshake` -> `decode_error`, `CertificateInvalid` -> `bad_certificate`, `SecretExportFailed` -> `internal_error`); `AlpnMismatch` is a local negotiation policy failure rather than a TLS protocol violation and is intentionally left unmapped (returns `null`). `alerts` is exported from `src/tls/root.zig`. The rest of #332's scope (centralized state transition validation, strict ordering enforcement, etc.) is left for follow-up work.

**Related issues**
Related to #332 (not closing it — this is only a preparatory slice of its scope).

**Tests**
Added focused unit tests in `src/tls/alerts.zig` covering each mapped `HandshakeError` variant, the `AlpnMismatch` -> `null` case, and the RFC 8446 numeric values of every `AlertDescription` member.

**Security impact**
Touches TLS alert vocabulary (additive only, no behavior change): no existing code path yet consumes `alerts.zig`, so there is no change to handshake behavior. Reviewed for correctness against RFC 8446 §6 alert codes.

**Performance impact**
None. No hot path touched.

**Checklist**
- [x] `zig build test-tls --summary all --error-style verbose` green (27/27 tests passed)
- [ ] `zig build test --summary all` (not run; scope limited to `src/tls`)
- [ ] `zig build test-integration` (not run; no protocol/proxy behavior changed)
- [ ] README / CHANGELOG updated (no operator-visible behavior changed)

---
_Generated by [Claude Code](https://claude.ai/code/session_01JwFwyWHBCneUk45NFws4Us)_

### DIFF
--- a/src/tls/alerts.zig
+++ b/src/tls/alerts.zig
@@ -18,22 +18,22 @@ pub const AlertDescription = enum(u8) {
     decrypt_error = 51,
     missing_extension = 109,
     unsupported_extension = 110,
+    no_application_protocol = 120,
     certificate_required = 116,
     internal_error = 80,
 };
 
 /// Maps a `tls_core.events.HandshakeError` to the fatal alert a peer must be
-/// sent, or `null` if the error is not a TLS protocol violation. ALPN
-/// mismatch is a local negotiation policy failure, not a TLS-level protocol
-/// error, so it is intentionally left unmapped here; transports that need an
-/// alert-like code for it (e.g. QUIC's `no_application_protocol`) define
-/// their own translation.
-pub fn fromHandshakeError(err: events.HandshakeError) ?AlertDescription {
+/// sent. Per RFC 7301 §3.2, a server that cannot negotiate a mutually
+/// supported application protocol sends fatal `no_application_protocol`;
+/// RFC 9001 §4.1.1 has QUIC translate that same TLS alert into its
+/// CRYPTO_ERROR space rather than defining a separate close reason.
+pub fn fromHandshakeError(err: events.HandshakeError) AlertDescription {
     return switch (err) {
         error.MalformedHandshake => .decode_error,
         error.CertificateInvalid => .bad_certificate,
         error.SecretExportFailed => .internal_error,
-        error.AlpnMismatch => null,
+        error.AlpnMismatch => .no_application_protocol,
     };
 }
 
@@ -51,11 +51,11 @@ test "secret export failure maps to internal_error" {
     try testing.expectEqual(AlertDescription.internal_error, fromHandshakeError(error.SecretExportFailed));
 }
 
-test "ALPN mismatch is not a TLS protocol alert" {
-    try testing.expectEqual(@as(?AlertDescription, null), fromHandshakeError(error.AlpnMismatch));
+test "ALPN mismatch maps to no_application_protocol" {
+    try testing.expectEqual(AlertDescription.no_application_protocol, fromHandshakeError(error.AlpnMismatch));
 }
 
-test "alert description values match RFC 8446 section 6" {
+test "alert description values match RFC 8446 section 6 and RFC 7301 section 3.2" {
     try testing.expectEqual(@as(u8, 0), @intFromEnum(AlertDescription.close_notify));
     try testing.expectEqual(@as(u8, 10), @intFromEnum(AlertDescription.unexpected_message));
     try testing.expectEqual(@as(u8, 42), @intFromEnum(AlertDescription.bad_certificate));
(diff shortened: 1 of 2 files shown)


src/tls/alerts.zig
```diff
@@ -0,0 +1,70 @@
...
+    internal_error = 80,
+};
+
+/// Maps a `tls_core.events.HandshakeError` to the fatal alert a peer must be
+/// sent, or `null` if the error is not a TLS protocol violation. ALPN
+/// mismatch is a local negotiation policy failure, not a TLS-level protocol
+/// error, so it is intentionally left unmapped here; transports that need an
+/// alert-like code for it (e.g. QUIC's `no_application_protocol`) define
+/// their own translation.
+pub fn fromHandshakeError(err: events.HandshakeError) ?AlertDescription {
```
Jaruso: **Blocking:** `MalformedHandshake` currently covers both malformed bytes **and an illegal message order** (`events.zig` says “malformed or arrived in an illegal order”), but those require different alerts. RFC 8446 reserves `decode_error` for messages that do not conform to the formal syntax; a valid handshake message received at the wrong point in the state machine is `unexpected_message`. Mapping this combined error unconditionally to `decode_error` loses information and will produce the wrong peer-visible error for ordering failures. Suggest splitting the core error into something like `MalformedHandshake` and `UnexpectedHandshakeMessage` (or otherwise carrying a reason), then mapping them separately.

src/tls/alerts.zig
```diff
@@ -0,0 +1,70 @@
...
+    return switch (err) {
+        error.MalformedHandshake => .decode_error,
+        error.CertificateInvalid => .bad_certificate,
```
Jaruso, line 34: **Blocking:** ALPN mismatch is not merely a local policy failure here. RFC 7301 requires a server with no mutually supported protocol to send fatal `no_application_protocol` (120), and RFC 9001 requires QUIC endpoints to close with the corresponding TLS alert / QUIC `CRYPTO_ERROR` 0x0178 when ALPN negotiation fails. QUIC does not define a separate non-TLS translation; it converts the TLS `AlertDescription` by adding 0x100. Please add `no_application_protocol = 120` and map `AlpnMismatch` to it (or split the error only if some genuinely local, non-negotiation condition also uses this variant).

Jaruso (commented):
The typed alert vocabulary is a good preparatory step, but two current mappings conflict with the TLS/QUIC specifications and would cause callers to emit the wrong wire error. I think the shared `HandshakeError` vocabulary needs to be made specific enough to preserve these distinctions before this mapping is consumed.

src/tls/alerts.zig
Jaruso: Fixed in 49a8a70: added `AlertDescription.no_application_protocol = 120` and mapped `AlpnMismatch` to it, per RFC 7301 §3.2 / RFC 9001 §4.1.1. Updated the unit tests accordingly.

---

src/tls/alerts.zig
Jaruso: Agreed this is a real gap — `decode_error` is wrong for the illegal-ordering cases. Leaving it as-is for this PR, though: splitting `MalformedHandshake` means changing the shared `HandshakeError` vocabulary in `events.zig` and reclassifying every call site across `src/tls/messages.zig`, `src/tls/negotiation.zig`, `src/quic/tls_handshake.zig`, and `src/quic/tls_backend.zig` (several dozen sites, many of which are genuinely ambiguous between "malformed bytes" and "message at the wrong point in the state machine" without deeper state-machine work). That's squarely #332's "centralize state transition validation" scope, not this alert-vocabulary prep PR.

Proposing this be picked up as part of #332 itself (or a dedicated follow-up issue) alongside the strict state-machine work, so the split and the call-site reclassification happen together with real ordering-violation detection rather than as a mechanical rename here. Happy to file that follow-up issue if useful.

---

PR #395, opened 2026-07-11
tls: move backend driver lifecycle into core

Jaruso:
## Summary
- add a generic tls_core.engine.Driver over the shared transport backend contract
- make the QUIC handshake wrapper delegate backend start/receive/failure lifecycle to the core driver
- keep QUIC-specific CRYPTO routing, secret installation, transport-parameter authentication, and CID binding in src/quic
- document the updated ownership split

Part of #329. This stays separate from #332 / PR #394 alert mapping work.

## Validation

### DIFF
--- a/src/tls/engine.zig
+++ b/src/tls/engine.zig
@@ -35,9 +35,153 @@ pub const Engine = struct {
     }
 };
 
+pub fn Driver(comptime Transport: type) type {
+    return struct {
+        role: state.Role,
+        backend: Transport.Backend,
+        state: state.DriverState = .idle,
+        failure_reason: ?Transport.Error = null,
+        sink: Transport.EventSink = .{},
+
+        const Self = @This();
+
+        pub fn init(role: state.Role, backend: Transport.Backend) Self {
+            return .{ .role = role, .backend = backend };
+        }
+
+        /// Start the backend and return the driver's internal event sink.
+        /// The returned pointer and every event payload slice borrow storage
+        /// owned by this driver; they stay valid only until the next `start` or
+        /// `receive` call, both of which reset and overwrite the sink.
+        pub fn start(self: *Self, params: Transport.TransportParametersType) Transport.Error!*Transport.EventSink {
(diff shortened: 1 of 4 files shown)


src/tls/engine.zig
```diff
@@ -35,9 +35,103 @@ pub const Engine = struct {
...
+
+        const Self = @This();
+
+        pub fn init(role: state.Role, backend: Transport.Backend) Self {
+            return .{ .role = role, .backend = backend };
+        }
+
+        pub fn start(self: *Self, params: Transport.TransportParametersType) Transport.Error!*Transport.EventSink {
+            self.state = .in_progress;
+            self.sink.reset();
```
Jaruso, line 59: Since this type now claims ownership of backend lifecycle state, should `start` reject or assert unless the driver is `.idle`? As written, calling `start` twice—or calling it after `.complete`/`.failed`—silently resets the sink, changes the state back to `.in_progress`, and invokes the backend again. The previous QUIC wrapper also did not guard this, but moving the behavior into a reusable core driver is a good point to define and test the valid transition contract.

src/tls/engine.zig
```diff
@@ -35,9 +35,103 @@ pub const Engine = struct {
...
+            self.backend.start(self.role, params, &self.sink) catch |err| return self.fail(err);
+            return &self.sink;
+        }
+
```
Jaruso, line 63: Could we document that this returned pointer is borrowed only until the next `start`/`receive` call? The next call resets and overwrites the same internal sink and scratch storage, so retaining the pointer or any event payload slice across calls is unsafe. That ownership rule existed at the old seam, but it is now part of this generic public API and is easy for a future record-layer caller to miss.

Jaruso (commented):
Nice extraction overall. Moving backend invocation, sink ownership, state, and failure capture into the protocol-neutral driver makes the QUIC wrapper substantially clearer while preserving the intended ownership split. I left two suggestions around making the new generic driver's lifecycle contract explicit and safe.

PR #397, opened 2026-07-11
tls: define record-mode handshake transport contract

Jaruso:
## Summary
- add `tls.record_transport` as the record-mode handshake contract for future TLS-over-TCP work
- model plaintext, handshake-protected, and application-protected epochs with asymmetric read/write key install events
- document/test borrowed event payload lifetime and secure zeroization of copied secrets

## Validation

Refs #337

### DIFF
--- a/src/tls/record_transport.zig
+++ b/src/tls/record_transport.zig
@@ -1,0 +1,255 @@
+//! Record-mode TLS handshake contract for future TLS-over-TCP support.
+//!
+//! This module is deliberately below any socket, readiness, ciphertext-buffer,
+//! or record-codec layer. A future record implementation decrypts TLS records,
+//! supplies the resulting handshake bytes to `Driver.receive`, and consumes the
+//! events emitted here to write plaintext handshake bytes, install traffic keys,
+//! surface negotiated metadata, send fatal alerts, and observe completion.
+//!
+//! Ownership rules are narrow and testable:
+//! - All byte slices in emitted events are copied into the driver's `EventSink`.
+//! - Those slices borrow the sink and remain valid only until the next driver
+//!   `start` or `receive`, both of which reset the sink.
+//! - Resetting or deinitializing the sink securely zeroes the used scratch
+//!   range, so copied traffic secrets do not survive the event lifetime.
+
+const std = @import("std");
+const alerts = @import("alerts.zig");
+const engine = @import("engine.zig");
+const events = @import("events.zig");
+const state = @import("state.zig");
+
+pub const RecordEpoch = enum {
+    /// Handshake bytes before record protection is active.
+    plaintext,
+    /// Handshake records protected with TLS handshake traffic keys.
+    handshake_protected,
+    /// Records protected with TLS application traffic keys.
+    application_protected,
+};
+
+pub const Error = events.HandshakeError || error{
+    InvalidHandshakeState,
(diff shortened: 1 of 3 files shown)


src/tls/record_transport.zig
```diff
@@ -0,0 +1,255 @@
...
+pub const max_byte_count = 16 * 1024;
+
+pub const TrafficKeys = struct {
+    epoch: RecordEpoch,
+    secret: []const u8,
+};
+
+pub const RecordHandshakeEvent = union(enum) {
+    /// Plaintext handshake bytes the record layer must frame and write.
+    emit_plaintext: []const u8,
```
Jaruso: `emit_plaintext` needs to carry the record-protection epoch (for example `{ epoch, data }`), not just bytes. The consumer cannot safely infer this from surrounding key-install events: in the fixture below the client installs application write keys and then emits `client finished`, but Finished must still be sent under handshake keys. A record layer applying events in order could therefore protect the message with the wrong keys. Please make the required outbound epoch explicit on every emitted handshake flight.

src/tls/record_transport.zig
```diff
@@ -0,0 +1,255 @@
...
+
+            if (epoch == .handshake_protected and std.mem.eql(u8, bytes, "client finished")) {
+                try sink.installReadKeys(.application_protected, "server read application");
+                try sink.installWriteKeys(.application_protected, "server write application");
+                try sink.emitComplete();
+                return;
+            }
+
+            try sink.emitFatalAlert(alerts.fromHandshakeError(error.UnexpectedHandshakeMessage));
+            return error.UnexpectedHandshakeMessage;
```
Jaruso, line 301: This fatal alert is not observable through `engine.Driver.receive`: returning the error causes `receive()` to return the error instead of the sink pointer, so the caller never gets the `.fatal_alert` event it is supposed to serialize. Please either return a result that contains both emitted events and an optional terminal error, or move alert mapping/emission into the driver so failure paths still expose the alert before shutdown. A regression test should exercise this branch through `Driver.receive` and assert that the alert is consumable.

src/tls/record_transport.zig
```diff
@@ -0,0 +1,255 @@
...
+pub const Contract = struct {
+    pub const TransportParametersType = void;
+    pub const EpochType = RecordEpoch;
+    pub const Error = record_transport.Error;
+    pub const Event = RecordHandshakeEvent;
+    pub const EventSink = record_transport.EventSink;
+    pub const Backend = record_transport.Backend;
+};
+
+pub const Driver = engine.Driver(Contract);
```
Jaruso: The sink has a secure `deinit`, but the exported API is `Driver = engine.Driver(Contract)` and that driver currently has no record-transport-specific cleanup method. That makes final secret zeroization easy to miss when a connection is destroyed without another `start`/`receive` reset. Consider exposing a `deinit()` at the driver/record-transport boundary that delegates to `sink.deinit()`, and test that teardown zeroes the last emitted traffic secrets.

Jaruso (commented):
The record-mode seam is a useful addition, but two API behaviors need clarification before a real record layer can consume it safely: outbound handshake data must carry its protection epoch, and failure alerts must remain observable when the backend returns an error. I also suggest making secret cleanup available at the Driver boundary.

PR #398, opened 2026-07-12
tls: add incremental TLS 1.3 record codec

Jaruso:
## Summary
- add `tls.record_codec` for bounded TLSPlaintext/TLSCiphertext parsing and serialization
- support incremental split-header/body, byte-at-a-time, and coalesced multi-record input through a bounded sink
- add TLSInnerPlaintext encode/decode helpers and a fuzz entrypoint for future security corpus work

## Validation

Refs #350

### DIFF
--- a/src/tls/record_codec.zig
+++ b/src/tls/record_codec.zig
@@ -105,12 +105,20 @@ pub const Parser = struct {
     /// Feed arbitrary TCP bytes into the parser. Completed records are copied
     /// into `sink`; incomplete headers/bodies stay buffered for the next feed.
     pub fn feed(self: *Parser, bytes: []const u8, sink: anytype) Error!void {
+        try self.drain(sink);
         for (bytes) |byte| {
             if (self.len == self.pending.len) return error.RecordBufferOverflow;
             self.pending[self.len] = byte;
             self.len += 1;
             try self.drain(sink);
         }
+    }
+
+    /// Retry emission of already-buffered complete records after the caller has
+    /// made room in `sink`. This supports bounded sink/backpressure loops
+    /// without appending unrelated bytes first.
+    pub fn drainReady(self: *Parser, sink: anytype) Error!void {
+        try self.drain(sink);
     }
 
     pub fn finish(self: *const Parser) Error!void {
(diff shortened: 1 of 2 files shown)


src/tls/record_codec.zig
```diff
@@ -0,0 +1,353 @@
...
+    }
+
+    fn drain(self: *Parser, sink: anytype) Error!void {
+        while (self.len >= header_len) {
+            const header = try parseHeader(self.pending[0..header_len], self.mode);
+            const record_len = header_len + header.payload_len;
+            if (self.len < record_len) return;
+            try sink.push(.{
+                .content_type = header.content_type,
+                .legacy_version = header.legacy_version,
```
Jaruso, line 135: `sink.push` can fail after the parser has buffered a complete record, but that record remains in `pending` and there is no public way to retry `drain`. Calling `feed` with an empty slice does nothing; calling it with more bytes appends before draining, and a maximum-sized record will immediately return `RecordBufferOverflow`. This makes `RecordSinkOverflow` effectively unrecoverable and breaks bounded/backpressure-style consumption. Consider exposing a retryable drain operation, draining before appending new bytes, or defining `feed` to return consumed-byte progress so callers can reset/consume the sink and resume without losing parser state. A regression test should fill a one-record sink with two coalesced records, reset the sink, and verify the second record can still be emitted.

src/tls/record_codec.zig
```diff
@@ -0,0 +1,353 @@
...
+    @memcpy(out[header_len..][0..payload.len], payload);
+    return out[0 .. header_len + payload.len];
+}
+
+pub fn encodeInnerPlaintext(content_type: ContentType, content: []const u8, padding_len: usize, out: []u8) Error![]const u8 {
+    if (content_type == .change_cipher_spec) return error.InvalidRecordType;
+    if (content.len > max_plaintext_fragment_len) return error.RecordTooLarge;
+    const total = content.len + 1 + padding_len;
+    if (total > max_ciphertext_fragment_len) return error.RecordTooLarge;
+    if (out.len < total) return error.RecordBufferOverflow;
```
Jaruso, line 196: This addition can overflow `usize` before the subsequent limit check when `padding_len` is caller-controlled. In safety-enabled builds that traps; in optimized unsafe builds it can wrap and undermine the output bounds checks. Please use checked arithmetic (for example `std.math.add`) or compare `padding_len` against the remaining allowance before computing `total`. It would be good to add a test with `padding_len = std.math.maxInt(usize)` and assert `RecordTooLarge`.

Jaruso (commented):
The codec is well-scoped and the incremental parsing tests cover the main fragmentation patterns nicely. I found two correctness issues worth fixing before merge: sink backpressure can leave the parser unrecoverably wedged, and the inner-plaintext size calculation can overflow before the bounds check.

PR #400, opened 2026-07-12
crypto: add typed secret ownership helpers

Jaruso:
## Summary
- add shared `crypto_secrets` helpers with fixed-size and bounded dynamic secret containers
- centralize secure zeroing and constant-time equality through the new helper module
- migrate QUIC TLS traffic-secret storage and record transport scratch wiping onto the helper
- document secret ownership, replacement, deinit, and non-formatting conventions

## Validation

Refs #372

### DIFF
--- a/src/crypto/secrets.zig
+++ b/src/crypto/secrets.zig
@@ -1,0 +1,203 @@
+//! Secret ownership and zeroization helpers (#372).
+//!
+//! These containers make secret lifetimes explicit. They copy caller-provided
+//! bytes into owned storage, expose only borrowed slices, wipe replaced values
+//! before reuse, and require `deinit` before the value is discarded.
+
+const std = @import("std");
+
+pub const Error = error{SecretTooLarge};
+
+pub fn secureZero(buffer: []u8) void {
+    std.crypto.secureZero(u8, buffer);
+}
+
+pub fn constantTimeEqual(a: []const u8, b: []const u8) bool {
+    if (a.len != b.len) return false;
+    return std.crypto.timing_safe.compare(u8, a, b, .big) == .eq;
+}
+
+pub fn FixedSecret(comptime capacity: usize) type {
+    return struct {
+        bytes: [capacity]u8 = [_]u8{0} ** capacity,
+        len: usize = 0,
+
+        const Self = @This();
+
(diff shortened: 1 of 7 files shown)


src/crypto/secrets.zig
```diff
@@ -0,0 +1,203 @@
...
+        pub fn init(value: []const u8) Error!Self {
+            var secret = Self{};
+            try secret.replace(value);
+            return secret;
+        }
+
+        pub fn replace(self: *Self, value: []const u8) Error!void {
+            if (value.len > self.bytes.len) return error.SecretTooLarge;
+            self.clear();
+            @memcpy(self.bytes[0..value.len], value);
```
Jaruso: `replace` is unsafe when `value` aliases this secret's own storage. For example, `secret.replace(secret.slice()[1..])` calls `clear()` first, zeroing the source, and then copies zeros back. The bounded variant has the same behavior. Could we either reject overlapping input explicitly or make replacement overlap-safe (with a temporary/in-place move strategy), and add a regression test for it?

src/crypto/secrets.zig
```diff
@@ -0,0 +1,203 @@
...
+
+pub const BoundedSecret = struct {
+    allocator: std.mem.Allocator,
+    bytes: []u8,
+    len: usize = 0,
+
+    pub fn initCapacity(allocator: std.mem.Allocator, capacity: usize) !BoundedSecret {
+        const bytes = try allocator.alloc(u8, capacity);
+        @memset(bytes, 0);
+        return .{ .allocator = allocator, .bytes = bytes };
```
Jaruso: This type owns an allocation but remains freely copyable as a normal Zig value. An assignment/pass-by-value such as `var b = a;` creates two apparent owners of the same `bytes`; deinitializing one leaves the other dangling, and deinitializing both can double-free. That conflicts with the ownership guarantee this helper is meant to provide. Could this be redesigned around a non-copyable handle/in-place initialization pattern, or at minimum make ownership transfer explicit and avoid returning/passing `BoundedSecret` by value?

Jaruso (commented):
The abstraction is a good direction, but there are two ownership/aliasing cases worth fixing before it becomes the shared secret-storage primitive.

PR #401, opened 2026-07-12
pki: model X.509 certificate fields and validation extensions (#341)

Jaruso:
**What changed and why**
Adds `src/pki/x509.zig`, a typed, policy-neutral X.509 certificate model (story 324-C of the Web PKI epic #324), parsing DER certificates into every field needed for identity matching, signature verification, path building, and RFC 5280 validation by #324-D through #324-G. The parser enforces structure and DER well-formedness only; trust decisions stay in validation policy.

`TBSCertificate` decodes into version, serial, inner/outer signature algorithms (encoding-identity enforced per RFC 5280 §4.1.2.3, with `SignatureAlgorithm.classify` covering RSA PKCS#1 SHA-2, RSA-PSS, ECDSA SHA-2, and Ed25519), issuer/subject `Name`s with structured RDNs plus byte-exact `raw` views for RFC 5280 binary comparison, validity times (UTCTime/GeneralizedTime with the 2050 cutoff enforced), SPKI with RSA/P-256/P-384/P-521/Ed25519 key-type identification, v2/v3 unique identifiers, and the signature `BIT STRING`. Exact TBSCertificate TLV bytes are preserved (`tbs_raw`) for signature verification. Extensions parse into typed representations — SAN (all `GeneralName` variants; unconsumed alternatives retained raw), Basic Constraints, Key Usage (DER named-bit canonicality enforced), EKU with purpose helpers, SKI, AKI (keyid/issuer/serial pairing rule enforced), Name Constraints (CIDR-form IP subtrees), AIA, CRL Distribution Points, and certificate policies — while unknown extensions are retained with OID, criticality, and raw value; `hasUnhandledCriticalExtension` surfaces unknown critical extensions for fail-closed validation.

Structural strictness: DER DEFAULT values must be omitted (explicit v1 version, `critical FALSE`, `cA FALSE` are rejected), duplicate extension OIDs are rejected, extensions require v3 and unique IDs v2/v3, optional trailing TBS fields must appear in ascending tag order, and every nested structure must consume its input exactly. Certificates borrow the caller's DER buffer (pairing with `pem.Certificate.der`) and allocate collections from an internal arena released by one `deinit`. Configurable `Limits` bound extension, RDN, GeneralName, EKU, policy, and subtree counts. A `fuzzParseCertificate` entrypoint feeds #327-G. `oid.zig`'s well-known table grows the signature, curve, extension, EKU, access-method, and name-attribute arcs.

**Related issues**
Closes #341

**Tests**
`src/pki/x509_tests.zig` adds 17 tests in two groups. Real-world fixtures (committed as `.crt` under `src/pki/testdata/`, generated with OpenSSL 3.0): an RSA-2048 self-signed CA with critical name constraints (permitted DNS + excluded IP/mask) and a 2059 GeneralizedTime expiry; a P-256 leaf signed by that CA carrying SAN (DNS/wildcard/IPv4/IPv6/email/URI), critical KU, EKU, SKI, AKI with `issuer:always` (all three fields), AIA (OCSP + caIssuers), CRLDP, and certificate policies; a self-signed Ed25519 certificate; and a genuine v1 certificate. Synthetic corpora via a minimal DER builder cover: v3 round-trip, inner/outer signature-algorithm mismatch, explicit-default and future versions, extensions-on-v1/v2 and unique-ID-on-v1 rejection with v2 unique-ID acceptance, duplicate extensions, unknown critical/non-critical extension retention, explicit `critical FALSE` and `cA FALSE`, truncation sweep across every prefix, trailing bytes, empty extension sequences, pre-2050 GeneralizedTime, invalid SAN IP lengths, empty RDN sets, EC SPKI without curve parameters, directoryName retention + `parseNameRaw` round-trip, extension count limits, `checkAllAllocationFailures` leak coverage, and a bit-flip fuzz sweep over a real certificate.

**Security impact**
This is TLS-area code parsing untrusted input (certificates). Reviewed: all collection growth is bounded by configurable limits before allocation; the bounded #339 DER reader enforces depth/length/element limits and no path reads outside the supplied slice; DER DEFAULT-omission and canonical named-bit rules are enforced so a certificate has exactly one accepted encoding; unknown critical extensions are surfaced rather than ignored so validation can fail closed; duplicate extension OIDs are rejected; the arena plus `errdefer` makes every error path leak-free (verified by `checkAllAllocationFailures`). No existing header parsing, routing, auth, or logging paths are touched — this is a new module not yet wired into the TLS runtime.

**Performance impact**
None. New standalone module off every hot path. Parsing is a single pass over the DER with one arena per certificate.

**Checklist**

---


src/pki/x509.zig
```diff
@@ -0,0 +1,1141 @@
...
+                .second = gen.second,
+                .encoding = .generalized,
+            };
+        },
+        else => return error.MalformedValidity,
+    }
+}
+
+fn validateDirectoryString(elem: der.Element) Error!void {
+    if (elem.tag.class != .universal or elem.tag.constructed) return;
```
Jaruso: This early return also accepts a **constructed** encoding of a known directory-string tag (for example, constructed UTF8String) without validation. DER requires these string encodings to be primitive, so this creates a structural-canonicality hole despite the parser's strictness guarantee. Could we return `MalformedName` when the tag is one of the recognized string types and `constructed == true`, while continuing to retain genuinely unknown attribute-value tags raw? A synthetic test with constructed UTF8String would cover it.

src/pki/x509.zig
```diff
@@ -0,0 +1,1141 @@
...
+            if (points.items.len >= self.limits.max_distribution_points) return error.CountLimitExceeded;
+            const point_elem = inner.readElement() catch return error.MalformedExtension;
+            if (!isSequence(point_elem)) return error.MalformedExtension;
+            var point_reader = inner.childReader(point_elem.content_offset, point_elem.content.len) catch return error.MalformedExtension;
+
+            var full_names: []const GeneralName = &.{};
+            while (point_reader.remaining() > 0) {
+                const elem = point_reader.readElement() catch return error.MalformedExtension;
+                if (elem.tag.class != .context_specific) return error.MalformedExtension;
+                switch (elem.tag.number) {
```
Jaruso, line 939: `DistributionPoint` is a SEQUENCE of three OPTIONAL fields in tag order, each appearing at most once, but this loop currently permits duplicates and reordering (for example `[1], [0]`, or two `[0]` fields). It also accepts `[1]`/`[2]` without checking their primitive/constructed form or validating their implicit BIT STRING / GeneralNames payload. That means malformed DER can be retained as `raw` even though this parser promises nested structural validation. Could we track `last_tag` as done for AKI/name constraints, reject duplicates, and at least validate the implicit field encodings before discarding their typed values? Tests for duplicate/out-of-order tags and malformed reasons/cRLIssuer would be useful.

Jaruso (commented):
Strong foundation overall, especially the bounded parsing, exact TBS preservation, duplicate-extension rejection, and negative corpus. I found two structural-validation gaps worth tightening before downstream validation starts relying on this model.

PR #402, opened 2026-07-12
crypto: codify supported provider capability profile

Jaruso:
## Summary
- add a checked-in crypto capability matrix for the #327-B provider profile
- expose programmatic provider capability queries plus TLS policy derivation/validation helpers
- document the matrix, Zig version floor, provider differences, and review path

Refs #371

## Validation

### DIFF
--- a/src/crypto/profile.zig
+++ b/src/crypto/profile.zig
@@ -1,0 +1,214 @@
+//! Checked-in cryptographic capability profile (#371, epic #327).
+//!
+//! This file is intentionally data-heavy. It is the build-visible matrix that
+//! maps protocol algorithms to the implementation family that supplies them,
+//! the current provider status, test coverage, and consumers.
+
+const std = @import("std");
+const provider = @import("provider.zig");
+
+pub const zig_version_floor = "0.16.0";
+
+pub const ProviderKind = enum {
+    pure_zig,
+    openssl,
+};
+
+pub const Status = enum {
+    supported,
+    provider_deferred,
+    unsupported,
+};
+
+pub const Implementation = enum {
+    zig_std_crypto,
+    project_code,
+    openssl_provider,
+    unavailable,
+};
+
+pub const Category = enum {
+    hash,
+    hkdf,
+    aead,
+    group,
+    signature,
+    certificate_helper,
+    entropy,
+};
+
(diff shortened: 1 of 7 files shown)


src/crypto/profile.zig
```diff
@@ -0,0 +1,327 @@
...
+//! the current provider status, test coverage, and consumers. TLS policy code
+//! can derive advertised protocol capabilities from a provider's typed
+//! `provider.Capabilities`, so configuration cannot offer algorithms the
+//! selected backend cannot execute.
+
+const std = @import("std");
+const provider = @import("provider.zig");
+const tls_core = @import("tls_core");
+const tls_policy = tls_core.policy;
+
```
Jaruso, line 9: This introduces a dependency from the generic crypto provider package back into `tls_core`. That reverses the boundary described in the provider design (protocol code consumes crypto capabilities, rather than crypto importing protocol policy types) and makes a future `tls_core -> crypto` dependency prone to a module cycle. Could the provider-neutral matrix stay in `src/crypto/profile.zig`, while `TlsCapabilities`, policy construction, and TLS enum mapping move into a TLS-side adapter such as `src/tls/crypto_profile.zig`?

Jaruso (commented):
The capability matrix is a useful addition, but I found two structural issues worth addressing before this becomes the long-term source of truth.

src/crypto/profile.zig
```diff
@@ -0,0 +1,327 @@
...
+                .openssl => entry.openssl_status,
+            };
+        }
+    }
+    return .unsupported;
+}
+
+pub fn capabilities(kind: ProviderKind) provider.Capabilities {
+    var caps = provider.Capabilities{};
+    if (status(kind, .{ .hash = .sha256 }) == .supported) caps.hashes.insert(.sha256);
```
Jaruso: The docs call `rows` the source of truth, but this function manually enumerates every algorithm a second time. A newly added supported row can be omitted here and silently disappear from the returned capability set; the coverage test only proves that a row exists, not that this projection is exhaustive. Please derive `Capabilities` by iterating `rows` and switching on each supported provider-algorithm row (ignoring HKDF/certificate/entropy categories), or add an exhaustive consistency test that fails whenever a supported row is not represented.

PR #403, opened 2026-07-12
pki: implement SAN-only DNS and IP identity matching (#342)

Jaruso:
**What changed and why**
Adds `src/pki/identity.zig`, RFC 9525-style service identity verification over the #341 certificate model (story 324-D of the Web PKI epic #324). Verification uses `subjectAltName` only: DNS references match only DNS-ID entries, IP references only IP-address entries, and Common Name fallback is disabled — the subject field is never consulted, so CN-only certificates always fail with a typed `no_subject_alt_name` mismatch.

Wildcard behavior is conservative: a wildcard is honored only as the complete left-most label of the presented identifier, matches exactly one label (never zero, never across label boundaries), and requires at least two literal labels after it, so public-suffix-style identifiers (`*.com`) never match. Partial-label (`f*o.example.com`) and interior (`a.*.example.com`) wildcards match nothing, and presented identifiers with invalid LDH syntax are skipped rather than compared.

DNS references carry an explicit A-label/IDNA input contract: non-ASCII input is rejected with `error.NonAsciiDnsReference` instead of being silently transformed — callers convert internationalized names to A-labels first. References are validated for LDH label syntax, ASCII case is ignored on both sides, exactly one trailing dot is stripped from both the reference and each presented identifier, references may never contain `*`, and IPv4-shaped names that fail strict IP-literal parsing (octal forms like `0177.0.0.1`, out-of-range octets) are rejected to prevent address confusion. IP literals (including bracketed IPv6) parse through `std.Io.net`'s strict canonical parsers and compare byte-wise with no v4/v6-mapped equivalence.

Matching is a pure, allocation-free function over one parsed certificate — fully independent of chain construction (#324-E) and signature validation — and mismatch results carry only the caller's reference identity plus a mismatch class (`no_subject_alt_name`, `no_entries_of_reference_type`, `no_matching_entry`), never certificate contents. `presentedDnsIdMatches` is exported for TLS-layer reuse.

**Related issues**
Closes #342

**Tests**
`src/pki/identity_tests.zig` adds 11 tests (84 total under `zig build test-pki`) covering the full acceptance matrix: exact DNS, mixed-case, trailing-dot, wildcard, and subdomain-depth matches against the #341 P-256 leaf fixture (SAN with DNS, wildcard, IPv4, IPv6, email, and URI entries); IPv4/IPv6/bracketed/uncompressed-form references with byte-wise comparison and no cross-family matching; CN-only failure via the real v1 fixture whose CN exactly equals the reference; `no_entries_of_reference_type` via the DNS-only Ed25519 fixture; mismatch results echoing reference identity and class; reference-parsing validation (A-label contract, empty/oversized/embedded-wildcard/empty-label/hyphen/underscore rejection, IPv4-shaped names); a 21-case presented-identifier table for wildcard and malformed-identifier behavior; and synthetic certificates presenting hostile SANs (`*.com`, `f*o.example.com`, `*.*.example.com`, `a.*.example.com`, IP-shaped DNS names) that must never match.

**Security impact**
This is TLS-area identity-verification logic. Reviewed: no Unicode transformations are performed on any input (explicit A-label contract); wildcard acceptance is strictly narrower than RFC 9525 allows (complete left-most label, one label deep, two-literal-label minimum); IP references never compare against DNS-IDs, blocking IP-shaped DNS-name confusion, and strict IP parsing rejects octal/non-canonical literals; presented identifiers from untrusted certificates are syntax-validated before comparison and malformed ones match nothing; failure results expose no certificate data beyond the mismatch class. No existing parsing, routing, auth, or logging paths are touched — this is a new pure module not yet wired into the TLS runtime.

**Performance impact**
None. Pure, allocation-free matching off every hot path; a single pass over SAN entries per verification.

**Checklist**

---

### DIFF
--- a/src/pki/identity.zig
+++ b/src/pki/identity.zig
@@ -105,17 +105,21 @@ pub fn dnsReference(host: []const u8) Error!Reference {
     return .{ .dns_name = name };
 }
 
-/// Parse an IPv4/IPv6 literal (brackets accepted) into a reference
-/// identity, or null when the input is not an IP literal.
+/// Parse an IPv4/IPv6 literal into a reference identity, or null when the
+/// input is not an IP literal. Brackets are URI host syntax for IPv6 only:
+/// `[::1]` is accepted, `[127.0.0.1]` is not an IP literal.
 pub fn ipReference(literal: []const u8) ?Reference {
-    var text = literal;
-    if (text.len >= 2 and text[0] == '[' and text[text.len - 1] == ']') {
-        text = text[1 .. text.len - 1];
+    if (literal.len >= 2 and literal[0] == '[' and literal[literal.len - 1] == ']') {
+        const inner = literal[1 .. literal.len - 1];
+        if (net.Ip6Address.parse(inner, 0)) |v6| {
+            return .{ .ip_address = .{ .v6 = v6.bytes } };
+        } else |_| {}
+        return null;
     }
-    if (net.Ip4Address.parse(text, 0)) |v4| {
+    if (net.Ip4Address.parse(literal, 0)) |v4| {
         return .{ .ip_address = .{ .v4 = v4.bytes } };
     } else |_| {}
-    if (net.Ip6Address.parse(text, 0)) |v6| {
+    if (net.Ip6Address.parse(literal, 0)) |v6| {
         return .{ .ip_address = .{ .v6 = v6.bytes } };
     } else |_| {}
     return null;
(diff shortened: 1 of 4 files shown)


src/pki/identity.zig
```diff
@@ -0,0 +1,271 @@
...
+    const ref = try reference(host);
+    return verify(certificate, ref);
+}
+
+/// Whether one presented DNS-ID matches a validated DNS reference name.
+/// `presented` comes from an untrusted SAN entry: identifiers with invalid
+/// label syntax or non-conforming wildcards match nothing (the certificate
+/// may still match through another entry). Exposed for TLS-layer reuse;
+/// `reference_name` must come from `dnsReference`.
+pub fn presentedDnsIdMatches(presented: []const u8, reference_name: []const u8) bool {
```
Jaruso: Because this is exported for TLS-layer reuse, the validation precondition is easy to violate: `reference_name` is just a raw slice, so a caller can pass malformed input that `dnsReference` would reject. For example, `presentedDnsIdMatches("*.example.com", ".example.com")` consumes the empty first label as the wildcard match and returns true. Could this either remain private, validate `reference_name` internally, or accept a type that can only be produced by the validated constructor? In security-sensitive identity code, I would avoid relying on a doc-only precondition.

src/pki/identity.zig
```diff
@@ -0,0 +1,271 @@
...
+pub fn dnsReference(host: []const u8) Error!Reference {
+    const name = try validateDnsReference(host);
+    return .{ .dns_name = name };
+}
+
+/// Parse an IPv4/IPv6 literal (brackets accepted) into a reference
+/// identity, or null when the input is not an IP literal.
+pub fn ipReference(literal: []const u8) ?Reference {
+    var text = literal;
+    if (text.len >= 2 and text[0] == '[' and text[text.len - 1] == ']') {
```
Jaruso: This strips brackets before trying both address families, so `[127.0.0.1]` is accepted as an IPv4 identity. Brackets are the URI host syntax for IPv6 literals, not IPv4. Accepting bracketed IPv4 can make this verifier classify input differently from upstream URL/SNI parsing. Suggest only permitting brackets when the inner parse succeeds as IPv6, and adding a rejection test for bracketed IPv4.

Jaruso (commented):
Strong overall implementation: the SAN-only split, strict IP byte comparison, CN fallback removal, and hostile-certificate tests are all well designed. I left two inline comments around hardening the exported API boundary and tightening bracket handling. CI was still running when I reviewed.

PR #404, opened 2026-07-12
tls: implement TLS 1.3 record protection state

Jaruso:
## Summary
- add TLS 1.3 record-protection read/write states over the shared crypto-provider seam
- derive per-suite traffic keys and IVs via HKDF labels, construct sequence nonces, and use TLSCiphertext headers as AEAD associated data
- cover all supported cipher suites plus tamper, wrong-key, malformed-inner, header, and sequence-exhaustion failures

Refs #351

## Validation

### DIFF
--- a/src/tls/record_protection.zig
+++ b/src/tls/record_protection.zig
@@ -1,0 +1,363 @@
+//! TLS 1.3 record protection and deprotection.
+//!
+//! This layer sits above `record_codec`: the codec owns TLSPlaintext,
+//! TLSCiphertext, and TLSInnerPlaintext framing, while this module owns traffic
+//! key derivation, per-record nonces, sequence accounting, and AEAD use through
+//! the shared crypto-provider boundary.
+
+const std = @import("std");
+const crypto = @import("crypto");
+const algorithms = @import("algorithms.zig");
+const record_codec = @import("record_codec.zig");
+
+const provider = crypto.provider;
+const secrets = crypto.secrets;
+
+pub const Error = record_codec.Error || provider.HkdfError || provider.SealError || provider.OpenError || error{
+    InvalidTrafficSecretLength,
+    SequenceExhausted,
+};
+
+const KeySecret = secrets.FixedSecret(provider.max_aead_key_len);
+const IvSecret = secrets.FixedSecret(provider.aead_nonce_len);
+
+pub const SuiteProfile = struct {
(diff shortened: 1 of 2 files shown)


src/tls/record_protection.zig
```diff
@@ -0,0 +1,320 @@
...
+    };
+}
+
+pub const TrafficKeys = struct {
+    profile: SuiteProfile,
+    key: KeySecret = .{},
+    iv: IvSecret = .{},
+
+    pub fn derive(crypto_provider: provider.CryptoProvider, cipher_suite: algorithms.CipherSuite, traffic_secret: []const u8) Error!TrafficKeys {
+        const profile = suiteProfile(cipher_suite);
```
Jaruso, line 44: Could we enforce `traffic_secret.len == profile.hash.digestLength()` here? TLS 1.3 traffic secrets produced by `Derive-Secret` are exactly `Hash.length`. Allowing arbitrary lengths means a key-schedule bug can still derive deterministic—but non-compliant—record keys instead of failing at the boundary. A typed `InvalidTrafficSecretLength` error plus tests for short/long secrets would make this API contract explicit.

src/tls/record_protection.zig
```diff
@@ -0,0 +1,320 @@
...
+    }
+
+    pub fn deinit(self: *WriteState) void {
+        self.keys.deinit();
+        self.sequence = 0;
+    }
+
+    pub fn seal(self: *WriteState, content_type: record_codec.ContentType, plaintext: []const u8, padding_len: usize, out: []u8) Error![]const u8 {
+        if (self.sequence == std.math.maxInt(u64)) return error.SequenceExhausted;
+
```
Jaruso, line 89: This rejects sequence number `2^64 - 1` before it is used, so each key permits only `2^64 - 1` records rather than the full counter range `0...2^64-1`. The same applies to `ReadState`. Consider storing a separate `exhausted` flag (set after successfully using `maxInt(u64)`) or otherwise representing the post-final-record state without increment overflow. A boundary test at `maxInt(u64)-1` and `maxInt(u64)` would pin the behavior down.

Jaruso (commented):
The record-protection split is clean, the nonce/AAD construction follows TLS 1.3, and the negative-path coverage is strong. I left two focused correctness suggestions around enforcing the traffic-secret contract and representing sequence exhaustion without sacrificing the final valid sequence number.

PR #405, opened 2026-07-12
build: enforce TLS dependency audit and pure-Zig cutover checklist (#379)

Jaruso:
**What changed and why**
Enforces Tardigrade's external-library policy (epic #327, story 327-J) in source, build configuration, CI, and release artifacts, so the Bare Systems appliance can be proven free of OpenSSL and any other foreign TLS/crypto implementation rather than relying on convention.

- **Build profiles.** A new `-Dtls-profile` option selects between `general` (default; links the single approved OpenSSL adapter) and `appliance` (forbids all foreign TLS/crypto linkage). `src/http/tls_backend.zig` and `src/http/acme_backend.zig` swap the OpenSSL adapters (`tls_termination.zig`, `acme_client.zig`) for no-OpenSSL stubs **at the module graph**, so an appliance binary never analyzes `@cImport("openssl/...")`, never links `libssl`/`libcrypto`, and has no runtime fallback. The stubs fail closed (`error.ContextInitFailed` / `error.AcmeProtocolError`) until the native TLS path lands under #391 — a deliberate, inspectable failure, not a silent no-op. The pure-Zig ACME `ChallengeStore` is extracted to `src/http/acme_challenge_store.zig` so both profiles and all callers share one implementation without pulling OpenSSL into the appliance graph.
- **Self-identifying artifacts.** `tardigrade version` now prints `tls-profile=…, tls-backend=…`, so an artifact reports its own profile.
- **Audit scripts.** `scripts/audit-dependencies.sh` fails if a forbidden dependency (ngtcp2, nghttp3, quiche, BoringSSL, mbedTLS, wolfSSL, GnuTLS, LibreSSL, …) is configured (comments stripped so prose can document policy), if an OpenSSL include escapes the adapter boundary, or if any `@cImport` appears in a native path (`src/tls`, `src/pki`, `src/quic`, `src/crypto`, `src/http3`). `scripts/audit-release-binary.sh` inspects a built binary's dynamic dependencies (`ldd`/`otool -L`), fails a forbidden linkage or a self-report that disagrees with actual linkage, and emits a machine-readable dependency inventory.
- **CI and release.** A required `TLS dependency audit` CI job runs the source audit, builds **both** profiles, and audits each binary's linkage, uploading the inventories. The release workflow re-runs the source audit, audits each published binary, and ships its inventory beside the SBOM.
- The stale `build.zig.zon` note claiming an optional ngtcp2/nghttp3 link path (removed in #328) is corrected. Policy and the appliance/general cutover checklist are documented in `docs/TLS_DEPENDENCY_POLICY.md`.

**Related issues**
Closes #379. Coordinates with #328 (ngtcp2/nghttp3 removal linkage checks) and #391 (appliance binary-linkage/runtime-path gate).

**Tests**
Verified locally with Zig 0.16.0:
- Appliance build links only `libc`/`libm` (`ldd`); binary audit passes with `links_openssl: false` and `tls-backend=native`.
- General build links `libssl`/`libcrypto`; binary audit passes with `tls-backend=openssl-adapter`.
- Negative check: auditing the general binary under `--profile appliance` fails with the OpenSSL-linkage and native-self-report violations, confirming the gate bites.
- Source audit passes; both audit scripts are shellcheck-clean.
- New unit tests: the TLS-termination and ACME stubs fail closed and share the pure-Zig challenge store.
- `zig build test` green under both `-Dtls-profile=general` (1119/1120, 1 skip) and `-Dtls-profile=appliance` (1117/1118, 1 skip); `zig build test-integration` green (63/67, 4 skip).

**Security impact**
Directly strengthens TLS supply-chain posture: it makes "no OpenSSL in the appliance" a machine-checked property of both the source graph and the produced binary, with no hidden runtime fallback, and forbids unapproved foreign TLS/crypto/QUIC/H3 libraries across all profiles. OpenSSL remains confined to the approved adapter boundary in the general profile (enforced by the source audit). No request-parsing, routing, auth, or logging behavior changes.

**Performance impact**
None. Build-graph selection and CI/release tooling only; no runtime code path on the event loop or hot paths changes.

**Checklist**
- [x] `zig build test --summary all` green (both profiles)
- [x] `zig build test-integration` green

---

### DIFF
--- a/scripts/audit-dependencies.sh
+++ b/scripts/audit-dependencies.sh
@@ -1,0 +1,124 @@
+#!/usr/bin/env bash
+# TLS dependency source/configuration audit (#379, epic #327).
+#
+# Enforces Tardigrade's external-library policy before anything is compiled:
+#
+#   1. Forbidden TLS/crypto/QUIC/H3 dependency names must not appear in build
+#      configuration, workflows, scripts, or packaging metadata. External
+#      implementations are allowed only as out-of-process interoperability
+#      peers (scripts/interop/), never in the link graph.
+#   2. OpenSSL headers may be included only inside the approved
+#      general-profile adapter boundary.
+#   3. Native implementation paths (TLS, PKI, QUIC, crypto, HTTP/3) must not
+#      contain any @cImport at all.
+#
+# Exit code 0 means the audit passed; any violation prints the offending
+# file/line and exits 1. Run from anywhere; paths resolve from the repo root.
+
+set -euo pipefail
+
+REPO_ROOT="$(cd "$(dirname "$0")/.." && pwd)"
+cd "$REPO_ROOT"
+
+failures=0
+
+fail() {
+    echo "AUDIT FAIL: $1" >&2
+    failures=$((failures + 1))
+}
+
+# ── 1. Forbidden dependency identifiers in build/config surfaces ─────────────
+# openssl itself is not in this list: the general profile's adapter is the
+# single approved exception, enforced separately by checks 2 and 3 and by the
+# binary audit (scripts/audit-release-binary.sh).
+#
+# The check targets configuration that would actually pull a dependency into
+# the build or link graph, so comments are stripped before matching: policy
+# permits naming these libraries in prose (e.g. documenting that external
+# peers run out-of-process). A real dependency declaration is never inside a
+# comment.
+FORBIDDEN_NAMES='ngtcp2|nghttp3|quiche|boringssl|mbedtls|wolfssl|gnutls|libressl|rustls|s2n-tls|libtls|botan'
+
+# Strip comments from a file according to its type, then print the result.
+# .zig/.zon use `//`; shell/yaml/toml/Dockerfiles use `#`. Stripping is
+# type-scoped so a URL scheme's `//` in a shell script is preserved.
+strip_comments() {
+    local file="$1"
+    case "$file" in
+    *.zig | *.zon) sed 's://.*$::' "$file" ;;
+    *.sh | *.yml | *.yaml | *.toml | *Dockerfile* | *.spec | *.control) sed 's:#.*$::' "$file" ;;
+    *) cat "$file" ;;
+    esac
+}
+
+# Build configuration, automation, and packaging surfaces. Interop tooling
+# (scripts/interop/) is excluded by policy: it drives external peers as
+# separate processes and must name them. The audit scripts are excluded
+# because they define the deny list.
+scan_files=(build.zig build.zig.zon)
+while IFS= read -r f; do
+    scan_files+=("$f")
+done < <(find .github/workflows packaging -type f 2>/dev/null | sort)
+if [ -d scripts ]; then
+    while IFS= read -r script; do
+        case "$script" in
+        scripts/interop/*) ;;
+        scripts/audit-dependencies.sh | scripts/audit-release-binary.sh) ;;
+        *) scan_files+=("$script") ;;
+        esac
+    done < <(find scripts -type f | sort)
+fi
+while IFS= read -r dockerfile; do
+    scan_files+=("$dockerfile")
+done < <(find . -path ./.zig-cache -prune -o -path ./zig-out -prune -o -name 'Dockerfile*' -type f -print | sed 's|^\./||' | sort)
+
+for file in "${scan_files[@]}"; do
+    [ -f "$file" ] || continue
+    if matches="$(strip_comments "$file" | grep -niE "$FORBIDDEN_NAMES" 2>/dev/null)"; then
+        while IFS= read -r line; do
+            [ -z "$line" ] && continue
+            fail "forbidden dependency configuration in $file: $line"
+        done <<<"$matches"
+    fi
+done
+
+# ── 2. OpenSSL includes restricted to the approved adapter boundary ──────────
+OPENSSL_INCLUDE_ALLOWLIST=(
+    src/http/tls_termination.zig
+    src/http/acme_client.zig
+)
+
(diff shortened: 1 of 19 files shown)


scripts/audit-dependencies.sh
```diff
@@ -0,0 +1,124 @@
...
+
+# Strip comments from a file according to its type, then print the result.
+# .zig/.zon use `//`; shell/yaml/toml/Dockerfiles use `#`. Stripping is
+# type-scoped so a URL scheme's `//` in a shell script is preserved.
+strip_comments() {
+    local file="$1"
+    case "$file" in
+    *.zig | *.zon) sed 's://.*$::' "$file" ;;
+    *.sh | *.yml | *.yaml | *.toml | *Dockerfile* | *.spec | *.control) sed 's:#.*$::' "$file" ;;
+    *) cat "$file" ;;
```
Jaruso: This comment stripping also removes everything after `https://` inside Zig/ZON string literals. That means a real dependency declaration such as a `.url = "https://github.com/ngtcp2/..."` is reduced to `.url = "https:` before the deny-list scan and passes unnoticed. Since `build.zig.zon` primarily declares dependencies through URLs, this bypasses the exact configuration surface the audit is intended to protect. Could this use a parser-aware approach, or at minimum scan quoted strings/URLs before stripping comments?

scripts/audit-release-binary.sh
```diff
@@ -0,0 +1,213 @@
...
+deps=""
+inspection_tool=""
+case "$os" in
+Linux)
+    inspection_tool="ldd"
+    # ldd returns nonzero for a static binary ("not a dynamic executable");
+    # that is a valid, dependency-free result, so tolerate the failure.
+    deps="$(ldd "$BINARY" 2>/dev/null || true)"
+    ;;
+Darwin)
```
Jaruso, line 115: This fails open for any `ldd`/`otool` error, not just a genuinely static binary. A missing inspection tool, unreadable/corrupt binary, loader failure, or architecture mismatch all become an empty dependency list and can produce a passing appliance audit. Please distinguish the expected static-binary result from unexpected inspection failures and reject the latter; using `readelf -d`/`objdump -p` on Linux would also avoid executing through `ldd`.

scripts/audit-release-binary.sh
```diff
@@ -0,0 +1,213 @@
...
+if printf '%s\n' "$dep_names" | grep -qiE 'libssl|libcrypto'; then
+    links_openssl=true
+fi
+
+forbidden_hits="$(printf '%s\n' "$dep_names" | grep -iE "$FORBIDDEN_LIB_PATTERNS" || true)"
+
+# ── Binary self-report (the version line records the built-in profile) ───────
+self_report="$("$BINARY" version 2>/dev/null || true)"
+reported_backend="unknown"
+case "$self_report" in
```
Jaruso, line 148: The script says it verifies the built-in profile, but it only parses `tls-backend`. An appliance binary audited with `--profile general` will pass when it reports `native` and has no OpenSSL linkage, so the requested profile can disagree with the artifact without a violation. Please parse `tls-profile=...` separately and require an exact match with `$PROFILE` for both profiles.

Jaruso (commented):
The overall profile split and artifact-audit direction are strong, but these audit scripts currently have a few fail-open cases that weaken the claim that the property is machine-enforced. I’d tighten these before relying on the inventories as release evidence.

scripts/audit-dependencies.sh
```diff
@@ -0,0 +1,124 @@
...
+    esac
+}
+
+# Build configuration, automation, and packaging surfaces. Interop tooling
+# (scripts/interop/) is excluded by policy: it drives external peers as
+# separate processes and must name them. The audit scripts are excluded
+# because they define the deny list.
+scan_files=(build.zig build.zig.zon)
```
Jaruso: This only scans the root `build.zig` and `build.zig.zon`, so forbidden linkage moved into an imported build helper is invisible. For example, `build.zig` could `@import("build/dependencies.zig")`, and that helper could call `linkSystemLibrary("gnutls")` without triggering this audit. Since Zig build logic can be split across arbitrary `.zig` files, please scan all repository build-configuration Zig sources (with explicit exclusions for normal application/docs code), or derive and inspect the imported build graph rather than assuming the root file contains every dependency declaration.

scripts/audit-dependencies.sh
```diff
@@ -0,0 +1,124 @@
...
+    fi
+done
+
+# ── 2. OpenSSL includes restricted to the approved adapter boundary ──────────
+OPENSSL_INCLUDE_ALLOWLIST=(
+    src/http/tls_termination.zig
+    src/http/acme_client.zig
+)
+
+if matches="$(grep -rn '@cInclude("openssl/' src/ 2>/dev/null)"; then
```
Jaruso: The adapter-boundary check depends on the exact token sequence `@cInclude("openssl/`. Equivalent valid spellings such as `@cInclude( "openssl/ssl.h")` are missed, and a compile-time string expression can evade it as well. That lets an OpenSSL include escape the allowlisted adapters while the audit still passes. Please make this whitespace-tolerant at minimum and consider checking every `@cInclude`/`@cImport` block for an `openssl/` header rather than matching one exact formatting form.

Jaruso (commented):
The build-profile split and module-graph selection look well structured, and CI is green. However, the new audit is the security boundary this PR is establishing, and there are currently several ways it can miss real forbidden configuration or fail open. In addition to the three existing unresolved findings on URL comment stripping, binary-inspection errors, and profile self-report matching, I found two further source-audit coverage gaps below. These should be fixed before merging so the documented guarantee matches what the scripts actually enforce.

PR #406, opened 2026-07-12
tls: bridge record epochs to traffic keys

Jaruso:
## Summary
- add a TLS record epoch bridge that maps shared TLS traffic-secret events into independent record read/write states
- support plaintext initial handshake records, protected handshake records, application-data records, and post-handshake handshake routing
- reject unsupported, duplicate, missing, and premature key transitions deterministically while discarding epoch state safely

Closes #352

## Tests
- zig build test-tls
- zig build test

### DIFF
--- a/src/tls/record_epoch_bridge.zig
+++ b/src/tls/record_epoch_bridge.zig
@@ -10,9 +10,12 @@
 const std = @import("std");
 const crypto = @import("crypto");
 const algorithms = @import("algorithms.zig");
+const engine = @import("engine.zig");
 const events = @import("events.zig");
 const record_codec = @import("record_codec.zig");
 const record_protection = @import("record_protection.zig");
+const tls_state = @import("state.zig");
+const transport = @import("transport.zig");
 
 const provider = crypto.provider;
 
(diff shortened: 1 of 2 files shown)


src/tls/record_epoch_bridge.zig
```diff
@@ -0,0 +1,380 @@
...
+    }
+
+    pub fn applyEvent(self: *Bridge, event: events.Event) Error!void {
+        switch (event) {
+            .traffic_secret => |traffic_secret| try self.installTrafficSecret(traffic_secret.epoch, traffic_secret.direction, traffic_secret.data),
+            .discard_epoch => |epoch| self.discardEpoch(epoch),
+            .handshake_complete => try self.markHandshakeComplete(),
+            .handshake_bytes,
+            .alpn,
+            .certificate,
```
Jaruso, line 77: `handshake_bytes` is the event this bridge is supposed to carry into TLS records, but it is silently discarded here. That means `applyEvent` only installs keys; it never actually connects the shared handshake engine's output to the record layer. The issue objective explicitly says to connect the engine events at the plaintext/handshake/application transitions, and the acceptance criterion requires a full handshake through records. Could `applyEvent` return or enqueue sealed records for `.handshake_bytes` based on the event epoch, or could this type expose a deterministic event-to-record API that the loopback test drives? Ignoring this event makes the bridge incomplete even though manual calls to `sealHandshake` work.

src/tls/record_epoch_bridge.zig
```diff
@@ -0,0 +1,380 @@
...
+        traffic_secret: []const u8,
+    ) Error!void {
+        switch (epoch) {
+            .handshake => switch (direction) {
+                .read => try self.installRead(&self.read_handshake, traffic_secret),
+                .write => try self.installWrite(&self.write_handshake, traffic_secret),
+            },
+            .application => switch (direction) {
+                .read => try self.installRead(&self.read_application, traffic_secret),
+                .write => try self.installWrite(&self.write_application, traffic_secret),
```
Jaruso: There is no phase/order validation here, so application secrets can be installed before any handshake secret, and a handshake secret can be installed again after `.discard_epoch(.handshake)` because discard only clears the slot. That misses the requirement that early and late transitions fail deterministically. Please track terminal epoch state (ideally independently for read/write), reject application installation before the corresponding handshake epoch has been activated, and reject any secret for an epoch after that direction has been discarded or advanced. A slot being null is not enough to distinguish “not installed yet” from “already discarded.”

src/tls/record_epoch_bridge.zig
```diff
@@ -0,0 +1,380 @@
...
+            },
+            .initial,
+            .zero_rtt,
+            => {},
+        }
+    }
+
+    pub fn markHandshakeComplete(self: *Bridge) Error!void {
+        if (self.read_application == null or self.write_application == null) return error.MissingApplicationKeys;
+        self.handshake_complete = true;
```
Jaruso, line 147: This allows `handshake_complete` after installing only the two application keys; no handshake keys or handshake-record progress are required. The test below actually relies on that behavior. For a bridge enforcing exact TLS 1.3 transitions, completion should be validated against a phase model, not just application-key presence. At minimum, require both directions to have progressed through handshake keys and into application keys, and ensure completion cannot be repeated or occur after an invalid/discarded transition.

src/tls/record_epoch_bridge.zig
```diff
@@ -0,0 +1,380 @@
...
+
+fn secret(comptime fill: u8) [32]u8 {
+    return [_]u8{fill} ** 32;
+}
+
+const testing = std.testing;
+
+test "record epoch bridge drives plaintext handshake, encrypted handshake, application data, and post-handshake routing" {
+    const cp = testProvider();
+    var client = Bridge.init(cp, .tls_aes_128_gcm_sha256);
```
Jaruso, line 317: This is a useful record-protection smoke test, but it is not the acceptance test described in #352. It manually injects arbitrary secrets and arbitrary strings such as `"client hello"`/`"client finished"`; no shared client/server TLS engine runs, no real handshake messages are parsed, and no engine events drive the bridge. Please add the deterministic in-memory loopback required by the issue: instantiate the real client and server handshake engines, shuttle emitted `handshake_bytes` through this bridge and `record_codec`, feed opened bytes back into the peer engine, and run until both emit `handshake_complete`. Also assert the first encrypted handshake and application ciphertext/nonce against an independent expected vector (or exported expected key/IV and sequence 0), because a round-trip using the same derivation on both sides can pass even if both sides derive the same wrong key.

Jaruso (commented):
I reviewed this against issue #352 rather than only the PR summary. The record-protection plumbing is clean and the independent read/write states are a good foundation, but the current implementation does not yet satisfy several P0 acceptance criteria: it does not actually route engine handshake-byte events into records, it does not enforce early/late transition ordering, and the main test is a synthetic record round-trip rather than a full client/server TLS 1.3 handshake driven by the shared engine. I left inline comments with concrete changes and test suggestions.

PR #407, opened 2026-07-12
tls: add nonblocking encrypted stream foundation

Jaruso:
## Summary
- add a backend-neutral TLS encrypted stream vtable for HTTP/1.1 and HTTP/2 consumers
- add a bounded pure-Zig record stream foundation with nonblocking plaintext read/write, carrier read/write driving, exact parser consumption, lifecycle/readiness state, fatal-error latching, and graceful close-notify shutdown
- cover fragmented handshake/application record routing, nonblocking socketpair transport, bounded drive loops, partial carrier writes, terminal-state cleanup, and vtable-shaped backend compatibility

## Scope note
This PR intentionally no longer closes #353. It provides the nonblocking pure-Zig record-stream foundation for #353, but the remaining #353 acceptance work is now tracked explicitly:

- #410: make the encrypted stream own and drive the real TLS record handshake over nonblocking socket I/O
- #411: add the production OpenSSL `EncryptedStream` adapter and shared conformance suite

Refs #353. Depends on the record-layer foundation cleanup tracked in #408 for full #353 closure.

## Tests
- zig build test-tls
- zig build test-quic
- zig build test

### DIFF
--- a/src/tls/encrypted_stream.zig
+++ b/src/tls/encrypted_stream.zig
@@ -284,6 +284,11 @@ pub const PureZigRecordStream = struct {
             return .{ .made_progress = false, .readiness = self.readiness() };
         }
 
+        if (self.lifecycle == .closing and !self.bridge.handshake_complete) {
+            if (try self.queueCloseNotify()) made_progress = true;
+            return .{ .made_progress = made_progress, .readiness = self.readiness() };
+        }
+
         if (self.carrier) |carrier| {
             var written_total: usize = 0;
             while (self.outbound_ciphertext.len > 0 and written_total < drive_write_budget) {
@@ -1134,6 +1139,50 @@ test "encrypted stream close during handshake drops queued output before closing
     try testing.expect(result.made_progress);
     try testing.expectEqual(Lifecycle.closed, stream_state.lifecycle);
     try testing.expectEqual(@as(usize, 0), stream_state.queuedCiphertextLen());
+    try testing.expectEqual(@as(usize, 1), carrier.close_count);
+
+    stream_state.deinit();
+    try testing.expectEqual(@as(usize, 1), carrier.close_count);
+}
+
+test "encrypted stream close during handshake does not write queued output" {
+    const WritableCarrier = struct {
+        written: usize = 0,
+        close_count: usize = 0,
+
+        fn carrier(self: *@This()) Carrier {
+            return .{ .ptr = self, .readFn = read, .writeFn = write, .closeFn = close, .owns_handle = true };
+        }
+
+        fn read(_: *anyopaque, _: []u8) Error!usize {
+            return error.WouldBlock;
+        }
+
+        fn write(ptr: *anyopaque, bytes: []const u8) Error!usize {
+            const self: *@This() = @ptrCast(@alignCast(ptr));
+            self.written += bytes.len;
+            return bytes.len;
+        }
+
+        fn close(ptr: *anyopaque) void {
+            const self: *@This() = @ptrCast(@alignCast(ptr));
+            self.close_count += 1;
+        }
+    };
+
+    const cp = testProvider();
+    var carrier = WritableCarrier{};
+    var stream_state = PureZigRecordStream.initWithCarrier(cp, .tls_aes_128_gcm_sha256, carrier.carrier());
+
+    try stream_state.applyEvent(.{ .handshake_bytes = .{ .epoch = .initial, .data = "queued client hello" } });
+    try testing.expect(stream_state.queuedCiphertextLen() > 0);
+
+    stream_state.stream().close();
+    const result = try stream_state.stream().drive();
+    try testing.expect(result.made_progress);
+    try testing.expectEqual(Lifecycle.closed, stream_state.lifecycle);
+    try testing.expectEqual(@as(usize, 0), stream_state.queuedCiphertextLen());
+    try testing.expectEqual(@as(usize, 0), carrier.written);
     try testing.expectEqual(@as(usize, 1), carrier.close_count);
 
     stream_state.deinit();
--- a/src/tls/encrypted_stream.zig
+++ b/src/tls/encrypted_stream.zig
@@ -120,6 +120,8 @@ pub const PureZigRecordStream = struct {
     closing: bool = false,
     closed: bool = false,
     peer_closed: bool = false,
+    close_notify_queued: bool = false,
+    failed: ?Error = null,
 
     pub fn init(crypto_provider: provider.CryptoProvider, cipher_suite: algorithms.CipherSuite) PureZigRecordStream {
         return .{ .bridge = record_epoch_bridge.Bridge.init(crypto_provider, cipher_suite) };
@@ -142,6 +144,8 @@ pub const PureZigRecordStream = struct {
         self.closing = false;
         self.closed = true;
         self.peer_closed = true;
+        self.close_notify_queued = false;
+        self.failed = null;
     }
 
     pub fn stream(self: *PureZigRecordStream) EncryptedStream {
@@ -149,21 +153,23 @@ pub const PureZigRecordStream = struct {
     }
 
     pub fn applyEvent(self: *PureZigRecordStream, event: events.Event) Error!void {
+        if (self.failed) |err| return err;
         if (self.closed) return error.StreamClosed;
         if (event == .handshake_bytes and self.outbound_ciphertext.available() < record_codec.max_ciphertext_record_len) return error.WouldBlock;
         var record_buf: [record_codec.max_ciphertext_record_len]u8 = undefined;
-        if (try self.bridge.applyEvent(event, &record_buf)) |record| {
-            try self.outbound_ciphertext.append(record);
+        if (self.bridge.applyEvent(event, &record_buf) catch |err| return self.fail(err)) |record| {
+            self.outbound_ciphertext.append(record) catch |err| return self.fail(err);
         }
     }
 
     pub fn feedHandshakeCiphertext(self: *PureZigRecordStream, epoch: events.EncryptionEpoch, bytes: []const u8) Error!usize {
+        if (self.failed) |err| return err;
         if (self.closed) return error.StreamClosed;
         if (self.inbound_handshake.available() < record_codec.max_plaintext_fragment_len) return error.WouldBlock;
         var sink = record_codec.RecordSink(1, record_codec.max_ciphertext_fragment_len){};
         const parser = self.parserForEpoch(epoch);
-        const consumed = try feedUntilOneRecord(parser, bytes, &sink);
-        try self.openHandshakeSink(epoch, &sink);
+        const consumed = feedUntilOneRecord(parser, bytes, &sink) catch |err| return self.fail(err);
+        self.openHandshakeSink(epoch, &sink) catch |err| return self.fail(err);
         return consumed;
     }
 
@@ -172,19 +178,20 @@ pub const PureZigRecordStream = struct {
     }
 
     pub fn feedCiphertext(self: *PureZigRecordStream, bytes: []const u8) Error!usize {
+        if (self.failed) |err| return err;
         if (self.closed) return error.StreamClosed;
         if (!self.canAcceptCarrierRead()) return error.WouldBlock;
         var sink = record_codec.RecordSink(1, record_codec.max_ciphertext_fragment_len){};
-        const consumed = try feedUntilOneRecord(&self.ciphertext_parser, bytes, &sink);
+        const consumed = feedUntilOneRecord(&self.ciphertext_parser, bytes, &sink) catch |err| return self.fail(err);
 
         var plaintext_buf: [record_codec.max_ciphertext_fragment_len]u8 = undefined;
         for (sink.items[0..sink.len]) |record| {
-            const opened = try self.bridge.openProtected(.application, record, &plaintext_buf);
+            const opened = self.bridge.openProtected(.application, record, &plaintext_buf) catch |err| return self.fail(err);
             switch (opened.inner.content_type) {
-                .application_data => try self.inbound_plaintext.append(opened.inner.content),
-                .handshake => try self.inbound_handshake.append(opened.inner.content),
-                .alert => try self.handleAlert(opened.inner.content),
-                .change_cipher_spec => return error.UnsupportedRecordContent,
+                .application_data => self.inbound_plaintext.append(opened.inner.content) catch |err| return self.fail(err),
+                .handshake => self.inbound_handshake.append(opened.inner.content) catch |err| return self.fail(err),
+                .alert => self.handleAlert(opened.inner.content) catch |err| return self.fail(err),
+                .change_cipher_spec => return self.fail(error.UnsupportedRecordContent),
             }
         }
         return consumed;
@@ -195,6 +202,7 @@ pub const PureZigRecordStream = struct {
     }
 
     pub fn readPlaintext(self: *PureZigRecordStream, out: []u8) Error!usize {
+        if (self.failed) |err| return err;
         if (self.closed) return error.StreamClosed;
         if (self.inbound_plaintext.read(out)) |n| return n;
         if (self.peer_closed) return error.EndOfStream;
@@ -202,6 +210,7 @@ pub const PureZigRecordStream = struct {
     }
 
     pub fn writePlaintext(self: *PureZigRecordStream, bytes: []const u8) Error!usize {
+        if (self.failed) |err| return err;
         if (self.closed or self.closing) return error.StreamClosed;
         if (bytes.len == 0) return 0;
         if (!self.bridge.handshake_complete) return error.WouldBlock;
@@ -209,8 +218,8 @@ pub const PureZigRecordStream = struct {
 
         const n = @min(bytes.len, record_codec.max_plaintext_fragment_len);
         var record_buf: [record_codec.max_ciphertext_record_len]u8 = undefined;
-        const record = try self.bridge.sealApplicationData(bytes[0..n], &record_buf);
-        try self.outbound_ciphertext.append(record);
+        const record = self.bridge.sealApplicationData(bytes[0..n], &record_buf) catch |err| return self.fail(err);
+        self.outbound_ciphertext.append(record) catch |err| return self.fail(err);
         return n;
     }
 
@@ -226,9 +235,15 @@ pub const PureZigRecordStream = struct {
 
     pub fn consumeCiphertext(self: *PureZigRecordStream, count: usize) Error!void {
         try self.outbound_ciphertext.discard(count);
+        if (self.closing and self.carrier == null and self.close_notify_queued and self.outbound_ciphertext.len == 0) {
+            self.closed = true;
+        }
     }
 
     pub fn readiness(self: *const PureZigRecordStream) Readiness {
+        if (self.failed != null) {
+            return .{ .peer_closed = self.peer_closed };
+        }
         return .{
             .wants_read = self.canAcceptCarrierRead(),
             .wants_write = self.outbound_ciphertext.len > 0 or (self.closing and !self.closed),
@@ -239,13 +254,19 @@ pub const PureZigRecordStream = struct {
     }
 
     pub fn drive(self: *PureZigRecordStream) Error!DriveResult {
+        if (self.failed) |err| return err;
         var made_progress = false;
+        if (self.closing and self.outbound_ciphertext.len == 0 and self.close_notify_queued) {
+            self.closed = true;
+            self.closeCarrier();
+            return .{ .made_progress = true, .readiness = self.readiness() };
+        }
         if (try self.queueCloseNotify()) made_progress = true;
         if (self.carrier) |carrier| {
             if (self.outbound_ciphertext.len > 0) {
                 const written = carrier.write(self.peekCiphertext()) catch |err| switch (err) {
                     error.WouldBlock => 0,
-                    else => return err,
+                    else => return self.fail(err),
                 };
                 if (written > 0) {
                     try self.consumeCiphertext(written);
@@ -266,14 +287,14 @@ pub const PureZigRecordStream = struct {
                         made_progress = true;
                         break :eof null;
                     },
-                    else => return err,
+                    else => return self.fail(err),
                 };
                 if (maybe_read_len) |read_len| {
                     if (read_len == 0) {
                         self.peer_closed = true;
                         made_progress = true;
                     } else {
-                        const consumed = try self.feedCiphertext(buf[0..read_len]);
+                        const consumed = try self.feedCarrierCiphertext(buf[0..read_len]);
                         std.debug.assert(consumed == read_len);
                         made_progress = true;
                     }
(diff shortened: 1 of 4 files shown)


src/tls/encrypted_stream.zig
```diff
@@ -0,0 +1,542 @@
...
+    pub fn stream(self: *PureZigRecordStream) EncryptedStream {
+        return .{ .ptr = self, .vtable = &pure_zig_record_vtable };
+    }
+
+    pub fn applyEvent(self: *PureZigRecordStream, event: events.Event) Error!void {
+        if (self.closed) return error.StreamClosed;
+        var record_buf: [record_codec.max_ciphertext_record_len]u8 = undefined;
+        if (try self.bridge.applyEvent(event, &record_buf)) |record| {
+            try self.outbound_ciphertext.append(record);
+        }
```
Jaruso, line 180: `bridge.applyEvent()` seals `handshake_bytes` immediately, which advances the selected write-state sequence number. If `outbound_ciphertext.append()` then returns `CiphertextBufferFull`, the record is lost but the sequence has already advanced, so retrying the same event produces a different nonce and a permanent gap. Please make this operation atomic with respect to backpressure—e.g. preflight enough queue space before sealing, or retain the produced record as pending state until it is fully queued.

src/tls/encrypted_stream.zig
```diff
@@ -0,0 +1,542 @@
...
+    pub fn feedCiphertext(self: *PureZigRecordStream, bytes: []const u8) Error!usize {
+        if (self.closed) return error.StreamClosed;
+        var sink = record_codec.RecordSink(8, record_codec.max_ciphertext_fragment_len * 8){};
+        try self.ciphertext_parser.feed(bytes, &sink);
+
+        var plaintext_buf: [record_codec.max_ciphertext_fragment_len]u8 = undefined;
+        for (sink.items[0..sink.len]) |record| {
+            const opened = try self.bridge.openApplicationData(record, &plaintext_buf);
+            try self.inbound_plaintext.append(opened.inner.content);
+        }
```
Jaruso, line 217: Routing every encrypted record through `openApplicationData()` means valid TLS 1.3 records whose inner type is `handshake` (post-handshake NewSessionTicket/KeyUpdate) or `alert` (`close_notify`) are rejected as `UnexpectedRecordContent`. That conflicts with the issue’s handshake-progress and graceful-shutdown scope. The stream needs to decrypt once and dispatch on the recovered inner content type, forwarding handshake bytes to the engine, application data to the plaintext queue, and alerts to shutdown/error state.

src/tls/encrypted_stream.zig
```diff
@@ -0,0 +1,542 @@
...
+        return bytes.len;
```
Jaruso: This append can fail only after the parser has consumed the record and `openApplicationData()` has successfully advanced the read sequence. A full plaintext queue therefore drops authenticated plaintext and leaves the connection unable to replay the record. The same pattern exists in `openHandshakeSink`. Please apply backpressure before consuming/decrypting a complete record, or persist the decoded record/plaintext in stream-owned pending storage so queue-full is retryable without advancing cryptographic state twice.

src/tls/encrypted_stream.zig
```diff
@@ -0,0 +1,542 @@
...
+        const n = @min(bytes.len, record_codec.max_plaintext_fragment_len);
+        var record_buf: [record_codec.max_ciphertext_record_len]u8 = undefined;
+        const record = try self.bridge.sealApplicationData(bytes[0..n], &record_buf);
+        try self.outbound_ciphertext.append(record);
+        return n;
+    }
+
+    pub fn drainCiphertext(self: *PureZigRecordStream, out: []u8) Error!usize {
+        if (self.outbound_ciphertext.read(out)) |n| return n;
+        if (self.closed) return error.StreamClosed;
```
Jaruso: A destructive `drainCiphertext()` is not sufficient for nonblocking socket writes: `write(2)` may return `EAGAIN` or accept only part of the drained slice, after which the removed bytes must still be retained. The test helper currently treats a partial write as a fatal assertion, but real event-loop behavior must preserve the remainder. Consider a `peekCiphertext()` + `consumeCiphertext(n)` contract, or have `drive()` own the write callback and pending offset.

src/tls/encrypted_stream.zig
```diff
@@ -0,0 +1,542 @@
...
+            .wants_write = self.outbound_ciphertext.len > 0,
+            .can_read_plaintext = self.inbound_plaintext.len > 0,
+            .can_write_plaintext = !self.closed and self.bridge.handshake_complete and self.outbound_ciphertext.available() >= record_codec.max_ciphertext_record_len,
+            .peer_closed = self.peer_closed,
+        };
+    }
+
+    pub fn drive(self: *PureZigRecordStream) Error!DriveResult {
+        return .{ .made_progress = false, .readiness = self.readiness() };
+    }
```
Jaruso, line 373: `drive()` never drives anything, and the common `EncryptedStream` interface has no fd or injected nonblocking read/write callbacks. As a result, socket reads, parser feeding, ciphertext writes, and handshake-event exchange all remain in backend-specific external helpers. That does not yet satisfy the requirement to keep event-loop registration and handshake/record progress behind a stable interface. Please make `drive()` perform a bounded amount of carrier and handshake work based on readiness, returning `made_progress=true` only when state or queues actually change.

src/tls/encrypted_stream.zig
```diff
@@ -0,0 +1,542 @@
...
+    const fds = try testSocketPair();
+    defer _ = std.c.close(fds[0]);
+    defer _ = std.c.close(fds[1]);
+
+    const cp = testProvider();
+    var client = PureZigRecordStream.init(cp, .tls_aes_128_gcm_sha256);
+    defer client.deinit();
+    var server = PureZigRecordStream.init(cp, .tls_aes_128_gcm_sha256);
+    defer server.deinit();
+    try establish(&client, &server);
```
Jaruso, line 867: This socket-pair test starts after `establish()` manually injects four arbitrary secrets and marks both handshakes complete, so no TLS handshake crosses the socket pair. Issue #353 explicitly requires the socket-pair test to complete the handshake and bidirectional application exchange under fragmented reads/writes. Please drive the real client/server engines, their emitted traffic-secret/handshake events, and the record stream over these nonblocking fds before sending application data; this should also exercise failure/cancellation cleanup.

src/tls/encrypted_stream.zig
```diff
@@ -0,0 +1,542 @@
...
+    try testing.expect(!before.wants_write);
+    try testing.expect(!before.can_read_plaintext);
+    try testing.expect(!before.can_write_plaintext);
+
+    const drive = try stream.drive();
+    try testing.expect(!drive.made_progress);
+    try testing.expectEqual(before, drive.readiness);
+}
+
+test "encrypted stream interface accepts OpenSSL-like and pure-Zig backends" {
```
Jaruso: This proves only that a test double can populate the vtable; it does not show that the existing OpenSSL adapter can implement the same EOF, would-block, readiness, close, and partial-write semantics. Since backend interchangeability is an acceptance criterion, could this wrap the real OpenSSL connection adapter (or a thin production adapter around it) and run the same behavioral conformance tests against both backends?

Jaruso (commented):
I reviewed this against issue #353’s P0 acceptance criteria. The backend-neutral surface is a useful starting point, but the current implementation still leaves transport driving, handshake progression, partial-write retention, fatal-state handling, and graceful shutdown outside the abstraction. There are also two cryptographic-state corruption paths where queue backpressure is reported only after a record has already been sealed/opened and its sequence number advanced. The socket-pair test bypasses the TLS handshake, and the OpenSSL compatibility test uses a fake backend rather than the existing adapter. CI is currently failing in both Linux unit-test jobs (ubuntu-latest and ubuntu-24.04-arm), while macOS passes.

src/tls/encrypted_stream.zig
```diff
@@ -0,0 +1,776 @@
...
+
+    pub fn readHandshake(self: *PureZigRecordStream, out: []u8) Error!usize {
+        return self.inbound_handshake.read(out) orelse error.WouldBlock;
+    }
+
+    pub fn feedCiphertext(self: *PureZigRecordStream, bytes: []const u8) Error!usize {
+        if (self.closed) return error.StreamClosed;
+        if (self.inbound_plaintext.available() < record_codec.max_plaintext_fragment_len or
+            self.inbound_handshake.available() < record_codec.max_plaintext_fragment_len)
+        {
```
Jaruso: This preflight guarantees capacity for only one maximum-size plaintext record, but `Parser.feed()` can emit up to eight coalesced records into `sink`. If one call contains multiple records whose aggregate plaintext exceeds the remaining queue, a later record is still opened (advancing the read sequence) before its append fails, recreating the non-retryable state corruption this check is intended to prevent. `feedHandshakeCiphertext()` has the same issue. Please process/decrypt one record at a time with pending storage, or reserve capacity for the entire emitted batch before opening any record; add a coalesced-record backpressure test.

src/tls/encrypted_stream.zig
```diff
@@ -0,0 +1,776 @@
...
+
+    pub fn consumeCiphertext(self: *PureZigRecordStream, count: usize) Error!void {
+        try self.outbound_ciphertext.discard(count);
+    }
+
+    pub fn readiness(self: *const PureZigRecordStream) Readiness {
+        return .{
+            .wants_read = !self.closed and !self.peer_closed and self.inbound_plaintext.available() > 0,
+            .wants_write = self.outbound_ciphertext.len > 0,
+            .can_read_plaintext = self.inbound_plaintext.len > 0,
```
Jaruso, line 274: `wants_read` no longer matches the condition under which `drive()` will actually call the carrier. Here it stays true whenever the plaintext queue has any space, while `drive()` requires *both* plaintext and handshake queues to have a full-fragment of space. If the handshake queue is partially full and the fd remains readable, an event loop can repeatedly call `drive()`, get `made_progress=false`, and still receive `wants_read=true`, which is exactly the busy-loop case in the acceptance criteria. Please derive both from one shared `canAcceptCarrierRead()` predicate.

src/tls/encrypted_stream.zig
```diff
@@ -0,0 +1,776 @@
...
+                }
+            }
+            if (!self.peer_closed and
+                (self.inbound_plaintext.available() >= record_codec.max_plaintext_fragment_len and
+                    self.inbound_handshake.available() >= record_codec.max_plaintext_fragment_len))
+            {
+                var buf: [record_codec.max_ciphertext_record_len]u8 = undefined;
+                const read_len = carrier.read(&buf) catch |err| switch (err) {
+                    error.WouldBlock => 0,
+                    error.EndOfStream => blk: {
```
Jaruso: The carrier contract returns `usize`, and a normal fd-backed `read(2)` returns `0` on orderly EOF. `drive()` only recognizes `error.EndOfStream`; a zero result falls through as no progress and never sets `peer_closed`, so an actual socket adapter built around the helper below will miss EOF indefinitely. Either specify and enforce that every carrier normalizes zero to `EndOfStream`, or handle `read_len == 0` here as EOF (with a test using a closed socket peer).

src/tls/encrypted_stream.zig
```diff
@@ -0,0 +1,776 @@
...
+}
+
+fn pureWrite(ptr: *anyopaque, bytes: []const u8) Error!usize {
+    const self: *PureZigRecordStream = @ptrCast(@alignCast(ptr));
+    return self.writePlaintext(bytes);
+}
+
+fn pureClose(ptr: *anyopaque) void {
+    const self: *PureZigRecordStream = @ptrCast(@alignCast(ptr));
+    self.closed = true;
```
Jaruso: This is not yet graceful TLS shutdown: `close()` only flips a flag, does not enqueue an application-epoch `close_notify`, and `drive()` does not check `closed` before continuing carrier reads/writes. The peer can therefore see a bare TCP close, while this side may keep doing I/O after the public stream reports closed. Consider explicit `open -> closing -> closed` state: seal and queue `close_notify`, flush pending ciphertext, stop accepting plaintext, then finish/close the carrier. Please cover partial-write and cancellation during shutdown.

src/tls/encrypted_stream.zig
```diff
@@ -0,0 +1,776 @@
...
+    pub fn drive(self: EncryptedStream) Error!DriveResult {
+        return self.vtable.driveFn(self.ptr);
+    }
+};
+
+pub const Carrier = struct {
+    ptr: *anyopaque,
+    readFn: *const fn (*anyopaque, []u8) Error!usize,
+    writeFn: *const fn (*anyopaque, []const u8) Error!usize,
+
```
Jaruso, line 106: There is still no carrier close/cancel/deinit operation, and `PureZigRecordStream.deinit()` never releases carrier resources. The socket-pair test closes both fds externally only on the success path, so the issue's FD-cleanup requirement cannot be verified through the abstraction on initialization failure, handshake failure, or cancellation. If `Carrier` is intentionally borrowed, document that ownership contract and add failure/cancellation tests proving the owner always closes it; otherwise add an ownership/close hook.

Jaruso (commented):
Follow-up review after `tls: harden encrypted stream backpressure`. The update addresses the original single-record write preflight, protected-record dispatch, partial-write retention, Linux test failure, and adds bounded carrier I/O; CI is now green. I still see correctness gaps around coalesced-record backpressure, readiness/drive agreement, EOF normalization, graceful local shutdown, and carrier ownership. The prior acceptance-criteria comments about completing the real TLS handshake over the socket pair and exercising a production OpenSSL adapter also remain unresolved.

src/tls/encrypted_stream.zig
```diff
@@ -0,0 +1,938 @@
...
+                    else => return err,
+                };
+                if (maybe_read_len) |read_len| {
+                    if (read_len == 0) {
+                        self.peer_closed = true;
+                        made_progress = true;
+                    } else {
+                        const consumed = try self.feedCiphertext(buf[0..read_len]);
+                        std.debug.assert(consumed == read_len);
+                        made_progress = true;
```
Jaruso, line 350: `drive()` still routes every carrier byte through `feedCiphertext()`, which decrypts with `openProtected(.application, ...)`. Before `handshake_complete`, this cannot accept either plaintext ClientHello/ServerHello records or handshake-epoch ciphertext, so the carrier-backed stream still cannot perform the socket-pair TLS handshake required by #353. The driver needs explicit record/epoch progression (initial → handshake → application), must deliver recovered handshake bytes to the TLS engine, and must apply the engine's emitted events before application traffic is enabled.

src/tls/encrypted_stream.zig
```diff
@@ -0,0 +1,938 @@
...
+        return self.inbound_handshake.read(out) orelse error.WouldBlock;
+    }
+
+    pub fn feedCiphertext(self: *PureZigRecordStream, bytes: []const u8) Error!usize {
+        if (self.closed) return error.StreamClosed;
+        if (!self.canAcceptCarrierRead()) return error.WouldBlock;
+        var sink = record_codec.RecordSink(1, record_codec.max_ciphertext_fragment_len){};
+        const consumed = try feedUntilOneRecord(&self.ciphertext_parser, bytes, &sink);
+
+        var plaintext_buf: [record_codec.max_ciphertext_fragment_len]u8 = undefined;
```
Jaruso, line 208: This is now a partial-consumption API: `feedUntilOneRecord()` can return after consuming only a prefix of `bytes`. However existing callers such as `pumpCiphertext`, `pumpHandshake`, and `readFdIntoStream` ignore the returned count and discard the whole supplied slice. If one TCP fragment completes a record and also contains bytes from the next record, that suffix is lost. Please either retain unconsumed input inside the stream or document the contract and update every caller to loop/preserve `bytes[consumed..]`; a test should split a read across a record boundary while including the start of the next record.

src/tls/encrypted_stream.zig
```diff
@@ -0,0 +1,938 @@
...
+        for (sink.items[0..sink.len]) |record| {
+            const opened = try self.bridge.openProtected(.application, record, &plaintext_buf);
+            switch (opened.inner.content_type) {
+                .application_data => try self.inbound_plaintext.append(opened.inner.content),
```
Jaruso: A parser, AEAD, record-content, or carrier error is returned without transitioning the stream into a terminal failure state. At this point the parser may already have consumed the record, so retrying is not recoverable, yet subsequent `drive()`/readiness calls can continue as though the stream were healthy. Since #353 explicitly requires defined fatal-error behavior, please latch the failure reason, stop further reads/writes, and close an owned carrier (or require explicit cancellation) consistently; add malformed-record and authentication-failure cleanup tests.

src/tls/encrypted_stream.zig
```diff
@@ -0,0 +1,938 @@
...
+    }
+
+    fn canAcceptCarrierRead(self: *const PureZigRecordStream) bool {
+        return !self.closed and !self.closing and !self.peer_closed and
+            self.inbound_plaintext.available() >= record_codec.max_plaintext_fragment_len and
+            self.inbound_handshake.available() >= record_codec.max_plaintext_fragment_len;
+    }
+
+    fn queueCloseNotify(self: *PureZigRecordStream) Error!bool {
+        if (!self.closing or self.closed or self.outbound_ciphertext.len > 0) return false;
```
Jaruso: There is still a close-state loop in manual/no-carrier mode. `drive()` queues `close_notify`; the caller drains it with `drainCiphertext()`/`consumeCiphertext()`; then the next `drive()` sees `closing && outbound empty` and queues a second `close_notify` before reaching the no-carrier completion branch. The stream can therefore never become `closed`. Please track whether the alert has already been queued/sent, or finalize closure when the closing queue is consumed, and cover `init()` + manual drain in a shutdown test.

Jaruso (commented):
Third-pass review against issue #353. This update fixes a meaningful set of prior findings: one-record backpressure is retry-safe, readiness and carrier-read gating share a predicate, zero-length reads become EOF, owned carriers have a cleanup hook, and graceful local close now queues and flushes `close_notify`. I resolved the corresponding threads, and CI is green.

The remaining acceptance blockers are still the real socket-pair TLS handshake, production OpenSSL conformance, and failure/cancellation coverage. I also found four concrete runtime issues in the new implementation below.

src/tls/encrypted_stream.zig
```diff
@@ -0,0 +1,1180 @@
...
+                    else => return self.fail(err),
+                };
+                if (written > 0) {
+                    try self.consumeCiphertext(written);
+                    made_progress = true;
+                }
+            }
+            if (self.closing and self.outbound_ciphertext.len == 0) {
+                self.closed = true;
+                self.closeCarrier();
```
Jaruso, line 322: Closing while application ciphertext is already queued still skips `close_notify`. `queueCloseNotify()` runs before this write and returns `false` because the queue is nonempty. If this write drains the remaining application bytes, this branch immediately closes the carrier even though `close_notify_queued` is still false. Please only finalize carrier closure after the alert has actually been queued and flushed; when ordinary ciphertext drains, queue `close_notify` before closing. Add a regression test that calls `write()`, then `close()`, with both full and partial carrier writes.

src/tls/encrypted_stream.zig
```diff
@@ -0,0 +1,1180 @@
...
+                made_progress = true;
+            }
+            if (self.canAcceptCarrierRead()) {
+                var buf: [1]u8 = undefined;
+                const maybe_read_len = carrier.read(&buf) catch |err| switch (err) {
+                    error.WouldBlock => null,
```
Jaruso, line 335: Reading exactly one carrier byte per `drive()` is not a viable readiness contract. A maximum TLS record requires roughly 16.6k drive calls, and an edge-triggered poller may not produce another notification because the fd remains readable after this call. Please read/process a bounded chunk or byte budget until `WouldBlock` or a handshake boundary, preserving any unconsumed suffix. The byte-at-a-time workaround is also the parser-contract debt now tracked in #408.

src/tls/encrypted_stream.zig
```diff
@@ -0,0 +1,1180 @@
...
+        var sink = record_codec.RecordSink(1, record_codec.max_ciphertext_fragment_len){};
+        const parser = self.parserForEpoch(epoch);
+        const consumed = feedUntilOneRecord(parser, bytes, &sink) catch |err| return self.fail(err);
+        self.openHandshakeSink(epoch, &sink) catch |err| return self.fail(err);
+        return consumed;
+    }
+
+    pub fn readHandshake(self: *PureZigRecordStream, out: []u8) Error!usize {
+        return self.inbound_handshake.read(out) orelse error.WouldBlock;
+    }
```
Jaruso, line 199: The new fatal latch does not cover all public stream operations. `readHandshake()` can still return previously queued handshake bytes after a parser/authentication failure, allowing the external engine to keep progressing a connection that is already terminal; `drainCiphertext()`/`peekCiphertext()` likewise leave stale output accessible. Please gate or clear every queue-facing helper on failure and release/reset the bridge/parser state consistently. A useful regression is: queue valid handshake input, trigger a fatal record, then verify no handshake or ciphertext can be consumed afterward.

Jaruso (commented):
The latest update fixes the internal partial-consumption callers and the repeated manual-close loop, and CI is green. I found three remaining runtime problems in the new implementation. The pre-application carrier path still does not integrate the handshake engine, the real socket-pair handshake and production OpenSSL conformance requirements remain open, and #408 still blocks treating this PR as satisfying #353.

src/tls/encrypted_stream.zig
```diff
@@ -0,0 +1,1381 @@
...
+
+    fn canAcceptCarrierRead(self: *const PureZigRecordStream) bool {
+        return self.lifecycle != .closed and self.lifecycle != .failed and self.lifecycle != .closing and !self.peer_closed and
+            self.inbound_carrier.available() > 0 and
+            self.inbound_plaintext.available() >= record_codec.max_plaintext_fragment_len and
+            self.inbound_handshake.available() >= record_codec.max_plaintext_fragment_len;
+    }
+
+    fn queueCloseNotify(self: *PureZigRecordStream) Error!bool {
+        if (self.lifecycle != .closing or self.close_notify_queued or self.outbound_ciphertext.len > 0) return false;
```
Jaruso: The new handshake-shutdown cleanup is still blocked by queued output. If the handshake has emitted a record, `close()` is called, and the carrier keeps returning `WouldBlock`, this condition returns before the `!handshake_complete` branch, so the owned fd is never closed and cancellation can stall indefinitely. Before handshake completion there is no `close_notify` to flush; please make closing abortive regardless of `outbound_ciphertext.len` (clear the queued handshake bytes and close the carrier), or explicitly provide a separate cancellation operation. A regression test should queue `handshake_bytes`, use a carrier whose write always would-block, then verify `close()` + `drive()` reaches `closed` and calls `closeFn` exactly once.

Jaruso (commented):
Follow-up review of the latest handshake-shutdown commit. CI is green, and the owned carrier now closes exactly once when shutdown begins with an empty outbound queue. One backpressured cancellation path still prevents the new cleanup guarantee from taking effect; see the inline comment.

src/tls/encrypted_stream.zig
Jaruso: Fixed in 0334eb8. Pre-handshake close now clears queued handshake ciphertext, closes the owned carrier abortively, and reaches `.closed`; added the blocking-carrier regression for this case.

PR #409, opened 2026-07-13
pki: implement certificate signature verification matrix (#343)

Jaruso:
## What changed and why

Adds `src/pki/verify.zig` for story 324-E of the Web PKI epic (#324). The module verifies X.509 certificate signatures through the shared crypto-provider seam and parsed X.509 views, without naming cryptographic primitives in PKI code or falling back implicitly to OpenSSL.

The merged provider-backed verification matrix is:

- **Ed25519** — supported.
- **ECDSA P-256/SHA-256** — supported.
- **RSA-PSS-RSAE/SHA-256** — parsed, parameter-validated, and rejected fail-closed as unsupported because Zig 0.16's standard-library EMSA-PSS verifier does not validate the complete PS zero-padding region required by RFC 8017. Conformant implementation work is tracked in #413.

`verifyCertificateSignature` validates the certificate signature algorithm and issuer SubjectPublicKeyInfo before provider delegation. This includes:

- Ed25519 and ECDSA signature parameters must be absent.
- Ed25519 SPKI parameters must be absent; `rsaEncryption` parameters must be absent or NULL.
- ECDSA public keys use valid compressed or uncompressed SEC1 encodings.
- ECDSA signatures use canonical DER and P-256-sized scalar encodings.
- RSA public-key DER has strictly positive modulus and exponent values.
- RSA-PSS parameters are constrained to SHA-256, MGF1-SHA-256, and a 32-byte salt.
- Signature and public-key BIT STRINGs must be octet-aligned.

Errors remain typed and fail closed: unsupported algorithm/configuration, issuer-key mismatch, malformed public key, malformed signature, and cryptographically invalid signature are distinct outcomes.

## Provider and capability changes

The pure-Zig provider gains ECDSA-P256/SHA-256 verification over `std.crypto`. The capability profile and TLS policy projection advertise Ed25519 and ECDSA-P256/SHA-256. RSA-PSS remains deliberately absent from provider capabilities until #413 supplies a conformant implementation.

The secp256r1 ECDH group remains deferred; this PR implements only the signature scheme over that curve.

## Tests

Coverage includes:

- OpenSSL-generated Ed25519 and P-256 self-signed certificate fixtures.
- Tampered TBS data, wrong issuer keys, and issuer-key-type mismatch.
- Unsupported PKCS#1 v1.5 signatures.
- RSA-PSS parameter validation and fail-closed capability behavior.
- Missing or forbidden AlgorithmIdentifier parameters.
- Non-positive RSA modulus/exponent values.
- Non-canonical and oversized ECDSA scalar encodings.
- Malformed BIT STRING padding and short malformed provider inputs.

The final head passed the complete CI workflow, including Linux and macOS unit tests, formatting, security checks, packaging, and the TLS dependency audit.

## Security impact

This is security-sensitive PKI code operating on untrusted certificates. Inputs are structurally validated before supported provider operations. Unsupported schemes fail closed, and RSA-PSS is not advertised while the available Zig 0.16 implementation has known verification strictness gaps.

## Related issues

- #343
- #413
- #324
- #327


src/pki/verify.zig
```diff
@@ -0,0 +1,270 @@
...
+pub fn verifySelfSignature(
+    crypto_provider: provider.CryptoProvider,
+    certificate: *const x509.Certificate,
+) Error!void {
+    return verifyCertificateSignature(crypto_provider, certificate, &certificate.subject_public_key_info);
+}
+
+fn resolveScheme(algorithm: *const x509.AlgorithmIdentifier) Error!Resolved {
+    return switch (x509.SignatureAlgorithm.classify(algorithm)) {
+        .ed25519 => .{ .scheme = .ed25519, .required_key_type = .ed25519 },
```
Jaruso: Ed25519 and `ecdsa-with-SHA256` have parameter rules too. RFC 8410 requires Ed25519 `AlgorithmIdentifier.parameters` to be absent, and RFC 5758 requires the ECDSA-with-SHA256 parameters to be absent. `SignatureAlgorithm.classify` only checks the OID, so a certificate carrying NULL or arbitrary parameters can currently reach verification. Please reject `parameters_raw != null` for these two cases and add negative tests.

src/pki/verify.zig
```diff
@@ -0,0 +1,270 @@
...
+        // parser did not recognize are outside the matrix.
+        else => error.UnsupportedSignatureAlgorithm,
+    };
+}
+
+fn issuerPublicKey(scheme: provider.SignatureScheme, issuer: *const x509.SubjectPublicKeyInfo) Error![]const u8 {
+    // A public key's BIT STRING must be octet-aligned.
+    if (issuer.subject_public_key.unused_bits != 0) return error.MalformedPublicKey;
+    const key = issuer.subject_public_key.data;
+    switch (scheme) {
```
Jaruso, line 151: The `key_type` classification does not fully validate the issuer SPKI `AlgorithmIdentifier`. The parser classifies Ed25519 and `rsaEncryption` by OID regardless of their parameters, so malformed SPKIs can reach the provider. Please enforce Ed25519 parameters being absent and an explicit RSA parameter policy here or in the X.509 parser (at minimum reject arbitrary non-NULL values), with regression tests.

src/pki/verify.zig
```diff
@@ -0,0 +1,270 @@
...
+    if (r.isNegative() or s.isNegative()) return error.MalformedSignature;
+}
+
+/// Return the RSA modulus length in bytes (minimal, leading-zero stripped)
+/// from a DER `RSAPublicKey`, validating the two-INTEGER structure.
+fn rsaModulusLen(rsa_public_key_der: []const u8) Error!usize {
+    var reader = der.Reader.init(rsa_public_key_der, .{});
+    var seq = reader.readSequence() catch return error.MalformedPublicKey;
+    reader.expectEnd() catch return error.MalformedPublicKey;
+    const modulus = seq.readInteger() catch return error.MalformedPublicKey;
```
Jaruso, line 235: `readInteger()` enforces minimal DER encoding but does not require a positive value. A negative modulus or exponent therefore passes this structural check, and `std.crypto.Certificate.rsa.PublicKey.parseDer` later treats the raw bytes as unsigned. That can accept an `RSAPublicKey` OpenSSL would reject. Please retain the exponent view and reject negative/zero modulus and exponent values before delegation.

src/pki/verify.zig
```diff
@@ -0,0 +1,270 @@
...
+/// for direct testing of the signature-encoding rules.
+pub fn validateEcdsaDerSignature(sig: []const u8) Error!void {
+    var reader = der.Reader.init(sig, .{});
+    var seq = reader.readSequence() catch return error.MalformedSignature;
+    reader.expectEnd() catch return error.MalformedSignature;
+    const r = seq.readInteger() catch return error.MalformedSignature;
+    const s = seq.readInteger() catch return error.MalformedSignature;
+    seq.expectEnd() catch return error.MalformedSignature;
+    // ECDSA r and s are positive.
+    if (r.isNegative() or s.isNegative()) return error.MalformedSignature;
```
Jaruso: Suggestion: also bound each P-256 scalar INTEGER to at most 32 magnitude bytes (33 only when the first byte is the required sign-padding zero). A canonical positive INTEGER larger than that passes this validator, then `Signature.fromDer` returns `InvalidInput`; the caller currently maps that to `MalformedPublicKey`, so the advertised error taxonomy is lost.

src/crypto/pure_zig.zig
```diff
@@ -388,7 +397,36 @@ fn verifyImpl(
...
+            const sig = EcdsaP256Sha256.Signature.fromDer(signature) catch return error.InvalidInput;
+            sig.verify(message, pk) catch return error.AuthenticationFailed;
+        },
+        .rsa_pss_rsae_sha256 => {
+            // `public_key` is the DER `RSAPublicKey` (SEQUENCE { modulus,
+            // publicExponent }) carried in the SPKI subjectPublicKey;
+            // `signature` is the raw EMSA-PSS signature, one modulus in
+            // length. SHA-256 is both the message hash and the MGF1 hash,
+            // with the salt length equal to the digest length (rsae).
+            const components = rsa.PublicKey.parseDer(public_key) catch return error.InvalidInput;
```
Jaruso: This stdlib DER helper is not total over arbitrary untrusted slices: its underlying `der.Element.parse` indexes the first two bytes before performing bounds checks. An empty or one-byte `public_key` can therefore panic before this `catch` runs. Now that RSA verification is advertised by the provider, `verify` should return `InvalidInput` for every byte slice. Please add explicit bounded pre-validation (ideally the project DER reader) and tests for empty, one-byte, truncated-length, and truncated-content inputs.

src/crypto/pure_zig.zig
```diff
@@ -388,7 +397,36 @@ fn verifyImpl(
...
+            if (components.exponent.len > components.modulus.len) return error.InvalidInput;
+            if (signature.len != components.modulus.len) return error.InvalidInput;
+            const pk = rsa.PublicKey.fromBytes(components.exponent, components.modulus) catch return error.InvalidInput;
+            switch (components.modulus.len) {
```
Jaruso: Byte length alone does not establish a 2048/3072/4096-bit modulus. For example, a minimal 256-byte modulus beginning with `0x01` is only 2041 bits. The stdlib PSS verifier then derives an encoded-message length smaller than `modulus_len` and hits its `debug.assert(emLen == em.len)` (with inconsistent behavior when assertions are removed). Please validate the exact modulus bit length before dispatch and add a low-top-bit modulus regression test.

src/crypto/pure_zig.zig
```diff
@@ -388,7 +397,36 @@ fn verifyImpl(
...
+                inline 256, 384, 512 => |modulus_len| {
+                    rsa.PSSSignature.verify(modulus_len, signature[0..modulus_len].*, message, pk, Sha256) catch
```
Jaruso: There is also a strictness issue in the current Zig 0.16 `std.crypto.Certificate.rsa.PSSSignature` implementation: its EMSA-PSS step 10 checks only the final octet of the zero-padding `PS` region, not that the entire region is zero as RFC 8017 requires. A holder of the private key can therefore create a malformed PSS encoding accepted here but rejected by OpenSSL. Since this PR promotes the scheme to supported and claims differential behavior, please add a crafted negative vector and either harden/patch the verifier before advertising support or keep this capability deferred.

Jaruso (commented):
Security-focused review of the new certificate verification paths. I found several fail-closed edge cases around AlgorithmIdentifier validation, RSA DER parsing, exact modulus sizing, and the current std.crypto PSS verifier. The inline comments include concrete regression cases and suggested fixes.

Jaruso (commented):
Second pass complete. The new commit addresses all seven security findings from the first review, and deferring RSA-PSS rather than relying on Zig 0.16's non-conformant EMSA-PSS verifier is the right fail-closed decision.

Two items remain before merge:

1. The PR description and `Closes #343` are now stale. They still say RSA-PSS verification is implemented, advertised, and successfully verifies the fixture, while the current provider intentionally returns `UnsupportedCapability` and the profile leaves the scheme unavailable. Either update the PR scope/title/body and keep #343 open (or create a clearly linked follow-up), or complete RSA-PSS with a conformant verifier before closing the story.
2. CI run 544 is currently failing in the macOS unit-test job, while the Ubuntu unit tests and the other visible jobs pass. That failure should be understood or successfully rerun before merge.

Aside from those scope/CI items, I did not find another security defect in the revised verification paths.

src/pki/verify_tests.zig
```diff
@@ -0,0 +1,401 @@
...
+    // fails it closed rather than mis-verifying it.
+    var rp = try Loaded.init(allocator, rsa_pss_crt);
+    defer rp.deinit(allocator);
+    try testing.expectEqual(x509.SignatureAlgorithm.rsa_pss, rp.cert.signatureAlgorithm());
+    try testing.expectError(error.UnsupportedSignatureAlgorithm, verify.verifySelfSignature(cp, &rp.cert));
+}
+
+test "the three supported schemes classify to the right key types" {
+    const allocator = testing.allocator;
+
```
Jaruso, line 85: Minor naming cleanup: RSA-PSS is now explicitly deferred, so this test no longer covers "three supported schemes." Consider renaming it to something like `the three matrix schemes classify to the right key types` to distinguish OID/key classification from provider support.

PR #414, opened 2026-07-13
tls: remediate merged record-layer foundation audit findings

Jaruso:
**What changed and why**
Resolves all seven confirmed findings from the retrospective audit of the merged record-mode foundation (#393, #395, #397, #398, #404, #406), surfaced while integrating the encrypted stream (#407):

1. **One canonical handshake transport contract** — `record_transport.zig` (a record-specific contract duplicating `tls.transport.Contract`'s guarantees with no production consumer) is removed. `tls.transport.Contract`/`ContractWithOptions` is now the single contract QUIC and TCP record mode both instantiate, and gained the ability to carry a terminal fatal alert (`Event.fatal_alert`/`emitFatalAlert`).
2. **Atomic, zeroizing event sinks** — `EventSink.reset()`/`deinit()` now securely wipe the used scratch range (not just counters), every `Driver` gained a `deinit()` that wipes its sink, and byte-bearing emitters preflight capacity before copying so a rejected emit never leaves secret bytes behind.
3. **Exact-consumption parser contract** — verified `record_codec.Parser.feedOne`/`drainReady` (already added alongside #407) fully close the ambiguous-partial-consumption gap; added the one missing edge case (retrying against an already-saturated sink consumes nothing).
4. **Legal `0x0301` ClientHello compatibility** — `record_codec.parseHeader` now accepts RFC 8446 §5.1's `0x0301` compatibility version, scoped to only the very first record a server-role initial-epoch parser consumes.
5. **Validated epoch-discard state transition** — `record_epoch_bridge.Bridge.discardEpoch` rejects early/late/duplicate discards instead of silently no-op'ing; `markHandshakeComplete` requires the initial epoch to already be discarded; the record-mode byte stream clears its initial-epoch parser in lockstep with the discard event.
6. **A real client/server handshake through the record stack** — new `src/quic/record_mode_handshake_test.zig` drives the production `Tls13Backend` engine (not a scripted/fabricated backend) end-to-end through two independent `record_epoch_bridge.Bridge` instances, with independently cross-checked first-record keys, sequence numbers, epoch discards, and cleanup on success and authentication-failure paths.
7. **Independent AEAD known-answer vectors** — `record_protection` now has vectors independently computed by a separate implementation (RFC 5869/RFC 8446 §7.1 HKDF via Python's stdlib `hmac`/`hashlib`, AEAD sealing via `pycryptodome`, cross-checked against Zig's own `std.crypto.tls.hkdfExpandLabel`) for all three supported cipher suites, so a bug shared between seal and open can't hide behind a self-consistent round trip.

This remediation was required before #407 is treated as satisfying the encrypted-stream acceptance criteria, per #408's stated scope.

**Related issues**
Closes #408

**Tests**
- New tests added per finding: `src/tls/transport.zig`, `src/tls/engine.zig`, `src/quic/tls_handshake.zig`, `src/quic/connection.zig` (finding 1/2); `src/tls/record_codec.zig` (findings 3/4); `src/tls/record_epoch_bridge.zig`, `src/tls/encrypted_stream.zig` (finding 5); `src/tls/record_protection.zig` (finding 7); new `src/quic/record_mode_handshake_test.zig` (finding 6, wired into `test-quic` via `src/quic/root.zig`).
- `zig build test-tls`, `zig build test-quic`, and `zig build test` all pass.
- `zig build test-integration` was run and fails identically on `main` at the merge-base (`ce2c607`) before this branch's changes — a pre-existing sandbox/networking limitation (`error.SocketUnconnected`), not a regression from this PR.

**Security impact**
Yes — this is a TLS record-layer change. Reviewed: traffic-secret zeroization on reset/failure/overflow/teardown, atomic event emission on overflow, epoch-discard ordering (no late plaintext acceptance, no stale keys after discard), and the `0x0301` compatibility acceptance is scoped strictly to the first record of a server-role initial-epoch parser (ciphertext-mode parsers reject it unconditionally as defense in depth). Finding 7's independent AEAD vectors and finding 6's real end-to-end handshake add cryptographic-correctness coverage beyond the existing self-consistent round trips.

**Performance impact**
None. No hot-path behavior changed; additions are validation/bookkeeping in already-present control flow (event emission preflight, epoch-transition checks) plus test-only code.

**Checklist**
- [x] `zig build test-integration` run (pre-existing unrelated sandbox failure, confirmed present at the merge-base before this branch)
- [x] README / CHANGELOG updated (`CHANGELOG.md` `[Unreleased]` → Features)


src/quic/connection.zig
```diff
@@ -486,6 +486,7 @@ pub const Connection = struct {
     }
 
     fn deinitPartial(self: *Connection) void {
+        self.handshake.deinit();
```
Jaruso, line 496: **Blocking:** `Connection.handshake` is declared `undefined`, while the `errdefer conn.deinitPartial()` is installed before `installInitialSecrets()` and before the assignment to `conn.handshake`. If initial-secret installation (or another fallible operation in that window) fails, this now calls `deinit()` on undefined storage. Please initialize the handshake before any fallible operation covered by this `errdefer`, or track initialization explicitly (for example, an optional/boolean guard) and add an early-init-failure regression test.

src/tls/transport.zig
```diff
@@ -51,6 +74,11 @@ pub fn ContractWithOptions(
             discard_epoch: Epoch,
             /// The handshake authenticated and completed on this side.
             handshake_complete,
+            /// A fatal alert the transport must send before closing. Whether
+            /// and when to emit one from a handshake error is transport
+            /// policy (record mode's is #354); this variant only lets the
+            /// contract carry it once a caller decides to.
+            fatal_alert: alerts.AlertDescription,
```
Jaruso, line 81: **Blocking:** this restores the ability to *represent* a fatal alert, but not the deleted record driver's terminal-result contract. `engine.Driver.receive()` still catches a backend error and returns only that error, so record mode has no supported receive result containing both the emitted sink (including a fatal alert or preceding output) and the terminal error. That terminal-error-plus-events behavior was one of finding 1's guarantees. Please expose a canonical result such as `{ sink, terminal_error }` (either in the generic driver or a record adapter around it) and test a backend that emits terminal output/alert and then fails, verifying the caller receives both.

src/tls/record_codec.zig
```diff
@@ -188,11 +233,17 @@ pub const RecordHeader = struct {
...
     if (bytes.len != header_len) return error.TruncatedRecord;
     const content_type = parseContentType(bytes[0]) catch return error.InvalidRecordType;
     const version = std.mem.readInt(u16, bytes[1..3], .big);
-    if (version != legacy_record_version) return error.InvalidRecordVersion;
+    // The compatibility version has no meaning once records are encrypted: a
+    // ciphertext-mode parser always requires exactly 0x0303, regardless of
+    // the caller-supplied policy, so a misconfigured parser instance still
+    // fails closed rather than accepting it on the wrong record stream.
+    const version_ok = version == legacy_record_version or
+        (mode == .plaintext and version_policy == .allow_initial_client_hello_compat and version == compat_client_hello_version);
```
Jaruso: **Blocking:** this permits `0x0301` for *any* first plaintext record handled by the server's initial parser. A first `alert` or `change_cipher_spec` record with version `0x0301` passes this header check, even though #408 requires compatibility only for the initial ClientHello. At minimum include `content_type == .handshake`; more robustly, consume the compatibility allowance only after the complete first record is verified to begin with a ClientHello handshake message. Please add negative tests for a `0x0301` alert/CCS and for a non-ClientHello handshake record.

src/tls/record_epoch_bridge.zig
```diff
@@ -117,31 +128,41 @@ pub const Bridge = struct {
...
-            .initial,
-            .zero_rtt,
-            => {},
+            .zero_rtt => return error.UnsupportedRecordEpoch,
         }
     }
 
     pub fn markHandshakeComplete(self: *Bridge) Error!void {
         if (self.handshake_complete) return error.InvalidEpochTransition;
+        if (!self.initial_discarded) return error.InvalidEpochTransition;
```
Jaruso: **Blocking:** finding 5 specifically requires handshake completion to prove that prior handshake keys were released. This check only requires initial discard plus installed application keys, so `markHandshakeComplete()` succeeds while both handshake key states remain live (the new unit test currently demonstrates exactly that path). Please require `handshake_discarded` before completion, or atomically clear the handshake states as part of completion, and add a test that completion fails when the required handshake discard is missing.

src/tls/record_epoch_bridge.zig
```diff
@@ -117,31 +128,41 @@ pub const Bridge = struct {
...
+                if (self.handshake_discarded) return error.EpochAlreadyDiscarded;
+                if (self.read_phase != .application or self.write_phase != .application) return error.EpochDiscardTooEarly;
                 clearRead(&self.read_handshake);
                 clearWrite(&self.write_handshake);
-                self.read_handshake_discarded = true;
-                self.write_handshake_discarded = true;
+                self.handshake_discarded = true;
             },
             .application => {
+                if (self.application_discarded) return error.EpochAlreadyDiscarded;
```
Jaruso, line 151: **Blocking:** application discard still has no early-transition validation. Calling `discardEpoch(.application)` on a fresh bridge succeeds, marks the epoch permanently discarded, and resets both phases, which violates the issue's requirement to reject early application discards. Since `deinit()` already provides unconditional failure/abandonment cleanup, the orderly transition here should require a completed handshake with live application state (for example, `handshake_complete` and both phases at `.complete`) before clearing it. Please cover early and pre-completion application discard in tests.

src/quic/record_mode_handshake_test.zig
```diff
@@ -0,0 +1,419 @@
...
+        // events of its own -- it just becomes ready to receive the
+        // ClientHello), so both sides must be started before anything is
+        // pumped.
+        _ = try self.server_driver.start(dummy_params);
+        const initial = try self.client_driver.start(dummy_params);
+        try pump(&self.client_driver, &self.client_bridge, .client, &self.server_driver, &self.server_bridge, .server, initial, &self.observed);
+    }
+
+    fn deinit(self: *Harness) void {
+        self.client_bridge.deinit();
```
Jaruso, line 341: The success/failure tests claim teardown coverage, but this harness only deinitializes the bridges. The two `CoreDriver` sinks can still contain copied traffic-secret bytes from their final event batch, so this does not prove the new driver teardown guarantee end to end. Please call `client_driver.deinit()` and `server_driver.deinit()` here (with an initialization guard if needed) and assert the used scratch ranges are zero after teardown.

Jaruso (commented):
The overall direction is good and CI is green, but several parts of #408's lifecycle/security contract are not fully satisfied yet. I found five correctness blockers plus one teardown-coverage gap: unsafe QUIC partial initialization, loss of terminal-error-plus-events semantics, overly broad `0x0301` acceptance, handshake completion with live handshake keys, unvalidated early application discard, and the end-to-end harness not deinitializing its drivers. Details and concrete remediation suggestions are inline.

src/quic/connection.zig
Jaruso: Fixed in 555d85a: `conn.handshake` is now constructed before `errdefer conn.deinitPartial()` is installed, so the fallible operations in between (`installInitialSecrets`, etc.) can no longer trigger a `deinit()` on undefined storage. `Handshake.initClient`/`initServer` are plain constructors with no dependency on installed secrets, so the reorder is free. Added a regression test that forces `installInitialSecrets` to fail (too-short `original_dcid`) and confirms `Connection.init` fails cleanly.

---

src/tls/transport.zig
Jaruso: Fixed in 555d85a: added `Driver.startOutcome()`/`receiveOutcome()`, returning `{ sink, terminal_error }` instead of discarding the sink on backend failure. Kept as additive methods alongside the existing `start()`/`receive()` so QUIC's simple `try`-based happy path is unaffected. Added a test with a backend that emits a fatal alert and then fails, confirming the caller gets both the alert and the terminal error from one call.

---

src/tls/record_codec.zig
Jaruso: Fixed in 555d85a: `parseHeader`'s compat condition now also requires `content_type == .handshake`, so an alert/change_cipher_spec record with `0x0301` is rejected before the payload is even looked at. Since `parseHeader` only sees the 5-byte header, added a second check in `Parser.drain`/`drainOne` once the full record is assembled: a `0x0301`-tagged record is rejected unless its payload actually begins with the ClientHello handshake message type (`msg_type == 1`). Added negative tests for both a `0x0301` alert/CCS and a `0x0301` handshake record that isn't a ClientHello.

---

src/tls/record_epoch_bridge.zig
Jaruso: Fixed in 555d85a: `markHandshakeComplete()` now also requires `handshake_discarded`, not just `initial_discarded`. This required reordering the scripted-backend `Harness.pump` tests (in both `record_epoch_bridge.zig` and `record_mode_handshake_test.zig`) to a two-pass design: seal each handshake-epoch message in the sink's own order (so a message is sealed before its epoch's key is discarded), but defer cross-side delivery/recursion to a second pass so a bridge's own remaining local events -- notably its own `handshake_complete` -- are applied before a peer's cascading response can depend on them. Extended the existing test to assert completion still fails when only `initial` (not `handshake`) has been discarded.

---

src/tls/record_epoch_bridge.zig
Jaruso: Fixed in 555d85a: `discardEpoch(.application)` now requires `handshake_complete` and both phases at `.complete` before clearing, so it fails closed on a fresh or in-progress bridge instead of succeeding unconditionally. `deinit()` remains the unconditional abandonment/failure cleanup path, unaffected. Added a test covering early discard both before any handshake progress and after application keys are installed but before completion, plus the already-discarded case.

---

src/quic/record_mode_handshake_test.zig
Jaruso: Fixed in 555d85a: `Harness.deinit()` now also calls `client_driver.deinit()`/`server_driver.deinit()` (guarded by a `drivers_ready` flag against the pre-`run()` undefined state), and both tests capture each driver's `sink.used` before teardown and assert the scratch range is actually zeroed after.

---

src/tls/record_codec.zig
```diff
@@ -150,29 +188,43 @@ pub const Parser = struct {
...
         self.discard(record_len);
+        self.parsed_first_record = true;
+    }
+
+    /// The very first record this parser consumes may use its configured
+    /// compatibility policy; every record after that is always `.strict`,
+    /// matching the RFC 8446 SS5.1 scoping to an initial (non-post-HRR)
+    /// ClientHello.
+    fn currentVersionPolicy(self: *const Parser) VersionPolicy {
+        return if (self.parsed_first_record) .strict else self.first_record_version_policy;
```
Jaruso, line 333: **Blocking:** this still scopes compatibility to the first *record*, not to the complete initial ClientHello. RFC 8446 §5.1 permits handshake messages to be fragmented across several records and says a record containing the initial ClientHello may use `0x0301`. A fragmented ClientHello whose continuation record also uses `0x0301` is rejected here because `parsed_first_record` becomes true after fragment 1; that continuation also cannot pass `checkCompatVersionCarriesClientHello` because it does not begin with a new `msg_type` byte. Please track the initial ClientHello message across record boundaries (including a handshake header split across records), keep the compatibility window open until its declared handshake length is fully consumed, and add a multi-record `0x0301` ClientHello test. The window should then close before any post-HRR ClientHello or other plaintext message.

src/quic/record_mode_handshake_test.zig
```diff
@@ -0,0 +1,458 @@
...
+    // *and* the drivers' own event sinks. The client driver's last receive
+    // (the server's flight) copied fresh application traffic secrets into
+    // its sink, so its scratch is provably nonempty before teardown; the
+    // server driver's last receive (the client's Finished) only carried
+    // non-byte-bearing discard/complete events, so its scratch may already
+    // be empty -- the post-teardown check still holds either way.
+    const client_driver_used = h.client_driver.sink.used;
+    const server_driver_used = h.server_driver.sink.used;
+    try std.testing.expect(client_driver_used > 0);
+    h.deinit();
```
Jaruso, line 426: Non-blocking: both tests still have `defer h.deinit()` and then call `h.deinit()` explicitly here to inspect the wiped state, so teardown runs twice despite the contract/documentation saying each owner calls it exactly once. It happens to be idempotent today, but this test would mask a future non-idempotent teardown bug. Please guard the deferred cleanup (for example with a `deinitialized` boolean) or move the post-deinit assertions into a helper that performs the one teardown call.

docs/TLS_RECORD_TRANSPORT.md
```diff
@@ -1,20 +1,44 @@
...
+
+The handshake backend emits events tagged with the caller's own `Epoch` type:
+handshake bytes to send, traffic-secret installation (read/write, per epoch),
+transport-owned peer parameters, negotiated ALPN, peer certificate state,
+epoch discard, completion, and a fatal alert.
+
+Event byte slices are copied into the driver-owned `EventSink`. They remain
+valid only until the next `Driver.start` or `Driver.receive` call, both of
+which reset the sink. Resetting or deinitializing the sink securely zeroes the
+used scratch range, so a copied traffic secret does not survive past that
```
Jaruso: Suggestion: the canonical terminal-error-plus-events behavior added in the fix is not documented here, and the lifetime sentence only names `start`/`receive` even though `startOutcome`/`receiveOutcome` also reset the sink. Please document when record-mode callers must use the outcome APIs, that `terminal_error` and the emitted sink are returned together, and that all four drive methods invalidate prior event slices. This was one of finding 1's key contract guarantees.

Jaruso (commented):
I re-reviewed the updated head. The six original findings are substantively fixed and the latest CI run is green. I found one remaining interoperability blocker in the `0x0301` compatibility state machine, plus two non-blocking cleanup/documentation issues. Details are inline.

src/tls/record_codec.zig
Jaruso: Fixed in 38ec958: `Parser` now tracks `client_hello_remaining` and keeps the compatibility window open across as many records as the ClientHello's declared 3-byte length says it needs, instead of closing after the first record unconditionally. A continuation record is accepted without being re-checked against `msg_type` (it's a raw fragment, not a new message). The window still closes for good, on any record, once that accounting says the message is fully consumed. Added a multi-record fragmented-ClientHello test, and fixed two existing tests whose payloads didn't carry a real length field (harmless before, but would've been misread as declaring a multi-megabyte ClientHello under the new tracking).

---

src/quic/record_mode_handshake_test.zig
Jaruso: Fixed in 38ec958: added a `deinitialized` guard to `Harness`, so `deinit()` is now a documented no-op on a second call rather than relying on incidental idempotency in `Bridge`/`Driver`'s own teardown.

---

docs/TLS_RECORD_TRANSPORT.md
Jaruso: Fixed in 38ec958: added a "Terminal error plus events" section documenting `startOutcome`/`receiveOutcome`, when a record-mode caller needs them over `start`/`receive`, and that `{ sink, terminal_error }` are always returned together. Updated the lifetime sentence to name all four drive methods.

---

Jaruso (commented):
I re-reviewed commit `38ec958`. The teardown guard and transport-contract documentation changes are good, and CI is green. The compatibility implementation now handles ordinary multi-record ClientHello fragmentation, but two parser correctness blockers remain: it still rejects a handshake header split across records, and its new fragment accounting is mutated before sink emission succeeds, so backpressure retries can count the same record twice. Details and suggested tests are inline.

src/tls/record_codec.zig
```diff
@@ -150,29 +197,74 @@ pub const Parser = struct {
...
+    ///   begin a ClientHello (`msg_type` then a 3-byte length). If the
+    ///   record does not carry the whole declared message, the window
+    ///   stays open until a later record supplies the rest.
+    fn advanceClientHelloCompatWindow(self: *Parser, header: RecordHeader, payload: []const u8) Error!bool {
+        if (self.client_hello_remaining > 0) {
+            self.client_hello_remaining -= @min(payload.len, self.client_hello_remaining);
+            return self.client_hello_remaining > 0;
+        }
+        if (header.legacy_version != compat_client_hello_version) return false;
+        if (payload.len < handshake_header_len or payload[0] != client_hello_msg_type) return error.InvalidRecordVersion;
```
Jaruso: **Blocking:** this still rejects the handshake-header-split case called out in the prior review. TLS handshake messages may fragment at any byte boundary, so the first `0x0301` record can legally contain only `msg_type`, or `msg_type` plus one/two length bytes. Here any payload shorter than four bytes immediately returns `InvalidRecordVersion`. Please accumulate the four-byte handshake header across records before decoding the declared length, while validating the first available byte is ClientHello, and add cases splitting after 1, 2, and 3 header bytes. The compatibility window should remain open through those fragments and close only after the declared body is fully consumed.

src/tls/record_codec.zig
```diff
@@ -150,29 +197,74 @@ pub const Parser = struct {
 
     fn drain(self: *Parser, sink: anytype) Error!void {
         while (self.len >= header_len) {
-            const header = try parseHeader(self.pending[0..header_len], self.mode);
+            const header = try parseHeader(self.pending[0..header_len], self.mode, self.currentVersionPolicy());
             const record_len = header_len + header.payload_len;
             if (self.len < record_len) return;
+            const payload = self.pending[header_len..record_len];
+            const still_fragmenting = try self.advanceClientHelloCompatWindow(header, payload);
```
Jaruso: **Blocking:** `advanceClientHelloCompatWindow()` mutates `client_hello_remaining` before `sink.push()` succeeds. If `push` returns `RecordSinkOverflow`, this record stays buffered, but the compatibility state has already advanced; `drainReady()` then processes the same record again and decrements/counts it a second time (and a failed first fragment can be reinterpreted as a continuation without revalidating `msg_type`). Please make the state transition transactional—compute the next compatibility state without mutating, commit it only after `sink.push()` succeeds—and add a compatibility-fragment regression using a sink that overflows, resets, then retries the buffered record.

src/tls/record_codec.zig
Jaruso: Fixed in a5086dc: `Parser.client_hello_state` is now a tagged union (`idle` / `collecting_header` / `collecting_body`) that accumulates the 4-byte `msg_type`+length header across as many records as needed, validating `msg_type` against the first available byte, before decoding the declared length. Added tests splitting after 1, 2, and 3 header bytes.

---

src/tls/record_codec.zig
Jaruso: Fixed in a5086dc: split into a pure `nextClientHelloState(current, header, payload)` that computes the next state without mutating the parser, and `drain`/`drainOne` now only commit it to `self.client_hello_state` (and `parsed_first_record`) after `sink.push()` succeeds. A failed push (`RecordSinkOverflow`) leaves the state untouched, so `drainReady()`'s retry recomputes from the same pre-push state instead of double-counting. Added a regression using a 1-capacity sink that overflows mid-fragment, then drains and retries.

---

src/tls/engine.zig
```diff
@@ -61,6 +71,21 @@ pub fn Driver(comptime Transport: type) type {
             return &self.sink;
         }
 
+        /// Like `start`, but on backend failure returns the sink alongside the
+        /// error instead of discarding it, so a caller that needs the backend's
+        /// terminal output (e.g. a fatal alert emitted just before failing) can
+        /// still reach it.
+        pub fn startOutcome(self: *Self, params: Transport.TransportParametersType) Outcome {
```
Jaruso, line 78: **Blocking:** `startOutcome()` does not preserve the terminal outcome once the driver is already failed. A second call enters this `state != .idle` branch, overwrites the original backend failure with `InvalidHandshakeState`, and returns the old sink under the new error. That contradicts the documented guarantee that repeated outcome calls after failure return the same terminal error and sink without re-invoking the backend. It also means an invalid second start can return stale traffic-secret events from the prior call without resetting them. Please special-case `.failed` exactly as `receiveOutcome()` does, and define/reset the sink behavior for other non-idle states so stale events are never represented as output from the invalid call. Add regressions for repeated `startOutcome` after backend failure and for a second start after a successful event batch containing a secret.

src/tls/record_codec.zig
```diff
@@ -53,6 +53,100 @@ pub const InnerPlaintext = struct {
...
+/// decides whether the window closes now (the message fits, with or
+/// without trailing bytes belonging to something else) or stays open for a
+/// further `.collecting_body` record.
+fn nextClientHelloState(current: ClientHelloState, header: RecordHeader, payload: []const u8) Error!ClientHelloState {
+    var header_buf: [handshake_header_len]u8 = undefined;
+    var header_have: usize = 0;
+
+    switch (current) {
+        .idle => {
+            if (header.legacy_version != compat_client_hello_version) return .idle;
```
Jaruso: **Blocking:** compatibility tracking only begins when the *first* fragment itself uses `0x0301`. A legal initial ClientHello can be fragmented with its first record at `0x0303` and a continuation record at `0x0301`; both records contain the initial ClientHello, but this branch returns `.idle`, marks the first record complete for policy purposes, and the continuation is then rejected under `.strict`. Track the initial ClientHello message regardless of whether its first record used `0x0303` or `0x0301`, while only permitting `0x0301` while that tracked initial message remains in progress. Add mixed-version header/body fragmentation tests.

src/tls/record_codec.zig
```diff
@@ -53,6 +53,100 @@ pub const InnerPlaintext = struct {
...
+            if (payload.len == 0) return error.InvalidRecordVersion;
+        },
+        .collecting_header => |progress| {
+            header_buf = progress.buf;
+            header_have = progress.len;
+        },
+        .collecting_body => |remaining| {
```
Jaruso, line 148: **Blocking:** while collecting a fragmented ClientHello, this state machine consumes payload bytes from *any* `0x0303` plaintext record without checking `header.content_type == .handshake`. An intervening alert or CCS record is therefore treated as ClientHello header/body continuation and corrupts `client_hello_state`; RFC 8446 §5.1 forbids interleaving other record types within a fragmented handshake message. Require handshake content type for every `collecting_header`/`collecting_body` continuation and reject without committing state. Add negative cases for alert/CCS during both header and body collection.

src/tls/encrypted_stream.zig
```diff
@@ -178,6 +190,16 @@ pub const PureZigRecordStream = struct {
...
             self.outbound_ciphertext.append(record) catch |err| return self.fail(err);
         }
+        // The initial epoch's plaintext parser is only safe to keep once
+        // its keys are gone: nothing should ever arrive at that epoch
+        // again after discard, so drop any partially-buffered state along
+        // with it. Never do this for `.handshake` discard: the ciphertext
+        // parser is shared across the handshake and application epochs,
+        // and bytes already buffered there may belong to the next
+        // (application) record.
+        if (event == .discard_epoch and event.discard_epoch == .initial) {
```
Jaruso, line 256: **Blocking:** only the initial parser is cleared at discard; a partially buffered ciphertext record is silently carried across `.handshake` discard into the application-key epoch. That is precisely the stale partial-record state the issue requires the transition to clear or reject. With `feedOne`, a legitimate following application-record suffix remains in the caller/carrier buffer rather than inside `ciphertext_parser`, so the parser should be empty at this key transition. Please fail closed if it is not empty (or use per-epoch parser state), then add a regression that buffers a partial handshake-epoch record and attempts the discard. Do not reinterpret unknown pre-transition bytes under application keys.

Jaruso (commented):
I completed a full review of the entire current PR, not only the latest two fixes. The earlier findings are substantially improved and CI is green, but I found four remaining correctness/contract blockers, so I do not recommend merging this head yet. They concern outcome-driver failed-state semantics, the ClientHello compatibility state machine in two additional legal/adversarial fragmentation cases, and stale ciphertext-parser state crossing the handshake key transition. Details and concrete regression cases are inline. (GitHub does not allow the PR author to submit a formal REQUEST_CHANGES review on their own PR, so this is a blocking COMMENT review.)

Non-blocking cleanup after those fixes: wire `record_mode_handshake_test.zig` through a test-only import instead of exporting it from `quic/root.zig`; the current wiring exposes a test module in the public QUIC namespace and adds `crypto` to the production module solely for the test.

src/tls/engine.zig
Jaruso: Fixed in 27ae13d: `startOutcome()` now checks `self.state == .failed` first, exactly like `receiveOutcome()`, returning the original terminal error and sink unchanged on repeat calls. For the other non-idle states it resets the sink before marking failed, so an invalid second call can't surface a stale event batch (including a copied traffic secret) as its own output. Added both requested regressions.

---

src/tls/record_codec.zig
Jaruso: Fixed in 27ae13d: tracking now starts at the first handshake-content-type record regardless of its declared version (0x0303 or 0x0301), instead of only when that first record itself claims 0x0301. A non-ClientHello first message is only an error if some record in the sequence actually invoked the exception for it, so ordinary 0x0303 traffic that happens not to be a ClientHello is unaffected. Added a mixed-version (0x0303 first fragment, 0x0301 continuation) test.

---

src/tls/record_codec.zig
Jaruso: Fixed in 27ae13d: continuation records in `.collecting_header`/`.collecting_body` now require `content_type == .handshake` and are rejected (`InvalidRecordType`, without committing state) otherwise. This is gated to compat-policy parsers only, since the same check would otherwise reject completely ordinary traffic on a `.strict` parser (an unrelated handshake record followed by an alert, which has nothing to do with any compatibility window). Added negative cases for alert/CCS during both header and body collection.

---

src/tls/encrypted_stream.zig
Jaruso: Fixed in 27ae13d: `applyEvent` now fails the stream closed (new `error.PartialRecordAtEpochTransition`) when `.handshake` discard lands with `ciphertext_parser.len != 0`. As you noted, `feedOne`'s exact-consumption contract means that can only represent a genuinely incomplete record straddling the key transition -- a legitimate next-record suffix stays in the caller/carrier buffer instead -- so there's no case where continuing to buffer it is safe. Added a regression that buffers a partial handshake-epoch record and then attempts the discard.

---

src/tls/record_codec.zig
```diff
@@ -150,31 +286,65 @@ pub const Parser = struct {
...
     fn drainOne(self: *Parser, sink: anytype) Error!void {
         if (self.len < header_len) return;
-        const header = try parseHeader(self.pending[0..header_len], self.mode);
+        const header = try parseHeader(self.pending[0..header_len], self.mode, self.currentVersionPolicy());
         const record_len = header_len + header.payload_len;
         if (self.len < record_len) return;
+        const payload = self.pending[header_len..record_len];
+        const next_client_hello_state = try nextClientHelloState(self.client_hello_state, header, payload);
         try sink.push(.{
             .content_type = header.content_type,
```
Jaruso, line 318: **Blocking:** `drain()` now correctly gates ClientHello tracking through `self.nextClientHelloStateForRecord(...)`, but `drainOne()` bypasses that guard and calls `nextClientHelloState(...)` directly. `feedOne()` is the stream-facing API, so a strict parser can still enter `.collecting_header`/`.collecting_body` after an ordinary fragmented handshake record and then reject a following alert or CCS as `InvalidRecordType`. Use the guarded helper here too, and add a strict-policy `feedOne()` regression so `feed()` and `feedOne()` have identical compatibility behavior.

Jaruso (commented):
Full re-review of `27ae13d`:

Three of the four prior blockers are correctly addressed:
- `startOutcome()` preserves an existing terminal failure and clears stale output on an invalid second start.
- the ClientHello tracker now supports a `0x0303` first fragment followed by `0x0301`, and rejects non-handshake interleaving.
- the stream fails closed if a partial ciphertext record exists at handshake-epoch discard.
- the record-mode handshake test is now test-only rather than exported from the production QUIC module.

One blocking regression remains in `record_codec.Parser`: `drain()` correctly calls `nextClientHelloStateForRecord()`, which gates tracking to `.allow_initial_client_hello_compat`, but `drainOne()` still calls `nextClientHelloState()` directly. Since `feedOne()` is the API used by the encrypted stream, strict parsers can enter ClientHello tracking after an ordinary fragmented first handshake record and then reject a following alert/CCS as `InvalidRecordType`. This contradicts the new helper's own stated purpose and makes behavior depend on whether the caller uses `feed()` or `feedOne()`.

Please make `drainOne()` use `self.nextClientHelloStateForRecord(header, payload)` as well, and add a strict-parser `feedOne()` regression proving a fragmented ordinary handshake record followed by an alert/CCS is not subjected to compat-window interleaving rules.

Because this is the author's own PR, GitHub only permits a COMMENT review rather than formal REQUEST_CHANGES. My merge recommendation remains **do not merge** until this is fixed.

Jaruso (commented):
Final re-review complete at `acc6955e`.

The remaining `record_codec` blocker is fixed correctly: `drainOne()` now uses the same policy-gated `nextClientHelloStateForRecord()` helper as `drain()`, and the added strict-parser regression proves `feed()` and `feedOne()` accept the same fragmented-handshake-then-alert sequence without entering ClientHello compatibility tracking.

I also reviewed the small OpenSSL pending-close test synchronization change. Waiting for the peer reader to complete and then driving the server again to consume the peer shutdown is consistent with the bidirectional TLS shutdown flow and does not alter production behavior.

All previously raised blocking findings remain addressed, all review threads are resolved, current CI is green, and GitHub reports the PR mergeable. I found no remaining code-level reason to hold this PR. Because this is the author's own PR, this is a COMMENT review rather than APPROVE.

PR #415, opened 2026-07-13
tls: add OpenSSL EncryptedStream adapter conformance

Jaruso:
## Summary
- expose the shared EncryptedStream vtable from production OpenSSL TlsConnection
- preserve the appliance no-OpenSSL profile with a matching inert stub surface
- add in-process OpenSSL socketpair coverage for would-block, pending readiness, partial reads, writes, close, EOF, fatal wire failure, and cleanup

Closes #411

## Validation
- zig build test-tls -Dtls-profile=appliance --summary all --error-style verbose
- zig build test -Dtls-profile=appliance --summary all --error-style verbose

### DIFF
--- a/src/http/tls_termination.zig
+++ b/src/http/tls_termination.zig
@@ -334,11 +334,12 @@ pub const TlsConnection = struct {
     stream_lifecycle: OpenSslStreamLifecycle = .open,
     retry_direction: OpenSslRetryDirection = .none,
     pending_write: ?OpenSslPendingWrite = null,
+    close_requested: bool = false,
     peer_closed: bool = false,
     stream_failure: ?encrypted_stream.Error = null,
 
     pub fn deinit(self: *TlsConnection) void {
-        if (self.stream_failure == null and self.stream_lifecycle != .closed) {
+        if (opensslShouldShutdownOnDeinit(self)) {
             c.ERR_clear_error();
             _ = c.SSL_shutdown(self.ssl);
         }
@@ -448,9 +449,14 @@ fn opensslClearRetry(self: *TlsConnection) void {
     self.pending_write = null;
 }
 
+fn opensslShouldShutdownOnDeinit(self: *const TlsConnection) bool {
+    return self.stream_failure == null and self.stream_lifecycle != .closed and self.pending_write == null;
+}
+
 fn opensslStreamFail(self: *TlsConnection, err: encrypted_stream.Error) encrypted_stream.Error {
     self.stream_lifecycle = .failed;
     opensslClearRetry(self);
+    self.close_requested = false;
     self.stream_failure = err;
     return err;
 }
@@ -481,6 +487,7 @@ fn opensslStreamRead(ptr: *anyopaque, out: []u8) encrypted_stream.Error!usize {
     if (self.stream_failure) |err| return err;
     if (self.stream_lifecycle != .open) return error.StreamClosed;
     if (self.pending_write != null) return error.RetryOperationPending;
+    if (self.close_requested) return error.StreamClosed;
     c.ERR_clear_error();
     const rc = c.SSL_read(self.ssl, out.ptr, @intCast(out.len));
     if (rc > 0) {
@@ -510,6 +517,7 @@ fn opensslStreamWrite(ptr: *anyopaque, bytes: []const u8) encrypted_stream.Error
     if (self.pending_write) |pending| {
         if (!pending.matches(bytes)) return error.RetryOperationPending;
     } else {
+        if (self.close_requested) return error.StreamClosed;
         if (bytes.len == 0) return 0;
     }
     c.ERR_clear_error();
@@ -537,6 +545,7 @@ fn opensslStreamWrite(ptr: *anyopaque, bytes: []const u8) encrypted_stream.Error
 fn opensslAdvanceShutdown(self: *TlsConnection) encrypted_stream.Error!bool {
     if (self.stream_failure) |err| return err;
     if (self.stream_lifecycle == .closed) return false;
+    self.close_requested = true;
     if (self.pending_write != null) return error.RetryOperationPending;
 
     self.stream_lifecycle = .closing;
@@ -545,6 +554,7 @@ fn opensslAdvanceShutdown(self: *TlsConnection) encrypted_stream.Error!bool {
     opensslRefreshPeerClosed(self);
     if (rc == 1) {
         self.stream_lifecycle = .closed;
+        self.close_requested = false;
         opensslClearRetry(self);
         return true;
     }
@@ -571,6 +581,7 @@ fn opensslAdvanceShutdown(self: *TlsConnection) encrypted_stream.Error!bool {
 fn opensslStreamClose(ptr: *anyopaque) void {
     const self: *TlsConnection = @ptrCast(@alignCast(ptr));
     if (self.stream_lifecycle == .closed or self.stream_lifecycle == .failed) return;
+    self.close_requested = true;
     _ = opensslAdvanceShutdown(self) catch {};
 }
 
@@ -595,6 +606,12 @@ fn opensslStreamReadiness(ptr: *anyopaque) encrypted_stream.Readiness {
             .peer_closed = self.peer_closed,
         };
     }
+    if (self.close_requested) {
+        return .{
+            .wants_write = true,
+            .peer_closed = self.peer_closed,
+        };
+    }
     return .{
         .wants_read = self.retry_direction == .read or (self.retry_direction == .none and pending == 0 and !self.peer_closed),
         .wants_write = self.retry_direction == .write,
@@ -607,7 +624,7 @@ fn opensslStreamReadiness(ptr: *anyopaque) encrypted_stream.Readiness {
 fn opensslStreamDrive(ptr: *anyopaque) encrypted_stream.Error!encrypted_stream.DriveResult {
     const self: *TlsConnection = @ptrCast(@alignCast(ptr));
     if (self.stream_failure) |err| return err;
-    const made_progress = if (self.stream_lifecycle == .closing)
+    const made_progress = if (self.stream_lifecycle == .closing or (self.close_requested and self.pending_write == null))
         try opensslAdvanceShutdown(self)
     else
         false;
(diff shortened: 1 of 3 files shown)


src/http/tls_termination.zig
```diff
@@ -402,6 +408,69 @@ pub const TlsConnection = struct {
...
+    return .openssl;
+}
+
+fn opensslStreamRead(ptr: *anyopaque, out: []u8) encrypted_stream.Error!usize {
+    const self: *TlsConnection = @ptrCast(@alignCast(ptr));
+    if (self.closed) return error.StreamClosed;
+    const rc = c.SSL_read(self.ssl, out.ptr, @intCast(out.len));
+    if (rc > 0) return @intCast(rc);
+    return switch (c.SSL_get_error(self.ssl, rc)) {
+        c.SSL_ERROR_WANT_READ, c.SSL_ERROR_WANT_WRITE => error.WouldBlock,
```
Jaruso: Both retry cases become `WouldBlock`, but `readiness()` later hard-codes `wants_read = true` and `wants_write = false`. OpenSSL allows `SSL_read` to return `WANT_WRITE` and `SSL_write` to return `WANT_READ`; polling the wrong edge can stall the connection indefinitely. Please retain the last required I/O direction in `TlsConnection`, clear it after progress, and reflect it in `readiness()`/`drive()`. This should have a test that forces socket backpressure rather than only the idle `WANT_READ` case.

src/http/tls_termination.zig
```diff
@@ -402,6 +408,69 @@ pub const TlsConnection = struct {
...
+        c.SSL_ERROR_ZERO_RETURN => error.EndOfStream,
+        else => error.SocketReadFailed,
```
Jaruso: This fatal error is returned but not latched. After `SSL_ERROR_SSL` or `SSL_ERROR_SYSCALL`, OpenSSL says no further I/O should occur and `SSL_shutdown()` must not be called; here the connection still reports writable, subsequent calls keep using the `SSL`, and `deinit()` calls shutdown. Please add a failed state that preserves the mapped error for later `read`/`write`/`drive` calls and skips shutdown during cleanup. Also clear the thread error queue immediately before each SSL I/O call so `SSL_get_error()` is reliable. The current “latches fatal wire failures” test checks only the first read, so it would not catch this.

src/http/tls_termination.zig
```diff
@@ -402,6 +408,69 @@ pub const TlsConnection = struct {
...
+        c.SSL_ERROR_WANT_READ, c.SSL_ERROR_WANT_WRITE => error.WouldBlock,
+        c.SSL_ERROR_ZERO_RETURN => error.EndOfStream,
+        else => error.SocketWriteFailed,
+    };
+}
+
+fn opensslStreamClose(ptr: *anyopaque) void {
+    const self: *TlsConnection = @ptrCast(@alignCast(ptr));
+    if (self.closed) return;
+    _ = c.SSL_shutdown(self.ssl);
```
Jaruso: A single `SSL_shutdown()` call is not necessarily a completed shutdown on a nonblocking socket: `0` means shutdown was initiated but is incomplete, and a negative result can require read or write readiness. Marking the stream closed immediately can drop the `close_notify`; returning `peer_closed = true` also conflates a local close request with receipt of the peer's authenticated close. Please model a closing state and have `drive()` continue shutdown according to the returned WANT direction (or explicitly document/test a fast-shutdown policy). Only set `peer_closed` after `SSL_ERROR_ZERO_RETURN`/`SSL_RECEIVED_SHUTDOWN`, and verify the client actually observes the close notification.

src/http/tls_termination.zig
```diff
@@ -869,3 +1045,62 @@ test "tls terminator copies sni specs for maintenance reload" {
...
+    try clientWriteAll(pair.client_ssl, "client-payload");
+    const partial = try stream.read(scratch[0..6]);
+    try std.testing.expectEqualStrings("client", scratch[0..partial]);
+    try std.testing.expect(stream.readiness().can_read_plaintext);
+
+    const rest = try stream.read(scratch[0..16]);
+    try std.testing.expectEqualStrings("-payload", scratch[0..rest]);
+    try std.testing.expect(!stream.readiness().can_read_plaintext);
+
+    const written = try stream.write("server-reply");
```
Jaruso, line 1333: This verifies only an uncontended full write, not partial/write-blocked behavior claimed by the PR and required by #411. Please force backpressure (for example, shrink `SO_SNDBUF`, stop reading on the client, and write enough data to reach `WANT_WRITE`), then drain the peer and verify readiness transitions, retry behavior, and no duplicated bytes. If the abstraction expects successful partial writes, the SSL needs `SSL_MODE_ENABLE_PARTIAL_WRITE`; otherwise the same-buffer retry requirement should be made explicit and tested.

Jaruso (commented):
Good direction, and the CI run for this head commit is green. I found three adapter-level correctness issues that can cause an event-loop stall or leave shutdown/failure state inconsistent.

One broader coverage concern: #411 calls for a shared conformance suite against both the pure-Zig and OpenSSL backends, but these are bespoke OpenSSL tests. That lets the two implementations drift on lifecycle/readiness semantics—the close behavior below is already an example. Please factor the common assertions into a backend-neutral helper and add observable graceful-shutdown, fatal-state, cleanup, and forced write-backpressure coverage.

src/http/tls_termination.zig
```diff
@@ -402,6 +416,176 @@ pub const TlsConnection = struct {
...
+    c.ERR_clear_error();
+    const rc = c.SSL_write(self.ssl, bytes.ptr, @intCast(bytes.len));
+    if (rc > 0) {
+        self.retry_direction = .none;
+        opensslRefreshPeerClosed(self);
+        return @intCast(rc);
+    }
+    const ssl_error = c.SSL_get_error(self.ssl, rc);
+    return switch (ssl_error) {
+        c.SSL_ERROR_WANT_READ, c.SSL_ERROR_WANT_WRITE => blocked: {
```
Jaruso, line 532: `SSL_MODE_ACCEPT_MOVING_WRITE_BUFFER` only permits the pointer to move; after `SSL_write` returns WANT_READ/WANT_WRITE, OpenSSL still requires the retry to use the same data and length. This adapter returns a generic `WouldBlock` while storing only `retry_direction`, so a caller can issue a different write—or call `read()`, which can overwrite the retry direction—before the required write retry. That is allowed by the current `EncryptedStream` contract but violates OpenSSL's retry contract. Please either retain/stage the pending write internally, or track the pending operation plus length and explicitly document/enforce the same-data retry rule. While an operation is pending, readiness should also avoid advertising another plaintext operation that can replace its retry state. Add a test that attempts an interleaved read or changed write slice after the forced `WANT_WRITE`.

Jaruso (commented):
The pushed commit addresses the original readiness-direction, fatal-state, graceful-shutdown, shared-conformance, and forced-backpressure concerns. One OpenSSL retry-contract issue remains before this is fully backend-neutral: a blocked `SSL_write` must be retried with the same data and length, but the adapter currently remembers only the socket direction, not the pending operation or arguments.

src/http/tls_termination.zig
```diff
@@ -402,6 +417,212 @@ pub const TlsConnection = struct {
...
+        else => opensslStreamFail(self, error.SocketWriteFailed),
+    };
+}
+
+fn opensslStreamClose(ptr: *anyopaque) void {
+    const self: *TlsConnection = @ptrCast(@alignCast(ptr));
+    if (self.stream_lifecycle == .closed or self.stream_lifecycle == .failed) return;
+    _ = opensslAdvanceShutdown(self) catch {};
+}
+
```
Jaruso, line 587: If `pending_write` is set, `opensslAdvanceShutdown()` returns `RetryOperationPending`, but `close()` swallows that error without recording that closure was requested. After the caller completes the exact write retry, `pending_write` is cleared while the lifecycle is still `.open`; `drive()` then does nothing, so the single `close()` call has been lost and the stream becomes writable again. Please latch a `close_requested`/closing intent before returning, then begin shutdown automatically after the pending retry succeeds (or define an explicit abort policy). Add a test that forces `WANT_WRITE`, calls `close()` once, completes the retry, and reaches `close_notify`/closed through `drive()` without calling `close()` a second time.

Jaruso (commented):
The pending-write retry enforcement addresses the previous review, and the latest CI run is green. One lifecycle edge remains: a close requested while an OpenSSL write retry is pending is silently lost, while `deinit()` still attempts `SSL_shutdown()` in that same state.

src/http/tls_termination.zig
```diff
@@ -328,9 +331,17 @@ pub const TlsTerminator = struct {
...
+    retry_direction: OpenSslRetryDirection = .none,
+    pending_write: ?OpenSslPendingWrite = null,
+    peer_closed: bool = false,
+    stream_failure: ?encrypted_stream.Error = null,
 
     pub fn deinit(self: *TlsConnection) void {
-        _ = c.SSL_shutdown(self.ssl);
+        if (self.stream_failure == null and self.stream_lifecycle != .closed) {
+            c.ERR_clear_error();
+            _ = c.SSL_shutdown(self.ssl);
```
Jaruso, line 344: This cleanup path still calls `SSL_shutdown()` when `pending_write != null`. At that point OpenSSL requires the blocked `SSL_write` operation to be retried with the same arguments; starting shutdown instead abandons that operation and can lose the application bytes. Please either skip graceful shutdown and directly free/abort when a write retry is outstanding, or arrange for the pending write to complete before shutdown. A cleanup test after forced write backpressure would make this behavior explicit.

Jaruso (commented):
Re-reviewed the latest commit. The pending-write retry contract, close intent, shutdown progression, and cleanup behavior now compose correctly: `close()` is latched while the exact OpenSSL write retry remains outstanding; normal plaintext I/O stays disabled; after the retry completes, `drive()` begins and finishes the bidirectional shutdown; and `deinit()` avoids calling `SSL_shutdown()` while OpenSSL still requires the pending write retry. The new tests exercise both the close-after-backpressure path and the cleanup guard. CI is green. I did not find any additional blocking issues in this pass.

PR #416, opened 2026-07-14
tls: implement alert truncation semantics

Jaruso:
## Summary
- add explicit pure-Zig TLS alert handling for close_notify, malformed alerts, and fatal alerts
- surface transport EOF before peer close_notify as TruncatedStream while keeping EOF after close_notify clean
- map OpenSSL unexpected EOF without close_notify to TruncatedStream and cover it with an interop-style adapter test

## Tests
- zig build test-tls
- zig build test-quic

Closes #354

### DIFF
--- a/src/tls/encrypted_stream.zig
+++ b/src/tls/encrypted_stream.zig
@@ -201,6 +201,7 @@ pub const PureZigRecordStream = struct {
     peer_closed: bool = false,
     carrier_eof: bool = false,
     close_notify_queued: bool = false,
+    pending_terminal_read_error: ?Error = null,
     failed: ?Error = null,
 
     pub fn init(role: tls_state.Role, crypto_provider: provider.CryptoProvider, cipher_suite: algorithms.CipherSuite) PureZigRecordStream {
@@ -238,6 +239,7 @@ pub const PureZigRecordStream = struct {
         self.peer_closed = true;
         self.carrier_eof = false;
         self.close_notify_queued = false;
+        self.pending_terminal_read_error = null;
         self.failed = null;
     }
 
@@ -293,7 +295,9 @@ pub const PureZigRecordStream = struct {
     pub fn readHandshake(self: *PureZigRecordStream, out: []u8) Error!usize {
         if (self.failed) |err| return err;
         if (self.lifecycle == .closed or self.lifecycle == .failed) return error.StreamClosed;
-        return self.inbound_handshake.read(out) orelse error.WouldBlock;
+        if (self.inbound_handshake.read(out)) |n| return n;
+        try self.raisePendingTerminalError();
+        return error.WouldBlock;
     }
 
     pub fn feedCiphertext(self: *PureZigRecordStream, bytes: []const u8) Error!usize {
@@ -319,13 +323,15 @@ pub const PureZigRecordStream = struct {
 
     pub fn markPeerEof(self: *PureZigRecordStream) Error!void {
         if (self.peer_closed) return;
-        return self.fail(error.TruncatedStream);
+        self.carrier_eof = true;
+        return self.deferTerminalReadError(error.TruncatedStream);
     }
 
     pub fn readPlaintext(self: *PureZigRecordStream, out: []u8) Error!usize {
         if (self.failed) |err| return err;
         if (self.lifecycle == .closed or self.lifecycle == .failed) return error.StreamClosed;
         if (self.inbound_plaintext.read(out)) |n| return n;
+        try self.raisePendingTerminalError();
         if (self.peer_closed) return error.EndOfStream;
         return error.WouldBlock;
     }
@@ -524,8 +530,8 @@ pub const PureZigRecordStream = struct {
         const made_progress = try self.processCarrierInputBudget(record_budget_remaining);
         if (self.peer_closed) return made_progress;
         if (self.inbound_carrier.len > 0) return made_progress;
-        if (self.inbound_plaintext.len > 0 or self.inbound_handshake.len > 0) return made_progress;
-        return self.fail(error.TruncatedStream);
+        try self.deferTerminalReadError(error.TruncatedStream);
+        return made_progress;
     }
 
     fn feedCarrierCiphertext(self: *PureZigRecordStream, bytes: []const u8) Error!usize {
@@ -578,6 +584,24 @@ pub const PureZigRecordStream = struct {
         return self.processCarrierInputBudget(&record_budget_remaining);
     }
 
+    fn hasBufferedInboundContent(self: *const PureZigRecordStream) bool {
+        return self.inbound_plaintext.len > 0 or self.inbound_handshake.len > 0;
+    }
+
+    fn deferTerminalReadError(self: *PureZigRecordStream, err: Error) Error!void {
+        if (self.hasBufferedInboundContent()) {
+            self.pending_terminal_read_error = err;
+            return;
+        }
+        return self.fail(err);
+    }
+
+    fn raisePendingTerminalError(self: *PureZigRecordStream) Error!void {
+        const err = self.pending_terminal_read_error orelse return;
+        if (self.hasBufferedInboundContent()) return;
+        return self.fail(err);
+    }
+
     fn closeCarrier(self: *PureZigRecordStream) void {
         if (self.carrier) |carrier| {
             std.debug.assert(!carrier.owns_handle or carrier.closeFn != null);
@@ -595,6 +619,7 @@ pub const PureZigRecordStream = struct {
         self.inbound_handshake.clear();
         self.initial_parser.reset();
         self.ciphertext_parser.reset();
+        self.pending_terminal_read_error = null;
         self.bridge.deinit();
         self.closeCarrier();
         return err;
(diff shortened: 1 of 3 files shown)


Jaruso (commented):
Two correctness issues are called out inline.

One additional gap: handshake-epoch alerts still never reach `handleAlert`. `feedHandshakeCiphertext` calls `Bridge.openHandshake`, which rejects any inner content type other than `.handshake`, so a normal fatal alert received during the handshake is latched as `UnexpectedRecordContent` rather than `PeerFatalAlert`. Since this PR describes explicit fatal-alert handling, please either route handshake-epoch `.alert` records through the alert handler or narrow/document the scope and add coverage for the current behavior.

src/tls/encrypted_stream.zig
```diff
@@ -487,11 +493,21 @@ pub const PureZigRecordStream = struct {
     }
 
     fn handleAlert(self: *PureZigRecordStream, alert: []const u8) Error!void {
-        if (alert.len >= 2 and alert[1] == 0) {
+        if (alert.len != 2) return self.fail(error.MalformedAlert);
+        const level = alert[0];
+        const description: alerts.AlertDescription = std.enums.fromInt(alerts.AlertDescription, alert[1]) orelse return self.fail(error.PeerFatalAlert);
+        if (level == 1 and description == .close_notify) {
```
Jaruso: **[P1] Classify TLS 1.3 alerts by description, not the legacy level byte.**

RFC 8446 §6 makes severity implicit in the alert description and says the `level` field can be ignored. As written, `{ 2, close_notify }` becomes `PeerFatalAlert`, while `{ 1, unexpected_message }` becomes `MalformedAlert` even though it must be treated as an error alert regardless of level. Please branch on `description` first (`close_notify` → `peer_closed`; error/unknown descriptions → `PeerFatalAlert`) and add coverage for both level-mismatch cases.

src/tls/encrypted_stream.zig
```diff
@@ -426,15 +432,15 @@ pub const PureZigRecordStream = struct {
                 const maybe_read_len = carrier.read(buf[0..read_cap]) catch |err| switch (err) {
                     error.WouldBlock => null,
                     error.EndOfStream => eof: {
-                        self.peer_closed = true;
+                        try self.handleCarrierEof();
```
Jaruso: **[P1] Drain already-buffered records before declaring truncation.**

`processCarrierInput` is capped at `drive_record_budget` (8) records per call, but one 4 KiB carrier read can contain far more than eight small TLS records. If `close_notify` is record 9+ and the next `carrier.read` returns EOF, this path calls `handleCarrierEof()` while the closure alert is still sitting in `inbound_carrier`, latching a false `TruncatedStream` and clearing buffered plaintext. On EOF, record the transport EOF and process the buffered ciphertext until no further progress is possible before deciding whether `peer_closed` was observed. A regression test with more than eight tiny records followed by `close_notify` in the same carrier read would cover this.

Jaruso (commented):
The pushed commit addresses both original P1 findings and the encrypted handshake-epoch alert gap. The new regression coverage is useful.

I left two remaining issues inline. One additional completeness suggestion: RFC 8446 §6.1 defines `user_canceled` (90) as a closure alert that is generally warning-level and should be followed by `close_notify`. Because `AlertDescription` omits it, `std.enums.fromInt` currently classifies it as an unknown error and latches `PeerFatalAlert`. Since #354 explicitly calls for warning-alert handling, consider adding `user_canceled` and a test that it does not become a protocol-fatal alert.

src/tls/encrypted_stream.zig
```diff
@@ -481,17 +487,37 @@ pub const PureZigRecordStream = struct {
     fn openHandshakeSink(self: *PureZigRecordStream, epoch: events.EncryptionEpoch, sink: anytype) Error!void {
         var plaintext_buf: [record_codec.max_ciphertext_fragment_len]u8 = undefined;
         for (sink.items[0..sink.len]) |record| {
-            const opened = try self.bridge.openHandshake(epoch, record, &plaintext_buf);
-            try self.inbound_handshake.append(opened.inner.content);
+            const opened = try self.bridge.openProtected(epoch, record, &plaintext_buf);
```
Jaruso, line 508: **[P1] Initial-epoch plaintext alerts still bypass alert handling.**

For `.initial`, `Bridge.openProtected` rejects every outer content type except `.handshake` before this switch can see `.alert`. A server can send a plaintext error alert (for example `protocol_version`) in response to `ClientHello` before handshake traffic keys exist, so this still surfaces as `UnexpectedRecordContent` rather than `PeerFatalAlert`. Please allow an initial plaintext alert record to reach `handleAlert` and add coverage for the pre-key path as well as the encrypted handshake-epoch path.

src/tls/encrypted_stream.zig
```diff
@@ -481,17 +487,37 @@ pub const PureZigRecordStream = struct {
...
             return;
         }
-        return self.fail(error.UnsupportedRecordContent);
+        return self.fail(error.PeerFatalAlert);
+    }
+
+    fn handleCarrierEof(self: *PureZigRecordStream) Error!bool {
+        if (self.peer_closed) return false;
+        var made_progress = false;
+        while (self.inbound_carrier.len > 0 and !self.peer_closed) {
```
Jaruso: **[P2] Preserve the bounded-work contract after transport EOF.**

This loop repeatedly invokes `processCarrierInput(drive_record_budget)` until the entire carrier queue is drained, so observing EOF effectively removes `drive_record_budget`. A peer can queue a large number of tiny valid TLS records and then FIN, causing a single nonblocking `drive()` call to decrypt the whole backlog. Consider latching a `carrier_eof` state, continuing to process only the normal record budget on each drive, and reporting `TruncatedStream` only once buffered/parser input is exhausted without `close_notify`. A regression test with substantially more than one budget of tiny records would protect the scheduling guarantee.

src/tls/encrypted_stream.zig
```diff
@@ -310,8 +317,9 @@ pub const PureZigRecordStream = struct {
         return consumed;
     }
 
-    pub fn markPeerClosed(self: *PureZigRecordStream) void {
-        self.peer_closed = true;
+    pub fn markPeerEof(self: *PureZigRecordStream) Error!void {
```
Jaruso, line 324: **[P1] Preserve authenticated plaintext when caller-fed input reaches EOF.**

`markPeerEof()` still calls `fail()` immediately, and `fail()` clears `inbound_plaintext` and `inbound_handshake`. In caller-fed mode, a sequence such as `feedCiphertext(valid_application_record)` followed by transport EOF therefore discards the authenticated record before the application can read it. This also differs from the new carrier-backed path, which deliberately defers truncation while plaintext is buffered. Please route this API through the same pending-EOF state: allow buffered data to drain, then surface and latch `TruncatedStream`. A regression test should feed one valid application record without `close_notify`, call `markPeerEof()`, successfully read the payload, and receive `TruncatedStream` on the next read.

src/tls/encrypted_stream.zig
```diff
@@ -481,17 +493,39 @@ pub const PureZigRecordStream = struct {
...
+        if (description == .user_canceled) return;
+        return self.fail(error.PeerFatalAlert);
+    }
+
+    fn handleCarrierEof(self: *PureZigRecordStream, record_budget_remaining: *usize) Error!bool {
+        if (self.peer_closed) return false;
+        const made_progress = try self.processCarrierInputBudget(record_budget_remaining);
+        if (self.peer_closed) return made_progress;
+        if (self.inbound_carrier.len > 0) return made_progress;
+        if (self.inbound_plaintext.len > 0 or self.inbound_handshake.len > 0) return made_progress;
```
Jaruso: **[P1] Make deferred truncation observable after the last buffered byte is read.**

When EOF follows valid application data without `close_notify`, this return preserves the plaintext, which is good. However, after the caller drains that plaintext, `readPlaintext()` returns `WouldBlock` and `readiness()` reports `wants_read = false`, `can_read_plaintext = false`, and `peer_closed = false`. A readiness-driven caller has no reason to invoke `drive()` again, so the pending `TruncatedStream` can remain hidden indefinitely. Please represent this as a pending terminal read error that is returned/latching once buffered plaintext is exhausted (or otherwise expose pending terminal work in readiness). Add a test that drives once through data+FIN, drains the plaintext, and verifies the next read surfaces `TruncatedStream` without requiring an extra externally-triggered drive.

Jaruso (commented):
The latest commit fixes the previously reported initial-alert, bounded-work, and `user_canceled` gaps, and the new CI run is green. Two EOF-state issues remain inline. They appear to share the same underlying need: represent truncation as a pending terminal read error so already-authenticated bytes can drain without losing the eventual `TruncatedStream`.

src/tls/encrypted_stream.zig
```diff
@@ -310,14 +321,17 @@ pub const PureZigRecordStream = struct {
         return consumed;
     }
 
-    pub fn markPeerClosed(self: *PureZigRecordStream) void {
-        self.peer_closed = true;
+    pub fn markPeerEof(self: *PureZigRecordStream) Error!void {
+        if (self.peer_closed) return;
+        self.carrier_eof = true;
+        return self.deferTerminalReadError(error.TruncatedStream);
```
Jaruso, line 327: **[P1] Stop accepting new plaintext writes once truncation is pending.**

This now preserves buffered authenticated data, but `pending_terminal_read_error` leaves the lifecycle `.open`, so `writePlaintext()` can still seal and queue new application data after transport EOF is known. The OpenSSL backend latches this same condition as failure immediately, and the shared failure contract expects writes to return the terminal error. Please reject new writes (and clear `can_write_plaintext`) while the terminal error is pending, without discarding inbound data. A regression test can mark EOF with plaintext buffered, verify that plaintext remains readable, and verify `write()` already returns `TruncatedStream`.

src/tls/encrypted_stream.zig
```diff
@@ -361,7 +375,7 @@ pub const PureZigRecordStream = struct {
             return .{ .peer_closed = self.peer_closed };
         }
         return .{
-            .wants_read = self.canAcceptCarrierRead(),
+            .wants_read = !self.carrier_eof and self.canAcceptCarrierRead(),
```
Jaruso, line 380: **[P2] Keep the deferred error visible through readiness.**

The new tests issue a second `read()` after consuming the payload, which surfaces `TruncatedStream`. A readiness-driven caller may instead consume exactly the advertised bytes, re-check readiness, and stop: then `wants_read`, `can_read_plaintext`, and `peer_closed` are all false while `pending_terminal_read_error` is still waiting. Please make the pending terminal condition read-ready (or latch it as the final buffered read drains) so the stream cannot settle into an idle state before the error is observed. A regression test should drain the payload, inspect readiness before another read, and confirm the terminal condition is still advertised.

Jaruso (commented):
The latest commit fixes both previous EOF findings and adds good regression coverage. CI is green.

Two contract-level gaps remain in the pending truncation state, called out inline. They both come from detecting a terminal transport condition while temporarily keeping the read side alive to drain authenticated bytes.

Jaruso (commented):
Final full-pass verdict: **not ready yet**. There is one remaining protocol-correctness blocker, and this is my consolidated final change request.

The OpenSSL backend currently treats receipt of peer `close_notify` as a full-duplex close. In `opensslStreamWrite`, `peer_closed` causes every later write to return `StreamClosed`, and `opensslStreamReadiness` also disables `can_write_plaintext`. TLS 1.3 closure is directional: peer `close_notify` closes only the peer-to-us direction. Our local write direction remains usable until we send our own `close_notify`. OpenSSL explicitly supports `SSL_write()` after a remotely initiated shutdown and before the local `SSL_shutdown()`.

Please make exactly these changes:

1. In `opensslStreamWrite`, remove `or self.peer_closed` from the initial lifecycle guard. Keep all existing protections for `stream_failure`, non-open lifecycle, `close_requested`, and pending-write retry identity.
2. In `opensslStreamReadiness`, stop gating `can_write_plaintext` on `!self.peer_closed`. After the read side observes peer `close_notify`, readiness should report `peer_closed = true` **and** keep `can_write_plaintext = true` while the local side is still open and no retry/close operation is pending.
3. Add an OpenSSL one-sided-shutdown regression test that:
   - has the client send `close_notify` without completing the reverse shutdown;
   - verifies the server stream reads `EndOfStream` and reports `peer_closed = true` while remaining writable;
   - writes a final application payload from server to client and verifies the client receives it;
   - then closes the server side and verifies the reciprocal `close_notify` completes cleanly.
4. Add equivalent pure-Zig coverage, or preferably a shared one-sided-close conformance helper exercised by both backends, so the directional contract cannot diverge again.

After those exact changes pass CI, I consider this PR ready. I found no other blockers in the complete PR, alert handling, EOF/truncation state machine, buffer draining, bounded-drive behavior, fatal-error latching, duplicate/data-after-close behavior, or existing bidirectional shutdown paths.

Jaruso (commented):
Final review: **ready to merge**.

The last protocol-correctness blocker is fixed. The OpenSSL backend now keeps the local write direction open after receiving peer `close_notify`, readiness advertises that state correctly, and both OpenSSL and pure-Zig regression tests cover sending final application data before the reciprocal shutdown. CI run 580 is green on this head.

I found no remaining blockers across the full PR.

Issue #139, opened 2026-06-06
perf: implement streaming reverse proxy mode with bounded buffering


Jaruso:
## Summary
Implement a streaming reverse-proxy path that relays request and response bodies incrementally instead of fully buffering upstream responses or large request bodies by default.

## Research basis
Nginx, Caddy, Apache mod_proxy, HAProxy, and Envoy all expose buffering and streaming controls because proxy buffering is a major performance and memory lever. Envoy emphasizes flow control and bounded buffering; Caddy exposes request/response buffer and flush controls; Nginx exposes `proxy_buffering` and `proxy_request_buffering`.

## Current Tardigrade context
The current proxy path includes code paths that build requests into memory and read upstream responses into buffers before returning parsed responses. That is simple and safe for small responses, but it limits efficiency for large bodies, slow clients, streaming APIs, and high-concurrency reverse-proxy workloads.

## Proposed design
- Add a streaming relay path for proxy responses.
- Stream client request bodies to upstream when route policy allows.
- Preserve full-buffer mode for features that require whole-body inspection/transformation.
- Support chunked and fixed-length response bodies.
- Ensure hop-by-hop headers are handled correctly.
- Add configurable flush behavior for low-latency streaming routes.
- Add max buffered bytes safeguards.

## Acceptance criteria
- Large upstream responses can be relayed without buffering the entire response in memory.
- Large client request bodies can be proxied without full request buffering when enabled.
- Existing small-response behavior remains correct.
- Streaming mode is configurable per route or proxy policy.
- Memory use remains bounded under slow-client and slow-origin scenarios.
- Tests cover fixed-length, chunked, no-body, HEAD, 204/304, upstream close, and client abort cases.

## Metrics to add or verify
- `tardigrade_proxy_streaming_requests_total`
- `tardigrade_proxy_buffered_requests_total`
- `tardigrade_proxy_buffered_bytes_current`
- `tardigrade_proxy_buffered_bytes_total`
- `tardigrade_proxy_client_aborts_total`
- `tardigrade_proxy_upstream_aborts_total`
- TTFB for proxied responses

## Benchmark requirements
Run:
- reverse proxy small response benchmark
- reverse proxy 1 MiB / 16 MiB response benchmark
- large upload benchmark
- slow-client download benchmark

Compare RSS, p99 latency, throughput, bytes buffered, and CPU/request before/after.

## Complexity
Medium to high.

## Risk
[...]

PR #417, opened 2026-07-15
perf: add route-level proxy streaming policy

Jaruso:
## Summary
- add inheritable per-location proxy streaming policy (`inherit`, `off`, `response`, `full`) for config-file and location-blob routes
- replace streaming eligibility booleans with typed fallback reasons for response streaming and full-mode upload decisions
- expose fallback reasons through `tardigrade_proxy_streaming_fallback_total{reason=...}` and document the route policy/metric
- add unit and integration coverage for route overrides, fallback metrics, and upload eligibility reasons

This implements the policy and observability slice recommended as PR 1 in #139. Request-upload completion and the broader gateway conformance matrix remain separate follow-up slices from the issue body.

## Validation
- `git diff --check`
- `zig build test-integration -- --test-filter "proxy route streaming policy overrides global mode"`

Closes part of #139

### DIFF
--- a/src/edge_gateway.zig
+++ b/src/edge_gateway.zig
@@ -1311,33 +1311,63 @@ fn targetSupportsStdHttpStreaming(cfg: *const edge_config.EdgeConfig, target: []
     return !(cfg.upstream_tls_client_cert.len > 0 and std.mem.startsWith(u8, cfg.upstream_base_url, "https://"));
 }
 
-fn streamingUploadAllowedBeforeBodyRead(
+const RequestUploadStreamingEligibility = union(enum) {
+    stream,
+    fallback: gproxy_runtime.StreamingFallbackReason,
+    not_applicable,
+};
+
+fn targetStreamingFallbackReason(
+    cfg: *const edge_config.EdgeConfig,
+    target: []const u8,
+) ?gproxy_runtime.StreamingFallbackReason {
+    const trimmed = std.mem.trim(u8, target, " \t\r\n");
+    if (trimmed.len == 0) return .unsupported_route_type;
+    if (gp.unixSocketPathFromEndpoint(trimmed) != null) return .unix_socket_target;
+    if (isHttpUrl(trimmed)) {
+        if (cfg.upstream_tls_client_cert.len > 0 and std.mem.startsWith(u8, trimmed, "https://")) return .upstream_mtls_target;
+        return null;
+    }
+    if (gp.unixSocketPathFromEndpoint(cfg.upstream_base_url) != null) return .unix_socket_target;
+    if (cfg.upstream_tls_client_cert.len > 0 and std.mem.startsWith(u8, cfg.upstream_base_url, "https://")) return .upstream_mtls_target;
+    return null;
+}
+
+fn streamingUploadEligibilityBeforeBodyRead(
     cfg: *const edge_config.EdgeConfig,
     request: *const http.Request,
-) bool {
-    if (!cfg.proxy_streaming_mode.requestStreamingEnabled()) return false;
-    if (!request.method.hasRequestBody()) return false;
-    if (request.hasTransferEncoding()) return false;
-    const content_length = request.contentLength() orelse return false;
-    if (content_length == 0) return false;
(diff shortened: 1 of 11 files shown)


src/edge_gateway.zig
```diff
@@ -1311,35 +1311,65 @@ fn targetSupportsStdHttpStreaming(cfg: *const edge_config.EdgeConfig, target: []
...
-        .proxy_pass => |target| targetSupportsStdHttpStreaming(cfg, target),
-        else => false,
+        .proxy_pass => |target| if (targetStreamingFallbackReason(cfg, target)) |reason| .{ .fallback = reason } else .stream,
+        else => .{ .fallback = .unsupported_route_type },
     };
 }
 
+fn mayNeedStreamingRequestBodyPreRead(cfg: *const edge_config.EdgeConfig) bool {
+    if (cfg.proxy_streaming_mode.requestStreamingEnabled()) return true;
+    for (cfg.location_blocks) |block| {
```
Jaruso, line 1369: This only scans the base config's global mode and top-level `location_blocks`. `resolveRequestConfig` can later replace `location_blocks` with those from the selected `server_block`, so a vhost-only `proxy_streaming full` route is missed here: the connection takes the full-body read path before the Host-specific config is resolved, and the upload can never stream. Please scan `cfg.server_blocks[*].location_blocks` as well (or otherwise make the pre-read decision server-block aware), and add an integration test with global `off` plus a `server` location set to `full`.

src/edge_gateway.zig
```diff
@@ -1425,47 +1455,56 @@ fn handleConnection(conn: anytype, session: *ConnectionSession, cfg: *const edge
...
-            } else {
-                session.pending_len = 0;
-            }
-            request = parse_result.request;
-            request_initialized = true;
+            },
+            .fallback, .not_applicable => {
+                if (upload_eligibility == .fallback) {
+                    const reason = upload_eligibility.fallback;
+                    state.metricsRecordProxyStreamingFallback(reason.metricLabel());
```
Jaruso, line 1476: This upload-phase increment is followed by an independent response-phase increment in `handleLocationProxyPass`. For example, a `POST` with `Content-Length` to a route with `proxy_streaming off` records `policy_disabled` twice for one request, while the Prometheus HELP text says this is the total number of proxied requests. The runtime path also increments before checking whether a started streaming upload must be rejected with 502, in which case no buffered path is used. Please either split/label the phases, explicitly define this as fallback *events*, or arrange to record once per request. A POST integration assertion would catch the double count.

src/edge_gateway.zig
```diff
@@ -1311,35 +1311,65 @@ fn targetSupportsStdHttpStreaming(cfg: *const edge_config.EdgeConfig, target: []
...
-) bool {
-    if (!cfg.proxy_streaming_mode.requestStreamingEnabled()) return false;
-    if (!request.method.hasRequestBody()) return false;
-    if (request.hasTransferEncoding()) return false;
-    const content_length = request.contentLength() orelse return false;
-    if (content_length == 0) return false;
-    if (content_length > cfg.request_limits.effectiveMaxBodySize()) return false;
+) RequestUploadStreamingEligibility {
+    if (!request.method.hasRequestBody()) return .not_applicable;
+    if (request.hasTransferEncoding()) return .{ .fallback = .chunked_request_upload };
```
Jaruso: Suggestion: consider resolving the matched route and its policy before these body-shape reasons. When any route enables `full`, every body-bearing request reaches this function; an explicitly `off` route with chunked framing or no `Content-Length` is currently reported as `chunked_request_upload` / `missing_content_length` rather than `policy_disabled`. That makes the operator-visible reason depend on request framing instead of the route's actual policy. A small precedence test would make the intended behavior explicit.

Jaruso (commented):
Overall, this is a solid implementation of the policy/observability slice from #139. The route policy model is clean, the fallback reasons are bounded, and the integration test covers both directions of the global/route override.

I found one functional gap that can prevent request streaming for virtual-host routes, plus a metric-semantics issue that can make the new counter misleading. I also left a suggestion about fallback-reason precedence. I would address the vhost pre-read case and decide whether this metric counts requests or phase-level fallback events before merging.

Issue #140, opened 2026-06-06
perf: add watermark-based backpressure for proxy and connection buffers


Jaruso:
## Summary
Add explicit high/low watermark controls for client, upstream, and per-stream buffers so Tardigrade can pause reads, resume reads, and bound memory use under slow-client, slow-origin, and large-body workloads.

## Research basis
Envoy treats flow control and buffer limits as core proxy behavior, with listener/cluster/protocol-level limits and metrics for paused/resumed reads. HAProxy and Apache HTTP/2 expose buffer and stream controls. Streaming proxying without watermarks simply moves memory pressure around; bounded buffering is required for first-class stability.

## Proposed design
- Define buffer accounting scopes: per connection, per request/stream, per upstream, global.
- Add configurable high/low watermarks.
- Pause client reads when downstream-to-upstream buffers exceed high watermark.
- Pause upstream reads when upstream-to-client buffers exceed high watermark.
- Resume reads when buffers fall below low watermark.
- Add hard caps that fail safely with clear error metrics.
- Integrate with streaming proxy mode.

## Acceptance criteria
- Watermarks are enforced for proxy request bodies and response bodies.
- Reads pause/resume without busy looping.
- Buffer caps produce deterministic 413/502/503-style behavior where appropriate.
- Metrics expose current buffered bytes and pause/resume counters.
- Tests cover slow client, slow upstream, large upload, large download, and abort paths.

## Metrics to add or verify
- `tardigrade_buffered_bytes_current{scope=...}`
- `tardigrade_buffer_high_watermark_events_total{scope=...}`
- `tardigrade_buffer_read_pauses_total{side=client|upstream}`
- `tardigrade_buffer_read_resumes_total{side=client|upstream}`
- `tardigrade_buffer_limit_exceeded_total{scope=...}`

## Benchmark requirements
Run streaming proxy benchmarks with:
- fast upstream / slow client
- slow upstream / fast client
- large uploads
- many concurrent medium responses

Compare RSS, p99, p999, abort correctness, and paused/resumed read behavior.

## Complexity
Medium.

## Risk
Medium. Incorrect flow control can deadlock or stall connections.

## Dependencies
Depends on #136. Strongly related to #139.

PR #418, opened 2026-07-15
proxy: add buffer accounting metrics

Jaruso:
## Summary
- add shared proxy buffer accounting directions, scopes, limits, and owner-local reservation helper
- add validated proxy buffer watermark env config and expose configured limits in Prometheus
- add bounded-cardinality buffer current/high-watermark/pause/resume/limit metrics, fold legacy buffered-response bytes into the shared metric surface, and account streaming relay buffers during active streaming proxy exchanges
- document the new operator-facing metrics and config knobs

Refs #140

## Validation
- git diff --check
- zig build test-integration -- --test-filter "proxy streams upstream response incrementally"

### DIFF
--- a/src/edge_config.zig
+++ b/src/edge_config.zig
@@ -403,6 +403,8 @@ pub const EdgeConfig = struct {
     max_connection_memory_bytes: usize,
     /// Maximum buffered bytes accepted from an upstream HTTP response body.
     max_buffered_upstream_response_bytes: usize,
+    /// Shared proxy body-buffer watermarks and future aggregate hard limits.
+    proxy_buffer_limits: http.proxy_buffer_account.Limits,
     /// Proxy streaming policy for data-plane HTTP proxy routes.
     proxy_streaming_mode: ProxyStreamingMode,
     /// Maximum relay buffer used per direction for streaming proxy transfers.
@@ -1165,6 +1167,13 @@ pub fn loadFromEnv(allocator: std.mem.Allocator) !EdgeConfig {
     const max_buffered_upstream_resp_str = envOrDefault(allocator, "TARDIGRADE_MAX_BUFFERED_UPSTREAM_RESPONSE_BYTES", "262144") catch unreachable;
     defer allocator.free(max_buffered_upstream_resp_str);
     const max_buffered_upstream_response_bytes = std.fmt.parseInt(usize, max_buffered_upstream_resp_str, 10) catch 256 * 1024;
+    const proxy_buffer_limits = http.proxy_buffer_account.Limits{
+        .per_stream_low_watermark = parseIntEnv(usize, allocator, "TARDIGRADE_PROXY_BUFFER_PER_STREAM_LOW_WATERMARK_BYTES", 256 * 1024),
+        .per_stream_high_watermark = parseIntEnv(usize, allocator, "TARDIGRADE_PROXY_BUFFER_PER_STREAM_HIGH_WATERMARK_BYTES", 768 * 1024),
+        .per_stream_hard_limit = parseIntEnv(usize, allocator, "TARDIGRADE_PROXY_BUFFER_PER_STREAM_HARD_LIMIT_BYTES", 1024 * 1024),
+        .per_origin_hard_limit = parseIntEnv(usize, allocator, "TARDIGRADE_PROXY_BUFFER_PER_ORIGIN_HARD_LIMIT_BYTES", 0),
+        .global_hard_limit = parseIntEnv(usize, allocator, "TARDIGRADE_PROXY_BUFFER_GLOBAL_HARD_LIMIT_BYTES", 0),
+    };
 
     const proxy_streaming_mode_str = envOrDefault(allocator, "TARDIGRADE_PROXY_STREAMING_MODE", "off") catch unreachable;
     defer allocator.free(proxy_streaming_mode_str);
@@ -1538,6 +1547,7 @@ pub fn loadFromEnv(allocator: std.mem.Allocator) !EdgeConfig {
         .connection_pool_size = connection_pool_size,
         .max_connection_memory_bytes = max_connection_memory_bytes,
         .max_buffered_upstream_response_bytes = max_buffered_upstream_response_bytes,
+        .proxy_buffer_limits = proxy_buffer_limits,
         .proxy_streaming_mode = proxy_streaming_mode,
         .proxy_stream_buffer_size = proxy_stream_buffer_size,
         .max_total_connection_memory_bytes = max_total_connection_memory_bytes,
@@ -2641,6 +2651,10 @@ pub fn validate(cfg: *const EdgeConfig) !void {
         std.log.err("config validation failed: TARDIGRADE_MAX_BUFFERED_UPSTREAM_RESPONSE_BYTES must be at least 1", .{});
         return error.InvalidConfigValue;
     };
+    cfg.proxy_buffer_limits.validate() catch {
+        std.log.err("config validation failed: proxy buffer watermarks must satisfy low < high <= hard and aggregate hard limits must be 0 or at least the per-stream hard limit", .{});
+        return error.InvalidConfigValue;
+    };
     validateProxyStreamBufferSize(cfg.proxy_stream_buffer_size) catch {
         std.log.err("config validation failed: TARDIGRADE_PROXY_STREAM_BUFFER_SIZE must be between 1 and 1048576 bytes", .{});
         return error.InvalidConfigValue;
@@ -3329,4 +3343,30 @@ test "validate proxy stream buffer size rejects zero and oversized buffers" {
     try validateProxyStreamBufferSize(1024 * 1024);
     try std.testing.expectError(error.InvalidConfigValue, validateProxyStreamBufferSize(0));
     try std.testing.expectError(error.InvalidConfigValue, validateProxyStreamBufferSize(1024 * 1024 + 1));
+}
+
+test "proxy buffer limits validate low high hard ordering" {
+    try (http.proxy_buffer_account.Limits{
+        .per_stream_low_watermark = 256 * 1024,
+        .per_stream_high_watermark = 768 * 1024,
+        .per_stream_hard_limit = 1024 * 1024,
+        .per_origin_hard_limit = 0,
+        .global_hard_limit = 0,
+    }).validate();
+
+    try std.testing.expectError(error.InvalidBufferLimits, (http.proxy_buffer_account.Limits{
+        .per_stream_low_watermark = 768 * 1024,
+        .per_stream_high_watermark = 256 * 1024,
+        .per_stream_hard_limit = 1024 * 1024,
+        .per_origin_hard_limit = 0,
+        .global_hard_limit = 0,
+    }).validate());
+
+    try std.testing.expectError(error.InvalidBufferLimits, (http.proxy_buffer_account.Limits{
+        .per_stream_low_watermark = 256 * 1024,
+        .per_stream_high_watermark = 768 * 1024,
+        .per_stream_hard_limit = 1024 * 1024,
+        .per_origin_hard_limit = 512 * 1024,
+        .global_hard_limit = 0,
+    }).validate());
 }
--- a/src/edge_config.zig
+++ b/src/edge_config.zig
@@ -2659,6 +2659,10 @@ pub fn validate(cfg: *const EdgeConfig) !void {
         std.log.err("config validation failed: TARDIGRADE_PROXY_STREAM_BUFFER_SIZE must be between 1 and 1048576 bytes", .{});
         return error.InvalidConfigValue;
     };
+    validateProxyBufferLimitsCoverRelayAllocations(cfg.proxy_buffer_limits, cfg.proxy_stream_buffer_size) catch {
+        std.log.err("config validation failed: TARDIGRADE_PROXY_BUFFER_PER_STREAM_HARD_LIMIT_BYTES must be at least the effective streaming relay allocation", .{});
+        return error.InvalidConfigValue;
+    };
 
     // Timeout relationship sanity checks — warn only, these are not hard errors
     // because operators may have intentional unusual configurations.
@@ -2706,6 +2710,11 @@ fn validateBufferedUpstreamResponseLimit(limit: usize) !void {
 
 fn validateProxyStreamBufferSize(size: usize) !void {
     if (size == 0 or size > 1024 * 1024) return error.InvalidConfigValue;
+}
+
+fn validateProxyBufferLimitsCoverRelayAllocations(limits: http.proxy_buffer_account.Limits, proxy_stream_buffer_size: usize) !void {
+    const effective_relay_bytes = @max(proxy_stream_buffer_size, 16 * 1024);
+    if (limits.per_stream_hard_limit < effective_relay_bytes) return error.InvalidConfigValue;
 }
 
 /// Emit log warnings for configurations that are valid but operationally risky.
@@ -3343,6 +3352,17 @@ test "validate proxy stream buffer size rejects zero and oversized buffers" {
     try validateProxyStreamBufferSize(1024 * 1024);
     try std.testing.expectError(error.InvalidConfigValue, validateProxyStreamBufferSize(0));
     try std.testing.expectError(error.InvalidConfigValue, validateProxyStreamBufferSize(1024 * 1024 + 1));
+}
+
+test "proxy buffer hard limit covers effective streaming relay allocation" {
+    const limits = http.proxy_buffer_account.Limits{
+        .per_stream_low_watermark = 1024,
+        .per_stream_high_watermark = 2048,
+        .per_stream_hard_limit = 4096,
+        .per_origin_hard_limit = 0,
+        .global_hard_limit = 0,
+    };
+    try std.testing.expectError(error.InvalidConfigValue, validateProxyBufferLimitsCoverRelayAllocations(limits, 16 * 1024));
 }
 
 test "proxy buffer limits validate low high hard ordering" {
--- a/src/gateway_proxy.zig
+++ b/src/gateway_proxy.zig
@@ -770,12 +770,16 @@ fn streamViaH2Pool(
                 if (streaming_body) |sb| {
                     var sent: usize = @min(sb.initial_bytes.len, sb.content_length);
                     if (sent > 0) {
+                        var initial_reservation = ProxyBufferReservation.init(.downstream_to_upstream, proxy_buffer_limits, proxy_buffer_observer);
+                        try initial_reservation.reserve(sent);
                         conn.writeStreamingRequestBody(stream, sb.initial_bytes[0..sent], sent == sb.content_length) catch |err| {
+                            initial_reservation.releaseAll();
                             conn.finishStreaming(stream);
                             if (!conn.healthy()) h2_pool.evict(key, conn);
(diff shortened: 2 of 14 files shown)


src/gateway_proxy_runtime.zig
```diff
@@ -522,6 +522,26 @@ pub fn handleLocationProxyPass(
     const fallback_reason: StreamingFallbackReason = switch (streamingEligibilityForDataPlaneProxyRequest(cfg, matched_block, &resolved, upstream_url.value, max_attempts)) {
         .stream => {
             state.recordUpstreamAttemptStart(selection.base_url);
+            const relay_buffer_bytes = @max(cfg.proxy_stream_buffer_size, 16 * 1024);
```
Jaruso: **Blocking:** This reserves only the caller-owned `read_buf` capacity. When `executeStreamingHttpProxyRequest` selects HTTP/2, upstream DATA is first retained in the `upstream_h2` per-stream body queue; those queued bytes are not represented here. As a result, `tardigrade_buffered_bytes_current` under-reports the path with the largest application-owned queue (potentially by the receive-window amount). Please account queue append/drain ownership in the H2 layer, or explicitly scope this instrumentation to HTTP/1 rather than presenting it as shared streaming accounting.

src/gateway_proxy_runtime.zig
```diff
@@ -522,6 +522,26 @@ pub fn handleLocationProxyPass(
...
+            state.metricsRecordProxyBufferBytes(.upstream_to_downstream, .stream, relay_buffer_bytes);
+            state.metricsRecordProxyBufferBytes(.upstream_to_downstream, .global, relay_buffer_bytes);
+            const account_upload_relay = streaming_request_body != null;
+            if (account_upload_relay) {
+                state.metricsRecordProxyBufferBytes(.downstream_to_upstream, .stream, relay_buffer_bytes);
```
Jaruso: **Blocking:** The streaming executor allocates a single `read_buf` and reuses it for the request upload and then the response relay; these are sequential phases, not two simultaneous relay allocations. Adding a second downstream-to-upstream reservation for the entire exchange doubles the same memory in both the `stream` and `global` gauges for streaming uploads. Please track one shared reservation and change its direction by phase, or account only genuinely distinct concurrently owned buffers.

src/gateway_state.zig
```diff
@@ -452,6 +452,7 @@ pub const GatewayState = struct {
     active_mux_subscriptions: usize, // runtime accounting [connection_mutex]
     connection_memory_estimate_bytes: usize, // cfg-derived; updated on reload [runtime_mutex]
     max_total_connection_memory_bytes: usize, // cfg snapshot; updated on reload [runtime_mutex]
+    proxy_buffer_limits: http.proxy_buffer_account.Limits, // cfg snapshot; updated on reload [runtime_mutex]
```
Jaruso, line 455: **Blocking:** This snapshot is initialized at startup, but `gateway_shutdown.applyReloadedRuntimeConfig` does not update it. After a successful hot reload, active requests use the new config while `tardigrade_buffer_config_limit_bytes` continues exporting the old limits. Also, because this field is documented as protected by `runtime_mutex`, Prometheus rendering should take a snapshot under that same lock before reading it.

src/http/metrics.zig
```diff
@@ -290,12 +315,81 @@ pub const Metrics = struct {
...
+                .global => &self.proxy_buffer_downstream_to_upstream_global_current,
+                else => return,
+            },
+            .upstream_to_downstream => switch (scope) {
+                .stream => &self.proxy_buffer_upstream_to_downstream_stream_current,
+                .global => &self.proxy_buffer_upstream_to_downstream_global_current,
+                else => return,
+            },
+        };
+        slot.* = if (value >= slot.*) 0 else slot.* - value;
```
Jaruso: **Suggestion:** Please avoid silently clamping an accounting underflow to zero. This slot is a process aggregate, so a duplicate or oversized release from one stream can erase reservations belonging to other active streams and hide the exactly-once ownership bug. A checked subtraction/assertion (plus an imbalance test) would make failures visible instead of corrupting the gauge.

Jaruso (commented):
The configuration and bounded-cardinality metric surface look well structured, and CI is green. I found three correctness issues that should be addressed before merging:

1. Streaming uploads count one shared relay allocation twice.
2. HTTP/2's per-stream body queue is not included, so the new current-byte gauge under-reports the path with the largest retained queue.
3. Hot reload leaves the exported limit snapshot stale and would be racy unless the read is locked too.

Please add focused tests that assert directional/current gauges during a streaming upload, account HTTP/2 queue append/drain ownership, and verify hot-reloaded limits appear in Prometheus. I also left a suggestion to make release underflow fail loudly rather than erase other active reservations.

src/gateway_proxy_runtime.zig
```diff
@@ -522,6 +522,16 @@ pub fn handleLocationProxyPass(
     const fallback_reason: StreamingFallbackReason = switch (streamingEligibilityForDataPlaneProxyRequest(cfg, matched_block, &resolved, upstream_url.value, max_attempts)) {
         .stream => {
             state.recordUpstreamAttemptStart(selection.base_url);
+            const relay_buffer_bytes = @max(cfg.proxy_stream_buffer_size, 16 * 1024);
+            state.metricsRecordProxyBufferBytes(.upstream_to_downstream, .stream, relay_buffer_bytes);
```
Jaruso: **Blocking:** The accounting needs to move to the actual protocol/phase owners. In HTTP/1, the heap `read_buf` is the upstream-to-downstream response relay, but `sendStreamingProxyRequest` has a distinct 16 KiB stack upload relay that is currently unaccounted. In HTTP/2, the heap `read_buf` is reused downstream-to-upstream during upload and upstream-to-downstream during response drain, while the much larger per-stream `Stream.body` queue is still omitted. Because this wrapper runs before protocol selection, it cannot represent those ownership/direction differences and it also counts H2 despite the docs saying this slice is HTTP/1-only. Please instrument the concrete owner sites (or pass an accounting object/callback into them) rather than assigning one upstream-to-downstream reservation around the entire executor.

src/edge_config.zig
```diff
@@ -2641,6 +2651,10 @@ pub fn validate(cfg: *const EdgeConfig) !void {
         std.log.err("config validation failed: TARDIGRADE_MAX_BUFFERED_UPSTREAM_RESPONSE_BYTES must be at least 1", .{});
         return error.InvalidConfigValue;
     };
+    cfg.proxy_buffer_limits.validate() catch {
```
Jaruso, line 2654: **Blocking:** These limits are not tied to the buffers they are supposed to bound. The response `read_buf` is always at least 16 KiB, and HTTP/1 full streaming adds a separate fixed 16 KiB upload relay; this config accepts a 4 KiB hard limit and the request still succeeds. The production path also bypasses `Account.reserve()`, so no `BufferLimitExceeded` result or limit-exceeded metric can occur. Please cross-validate each effective direction-specific reservation against `per_stream_hard_limit` and/or make the concrete buffer owner reserve through `Account` before allocating or entering the relay.

src/http/metrics.zig
```diff
@@ -290,12 +315,83 @@ pub const Metrics = struct {
...
+
+    pub fn recordProxyBufferHighWatermark(self: *Metrics, direction: proxy_buffer_account.Direction, scope: proxy_buffer_account.Scope) void {
+        if (scope != .stream) return;
+        switch (direction) {
+            .downstream_to_upstream => self.proxy_buffer_high_watermark_downstream_to_upstream_stream += 1,
+            .upstream_to_downstream => self.proxy_buffer_high_watermark_upstream_to_downstream_stream += 1,
+        }
+    }
+
+    pub fn recordProxyBufferLimitExceeded(self: *Metrics, direction: proxy_buffer_account.Direction, scope: proxy_buffer_account.Scope) void {
```
Jaruso, line 404: **Suggestion:** This counter, along with the pause/resume counters below, currently has no production caller; only the serializer test increments it. The operator docs state that hard-limit exceedance and pause/resume transitions are observable, but these series will remain permanently zero in real traffic. Either wire the actual transition sites in this PR or describe these series explicitly as reserved for a later enforcement slice so operators do not interpret zero as evidence that no events occurred.

Jaruso (commented):
Second pass: the reload snapshot/race and aggregate underflow handling from the first review are addressed, and the double reservation itself is gone. Two correctness issues remain before this metric surface is reliable:

1. The relay reservation is still outside the protocol- and phase-specific owner, so HTTP/2 is partially/misleadingly counted and streaming uploads are charged to the response direction for their entire upload phase.
2. Config validation permits a per-stream hard limit below the relay allocation that runtime actually makes; the new integration test demonstrates a 4 KiB hard limit with a 16 KiB reservation that succeeds without a limit-exceeded event.

I also left a note that the exported pause/resume and limit-exceeded counters currently have no production call sites, despite the operator docs describing them as observable events.

src/http/proxy_buffer_account.zig
```diff
@@ -0,0 +1,177 @@
...
+        self.current = next;
+        if (!self.above_high_watermark and self.current >= self.limits.per_stream_high_watermark) {
+            self.above_high_watermark = true;
+            self.high_watermark_events += 1;
+        }
+    }
+
+    pub fn release(self: *Account, bytes: usize) void {
+        self.current = if (bytes >= self.current) 0 else self.current - bytes;
+        if (self.above_high_watermark and self.current <= self.limits.per_stream_low_watermark) {
```
Jaruso, line 120: **Blocking:** The owner-local account still silently accepts an oversized or duplicate release by clamping to zero. The aggregate metrics layer was changed to reject underflow, but this foundational helper can hide the exactly-once ownership bug before it reaches that layer. Please make `release` checked (`!void` or an assertion) and add a test proving an over-release is reported and leaves `current` unchanged.

src/http/proxy_buffer_account.zig
```diff
@@ -0,0 +1,177 @@
...
+    }
+
+    pub fn reserve(self: *Account, bytes: usize) !void {
+        const next = std.math.add(usize, self.current, bytes) catch {
+            self.limit_exceeded_events += 1;
+            return error.BufferLimitExceeded;
+        };
+        if (next > self.limits.per_stream_hard_limit) {
+            self.limit_exceeded_events += 1;
+            return error.BufferLimitExceeded;
```
Jaruso, line 108: **Suggestion:** `Account` accepts `.connection`, `.origin`, and `.global` scopes, but `reserve` always enforces `per_stream_hard_limit`. Either restrict this owner-local type to `.stream` in this slice, or select the applicable limit by scope; otherwise a future aggregate account will silently enforce the wrong policy.

src/gateway_proxy_runtime.zig
```diff
@@ -522,6 +522,16 @@ pub fn handleLocationProxyPass(
...
+            state.metricsRecordProxyBufferBytes(.upstream_to_downstream, .global, relay_buffer_bytes);
```
Jaruso: **Blocking:** The stream and global values represent one logical reservation, but these two calls take and release `metrics_mutex` independently. `metricsToPrometheus` can snapshot between them and expose an impossible state (`stream=16KiB, global=0`, or the inverse during release). Please add one state API that records/releases all scopes for a reservation under a single lock, ideally fed by the concrete owner-local account.

tests/integration.zig
```diff
@@ -3625,6 +3628,18 @@ test "proxy full streaming mode relays fixed-length upload beyond request buffer
...
     try std.testing.expectEqualStrings(payload, captured);
+
+    var metrics = try sendRequest(allocator, tardigrade.port, .{
+        .method = "GET",
+        .path = "/status/metrics",
+        .body = "",
+        .headers = &.{},
+    });
+    defer metrics.deinit();
+    try std.testing.expect(std.mem.find(u8, metrics.body, "tardigrade_buffer_high_watermark_events_total{direction=\"upstream_to_downstream\",scope=\"stream\"} 1") != null);
```
Jaruso, line 3668: **Test gap:** Scraping only after the request completes cannot verify a `current` gauge or directional ownership; it only verifies eventual zero, so this still passes if the upload relay was never recorded. Hold the upload or response deliberately stalled, scrape while the relay is active, assert the expected downstream-to-upstream/upstream-to-downstream current values, then unblock and assert return to zero. Also, this modified test is not run by the current CI workflow, and the PR validation section lists a different integration-test filter—please run and record this exact test.

src/gateway_state.zig
```diff
@@ -3377,4 +3431,29 @@ test "served Prometheus metrics expose h2 streaming upload fallback counter" {
...
+
+test "served Prometheus metrics reflect updated proxy buffer limit snapshot" {
+    var gs: GatewayState = undefined;
+    initMetricsJsonTestState(&gs, std.testing.allocator);
+    defer gs.h2_pool.deinit();
+    defer gs.upstream_pool.deinit();
+    defer gs.mux_subscriptions_by_device.deinit();
+
+    gs.runtime_mutex.lock();
+    gs.proxy_buffer_limits = .{
```
Jaruso, line 3469: **Suggestion:** This proves Prometheus rendering reads the snapshot, but it does not guard the actual reload regression. The original bug was the missing assignment in `applyReloadedRuntimeConfig`; this test manually performs that assignment and would still pass if the production reload line were deleted. Add a focused apply/reload test that changes the config through the production helper and then verifies the exported values.

Jaruso (commented):
I completed a clean-room review of the current head across all 11 changed files and traced the surrounding HTTP/1, HTTP/2, reload, metrics serialization, and test code.

I corrected and resolved the earlier overbroad double-count thread: HTTP/1 has a distinct stack upload relay, while HTTP/2 reuses the heap relay. The remaining canonical ownership comment now describes the protocol-specific problem precisely.

Blocking findings remaining at this head:
1. Accounting is attached above protocol/phase selection, so HTTP/1 upload ownership, HTTP/2 direction changes, and the HTTP/2 body queue cannot be represented correctly.
2. Configured hard limits are not connected to effective allocations or `Account.reserve()`.
3. The owner-local account still hides over-release.
4. Stream/global parts of one logical reservation can be exposed as an inconsistent scrape snapshot.

I also left targeted suggestions for scope semantics and tests. CI is green, but CI only runs unit tests; the modified integration test is not part of that workflow, and the PR validation section lists a different integration-test filter.

src/http/upstream_h2.zig
```diff
@@ -740,6 +791,7 @@ pub fn H2Conn(comptime Transport: type) type {
                     const n = @min(avail, out.len);
                     @memcpy(out[0..n], stream.body.items[stream.body_read_off..][0..n]);
                     stream.body_read_off += n;
+                    stream.releaseQueuedBody(n);
```
Jaruso: **Blocking:** This releases the queue reservation—and shortly afterward replenishes the HTTP/2 stream window—as soon as the bytes are copied into the caller’s `out` buffer. The gateway has not written those bytes downstream yet. While that write is blocked, the same proxy stream owns the unaccounted `read_buf` bytes and the origin may refill the queue using the newly granted credit, so `per_stream_hard_limit` does not bound the combined queue + relay ownership.

Please keep the reservation/stream credit until the downstream write succeeds, or transfer the reservation explicitly to a separately accounted relay owner and acknowledge consumption afterward. Also compact or rotate consumed prefixes: `clearRetainingCapacity()` only runs when the consumer catches the queue completely, so a continuously non-empty stream can leave `body.items.len`/capacity growing with total response bytes even though the gauge tracks only unread bytes.

src/http/upstream_h2.zig
```diff
@@ -1139,6 +1191,17 @@ pub fn H2Conn(comptime Transport: type) type {
                         deliver = false;
                     }
                 }
+                if (deliver) {
+                    s.recordQueuedBody(fr.payload.len) catch |err| {
+                        if (err == error.BufferLimitExceeded) {
+                            if (s.err == null) s.err = err;
```
Jaruso, line 1220: **Blocking:** A local hard-limit failure is stored in `s.err`, but `finishStreaming()` sends `RST_STREAM` only when `stream.err == null`. This therefore prevents the reset that should stop the origin. After the worker removes the stream, later DATA is treated as unknown and the current `replenish_stream = true` path can send stream-level `WINDOW_UPDATE` for that removed stream.

Please distinguish locally generated aborts from peer/connection errors and send `RST_STREAM(CANCEL)` after releasing `state_mutex` when this reservation fails. Add a test proving the origin sees the reset and all queued reservations return to zero.

src/gateway_proxy.zig
```diff
@@ -1685,6 +1762,9 @@ fn streamProxyOverTransport(
     var aborted = false;
     var reusable = head.http_1_1 and !head.connection_close;
     if (body_allowed) {
+        var response_reservation = ProxyBufferReservation.init(.upstream_to_downstream, proxy_buffer_limits, proxy_buffer_observer);
+        try response_reservation.reserve(read_buf.len);
```
Jaruso: **Blocking:** `readUpstreamHead()` can leave already-read response-body bytes in `rb/read_buf`, but this reservation is created only after `writeStreamedUpstreamResponseHeadFromHeaders()` completes. If the downstream header write blocks, the proxy already owns body bytes in `read_buf` while both current gauges remain zero.

Start the response-buffer reservation before reading the upstream head (or at least before any downstream head write after that read), and retain it through the relay/teardown paths.

src/gateway_proxy.zig
```diff
@@ -740,12 +794,16 @@ fn streamViaH2Pool(
                             h2_pool.release(conn);
                             return error.ClientAborted;
                         }
+                        var upload_reservation = ProxyBufferReservation.init(.downstream_to_upstream, proxy_buffer_limits, proxy_buffer_observer);
+                        try upload_reservation.reserve(n);
```
Jaruso: **Blocking:** This reservation covers only chunks read later into `read_buf`. `sb.initial_bytes` is written before this loop and can contain request-body bytes received in the same read as the headers. On H2 those bytes are completely unaccounted; on H1 the fixed 16 KiB relay reservation can under-report a much larger initial slice held in the 256 KiB request buffer.

Account the initial body ownership before writing it upstream, with the reservation covering its actual byte count or an explicit ownership transfer from the inbound request buffer. Add a test where headers and a substantial body prefix arrive together; the current integration test sends them separately and cannot catch this.

Jaruso (commented):
Full re-review of the current head. The owner placement, relay-limit validation, checked release, atomic stream/global metrics updates, active HTTP/1 upload scrape, and production reload test are now addressed. Four correctness blockers remain:

1. HTTP/2 queue bytes are released and stream credit is replenished before the copied relay bytes have been written downstream, so ownership temporarily becomes unaccounted and the hard limit does not bound queue + relay memory.
2. A local HTTP/2 `BufferLimitExceeded` marks `stream.err`, but `finishStreaming` only resets streams with `err == null`; this can abandon the stream without sending RST_STREAM and later grant WINDOW_UPDATE credit to the removed stream.
3. HTTP/1 can already hold response-body bytes in `read_buf` after parsing the head, but the reservation begins only after the downstream response head has been written.
4. Request-body bytes pre-read with the request headers (`initial_bytes`) bypass the H2 upload reservation and can exceed the fixed H1 relay amount being reported.

Please add focused tests for H2 slow-downstream ownership/compaction, hard-limit reset and zero-on-teardown behavior, HTTP/1 body bytes arriving with the response head, and uploads whose body bytes arrive in the initial header read. CI is green, but the workflow still does not run `test-integration`, and the PR validation section still names a different integration-test filter than the modified upload test.

PR #419, opened 2026-07-15
Finalize tardi as the public CLI

Jaruso:
### Motivation
- Make the shorter `tardi` name the canonical public CLI while retaining `tardigrade` as a compatibility alias for existing installs and scripts.
- Ensure packaging, install scripts, CI/release workflows, service units, docs, and benchmarks all consistently reference the canonical CLI name.

### Description
- Change the build output name to `tardi` and update integration/test install lookup to `zig-out/bin/tardi` via `build.zig` and related build options.
- Make `tardi` the primary binary in release archives and in `scripts/package-release-archive.sh` while creating a `tardigrade` compatibility symlink in archives and package trees.
- Update the installer `scripts/install.sh`, packaging (DEB/RPM/Homebrew) scripts, systemd unit, CI/release workflows, audit script, benchmarks, test harnesses, and README/docs to invoke `tardi` as the canonical CLI and keep `tardigrade` as an alias.
- Adjust packaging build scripts (`packaging/deb/build.sh`, `packaging/rpm/*`, `packaging/homebrew/*`) and test scripts to install or assert `/usr/bin/tardi` and preserve `/usr/bin/tardigrade` as a symlink where appropriate.

### Testing
- Ran `git diff --check` and a shell-syntax check with `bash -n` across the modified scripts which completed successfully.
- Exercised release packaging locally with `scripts/package-release-archive.sh --archive-name test --binary <tmp/tardi> --output-dir <tmp/out> --staging-dir <tmp/stage>` and inspected the produced archive, which succeeded.
- Ran the install smoke harness `scripts/test-install.sh` using a temporary dummy `zig-out/bin/tardi`, which completed successfully.
- Performed a broad `bash -n` lint of packaging and benchmark scripts which passed, and updated unit/integration binary paths in `build.zig` and tests; running `zig build` was attempted but could not be executed because `zig` is not installed in this environment and network-based Zig installer / `apt-get` failed with HTTP 403 from the environment proxy, so a full build/test could not be run here.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a57c0d4cf9c83279d0d30c7ebf0cc48)


build.zig
```diff
@@ -104,7 +104,7 @@ pub fn build(b: *std.Build) void {
     exe_mod.addImport("tls_core", tls_core_mod);
 
     const exe = b.addExecutable(.{
-        .name = "tardigrade",
+        .name = "tardi",
```
Jaruso, line 107: This changes the only Zig build artifact to `tardi`, but `.dockerignore` still whitelists only `!zig-out/bin/tardigrade` after excluding `zig-out/bin/*`. Please update that whitelist; otherwise the new binary is omitted from Docker build contexts.

scripts/test-install.sh
```diff
@@ -69,8 +69,8 @@ grep -F "unsupported architecture: sparc64" "${TMPDIR}/unsupported.log" >/dev/nu
 TARDIGRADE_RELEASE_BASE_URL="file://${RELEASE_ROOT}" \
     "${REPO_ROOT}/scripts/install.sh" --version "$VERSION" --dir "$INSTALL_DIR"
 
-test -x "${INSTALL_DIR}/tardigrade"
-test -e "${INSTALL_DIR}/tardi"
+test -x "${INSTALL_DIR}/tardi"
+test -e "${INSTALL_DIR}/tardigrade"
```
Jaruso, line 73: Please execute the compatibility command too, e.g. `"${INSTALL_DIR}/tardigrade" version >/dev/null`. The PR's compatibility guarantee is specifically that existing `tardigrade` invocations keep working; checking only that a path exists does not directly exercise that contract. The DEB/RPM/Homebrew smoke tests should cover the alias similarly.

Jaruso (commented):
The overall direction looks right, and the archive/package layout is consistent: `tardi` is the real binary and `tardigrade` remains the compatibility entry point. I found a few gaps that should be fixed before merge:

1. **The CLI still identifies the old command in its own help output.** `src/main.zig::printUsage` still prints every example as `tardigrade ...`. After this change, running `tardi --help` will tell users to invoke a different command. Please change those usage lines to `tardi` and add a regression assertion for the canonical name.

2. **`.dockerignore` still excludes the new build artifact.** It currently whitelists `!zig-out/bin/tardigrade`, while `zig-out/bin/tardi` remains matched by `zig-out/bin/*`. Any Docker build context relying on this whitelist will omit the canonical binary. Please whitelist `tardi` instead (or both names if both are intentionally present in the build tree).

3. **The BearClaw example still uses the removed build-output path.** `examples/bearclaw/README.md` invokes `./zig-out/bin/tardigrade` in both the runtime-config example and the suggested startup section. Those commands will fail after this PR because Zig now emits only `zig-out/bin/tardi`.

4. **Exercise the compatibility alias, not only its presence.** The installer, DEB, RPM, and Homebrew changes promise that existing `tardigrade` invocations continue working, but the updated smoke tests execute only `tardi`. Please invoke `tardigrade version` as well and verify both names are removed during package uninstall where applicable.

CI is green, but these paths are not currently covered by it.

Jaruso (commented):
Second independent pass on the same head commit. The four findings from the first review are still outstanding, and I found three additional issues:

1. **Source builds do not receive the promised compatibility alias.** `build.zig` now installs only `zig-out/bin/tardi`. Release archives and OS packages add `tardigrade`, but anyone building from source—or any external script using the documented build tree—loses `zig-out/bin/tardigrade` entirely. That conflicts with the PR's stated compatibility goal for existing scripts. Please add a build-install alias/copy for `tardigrade`, or explicitly narrow the compatibility promise to packaged installations.

2. **The Homebrew test invokes an unsupported CLI form.** `src/main.zig` recognizes the `version` subcommand but not `--version`; the formula currently executes `tardi --version`, so `brew test` should exit with an unknown-option error. Please use `tardi version`. Since compatibility is part of this PR, testing `tardigrade version` there too would be useful.

3. **The operator-visible rename is missing from the changelog.** The `[Unreleased]` section still says ``tardigrade version`` for the current TLS-profile feature and contains no entry explaining that `tardi` is becoming canonical while `tardigrade` remains an alias. The release checklist explicitly requires complete changelog entries for operator-visible changes.

CI, including the Linux packaging smoke job, is green; Homebrew and source-build alias compatibility are not covered by those checks.

build.zig
Jaruso, line 107: Changing the artifact name here means a normal `zig build` produces only `zig-out/bin/tardi`; unlike the release/package scripts, the build graph never creates `zig-out/bin/tardigrade`. That breaks existing source-tree scripts despite the PR's compatibility goal. Please install an alias/copy for the old name as part of the build output, or narrow the stated compatibility scope.

packaging/homebrew/tardigrade.rb
```diff
@@ -54,7 +55,7 @@ def install
 
   test do
     pid = fork do
-      exec bin/"tardigrade", "--version"
+      exec bin/"tardi", "--version"
```
Jaruso, line 58: The CLI parser supports `version`, not `--version`, so this formula test should fail with an unknown option. Please change this to `exec bin/"tardi", "version"`; ideally also execute `bin/"tardigrade", "version"` to cover the compatibility symlink.

build.zig
Jaruso: Please make this exact `.dockerignore` block:

```text
zig-out/*
!zig-out/bin/
zig-out/bin/*
!zig-out/bin/tardi
!zig-out/bin/tardigrade
```

Both entries are required because this PR promises that a normal source build exposes the canonical `tardi` command and the `tardigrade` compatibility entry point. Do not resolve this thread with only `!zig-out/bin/tardi`; that would still exclude the compatibility artifact.

Jaruso: Required result: `zig build` must install **both** executable paths into `zig-out/bin`:

```text
zig-out/bin/tardi
zig-out/bin/tardigrade
```

`tardi` is the real artifact; `tardigrade` may be a relative symlink to `tardi` or a byte-identical fallback copy when symlinks are unavailable. Please implement this in the Zig install graph rather than only in packaging scripts.

Do not resolve until these commands pass after deleting `zig-out`:

```bash
rm -rf zig-out
zig build
test -x zig-out/bin/tardi
test -x zig-out/bin/tardigrade
./zig-out/bin/tardi version >/dev/null
./zig-out/bin/tardigrade version >/dev/null
```

Also add an automated regression check for those exact conditions.

scripts/test-install.sh
Jaruso: Please replace this portion of the installer smoke test with the exact behavioral checks below:

```bash
test -x "${INSTALL_DIR}/tardi"
test -x "${INSTALL_DIR}/tardigrade"
"${INSTALL_DIR}/tardi" version >/dev/null
"${INSTALL_DIR}/tardigrade" version >/dev/null
```

The compatibility path must be executed, not merely checked with `test -e`. Keep the bad-install assertions for both paths. Apply the same rule to the DEB and RPM smoke tests: execute both names after installation and assert both are absent after uninstall.

packaging/homebrew/tardigrade.rb
Jaruso: Please replace the entire formula test with this exact block:

```ruby
test do
  system bin/"tardi", "version"
  system bin/"tardigrade", "version"
end
```

Do not use `--version`; this CLI implements the `version` subcommand. Testing both commands is required because the formula installs and promises both entry points.

README.md
```diff
@@ -105,7 +105,7 @@ For a release-mode binary:
 
 ```bash
-./zig-out/bin/tardigrade --help
+./zig-out/bin/tardi --help
 ```
```
Jaruso, line 110: The install section above still says the installer puts `tardigrade` in `$HOME/.local/bin` as the installed binary. After this change the real binary is `tardi` and `tardigrade` is the compatibility alias. Please update that paragraph and explicitly document both names so the public install contract matches the implementation.

scripts/profile.sh
```diff
@@ -128,20 +128,20 @@ Memory / Allocation Profiling on Linux
...
 Option A — Valgrind massif (allocation-heavy path identification):
   valgrind --tool=massif --pages-as-heap=yes \
-    ./zig-out/bin/tardigrade run
+    ./zig-out/bin/tardi run
   # After exit:
   ms_print massif.out.* | head -100
 
 Option B — heaptrack (lower overhead, live view):
-  heaptrack ./zig-out/bin/tardigrade run
+  heaptrack ./zig-out/bin/tardi run
```
Jaruso, line 136: This now profiles a process named `tardi`, but the next instruction still opens `heaptrack.tardigrade.*.gz`. Update the expected output glob to the new executable name (or use a generic `heaptrack.*.gz` instruction) so the documented command actually finds the generated report.

packaging/systemd/tardigrade.service
```diff
@@ -9,9 +9,9 @@ User=tardigrade
 Group=tardigrade
 WorkingDirectory=/var/lib/tardigrade
 EnvironmentFile=-/etc/tardigrade/tardigrade.env
-ExecStart=/usr/bin/tardigrade run -c /etc/tardigrade/tardigrade.conf
-ExecReload=/usr/bin/tardigrade reload -c /etc/tardigrade/tardigrade.conf
-ExecStop=/usr/bin/tardigrade stop -c /etc/tardigrade/tardigrade.conf
+ExecStart=/usr/bin/tardi run -c /etc/tardigrade/tardigrade.conf
```
Jaruso, line 12: The systemd unit now uses the canonical executable, but the sibling `packaging/launchd/io.barelabs.tardigrade.plist` still has `/usr/local/bin/tardigrade` in `ProgramArguments`. Please update the launchd unit too so all shipped service definitions consistently use `tardi` as described by the PR.

Jaruso (commented):
Consolidated review after a full pass over all 26 changed files plus repository-wide searches for build-output paths, installed executable paths, CLI invocations, service definitions, and compatibility coverage.

### Functional issues

1. **The compiled CLI still presents `tardigrade` as the command name.** `src/main.zig::printUsage` prints every usage line with `tardigrade`, and the daemon success message still says `started tardigrade in background`. The canonical public CLI should identify itself as `tardi`; add a regression test around the generated usage text.

2. **Source-tree compatibility is incomplete.** `build.zig` installs only `zig-out/bin/tardi`, so existing source-tree scripts that invoke `zig-out/bin/tardigrade` break. This also leaves `.dockerignore` whitelisting the removed output name and leaves both executable examples in `examples/bearclaw/README.md` broken. Either create the compatibility alias in the Zig install tree or narrow the stated compatibility promise, and update the remaining paths.

3. **The Homebrew formula test invokes an unsupported command.** The parser accepts `tardi version`, not `tardi --version`. The formula test should use the supported form and should also execute the `tardigrade` compatibility symlink.

4. **The compatibility guarantee is not exercised end to end.** The installer, DEB, RPM, and Homebrew tests verify or execute `tardi`, but do not consistently execute `tardigrade version`; package uninstall checks also verify only the canonical path. Please test both names and verify both are removed on uninstall. The archive test should likewise verify that `tardigrade` resolves to and executes the packaged `tardi` binary.

5. **The standalone launchd service was missed.** `packaging/launchd/io.barelabs.tardigrade.plist` still launches `/usr/local/bin/tardigrade`, while the systemd and Homebrew service definitions were moved to the canonical executable. The alias keeps this functional, but it contradicts the PR's stated goal that service units consistently use `tardi`.

### Documentation and release consistency

6. The README install section still says the installer installs `tardigrade` as the primary binary, rather than `tardi` plus a compatibility alias.
7. The Linux heaptrack instructions launch `tardi` but still tell users to open `heaptrack.tardigrade.*.gz`.
8. `CHANGELOG.md` has no entry for the public CLI rename and its current Unreleased text still says `tardigrade version`.

I also checked the installer replacement order, reusable archive staging directory, archive symlink direction, DEB/RPM manifests, package service paths, release workflow, checksum flow, and binary-audit invocation. I found no additional problems in those areas beyond the items above.

PR #422, opened 2026-07-15
Refactor gateway route execution stages

Jaruso:
## Summary

- extract HTTP/1 direct-response rendering into a shared return-response stage
- split location auth and location action execution out of `routeRequest`
- reuse the direct-response stage for conditional/rewrite returns while preserving existing rewrite return behavior
- make return-response metrics ownership explicit with `{ status, error_code }` results from the renderer
- reject `auth required` HTTP/3 locations with `401` until a protocol-neutral auth stage is available

## Behavior notes

- conditional/rewrite returned responses now go through the same direct-response writer as location returns, so they emit both `X-Request-ID` and `X-Correlation-ID` instead of only `X-Correlation-ID`
- conditional/rewrite returned responses now populate `ctx.response_bytes` for access logging/metrics context: `0` for redirects and `body.len` for body responses
- non-GET/HEAD location `return_response` 405s now populate `ctx.response_bytes` for their JSON error body and report `invalid_request` once through caller-owned metrics
- invalid auth now records `forbidden` in the error metrics registry, matching the existing API error code

Closes #201

## Review follow-up

- added direct `writeReturnResponsePlan` tests for body, redirect, request IDs, configured/security headers, response byte accounting, and exactly-once 405 metrics
- added auth-stage tests for invalid session `401` and invalid bearer `403`, each proving exactly one request/error metric increment
- added an HTTP/3 route test proving `auth required` locations are rejected before action execution
- changed `planDirectResponse` to delegate to `planReturnResponse(true, ...)` to avoid drift

## Validation

- `git diff --check`

`zig build test-integration` was run locally on this branch and passed: 65 passed, 4 skipped. The skipped integration checks in CI match the normal pipeline pattern for this repo.

### DIFF
--- a/src/gateway_handlers.zig
+++ b/src/gateway_handlers.zig
@@ -118,11 +118,91 @@ pub const ReturnResponsePlan = union(enum) {
     body: struct { status: u16, body: []const u8 },
 };
 
+pub const ReturnResponseWriteResult = struct {
+    status: u16,
+    error_code: ?[]const u8 = null,
+};
+
 pub fn planReturnResponse(request_is_get_or_head: bool, status: u16, body: []const u8) ReturnResponsePlan {
     const is_redirect = status >= 300 and status < 400;
     if (!is_redirect and !request_is_get_or_head) return .method_not_allowed;
     if (is_redirect and body.len > 0) return .{ .redirect = .{ .status = status, .location = body } };
     return .{ .body = .{ .status = status, .body = body } };
+}
+
+pub fn planDirectResponse(status: u16, body: []const u8) ReturnResponsePlan {
+    return planReturnResponse(true, status, body);
+}
+
+/// Render a configured direct response for HTTP/1.x routes and rewrite stages.
+///
+/// Route matching, rewrite evaluation, and metrics/logging stay with the caller;
+/// this stage owns shared status/header/body shaping for the socket response.
+pub fn writeReturnResponsePlan(
+    allocator: std.mem.Allocator,
+    writer: anytype,
+    state: *GatewayState,
+    ctx: *http.request_context.RequestContext,
+    plan: ReturnResponsePlan,
+    correlation_id: []const u8,
+    keep_alive: bool,
+) !ReturnResponseWriteResult {
+    switch (plan) {
+        .method_not_allowed => {
+            const payload = try buildApiErrorJson(allocator, "invalid_request", "Method Not Allowed", correlation_id);
+            defer allocator.free(payload);
+            var response = http.Response.json(allocator, payload);
+            defer response.deinit();
+            _ = response.setStatus(.method_not_allowed).setConnection(keep_alive);
+            setRequestIdHeaders(&response, correlation_id);
+            ctx.response_bytes = payload.len;
+            applyResponseHeaders(state, &response);
+            try response.write(writer);
+            return .{ .status = 405, .error_code = "invalid_request" };
+        },
+        .redirect => |r| {
+            var response = http.Response.redirect(allocator, r.location, @enumFromInt(r.status));
+            defer response.deinit();
+            _ = response.setConnection(keep_alive);
+            setRequestIdHeaders(&response, correlation_id);
+            ctx.response_bytes = 0;
+            applyResponseHeaders(state, &response);
+            try response.write(writer);
+            return .{ .status = r.status };
+        },
+        .body => |b| {
+            var response = http.Response.init(allocator);
+            defer response.deinit();
+            _ = response.setStatus(@enumFromInt(b.status))
+                .setBody(b.body)
+                .setContentType("text/plain; charset=utf-8")
+                .setConnection(keep_alive);
+            setRequestIdHeaders(&response, correlation_id);
+            ctx.response_bytes = b.body.len;
+            applyResponseHeaders(state, &response);
+            try response.write(writer);
+            return .{ .status = b.status };
+        },
+    }
+}
+
+pub fn writeReturnResponse(
+    allocator: std.mem.Allocator,
+    writer: anytype,
+    state: *GatewayState,
+    ctx: *http.request_context.RequestContext,
+    request_is_get_or_head: bool,
+    status: u16,
+    body: []const u8,
+    correlation_id: []const u8,
+    keep_alive: bool,
+) !ReturnResponseWriteResult {
+    return writeReturnResponsePlan(allocator, writer, state, ctx, planReturnResponse(request_is_get_or_head, status, body), correlation_id, keep_alive);
+}
+
+fn recordReturnResponseMetrics(state: *GatewayState, result: ReturnResponseWriteResult) void {
+    state.metricsRecord(result.status);
+    if (result.error_code) |code| state.metricsRecordErrorCode(code);
 }
 
 test "planReturnResponse: non-redirect return is GET/HEAD only" {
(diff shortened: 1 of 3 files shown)


src/gateway_handlers.zig
```diff
@@ -125,6 +125,70 @@ pub fn planReturnResponse(request_is_get_or_head: bool, status: u16, body: []con
...
+    ctx: *http.request_context.RequestContext,
+    plan: ReturnResponsePlan,
+    correlation_id: []const u8,
+    keep_alive: bool,
+) !u16 {
+    switch (plan) {
+        .method_not_allowed => {
+            try sendApiError(allocator, writer, .method_not_allowed, "invalid_request", "Method Not Allowed", correlation_id, keep_alive, state);
+            return 405;
+        },
```
Jaruso, line 162: **Blocking:** `sendApiError` already records the 405 request and `invalid_request` error code. `executeLocationAction` then records the returned status again, despite this function's contract saying metrics remain with the caller. Please give metrics exactly one owner and add a test asserting a non-GET/HEAD `return_response` increments the request counter once.

src/gateway_handlers.zig
```diff
@@ -173,107 +245,23 @@ pub fn routeRequest(
...
-                    const auth_message = if (auth_res.failure_reason == .invalid) "Forbidden" else "Unauthorized";
-                    const auth_status_code: u16 = @intFromEnum(auth_status);
-                    try sendApiError(allocator, writer, auth_status, auth_code, auth_message, correlation_id, keep_alive.*, state);
-                    state.metricsRecord(auth_status_code);
-                    state.metricsRecordErrorCode(auth_code);
-                    return auth_status_code;
-                }
+            if (try enforceLocationAuth(allocator, writer, cfg, state, ctx, request, matched, correlation_id, keep_alive.*, client_ip)) |status| {
+                return status;
             }
```
Jaruso, line 375: **Blocking/security question:** this auth stage is invoked only by the HTTP/1 `routeRequest`. `routeHttp3Location` resolves and directly executes the same location blocks without an auth gate that I can find. Please share this auth decision with HTTP/3, reject auth-required h3 locations, or add a test proving enforcement occurs elsewhere; otherwise an `auth required` route appears protocol-dependent.

src/gateway_handlers.zig
```diff
@@ -288,6 +276,129 @@ pub fn routeRequest(
...
+        return null;
+    }
+
+    if (http.session.fromHeaders(&request.headers)) |session_token| {
+        if (state.validateSessionIdentity(allocator, session_token)) |identity| {
+            ctx.setIdentity(identity);
+            return null;
+        }
+        try sendApiError(allocator, writer, .unauthorized, "unauthorized", "Unauthorized", correlation_id, keep_alive, state);
+        state.metricsRecord(401);
```
Jaruso: `sendApiError` already calls both `metricsRecord(401)` and `metricsRecordErrorCode("unauthorized")`; these two calls count the same failed request a second time. The final 401/403 branch below has the same duplication. Please remove one ownership layer and regression-test both branches.

src/gateway_handlers.zig
```diff
@@ -125,6 +125,70 @@ pub fn planReturnResponse(request_is_get_or_head: bool, status: u16, body: []con
...
+            var response = http.Response.init(allocator);
+            defer response.deinit();
+            _ = response.setStatus(@enumFromInt(b.status))
+                .setBody(b.body)
+                .setContentType("text/plain; charset=utf-8")
+                .setConnection(keep_alive);
+            setRequestIdHeaders(&response, correlation_id);
+            ctx.response_bytes = b.body.len;
+            applyResponseHeaders(state, &response);
+            try response.write(writer);
```
Jaruso, line 183: This changes observable access-log behavior for conditional/rewrite returns: the old edge branches left `ctx.response_bytes` at zero, while `logAccessForRequest` uses this field as `bytes_sent`/`response_bytes`. That may be a good correction, but please document and test it (and account for bytes consistently in the 405 JSON branch) rather than letting it ride as an unmentioned refactor side effect.

Jaruso (commented):
## Detailed review — changes requested

I reviewed the complete diff, the pre-refactor implementations, the surrounding HTTP/1 and HTTP/3 dispatch paths, the response writer/error helper, access-log byte accounting, issue #201's acceptance criteria, and the green CI run.

The extraction is directionally good: `routeRequest` is materially easier to read, the action switch appears to preserve the existing HTTP/1 control flow, and sharing the direct-response renderer removes meaningful duplication. I do not see a problem with the basic auth/action extraction itself.

I do want the following addressed before merging:

### 1. Blocking: establish exactly one owner for request/error metrics

`writeReturnResponsePlan` says metrics/logging remain with the caller, but its `.method_not_allowed` branch calls `sendApiError`, which already records both the request status and error code. `executeLocationAction` then records the returned status again. A `POST` to a non-redirect `return_response` location therefore increments the request counter twice.

The extracted auth stage has the same problem more broadly: both error branches call `sendApiError` and then explicitly call `metricsRecord` and `metricsRecordErrorCode`, so 401/403 request and error-code counters are double-counted.

This behavior existed in the inlined code, but this PR is explicitly establishing stage ownership and closing #201. This is the right point to make the boundary correct rather than preserve ambiguous/double accounting.

Please choose one model and apply it consistently:

- preferably make the response renderer side-effect-free with respect to metrics and let the stage caller own all metrics, or
- keep `sendApiError` as the metrics owner and remove all duplicate caller-side increments.

Please add regression tests proving exactly one request increment and one error-code increment for at least:

- non-GET/HEAD `return_response` -> 405;
- invalid session -> 401;
- invalid authorization input -> 403.

### 2. Blocking/security: the new location-auth stage is still HTTP/1-only

`enforceLocationAuth` is called only from the HTTP/1 `routeRequest`. The HTTP/3 `routeHttp3Location` path resolves the same location blocks and directly executes `proxy_pass`, `return_response`, rewrites, and static responses without an equivalent auth gate that I can find.

If HTTP/3 can serve a location configured with `auth required`, this appears to be an authorization bypass: the same route is protected over HTTP/1 but unprotected over HTTP/3.

Before this PR closes #201, please do one of the following:

- extract a protocol-neutral auth decision/stage and invoke it from both HTTP/1 and HTTP/3 dispatch;
- explicitly reject auth-required location actions over HTTP/3 until the shared stage exists; or
- point me to the existing HTTP/3 enforcement path I missed and add a test demonstrating that an unauthenticated HTTP/3 request cannot reach an auth-required location.

If that work is intentionally out of scope, this PR should not close #201 without a clearly linked security follow-up that defines the gap and prevents it from being lost.

### 3. Behavior change: direct rewrite/conditional returns now alter access-log byte counts

The old conditional/rewrite `.returned` branches wrote the body but did not populate `ctx.response_bytes`. The shared renderer now sets it to `body.len`, and `logAccessForRequest` uses that field for both `bytes_sent` and `response_bytes`.

That may be a desirable correction, but it is observably different logging behavior in a PR described as preserving existing rewrite-return behavior. Please either:

- document this as an intentional correction and add a regression test for the access-log byte count, or
- preserve the previous behavior in this refactor and make the accounting change separately.

Please also make byte accounting consistent across all plan variants. The body path sets it, redirect sets zero, but the 405 JSON path currently leaves it untouched even though bytes were sent.

### 4. Add tests around the extracted I/O stage, not only its planner

The new unit test validates only `planDirectResponse`. The higher-risk change is the renderer and ownership boundary. Please add focused tests that exercise `writeReturnResponsePlan` with an in-memory writer and verify:

- status line and body for a normal direct response;
- `Location` and empty body for a redirect;
- both request-ID headers;
- configured response/security headers;
- `ctx.response_bytes`;
- exactly-once metrics behavior for 405/error rendering.

An integration regression covering one conditional return and one rewrite return would also directly protect the two call sites moved in `edge_gateway.zig`.

### 5. Non-blocking follow-ups worth handling while this stage is centralized

- `planDirectResponse` duplicates the redirect/body decision in `planReturnResponse`; it can be implemented as `planReturnResponse(true, status, body)` so the two planners cannot drift.
- The renderer treats GET and HEAD equivalently and always calls `Response.write`, so a configured `return 200 <body>` can emit a body on HEAD. The repository already has `responseBodyAllowed` and `Response.writeHead`; the new response stage is a natural place to make HEAD, 1xx, 204, and 304 body suppression explicit. This is pre-existing behavior, so I am not making it independently blocking for this refactor, but I would prefer a follow-up issue/test if it is not handled here.

Once the metrics ownership and HTTP/3 auth questions are resolved and the extracted renderer has direct regression coverage, I think the structural direction is solid.

Jaruso (commented):
## Re-review complete — no further changes requested

I re-reviewed the updated PR at `5aca8b4d219271ae97c2a7a3fab90c44ebc964e2`, including the complete three-file diff, the replacement response writer, HTTP/1 auth-stage behavior, HTTP/3 location dispatch, metrics serialization, the new regression tests, and the latest successful CI run.

The previous blocking findings are addressed:

- **Return-response metrics now have a single owner.** `writeReturnResponsePlan` no longer calls the metrics-mutating `sendApiError`; it returns `{ status, error_code }`, and each caller records that result once. The 405 test verifies the renderer itself leaves counters untouched and the caller increments the request and `invalid_request` counters exactly once.
- **Auth rejection metrics are no longer duplicated.** `enforceLocationAuth` now relies solely on `sendApiError` for 401/403 status and error-code metrics. Dedicated tests cover invalid session and invalid bearer cases and assert exactly one increment.
- **HTTP/3 auth-required routes fail closed.** `routeHttp3Location` checks the matched location before executing its action and returns a 401 JSON response for every `auth required` location. The new test proves the protected action body is not executed and verifies status/error metrics. This temporarily rejects even valid credentials over HTTP/3, but that limitation is explicit in the PR description and is preferable to the previous protocol-specific authorization bypass.
- **Response-byte and header behavior is explicit.** The shared writer sets body bytes, zero redirect bytes, and JSON error bytes; emits both request-ID headers; applies configured/security headers; and has direct in-memory writer coverage. The PR description now calls out the intentional access-log behavior change.
- **Planner drift is removed.** `planDirectResponse` delegates to `planReturnResponse(true, ...)`.
- **403 observability is complete.** The new `forbidden` counter is initialized, recorded, and exposed in both Prometheus and JSON output in the correct argument order.

I also checked the remaining pre-existing HEAD/1xx/204/304 body-shaping concern. This PR does not worsen that behavior, so I continue to regard it as a separate follow-up rather than a merge blocker.

GitHub does not permit this account to formally approve its own pull request, but from this re-review I have **no additional changes to request**.

PR #423, opened 2026-07-15
crypto: add deterministic TLS and QUIC vector harness

Jaruso:
## Summary

- add a standalone `test-crypto-vectors` target for deterministic TLS/QUIC crypto vectors
- cover vector provenance, provider capability coverage/waivers, HKDF labels, TLS 1.3 key schedule and Finished values, transcript HRR `message_hash` handling, TLS record protection, QUIC initial packet protection, AEAD KATs, X25519, Ed25519, and deferred negative cases
- add `tests/vectors/generate_crypto_vectors.js` to reproduce project-owned HKDF, transcript, TLS record, and QUIC packet fixtures without new dependencies
- wire the harness into `zig build test` and `zig build test-crypto`

## Scope note

This PR establishes the pure-Zig executable vector harness slice for #373. It intentionally does **not** close #373 yet: OpenSSL `CryptoProvider` execution and the remaining full provider-neutral vector expansion are still tracked there.

## Validation

- `node tests/vectors/generate_crypto_vectors.js`
- `zig build test-crypto-vectors --summary all --error-style verbose`
- `git diff --check`

Refs #373

### DIFF
--- a/tests/crypto_vectors.zig
+++ b/tests/crypto_vectors.zig
@@ -1,0 +1,323 @@
+//! Deterministic TLS/QUIC cryptographic vector harness (#373).
+//!
+//! The cases in this file intentionally sit outside the module-local unit
+//! tests. They bind the provider capability matrix to externally sourced
+//! vectors and protocol derivation stages so adding or advertising a crypto
+//! capability requires either positive vector coverage or an explicit waiver.
+
+const std = @import("std");
+const crypto_pkg = @import("crypto");
+const quic = @import("quic");
+const tls_core = @import("tls_core");
+
+const testing = std.testing;
+const provider = crypto_pkg.provider;
+const profile = crypto_pkg.profile;
+const pure_zig = crypto_pkg.pure_zig;
+
+const Coverage = enum {
+    positive,
+    negative,
+    waived,
+};
+
+const VectorMeta = struct {
+    name: []const u8,
+    source: []const u8,
+    license: []const u8,
+};
+
+const vector_meta = [_]VectorMeta{
+    .{ .name = "hkdf-rfc5869-case-1", .source = "RFC 5869 Appendix A.1", .license = "IETF Trust" },
+    .{ .name = "tls13-rfc8448-simple-1rtt", .source = "RFC 8448 Section 3", .license = "IETF Trust" },
+    .{ .name = "tls13-record-protection-aes128-gcm", .source = "Independently computed with Python hmac/hashlib and pycryptodome", .license = "project fixture" },
+    .{ .name = "quic-v1-initial-rfc9001-a1", .source = "RFC 9001 Appendix A.1", .license = "IETF Trust" },
+    .{ .name = "aes-128-gcm-nist-zero-block", .source = "NIST SP 800-38D / CAVP GCM zero-block vector", .license = "public domain" },
+    .{ .name = "aes-256-gcm-nist-zero-block", .source = "NIST SP 800-38D / CAVP GCM zero-block vector", .license = "public domain" },
+    .{ .name = "chacha20-poly1305-rfc8439", .source = "RFC 8439 Section 2.8.2", .license = "IETF Trust" },
+    .{ .name = "x25519-rfc7748-alice-bob", .source = "RFC 7748 Section 6.1", .license = "IETF Trust" },
+    .{ .name = "ed25519-rfc8032-test-1", .source = "RFC 8032 Section 7.1", .license = "IETF Trust" },
+};
+
+fn hexBytes(comptime hex: []const u8) [hex.len / 2]u8 {
+    var bytes: [hex.len / 2]u8 = undefined;
+    _ = std.fmt.hexToBytes(&bytes, hex) catch unreachable;
+    return bytes;
+}
+
+fn expectStage(stage: []const u8, expected: []const u8, actual: []const u8) !void {
+    if (!std.mem.eql(u8, expected, actual)) {
+        std.debug.print("crypto vector mismatch at stage: {s}\n", .{stage});
+        return error.VectorMismatch;
+    }
+}
+
+fn expectVectorError(stage: []const u8, expected_error: anyerror, actual_error: anyerror) !void {
+    if (actual_error != expected_error) {
+        std.debug.print("crypto vector error mismatch at stage: {s}; got {s}\n", .{ stage, @errorName(actual_error) });
+        return error.VectorMismatch;
+    }
+}
+
+fn cryptoProvider() provider.CryptoProvider {
+    const Holder = struct {
+        var entropy = pure_zig.DeterministicEntropy.init(0x373);
+        var provider_instance = pure_zig.Provider.init(entropy.entropy());
+    };
+    return Holder.provider_instance.cryptoProvider();
+}
+
+fn coverageForAlgorithm(algorithm: profile.Algorithm) Coverage {
+    return switch (algorithm) {
+        .hash => |hash| switch (hash) {
+            .sha256 => .positive,
+            .sha384 => .positive,
+        },
+        .hkdf => |hash| switch (hash) {
+            .sha256 => .positive,
+            .sha384 => .positive,
+        },
+        .aead => |aead| switch (aead) {
+            .aes_128_gcm => .positive,
+            .aes_256_gcm => .positive,
+            .chacha20_poly1305 => .positive,
+        },
+        .group => |group| switch (group) {
+            .x25519 => .positive,
+            .secp256r1 => .negative,
+        },
+        .signature => |scheme| switch (scheme) {
+            .ed25519 => .positive,
+            .ecdsa_secp256r1_sha256 => .positive,
+            .rsa_pss_rsae_sha256 => .negative,
+        },
+        .certificate_helper => |helper| switch (helper) {
+            .der_parser => .waived,
+            .chain_builder => .waived,
+            .webpki_validation => .waived,
+        },
+        .entropy => |capability| switch (capability) {
+            .injected_random_bytes => .positive,
+            .secure_zero => .positive,
+            .constant_time_compare => .positive,
+        },
+    };
+}
+
+test "vector manifest records provenance and licensing" {
+    for (vector_meta) |entry| {
+        try testing.expect(entry.name.len > 0);
+        try testing.expect(entry.source.len > 0);
+        try testing.expect(entry.license.len > 0);
+    }
+}
+
+test "capability matrix has vector coverage or explicit waivers" {
+    for (profile.rows) |row| {
+        const coverage = coverageForAlgorithm(row.algorithm);
+        switch (row.pure_zig_status) {
+            .supported => try testing.expectEqual(Coverage.positive, coverage),
+            .provider_deferred, .unsupported => try testing.expect(coverage == .negative or coverage == .waived),
+        }
+    }
+}
+
+test "provider HKDF stages match RFC 5869 and TLS expand-label vectors" {
+    const cp = cryptoProvider();
+
+    const ikm = hexBytes("0b0b0b0b0b0b0b0b0b0b0b0b0b0b0b0b0b0b0b0b0b0b");
+    const salt = hexBytes("000102030405060708090a0b0c");
+    const info = hexBytes("f0f1f2f3f4f5f6f7f8f9");
+    const expected_prk = hexBytes("077709362c2e32df0ddc3f0dc47bba6390b6c73bb50f9c3122ec844ad7c2b3e5");
+    var prk: [provider.Hash.sha256.digestLength()]u8 = undefined;
+    try cp.hkdfExtract(.sha256, &salt, &ikm, &prk);
+    try expectStage("hkdf extract sha256 / RFC 5869 A.1", &expected_prk, &prk);
+
+    const HkdfSha256 = std.crypto.kdf.hkdf.HkdfSha256;
+    const expected_okm = std.crypto.tls.hkdfExpandLabel(HkdfSha256, prk, "derived", &info, 42);
+    var okm: [42]u8 = undefined;
+    try cp.hkdfExpandLabel(.sha256, &prk, "derived", &info, &okm);
+    try expectStage("hkdf expand-label sha256 / TLS label", &expected_okm, &okm);
+
+    const secret384 = [_]u8{0x42} ** provider.Hash.sha384.digestLength();
+    const HkdfSha384 = std.crypto.kdf.hkdf.Hkdf(std.crypto.auth.hmac.sha2.HmacSha384);
+    const expected384 = std.crypto.tls.hkdfExpandLabel(HkdfSha384, secret384, "c hs traffic", "", 48);
+    var out384: [48]u8 = undefined;
+    try cp.hkdfExpandLabel(.sha384, &secret384, "c hs traffic", "", &out384);
+    try expectStage("hkdf expand-label sha384 / TLS label", &expected384, &out384);
+}
+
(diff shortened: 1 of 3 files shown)


tests/crypto_vectors.zig
```diff
@@ -0,0 +1,323 @@
...
+    source: []const u8,
+    license: []const u8,
+};
+
+const vector_meta = [_]VectorMeta{
+    .{ .name = "hkdf-rfc5869-case-1", .source = "RFC 5869 Appendix A.1", .license = "IETF Trust" },
+    .{ .name = "tls13-rfc8448-simple-1rtt", .source = "RFC 8448 Section 3", .license = "IETF Trust" },
+    .{ .name = "tls13-record-protection-aes128-gcm", .source = "Independently computed with Python hmac/hashlib and pycryptodome", .license = "project fixture" },
+    .{ .name = "quic-v1-initial-rfc9001-a1", .source = "RFC 9001 Appendix A.1", .license = "IETF Trust" },
+    .{ .name = "aes-128-gcm-nist-zero-block", .source = "NIST SP 800-38D / CAVP GCM zero-block vector", .license = "public domain" },
```
Jaruso: **Provenance is not reproducible enough for a security vector fixture.**

`"Independently computed with Python hmac/hashlib and pycryptodome"` does not identify the exact generator, dependency version, input encoding, command, or output derivation. A future maintainer cannot independently regenerate this record vector or determine whether it was copied from the implementation under test. The current provenance test only proves that this string is non-empty.

Please check in the generator (for example under `tools/` or `tests/vectors/`) or document the exact, deterministic reproduction procedure alongside the literal fixture. I would include:

- a stable vector/case ID;
- the exact algorithm and parameters;
- the full plaintext/AAD/sequence-number inputs;
- the exact tool and dependency versions;
- the command used to produce the bytes;
- a precise license identifier/source reference;
- the generated key, IV/nonce, ciphertext, and tag as separately named outputs.

CI does not need to run Python; the important part is that the checked-in bytes are independently auditable. It would also help to make test cases reference the metadata entry by ID so metadata cannot become orphaned from the executable test.

tests/crypto_vectors.zig
```diff
@@ -0,0 +1,323 @@
...
+
+fn cryptoProvider() provider.CryptoProvider {
+    const Holder = struct {
+        var entropy = pure_zig.DeterministicEntropy.init(0x373);
+        var provider_instance = pure_zig.Provider.init(entropy.entropy());
+    };
+    return Holder.provider_instance.cryptoProvider();
+}
+
+fn coverageForAlgorithm(algorithm: profile.Algorithm) Coverage {
```
Jaruso: **This coverage table is a hard-coded claim, not coverage derived from executable vector cases.**

The test below can only verify that this switch returns the value written here. It cannot determine whether a vector actually exists or ran. For example, this file marks `ecdsa_secp256r1_sha256` as positive but contains no ECDSA vector; it also marks injected entropy, secure zero, and constant-time comparison positive without a corresponding vector case in this harness. As written, those rows remain green even if the relevant test is deleted.

There is a second modeling problem: a single `Coverage` value cannot satisfy #373's requirement that each supported capability point to **positive and negative** coverage. Supported AEADs, signatures, and key exchange should generally have both valid and invalid cases; `.positive` discards that distinction.

Please derive coverage from a case registry rather than a parallel switch. One workable shape would be a manifest entry containing:

```zig
algorithm: profile.Algorithm,
providers: ProviderSet,
positive_case_ids: []const []const u8,
negative_case_ids: []const []const u8,
waiver: ?Waiver,
```

The runner can record which case IDs actually executed, and the matrix assertion can require the appropriate non-empty sets for each provider/status. A waiver should carry a reason and tracking issue, not just a boolean state. That makes deleting or forgetting a vector fail the matrix test automatically.

tests/crypto_vectors.zig
```diff
@@ -0,0 +1,323 @@
...
+        try testing.expect(entry.name.len > 0);
+        try testing.expect(entry.source.len > 0);
+        try testing.expect(entry.license.len > 0);
+    }
+}
+
+test "capability matrix has vector coverage or explicit waivers" {
+    for (profile.rows) |row| {
+        const coverage = coverageForAlgorithm(row.algorithm);
+        switch (row.pure_zig_status) {
```
Jaruso: **This validates only the pure-Zig column and therefore is not provider-neutral.**

The issue explicitly asks for pure-Zig and OpenSSL providers to run the same compatible cases. This loop ignores `row.openssl_status`, `cryptoProvider()` is hard-coded to `pure_zig.Provider`, and the build target imports only the generic crypto module/TLS/QUIC modules; it never constructs or configures an OpenSSL `CryptoProvider` implementation.

I suggest making the vector cases provider-agnostic functions and invoking them through a provider factory, for example once for `.pure_zig` and once for `.openssl` when that adapter is available in the selected build profile. The coverage assertion should evaluate each provider status independently and report the provider name, algorithm, and missing case IDs.

The repository currently documents the OpenSSL `CryptoProvider` adapter as future work. If that means dual-provider execution cannot be implemented in this PR, then the honest resolution is to split/narrow #373 and avoid `Closes #373` here. Otherwise this test will certify only one side of an acceptance criterion while the PR summary describes the harness as provider-neutral.

tests/crypto_vectors.zig
```diff
@@ -0,0 +1,323 @@
...
+    const ikm = hexBytes("0b0b0b0b0b0b0b0b0b0b0b0b0b0b0b0b0b0b0b0b0b0b");
+    const salt = hexBytes("000102030405060708090a0b0c");
+    const info = hexBytes("f0f1f2f3f4f5f6f7f8f9");
+    const expected_prk = hexBytes("077709362c2e32df0ddc3f0dc47bba6390b6c73bb50f9c3122ec844ad7c2b3e5");
+    var prk: [provider.Hash.sha256.digestLength()]u8 = undefined;
+    try cp.hkdfExtract(.sha256, &salt, &ikm, &prk);
+    try expectStage("hkdf extract sha256 / RFC 5869 A.1", &expected_prk, &prk);
+
+    const HkdfSha256 = std.crypto.kdf.hkdf.HkdfSha256;
+    const expected_okm = std.crypto.tls.hkdfExpandLabel(HkdfSha256, prk, "derived", &info, 42);
```
Jaruso: **The expected Expand-Label output is generated by another in-process implementation instead of being a fixed external vector.**

This is useful as a differential test, but it is not equivalent to an authoritative known-answer test. Both sides ultimately rely on the same Zig standard-library cryptographic primitives and assumptions, so a shared bug or shared interpretation error can produce the same wrong result and still pass. The SHA-384 case below is entirely self-derived in the same manner.

Please hard-code the expected output bytes from an independently generated/checkable fixture and retain the stage label. Ideally:

- use an RFC 8448/8446-derived Expand-Label value where available;
- otherwise generate the literal with the checked-in independent script described in the provenance metadata;
- record the secret, label, context, output length, hash, and expected bytes explicitly;
- keep the `std.crypto.tls.hkdfExpandLabel` comparison as an additional differential assertion, not as the oracle.

That gives three useful comparisons: provider vs fixed vector, stdlib vs fixed vector, and provider vs stdlib. The fixed vector is what prevents both implementations from agreeing on the same defect.

Jaruso (commented):
## Review summary

This is a strong starting point: the standalone build target is useful, the stage-specific mismatch reporting is appropriately restrained, and the RFC-backed TLS/AEAD/X25519/Ed25519 literals provide meaningful regression value.

I would still treat the items below as **blocking before merge and before closing #373**. The main problem is that the PR currently describes a complete, provider-neutral vector harness, but several acceptance criteria are either represented only by hard-coded assertions or are not exercised at all.

### Blocking findings

1. **The capability-coverage check is self-attestation rather than a check of registered vectors.** `coverageForAlgorithm` can label an algorithm `positive` even when no corresponding case exists. ECDSA-P256, injected entropy, secure zero, and constant-time comparison are immediate examples. The single `Coverage` enum also cannot represent the issue’s requirement for both positive and negative coverage for a supported capability.
2. **The harness only runs the pure-Zig provider.** It checks only `pure_zig_status`, constructs only `pure_zig.Provider`, and the test artifact has no OpenSSL-provider wiring. That does not satisfy the provider-neutral/shared-case requirement in #373. If the OpenSSL `CryptoProvider` adapter is intentionally future work, this PR should not close the portion of #373 that requires it; split or narrow the issue explicitly.
3. **The HKDF Expand-Label “expected” values are computed at test runtime by `std.crypto`.** That is useful differential testing, but it is not an externally fixed vector and can allow a shared implementation/assumption bug to pass. The SHA-384 case has no independent expected literal at all.
4. **The TLS case bypasses transcript construction and has no HelloRetryRequest coverage.** Supplying precomputed transcript hashes validates key-schedule functions, but it cannot detect hashing the wrong handshake bytes, ordering errors, or the TLS 1.3 HRR `message_hash` transcript rewrite required by the issue.
5. **The QUIC packet-protection portion is primarily a seal/open round trip.** It does not compare ciphertext, tag, protected header, or final packet bytes with RFC 9001. Additionally, the copied RFC header’s encoded Length field does not describe the synthetic 32-byte payload used here, so this is not a valid complete RFC packet fixture.
6. **The X25519 low-order-point negative case can pass when the invalid point is accepted.** The catch-only form verifies the error only when an error happens; successful derivation falls through and the test succeeds.
7. **CI is currently red on `Test (ubuntu-latest)`.** The ARM and macOS test jobs passed, but the Ubuntu x86 unit-test job failed during `Run unit tests`. The connector log was truncated before the specific failure, so I am not attributing a cause without evidence, but the failing check is independently merge-blocking.

### Recommended structure

A more robust design would make vectors data-driven and make the coverage report derive from executable registrations rather than from a parallel switch. For example, each case could carry a stable case ID, source/version/license metadata, algorithm, provider applicability, and positive/negative classification. Then the harness can build coverage from the cases it actually executes and compare that result with every row in `profile.rows`. Waivers should include a non-empty reason and ideally a tracking issue.

For provider parity, put the reusable cases behind a small provider factory/runner and instantiate the suite once per available provider. In a profile where an adapter is unavailable, keep that absence explicit rather than treating pure-Zig execution as provider-neutral coverage.

I left detailed implementation suggestions inline for each finding.

tests/crypto_vectors.zig
```diff
@@ -0,0 +1,323 @@
...
+    var out384: [48]u8 = undefined;
+    try cp.hkdfExpandLabel(.sha384, &secret384, "c hs traffic", "", &out384);
+    try expectStage("hkdf expand-label sha384 / TLS label", &expected384, &out384);
+}
+
+test "TLS 1.3 key schedule and Finished value match RFC 8448" {
+    const KeySchedule = tls_core.key_schedule.KeySchedule;
+
+    const shared = hexBytes("8bd4054fb55b9d63fdfbacf9f04b9f0d35e6d63f537563efd46272900f89492d");
+    const hello_hash = hexBytes("860c06edc07858ee8e78f0e7428c58edd6b43f2ca3e6e95f02ed063cf0e1cad8");
```
Jaruso, line 284: **This validates the key schedule from precomputed hashes, but it does not test transcript handling or HRR.**

Passing `hello_hash` and later `finished_hash` directly means the harness cannot catch errors in which handshake bytes are included, their ordering, handshake framing, transcript updates, or the TLS 1.3 HelloRetryRequest `message_hash` transformation. #373 specifically calls out transcript/HRR fixtures, so this is an important missing layer rather than a naming nit.

Please keep this test—it is a good isolated key-schedule KAT—but add transcript-driven cases that:

1. feed the actual RFC handshake-message bytes into the project transcript implementation;
2. assert intermediate transcript hashes at named stages;
3. derive the traffic secrets from those computed hashes;
4. include an HRR trace that exercises the synthetic `message_hash` replacement of ClientHello1;
5. fail with the exact transcript stage rather than only at the final secret.

Until those cases exist, I would rename this test to make its narrower scope explicit and avoid claiming transcript coverage in the capability matrix.

tests/crypto_vectors.zig
```diff
@@ -0,0 +1,323 @@
...
+
+    const sample = hexBytes("d1b1c98dd7689fb8ec11d242b123dc9b");
+    try expectStage("quic header-protection mask / RFC 9001 A.3", &hexBytes("437b9aec36"), &secrets.client.headerProtectionMask(sample));
+
+    var header = hexBytes("c300000001088394c8f03e5157080000449e00000002");
+    var plaintext = [_]u8{0} ** 32;
+    var sealed: [plaintext.len + quic.tls_adapter.packet_protection_tag_len]u8 = undefined;
+    const protected = try secrets.client.sealPayload(2, &header, &plaintext, &sealed);
+    var opened: [plaintext.len]u8 = undefined;
+    try expectStage("quic packet-protection open after seal", &plaintext, try secrets.client.openPayload(2, &header, protected, &opened));
```
Jaruso, line 399: **This round trip does not establish that QUIC packet protection matches RFC 9001.**

`sealPayload` and `openPayload` can share the same mistake and still round-trip successfully. The test already has useful fixed checks for initial secrets, nonce, and a header-protection mask, but it never compares the produced ciphertext/tag or the protected header/final packet with the RFC fixture.

There is also a fixture-consistency issue: `header` is copied from the RFC packet and retains that packet's encoded Length field, while the test encrypts only a synthetic 32-byte plaintext. The resulting bytes are cryptographically usable as AEAD AAD, but they do not describe a valid complete packet of the payload size being produced, so this should not be presented as the RFC packet-protection vector.

Please either:

- run the exact RFC Appendix A packet fixture end to end and assert the ciphertext, tag, header-protection sample/mask, protected first byte and packet-number bytes, and final packet; or
- use a smaller independently generated fixture whose QUIC Length field is internally consistent, with all expected bytes checked in.

The tampered-tag rejection should remain as a separate negative case after the known-answer assertions. That combination detects asymmetric implementation bugs and validates the wire representation, not just internal seal/open compatibility.

tests/crypto_vectors.zig
```diff
@@ -0,0 +1,323 @@
...
+    const bob_public = hexBytes("de9edb7d7b7dc1b4d35b61c2ece435373f8343c85b78674dadfc7e146f882b4f");
+    const expected_shared = hexBytes("4a5d9d5ba4ce2de1728e3bf480350f25e07e21c947d19e3376f09b3c1e161742");
+    var shared: [32]u8 = undefined;
+    try cp.deriveSharedSecret(.x25519, &alice_private, &bob_public, &shared);
+    try expectStage("x25519 shared secret / RFC 7748", &expected_shared, &shared);
+
+    const zero_point = [_]u8{0} ** 32;
+    cp.deriveSharedSecret(.x25519, &alice_private, &zero_point, &shared) catch |err| {
+        try expectVectorError("x25519 low-order point rejection", error.InvalidInput, err);
+    };
```
Jaruso: **This negative test silently passes if the low-order point is accepted.**

The `catch` body checks the error value only when `deriveSharedSecret` returns an error. If the provider unexpectedly succeeds, execution falls through to the Ed25519 section and the test still passes, which defeats the security regression this case is meant to enforce.

Use an assertion that requires the error, for example:

```zig
try testing.expectError(
    error.InvalidInput,
    cp.deriveSharedSecret(.x25519, &alice_private, &zero_point, &shared),
);
```

Alternatively, use `catch |err| { ...; return; }` followed by `return error.ExpectedLowOrderPointRejection;`, matching the pattern already used by the AEAD and QUIC authentication-failure tests. The direct `expectError` form is clearer here.

For additional hardening, initialize `shared` to a sentinel before the call and assert it does not contain a usable all-zero shared secret after rejection, if the provider contract specifies output behavior on failure.

PR #426, opened 2026-07-15
Add bounded crypto corpus runner

Jaruso:
## Summary

Refs #374. This PR adds the first pure-Zig-provider slice of the Wycheproof-style negative/edge corpus work. It intentionally does not close #374 because OpenSSL differential execution, RSA-PSS coverage, ECDSA expansion, and broader corpus growth remain follow-up work.

## Included Algorithms and Counts

- AES-128-GCM: 2 cases (1 valid, 1 invalid)
- AES-256-GCM: 2 cases (1 valid, 1 invalid)
- ChaCha20-Poly1305: 3 cases (1 valid, 2 invalid, including invalid nonce length)
- X25519: 2 cases (1 valid, 1 acceptable with explicit `invalid-input` provider policy)
- Ed25519 verification: 2 cases (1 valid, 1 invalid)

Total executed corpus: 5 suites, 11 cases, 5 valid, 5 invalid, 1 acceptable.

## What Changed

- Added `tests/vectors/wycheproof/corpus.json`, a reduced offline schema preserving upstream source file, group index, `tcId`, classification, comments, flags, and reproducible inputs.
- Added `tests/vectors/wycheproof/README.md` with origin/license metadata, upstream commit, schema, parser limits, acceptable-case policy, update procedure, and skipped suites.
- Added `tests/crypto_corpus.zig`, a bounded parser and executor that runs cases through the shared `CryptoProvider` interface.
- Added `tests/crypto_corpus_manifest.zig` and registered the suite IDs with the existing crypto-vector coverage log.
- Added `zig build test-crypto-corpus` and wired it into `test-crypto` and the normal `zig build test` path.

## Parser Limits

- file size: 64 KiB
- JSON nesting depth: 12
- groups per suite: 8
- total cases: 64
- identifier length: 96 bytes
- comments: 256 bytes
- flags per case: 8
- encoded hex field: 512 bytes
- decoded bytes across executable fields: 8192 bytes

The parser rejects duplicate case IDs, unknown classifications, malformed hex, oversized values, unsupported schema versions, missing required fields, unknown algorithms, unsupported operation mappings, unknown execution-affecting fields, and acceptable cases without explicit provider policy.

## Deferred Suites

- RSA-PSS: provider capability remains deferred for #374 follow-up.
- ECDSA-P256-SHA256: supported provider operation, but outside this first merge-sized #374 corpus slice.
- X448 and broader asymmetric formats: unsupported by the current provider capability matrix.
- OpenSSL differential execution: deferred until the OpenSSL `CryptoProvider` adapter exists.

## Validation

- `zig build test-crypto-corpus --summary all --error-style verbose`
- `zig build test-crypto-vectors --summary all --error-style verbose`
- `git diff --check`

### DIFF
--- a/tests/crypto_corpus.zig
+++ b/tests/crypto_corpus.zig
@@ -174,12 +174,14 @@ fn parseCorpus(backing_allocator: std.mem.Allocator, input: []const u8, limits: 
     const skipped_suites = try parseSkippedSuites(allocator, try requireArray(root_object, "skippedSuites"), limits);
 
     var seen_cases = std.StringHashMap(void).init(allocator);
+    var parsed_cases: usize = 0;
     var decoded_total: usize = 0;
     const suites = try parseSuites(
         allocator,
         try requireArray(root_object, "suites"),
         allowlist,
         &seen_cases,
+        &parsed_cases,
         &decoded_total,
         limits,
     );
@@ -234,6 +236,7 @@ fn parseSuites(
     array: std.json.Array,
     allowlist: []const Algorithm,
     seen_cases: *std.StringHashMap(void),
+    parsed_cases: *usize,
     decoded_total: *usize,
     limits: Limits,
 ) ![]Suite {
@@ -252,6 +255,7 @@ fn parseSuites(
             algorithm,
             operation,
             seen_cases,
+            parsed_cases,
             decoded_total,
             limits,
         );
@@ -272,6 +276,7 @@ fn parseGroups(
     algorithm: Algorithm,
     operation: Operation,
     seen_cases: *std.StringHashMap(void),
+    parsed_cases: *usize,
     decoded_total: *usize,
     limits: Limits,
 ) ![]Group {
@@ -289,6 +294,7 @@ fn parseGroups(
                 algorithm,
                 operation,
                 seen_cases,
+                parsed_cases,
                 decoded_total,
                 limits,
             ),
@@ -303,6 +309,7 @@ fn parseCases(
     algorithm: Algorithm,
     operation: Operation,
     seen_cases: *std.StringHashMap(void),
+    parsed_cases: *usize,
     decoded_total: *usize,
     limits: Limits,
 ) ![]Case {
@@ -318,6 +325,8 @@ fn parseCases(
 
         const id = try requireId(obj, "id", limits);
         if (seen_cases.contains(id)) return error.DuplicateCaseId;
+        if (parsed_cases.* >= limits.max_cases) return error.TooManyCases;
+        parsed_cases.* += 1;
         try seen_cases.put(id, {});
 
         const classification = try parseClassification(try requireString(obj, "classification", limits));
@@ -387,9 +396,9 @@ fn runCorpus(corpus: *const Corpus) !Report {
 
 fn executeCase(cp: provider.CryptoProvider, suite: Suite, group: Group, case: Case) !void {
     const observed = switch (suite.operation) {
-        .aead_open => executeAeadOpen(cp, suite.algorithm.aead().?, case),
-        .derive_shared_secret => executeX25519(cp, case),
-        .verify => executeEd25519Verify(cp, case),
+        .aead_open => try executeAeadOpen(cp, suite.algorithm.aead().?, case),
+        .derive_shared_secret => try executeX25519(cp, case),
+        .verify => try executeEd25519Verify(cp, case),
     };
     if (observed != case.expected) {
         std.debug.print(
@@ -409,38 +418,38 @@ fn executeCase(cp: provider.CryptoProvider, suite: Suite, group: Group, case: Ca
     }
 }
 
-fn executeAeadOpen(cp: provider.CryptoProvider, aead: provider.Aead, case: Case) Expected {
-    const plaintext = testing.allocator.alloc(u8, case.ciphertext.len) catch return .invalid_input;
+fn executeAeadOpen(cp: provider.CryptoProvider, aead: provider.Aead, case: Case) !Expected {
+    const plaintext = try testing.allocator.alloc(u8, case.ciphertext.len);
     defer testing.allocator.free(plaintext);
     cp.aeadOpen(aead, case.key, case.nonce, case.aad, case.ciphertext, case.tag, plaintext) catch |err| {
-        return observedFromError(err);
+        return try observedFromError(err);
     };
     if (!std.mem.eql(u8, plaintext, case.message)) return .authentication_failed;
     return .success;
 }
 
-fn executeX25519(cp: provider.CryptoProvider, case: Case) Expected {
+fn executeX25519(cp: provider.CryptoProvider, case: Case) !Expected {
     var out: [provider.max_shared_secret_len]u8 = undefined;
     if (case.shared.len > out.len) return .invalid_input;
     cp.deriveSharedSecret(.x25519, case.private_key, case.public_key, out[0..case.shared.len]) catch |err| {
-        return observedFromError(err);
+        return try observedFromError(err);
     };
     if (!std.mem.eql(u8, out[0..case.shared.len], case.shared)) return .authentication_failed;
     return .success;
 }
 
-fn executeEd25519Verify(cp: provider.CryptoProvider, case: Case) Expected {
+fn executeEd25519Verify(cp: provider.CryptoProvider, case: Case) !Expected {
     cp.verify(.ed25519, case.public_key, case.message, case.signature) catch |err| {
-        return observedFromError(err);
+        return try observedFromError(err);
     };
     return .success;
 }
 
-fn observedFromError(err: anyerror) Expected {
+fn observedFromError(err: anyerror) !Expected {
     return switch (err) {
         error.AuthenticationFailed => .authentication_failed,
         error.InvalidInput => .invalid_input,
-        else => .invalid_input,
+        else => err,
     };
 }
 
@@ -671,6 +680,8 @@ test "crypto corpus parser rejects bounded failure paths" {
     try testing.expectError(error.OversizedValue, parseCorpus(testing.allocator, minimalCorpus(.oversized_id), .{}));
     try testing.expectError(error.DuplicateCaseId, parseCorpus(testing.allocator, minimalCorpus(.duplicate_case_id), .{}));
     try testing.expectError(error.UnknownField, parseCorpus(testing.allocator, minimalCorpus(.unknown_field), .{}));
+    try testing.expectError(error.TooManyCases, parseCorpus(testing.allocator, twoGroupCorpus(), .{ .max_cases = 1 }));
+    try testing.expectError(error.UnsupportedCapability, observedFromError(error.UnsupportedCapability));
 }
 
 test "crypto corpus executes through provider and registered manifest" {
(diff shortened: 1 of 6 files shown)


tests/crypto_corpus.zig
```diff
@@ -0,0 +1,762 @@
...
+            .verify => try ensureOnlyFields(obj, &.{ "id", "upstreamTcId", "classification", "expected", "comment", "flags", "publicKey", "message", "signature" }),
+        }
+
+        const id = try requireId(obj, "id", limits);
+        if (seen_cases.contains(id)) return error.DuplicateCaseId;
+        try seen_cases.put(id, {});
+
+        const classification = try parseClassification(try requireString(obj, "classification", limits));
+        const expected = try parseExpected(try requireString(obj, "expected", limits));
+        if (classification == .acceptable and expected == .success) return error.AcceptableCaseNeedsPolicy;
```
Jaruso: The classification and expected outcome need a validated relationship. As written, a typo such as `valid + authentication-failed` or `invalid + success` is accepted and can bless broken provider behavior. At the same time, `acceptable + success` is rejected even though the README says the explicit `expected` field is the provider policy for ambiguous cases.

Please centralize the policy, for example:

```zig
fn validateExpected(classification: Classification, expected: Expected) !void {
    switch (classification) {
        .valid => if (expected != .success)
            return error.InvalidExpectedOutcome,
        .invalid => if (expected == .success)
            return error.InvalidExpectedOutcome,
        .acceptable => {}, // the explicit expected value is the policy
    }
}
```

Then test every allowed/rejected combination. If `acceptable` needs a stricter policy than an explicit expected outcome, add a separate policy field rather than treating success as inherently invalid.

tests/crypto_corpus.zig
```diff
@@ -0,0 +1,762 @@
...
+    }
+}
+
+fn executeAeadOpen(cp: provider.CryptoProvider, aead: provider.Aead, case: Case) Expected {
+    const plaintext = testing.allocator.alloc(u8, case.ciphertext.len) catch return .invalid_input;
+    defer testing.allocator.free(plaintext);
+    cp.aeadOpen(aead, case.key, case.nonce, case.aad, case.ciphertext, case.tag, plaintext) catch |err| {
+        return observedFromError(err);
+    };
+    if (!std.mem.eql(u8, plaintext, case.message)) return .authentication_failed;
```
Jaruso: This converts a **successful provider return with incorrect plaintext** into `authentication_failed`. That is unsafe for a negative tag case: a provider that incorrectly accepts the tag and emits garbage would still match the fixture’s expected `authentication-failed` result, hiding an authentication-bypass bug.

A successful call and an authentication error must remain distinguishable. For example:

```zig
const Observed = enum {
    success,
    authentication_failed,
    invalid_input,
    output_mismatch,
};

try cp.aeadOpen(...);
if (!std.mem.eql(u8, plaintext, case.message)) return .output_mismatch;
return .success;
```

`output_mismatch` should always make `executeCase` fail, regardless of the fixture’s expected outcome. Please add a fake provider that returns success with wrong bytes for an invalid-tag case and prove the runner rejects it.

Jaruso (commented):
**Review outcome: changes required.** GitHub does not allow an author to submit `REQUEST_CHANGES` on their own pull request, so this is recorded as a commented review rather than an approval/blocking-state review.

I reviewed all six changed files, the full acceptance criteria for #374, the provider error contracts, the existing vector-harness coverage machinery, CI, and every one of the 11 reduced vectors against the pinned C2SP Wycheproof commit.

This is appropriately presented as a partial slice of #374 and should **not** close that issue. The offline CI integration, documentation, explicit skipped-suite records, and selected pure-Zig vectors are useful progress. However, the corpus runner is security-sensitive test infrastructure, and there are several ways it can currently report a passing case without having tested the behavior the fixture claims to test.

## Existing unresolved blockers

### 1. Enforce `max_cases` across the whole corpus, not independently per group

The README and PR description define 64 as a total-case bound, but `parseCases` applies it to each individual group. A corpus containing several groups of 64 cases therefore bypasses the documented bound.

Please carry one total through the parser, for example:

```zig
var total_cases: usize = 0;
const suites = try parseSuites(
    allocator,
    try requireArray(root_object, "suites"),
    allowlist,
    &seen_cases,
    &total_cases,
    &decoded_total,
    limits,
);
```

Then, before parsing a group’s case array:

```zig
if (array.items.len > limits.max_cases or
    total_cases.* > limits.max_cases - array.items.len)
{
    return error.TooManyCases;
}
total_cases.* += array.items.len;
```

Please add a regression with multiple groups whose individual sizes are legal but whose aggregate exceeds the limit.

### 2. Unexpected provider and allocator errors must fail the runner

`observedFromError` maps every error other than `AuthenticationFailed` to `invalid-input`, and allocation failure in `executeAeadOpen` is also converted to `invalid-input`. That can turn infrastructure/provider failure into a passing negative case and violates the provider’s deliberately distinct error taxonomy.

The execution helpers should return an error union and only translate the two expected case outcomes:

```zig
fn observedFromError(err: anyerror) !Expected {
    return switch (err) {
        error.AuthenticationFailed => .authentication_failed,
        error.InvalidInput => .invalid_input,
        else => err,
    };
}
```

Likewise, use `try testing.allocator.alloc(...)` rather than treating OOM as malformed crypto input. Add a fake-provider test proving `UnsupportedCapability` or another unexpected error fails the corpus run rather than satisfying an `invalid-input` case.

## Additional blockers

I left inline comments for five additional issues:

1. A successful provider call that produces wrong bytes is currently reclassified as authentication failure, allowing an authentication-bypass defect to pass an invalid-tag fixture.
2. The parser does not validate the relationship between upstream classification and the declared expected outcome, while simultaneously rejecting a legitimate explicit `acceptable + success` policy.
3. X25519 uses the fixture’s expected shared-secret length as the provider output-buffer length, allowing a malformed expected value to manufacture `InvalidInput` before the peer key is exercised.
4. Ed25519 `tcId` 10 is assigned upstream group index 0 even though it comes from the second upstream group.
5. Manifest coverage validates suite ID/count/source but not algorithm, so metadata drift can credit the wrong algorithm with corpus coverage.

## Parser/executor tests needed

The project review checklist requires every expected error path to be tested. Please add focused coverage for at least:

- aggregate `TooManyCases`
- `TooManyGroups`
- `TooManyFlags`
- `DecodedByteLimitExceeded`
- `UnknownExpectedOutcome`
- `UnsupportedOperation`
- classification/expected-policy combinations
- unexpected provider errors
- successful provider return with wrong output
- malformed X25519 expected-output length

## Acceptance-criteria assessment

- **Selected malformed signatures, X25519 low-order input, AEAD modified tags, and invalid nonce:** partially covered by this slice.
- **Stable offline CI corpus:** implemented; the default `zig build test` path runs it, and current CI is green across the configured test platforms.
- **Explicit skipped algorithms:** represented in the checked-in corpus/manifest, though the missing algorithm cross-check currently weakens the integrity of that accounting.
- **Provider disagreements with reproducible identifiers:** not yet implemented because OpenSSL differential execution is explicitly deferred.
- **Bounded, test-covered parsing:** not yet satisfied due to the global-bound defect and untested/error-collapsing paths above.
- **RSA-PSS, ECDSA expansion, OpenSSL differential execution, and broader corpus growth:** correctly remain open follow-up work under #374.

Once these issues are corrected and the adversarial tests are present, this first merge-sized slice should be in good shape without pretending that #374 itself is complete.

tests/crypto_corpus.zig
```diff
@@ -0,0 +1,762 @@
...
+    return .success;
+}
+
+fn executeX25519(cp: provider.CryptoProvider, case: Case) Expected {
+    var out: [provider.max_shared_secret_len]u8 = undefined;
+    if (case.shared.len > out.len) return .invalid_input;
+    cp.deriveSharedSecret(.x25519, case.private_key, case.public_key, out[0..case.shared.len]) catch |err| {
```
Jaruso: The provider output-buffer length is controlled by the fixture’s expected `shared` value. If `shared` is empty or short, the provider returns `InvalidInput` because X25519 requires a 32-byte output buffer; an `invalid-input` fixture can then pass without testing the supplied public key at all.

Use the algorithm contract for the execution buffer and validate expected output separately:

```zig
const shared_len = provider.Group.x25519.sharedSecretLength();
var out: [provider.max_shared_secret_len]u8 = undefined;

cp.deriveSharedSecret(
    .x25519,
    case.private_key,
    case.public_key,
    out[0..shared_len],
) catch |err| return try observedFromError(err);

if (case.expected == .success and case.shared.len != shared_len)
    return error.InvalidCorpus;
if (!std.mem.eql(u8, out[0..shared_len], case.shared))
    return .output_mismatch;
return .success;
```

Please add a regression where an otherwise valid X25519 case has an empty/short `shared` field; parsing or execution must reject the fixture rather than report the expected provider error.

tests/crypto_corpus.zig
```diff
@@ -0,0 +1,762 @@
...
+fn validateManifestCoverage(corpus: *const Corpus, report: Report) !void {
+    try testing.expectEqual(corpus_manifest.suites.len, report.executed_suites);
+    try testing.expectEqual(corpus_manifest.skipped_suites.len, report.skipped_suites);
+    for (corpus_manifest.suites) |expected_suite| {
+        var matched = false;
+        for (corpus.suites) |actual_suite| {
+            if (!std.mem.eql(u8, expected_suite.id, actual_suite.id)) continue;
+            matched = true;
+            try testing.expectEqual(expected_suite.case_count, countCases(actual_suite));
+            try testing.expectEqualStrings(expected_suite.source_file, actual_suite.source_file);
```
Jaruso, line 496: Manifest validation checks suite ID, case count, and source file, but never checks that the corpus suite’s algorithm matches `corpus_manifest.Suite.algorithm`. A corpus entry could therefore execute ChaCha20 under an AES suite ID while the vector registry still credits AES with negative coverage.

Please map the local algorithm enum to `profile.Algorithm` and compare it here:

```zig
fn profileAlgorithm(algorithm: Algorithm) profile.Algorithm {
    return switch (algorithm) {
        .aes_128_gcm => .{ .aead = .aes_128_gcm },
        .aes_256_gcm => .{ .aead = .aes_256_gcm },
        .chacha20_poly1305 => .{ .aead = .chacha20_poly1305 },
        .x25519 => .{ .group = .x25519 },
        .ed25519 => .{ .signature = .ed25519 },
    };
}

try testing.expect(std.meta.eql(
    expected_suite.algorithm,
    profileAlgorithm(actual_suite.algorithm),
));
```

Also compare the `CaseMeta.algorithm` returned by `log.execute(suite.id)` against `suite.algorithm` in `registerWycheproofCorpusSuites`, so all three metadata layers remain tied together.

tests/vectors/wycheproof/corpus.json
```diff
@@ -0,0 +1,254 @@
...
+    },
+    {
+      "id": "wycheproof-ed25519-verify-reduced",
+      "algorithm": "ED25519",
+      "operation": "verify",
+      "sourceFile": "testvectors_v1/ed25519_test.json",
+      "groups": [
+        {
+          "id": "ed25519-verify-valid-and-invalid",
+          "upstreamGroupIndex": 0,
```
Jaruso, line 225: This group index is not valid for both cases. At the pinned upstream commit, Ed25519 `tcId: 1` is in the first `testGroups` entry, while `tcId: 10` is in the second entry. Combining them under `upstreamGroupIndex: 0` loses the provenance the README and PR description promise to preserve.

Please split this into two reduced groups, for example:

```json
"groups": [
  {
    "id": "ed25519-valid-group",
    "upstreamGroupIndex": 0,
    "cases": [/* tcId 1 */]
  },
  {
    "id": "ed25519-invalid-signature-group",
    "upstreamGroupIndex": 1,
    "cases": [/* tcId 10 */]
  }
]
```

The vector bytes, classifications, comments, and flags otherwise match the pinned upstream data.

PR #429, opened 2026-07-16
crypto: add OpenSSL differential oracle tests

Jaruso:
## Summary
- make the OpenSSL differential registry executable by attaching each declared case to a runner
- compare TLS record, QUIC initial, TLS key schedule, transcript, and Finished outputs against OpenSSL using oracle-side constants/intermediates independent of the production helper under test
- keep the general-profile `test-crypto-openssl` target out-of-process and keep appliance builds free of OpenSSL invocation

Refs #377. This PR covers HKDF, TLS/QUIC derivation, selected transcript behavior, and Finished HMAC independence. Certificate-chain decisions remain delegated to #348, full TLS-over-TCP interop remains delegated to #358, PSK binder failures remain delegated to #362, and detached-tag AEAD / raw-key X25519 / raw-key signature primitive parity is now explicitly tracked by #431.

## Validation
- `zig build test-crypto-openssl --summary all --error-style verbose`
- `zig build test-crypto -Dtls-profile=appliance --summary all --error-style verbose`
- `git diff --check`

## Staged-tree hygiene
- `git diff --cached --check` before commit
- staged files scanned for private-key/token-like patterns before commit

### DIFF
--- a/tests/crypto_openssl_diff.zig
+++ b/tests/crypto_openssl_diff.zig
@@ -8,10 +8,12 @@
 const std = @import("std");
 const compat = @import("zig_compat");
 const crypto_pkg = @import("crypto");
+const quic = @import("quic");
 const tls_core = @import("tls_core");
 
 const testing = std.testing;
 const provider = crypto_pkg.provider;
+const profile = crypto_pkg.profile;
 const pure_zig = crypto_pkg.pure_zig;
 
 const OpenSslError = error{
@@ -20,10 +22,105 @@ const OpenSslError = error{
     DifferentialMismatch,
 };
 
+const CoverageKind = enum {
+    hkdf_extract,
+    hkdf_expand_label,
+    tls_record_keys,
+    quic_initial,
+    transcript_hash,
+    finished_hmac,
+    aead,
+    key_exchange,
+    signature_verify,
+};
+
+const CoverageClass = enum {
+    positive,
+    negative,
+};
+
+const CoverageCase = struct {
+    kind: CoverageKind,
+    algorithm: ?profile.Algorithm,
+    class: CoverageClass,
+    rationale: []const u8,
+};
+
+const Waiver = struct {
+    kind: CoverageKind,
+    algorithm: profile.Algorithm,
+    class: CoverageClass,
+    reason: []const u8,
+    tracking_issue: []const u8,
+};
+
+const executed_cases = [_]CoverageCase{
+    .{ .kind = .hkdf_extract, .algorithm = .{ .hkdf = .sha256 }, .class = .positive, .rationale = "provider HKDF-Extract compared to OpenSSL EXTRACT_ONLY" },
+    .{ .kind = .hkdf_extract, .algorithm = .{ .hkdf = .sha384 }, .class = .positive, .rationale = "provider HKDF-Extract compared to OpenSSL EXTRACT_ONLY" },
+    .{ .kind = .hkdf_expand_label, .algorithm = .{ .hkdf = .sha256 }, .class = .positive, .rationale = "provider TLS HKDF-Expand-Label compared to OpenSSL EXPAND_ONLY" },
+    .{ .kind = .hkdf_expand_label, .algorithm = .{ .hkdf = .sha384 }, .class = .positive, .rationale = "provider TLS HKDF-Expand-Label compared to OpenSSL EXPAND_ONLY" },
+    .{ .kind = .tls_record_keys, .algorithm = .{ .aead = .aes_128_gcm }, .class = .positive, .rationale = "TrafficKeys.derive compared to OpenSSL key/iv labels" },
+    .{ .kind = .tls_record_keys, .algorithm = .{ .aead = .aes_256_gcm }, .class = .positive, .rationale = "TrafficKeys.derive compared to OpenSSL key/iv labels" },
+    .{ .kind = .tls_record_keys, .algorithm = .{ .aead = .chacha20_poly1305 }, .class = .positive, .rationale = "TrafficKeys.derive compared to OpenSSL key/iv labels" },
+    .{ .kind = .quic_initial, .algorithm = .{ .aead = .aes_128_gcm }, .class = .positive, .rationale = "deriveInitialSecretsV1 compared to OpenSSL HKDF extract/labels" },
+    .{ .kind = .transcript_hash, .algorithm = .{ .hash = .sha256 }, .class = .positive, .rationale = "Transcript update/HRR replacement compared to OpenSSL SHA-256" },
+    .{ .kind = .transcript_hash, .algorithm = .{ .hash = .sha256 }, .class = .negative, .rationale = "mutated transcript hash produces different Finished verify_data" },
+    .{ .kind = .finished_hmac, .algorithm = .{ .hkdf = .sha256 }, .class = .positive, .rationale = "KeySchedule Finished HMAC compared to OpenSSL HMAC" },
+    .{ .kind = .finished_hmac, .algorithm = .{ .hkdf = .sha256 }, .class = .negative, .rationale = "transcript mutation changes Finished verify_data" },
+};
+
+const waivers = [_]Waiver{
+    .{ .kind = .aead, .algorithm = .{ .aead = .aes_128_gcm }, .class = .positive, .reason = "OpenSSL CLI cannot seal/open AEAD with tags portably; add test-only EVP oracle executable.", .tracking_issue = "#377" },
+    .{ .kind = .aead, .algorithm = .{ .aead = .aes_128_gcm }, .class = .negative, .reason = "Invalid-tag parity requires the same test-only EVP oracle executable.", .tracking_issue = "#377" },
+    .{ .kind = .aead, .algorithm = .{ .aead = .aes_256_gcm }, .class = .positive, .reason = "OpenSSL CLI cannot seal/open AEAD with tags portably; add test-only EVP oracle executable.", .tracking_issue = "#377" },
+    .{ .kind = .aead, .algorithm = .{ .aead = .aes_256_gcm }, .class = .negative, .reason = "Invalid-tag parity requires the same test-only EVP oracle executable.", .tracking_issue = "#377" },
+    .{ .kind = .aead, .algorithm = .{ .aead = .chacha20_poly1305 }, .class = .positive, .reason = "OpenSSL CLI cannot seal/open AEAD with tags portably; add test-only EVP oracle executable.", .tracking_issue = "#377" },
+    .{ .kind = .aead, .algorithm = .{ .aead = .chacha20_poly1305 }, .class = .negative, .reason = "Invalid-tag parity requires the same test-only EVP oracle executable.", .tracking_issue = "#377" },
+    .{ .kind = .key_exchange, .algorithm = .{ .group = .x25519 }, .class = .positive, .reason = "X25519 parity needs a stable out-of-process EVP derive oracle, not ad hoc CLI files.", .tracking_issue = "#377" },
+    .{ .kind = .key_exchange, .algorithm = .{ .group = .x25519 }, .class = .negative, .reason = "Low-order/invalid-key parity needs the same EVP derive oracle.", .tracking_issue = "#377" },
+    .{ .kind = .signature_verify, .algorithm = .{ .signature = .ed25519 }, .class = .positive, .reason = "Signature parity needs a stable out-of-process EVP verify oracle.", .tracking_issue = "#377" },
+    .{ .kind = .signature_verify, .algorithm = .{ .signature = .ed25519 }, .class = .negative, .reason = "Invalid-signature parity needs the same EVP verify oracle.", .tracking_issue = "#377" },
+    .{ .kind = .signature_verify, .algorithm = .{ .signature = .ecdsa_secp256r1_sha256 }, .class = .positive, .reason = "ECDSA parity needs a stable out-of-process EVP verify oracle.", .tracking_issue = "#377" },
+    .{ .kind = .signature_verify, .algorithm = .{ .signature = .ecdsa_secp256r1_sha256 }, .class = .negative, .reason = "Invalid-signature parity needs the same EVP verify oracle.", .tracking_issue = "#377" },
+};
+
 fn hexBytes(comptime hex: []const u8) [hex.len / 2]u8 {
     var bytes: [hex.len / 2]u8 = undefined;
     _ = std.fmt.hexToBytes(&bytes, hex) catch unreachable;
     return bytes;
+}
+
+fn algorithmEql(a: profile.Algorithm, b: profile.Algorithm) bool {
+    return switch (a) {
+        .hash => |value| b == .hash and b.hash == value,
+        .hkdf => |value| b == .hkdf and b.hkdf == value,
+        .aead => |value| b == .aead and b.aead == value,
+        .group => |value| b == .group and b.group == value,
+        .signature => |value| b == .signature and b.signature == value,
+        .certificate_helper => |value| b == .certificate_helper and b.certificate_helper == value,
+        .entropy => |value| b == .entropy and b.entropy == value,
+    };
+}
+
+fn hasCoverage(kind: CoverageKind, algorithm: profile.Algorithm, class: CoverageClass) bool {
+    for (executed_cases) |case| {
+        const case_algorithm = case.algorithm orelse continue;
+        if (case.kind == kind and case.class == class and algorithmEql(case_algorithm, algorithm)) return true;
+    }
+    return false;
+}
+
+fn hasWaiver(kind: CoverageKind, algorithm: profile.Algorithm, class: CoverageClass) bool {
+    for (waivers) |waiver| {
+        if (waiver.kind == kind and waiver.class == class and algorithmEql(waiver.algorithm, algorithm)) return true;
+    }
+    return false;
+}
+
+fn expectCoverageOrWaiver(kind: CoverageKind, algorithm: profile.Algorithm, class: CoverageClass) !void {
+    if (hasCoverage(kind, algorithm, class) or hasWaiver(kind, algorithm, class)) return;
+    std.debug.print("missing OpenSSL differential coverage or waiver: kind={s} class={s} algorithm={any}\n", .{ @tagName(kind), @tagName(class), algorithm });
+    return error.MissingDifferentialCoverage;
 }
 
 fn cryptoProvider() provider.CryptoProvider {
@@ -173,6 +270,48 @@ fn runOpenSsl(allocator: std.mem.Allocator, stage: []const u8, argv: []const []c
     return error.OpenSslOracleFailed;
 }
 
+fn opensslHkdfExtract(
+    allocator: std.mem.Allocator,
+    stage: []const u8,
+    hash: provider.Hash,
+    salt: []const u8,
+    ikm: []const u8,
+) ![]u8 {
+    const salt_hex = try hexAlloc(allocator, salt);
+    defer allocator.free(salt_hex);
+    const ikm_hex = try hexAlloc(allocator, ikm);
+    defer allocator.free(ikm_hex);
+    const salt_opt = try std.fmt.allocPrint(allocator, "hexsalt:{s}", .{salt_hex});
+    defer allocator.free(salt_opt);
+    const key_opt = try std.fmt.allocPrint(allocator, "hexkey:{s}", .{ikm_hex});
+    defer allocator.free(key_opt);
+    const key_len = try std.fmt.allocPrint(allocator, "{d}", .{hash.digestLength()});
+    defer allocator.free(key_len);
+    const digest = switch (hash) {
+        .sha256 => "digest:SHA256",
+        .sha384 => "digest:SHA384",
+    };
+    const openssl = try opensslBinary(allocator);
+    defer openssl.deinit(allocator);
+
+    return runOpenSsl(allocator, stage, &.{
+        openssl.path,
+        "kdf",
+        "-keylen",
+        key_len,
+        "-binary",
+        "-kdfopt",
+        digest,
+        "-kdfopt",
+        "mode:EXTRACT_ONLY",
+        "-kdfopt",
+        salt_opt,
+        "-kdfopt",
+        key_opt,
+        "HKDF",
+    });
+}
+
 fn opensslHkdfExpandLabel(
     allocator: std.mem.Allocator,
     stage: []const u8,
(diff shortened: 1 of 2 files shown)


tests/crypto_openssl_diff.zig
```diff
@@ -0,0 +1,326 @@
...
+        "-mac",
+        "HMAC",
+        "-macopt",
+        key_opt,
+        "-binary",
+        path,
+    });
+}
+
+test "OpenSSL HKDF oracle matches TLS 1.3 and QUIC label derivations" {
```
Jaruso: **Blocking:** this test covers only positive `HKDF-Expand-Label` paths. Issue #377 also requires HKDF extract, AEAD seal/open parity and failure behavior, supported key exchange/signature behavior, and invalid tags/signatures/binders/keys/transcript mutations. #348 and #358 only delegate certificate decisions and full TLS-over-TCP interoperability; they do not own these missing primitive-level checks. Please add the missing differential matrix and an explicit waiver registry for genuinely deferred cases.

tests/crypto_openssl_diff.zig
```diff
@@ -0,0 +1,326 @@
...
+    const secret384 = [_]u8{0x42} ** provider.Hash.sha384.digestLength();
+    var pure_sha384: [48]u8 = undefined;
+    try cp.hkdfExpandLabel(.sha384, &secret384, "c hs traffic", "", &pure_sha384);
+    const openssl_sha384 = try opensslHkdfExpandLabel(allocator, "hkdf expand-label sha384 / client handshake traffic", .sha384, &secret384, "c hs traffic", "", pure_sha384.len);
+    defer allocator.free(openssl_sha384);
+    try expectStage("hkdf expand-label sha384 / client handshake traffic", &pure_sha384, openssl_sha384);
+
+    const tls_record_secret = hexBytes("0102030405060708090a0b0c0d0e0f101112131415161718191a1b1c1d1e1f20");
+    var pure_record_key: [16]u8 = undefined;
+    try cp.hkdfExpandLabel(.sha256, &tls_record_secret, "key", "", &pure_record_key);
```
Jaruso: **Blocking false-positive:** calling `cp.hkdfExpandLabel(..., "key", ...)` directly does not test TLS record-key derivation. A regression in `tls_core.record_protection.TrafficKeys.derive`—wrong suite-to-hash mapping, wrong label, wrong key length, or omitted IV derivation—would still pass this test.

Please exercise the production helper and compare its outputs to OpenSSL, e.g.:

```zig
var keys = try tls_core.record_protection.TrafficKeys.derive(
    cp,
    .tls_aes_128_gcm_sha256,
    &tls_record_secret,
);
defer keys.deinit();
try expectStage("tls record aes128 key", keys.key.slice(), openssl_record_key);
try expectStage("tls record aes128 iv", keys.iv.slice(), openssl_record_iv);
```

Repeat this for every supported TLS record suite so SHA-384/AES-256 and ChaCha20 mappings are covered too.

tests/crypto_openssl_diff.zig
```diff
@@ -0,0 +1,326 @@
...
+
+    var pure_record_iv: [provider.aead_nonce_len]u8 = undefined;
+    try cp.hkdfExpandLabel(.sha256, &tls_record_secret, "iv", "", &pure_record_iv);
+    const openssl_record_iv = try opensslHkdfExpandLabel(allocator, "tls record aes128 iv", .sha256, &tls_record_secret, "iv", "", pure_record_iv.len);
+    defer allocator.free(openssl_record_iv);
+    try expectStage("tls record aes128 iv", &pure_record_iv, openssl_record_iv);
+
+    const quic_initial_secret = hexBytes("7db5df06e7a69e432496adedb00851923595221596ae2ae9fb8115c1e9ed0a44");
+    var pure_client_initial: [32]u8 = undefined;
+    try cp.hkdfExpandLabel(.sha256, &quic_initial_secret, "client in", "", &pure_client_initial);
```
Jaruso: **Blocking false-positive:** starting from a hard-coded `quic_initial_secret` bypasses the QUIC initial HKDF-Extract and the production `deriveInitialSecretsV1` / `deriveAes128GcmKeys` code. It also omits the server secret, IV, and header-protection key.

Import the `quic` module into this test target and compare the actual production result:

```zig
const dcid = hexBytes("8394c8f03e515708");
const zig_initial = try quic.tls_adapter.deriveInitialSecretsV1(&dcid);

try expectStage("quic v1 initial secret", &zig_initial.initial_secret, openssl_initial_secret);
try expectStage("quic client initial secret", &zig_initial.client.secret, openssl_client_secret);
try expectStage("quic client key", &zig_initial.client.key, openssl_key);
try expectStage("quic client iv", &zig_initial.client.iv, openssl_iv);
try expectStage("quic client hp", &zig_initial.client.hp, openssl_hp);
```

The build module will also need `crypto_openssl_diff_mod.addImport("quic", quic_mod);`.

tests/crypto_openssl_diff.zig
```diff
@@ -0,0 +1,326 @@
...
+
+    var tmp = testing.tmpDir(.{});
+    defer tmp.cleanup();
+
+    const client_hello_1 = hexBytes("01000003aabbcc");
+    try tmp.dir.writeFile(io, .{ .sub_path = "client_hello_1.bin", .data = &client_hello_1 });
+    const client_hello_path = try tmp.dir.realPathFileAlloc(io, "client_hello_1.bin", allocator);
+    defer allocator.free(client_hello_path);
+
+    const openssl_ch1_hash = try opensslSha256File(allocator, "tls transcript ClientHello1 hash", client_hello_path);
```
Jaruso: **Blocking:** this is not a differential test of Tardigrade's transcript implementation. It runs OpenSSL SHA-256 and compares the result to a hard-coded literal; no pure-Zig transcript code is invoked, so `tls_core.transcript.Transcript` could be completely broken and this test would remain green.

Use the production transcript helper as the expected side:

```zig
var transcript = tls_core.transcript.Transcript{};
transcript.update(&client_hello_1);
const zig_ch1_hash = transcript.peek();
try expectStage("tls transcript ClientHello1 hash", &zig_ch1_hash, openssl_ch1_hash);
```

Then cover the issue's selected transcript behavior, especially HRR rewriting and mutations:

```zig
const ch1_hash = transcript.peek();
transcript.replace(ch1_hash);
transcript.update(&hello_retry_request);
transcript.update(&client_hello_2);
const zig_hrr_hash = transcript.peek();
// Hash message_hash(CH1) || HRR || CH2 independently through OpenSSL and compare.
```

Also mutate one transcript byte and assert the derived Finished verify_data diverges/rejects, instead of feeding an unrelated precomputed `finished_hash` directly into both HMAC paths.

Jaruso (commented):
## Review outcome: changes required

I reviewed both changed files, the full acceptance criteria in #377, its dependencies (#373, #348, and #358), the existing crypto capability/profile matrix, and the successful CI run.

The OpenSSL binary resolver and general/appliance build gating look reasonable, and CI is green. However, this PR does **not** complete the referenced issue and currently creates false confidence around several claims in the PR summary.

### Blocking acceptance-criteria gaps

The differential harness currently covers only positive `HKDF-Expand-Label` cases, one SHA-256 digest checked against a literal, and one Finished HMAC. Issue #377 additionally requires:

- HKDF **extract** parity;
- AEAD seal/open output parity and failure behavior for supported suites (`AES-128-GCM`, `AES-256-GCM`, `ChaCha20-Poly1305`);
- supported key-exchange parity (`X25519`);
- supported signature parity (`Ed25519` and currently advertised ECDSA verification);
- negative cases for invalid tags, signatures, binders, keys, and transcript mutations;
- written allowlists for intentionally unsupported/policy-different cases.

The delegation to #348 and #358 is valid for certificate-chain decisions and full TLS-over-TCP interop, but those issues do not own the missing primitive/key-schedule negative coverage above.

### Expected implementation shape

At minimum, add an OpenSSL extract helper and compare the real provider call:

```zig
fn opensslHkdfExtract(
    allocator: std.mem.Allocator,
    stage: []const u8,
    hash: provider.Hash,
    salt: []const u8,
    ikm: []const u8,
) ![]u8 {
    const salt_hex = try hexAlloc(allocator, salt);
    defer allocator.free(salt_hex);
    const ikm_hex = try hexAlloc(allocator, ikm);
    defer allocator.free(ikm_hex);

    const salt_opt = try std.fmt.allocPrint(allocator, "hexsalt:{s}", .{salt_hex});
    defer allocator.free(salt_opt);
    const key_opt = try std.fmt.allocPrint(allocator, "hexkey:{s}", .{ikm_hex});
    defer allocator.free(key_opt);
    const key_len = try std.fmt.allocPrint(allocator, "{d}", .{hash.digestLength()});
    defer allocator.free(key_len);

    const digest = switch (hash) {
        .sha256 => "digest:SHA256",
        .sha384 => "digest:SHA384",
    };
    const openssl = try opensslBinary(allocator);
    defer openssl.deinit(allocator);

    return runOpenSsl(allocator, stage, &.{
        openssl.path, "kdf", "-keylen", key_len, "-binary",
        "-kdfopt", digest,
        "-kdfopt", "mode:EXTRACT_ONLY",
        "-kdfopt", salt_opt,
        "-kdfopt", key_opt,
        "HKDF",
    });
}
```

```zig
var zig_prk: [provider.Hash.sha256.digestLength()]u8 = undefined;
try cp.hkdfExtract(.sha256, &salt, &ikm, &zig_prk);
const openssl_prk = try opensslHkdfExtract(
    allocator,
    "hkdf extract sha256 / RFC 5869 A.1",
    .sha256,
    &salt,
    &ikm,
);
defer allocator.free(openssl_prk);
try expectStage("hkdf extract sha256 / RFC 5869 A.1", &zig_prk, openssl_prk);
```

For AEAD, X25519, and signature operations, a small **test-only out-of-process EVP oracle executable** is an acceptable implementation: it may link OpenSSL, but the Zig test binary must continue to invoke it as a child process. The oracle should accept deterministic fixture inputs, return output/status only, and support at least:

```text
hkdf-extract
hkdf-expand
seal / open
x25519-derive
verify-ed25519
verify-ecdsa-p256-sha256
```

Then add positive and negative table-driven cases. For example, an invalid-tag case should require both implementations to reject and should also verify the pure-Zig plaintext buffer was cleared:

```zig
var plaintext = [_]u8{0xA5} ** ciphertext.len;
var bad_tag = valid_tag;
bad_tag[0] ^= 1;

try testing.expectError(
    error.AuthenticationFailed,
    cp.aeadOpen(.aes_128_gcm, &key, &nonce, ad, ciphertext, &bad_tag, &plaintext),
);
try testing.expectEqualSlices(u8, &([_]u8{0} ** ciphertext.len), &plaintext);
try testing.expect(!(try opensslOracleOpenAccepted(
    allocator,
    .aes_128_gcm,
    &key,
    &nonce,
    ad,
    ciphertext,
    &bad_tag,
)));
```

Finally, add a checked-in differential case registry/waiver table so omitted capabilities cannot disappear silently. Each supported `profile.Row` should have positive coverage and, where applicable, negative coverage; each unsupported/deferred operation should have a written rationale and tracking issue rather than a blanket omission.

The inline comments identify two additional false-positive paths in the current tests: production TLS/QUIC derivation helpers are bypassed, and the transcript check never invokes the Zig transcript implementation.

tests/crypto_openssl_diff.zig
Jaruso: HKDF-Extract coverage and a registry were added, which is good progress, but this thread is not resolved yet. The new entries waive AEAD, X25519, Ed25519, and ECDSA—the core operations explicitly required by #377—and every waiver points back to #377 itself rather than a separate deferred dependency. Those are implementation/tooling gaps, not the “expected policy differences” the AC permits allowlisting. The registry also omits binder failures and signature *signing* entirely. Please either implement the out-of-process EVP oracle needed for these cases, or add explicit dependency waivers (for example binder work should point to #362) while keeping #377 open and updating the PR scope so it does not imply only #348/#358 remain.

tests/crypto_openssl_diff.zig
Jaruso: The production helper is now exercised for all three suites, but the original false-positive is still present because the OpenSSL side takes `hash` and key length from the same production `suiteProfile(suite)` being tested. If AES-256 were accidentally mapped to SHA-256/AES-128, both sides would use the wrong mapping and still match. Please make the oracle parameters independent and explicit, e.g. `.tls_aes_128_gcm_sha256 => { .hash = .sha256, .key_len = 16 }`, `.tls_aes_256_gcm_sha384 => { .hash = .sha384, .key_len = 32 }`, and ChaCha => SHA-256/32, then assert the production output against those constants.

tests/crypto_openssl_diff.zig
Jaruso: The original bypass is addressed: the test now calls `deriveInitialSecretsV1`, imports `quic`, and checks client/server secrets plus key/IV/HP outputs. I’m resolving this conversation. I found a separate independence issue in the new implementation—the OpenSSL side reuses the production salt, output-length constants, and Zig-derived intermediates—which I’m tracking on the current lines.

tests/crypto_openssl_diff.zig
Jaruso: The positive transcript path and HRR replacement are now genuinely differential and use `tls_core.transcript.Transcript`, so that portion is fixed. The negative portion is still incomplete: it mutates the already-computed digest (`mutated_hash[0]`) rather than a handshake byte, never rebuilds the production transcript, and never compares the mutated path to OpenSSL. Please mutate `client_hello_2` (or HRR), rebuild both `Transcript` and the synthetic OpenSSL input, compare the mutated digest on both sides, derive mutated Finished data through both sides, and assert it differs from the unmutated value.

tests/crypto_openssl_diff.zig
```diff
@@ -0,0 +1,591 @@
...
+    kind: CoverageKind,
+    algorithm: profile.Algorithm,
+    class: CoverageClass,
+    reason: []const u8,
+    tracking_issue: []const u8,
+};
+
+const executed_cases = [_]CoverageCase{
+    .{ .kind = .hkdf_extract, .algorithm = .{ .hkdf = .sha256 }, .class = .positive, .rationale = "provider HKDF-Extract compared to OpenSSL EXTRACT_ONLY" },
+    .{ .kind = .hkdf_extract, .algorithm = .{ .hkdf = .sha384 }, .class = .positive, .rationale = "provider HKDF-Extract compared to OpenSSL EXTRACT_ONLY" },
```
Jaruso: **Blocking coverage false-positive:** `executed_cases` is a static declaration, not a record of tests that actually ran. Deleting an entire differential test—or accidentally removing one `expectStage`—would leave this registry green because nothing connects these entries to execution.

Please make the registry executable. One straightforward shape is to give each case a runner and iterate the same table from one test:

```zig
const DifferentialCase = struct {
    kind: CoverageKind,
    algorithm: profile.Algorithm,
    class: CoverageClass,
    rationale: []const u8,
    run: *const fn (std.mem.Allocator) anyerror!void,
};

const cases = [_]DifferentialCase{
    .{
        .kind = .hkdf_extract,
        .algorithm = .{ .hkdf = .sha256 },
        .class = .positive,
        .rationale = "RFC 5869 A.1",
        .run = runHkdfExtractSha256,
    },
    // ...
};

test "OpenSSL differential cases" {
    for (cases) |case| {
        errdefer std.debug.print("failed case: {s}/{s}\n", .{ @tagName(case.kind), case.rationale });
        try case.run(testing.allocator);
    }
}
```

An execution log like the one in `tests/crypto_vectors.zig` is also fine. The key requirement is that CI cannot report coverage for a case whose oracle comparison did not execute.

tests/crypto_openssl_diff.zig
```diff
@@ -0,0 +1,591 @@
...
+    .{ .kind = .transcript_hash, .algorithm = .{ .hash = .sha256 }, .class = .negative, .rationale = "mutated transcript hash produces different Finished verify_data" },
+    .{ .kind = .finished_hmac, .algorithm = .{ .hkdf = .sha256 }, .class = .positive, .rationale = "KeySchedule Finished HMAC compared to OpenSSL HMAC" },
+    .{ .kind = .finished_hmac, .algorithm = .{ .hkdf = .sha256 }, .class = .negative, .rationale = "transcript mutation changes Finished verify_data" },
+};
+
+const waivers = [_]Waiver{
+    .{ .kind = .aead, .algorithm = .{ .aead = .aes_128_gcm }, .class = .positive, .reason = "OpenSSL CLI cannot seal/open AEAD with tags portably; add test-only EVP oracle executable.", .tracking_issue = "#377" },
+    .{ .kind = .aead, .algorithm = .{ .aead = .aes_128_gcm }, .class = .negative, .reason = "Invalid-tag parity requires the same test-only EVP oracle executable.", .tracking_issue = "#377" },
+    .{ .kind = .aead, .algorithm = .{ .aead = .aes_256_gcm }, .class = .positive, .reason = "OpenSSL CLI cannot seal/open AEAD with tags portably; add test-only EVP oracle executable.", .tracking_issue = "#377" },
```
Jaruso: **Blocking AC gap:** these are waivers for missing required implementations, not allowlisted *policy differences*. They all point back to #377 itself, so they do not identify an external dependency or follow-up that can ever discharge the waiver. The issue explicitly requires AEAD seal/open, supported key exchange, signature verification/signing, and failures for tags/signatures/binders/keys.

The registry also omits:

- TLS handshake/application traffic-secret derivation through `KeySchedule.init` and `applicationSecrets`;
- signature **signing** behavior (only `signature_verify` exists);
- binder failures entirely, even though #362 is the actual implementation dependency.

Please either add the out-of-process EVP oracle now, or model genuine dependency waivers with distinct tracking issues while keeping #377 open. For example:

```zig
const CoverageKind = enum {
    // existing kinds...
    signature_sign,
    psk_binder,
    tls_key_schedule,
};

const waivers = [_]Waiver{
    .{
        .kind = .psk_binder,
        .algorithm = .{ .hkdf = .sha256 },
        .class = .negative,
        .reason = "Pure-Zig binder generation/verification is not available yet.",
        .tracking_issue = "#362",
    },
};
```

For AEAD/X25519/signatures, #377 is the story that owns the oracle, so self-waiving those operations does not satisfy its scope. A small standalone test executable linked to libcrypto and invoked out of process would preserve the dependency boundary while supplying stable EVP seal/open, derive, sign, and verify commands.

tests/crypto_openssl_diff.zig
```diff
@@ -0,0 +1,591 @@
...
+    const chacha_secret = [_]u8{0x33} ** provider.Hash.sha256.digestLength();
+    try expectTlsTrafficKeys(allocator, "tls record chacha20 key", cp, .tls_chacha20_poly1305_sha256, &chacha_secret);
+}
+
+test "OpenSSL HKDF oracle matches QUIC production initial derivation" {
+    const allocator = testing.allocator;
+    const dcid = hexBytes("8394c8f03e515708");
+    const zig_initial = try quic.tls_adapter.deriveInitialSecretsV1(&dcid);
+
+    const openssl_initial = try opensslHkdfExtract(allocator, "quic v1 initial secret", .sha256, &quic.tls_adapter.initial_salt_v1, &dcid);
```
Jaruso: **Blocking differential-independence issue:** the production helper is now called, but the OpenSSL side reuses the production salt and production output-size constants. A regression in `initial_salt_v1`, `traffic_secret_len`, `aead_key_len`, `aead_iv_len`, or `header_protection_key_len` can therefore be mirrored by the oracle and remain green.

Use fixed RFC 9001 inputs and OpenSSL-derived intermediates on the oracle side:

```zig
const rfc9001_initial_salt = hexBytes(
    "38762cf7f55934b34d179ae6a4c80cadccbb7f0a",
);
const openssl_initial = try opensslHkdfExtract(
    allocator,
    "quic v1 initial secret",
    .sha256,
    &rfc9001_initial_salt,
    &dcid,
);
try expectStage("quic v1 initial secret", &zig_initial.initial_secret, openssl_initial);

const openssl_client_secret = try opensslHkdfExpandLabel(
    allocator,
    "quic client initial secret",
    .sha256,
    openssl_initial,
    "client in",
    "",
    32,
);
const openssl_client_key = try opensslHkdfExpandLabel(
    allocator,
    "quic client initial key",
    .sha256,
    openssl_client_secret,
    "quic key",
    "",
    16,
);
// IV = 12, HP = 16; repeat independently for server.
```

This makes the test fail when the production protocol constants—not merely the crypto primitive—are wrong.

tests/crypto_openssl_diff.zig
```diff
@@ -0,0 +1,591 @@
...
+    const synthetic_hrr_path = try tmp.dir.realPathFileAlloc(io, "synthetic_hrr.bin", allocator);
+    defer allocator.free(synthetic_hrr_path);
+    try expectOpenSslSha256File(allocator, "tls transcript HRR hash", &zig_hrr_hash, synthetic_hrr_path);
+
+    const traffic_secret = hexBytes("b67b7d690cc16c4e75e54213cb2d37b4e9c912bcded9105d42befd59d391ad38");
+    try tmp.dir.writeFile(io, .{ .sub_path = "finished_hash.bin", .data = &zig_hrr_hash });
+    const finished_hash_path = try tmp.dir.realPathFileAlloc(io, "finished_hash.bin", allocator);
+    defer allocator.free(finished_hash_path);
+
+    const finished_key = KeySchedule.finishedKey(traffic_secret);
```
Jaruso: **Blocking Finished false-positive:** OpenSSL receives `KeySchedule.finishedKey(traffic_secret)` from the Zig implementation, while `KeySchedule.verifyData` internally calls the same `finishedKey` function again. If the production label changes from `"finished"` to an incorrect value, both sides use the same wrong key and this test still passes.

Derive the Finished key independently through OpenSSL, compare it first, and then use the OpenSSL-derived key for the HMAC:

```zig
const zig_finished_key = KeySchedule.finishedKey(traffic_secret);
const openssl_finished_key = try opensslHkdfExpandLabel(
    allocator,
    "tls13 server Finished key",
    .sha256,
    &traffic_secret,
    "finished",
    "",
    32,
);
defer allocator.free(openssl_finished_key);
try expectStage(
    "tls13 server Finished key",
    &zig_finished_key,
    openssl_finished_key,
);

const openssl_verify_data = try opensslHmacSha256File(
    allocator,
    "tls13 server Finished verify_data",
    openssl_finished_key,
    finished_hash_path,
);
try expectStage(
    "tls13 server Finished verify_data",
    &KeySchedule.verifyData(traffic_secret, zig_hrr_hash),
    openssl_verify_data,
);
```

The same independence rule should be used when adding `KeySchedule.init` and application-traffic-secret comparisons.

Jaruso (commented):
Second-pass review at `aafb7549a0456fc37b4286798efc0ac187097d05`: **changes are still required**. CI is green and the update materially improves HKDF, TLS record, QUIC, and transcript coverage. I resolved the original QUIC-production-path conversation because that specific bypass was corrected. The other three original threads remain partially open, and the current code adds four additional blocking false-positive/acceptance-criteria gaps documented inline below.

The main theme is oracle independence: OpenSSL must receive protocol constants and intermediate values that are independent of the production helper under test. Otherwise both sides can make the same configuration mistake and the differential test remains green.

tests/crypto_openssl_diff.zig
```diff
@@ -0,0 +1,814 @@
...
+    .{ .kind = .signature_verify, .algorithm = .{ .signature = .ed25519 }, .class = .positive, .reason = "Raw-key Ed25519 verify parity requires the dedicated EVP oracle follow-up.", .tracking_issue = "#431" },
+    .{ .kind = .signature_verify, .algorithm = .{ .signature = .ed25519 }, .class = .negative, .reason = "Invalid-signature Ed25519 parity requires the same raw-key EVP oracle.", .tracking_issue = "#431" },
+    .{ .kind = .signature_sign, .algorithm = .{ .signature = .ecdsa_secp256r1_sha256 }, .class = .positive, .reason = "The provider seam exposes ECDSA-P256 verification only; signing support needs a separate provider story before differential coverage can run.", .tracking_issue = "#343" },
+    .{ .kind = .signature_verify, .algorithm = .{ .signature = .ecdsa_secp256r1_sha256 }, .class = .positive, .reason = "SEC1-key ECDSA verification parity requires the dedicated EVP oracle follow-up.", .tracking_issue = "#431" },
+    .{ .kind = .signature_verify, .algorithm = .{ .signature = .ecdsa_secp256r1_sha256 }, .class = .negative, .reason = "Invalid-signature ECDSA parity requires the same SEC1-key EVP oracle.", .tracking_issue = "#431" },
```
Jaruso, line 90: **Blocking waiver cannot be discharged:** `#343` is already closed and its scope was certificate **verification**; it never tracked ECDSA signing support. This entry therefore satisfies the registry syntactically while pointing to no open work that can replace the waiver.

ECDSA signing is not currently a supported provider operation, so the cleanest correction is to remove this waiver and the corresponding required lookup:

```zig
// Remove this waiver:
// .{ .kind = .signature_sign,
//    .algorithm = .{ .signature = .ecdsa_secp256r1_sha256 }, ... }

// And remove:
try expectCoverageOrWaiver(
    .signature_sign,
    .{ .signature = .ecdsa_secp256r1_sha256 },
    .positive,
);
```

That matches #377's requirement to compare signing behavior **where supported**. Alternatively, create a dedicated open provider story for ECDSA `SigningKey` support and point the waiver to that issue. Do not use closed verification issue #343 as the signing dependency.

Jaruso (commented):
One blocking tracking defect remains; all other previously raised findings are addressed and their conversations have been resolved.

PR #439, opened 2026-07-16
test(quic): bound deterministic H3 driver scenarios

Jaruso:
## Summary

- harden the existing native QUIC/H3 simulated network with aggregate queued-datagram and queued-byte limits
- bound driver iterations, simulated elapsed time, packets produced per iteration, and timer work with typed failures
- derive deterministic TLS test entropy from a reported scenario seed and retain failure snapshots with both endpoint states
- add focused regression coverage for every bound plus seed reproducibility
- expose the fast suite as `zig build test-quic-h3-driver` while retaining it in `test-quic` and the default test aggregate

## Investigation and reuse

This slice reuses the production `quic.connection.Connection` driver and its packet/frame codec, TLS adapter/backend, ACK and recovery timers, congestion state, streams and flow control, plus the production `http3.conn`/QPACK/session entrypoints. The existing in-process smoke and E2E harnesses remain the integration seams; no simplified protocol implementation or real-socket dependency was added.

The live tree already includes the baseline handshake and H3 exchange, deterministic loss/reordering/duplication/delay, real-UDP smoke, and optional external interop tooling. This PR therefore concentrates on the missing resource/progress bounds and reproducible failure diagnostics.

## Validation

- `zig build test-quic-h3-driver --summary all --error-style verbose` — 21/21 passed
- `zig build test-quic --summary all --error-style verbose` — 410/410 passed
- `zig build test --summary all --error-style verbose` — 1376/1377 passed, 1 existing skip
- `git diff --check`

## Scope

This is a merge-sized deterministic driver-hardening slice of #247. It does not claim to complete the remaining scheduled soak, benchmark, wider malformed corpus, or full interop-runner acceptance criteria, so #247 should remain open after this PR merges.

Part of #247.

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -366,6 +366,8 @@ pub fn build(b: *std.Build) void {
     quic_h3_e2e_mod.addImport("stream_transport", stream_transport_mod);
     const quic_h3_e2e_tests = b.addTest(.{ .root_module = quic_h3_e2e_mod });
     const run_quic_h3_e2e_tests = b.addRunArtifact(quic_h3_e2e_tests);
+    const quic_h3_driver_step = b.step("test-quic-h3-driver", "Run deterministic native QUIC/H3 driver scenarios");
+    quic_h3_driver_step.dependOn(&run_quic_h3_e2e_tests.step);
     quic_step.dependOn(&run_quic_h3_e2e_tests.step);
     test_step.dependOn(&run_quic_h3_e2e_tests.step);
 
--- a/tests/quic_h3_e2e.zig
+++ b/tests/quic_h3_e2e.zig
@@ -30,6 +30,32 @@ const testing = std.testing;
 
 const H3 = http3.conn.Conn(Connection);
 
+const HarnessFailure = error{
+    NoProgress,
+    IterationLimit,
+    SimulatedTimeLimit,
+    QueueDatagramLimit,
+    QueueByteLimit,
+    PacketProductionLimit,
+    TimerEventLimit,
+    UnexpectedConnectionClosure,
+    DatagramTooLarge,
+};
+
+const Direction = enum { to_server, to_client };
+
+/// Hard resource and progress limits for every deterministic scenario. These
+/// are deliberately part of the harness configuration rather than implicit
+/// test-runner timeouts, so adversarial cases fail at a reproducible boundary.
+const Limits = struct {
+    max_iterations: usize = 16_384,
+    max_simulated_time_us: u64 = 90_000_000,
+    max_queued_datagrams: usize = 128,
+    max_queued_bytes: usize = 256 * 1024,
+    max_packets_per_iteration: usize = 64,
+    max_timer_events: usize = 2_048,
+};
+
 /// Per-direction delivery rules, indexed by datagram sequence number in that
 /// direction (0-based, counted at the sender).
 const Rules = struct {
@@ -59,8 +85,10 @@ const Rules = struct {
 const InFlight = struct {
     deliver_at_us: u64,
     len: usize,
-    bytes: [2048]u8,
+    bytes: [max_test_datagram_size]u8,
 };
+
+const max_test_datagram_size = 2048;
 
 const Pipe = struct {
     rules: Rules,
@@ -71,7 +99,20 @@ const Pipe = struct {
         self.queue.deinit(allocator);
     }
 
-    fn push(self: *Pipe, allocator: std.mem.Allocator, bytes: []const u8, now_us: u64) !void {
+    fn append(self: *Pipe, allocator: std.mem.Allocator, bytes: []const u8, deliver_at_us: u64) !void {
+        if (bytes.len > max_test_datagram_size) return error.DatagramTooLarge;
+        var entry = InFlight{ .deliver_at_us = deliver_at_us, .len = bytes.len, .bytes = undefined };
+        @memcpy(entry.bytes[0..bytes.len], bytes);
+        try self.queue.append(allocator, entry);
+    }
+
+    fn queuedBytes(self: *const Pipe) usize {
+        var total: usize = 0;
+        for (self.queue.items) |entry| total += entry.len;
+        return total;
+    }
+
+    fn schedule(self: *Pipe, allocator: std.mem.Allocator, bytes: []const u8, now_us: u64) !void {
         const index = self.sent;
         self.sent += 1;
         if (self.rules.shouldDrop(index)) return;
@@ -80,12 +121,9 @@ const Pipe = struct {
             // Delivered after its successor: push it further out.
             deliver_at += 2 * self.rules.latency_us;
         }
-        var entry = InFlight{ .deliver_at_us = deliver_at, .len = bytes.len, .bytes = undefined };
-        @memcpy(entry.bytes[0..bytes.len], bytes);
-        try self.queue.append(allocator, entry);
+        try self.append(allocator, bytes, deliver_at);
         if (self.rules.shouldDuplicate(index)) {
-            entry.deliver_at_us += 1;
-            try self.queue.append(allocator, entry);
+            try self.append(allocator, bytes, deliver_at + 1);
         }
     }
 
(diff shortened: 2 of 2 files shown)


Jaruso (commented):
## Review result: changes required

I verified this against the referenced #247 acceptance criteria and the PR's explicitly narrower scope. I am **not** asking this PR to complete the remaining interop runner, fuzz corpus, soak, benchmark, or CI-documentation work. The new standalone build step is correctly wired into `test-quic` and the default `test` aggregate, and the current CI run is green.

However, four issues prevent this slice from fully delivering its own stated guarantees.

### 1. Blocking — ordinary scenario failures do not retain/report the seed and endpoint states

`Sim.fail` is only reached for the new `HarnessFailure` paths. Errors from `H3.start`, `sendRequest`, `pump`, `pollRequest`, `pollResponse`, stream operations, allocator failures, and every `testing.expect*` return directly. Those are the protocol regressions for which deterministic seed diagnostics are most valuable, yet they produce no `FailureSnapshot` and no seed/state log.

For example, a response-body mismatch or QPACK error in `runH3Exchange` returns directly without calling `Sim.fail`.

Please make snapshot recording accept `anyerror` and install an outer error capture for every scenario. One low-impact pattern is:

```zig
const FailureSnapshot = struct {
    reason: anyerror,
    scenario: []const u8,
    seed: u64,
    now_us: u64,
    iterations: usize,
    timer_events: usize,
    queued_datagrams: usize,
    queued_bytes: usize,
    client_state: connection.State,
    server_state: connection.State,
};

fn recordFailure(self: *Sim, reason: anyerror) void {
    if (self.last_failure != null) return;
    self.last_failure = .{
        .reason = reason,
        .scenario = self.scenario,
        .seed = self.seed,
        .now_us = self.now_us,
        .iterations = self.iterations,
        .timer_events = self.timer_events,
        .queued_datagrams = self.network.queuedDatagrams(),
        .queued_bytes = self.network.queuedBytes(),
        .client_state = self.client.state(),
        .server_state = self.server.state(),
    };
    if (self.log_failures) self.logFailure();
}

fn fail(self: *Sim, reason: anyerror) anyerror {
    self.recordFailure(reason);
    return reason;
}
```

Then, immediately after each `Sim.init`, capture every escaping test error:

```zig
var sim = try Sim.init(testing.allocator, config);
defer sim.deinit();
errdefer |err| sim.recordFailure(err);
```

Because `recordFailure` preserves the first error, errors already routed through `fail` will not be overwritten. Please add a regression proving that a non-`HarnessFailure` error also retains the scenario, seed, and both states.

### 2. Blocking — existing scenarios bypass the simulated-time and timer-work limits

The lossless close/drain test directly executes:

```zig
sim.now_us += 10_000_000;
sim.client.onTimeout(sim.now_us);
sim.server.onTimeout(sim.now_us);
```

The unexpected-closure test later directly mutates the clock and invokes `client.onTimeout` as well. These calls do not enforce `max_simulated_time_us`, do not enforce/increment `max_timer_events`, and leave failure snapshots under-reporting timer work. That contradicts the new comment that the hard limits apply to every deterministic scenario.

Please centralize explicit clock advancement and timeout execution through the bounded harness API, for example:

```zig
fn advanceAndFireTimers(self: *Sim, delta_us: u64) !void {
    const target = std.math.add(u64, self.now_us, delta_us) catch
        return self.fail(error.SimulatedTimeLimit);
    if (target - self.started_at_us > self.limits.max_simulated_time_us)
        return self.fail(error.SimulatedTimeLimit);

    self.now_us = target;
    try self.fireTimers();
}
```

Then replace both direct-timeout sites with:

```zig
try sim.advanceAndFireTimers(10_000_000);
```

This also avoids unchecked clock-addition overflow.

### 3. Blocking test gap — the seed regression would pass if the seed were ignored

The new test only compares two runs using the same seed. The pre-PR constant TLS entropy would also produce identical Initial datagrams, so this test does not protect the requirement that the reported seed actually controls the wire image.

Please add a third run using a different seed and assert that its protected Initial differs:

```zig
var different = try Sim.init(testing.allocator, .{
    .scenario = "seed-different",
    .seed = seed + 1,
});
defer different.deinit();
try testing.expect(try different.step());

const different_initial = different.network.to_server.queue.items[0];
try testing.expect(
    first_initial.len != different_initial.len or
        !std.mem.eql(
            u8,
            first_initial.bytes[0..first_initial.len],
            different_initial.bytes[0..different_initial.len],
        ),
);
```

That makes the test fail if seed derivation is removed or accidentally disconnected while preserving the same-seed reproducibility assertion.

### 4. Blocking test gap — aggregate queue limits are not tested across directions

Both queue tests exercise only client-to-server queuing. A future implementation with independent per-pipe limits would still pass, despite violating the PR's advertised aggregate `to_server + to_client` budget.

Please add focused network-level regressions. For the datagram count:

```zig
var network = TestNetwork{
    .to_server = .{ .rules = .{} },
    .to_client = .{ .rules = .{} },
    .limits = .{ .max_queued_datagrams = 1 },
};
defer network.deinit(testing.allocator);

const byte = [_]u8{0xaa};
try network.enqueue(testing.allocator, .to_server, &byte, 0);
try testing.expectError(
    error.QueueDatagramLimit,
    network.enqueue(testing.allocator, .to_client, &byte, 0),
);
```

Use a separate network instance with a one-byte aggregate byte budget to prove `QueueByteLimit` across directions as well; otherwise the datagram limit will mask the byte check.

After addressing these, please rerun:

```text
zig build test-quic-h3-driver --summary all --error-style verbose
zig build test-quic --summary all --error-style verbose
zig build test --summary all --error-style verbose
zig fmt --check build.zig src/ tests/
git diff --check
```

Jaruso (commented):
## Second review result: two follow-ups still required

I re-reviewed the full updated PR at `a0dad18e4d6067d7bf32c71942fc65967c13a88e`, not only the follow-up commit. The original four findings are substantially addressed:

- `FailureSnapshot.reason` and `recordFailure` now accept ordinary errors;
- every single-`Sim` scenario installs an outer error capture;
- explicit clock advancement and timer execution now use the bounded helper;
- the different-seed assertion is present;
- cross-direction aggregate datagram and byte tests are present;
- the updated CI workflow is green.

Two issues remain in the failure-diagnostic regression coverage.

### 1. Blocking — the multi-simulation seed test falsely attributes one error to all three scenarios

The seed test keeps `first`, `second`, and `different` alive in the same error-returning scope and registers an `errdefer` for each. If any later `try` or assertion fails, **all three** errdefers execute with the same escaping error.

For example, if `first.step()` fails, the test will record and log that same error against:

- `seed-different`, although that simulation was never stepped;
- `seed-second`, although that simulation was never stepped;
- `seed-first`, where the error actually occurred.

Because errdefers run in reverse registration order, the misleading `seed-different` diagnostic may be printed first. This conflicts directly with the purpose of reporting a trustworthy scenario/seed/state tuple.

Please isolate each simulation's capture scope. A helper is the cleanest approach:

```zig
fn initialWireImage(scenario: []const u8, seed: u64) !InFlight {
    var sim = try Sim.init(testing.allocator, .{
        .scenario = scenario,
        .seed = seed,
    });
    defer sim.deinit();
    errdefer |err| sim.recordFailure(err);

    try testing.expect(try sim.step());
    try testing.expectEqual(
        @as(usize, 1),
        sim.network.to_server.queue.items.len,
    );
    return sim.network.to_server.queue.items[0];
}

test "e2e harness: scenario seed reproduces the protected wire image" {
    const seed = 0x247_d37e;
    const first_initial = try initialWireImage("seed-first", seed);
    const second_initial = try initialWireImage("seed-second", seed);
    const different_initial = try initialWireImage("seed-different", seed + 1);

    try testing.expectEqual(first_initial.len, second_initial.len);
    try testing.expectEqualSlices(
        u8,
        first_initial.bytes[0..first_initial.len],
        second_initial.bytes[0..second_initial.len],
    );
    try testing.expect(
        first_initial.len != different_initial.len or
            !std.mem.eql(
                u8,
                first_initial.bytes[0..first_initial.len],
                different_initial.bytes[0..different_initial.len],
            ),
    );
}
```

Each helper invocation then owns exactly one scenario and can only attach an escaping error to that scenario.

### 2. Blocking test gap — the ordinary-error regression calls `recordFailure` directly and never exercises automatic error capture

The new test installs:

```zig
errdefer |err| sim.recordFailure(err);
```

but then directly invokes:

```zig
sim.recordFailure(error.QpackRegression);
```

The test returns successfully, so its errdefer never runs. Consequently, the test would still pass if the automatic-capture errdefers were removed or accidentally placed after cleanup. It proves `recordFailure(anyerror)` works, but not the requested behavior that an **escaping ordinary scenario error** is automatically captured.

Please make a nested helper actually return an ordinary error through an errdefer, then catch it from the test:

```zig
fn returnCapturedOrdinaryError(sim: *Sim) !void {
    errdefer |err| sim.recordFailure(err);
    return error.QpackRegression;
}

test "e2e harness: ordinary scenario errors retain deterministic context" {
    const scenario = "ordinary-protocol-error";
    const seed = 0x247_fa11;
    var sim = try Sim.init(testing.allocator, .{
        .scenario = scenario,
        .seed = seed,
        .log_failures = false,
    });
    defer sim.deinit();

    const client_state = sim.client.state();
    const server_state = sim.server.state();
    try testing.expectError(
        error.QpackRegression,
        returnCapturedOrdinaryError(sim),
    );

    // Preserve-first behavior remains worth checking.
    sim.recordFailure(error.LaterFailureMustNotOverwrite);
    try expectFailureSnapshot(sim, error.QpackRegression, scenario, seed);
    try testing.expectEqual(client_state, sim.last_failure.?.client_state);
    try testing.expectEqual(server_state, sim.last_failure.?.server_state);
}
```

This fails if automatic capture is disconnected and still verifies that a later error cannot overwrite the original snapshot.

### Non-blocking documentation note

The PR description still reports the pre-fix test totals (`18/18`, `407/407`, and `1373/1374`) even though three tests were added in the follow-up commit. Please update the validation section with the actual post-change counts from the final rerun.

After the two focused test fixes, rerun the same validation commands. I did not find an additional protocol-driver, queue-accounting, time-bound, or build-graph defect in this pass.

Jaruso (commented):
## Final review: ready to merge

I re-reviewed the complete current PR at `36cedc1dcc4b307bfb4323d55a73c9a6c3ecba6f`, not only the latest patch.

The two remaining findings from my previous pass are now correctly resolved:

- `initialWireImage` gives each seeded simulation its own scope and `errdefer`, so a failure is attributed only to the scenario that actually failed. Returning `InFlight` by value also keeps the compared datagram valid after the temporary simulation is deinitialized.
- `returnCapturedOrdinaryError` now produces a real escaping non-harness error and proves that `errdefer` records it. The subsequent failure confirms that the first snapshot is retained rather than overwritten.

I also rechecked the earlier changes across the full harness:

- ordinary protocol/assertion errors retain the scenario, seed, queue counters, timer/iteration counters, and both endpoint states;
- explicit clock advances and timeout execution use the bounded helper;
- iteration, simulated-time, packet-production, timer-work, aggregate datagram, and aggregate byte limits have focused coverage;
- the seed controls the protected Initial while identical seeds reproduce it;
- aggregate queue limits are tested across both directions;
- `test-quic-h3-driver` remains wired into the same artifact used by `test-quic` and the default test aggregate;
- the PR validation counts were updated to the current 21/21, 410/410, and 1376/1377 results;
- the latest GitHub Actions run completed successfully across the test, format, lint, build, security, dependency, packaging, TLS-audit, and performance-smoke jobs.

I found no remaining blocking or non-blocking issue in the code changed by this PR. This deterministic driver-hardening slice is ready to merge. As the PR states, issue #247 should remain open for its intentionally deferred soak, benchmark, wider malformed-corpus, and full interop-runner work.

PR #440, opened 2026-07-16
pki: add deterministic X.509 candidate-path construction engine (#344)

### LINKED ISSUE #344: pki: [324-F] build deterministic X.509 path-construction engine
## Parent epic

- #324
- Research story: **324-F** | Priority **P1** | Complexity **XL** | Owner **PKI**

## Objective

Construct candidate certificate paths from a leaf, peer-supplied intermediates, and configured trust anchors without mixing path discovery with RFC 5280 policy validation.

## Scope

- Index candidates by normalized issuer/subject and AKI/SKI hints.
- Build bounded, cycle-safe candidate paths.
- Handle cross-signed roots, duplicate certificates, missing intermediates, and multiple possible issuers deterministically.
- Rank or enumerate candidates using documented rules without accepting a path before validation.
- Bound depth, candidate fanout, and work to resist adversarial chain explosions.
- Keep AIA network fetching out of the initial implementation.

## Acceptance criteria

- Straight, cross-signed, ambiguous, duplicate, cyclic, and incomplete chain fixtures behave deterministically.
- Candidate construction is policy-independent and returns paths for #324-G to validate.
- Search limits prevent exponential or unbounded work.
- Trust anchors are treated as anchors, not blindly as ordinary intermediates.
- Errors distinguish no-path from parser/signature/policy failures.

## Dependencies

- #341 and #343.
- Enables #324-G and #324-H.

Jaruso:
**What changed and why**
Adds `src/pki/path_builder.zig` (story **324-F** of epic #324): a deterministic candidate certification-path construction engine that builds paths from a leaf certificate, peer-supplied intermediates, and configured trust anchors. This is path *discovery* only (RFC 4158 territory) — no signature verification, no validity/key-usage/name-constraint/Basic-Constraints policy, and no path acceptance. Every returned path is a candidate for RFC 5280 validation (#324-G) to check in order.

Design, mapped to the issue's scope:

- **Indexing** — each pool is deduplicated by exact DER (first occurrence wins) and indexed by RFC 5280 §7.1 issuer/subject name chaining through `x509.Name.chaining_key`, not byte-exact `Name.raw`. The key preserves RDN sequence order, compares attributes within an RDN as a set, includes attribute OIDs, handles `domainComponent` with caseIgnoreIA5Match, and prepares required PrintableString/UTF8String DirectoryString values using RFC 4518 stored-value preparation backed by generated RFC 3454 / Unicode 3.2 data (B.2 full case mapping, mappings to nothing, NFKC, prohibited/unassigned rejection, and insignificant-space handling). Runtime preparation rejects Unicode 3.2 unassigned input before mapping, and the generator pins B.3 fallback case mapping to `tools/unicode/UnicodeData-3.2.0.txt` instead of the host Python Unicode version. RFC 4518 insignificant-space handling uses the literal normative Appendix A combining-mark ranges, so SPACE followed by Appendix A marks remains significant while Unicode-category-only marks such as U+05BD follow the definitive RFC table. Other unsupported value types retain deterministic exact-bytes behavior under their own tag. AKI/SKI key identifiers are ranking hints, never filters — real-world AKIs are wrong often enough that RFC 4158 §2.4.2 treats them as sorting data.
- **Determinism** — sibling candidates are tried in a documented total order: key-identifier agreement (match / no evidence / mismatch), then anchors before intermediates, then ascending input index. Paths are emitted in depth-first order under that ordering, so the enumeration is a pure function of inputs and limits.
- **Anchors are anchors** — they terminate paths and are never extended or traversed as intermediates; a cross-signed root reaches a second anchor only through an explicit cross-certificate in the pool.
- **Cycle/duplicate safety** — byte-exact loop detection: a certificate never appears twice in one path, so cyclic and duplicate inputs (including the peer re-sending the root) terminate deterministically.
- **Bounds** — configurable caps on input pool sizes, path length, per-node fanout (applied after ranking, so the kept candidates are the best-ranked), returned paths, and a global candidate-visit budget that defeats adversarial chain explosions. Traversal is iterative (explicit stack), so search depth never becomes attacker-controlled recursion. DirectoryString canonical combining ordering is stable and bounded by a counting-sort pass over starter-delimited runs rather than an inversion-based sort.
- **Error taxonomy** — `NoCandidatePath` means the bounded search space was provably exhausted; `SearchLimitExceeded` means a limit stopped enumeration first (paths already found are still returned, flagged `truncated`). The path-builder error set contains no parser/signature/policy variants by construction: inputs are already-parsed #341 views and no cryptographic or policy judgement happens here.

AIA network fetching is intentionally out of scope per the issue. Wired into `src/pki/root.zig` exports and its `test {}` block like the existing modules; `CHANGELOG.md` updated.

**Related issues**
Closes #344. Parent epic: #324 (research story 324-F). Consumes #341 (typed certificate model) and sits beside #343 (signature verification); enables #324-G (path validation) and #324-H.

**Tests**
`src/pki/path_builder_tests.zig` covers straight chains, cross-signed roots, ambiguous issuers, deterministic replay, duplicate collapse, redundant roots, cycles, incomplete chains, anchor termination, limits, visit-budget starvation regressions, RFC 5280/RFC 4518 issuer-name chaining, significant SPACE+Appendix-A-combining-mark non-matching, RFC Appendix-A divergence candidate discovery, and allocation-failure leak checks.

`src/pki/x509_tests.zig` covers name-key structure, PrintableString/UTF8String unification, domainComponent caseIgnoreIA5Match, RFC 4518 DirectoryString regressions (`Straße`/`STRASSE`, SOFT HYPHEN removal, composed/decomposed `é`, full-width Latin compatibility normalization, NBSP space handling, significant SPACE+Mn and SPACE+Mc/CCC0 non-equivalence, RFC Appendix-A divergence for U+05BD, and a negative case), prohibited/unassigned stored-value rejection including post-Unicode-3.2 capitals with modern mappings, Unicode-3.2-assigned characters absent from RFC 3454 B.2 remaining accepted (`U+10A0`, `U+04C0`), `Certificate.isSelfIssued()` using nontrivial RFC 4518 chaining, and allocation-failure coverage for certificate parsing and name-key construction. `src/pki/x509.zig` internal tests cover stable canonical ordering, starter boundaries, leading non-starters, and a 4096+4096 adversarial descending-class run with a linear lookup-count bound.

Validated locally:

- [x] `zig build test-pki --summary all` — 133/133 passed
- [x] `zig build test --summary all` — 1398/1399 passed, 1 existing skip
- [x] `zig build test-integration --summary all` — 65/69 passed, 4 skips
- [x] `git diff --check`
- [x] regenerated `src/pki/rfc4518_data.zig` and verified deterministic output across generate+format cycles

**Security impact**
This is untrusted-input-adjacent TLS code: intermediates come from network peers. The engine's work is bounded on every axis an attacker influences — pool sizes are capped before indexing, the global candidate-visit budget caps total search work regardless of pool shape, fanout and depth are capped per node/path, and loop detection plus DER-exact deduplication prevent cycle and duplicate amplification. No recursion; all state lives in one arena with `errdefer` teardown, and allocation-failure sweeps cover both path construction and the new name-key construction path. The name-preparation path rejects Unicode 3.2 unassigned input before any mapping can erase it, uses pinned Unicode 3.2 fallback case data, classifies RFC 4518 significant spaces with the literal Appendix A combining-mark table, and uses bounded canonical ordering for combining marks. No trust decision is made here — the module cannot accept a chain, only propose candidates for #345 to validate, and anchors are kept structurally distinct from peer input via `Element.source`. No parsing of raw bytes happens in the path-builder module (inputs are #341 parsed views).

**Performance impact**
Not on any existing hot path — nothing calls this module yet (#324-G will). Search work is bounded by the limits above (defaults: 256 candidate visits, depth 8, fanout 8); indexing is O(n log n) in the deduplicated pool size. Name chaining uses generated static tables and binary-search lookups over DER-bounded X.509 name inputs; canonical combining ordering is stable O(n) per prepared string.

**Checklist**
- [x] README / CHANGELOG updated if operator-visible behavior changed (CHANGELOG entry updated; no operator-visible behavior change)


src/pki/path_builder.zig
```diff
@@ -0,0 +1,403 @@
...
+    fn collect(
+        self: *Search,
+        out: *std.ArrayList(Candidate),
+        pool: *const Pool,
+        source: Source,
+        child: *const x509.Certificate,
+        child_keyid: ?[]const u8,
+        current_path: []const Element,
+    ) Error!void {
+        for (pool.matchSubject(child.issuer.raw)) |entry| {
```
Jaruso: **Blocking — this does not satisfy the normalized issuer/subject acceptance criterion.** `Name.raw` equality is encoding equality, not RFC 5280 §7.1 distinguished-name equality. DER canonicalizes the ASN.1 encoding rules, but it does not normalize DirectoryString values: for example, `CN=Example CA` encoded as UTF8String and the same value encoded as PrintableString have different `raw` bytes and must still match for name chaining. RFC 5280 also requires case folding and insignificant-space handling for the mandatory PrintableString/UTF8String forms.

With the current index, a leaf whose issuer is UTF8String `Example CA` and an otherwise matching CA whose subject is PrintableString `Example CA` deterministically returns `NoCandidatePath`, so the builder cannot construct some conforming public-style chains.

Please add a bounded path-chaining comparison/canonical key to `x509.Name` and index/lookup using that representation rather than `raw`. Conceptually:

```zig
pub fn eqlForChaining(a: *const Name, b: *const Name) bool {
    if (a.rdns.len != b.rdns.len) return false;
    for (a.rdns, b.rdns) |a_rdn, b_rdn| {
        if (!rdnEqlForChaining(a_rdn, b_rdn)) return false;
    }
    return true;
}
```

`rdnEqlForChaining` must compare the RDN attributes as a set, match attribute OIDs, and normalize supported DirectoryString values per RFC 5280 §7.1/RFC 4518 (at minimum the required PrintableString and UTF8String forms, including caseIgnoreMatch and insignificant spaces). For efficient indexing, the parser can produce an arena-owned canonical chaining key when the `Name` is parsed; alternatively, use a coarse index plus bounded semantic equality filtering.

Please add regressions where issuer and subject are semantically equal but differ by UTF8String versus PrintableString encoding, and where case/insignificant spaces differ. The current all-UTF8, byte-identical fixtures cannot detect this failure.

src/pki/path_builder.zig
```diff
@@ -0,0 +1,403 @@
...
+            if (self.visit_budget == 0) {
```
Jaruso: **Blocking — the work budget is charged before ranking/fanout, so it can suppress the best-ranked path.** `collect` scans anchors first, decrements `visit_budget` for each raw match, and only later does `gatherCandidates` sort by key-id agreement and shrink to `max_fanout`. Two correctness problems follow:

1. A group of mismatching/unknown anchors can exhaust the budget before an AKI/SKI-matching intermediate is collected, even though the documented order says the matching intermediate must be tried first.
2. Candidates that will be discarded by `max_fanout` still consume the global budget. A wide first node can therefore leave zero budget for the retained best candidate to gather its own issuer, returning `SearchLimitExceeded` even when the intended bounded traversal needs only two edge visits.

For example, with a leaf AKI `good`, several same-subject anchors with SKI `bad`, one same-subject intermediate with SKI `good` leading to a root, `max_fanout = 1`, and a small visit budget, this implementation spends the budget on the bad anchors and never returns the documented highest-ranked path.

The budget should be applied to candidates actually attempted by the DFS *after* sibling ranking and fanout, not to unsorted pool entries scanned while constructing the candidate list. A safe shape is:

```zig
fn gatherCandidates(...) ![]const Candidate {
    try collectAllSubjectMatches(...); // bounded by pool-count limits
    std.mem.sort(Candidate, candidates.items, {}, candidateLessThan);
    if (candidates.items.len > limits.max_fanout) {
        truncated = true;
        candidates.shrinkRetainingCapacity(limits.max_fanout);
    }
    return candidates.items;
}

// Immediately before attempting the selected sorted candidate:
if (search.visit_budget == 0) {
    search.truncated = true;
    break; // stop globally, preserving pending-work/truncation semantics
}
search.visit_budget -= 1;
```

If scanning itself needs a separate global cap, introduce a distinct scan budget and preserve rank with a bounded top-k/rank-aware collection strategy; do not let source iteration order redefine the documented candidate order.

Please add regressions for both cases: (a) a matching intermediate must outrank earlier-collected mismatching anchors under a tight budget, and (b) entries discarded by `max_fanout` must not consume the traversal budget needed to reach the selected candidate's anchor.

Jaruso (commented):
## Review verdict: not ready to merge

I completed a full pass against issue #344 and the parent #324 path-building requirements, including the future #345 validator contract, all four changed files, the full new test suite, and CI run 671. The implementation is strong in its separation of construction from validation, iterative DFS structure, anchor termination, exact-DER duplicate handling, allocation cleanup, and typed no-path versus truncated-search outcomes. CI is green across the reported build/test/format/security jobs.

There are two blocking correctness/interoperability issues:

1. **The global candidate-visit budget is consumed before sibling ranking and fanout are applied.** Because anchors are collected first, lower-ranked anchors can exhaust the budget before a higher-ranked AKI/SKI-matching intermediate is even observed. Candidates later discarded by `max_fanout` also consume the entire global budget, which can prevent the retained best candidate from looking up its own issuer. This violates the documented total order and the acceptance criterion that candidates are ranked/enumerated using the documented rules.

2. **Issuer/subject matching is byte-exact DER rather than normalized distinguished-name matching.** RFC 5280 §7.1 requires PrintableString/UTF8String comparison using StringPrep/caseIgnoreMatch and insignificant-space handling; DER does not make semantically equal DirectoryStrings byte-identical. The issue explicitly requires normalized issuer/subject indexing, so valid chains whose issuer and subject use equivalent but differently encoded DNs currently produce `NoCandidatePath`.

I left detailed inline comments with concrete implementation shapes and regression tests. After these are fixed, I would re-review the affected path-builder and X.509 name-comparison code plus the added tests; the remaining acceptance criteria appear covered by this revision.

src/pki/path_builder.zig
Jaruso: Fixed in 4d26df1 via the canonical-key option. `x509.Name` now carries an arena-owned `chaining_key` computed at parse time (plus `eqlForChaining`): RDNs in sequence order, RDN attributes compared as sorted sets, attribute types by decoded OID, and primitive PrintableString/UTF8String values unified into one caseIgnore class — leading/trailing white space trimmed, internal runs collapsed, ASCII case folded. Other value types stay exact-bytes under their own tag (§7.1 rules (a)/(b)); full RFC 4518 StringPrep (non-ASCII folding) is documented as out of scope, matching the OpenSSL canonical-name baseline. The path builder now indexes and looks up pools by that key.

Regressions added at both levels: UTF8String↔PrintableString equality (chains, while `eqlEncoding` differs), case/insignificant-space variants, BMPString staying outside the class, key structure-sensitivity (one two-attribute RDN ≠ two single-attribute RDNs), and a path-builder test where a noisy UTF8String issuer chains to a canonical PrintableString anchor.

---

src/pki/path_builder.zig
Jaruso: Fixed in 4d26df1, taking the recommended shape: `collect` no longer touches the budget — it gathers all subject matches (bounded by the deduplicated pool-count limits), `gatherCandidates` sorts and applies `max_fanout`, and the DFS loop charges `visit_budget` immediately before attempting each sorted, fanout-bounded candidate, breaking globally with `truncated = true` on exhaustion. Scan work stays bounded without a separate cap: each attempted expansion performs one binary-search lookup per pool, so total scanning is O(attempts × log pool + matches), and matches are bounded by pool size.

Both requested regressions added: (a) three mismatching same-subject anchors vs. one AKI/SKI-matching intermediate under `max_fanout = 1` and a two-visit budget now returns the documented best-ranked path through the matching intermediate; (b) five fanout-discarded same-subject decoys no longer consume the budget needed for the retained candidate to reach its anchor.

---

src/pki/x509.zig
```diff
@@ -614,7 +631,54 @@ const Parser = struct {
...
+        var blob: std.ArrayList(u8) = .empty;
+        const components = attribute.type.components();
+        try appendCount(&blob, self.arena, components.len);
+        for (components) |component| try appendBe32(&blob, self.arena, component);
+        if (isCaseIgnoreStringTag(attribute.value_tag)) {
+            try blob.append(self.arena, 0x00);
+            const normalized = try normalizeDirectoryString(self.arena, attribute.value);
+            try appendCount(&blob, self.arena, normalized.len);
+            try blob.appendSlice(self.arena, normalized);
+        } else {
```
Jaruso, line 697: **Blocking — this still does not implement the RFC 5280 §7.1 comparison that `chaining_key` claims to represent.**

Everything outside primitive PrintableString/UTF8String enters this byte-exact branch. In particular, `domainComponent` is an IA5String, and RFC 5280 §7.1 explicitly requires it to be compared under §7.3's case-insensitive exact rule. A chain containing `DC=EXAMPLE,DC=COM` in one DN and `dc=example,dc=com` in the other therefore returns `NoCandidatePath` today.

The PrintableString/UTF8String branch is also only ASCII folding plus ASCII whitespace compression. RFC 5280 requires the six-step RFC 4518 preparation: Unicode mapping and case folding, Form KC normalization, prohibited-code-point handling, and insignificant-space handling. Concrete conforming pairs that this key keeps different include UTF8String `ÉCOLE` vs `école`, and `Example\u{00A0}CA` vs `Example CA` (NBSP maps to SPACE).

Please select the matching rule by attribute type and generate the key from the rule's prepared Unicode/ASCII representation. A suitable shape is:

```zig
fn attributeChainingBlob(self: *Parser, attribute: *const AttributeTypeAndValue) Error![]const u8 {
    if (attribute.type.eqlComponents(&wk.domain_component)) {
        if (!isPrimitiveIa5(attribute.value_tag)) return error.MalformedName;
        return self.domainComponentBlob(attribute.type, attribute.value); // ASCII case-folded
    }

    if (isRequiredDirectoryString(attribute.value_tag)) {
        const prepared = try stringprep.prepareStoredCaseIgnore(
            self.arena,
            attribute.value_tag,
            attribute.value,
        ); // transcode, map/case-fold, NFKC, prohibit, space handling
        return self.preparedAttributeBlob(attribute.type, .directory_case_ignore, prepared);
    }

    return self.exactAttributeBlob(attribute);
}
```

At minimum add regressions for:

```zig
// Required DirectoryString behavior.
expectChains(CN_UTF8("ÉCOLE"), CN_UTF8("école"));
expectChains(CN_UTF8("Example\u{00A0}CA"), CN_UTF8("Example CA"));

// Mandatory domainComponent behavior.
expectChains(DN("DC=EXAMPLE,DC=COM"), DN("dc=example,dc=com"));
```

If full RFC 4518 behavior is intentionally deferred, the code and changelog cannot describe this as an RFC 5280 §7.1 key, and issue #344's normalized-name acceptance criterion remains incomplete.

Jaruso (commented):
## Re-review verdict: not ready to merge

I re-reviewed the complete PR at `4d26df1`, not only the follow-up commit, against issue #344 and the downstream #345 validator contract.

The **candidate-visit budget fix is correct**: pool scans no longer consume the traversal budget, ranking and `max_fanout` happen first, and the budget is charged immediately before the DFS attempts a retained candidate. The two new regressions cover both previously reported starvation cases. The canonical-key serialization is also structurally sound: RDN sequence order is retained, attributes inside an RDN are sorted as a set, and the length/count prefixes avoid concatenation collisions.

Two blocking name-comparison issues remain:

1. The canonical key is still only an ASCII/OpenSSL-style subset, not the RFC 5280 §7.1 behavior claimed by the API and documentation. PrintableString/UTF8String comparison omits the required RFC 4518 Unicode mapping, case folding, Form KC normalization, and prohibited-character handling, and `domainComponent` IA5String values fall through to byte-exact comparison instead of the required case-insensitive comparison.
2. `Certificate.isSelfIssued()` still compares `issuer` and `subject` with `eqlEncoding`. RFC 5280 defines self-issued using the same §7.1 DN comparison, so a cross-encoding/case/space-equivalent CA is misclassified. That will produce incorrect path-length, name-constraint, and policy-counter behavior when #345 consumes this API.

I left detailed inline comments with implementation shapes and regression cases.

CI run 673 completed successfully. Before merge, please also update the PR description—it still says the builder indexes byte-exact `Name.raw` and lists the old test counts—and run `zig build test-integration`, which remains explicitly unchecked despite the repository checklist requiring it for TLS changes.

src/pki/x509.zig
```diff
@@ -141,17 +141,34 @@ pub const RelativeDistinguishedName = struct {
...
+    /// compare as exact bytes under their own tag (rules (a)/(b)). Two names
+    /// chain exactly when their keys are byte-equal.
+    chaining_key: []const u8,
 
     pub fn eqlEncoding(self: *const Name, other: *const Name) bool {
         return std.mem.eql(u8, self.raw, other.raw);
     }
 
+    /// RFC 5280 §7.1 name-chaining equality (see `chaining_key`).
+    pub fn eqlForChaining(self: *const Name, other: *const Name) bool {
```
Jaruso, line 179: **Blocking — the new chaining equality is not used by `Certificate.isSelfIssued()`.**

`isSelfIssued()` still returns `self.issuer.eqlEncoding(&self.subject)`. RFC 5280 defines a certificate as self-issued when those DNs match according to §7.1, not when their DER encodings are identical. A CA with an issuer encoded as noisy UTF8String and an equivalent PrintableString subject now chains correctly in the path builder but is still reported as non-self-issued.

That is not merely descriptive: #345 will use self-issued status when applying path-length constraints, name constraints, and policy counters. Misclassification changes validation results.

Please update the existing method:

```zig
pub fn isSelfIssued(self: *const Certificate) bool {
    return self.issuer.eqlForChaining(&self.subject);
}
```

Add a certificate-level regression whose issuer and subject differ in encoding/case/insignificant spaces but satisfy `eqlForChaining`, and assert both:

```zig
try testing.expect(!cert.issuer.eqlEncoding(&cert.subject));
try testing.expect(cert.issuer.eqlForChaining(&cert.subject));
try testing.expect(cert.isSelfIssued());
```

src/pki/x509.zig
Jaruso: Addressed in 320f0b4, but with a scope I want to be upfront about rather than claim more than is actually there.

Implemented:
- `domainComponent` now compares with `caseIgnoreIA5Match` (RFC 4517 §4.2.3) — ASCII case-fold over the already-validated IA5String bytes. Since IA5String is ASCII-only by definition, this is an *exact* implementation of the rule, not an approximation.
- The caseIgnore class now folds Latin-1 Supplement letters (U+00C0-U+00DE excluding U+00D7) in addition to ASCII, via codepoint-level normalization (`std.unicode.Utf8View`/`utf8Encode` — the input is guaranteed valid UTF-8 by the point this runs, since UTF8String content already passed `der.validateUtf8` and PrintableString is an ASCII subset). This covers `ÉCOLE`/`école` exactly, and all three of your concrete test vectors pass: `expectChains(CN_UTF8("ÉCOLE"), CN_UTF8("école"))`, `expectChains(CN_UTF8("Example\u{00A0}CA"), CN_UTF8("Example CA"))`, `expectChains(DN("DC=EXAMPLE,DC=COM"), DN("dc=example,dc=com"))`.
- U+00A0 NO-BREAK SPACE now folds into the space-collapsing algorithm alongside ASCII white space.

Deliberately not implemented, and now documented as such on `Name.chaining_key` and in the CHANGELOG rather than silently assumed: full RFC 4518 StringPrep — Unicode case folding for scripts beyond Latin-1 (Cyrillic, Greek, etc.), NFKC compatibility normalization, and prohibited-codepoint rejection. That requires generated Unicode Character Database tables (decomposition/composition mappings, combining classes, full case-folding data) — thousands of entries I'm not willing to hand-write from memory into security-sensitive name-matching code without a proper sourcing/generation pipeline and dedicated review of the data itself, separate from this PR's logic. A CA-issued name using non-Latin-1 case variation won't chain today.

If you'd like, I can open a follow-up issue under epic #324 scoped specifically to full RFC 4518 StringPrep (with a real data-generation approach, e.g. vendoring/generating from the Unicode Character Database rather than hand-transcribing), so it gets the review depth that data deserves rather than being folded into path-building review. Let me know if you'd rather block this PR on it instead.

Regressions added in `x509_tests.zig`: `"name chaining folds Latin-1 Supplement case and NO-BREAK SPACE"`, `"domainComponent RDN values compare with caseIgnoreIA5Match"`.

---

src/pki/x509.zig
Jaruso: Fixed in 320f0b4 exactly as suggested — `isSelfIssued()` now returns `self.issuer.eqlForChaining(&self.subject)`.

Added the requested regression (`"isSelfIssued uses name chaining, not encoding equality"`): a certificate whose issuer is UTF8String `"  Example   CA "` and subject is PrintableString `"example ca"`, asserting all three:
```zig
try testing.expect(!cert.issuer.eqlEncoding(&cert.subject));
try testing.expect(cert.issuer.eqlForChaining(&cert.subject));
try testing.expect(cert.isSelfIssued());
```

---

src/pki/x509.zig
```diff
@@ -1139,6 +1233,93 @@ fn validateDirectoryString(elem: der.Element) Error!void {
...
+        ' ', '\t', '\n', 0x0b, 0x0c, '\r', 0xa0 => true,
+        else => false,
+    };
+}
+
+/// Case-fold one codepoint for the caseIgnore class: ASCII and Latin-1
+/// Supplement letters only (see `Name.chaining_key`). U+00D7 MULTIPLICATION
+/// SIGN sits inside the Latin-1 uppercase-letter range but is not a letter
+/// and has no lowercase counterpart, so it is excluded.
+fn foldCaseIgnoreCodepoint(codepoint: u21) u21 {
```
Jaruso: **Blocking — the replacement still does not implement the claimed Latin-1/RFC 5280 caseIgnore behavior.** This function can only return one code point for one input code point, but RFC 5280 §7.1 requires RFC 4518 preparation with RFC 3454 Appendix B.2 *full* case mapping and NFKC. Those mappings are not limited to `uppercase + 0x20`.

A concrete counterexample is `U+00DF LATIN SMALL LETTER SHARP S`: RFC 3454 B.2 maps it to two code points, `"ss"`. Therefore `CN=Straße` and `CN=STRASSE` must compare equal after preparation, but this implementation leaves `ß` unchanged and lowers the other value to `strasse`, so the path builder returns `NoCandidatePath`. That is inside the Latin-1 range the new documentation says is covered exactly. Other missed Latin-1 cases include `U+00B5 -> U+03BC` and `U+00AD SOFT HYPHEN -> nothing`; NFKC also handles compatibility and composed/decomposed equivalents.

This affects both consumers of the key: issuer/subject lookup and `Certificate.isSelfIssued()`. It cannot safely be deferred to #345 because a path omitted here never reaches validation.

Please replace the hand-written simple fold with a bounded RFC 4518 preparation routine backed by authoritative Unicode 3.2/RFC 3454 data. The key-building shape needs to allow one-to-none and one-to-many output rather than allocating `content.len` and returning one folded code point:

```zig
const Prepared = union(enum) {
    value: []const u8,
    undefined,
};

fn prepareStoredDirectoryString(
    arena: std.mem.Allocator,
    content: []const u8,
) !Prepared {
    var mapped: std.ArrayList(u21) = .empty;
    // 1. transcode already-validated PrintableString/UTF8String
    // 2. apply RFC 4518 mappings, including RFC 3454 B.1 and full B.2
    // 3. normalize using Unicode 3.2 NFKC
    // 4. return .undefined for prohibited/unassigned code points
    // 5. bidi is ignored by RFC 4518
    // 6. canonicalize insignificant spaces
    return .{ .value = try encodeCanonical(arena, mapped.items) };
}
```

The data tables should be generated from pinned authoritative inputs rather than manually approximated. If preparation is undefined, either retain that state in `Name` so equality is always false, or reject it consistently as `MalformedName`; do not create a normal key for it.

Please add at least these regressions, plus one path-builder-level check:

```zig
try expectChains("Straße", "STRASSE");       // B.2 one-to-many
try expectChains("soft\u{00AD}hyphen", "softhyphen"); // B.1 removal
try expectChains("é", "e\u{0301}");         // NFKC
try expectChains("ＡＣＭＥ", "ACME");       // compatibility normalization
```

Also add an `isSelfIssued()` variant using one of these pairs so the validator-facing API cannot regress independently.

Jaruso (commented):
## Third review verdict: one blocker remains

The previous two findings are addressed correctly:

- the traversal budget is now charged only for ranked, fanout-bounded candidates actually attempted;
- `domainComponent` comparison and `Certificate.isSelfIssued()` now use the intended chaining semantics;
- the added regressions are focused and CI run 675 is green.

However, the new DirectoryString canonicalization is still not sufficient to satisfy the RFC 5280 §7.1 normalized-name requirement, and it is incorrect even within the narrower scope it now claims to support exactly. I left one detailed inline blocker with concrete counterexamples and an implementation/test shape.

The central problem is that RFC 5280 requires RFC 4518's six-step preparation using RFC 3454 B.2 full case mapping and NFKC. The current one-codepoint folding function can only perform one-to-one mappings. RFC 3454 includes required one-to-many and removal mappings such as `U+00DF LATIN SMALL LETTER SHARP S -> "ss"` and `U+00AD SOFT HYPHEN -> nothing`. Consequently, semantically matching names such as `CN=Straße` and `CN=STRASSE` still produce different chaining keys, causing both a path-building false negative and incorrect `isSelfIssued()` classification.

Because a false negative prevents #345 from ever evaluating an otherwise valid candidate path, this remains blocking rather than a validator-policy follow-up. The PR description also still describes the original byte-exact `Name.raw` design and should be refreshed once the implementation is final.

tools/generate_rfc4518_data.py
```diff
@@ -0,0 +1,222 @@
...
+
+
+def rfc4518_map(cp: int) -> list[int]:
+    if map_to_space(cp):
+        return [0x20]
+    if map_to_nothing(cp):
+        return []
+    return chars_to_scalars(stringprep.map_table_b2(chr(cp)))
+
+
```
Jaruso, line 271: **Blocking — the generated table can erase Unicode-3.2 unassigned code points before the A.1 prohibition check.**

This calls `stringprep.map_table_b2()` for every scalar. Python's `stringprep` module uses the Unicode 3.2 database for its classification/normalization tables, but its generated B.3 helper falls back to the interpreter's `str.lower()` for non-exception mappings. That means code points that were unassigned in Unicode 3.2 but assigned in newer Unicode versions can receive modern lowercase mappings.

The checked-in output demonstrates the bug:

- `rfc4518_data.map` contains an entry for U+1C90.
- Its `map_data` output is U+10D0.
- U+1C90 is in RFC 3454 table A.1, while U+10D0 is assigned.
- `normalizeDirectoryString()` checks `unassigned` only after applying this mapping and NFKC.

Therefore a UTF8String containing U+1C90 is accepted and prepares identically to U+10D0 instead of returning `error.NamePreparationFailed`. This is both a standards failure and a false name equivalence in a certificate-chain index.

Preserve the explicit RFC 4518 space/nothing mappings, but do not apply B.2 to A.1 scalars. At minimum:

```python

def rfc4518_map(cp: int) -> list[int]:
    if map_to_space(cp):
        return [0x20]
    if map_to_nothing(cp):
        return []
    ch = chr(cp)
    if stringprep.in_table_a1(ch):
        return [cp]  # must survive until the prohibit step
    return chars_to_scalars(stringprep.map_table_b2(ch))
```

Also add generator invariants so regeneration fails if any A.1 scalar outside the explicit space/nothing sets gets a non-identity mapping, and preferably verify the generated B.2 table against a fixed authoritative RFC-3454-derived source rather than runtime Unicode behavior.

Required regression:

```zig
try testing.expectError(
    error.NamePreparationFailed,
    x509.parseNameRaw(arena, try nameWithCnTag(arena, 0x0c, "\u{1C90}"), .{}),
);
```

Also test another post-3.2 capital with a modern mapping, such as U+1E900 or U+A7B0, and assert that the generated table never maps A.1 inputs into assigned outputs.

src/pki/x509.zig
```diff
@@ -1139,6 +1227,279 @@ fn validateDirectoryString(elem: der.Element) Error!void {
...
+    }
+}
+
+fn canonicalOrder(codepoints: []u21) void {
+    var index: usize = 1;
+    while (index < codepoints.len) : (index += 1) {
+        const ccc = combiningClass(codepoints[index]);
+        if (ccc == 0) continue;
+        var swap_index = index;
+        while (swap_index > 0 and combiningClass(codepoints[swap_index - 1]) > ccc) : (swap_index -= 1) {
```
Jaruso: **Blocking security issue — canonical ordering is O(n²) on an untrusted DirectoryString.**

This insertion-sort loop swaps once per inversion, and each comparison performs another binary search through the combining-class table. A peer can construct a valid UTF8String such as:

```text
"A" + U+0315 repeated N times + U+0300 repeated N times
```

U+0315 has CCC 232 and U+0300 has CCC 230. Both are valid, assigned, non-prohibited combining marks. Every U+0300 must move before every U+0315, producing approximately N² swaps. The default DER limit permits a 1 MiB element, so this can reach hundreds of thousands of combining marks and tens of billions of swaps during certificate parsing, before the path builder's visit/fanout/depth bounds can help.

Canonical ordering must be stable for equal CCC values and must not cross a starter, but CCC is only a `u8`. Implement this in bounded linear or O(n log n) work, for example:

```zig
fn canonicalOrder(codepoints: []u21) void {
    var start: usize = 0;
    while (start < codepoints.len) {
        // Find one starter-delimited combining sequence.
        var end = start + 1;
        while (end < codepoints.len and combiningClass(codepoints[end]) != 0) : (end += 1) {}

        // Stable counting/bucket ordering for classes 1...255, preserving
        // original order among equal classes. Keep the starter at `start`.
        stableOrderByCombiningClass(codepoints[start..end]);
        start = end;
    }
}
```

A stable merge sort per non-starter run is also acceptable, but precompute CCC values rather than repeatedly binary-searching inside an inversion loop. An additional strict scalar/output bound can be defense in depth, not a substitute for removing the quadratic algorithm.

Add a regression using a sizeable descending-class run and a testable work bound (operation counter in a test-only helper or a linear-algorithm structural test), plus a correctness case proving equal CCC order is stable and starters remain sequence boundaries.

Jaruso (commented):
## Review verdict: still not ready to merge

The latest commit resolves the prior RFC 4518 functional gaps: one-to-many B.2 mappings, mappings to nothing, Unicode 3.2 NFKC, prohibited/unassigned checks, insignificant-space encoding, the path-builder regression, nontrivial `isSelfIssued`, updated PR documentation, allocation-failure coverage, and integration-test reporting are all present. The previously reviewed traversal-budget, deterministic ordering, anchor termination, cycle/duplicate, error-taxonomy, and search-bound behavior remain intact. CI run 677 is green.

However, the generated-data implementation introduced two blocking issues in the new X.509 name-preparation path:

1. **Unicode 3.2 unassigned values can be mapped into assigned values before the A.1 check.** The generator invokes Python `stringprep.map_table_b2()` for every scalar, including A.1 values. CPython's generated `stringprep` implementation ultimately falls back to the running interpreter's `str.lower()` for non-exception case mappings, so characters assigned after Unicode 3.2 can acquire modern mappings. The checked-in table proves this has happened: U+1C90 is emitted as a mapping entry whose output is U+10D0. U+1C90 is unassigned in Unicode 3.2, but the runtime checks `unassigned` only after mapping/NFKC, by which time the original scalar has disappeared. A name containing U+1C90 is therefore accepted and canonicalizes equal to a valid U+10D0 name instead of returning `NamePreparationFailed`.

2. **Canonical ordering is quadratic on attacker-controlled input.** `canonicalOrder` is insertion sort and performs a table lookup plus swap for every inversion. A valid UTF8String containing a starter, a long run of U+0315 (CCC 232), then a long run of U+0300 (CCC 230) causes approximately N²/4 swaps. Both marks survive RFC 4518 preparation. With the existing 1 MiB DER element limit, a peer certificate can contain hundreds of thousands of such marks, turning certificate parsing into tens of billions of swaps before path search limits are relevant.

I left detailed inline comments with concrete fixes and regression requirements. Because the connected account owns this PR, GitHub will not permit a formal `REQUEST_CHANGES`; this is submitted as `COMMENT`, but both findings are blocking.

tools/generate_rfc4518_data.py
Jaruso, line 271: **Blocking — reject Unicode 3.2 unassigned input before any mapping can erase it.**

The current generator calls `stringprep.map_table_b2(chr(cp))` for every scalar that is not explicitly mapped to space/nothing. Python’s `stringprep` classifications are based on Unicode 3.2, but its B.2 implementation can use the running interpreter’s modern lowercase behavior for code points outside the original table. That allows a scalar that was **unassigned in Unicode 3.2** to be transformed into an assigned Unicode-3.2 scalar before the Zig runtime checks table A.1.

Concrete reproduction from the checked-in generated data:

```text
input:  U+1C90  (unassigned in Unicode 3.2 / RFC 3454 A.1)
map:    U+10D0  (assigned Georgian letter)
result: accepted and equal to a genuine U+10D0 name
```

`normalizeDirectoryString()` currently checks `unassigned` only on the mapped/NFKC output, so the invalid original scalar has already disappeared. A certificate containing U+1C90 can therefore produce a normal chaining key instead of `error.NamePreparationFailed`, creating a false issuer/subject equivalence.

Please fix this at **both layers**:

### 1. Make generation Unicode-3.2-safe

Do not run B.2 mapping on A.1 input scalars. Preserve them unchanged so they remain rejectable, while retaining the explicit RFC 4518 mappings to space/nothing:

```python
def rfc4518_map(cp: int) -> list[int]:
    if map_to_space(cp):
        return [0x20]
    if map_to_nothing(cp):
        return []

    ch = chr(cp)
    if stringprep.in_table_a1(ch):
        # RFC 4518 stored-value preparation must reject this input.
        # Never allow the host Python Unicode version to case-map it.
        return [cp]

    return chars_to_scalars(stringprep.map_table_b2(ch))
```

Add generation-time invariants so a future Python upgrade cannot silently reintroduce the bug:

```python
for cp in range(MAX_SCALAR + 1):
    if not valid_scalar(cp):
        continue

    mapped = rfc4518_map(cp)
    if stringprep.in_table_a1(chr(cp)) and not map_to_space(cp) and not map_to_nothing(cp):
        assert mapped == [cp], (
            f"Unicode 3.2 unassigned U+{cp:04X} was mapped to "
            f"{[f'U+{x:04X}' for x in mapped]}"
        )
```

Preferably also generate from a pinned, repository-controlled Unicode-3.2/RFC-3454 source or validate representative table hashes/counts so regeneration is not dependent on undocumented host-runtime behavior.

### 2. Add a runtime defense-in-depth check before mapping

Even with a correct generator, the parser should reject A.1 input before consulting the mapping table. That prevents a corrupted or incorrectly regenerated mapping table from bypassing the prohibition:

```zig
while (iterator.nextCodepoint()) |codepoint| {
    if (containsRange(rfc4518_data.unassigned[0..], codepoint)) {
        return error.NamePreparationFailed;
    }

    if (try mappingFor(
        rfc4518_data.map[0..],
        rfc4518_data.map_data[0..],
        codepoint,
    )) |replacement| {
        try appendScalars(&mapped, arena, replacement);
    } else {
        try appendScalar(&mapped, arena, codepoint);
    }
}
```

Keep the existing post-normalization unassigned/prohibited check too; the pre-check protects the input repertoire, while the post-check protects mapped/normalized output.

### Required regressions

```zig
try testing.expectError(
    error.NamePreparationFailed,
    x509.parseNameRaw(
        arena,
        try nameWithCnTag(arena, 0x0c, "\u{1C90}"),
        .{},
    ),
);
```

Add at least one more code point assigned after Unicode 3.2 that has a modern case mapping, such as U+1E900 or U+A7B0, and assert it is rejected. Also add a generator test/invariant proving no A.1 input can have a non-identity B.2 mapping outside the explicit space/nothing rules.

Finally, regenerate the Zig table and confirm U+1C90 is absent from `map` (or maps only to itself) and remains present in `unassigned`.

src/pki/x509.zig
Jaruso: **Blocking security issue — replace this inversion-based insertion sort with stable bounded ordering.**

`canonicalOrder()` performs one swap per combining-class inversion. Each loop comparison also calls `combiningClass()`, which binary-searches the generated CCC table. The input is certificate-controlled, and the default DER element limit is 1 MiB, so this is not a small-string concern.

An attacker can construct a valid UTF8String equivalent to:

```text
"A" + U+0315 × N + U+0300 × N
```

- U+0315 has canonical combining class 232.
- U+0300 has canonical combining class 230.
- Both are valid, assigned, non-prohibited marks.
- Canonical ordering must move every U+0300 before every U+0315.

The current implementation performs approximately `N²` swaps. With hundreds of thousands of marks fitting in a large DER element, certificate parsing can require tens of billions of operations **before** any path-builder depth/fanout/visit bound applies.

Please implement stable ordering in O(n) or O(n log n), scoped to each starter-delimited combining sequence. Because CCC is a `u8`, stable counting sort is straightforward and gives deterministic linear work.

A suitable implementation shape is:

```zig
fn canonicalOrder(
    arena: std.mem.Allocator,
    codepoints: []u21,
) Error!void {
    if (codepoints.len < 2) return;

    const scratch = try arena.alloc(u21, codepoints.len);
    var segment_start: usize = 0;

    while (segment_start < codepoints.len) {
        // A leading non-starter run is itself one sortable segment. When a
        // starter is present, keep it fixed and sort only the following marks.
        const first_ccc = combiningClass(codepoints[segment_start]);
        const sortable_start = segment_start + @intFromBool(first_ccc == 0);

        var segment_end = sortable_start;
        while (segment_end < codepoints.len and
            combiningClass(codepoints[segment_end]) != 0)
        {
            segment_end += 1;
        }

        try stableOrderByCcc(
            codepoints[sortable_start..segment_end],
            scratch[0 .. segment_end - sortable_start],
        );

        segment_start = segment_end;
    }
}

fn stableOrderByCcc(values: []u21, scratch: []u21) Error!void {
    if (values.len < 2) return;

    var counts = [_]usize{0} ** 256;
    for (values) |cp| {
        const ccc = combiningClass(cp);
        // A starter here would mean the caller selected the wrong segment.
        if (ccc == 0) return error.NamePreparationFailed;
        counts[ccc] = std.math.add(usize, counts[ccc], 1) catch
            return error.NamePreparationFailed;
    }

    var offsets = [_]usize{0} ** 256;
    var next: usize = 0;
    for (1..256) |ccc| {
        offsets[ccc] = next;
        next = std.math.add(usize, next, counts[ccc]) catch
            return error.NamePreparationFailed;
    }

    // Forward traversal plus per-class cursors preserves original order for
    // equal CCC values, as Unicode canonical ordering requires.
    var cursors = offsets;
    for (values) |cp| {
        const ccc = combiningClass(cp);
        scratch[cursors[ccc]] = cp;
        cursors[ccc] += 1;
    }

    @memcpy(values, scratch[0..values.len]);
}
```

The exact API can differ. Important properties are:

1. Stable ordering among equal CCC values.
2. No movement across a CCC-0 starter.
3. Correct handling of an initial run that begins with non-starters.
4. O(n) or O(n log n) worst-case work.
5. Checked allocation/arithmetic and typed failure propagation.

Update the caller from:

```zig
canonicalOrder(decomposed.items);
```

to something that propagates allocation/failure:

```zig
try canonicalOrder(arena, decomposed.items);
```

### Required tests

Add a large adversarial run, not merely a three-character correctness example:

```zig
// Build: 'A', 4096 × U+0315 (CCC 232), 4096 × U+0300 (CCC 230).
// After ordering: starter, all U+0300, then all U+0315.
```

Also add:

- Equal-CCC stability: two distinct CCC-230 marks must retain their input order.
- Starter boundaries: marks following `A` must not reorder across a later `B` starter.
- Leading non-starters: an initial combining sequence must still be ordered correctly.
- A testable work bound. Ideally expose a test-only operation counter or structure the helper so the test can assert writes/lookups are proportional to input length rather than relying on a wall-clock timeout.

A parser/name-preparation limit may be useful defense in depth, but it is not a substitute for removing this quadratic algorithm.

Jaruso (commented):
## Follow-up implementation guidance — two blocking fixes remain

I expanded both findings below into concrete implementation and regression requirements. The RFC 4518 functional behavior added in this revision is otherwise on the right track, but these two issues must be fixed before merge because one creates incorrect name equivalences and the other allows attacker-controlled superlinear certificate-parse work.

After fixing them, regenerate `src/pki/rfc4518_data.zig`, verify the generated file is deterministic, and rerun formatting, `test-pki`, the full test suite, integration tests, and allocation-failure coverage.

Jaruso (commented):
## Re-review verdict: two blockers remain

I re-reviewed head `9fdbdd7` against issue #344 after the latest fixes. The two findings from the previous pass are addressed correctly:

- Unicode 3.2 A.1 input is now rejected before mapping, with generator invariants and regressions.
- Canonical combining ordering is now stable and linear-time over starter-delimited runs, with adversarial work-bound coverage.

CI run 678 is green, and I did not find a regression in the path-builder traversal, ranking, anchor handling, cycle/duplicate handling, or search-limit behavior.

Two RFC 4518 correctness issues remain in the name-preparation implementation:

1. Insignificant-space handling currently classifies every U+0020 as a removable/collapsible space. RFC 4518 §2.6.1 defines a space more narrowly: U+0020 followed by **no combining marks**. A space carrying a combining mark is significant. The current code therefore creates false name equivalences.
2. The generator still derives B.2 mappings through Python's host-runtime `str.lower()` fallback for Unicode-3.2-assigned characters. The A.1-input guard fixes the previously reported bypass, but assigned inputs such as U+10A0 and U+04C0 are still given mappings that did not exist in RFC 3454 B.2, causing valid Unicode 3.2 names to fail preparation.

Both affect the canonical issuer/subject index itself. They can either merge distinct DNs into one candidate bucket or reject an otherwise valid DN before path construction, so they cannot be deferred to #345.

Detailed inline comments include implementation shapes and required tests. Because this is a review on the PR author's own account, GitHub records it as `COMMENTED` rather than `REQUEST_CHANGES`.

src/pki/x509.zig
```diff
@@ -1139,6 +1227,339 @@ fn validateDirectoryString(elem: der.Element) Error!void {
...
+    }
+    return null;
+}
+
+fn encodeCaseIgnoreAttributeValue(arena: std.mem.Allocator, codepoints: []const u21) Error![]const u8 {
+    var out: std.ArrayList(u8) = .empty;
+    var index: usize = 0;
+    var wrote_run = false;
+
+    while (index < codepoints.len and codepoints[index] == ' ') : (index += 1) {}
```
Jaruso: **Blocking — insignificant-space handling incorrectly removes significant SPACE+combining-mark sequences.**

RFC 4518 §2.6.1 does **not** define every U+0020 as a space for this step. Its exact definition is: a SPACE (U+0020) code point **followed by no combining marks**. Appendix A is normative and defines combining marks as Unicode properties `Mn`, `Mc`, or `Me`.

This implementation checks only:

```zig
codepoints[index] == ' '
```

at the leading-space scan, inner-space scan, and content-run boundary. Consequently, a U+0020 followed by a combining mark is discarded or collapsed even though the standard treats that U+0020 as significant content.

Concrete false equivalence:

```text
CN="<SPACE>\u0301A"
CN="\u0301A"
```

The first SPACE is followed by U+0301 COMBINING ACUTE ACCENT, so it is **not** an insignificant space. The two prepared values must differ. The current encoder removes the first code point from the first value and produces the same key for both names.

Do not implement this using `combiningClass(cp) != 0`. RFC 4518 uses the Unicode mark properties, and some normative combining marks have canonical combining class zero. U+093E DEVANAGARI VOWEL SIGN AA is an `Mc` combining mark listed in Appendix A but has CCC 0. A CCC-only test would leave the bug in place.

Please generate a pinned Unicode-3.2/RFC-4518 Appendix-A range table alongside the existing tables. For example, the generator can use the pinned database already present:

```python
combining_mark_values: list[int] = []

for cp in range(MAX_SCALAR + 1):
    if not valid_scalar(cp):
        continue
    if UCD.category(chr(cp)) in ("Mn", "Mc", "Me"):
        combining_mark_values.append(cp)

emit_ranges(
    "combining_marks",
    coalesce_ranges(combining_mark_values),
    lines,
)
```

Then classify insignificant spaces by position:

```zig
fn isCombiningMark(codepoint: u21) bool {
    return containsRange(rfc4518_data.combining_marks[0..], codepoint);
}

fn isInsignificantSpaceAt(codepoints: []const u21, index: usize) bool {
    if (codepoints[index] != ' ') return false;
    return index + 1 == codepoints.len or
        !isCombiningMark(codepoints[index + 1]);
}
```

Refactor `encodeCaseIgnoreAttributeValue()` so only `isInsignificantSpaceAt(...)` entries participate in leading/trailing removal and inner-run replacement. A significant U+0020 followed by a mark must be emitted as normal content together with its combining sequence.

Required regressions:

```zig
const spaced_mn = parseName(" \u{0301}A");
const bare_mn = parseName("\u{0301}A");
try testing.expect(!spaced_mn.eqlForChaining(&bare_mn));

// Prevent the tempting but incorrect CCC != 0 implementation.
const spaced_mc_ccc0 = parseName(" \u{093E}A");
const bare_mc_ccc0 = parseName("\u{093E}A");
try testing.expect(!spaced_mc_ccc0.eqlForChaining(&bare_mc_ccc0));
```

Also retain positive cases proving ordinary leading, trailing, and repeated U+0020 values still normalize as before. Add a path-builder regression where the leaf issuer uses `" \u{0301}A"` and an anchor subject uses `"\u{0301}A"`; it must return `NoCandidatePath`, not a false candidate path.

tools/generate_rfc4518_data.py
```diff
@@ -0,0 +1,230 @@
...
+    if map_to_space(cp):
+        return [0x20]
+    if map_to_nothing(cp):
+        return []
+    ch = chr(cp)
+    if stringprep.in_table_a1(ch):
+        return [cp]
+    return chars_to_scalars(stringprep.map_table_b2(ch))
+
+
```
Jaruso, line 271: **Blocking — the generator still uses the host Python Unicode version for B.2 mappings of Unicode-3.2-assigned input.**

The new A.1 guard correctly prevents an input that was unassigned in Unicode 3.2 from being modern-lowercased before rejection. However, `stringprep.map_table_b2()` is still unsafe as the source of the authoritative RFC 3454 B.2 table for **assigned** inputs.

CPython's generated `map_table_b3()` uses an exception dictionary and otherwise falls back to `code.lower()`. That fallback follows the running Python interpreter's modern Unicode data, not Unicode 3.2. Therefore the checked-in generator can add mappings for characters that were assigned in Unicode 3.2 but had no B.2 mapping in RFC 3454.

The current checked-in table demonstrates this:

```text
U+10A0 -> U+2D00
U+10A1 -> U+2D01
...
```

U+10A0 is assigned in Unicode 3.2. RFC 3454 B.2 contains no `10A0` mapping, so it must remain U+10A0 through the mapping step. U+2D00 falls inside RFC 3454 A.1's Unicode-3.2-unassigned range `2A6D7-2F7FF`. The runtime therefore accepts U+10A0 at the new pre-check, maps it to U+2D00, and rejects it at the post-normalization A.1 check. A valid Unicode-3.2 DirectoryString is incorrectly reported as `NamePreparationFailed`.

U+04C0 has the same class of failure: current Python lowercasing maps it to U+04CF, but RFC 3454 B.2 has no `04C0; 04CF` entry and U+04CF was unassigned in Unicode 3.2.

The fix must pin the **actual RFC 3454 B.2 mapping**, not merely guard one category of input around a host-version-dependent mapping function. Suitable approaches include:

1. Check in a small authoritative source file containing the RFC 3454 B.2 entries and have this script parse it.
2. Check in the Unicode 3.2 `CaseFolding-3.2.0.txt` source plus any explicitly generated RFC B.2 additional-folding data, and reproduce the RFC algorithm using only the pinned Unicode 3.2 database.
3. Generate a repository-controlled B.2 table once from the RFC and validate it with a committed digest/count and representative exact mappings.

Do **not** call current `str.lower()`—directly or indirectly—for fallback behavior. An input absent from the pinned B.2 table must map to itself.

Conceptually:

```python
RFC3454_B2: dict[int, tuple[int, ...]] = load_pinned_b2_table()

def rfc4518_map(cp: int) -> list[int]:
    if map_to_space(cp):
        return [0x20]
    if map_to_nothing(cp):
        return []
    if stringprep.in_table_a1(chr(cp)):
        return [cp]
    return list(RFC3454_B2.get(cp, (cp,)))
```

Add generation invariants that compare the entire emitted B.2 mapping against the pinned source, not only A.1 behavior. At minimum assert:

```python
assert rfc4518_map(0x10A0) == [0x10A0]
assert rfc4518_map(0x04C0) == [0x04C0]
assert rfc4518_map(0x00DF) == [0x0073, 0x0073]
```

Required Zig regressions:

```zig
// Both were assigned in Unicode 3.2 and have no RFC 3454 B.2 mapping.
_ = try parseNameRaw(... "\u{10A0}" ...);
_ = try parseNameRaw(... "\u{04C0}" ...);
```

Also assert that these values remain distinct from the modern lowercase code points and that regeneration removes the `0x10A0 -> 0x2D00` and `0x04C0 -> 0x04CF` entries. Keep the input and output A.1 checks; they are still valuable defense in depth after the table source is corrected.

tools/generate_rfc4518_data.py
```diff
@@ -0,0 +1,281 @@
...
+        decomp = nfkd_mapping(cp)
+        if decomp != [cp]:
+            nfkd_entries.append((cp, decomp))
+
+        ccc = UCD.combining(chr(cp))
+        if ccc != 0:
+            ccc_values.append((cp, ccc))
+        if UCD.category(chr(cp)) in ("Mn", "Mc", "Me"):
+            combining_mark_values.append(cp)
+
```
Jaruso, line 363: **Blocking — RFC 4518 says Appendix A is the definitive combining-mark table; deriving this from the Unicode 3.2 general category produces a different matching rule.**

RFC 4518 Appendix A is normative and explicitly states that its listed code points “are to be considered definitive for the purposes of implementation.” It is close to, but not identical to, `unicodedata.ucd_3_2_0`’s `Mn`/`Mc`/`Me` set.

The exact delta is:

```text
UCD 3.2 mark but NOT RFC 4518 Appendix A: U+05BD
RFC 4518 Appendix A but NOT UCD 3.2 mark: U+094E, U+094F
```

U+094E/U+094F are Unicode-3.2-unassigned and are already rejected, but U+05BD is assigned and accepted. That creates a real false negative:

```text
CN=" \u{05BD}A"
CN="\u{05BD}A"
```

Under RFC 4518, U+05BD is **not** a combining mark because it is absent from the definitive Appendix A table. The leading U+0020 is therefore a space “followed by no combining marks,” is insignificant, and the two values prepare equally. The generated table currently includes U+05BD because its UCD category is `Mn`, so `isInsignificantSpaceAt()` preserves that U+0020 and the values compare unequal. In the path builder this can again become `NoCandidatePath` for a conforming name pair.

Please generate `combining_marks` from the literal RFC 4518 Appendix A ranges, not from `UCD.category()`. A suitable generator shape is:

```python
RFC4518_COMBINING_MARK_RANGES = (
    (0x0300, 0x034F),
    (0x0360, 0x036F),
    # ...copy the complete normative Appendix A list...
    (0x1D1AA, 0x1D1AD),
)


def expand_ranges(ranges: tuple[tuple[int, int], ...]) -> list[int]:
    return [cp for first, last in ranges for cp in range(first, last + 1)]

combining_mark_values = expand_ranges(RFC4518_COMBINING_MARK_RANGES)
```

The ranges may instead live in a small checked-in data file, but their provenance should explicitly identify RFC 4518 Appendix A. Add invariants that make this intentional divergence visible:

```python
rfc_marks = set(expand_ranges(RFC4518_COMBINING_MARK_RANGES))
ucd_marks = {
    cp for cp in range(MAX_SCALAR + 1)
    if valid_scalar(cp) and UCD.category(chr(cp)) in ("Mn", "Mc", "Me")
}
assert ucd_marks - rfc_marks == {0x05BD}
assert rfc_marks - ucd_marks == {0x094E, 0x094F}
```

Required name-level regression:

```zig
const spaced = try x509.parseNameRaw(
    arena,
    try nameWithCnTag(arena, 0x0c, " \u{05BD}A"),
    .{},
);
const bare = try x509.parseNameRaw(
    arena,
    try nameWithCnTag(arena, 0x0c, "\u{05BD}A"),
    .{},
);
try testing.expect(spaced.eqlForChaining(&bare));
```

Also add a path-builder regression proving the spaced issuer successfully finds the bare-subject anchor. Keep the existing U+0301 and U+093E tests because both are present in the definitive Appendix A table.

After changing the source table, regenerate `src/pki/rfc4518_data.zig` and verify U+05BD is absent from `combining_marks`.

Jaruso (commented):
## Review verdict: one final standards blocker remains

The two findings from the previous pass are correctly addressed:

- B.3 fallback case mapping is now pinned to Unicode 3.2 data instead of the host Python runtime.
- SPACE followed by a combining mark is preserved, including CCC-zero marks, with both name-level and path-builder regressions.

I re-ran the complete issue #344 acceptance-criteria review and rechecked the RFC 4518 preparation pipeline. Candidate ordering, bounds, anchor handling, cycle/duplicate behavior, error taxonomy, policy separation, name preparation, allocation safety, and the previous security fixes all remain sound.

The remaining blocker is narrower: the new `combining_marks` table is derived from Unicode 3.2 general categories, but RFC 4518 Appendix A explicitly says its own list is normative and definitive. Those sets are not identical. In particular, Unicode 3.2 classifies U+05BD as `Mn`, but RFC 4518 Appendix A omits it. This changes whether a preceding U+0020 is insignificant and produces a valid issuer/subject false negative. I left an implementation-ready inline comment with the exact regression and expected generator structure.

Because this PR is authored by the authenticated account, GitHub will not allow a formal request-changes review; this COMMENT should be treated as blocking until the Appendix A table is used verbatim and the regression passes.

Jaruso (commented):
## Final review verdict: ready to merge

I re-reviewed the complete PR at `3f1e29a1587c5891eb69d161c47bc575402fb1e9`, including the latest Appendix A fix, the full RFC 4518 name-preparation path, all prior blocking findings, the path-builder behavior, issue #344's acceptance criteria, and CI run 680.

The final blocker is correctly resolved:

- The generator now carries the literal normative RFC 4518 Appendix A combining-mark ranges rather than deriving them from the Unicode 3.2 `Mn`/`Mc`/`Me` categories.
- The table correctly excludes U+05BD and includes U+094E/U+094F, with generation-time invariants protecting both divergence directions.
- The checked-in generated Zig table reflects those exact changes.
- Name-level and path-builder-level regressions verify the resulting chaining behavior.

I also rechecked the previously fixed areas:

- RFC 3454 B.2 full case mapping is pinned to Unicode 3.2 data rather than host-Python Unicode behavior.
- Unicode 3.2 unassigned input is rejected before mapping and after normalization.
- NFKC, one-to-none, and one-to-many transformations are represented correctly.
- Significant SPACE-plus-combining-mark sequences are preserved using the definitive RFC table.
- Canonical combining ordering is stable and has linear worst-case work.
- `Certificate.isSelfIssued()` uses chaining equality.
- Candidate ranking, fanout, visit-budget accounting, cycle detection, duplicate handling, anchor termination, and error taxonomy continue to satisfy #344.
- Construction remains policy-independent for the downstream #345 validator.

CI run 680 completed successfully. I found no remaining blocking or non-blocking code issues in this review. This is ready to merge.

GitHub does not permit the PR author to approve their own pull request, so this is submitted as a COMMENT review rather than APPROVE.

PR #443, opened 2026-07-16
tls: make PureZigRecordStream own the record handshake driver (#410)

Jaruso:
## Summary

`PureZigRecordStream` now **owns the shared TLS handshake driver** and progresses a real TLS 1.3 handshake inside `drive()` — completing the pure-Zig TLS-over-TCP handshake ownership required by #353. The successful socket-pair path no longer installs fabricated traffic secrets, hand-applies arbitrary events, or declares completion manually.

### Ownership model
The stream owns an `engine.Driver` (`RecordHandshakeDriver`) over a canonical record-mode `transport.Contract` (`events.EncryptionEpoch`, no transport params), plus explicit inbound/outbound record epochs, the negotiated ALPN/certificate state, and a pending fatal-alert. The concrete backend is **injected** (`initWithCarrierAndBackend`): a thin `RecordModeBackend` wrapper adapts the pure-Zig `Tls13Backend` (which lives in the `quic` module and is QUIC-shaped) into the record-mode contract, since `tls_core` cannot import `quic`. This is the "smallest wrapper needed"; production HTTP dispatch over it is out of scope (#356).

### `drive()` progression
Starts the driver once → flushes pending ciphertext → reads bounded carrier ciphertext → parses records incrementally → routes handshake plaintext to `driver.receiveOutcome(epoch, …)` at the **explicit** inbound epoch → applies every emitted event before the next driver call → seals handshake bytes → installs read/write secrets → applies epoch discards → captures ALPN/cert → transitions to application mode only on the real `handshake_complete` → serializes a pending fatal alert before closing → stops on would-block / budget / completion / terminal failure.

### Handshake events handled
`handshake_bytes`, `traffic_secret`, `alpn`, `certificate`, `discard_epoch`, `handshake_complete`, `fatal_alert` (`peer_transport_parameters` is QUIC-only and ignored).

### Event lifetime & atomicity
Each borrowed event batch is applied fully — copied, installed, or serialized — before any subsequent driver call resets the sink. An outbound-capacity **preflight** (`handshake_output_reserve = 3 × max_ciphertext_record_len`) guards each `start`/`receive` so a full batch serializes atomically instead of partially overwriting the still-borrowed sink. Traffic-secret bytes are installed immediately and never retained. `startOutcome`/`receiveOutcome` keep a terminal error and its emitted fatal alert observable together.

### Epoch rules
`read`/`write` epochs advance to `.handshake` when that direction's handshake secret installs, and to `.application` **only** at authenticated `handshake_complete` — not merely because an application key is installed (the peer's Finished is still handshake-epoch after this side installs its app read secret). Initial-epoch plaintext, handshake-protected, and application-protected records are routed by the explicit epoch; a partial record spanning a `.handshake` discard fails closed; the RFC 8446 §5.1 `0x0301` ClientHello and existing compatibility handling are preserved.

### Error / alert behavior
A terminal handshake failure is deferred: any emitted fatal alert is sealed into the normal bounded outbound queue and flushed to the peer (best-effort, nonblocking) before the stream latches the **preserved** failure. Alert-send failure never erases the underlying error; repeated operations return the stable terminal failure; `close_notify`, peer fatal alerts, malformed alerts, truncation, and clean EOF stay distinct.

## Test plan
- [x] `zig fmt --check build.zig src/ tests/` — clean
- [x] `git diff --check` — clean
- [x] `zig build test-tls` — 135/135 (adds driver-owned duplex handshake across a fragmentation matrix + a fatal-alert/latched-failure path, all existing low-level record tests still green)
- [x] `zig build test-quic` — 413/413 (adds the real-backend socket-pair suite)
- [x] `zig build test-integration` — 65/69 (4 pre-existing skips, exit 0)
- [x] `zig build test` — 1414 passed, 1 skipped
- [x] Fragmented socket-pair scenarios run 5× — deterministic (exit 0 each)

### Socket-pair scenarios (real `Tls13Backend` on both ends)
**Success:** client produces the initial flight via `drive()`; server processes it in record mode; fragmentation matrix `{1,1}`,`{2,3}`,`{3,2}`,`{5,5}`,`{7,64}`,`{64,7}`, whole-record (covers split initial headers, fragmented/coalesced records, tiny reads & partial writes, would-block between flights); both install genuine derived 1-RTT secrets; both complete with **no `establish()`**; ALPN `h3` + valid certificate captured; bidirectional application plaintext; orderly `close_notify` shutdown.
**Failure:** in-flight ClientHello tamper (transcript divergence → `AuthenticationFailed`); carrier EOF before `close_notify` → `TruncatedStream`. The tls_core scripted-backend test additionally proves the fatal-alert flush + peer-alert + latched terminal error deterministically.

### Cleanup validation
Driver teardown runs exactly once (`teardownDriver` → `Driver.deinit` wipes the borrowed sink + backend); the wrapper securely zeroes its private QUIC sink; the bridge wipes all key material; ALPN/epoch metadata is cleared on both `deinit` and fatal `fail`. Owned carriers close exactly once and borrowed carriers are never closed by the stream (existing + new suites). Testing allocator + FD tracking pass on success, failure, and truncation paths.

## Scope
Focused on #410. No PKI files (#345/#346), HTTP dispatch (#356), watermarks (#355), KeyUpdate (#357), or QUIC handshake changes. Files changed: `src/tls/encrypted_stream.zig`, `src/quic/record_mode_handshake_test.zig`.

## #353
This clears the remaining #353 handshake-ownership acceptance criteria (real handshake + bidirectional app data over a fragmented nonblocking socket pair; no fabricated secrets on the success path; would-block without busy-loop; cleanup on success/failure/cancellation). #353 also spans the OpenSSL/pure-Zig abstraction unification and HTTP-startup interfaces (#356), so it is intentionally **not** closed here. The exact prerequisite now cleared for **#355** (watermarks/backpressure): a driver-owned `drive()` with bounded outbound serialization and stable would-block/readiness semantics for HTTP backpressure to hook into.

Closes #410
Refs #353

### DIFF
--- a/src/quic/record_mode_handshake_test.zig
+++ b/src/quic/record_mode_handshake_test.zig
@@ -510,6 +510,7 @@ const RecordModeBackend = struct {
         const result = self.engine.backend().transport.start(role, dummy_params, &self.scratch);
         try translate(&self.scratch, sink);
         result catch |err| return mapError(err);
+        if (self.deferredPolicyError()) |err| return err;
     }
 
     fn receive(ptr: *anyopaque, epoch: events.EncryptionEpoch, bytes: []const u8, sink: *es.RecordTransport.EventSink) es.RecordHandshakeError!void {
@@ -518,6 +519,24 @@ const RecordModeBackend = struct {
         const result = self.engine.backend().transport.receive(toLevel(epoch), bytes, &self.scratch);
         try translate(&self.scratch, sink);
         result catch |err| return mapError(err);
+        if (self.deferredPolicyError()) |err| return err;
+    }
+
+    /// The concrete engine reports ALPN/certificate policy results as events and
+    /// marks its own core failed while returning success -- it expects its
+    /// driver to convert those into terminal errors (the QUIC driver does the
+    /// same via `AlpnMismatch`/`CertificateInvalid`). Surface that here so the
+    /// record handshake fails closed instead of stalling with the peer's
+    /// `Finished` still buffered and no further record ever arriving.
+    fn deferredPolicyError(self: *RecordModeBackend) ?es.RecordHandshakeError {
+        if (self.engine.core.handshake_lifecycle != .failed) return null;
+        for (self.scratch.items[0..self.scratch.len]) |event| {
+            if (event == .certificate and event.certificate == .invalid) return error.CertificateInvalid;
+        }
+        for (self.scratch.items[0..self.scratch.len]) |event| {
+            if (event == .alpn) return error.AlpnMismatch;
+        }
+        return error.UnexpectedHandshakeMessage;
     }
 
     /// Securely wipe the last batch's copied secrets from the private sink. The
@@ -694,6 +713,7 @@ const FdCarrier = struct {
 /// keep stable pointers.
 const SocketHarness = struct {
     fds: [2]std.posix.fd_t,
+    fds_closed: [2]bool,
     client_engine: tls_backend.Tls13Backend,
     server_engine: tls_backend.Tls13Backend,
     client_wrapper: RecordModeBackend = undefined,
@@ -703,24 +723,45 @@ const SocketHarness = struct {
     client: es.PureZigRecordStream = undefined,
     server: es.PureZigRecordStream = undefined,
 
-    fn create(client_chunk: usize, server_chunk: usize) !*SocketHarness {
+    const Options = struct {
+        client_chunk: usize = std.math.maxInt(usize),
+        server_chunk: usize = std.math.maxInt(usize),
+        client_trust: tls_backend.Trust = .{ .pinned_certificate = tls_backend.testdata.certificate_der },
+        client_alpn: []const u8 = "h3",
+        server_alpn: []const u8 = "h3",
+    };
+
+    fn create(opts: Options) !*SocketHarness {
         const self = try std.testing.allocator.create(SocketHarness);
         errdefer std.testing.allocator.destroy(self);
         self.fds = try testSocketPair();
+        self.fds_closed = .{ false, false };
 
-        self.client_engine = tls_backend.Tls13Backend.initClient(clientEntropy(), .{ .pinned_certificate = tls_backend.testdata.certificate_der });
+        self.client_engine = tls_backend.Tls13Backend.initClient(clientEntropy(), opts.client_trust);
         self.client_engine.omit_transport_parameters = true;
+        self.client_engine.alpn = opts.client_alpn;
         self.server_engine = tls_backend.Tls13Backend.initServer(serverEntropy(), fixtureIdentity());
         self.server_engine.omit_transport_parameters = true;
+        self.server_engine.alpn = opts.server_alpn;
 
         self.client_wrapper = RecordModeBackend.init(&self.client_engine);
         self.server_wrapper = RecordModeBackend.init(&self.server_engine);
-        self.client_carrier = .{ .fd = self.fds[0], .max_chunk = client_chunk };
-        self.server_carrier = .{ .fd = self.fds[1], .max_chunk = server_chunk };
+        self.client_carrier = .{ .fd = self.fds[0], .max_chunk = opts.client_chunk };
+        self.server_carrier = .{ .fd = self.fds[1], .max_chunk = opts.server_chunk };
 
         self.client = es.PureZigRecordStream.initWithCarrierAndBackend(.client, clientProvider(), suite, self.client_carrier.carrier(), self.client_wrapper.backend());
         self.server = es.PureZigRecordStream.initWithCarrierAndBackend(.server, serverProvider(), suite, self.server_carrier.carrier(), self.server_wrapper.backend());
         return self;
+    }
+
+    /// Close one endpoint exactly once (idempotent), so a test can close an
+    /// endpoint early to model peer EOF without `destroy` double-closing a
+    /// potentially-recycled descriptor.
+    fn closeEndpoint(self: *SocketHarness, index: usize) void {
+        if (!self.fds_closed[index]) {
+            closeFd(self.fds[index]);
+            self.fds_closed[index] = true;
+        }
     }
 
     fn destroy(self: *SocketHarness) void {
@@ -728,8 +769,8 @@ const SocketHarness = struct {
         self.server.deinit();
         self.client_engine.deinit();
         self.server_engine.deinit();
-        closeFd(self.fds[0]);
-        closeFd(self.fds[1]);
+        self.closeEndpoint(0);
+        self.closeEndpoint(1);
         std.testing.allocator.destroy(self);
     }
 
@@ -765,7 +806,7 @@ test "record stream completes a real TLS 1.3 handshake over a nonblocking socket
         .{ record_codec.max_ciphertext_record_len, record_codec.max_ciphertext_record_len },
     };
     for (chunks) |chunk| {
-        const h = try SocketHarness.create(chunk[0], chunk[1]);
+        const h = try SocketHarness.create(.{ .client_chunk = chunk[0], .server_chunk = chunk[1] });
         defer h.destroy();
 
         try h.driveUntil(SocketHarness.bothComplete);
@@ -815,7 +856,7 @@ test "record stream completes a real TLS 1.3 handshake over a nonblocking socket
 }
 
 test "record stream handshake fails closed when a ClientHello is corrupted in flight" {
-    const h = try SocketHarness.create(std.math.maxInt(usize), std.math.maxInt(usize));
+    const h = try SocketHarness.create(.{});
     defer h.destroy();
     // Flip a byte inside the ClientHello's 32-byte random (past the 5-byte
     // record header, 4-byte handshake header, and 2-byte legacy_version) as the
@@ -851,13 +892,14 @@ test "record stream handshake fails closed when a ClientHello is corrupted in fl
 }
 
 test "record stream handshake treats carrier EOF before close_notify as truncation" {
-    const h = try SocketHarness.create(std.math.maxInt(usize), std.math.maxInt(usize));
+    const h = try SocketHarness.create(.{});
     defer h.destroy();
 
     // Complete the handshake, then the server abruptly closes its socket
-    // instead of sending close_notify.
+    // instead of sending close_notify. `closeEndpoint` makes the harness the
+    // sole owner of each descriptor, so `destroy` never double-closes it.
     try h.driveUntil(SocketHarness.bothComplete);
-    closeFd(h.fds[1]);
+    h.closeEndpoint(1);
     h.server.stream().close();
 
     var client_error: ?anyerror = null;
@@ -870,4 +912,63 @@ test "record stream handshake treats carrier EOF before close_notify as truncati
     }
     try std.testing.expectEqual(@as(?anyerror, error.TruncatedStream), client_error);
     try std.testing.expect(h.client.lifecycle == .failed);
+}
+
+/// Drive both streams until one returns a terminal error or progress stalls,
+/// returning the first error observed (client checked before server each round).
+fn driveUntilError(h: *SocketHarness) ?anyerror {
+    var rounds: usize = 0;
+    while (rounds < 500) : (rounds += 1) {
+        const c = h.client.stream().drive() catch |err| return err;
+        const s = h.server.stream().drive() catch |err| return err;
+        if (h.client.lifecycle == .failed or h.server.lifecycle == .failed) {
+            // Give the failing side one more drive to surface its latched error.
+            _ = h.client.stream().drive() catch |err| return err;
+            _ = h.server.stream().drive() catch |err| return err;
+            return null;
+        }
+        if (!c.made_progress and !s.made_progress) return null;
+    }
+    return null;
+}
+
+test "record stream handshake fails closed with CertificateInvalid on a wrong pinned certificate" {
+    // The client pins a certificate that does not byte-equal the one the server
+    // presents. Proof-of-possession still checks out, so the engine emits
+    // `.certificate(.invalid)`, marks its core failed, and returns success --
+    // record mode must convert that into a terminal `CertificateInvalid` rather
+    // than stalling with the server `Finished` still buffered.
+    var wrong_pin: [tls_backend.testdata.certificate_der.len]u8 = undefined;
+    @memcpy(&wrong_pin, tls_backend.testdata.certificate_der);
+    wrong_pin[wrong_pin.len / 2] ^= 0xff;
+
+    const h = try SocketHarness.create(.{ .client_trust = .{ .pinned_certificate = &wrong_pin } });
+    defer h.destroy();
+
+    const failure = driveUntilError(h);
+    try std.testing.expect(!h.client.bridge.handshake_complete);
+    try std.testing.expect(!h.server.bridge.handshake_complete);
+    try std.testing.expect(h.client.lifecycle == .failed);
+    try std.testing.expect(failure != null);
+    try std.testing.expect(failure.? == error.CertificateInvalid);
+    // A repeated drive returns the stable, latched terminal error, and the
+    // fatal failure wiped the captured negotiation metadata.
+    try std.testing.expectError(error.CertificateInvalid, h.client.stream().drive());
+    try std.testing.expectEqual(events.CertificateState.not_checked, h.client.certificateState());
+}
+
+test "record stream handshake fails closed with AlpnMismatch when the server selects a different protocol" {
+    // The client offers h3; the server supports only h2. The engine reports the
+    // offered protocol, marks its core failed, and returns success, deferring
+    // the AlpnMismatch decision to record mode.
+    const h = try SocketHarness.create(.{ .server_alpn = "h2" });
+    defer h.destroy();
+
+    const failure = driveUntilError(h);
+    try std.testing.expect(!h.client.bridge.handshake_complete);
+    try std.testing.expect(!h.server.bridge.handshake_complete);
+    try std.testing.expect(h.server.lifecycle == .failed);
+    try std.testing.expect(failure != null);
+    try std.testing.expect(failure.? == error.AlpnMismatch);
+    try std.testing.expectError(error.AlpnMismatch, h.server.stream().drive());
 }
--- a/src/quic/record_mode_handshake_test.zig
+++ b/src/quic/record_mode_handshake_test.zig
@@ -761,6 +761,7 @@ const SocketHarness = struct {
 
         self.client = es.PureZigRecordStream.initWithCarrierAndBackend(.client, clientProvider(), suite, self.client_carrier.carrier(), self.client_wrapper.backend());
         self.server = es.PureZigRecordStream.initWithCarrierAndBackend(.server, serverProvider(), suite, self.server_carrier.carrier(), self.server_wrapper.backend());
+        self.client.setExpectedAlpn(opts.client_alpn) catch unreachable;
         return self;
     }
 
--- a/src/quic/record_mode_handshake_test.zig
+++ b/src/quic/record_mode_handshake_test.zig
@@ -680,9 +680,15 @@ const FdCarrier = struct {
     max_chunk: usize = std.math.maxInt(usize),
     read_offset: usize = 0,
     corrupt_at: ?usize = null,
+    one_write_per_drive: bool = false,
+    write_armed: bool = true,
 
     fn carrier(self: *FdCarrier) es.Carrier {
         return .{ .ptr = self, .readFn = read, .writeFn = write };
+    }
+
+    fn rearmWrite(self: *FdCarrier) void {
+        if (self.one_write_per_drive) self.write_armed = true;
     }
 
     fn read(ptr: *anyopaque, out: []u8) es.Error!usize {
@@ -702,9 +708,12 @@ const FdCarrier = struct {
 
     fn write(ptr: *anyopaque, bytes: []const u8) es.Error!usize {
         const self: *FdCarrier = @ptrCast(@alignCast(ptr));
+        if (self.one_write_per_drive and !self.write_armed) return error.WouldBlock;
         const cap = @min(bytes.len, self.max_chunk);
         if (cap == 0) return error.WouldBlock;
-        return writeFd(self.fd, bytes[0..cap]);
+        const written = try writeFd(self.fd, bytes[0..cap]);
+        if (self.one_write_per_drive and written > 0) self.write_armed = false;
+        return written;
     }
 };
 
@@ -729,6 +738,7 @@ const SocketHarness = struct {
         client_trust: tls_backend.Trust = .{ .pinned_certificate = tls_backend.testdata.certificate_der },
         client_alpn: []const u8 = "h3",
         server_alpn: []const u8 = "h3",
+        one_write_per_drive: bool = false,
     };
 
     fn create(opts: Options) !*SocketHarness {
@@ -746,8 +756,8 @@ const SocketHarness = struct {
 
         self.client_wrapper = RecordModeBackend.init(&self.client_engine);
         self.server_wrapper = RecordModeBackend.init(&self.server_engine);
-        self.client_carrier = .{ .fd = self.fds[0], .max_chunk = opts.client_chunk };
-        self.server_carrier = .{ .fd = self.fds[1], .max_chunk = opts.server_chunk };
+        self.client_carrier = .{ .fd = self.fds[0], .max_chunk = opts.client_chunk, .one_write_per_drive = opts.one_write_per_drive };
+        self.server_carrier = .{ .fd = self.fds[1], .max_chunk = opts.server_chunk, .one_write_per_drive = opts.one_write_per_drive };
 
         self.client = es.PureZigRecordStream.initWithCarrierAndBackend(.client, clientProvider(), suite, self.client_carrier.carrier(), self.client_wrapper.backend());
         self.server = es.PureZigRecordStream.initWithCarrierAndBackend(.server, serverProvider(), suite, self.server_carrier.carrier(), self.server_wrapper.backend());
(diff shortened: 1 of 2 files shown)


src/tls/encrypted_stream.zig
```diff
@@ -261,24 +383,153 @@ pub const PureZigRecordStream = struct {
...
+                    var record_buf: [record_codec.max_ciphertext_record_len]u8 = undefined;
+                    const record = self.bridge.sealHandshake(hb.epoch, hb.data, &record_buf) catch |err| return self.fail(err);
+                    self.outbound_ciphertext.append(record) catch |err| return self.fail(err);
+                },
+                .traffic_secret => |ts| {
+                    self.bridge.installTrafficSecret(ts.epoch, ts.direction, ts.data) catch |err| return self.fail(err);
+                    self.advanceEpochOnSecret(ts.epoch, ts.direction);
+                },
+                .alpn => |protocol| self.captureAlpn(protocol),
+                .certificate => |cert_state| self.certificate_state = cert_state,
```
Jaruso: **Blocking: `.certificate(.invalid)` is currently treated as metadata instead of a handshake failure.**

The concrete `Tls13Backend` deliberately emits the certificate state and returns success. On an invalid certificate it sets its internal handshake lifecycle to failed; its own comment says the *driver* must convert the event into `CertificateInvalid`. The QUIC driver does exactly that. Here, record mode only stores the enum.

With the real backend, a wrong certificate pin therefore behaves like this:

1. `CertificateVerify` produces `.certificate(.invalid)`.
2. This function stores `.invalid` and returns success.
3. The backend stops consuming its coalesced handshake input with `Finished` still buffered.
4. No further record arrives to trigger another backend call.
5. The stream remains in `.handshaking`/`wants_read` forever instead of failing closed.

A generic injected backend could be even worse and emit `.certificate(.invalid)` followed by `.handshake_complete`, which this loop would currently accept and open.

At minimum, apply the event as a terminal failure and stop processing later success events in the batch:

```zig
.certificate => |cert_state| {
    self.certificate_state = cert_state;
    if (cert_state == .invalid) {
        self.deferHandshakeFailure(error.CertificateInvalid, null);
        break;
    }
},
```

If `.not_checked` is supported, make that an explicit construction/policy option rather than accidentally treating every non-invalid value as acceptable. `handshake_complete` should also validate the final client certificate policy before opening.

Please add a real-backend test using a deliberately wrong pinned certificate. It should prove that neither side reaches open, `drive()` returns a stable `CertificateInvalid`, sensitive state is wiped, and carrier/backend teardown runs exactly once.

There is an analogous ALPN hole: on server-side mismatch, `Tls13Backend` emits the offered ALPN, marks its core failed, and returns success because its transport driver is expected to return `AlpnMismatch`. Record mode needs an explicit expected-ALPN policy (or the record adapter must return `AlpnMismatch`) so that path terminates instead of stalling.

src/tls/encrypted_stream.zig
```diff
@@ -476,6 +732,14 @@ pub const PureZigRecordStream = struct {
             if (try self.processCarrierInput(drive_record_budget)) made_progress = true;
         }
 
+        // A terminal handshake failure surfaced while routing records: flush the
+        // fatal alert the backend emitted (best effort, nonblocking) so the peer
+        // learns why, then latch the preserved failure and close.
+        if (self.pending_terminal) |err| {
+            _ = self.flushPendingAlert();
```
Jaruso: **Blocking: a normal `WouldBlock` or incomplete alert write drops the pending fatal alert.**

`flushPendingAlert()` is attempted once here, but its result is ignored and `fail()` is called immediately. `fail()` clears `outbound_ciphertext` and closes the carrier. Therefore all of these ordinary nonblocking cases lose the alert:

- the first alert write returns `WouldBlock`;
- the carrier accepts only part of the queued output and then blocks;
- the backend emitted handshake bytes before its fatal alert, so the alert is behind earlier queued records;
- the peer closes while the alert is pending.

This directly conflicts with the permanent #354 rule: record processing may queue a typed alert, but it must expose write readiness and apply a bounded flush/deadline policy; it must not synchronously retry-and-discard from the read progression path.

Please model `pending_terminal` as a real state:

```zig
fn readiness(self: *const PureZigRecordStream) Readiness {
    if (self.pending_terminal != null) return .{
        .wants_read = false,
        .wants_write = self.outbound_ciphertext.len != 0,
        .can_read_plaintext = false,
        .can_write_plaintext = false,
        .peer_closed = self.peer_closed,
    };
    // existing readiness...
}
```

After record processing first creates the pending failure, return a normal `DriveResult` without another hidden socket write. On later writable progression, flush only the bounded write budget, preserve the original handshake error, and latch it only after the queued terminal output drains (or an explicit bounded abort/deadline policy fires):

```zig
if (self.pending_terminal) |root_error| {
    const wrote = self.flushOutboundBudgetPreserving(root_error);
    made_progress = made_progress or wrote;

    if (self.outbound_ciphertext.len != 0) {
        return .{ .made_progress = made_progress, .readiness = self.readiness() };
    }
    return self.fail(root_error);
}
```

The flush helper must not replace `root_error` with a carrier-write error. Please add deterministic tests for: first write `WouldBlock` then succeeds; one-byte/partial alert writes; handshake bytes preceding the fatal alert; peer close during pending alert; and cleanup when the alert can never be transmitted.

src/quic/record_mode_handshake_test.zig
```diff
@@ -464,3 +464,410 @@ test "a tampered application record fails closed and cleanup still wipes secrets
...
+}
+
+test "record stream handshake treats carrier EOF before close_notify as truncation" {
+    const h = try SocketHarness.create(std.math.maxInt(usize), std.math.maxInt(usize));
+    defer h.destroy();
+
+    // Complete the handshake, then the server abruptly closes its socket
+    // instead of sending close_notify.
+    try h.driveUntil(SocketHarness.bothComplete);
+    closeFd(h.fds[1]);
```
Jaruso: **This FD is closed twice.**

The test closes `h.fds[1]` here, then the deferred `h.destroy()` unconditionally calls `closeFd` on both numeric descriptors again. Besides invalidating the PR's exact-once cleanup claim, a second close on a recycled descriptor can close an unrelated resource.

Please make endpoint ownership one-shot, for example:

```zig
fds: [2]?std.posix.fd_t,

fn closeEndpoint(self: *SocketHarness, index: usize) void {
    if (self.fds[index]) |fd| {
        closeFd(fd);
        self.fds[index] = null;
    }
}

fn destroy(self: *SocketHarness) void {
    self.client.deinit();
    self.server.deinit();
    self.client_engine.deinit();
    self.server_engine.deinit();
    self.closeEndpoint(0);
    self.closeEndpoint(1);
    std.testing.allocator.destroy(self);
}
```

Then call `h.closeEndpoint(1)` here. Alternatively, use `shutdown(..., SHUT_WR)` to produce peer EOF while leaving the harness as the sole closer. The latter often models abrupt one-directional EOF more cleanly.

Jaruso (commented):
## Blocking review — do not merge yet

I traced this PR back through #410, the parent #353 contract, the completed #354 alert semantics, and the surrounding `engine.Driver`, `transport.Contract`, `record_epoch_bridge`, and concrete `Tls13Backend` implementations. I also inspected both changed files in full and checked CI run 685, which is green.

The main success path is strong:

- ✅ A real pure-Zig TLS 1.3 client/server handshake is driven exclusively through `PureZigRecordStream.drive()`.
- ✅ The socket-pair fragmentation matrix covers split headers, partial writes, asymmetric chunking, and coalesced records.
- ✅ The success test does not call `establish()`, manually install secrets, or manually declare completion.
- ✅ Bidirectional application data and orderly `close_notify` are exercised.
- ✅ The explicit handshake/application epoch transition is correctly designed for the client-Finished ordering case.

However, three correctness gaps remain blocking:

1. **Certificate and ALPN failure events are not converted into terminal stream failures.** The concrete backend intentionally relies on its transport driver to enforce these policy events; record mode currently only records them. An invalid certificate can therefore leave the stream permanently handshaking instead of returning `CertificateInvalid`, and ALPN mismatch has the analogous failure mode.
2. **A pending fatal alert is discarded on normal nonblocking backpressure.** The implementation retries output synchronously and then calls `fail()` even after `WouldBlock` or an incomplete write, which clears the queued alert. This contradicts the permanent #354 invariant that read progression queues the alert, requests write readiness, and applies a bounded retry/deadline policy without obscuring the root error.
3. **The new truncation test double-closes an FD, and the driver-owned cancellation/owned-carrier cleanup AC is still not actually proven.** The real socket harness uses borrowed carriers, while the existing owned-carrier cancellation tests do not install the new handshake driver/backend.

Please also address Copilot's existing ALPN-length thread; silently truncating an impossible protocol name is not a safe fail-closed behavior.

### Required cancellation/teardown coverage

Please add a driver-owned cancellation test—not another `initWithCarrier` plumbing-only test—with all of these assertions:

```zig
var backend = CountingRecordBackend{
    // start emits at least one handshake traffic secret and handshake bytes
};
var carrier = CountingOwnedCarrier{};
var stream = PureZigRecordStream.initWithCarrierAndBackend(
    .client,
    cp,
    .tls_aes_128_gcm_sha256,
    carrier.carrier(),
    backend.recordBackend(),
);

_ = try stream.stream().drive(); // driver started; sensitive state exists
stream.stream().close();
_ = try stream.stream().drive();
stream.deinit();

try testing.expectEqual(@as(usize, 1), carrier.close_count);
try testing.expectEqual(@as(usize, 1), backend.deinit_count);
try testing.expect(!stream.bridge.hasReadKeys(.handshake));
try testing.expect(!stream.bridge.hasWriteKeys(.handshake));
// Also assert the driver's/backend's used secret scratch was zeroed.
```

Cover both cancellation before the first flight drains and cancellation after handshake keys have been installed. That closes the exact #410/#353 cleanup requirement for the newly introduced ownership path.

Because the authenticated account is also the PR author, GitHub does not permit me to submit `REQUEST_CHANGES`; this `COMMENT` review should be treated as blocking.

src/tls/encrypted_stream.zig
```diff
@@ -261,24 +397,178 @@ pub const PureZigRecordStream = struct {
...
+        self.alpn_captured = true;
+    }
+
+    /// Record a terminal handshake failure without immediately latching, first
+    /// queuing any emitted fatal alert into the normal bounded outbound queue.
+    /// `drive()` then flushes that output within a bounded write budget and
+    /// latches the preserved error only once it drains (or the flush deadline
+    /// fires) -- never by a hidden synchronous retry-and-discard.
+    fn deferHandshakeFailure(self: *PureZigRecordStream, err: Error, maybe_alert: ?alerts.AlertDescription) void {
+        if (maybe_alert) |desc| self.queueFatalAlert(desc);
```
Jaruso: **Blocking: terminal handshake errors from the real backend still close without sending their mapped TLS alert.**

`deferHandshakeFailure` only queues `maybe_alert`, but the real `Tls13Backend` does not emit `.fatal_alert` when its driver converts certificate or ALPN policy events into `CertificateInvalid` / `AlpnMismatch`. The new wrapper therefore returns the correct local error, while this function leaves the outbound queue empty. The peer receives an abrupt close/truncation instead of `bad_certificate` or `no_application_protocol`.

The repository already has the canonical transport-policy mapping in `alerts.fromHandshakeError`:

```zig
MalformedHandshake          -> decode_error
IllegalParameter            -> illegal_parameter
UnexpectedHandshakeMessage  -> unexpected_message
CertificateInvalid          -> bad_certificate
AlpnMismatch                -> no_application_protocol
SecretExportFailed          -> internal_error
InvalidHandshakeState       -> internal_error
```

Record mode should synthesize that alert whenever the backend did not already emit a more specific `.fatal_alert`:

```zig
fn mappedFatalAlert(err: Error) ?alerts.AlertDescription {
    return switch (err) {
        error.MalformedHandshake,
        error.IllegalParameter,
        error.UnexpectedHandshakeMessage,
        error.CertificateInvalid,
        error.AlpnMismatch,
        error.SecretExportFailed,
        error.InvalidHandshakeState,
        => alerts.fromHandshakeError(@errorCast(err)),
        else => null,
    };
}

fn deferHandshakeFailure(
    self: *PureZigRecordStream,
    err: Error,
    emitted_alert: ?alerts.AlertDescription,
) void {
    const alert = emitted_alert orelse mappedFatalAlert(err);
    if (alert) |desc| self.queueFatalAlert(desc);
    self.pending_terminal = err;
    self.terminal_flush_attempts = 0;
}
```

Please extend both real-backend policy tests to drive the peer as well:

- wrong pin: client returns stable `CertificateInvalid`, server opens a fatal `bad_certificate` alert and returns `PeerFatalAlert`;
- ALPN mismatch: server returns stable `AlpnMismatch`, client receives fatal `no_application_protocol` and returns `PeerFatalAlert`.

Those tests should include blocked/partial output so the synthesized-alert path uses the same terminal flushing state machine.

src/tls/encrypted_stream.zig
```diff
@@ -392,11 +690,33 @@ pub const PureZigRecordStream = struct {
...
+        // work. Flush its queued output (the fatal alert, behind any earlier
+        // handshake bytes it sits after) within the bounded write budget while
+        // preserving the root error, and latch only once the queue drains or
+        // the bounded flush deadline fires. This is the sole socket write on
+        // this path -- readiness advertises `wants_write` so the event loop
+        // drives us back when the carrier is writable, rather than a hidden
+        // synchronous retry that would discard the alert on `WouldBlock`.
+        if (self.pending_terminal) |root| {
+            if (self.flushPendingAlert() > 0) made_progress = true;
+            self.terminal_flush_attempts += 1;
```
Jaruso: **Blocking: this deadline advances even when the alert is making legitimate partial-write progress.**

`flushPendingAlert()` loops until the carrier blocks, but a normal nonblocking carrier can accept a small prefix and then return `WouldBlock` until the next writable event. This code increments `terminal_flush_attempts` once per drive regardless of whether bytes were written.

A handshake-protected alert is larger than 16 bytes. A carrier that accepts one byte per writable event therefore reaches `max_terminal_flush_attempts` with part of the alert still queued; `fail(root)` then clears the remaining ciphertext. The same failure is even easier when handshake bytes precede the fatal alert in the queue. This is exactly the partial-write case the previous review asked to cover.

Count **consecutive no-progress attempts**, not all drives:

```zig
const flushed = self.flushPendingAlert();
if (flushed > 0) {
    made_progress = true;
    self.terminal_flush_attempts = 0;
} else {
    self.terminal_flush_attempts += 1;
}

if (self.outbound_ciphertext.len == 0 or
    self.terminal_flush_attempts >= max_terminal_flush_attempts)
{
    return self.fail(root);
}
```

The queue is bounded, so continued byte progress cannot wedge memory or make one `drive()` unbounded. The deadline still handles a permanently blocked/dead carrier.

Please add a deterministic carrier that is re-armed for one small write per `drive()` and then returns `WouldBlock`. Trigger the failure after handshake keys are installed so the fatal alert is protected and longer than 16 bytes; also cover an earlier handshake record ahead of it. Assert the entire queue drains, the peer receives the fatal alert, and the root error is not latched early.

src/tls/encrypted_stream.zig
```diff
@@ -261,24 +397,178 @@ pub const PureZigRecordStream = struct {
...
+                        self.deferHandshakeFailure(error.CertificateInvalid, fatal_alert);
+                        break;
+                    }
+                },
+                .discard_epoch => |epoch| {
+                    self.bridge.discardEpoch(epoch) catch |err| return self.fail(err);
+                    try self.applyDiscardSideEffects(epoch);
+                },
+                .handshake_complete => {
+                    self.bridge.markHandshakeComplete() catch |err| return self.fail(err);
```
Jaruso, line 555: **Blocking security policy gap: a client can still complete with `.certificate(.not_checked)` without explicit opt-in.**

The concrete backend's `Trust.insecure_no_verification` contract explicitly says it reports `not_checked` and completes **only when the driver explicitly opts into** `allow_unverified_certificate`. QUIC enforces that with a driver flag. Record mode now rejects `.invalid`, but this completion branch accepts `.not_checked`—or no certificate event at all—and opens the stream unconditionally.

Please add an explicit record-stream policy, defaulting to verified certificates for the client role:

```zig
role: tls_state.Role,
allow_unverified_certificate: bool = false,

fn validateCompletionPolicy(self: *PureZigRecordStream) Error!void {
    if (self.role == .client and
        !self.allow_unverified_certificate and
        self.certificate_state != .valid)
    {
        self.deferHandshakeFailure(error.CertificateInvalid, null);
        return error.CertificateInvalid; // or signal the caller to stop the batch
    }
}
```

Apply that check before `bridge.markHandshakeComplete()` and before setting `lifecycle = .open`. Do not require a certificate on the server role because the current profile does not authenticate clients.

Tests expected:

```zig
// Default policy: insecure backend does not silently open.
Tls13Backend.initClient(..., .insecure_no_verification)
// => CertificateInvalid, never handshake_complete/open

// Explicit opt-in: the same backend may complete, and certificateState() is not_checked.
stream.allow_unverified_certificate = true;
```

The scripted client handshake should either emit `.certificate(.valid)` or explicitly opt into unverified mode so the default remains secure.

Jaruso (commented):
## Second review — still blocking

I compared the new head (`6aef82d`) directly against the previously reviewed commit, re-read both changed files in full context, checked the shared `transport.Contract`, `engine.Driver`, `alerts`, concrete `Tls13Backend`, and the #410/#353/#354 requirements, and verified CI run 690 is green.

The previous findings are substantively addressed:

- ✅ invalid certificate and ALPN mismatch no longer stall silently;
- ✅ oversized ALPN values fail closed;
- ✅ fatal output now survives an initial `WouldBlock`;
- ✅ the socket harness no longer double-closes its FD;
- ✅ driver-owned cancellation now proves exact-once carrier/backend teardown and secret wiping.

Three issues remain blocking:

1. **Real-backend certificate/ALPN failures still do not send the required TLS fatal alert.** The wrapper now returns `CertificateInvalid`/`AlpnMismatch`, but the stream only queues an alert when the backend explicitly emitted `.fatal_alert`. The concrete backend does not emit one for these policy failures, despite the existing canonical `alerts.fromHandshakeError` mapping being intended for record transport policy.
2. **The flush deadline counts successful partial-write drives as failures.** A legitimate carrier that accepts a small prefix and then returns `WouldBlock` can hit 16 attempts while continually making progress, causing the stream to discard the remainder of a protected alert or earlier queued handshake bytes.
3. **`.certificate(.not_checked)` still completes without the explicit driver opt-in required by the concrete backend's trust contract.** Record mode now rejects `.invalid`, but it has no equivalent of QUIC's `allow_unverified_certificate` and opens unconditionally on `handshake_complete`.

Because the authenticated GitHub account is the PR author, GitHub will not accept a formal `REQUEST_CHANGES`; this comment review is blocking.

src/tls/encrypted_stream.zig
```diff
@@ -261,24 +406,206 @@ pub const PureZigRecordStream = struct {
...
+                .handshake_bytes => |hb| {
+                    var record_buf: [record_codec.max_ciphertext_record_len]u8 = undefined;
+                    const record = self.bridge.sealHandshake(hb.epoch, hb.data, &record_buf) catch |err| return self.fail(err);
+                    self.outbound_ciphertext.append(record) catch |err| return self.fail(err);
+                },
+                .traffic_secret => |ts| {
+                    self.bridge.installTrafficSecret(ts.epoch, ts.direction, ts.data) catch |err| return self.fail(err);
+                    self.advanceEpochOnSecret(ts.epoch, ts.direction);
+                },
+                .alpn => |protocol| try self.captureAlpn(protocol),
```
Jaruso: **Blocking: client-side ALPN policy is still not enforced.**

This branch only copies the selected protocol. The concrete `Tls13Backend` client does the same in `onEncryptedExtensions`: it emits whatever protocol the server selected, but does **not** compare it with `engine.alpn` or mark its core failed. `RecordModeBackend.deferredPolicyError()` therefore catches only the server-side no-overlap path, where the server itself set `handshake_lifecycle = .failed`.

A malicious/nonconforming server backend can consequently select `h2` after the client offered `h3`—or emit no ALPN event at all—and this client still reaches `.open`. The QUIC driver closes both holes by checking the ALPN event immediately and checking again at completion.

Record mode needs an explicit expected-ALPN policy, not only captured metadata. For example, copy the required protocol into stream-owned storage and validate both the event and completion:

```zig
expected_alpn_storage: [max_alpn_len]u8 = undefined,
expected_alpn_len: usize = 0,
require_alpn: bool = false,

pub fn setExpectedAlpn(self: *PureZigRecordStream, protocol: []const u8) Error!void {
    if (protocol.len == 0 or protocol.len > max_alpn_len) return error.MalformedHandshake;
    @memcpy(self.expected_alpn_storage[0..protocol.len], protocol);
    self.expected_alpn_len = protocol.len;
    self.require_alpn = true;
}

fn alpnPolicyError(self: *const PureZigRecordStream, selected: ?[]const u8) ?Error {
    if (!self.require_alpn) return null;
    const protocol = selected orelse return error.AlpnMismatch;
    if (!std.mem.eql(u8, protocol, self.expected_alpn_storage[0..self.expected_alpn_len])) {
        return error.AlpnMismatch;
    }
    return null;
}
```

On a mismatched `.alpn`, defer `AlpnMismatch` and stop the batch before applying later success events. At `handshake_complete`, fail if the required ALPN was never emitted. The mapped alert must be `no_application_protocol`.

Please add two adversarial backend tests: one selects an unoffered protocol and one omits ALPN, then otherwise emits normal application secrets/Finished/discard/completion. In both cases the client must never open or send Client Finished, must return stable `AlpnMismatch`, and the peer must receive the fatal alert.

src/tls/encrypted_stream.zig
```diff
@@ -261,24 +406,206 @@ pub const PureZigRecordStream = struct {
...
+                    {
+                        self.deferHandshakeFailure(error.CertificateInvalid, fatal_alert);
+                        break;
+                    }
+                },
+                .discard_epoch => |epoch| {
+                    self.bridge.discardEpoch(epoch) catch |err| return self.fail(err);
+                    try self.applyDiscardSideEffects(epoch);
+                },
+                .handshake_complete => {
```
Jaruso, line 554: **Blocking: this fallback authentication check occurs after the batch has already queued Client Finished and discarded the keys needed to send its failure alert.**

The canonical client completion batch emitted by `Tls13Backend.onServerFinished` is ordered as:

1. application write/read secrets;
2. Client Finished handshake bytes;
3. `.discard_epoch(.handshake)`;
4. `.handshake_complete`.

If an injected backend omits the certificate event, `certificate_state` remains `.not_checked`. This loop still applies steps 1–3 first: it installs application keys, queues Client Finished, and `bridge.discardEpoch(.handshake)` clears the handshake write keys. Only here does it detect the authentication failure.

`deferHandshakeFailure(CertificateInvalid, ...)` then maps `bad_certificate`, but `queueFatalAlert()` attempts to seal at `write_epoch == .handshake`; `sealProtected` returns `MissingWriteKeys`, which is silently swallowed. The pending-terminal flush therefore sends **only Client Finished**. The peer can mark the handshake complete and then see an abrupt EOF instead of the fatal alert.

Completion policy must be evaluated before applying any success effects from a batch containing `handshake_complete`. A preflight is the cleanest shape:

```zig
fn completionPolicyError(
    self: *const PureZigRecordStream,
    sink: *const RecordTransport.EventSink,
) ?Error {
    var cert = self.certificate_state;
    var selected_alpn: ?[]const u8 = self.negotiatedAlpn();
    var completes = false;

    for (sink.items[0..sink.len]) |event| switch (event) {
        .certificate => |state| cert = state,
        .alpn => |protocol| selected_alpn = protocol,
        .handshake_complete => completes = true,
        else => {},
    };

    if (!completes) return null;
    if (self.role == .client and !self.allow_unverified_certificate and cert != .valid) {
        return error.CertificateInvalid;
    }
    return self.alpnPolicyError(selected_alpn);
}

fn applyDriverOutcome(self: *PureZigRecordStream, outcome: RecordHandshakeDriver.Outcome) Error!void {
    if (self.completionPolicyError(outcome.sink)) |err| {
        self.deferHandshakeFailure(err, findEmittedFatalAlert(outcome.sink));
        return; // no Finished, discard, or completion is applied
    }
    // existing event application
}
```

A variation that stages completion-batch output until policy validation is also fine, but merely moving the check a few lines earlier than `markHandshakeComplete()` is insufficient—the Client Finished and discard events are earlier in the batch.

Please add a scripted client backend that emits application secrets + Client Finished + handshake discard + completion with **no certificate event**. Assert that the stream queues/sends `bad_certificate`, does not send Client Finished, the peer never reaches handshake complete, and repeated operations return stable `CertificateInvalid`.

Jaruso (commented):
## Blocking follow-up — two completion-policy defects remain

I re-reviewed the full third commit against #410/#353, the permanent #354 alert semantics, the concrete `Tls13Backend`, the QUIC driver's policy enforcement, and every previously unresolved thread. The three findings from the prior pass are now correctly addressed:

- ✅ Handshake errors synthesize their canonical TLS alerts when the backend does not emit one.
- ✅ The fatal-alert deadline counts consecutive zero-progress attempts, and the new one-byte carrier test proves partial writes can exceed the deadline count while draining fully.
- ✅ `.certificate(.not_checked)` requires explicit client opt-in, with real-backend strict/opt-in tests.
- ✅ CI run 691 is fully green.

Two policy-ordering gaps are still blocking:

1. **The client never validates that the server-selected ALPN was offered/expected, and it does not require an ALPN before completion.** The concrete client backend emits whatever it parses from EncryptedExtensions without marking its core failed; record mode only captures that value. This is distinct from the now-fixed server-side no-overlap path.
2. **The fallback certificate check runs after success output and handshake-key discard events from the same completion batch.** If a backend omits the certificate event, the stream queues Client Finished, discards the keys needed to seal `bad_certificate`, and only then detects `certificate_state != .valid`. It consequently flushes a success flight before failing and cannot send the mapped fatal alert.

Both need to be validated before applying completion-batch success effects. I included concrete implementation and regression-test shapes inline.

Because the authenticated GitHub account is also the PR author, GitHub does not permit a `REQUEST_CHANGES` review; this `COMMENT` remains blocking.

Jaruso (commented):
## Follow-up review — merge-ready

I re-reviewed the complete delta from `7ff2e224` through `453c3811`, rechecked the full event-ordering context in `PureZigRecordStream`, and verified the latest CI run.

Both remaining blocking findings are correctly addressed:

- Record mode now owns an explicit expected-ALPN policy, rejects an unoffered selection immediately, rejects missing ALPN at completion, and maps both paths to `no_application_protocol` without sending Client Finished.
- Completion policy is preflighted against the complete borrowed event batch before application secrets, Client Finished, handshake-key discard, or `handshake_complete` are applied. The missing-certificate path therefore retains handshake write keys long enough to send `bad_certificate`, and the peer never reaches completion.

The new adversarial scripted-backend tests directly prove mismatched ALPN, absent ALPN, and absent certificate behavior, including stable local errors, peer fatal-alert receipt, and no Client Finished. The real socket-pair harness now also configures expected ALPN consistently.

I also rechecked the previously fixed certificate policy, synthesized-alert mapping, partial-write terminal flushing, exact-once teardown, fragmentation, application-data exchange, cancellation, EOF, and close-notify paths. I found no additional correctness, security, lifecycle, or acceptance-criteria gaps.

CI run 693 completed successfully. Because this PR is authored by the authenticated account, GitHub does not permit a self-approval; this COMMENT records my approval-equivalent verdict.

**Verdict: good to merge.**

PR #444, opened 2026-07-16
pki: enforce bounded RFC 5280 Name Constraints

Jaruso:
## Summary

- add bounded RFC 5280 Name Constraints processing for `directoryName`, `dNSName`, `rfc822Name`, URI hosts, IPv4, and IPv6
- integrate constraint state into candidate-path validation without changing anchor-local policy, signature/time/BC/KU/EKU checks, or post-policy identity matching
- add signed synthetic coverage, fixed OpenSSL-generated certificates, and an opt-in out-of-process differential target

Refs #345

## Processing model

Permitted subtrees from one CA are stored as an OR group; every group introduced by a different superior CA must match, giving the RFC intersection semantics without constructing an exponentially growing explicit intersection. Excluded subtrees are one cumulative union and are checked first. State is separate per GeneralName form.

The path is processed anchor-side toward the leaf. A certificate is checked against inherited state before its own critical CA extension is incorporated, so a constraint never applies to its issuer certificate itself. Configured-anchor extensions remain local trust policy and are ignored. Self-issued intermediates skip inherited checking but still contribute constraints; the final target is always checked.

## Name behavior

- `directoryName`: canonical RDN-prefix matching reuses the existing RFC 4518/RFC 4517 normalization and RDN set semantics through per-RDN chaining keys.
- `dNSName`: case-insensitive label-boundary subtrees. For OpenSSL compatibility, `.example.com` matches one-or-more-label descendants but not the exact base.
- `rfc822Name`: exact mailbox compatibility uses a case-sensitive local part and case-insensitive host; exact-host and leading-dot domain constraints are supported. Legacy subject-DN `emailAddress` attributes are checked in addition to every SAN instance.
- URI: a bounded authority-only helper extracts a DNS host, accepts ports and userinfo, and rejects missing authorities, IP literals, bracketed IPv6 hosts, and malformed host/port forms when URI constraints are active. A plain constraint is an exact host; a leading-dot constraint is descendants-only.
- IP: parsed octets are compared against canonical CIDR-style address/mask values. Masks must be contiguous, host bits in the base must be zero, and IPv4/IPv6 families must match. `/0`, host masks, boundaries, adjacent subnets, and both families are covered.

No CN-to-DNS fallback, directory-component-to-DNS synthesis, issuerAltName processing, or first-name-wins behavior is introduced. A constrained form may be absent; when present, every instance must pass.

## Bounds and failures

Default configurable bounds:

- path length: 8
- permitted groups per form: 8
- permitted subtrees: 64
- excluded subtrees: 64
- names examined per certificate: 128
- comparisons per validation: 4096
- directoryName parses: 64
- RDNs per parsed directoryName: 32
- attributes per parsed RDN: 8
- URI length/parsing work: 2048 bytes

Stable validator reasons are `name_constraints_violation`, `name_constraints_unsupported`, and `name_constraints_resource_limit_exceeded` (with existing `out_of_memory`). Diagnostics carry the leaf-first certificate index, permitted/excluded source where applicable, GeneralName form, introducing certificate index, and the Name Constraints OID without copying peer names.

Noncritical CA constraints, constraints on leaves, unsupported constrained GeneralName forms, malformed typed constraints, non-contiguous/noncanonical IP masks, and bound exhaustion fail closed.

## Fixtures and differential results

Synthetic signed fixtures cover propagation, permitted intersection, excluded union, excluded-over-permitted, multiple names, absent forms, empty-subject SAN-only leaves, self-issued rollover, final self-issued targets, all supported forms, malformed inputs, bounds, allocation-failure cleanup, and alternate-candidate fallback.

The independent OpenSSL fixture matrix covers:

- permitted DNS success
- excluded/permitted overlap rejection
- IPv4 subnet success and adjacent-subnet rejection
- directoryName rejection
- constrained intermediates
- leading-dot subdomain success and exact-base rejection

`zig build test-pki-openssl` compares all seven Tardigrade decisions with OpenSSL 3.6.2; all match. The generation script keeps private keys in a temporary directory, and only public certificates are checked in. Self-issued rollover remains in the deterministic signed suite rather than the OpenSSL CA-database generator.

Intentional compatibility policy is documented beside the fixtures: noncritical constraints fail closed, exact-mailbox constraints remain supported for historical compatibility despite RFC 9549's issuance update, unsupported forms/masks fail closed, and internationalized hosts must already be ASCII A-label form (no U-label conversion or full IDNA validation here).

## Validation

- `zig build test-pki --summary all --error-style verbose` (163/163)
- `zig build test-pki-openssl --summary all --error-style verbose` (7/7 fixture decisions)
- `zig build test-crypto --summary all --error-style verbose` (43/43)
- `zig build test-tls --summary all --error-style verbose` (133/133)
- `zig build test --summary all --error-style verbose` (1439 passed, 1 existing skip)
- `git diff --check`

## Remaining #345 work

Certificate-policy processing remains deferred and is intentionally outside this slice: bounded policy-tree processing, policy mappings, and `inhibitAnyPolicy`/related RFC 9618-compatible behavior. Noncritical `certificatePolicies` still follows the existing implicit-any policy and critical policy extensions still fail closed. Issue #345 must remain open until that work is completed.


src/pki/name_constraints.zig
```diff
@@ -0,0 +1,678 @@
...
+fn validateDnsName(name: []const u8) NameError!void {
+    if (name.len == 0 or name.len > 253 or name[0] == '.' or name[name.len - 1] == '.') return error.Malformed;
+    var label_start: usize = 0;
+    for (name, 0..) |byte, index| {
+        if (byte == '.') {
+            try validateDnsLabel(name[label_start..index]);
+            label_start = index + 1;
+        } else if (!std.ascii.isAlphanumeric(byte) and byte != '-') {
+            return error.Malformed;
+        }
```
Jaruso, line 474: **Blocking: this rejects every wildcard DNS-ID whenever a dNSName constraint is active.**

Tardigrade's existing `identity.zig` profile deliberately accepts RFC 9525-style presented identifiers such as `*.example.com` (complete left-most label, exactly one label deep). Here, however, `preparePresented()` sends that same SAN through `validateDnsName()`, and `*` is rejected as malformed. As a result, a leaf that is valid for `verifyHost()` becomes a `name_constraints_violation` as soon as an intermediate contains any dNSName constraint, including a broad permitted subtree such as `example.com`.

Please handle presented wildcard patterns explicitly rather than merely allowing `*` through this ordinary hostname validator. The matching must use **set/intersection semantics**, because treating the pattern as an ordinary suffix can create a constraint bypass. In particular, regression coverage should establish at least:

```text
permitted example.com       + SAN *.example.com => accept
permitted .example.com      + SAN *.example.com => accept
permitted foo.example.com   + SAN *.example.com => reject
excluded foo.example.com    + SAN *.example.com => reject
excluded .foo.example.com   + SAN *.example.com => accept (the wildcard covers only foo.example.com, not descendants below it)
excluded .example.com       + SAN *.example.com => reject
```

Malformed patterns (`*`, `a.*.example.com`, `f*o.example.com`, `*.com`) must still fail closed. I strongly recommend representing a presented DNS pattern distinctly (or returning wildcard metadata from validation) so `checkPresented()` can apply dedicated permitted-subset and excluded-intersection logic. This exact edge is security-sensitive: CVE-2025-61727 was caused by an excluded subdomain constraint failing to intersect a wildcard SAN.

src/pki/name_constraints.zig
```diff
@@ -0,0 +1,678 @@
...
+        var names_examined: usize = 0;
+
+        if (!certificate.subject.isEmpty() and self.hasState(.directory_name)) {
+            if (consumeName(&names_examined, self.limits.maximum_names_per_certificate, certificate_index)) |failed| return failed;
+            if (self.checkPresented(.{ .directory_name = &certificate.subject }, certificate_index)) |failed| return failed;
+        }
+
+        if (self.hasState(.rfc822_name)) {
+            for (certificate.subject.rdns) |rdn| {
+                for (rdn.attributes) |*attribute| {
```
Jaruso, line 288: **Blocking: legacy subject `emailAddress` attributes are being constrained even when the certificate has a subjectAltName extension.**

RFC 5280 §4.2.1.10 limits this fallback explicitly: the rfc822Name constraint is applied to a subject-DN `emailAddress` attribute only when the certificate **does not include a subjectAltName extension**. This loop currently runs unconditionally whenever rfc822Name state exists, before SAN processing below.

That means a certificate such as:

```text
subject emailAddress = legacy@outside.invalid
SAN dNSName           = api.example.com
inherited rfc822Name permitted subtree = .example.com
```

is rejected here, even though the legacy subject attribute must be ignored for rfc822Name constraint processing because the SAN extension is present (and no rfc822Name SAN is claimed).

Please cache the SAN lookup at the beginning of `checkCertificate()` and guard this fallback, for example conceptually:

```zig
const subject_alt_names = certificate.subjectAltName();
if (subject_alt_names == null and self.hasState(.rfc822_name)) {
    // inspect legacy subject emailAddress attributes
}

if (subject_alt_names) |names| {
    // existing SAN processing
}
```

Please add both sides of the regression: a legacy subject email must still be constrained when SAN is absent, but must not be considered when any SAN extension is present; an actual rfc822Name SAN must of course still be constrained.

src/pki/name_constraints.zig
```diff
@@ -0,0 +1,678 @@
...
+        std.ascii.eqlIgnoreCase(name[name.len - domain.len ..], domain);
+}
+
+fn splitMailbox(value: []const u8) NameError!struct { local: []const u8, host: []const u8 } {
+    const at = std.mem.lastIndexOfScalar(u8, value, '@') orelse return error.Malformed;
+    if (at == 0 or at + 1 == value.len) return error.Malformed;
+    const local = value[0..at];
+    if (std.mem.indexOfScalar(u8, local, '@') != null) return error.Malformed;
+    for (local) |byte| {
+        if (byte < 0x21 or byte > 0x7e) return error.Malformed;
```
Jaruso: **Blocking: `splitMailbox()` is neither an RFC 2821 Mailbox parser nor a safe strict subset.**

The current printable-ASCII check rejects valid quoted local-parts and accepts invalid dot-strings. Examples:

```text
"a@b"@example.com   // valid quoted local-part, rejected because local contains '@'
.a@example.com       // invalid dot-string, currently accepted
a..b@example.com     // invalid consecutive dots, currently accepted
a.@example.com       // invalid trailing dot, currently accepted
a(b)@example.com     // invalid unquoted atom character, currently accepted
```

This parser is used for both presented rfc822Name values and exact-mailbox constraints, so these are not just parser niceties: malformed certificates can be accepted, and valid historical exact-mailbox constraints can be rejected as unsupported. Exact comparison also needs to account for RFC 2821's requirement that equivalent quoted forms compare as the same local-part (for example `root@example.com` and `"root"@example.com`).

Please implement a bounded `Local-part = Dot-string / Quoted-string` parser (including quoted-pair handling), returning a normalized/canonical local-part suitable for exact comparison. If full quoted-string support is intentionally deferred, then make that an explicit unsupported/fail-closed case **and** strictly validate dot-atoms so malformed inputs cannot pass. Add helper and signed-path tests for all examples above.

src/pki/name_constraints.zig
```diff
@@ -0,0 +1,678 @@
...
+        if (byte == '/' or byte == '?' or byte == '#') {
+            authority_end = index;
+            break;
+        }
+    }
+    var authority = uri[authority_start..authority_end];
+    if (authority.len == 0) return error.Malformed;
+    if (std.mem.lastIndexOfScalar(u8, authority, '@')) |at| authority = authority[at + 1 ..];
+    if (authority.len == 0 or authority[0] == '[') return error.Malformed;
+
```
Jaruso, line 716: **Blocking: taking the last `@` silently accepts malformed URI authorities.**

RFC 3986 authority syntax permits at most one delimiter `@`; a raw `@` is not allowed inside `userinfo`, and percent escapes must be well formed. With `lastIndexOfScalar`, inputs such as these are accepted and reduced to the apparently safe host `api.example.com`:

```text
https://bad@@api.example.com/
https://bad user@api.example.com/
https://bad%ZZ@api.example.com/
```

All three are malformed URIs/authorities, but this function will discard the invalid prefix and pass the host through DNS validation. That contradicts the fail-closed behavior expected from this path validator.

Please parse/validate the authority components explicitly: allow zero or one `@`, validate `userinfo` against the bounded RFC 3986 character set (`unreserved`, valid `%HH`, `sub-delims`, and `:`), and then parse host/port. A simpler conservative alternative is to reject userinfo entirely, but that would reject otherwise valid URI SANs, so a small bounded validator is preferable. Please add negative tests for multiple `@`, spaces/control characters, and malformed percent escapes, plus a positive userinfo case.

Jaruso (commented):
I found four decision-affecting correctness/interoperability issues that should be addressed before merge. The bounded state model, CA-group intersection, cumulative exclusions, self-issued handling, and structured diagnostics are all thoughtfully implemented, but the current matcher rejects valid wildcard and legacy-email cases while accepting malformed mailbox/URI inputs.

These are not cosmetic concerns: each can change a chain from accepted to rejected or vice versa. Please add signed-path regression coverage for each behavior, in addition to helper-level tests.

I was able to inspect the complete one-commit diff and all core changed sources. I could not independently run Zig tests in this environment because Zig is unavailable, and GitHub's workflow/status endpoints were returning an upstream 502, so I have not independently verified the reported green checks.

Jaruso (commented):
Re-review complete at `9557fc39a2200d788f56270d0385640a7c44da9e`. I do not have any further change requests.

All four prior blockers are addressed:

- wildcard dNSName SANs now have dedicated permitted-subset / excluded-intersection handling rather than being treated as ordinary hostnames; the regression matrix covers the security-sensitive direct-child and leading-dot cases;
- legacy subject-DN `emailAddress` fallback is now gated on the absence of the subjectAltName extension, while actual rfc822Name SANs remain constrained;
- mailbox parsing now validates dot-strings, supports bounded quoted strings and quoted-pairs, and canonicalizes local parts for exact-mailbox comparison;
- URI userinfo now rejects multiple `@` delimiters, invalid characters/control bytes, and malformed percent escapes before extracting the host.

I also reviewed the two follow-up commits for second-order regressions. The wildcard set logic matches the intended one-label wildcard semantics, the local-part iterator is safe under the parser invariants, and the URI authority validation remains bounded by the existing URI length limit. The new signed-path tests exercise the fixes rather than relying only on helper tests.

The implementation satisfies the Name Constraints portion of #345 while correctly leaving certificate-policy processing open. From code review, this is ready to merge.

GitHub currently reports the PR as mergeable, but the commit-status endpoint returned no status contexts, so this conclusion is based on the code and committed test coverage rather than independently observed CI results. GitHub does not allow me to submit an APPROVE review on the PR author's own pull request, so this is recorded as a comment review.

PR #445, opened 2026-07-17
pki: add bounded RFC 9618 certificate policy validation

Jaruso:
## Summary

- parse typed `certificatePolicies`, `policyMappings`, `policyConstraints`, `inhibitAnyPolicy`, `anyPolicy`, and bounded policy qualifiers
- add a deterministic RFC 9618 policy graph and integrate it after Name Constraints but before leaf identity acceptance
- expose borrowed initial-policy configuration plus owned, sorted authority- and user-constrained policy outputs
- add signed synthetic chains, malformed DER/counter boundaries, allocation sweeps, alternate-candidate isolation, and fixed OpenSSL differential fixtures

## Graph representation and bound

The validator stores one node for each policy OID at each graph depth. Nodes use integer parent indices and unique bounded expected-policy arrays. When several policy paths converge, the child gains parent references instead of being copied, so mappings cannot expand into the obsolete RFC 5280 policy tree.

Every edge joins adjacent depths. Processing and ancestor pruning are iterative. Configurable limits cover path depth, policies and qualifiers per certificate, qualifier bytes, mappings, nodes per depth, total nodes, edges, expected entries, parent references, user inputs, outputs, and total operations. Successful results expose deterministic resource accounting; tests assert converging graphs retain exact node/edge counts across repeated runs.

## Policy behavior

- Default input is `anyPolicy`, with explicit policy, mapping inhibition, and anyPolicy inhibition disabled for ordinary compatibility.
- Trust-anchor certificate extensions do not initialize or constrain policy state.
- Processing runs from the certificate below the anchor toward the leaf while diagnostics retain leaf-first indices.
- Explicit-policy, mapping, and anyPolicy counters use checked nonnegative arithmetic, self-issued intermediate exceptions, and separate target wrap-up.
- Missing `certificatePolicies` nulls the graph; explicit-policy requirements then fail at the applicable depth.
- One-to-many, many-to-one, and multi-level mappings merge in the graph; mappings to/from anyPolicy and mappings on targets fail closed.
- Results are unique lexically sorted OID value copies. The deprecated graph/tree is not returned.

CPS Pointer and User Notice qualifiers are syntax-checked and bounded but never fetched or displayed. Unsupported qualifiers are structurally consumed; critical unsupported qualifiers and unsupported anyPolicy qualifiers reject without copying peer text into diagnostics.

Structured failures distinguish invalid/required policy, unsupported qualifiers, invalid mappings/constraints/inhibitAnyPolicy, policy resource exhaustion, and allocation failure. Diagnostics include the leaf-first certificate index, extension/policy OID when available, graph depth, and processing stage.

## Fixtures and differential checks

The checked-in public Ed25519 chains cover direct policy success, missing required policy, anyPolicy success/inhibition, simple mapping, caller mapping inhibition, mapping inhibition from `policyConstraints`, `requireExplicitPolicy`, and certificate-enforced `inhibitAnyPolicy`. `generate.sh` keeps private keys in a temporary directory and verifies the expected OpenSSL decisions before completing.

## Validation

- `zig build test-pki test-pki-openssl test-pki-policy-openssl test-crypto test-tls test --summary all --error-style verbose`
  - 43/43 build steps succeeded
  - 1,466 tests passed; 1 existing test skipped
  - 187 PKI tests passed
  - OpenSSL PKI matrix: 2/2 passed
  - focused policy/OpenSSL matrix: 1/1 passed
- `zig build test-pki` repeated ten times after exact graph resource assertions
- `git diff --check`
- staged-tree private-key/token scan

Current main already contains the merged RFC 4518 name-chaining work and its normalization/prohibited-input fixtures, satisfying the compatibility gate noted on #345. The broader second-validator hostile corpus remains tracked by #348 and is outside the original #345 acceptance criteria. This completes every original criterion without touching PR #443 TLS/QUIC ownership files.

Closes #345


src/pki/x509.zig
```diff
@@ -1114,19 +1164,207 @@ const Parser = struct {
...
+            const elem = inner.readElement() catch return error.MalformedExtension;
+            if (elem.tag.class != .context_specific or elem.tag.constructed or elem.tag.number > 1) {
+                return error.MalformedExtension;
+            }
+            if (last_tag) |last| if (elem.tag.number <= last) return error.MalformedExtension;
+            last_tag = elem.tag.number;
+            der.validateInteger(elem.content, self.limits.der.max_integer_bytes) catch return error.MalformedExtension;
+            const integer = der.IntegerView{ .content = elem.content };
+            if (integer.isNegative()) return error.MalformedExtension;
+            const count = integerToU32(integer) orelse return error.MalformedExtension;
```
Jaruso: **[merge-blocking] Do not narrow `SkipCerts ::= INTEGER (0..MAX)` to `u32`.**

Both `PolicyConstraints` fields and `InhibitAnyPolicy` use RFC 5280's `SkipCerts`, whose ASN.1 domain is every non-negative INTEGER, not `0..2^32-1`. The currently added test vector `01 00 00 00 00` is the canonical positive DER encoding of `2^32`; classifying it as `MalformedExtension` rejects a conforming certificate and can change the path decision.

The engine never needs the exact arbitrarily large value: it only takes `min(current_counter, skip_certs)`, and `current_counter` is at most the bounded path length plus one. Parse into the counter's domain with saturation rather than rejecting overflow. For example:

```zig
pub const PolicyConstraints = struct {
    require_explicit_policy: ?usize,
    inhibit_policy_mapping: ?usize,
};

fn integerToSaturatingUsize(integer: der.IntegerView) usize {
    var value: usize = 0;
    for (integer.content) |byte| {
        value = std.math.mul(usize, value, 256) catch
            return std.math.maxInt(usize);
        value = std.math.add(usize, value, @as(usize, byte)) catch
            return std.math.maxInt(usize);
    }
    return value;
}
```

Keep the negative and DER-canonicality checks, then use the saturating helper for both policy-constraint fields and `inhibitAnyPolicy`. Please replace the two `2^32` “malformed” cases with positive tests showing that a count larger than the path behaves like an uninhibiting count for `requireExplicitPolicy`, `inhibitPolicyMapping`, and `inhibitAnyPolicy`.

src/pki/x509.zig
```diff
@@ -1114,19 +1164,207 @@ const Parser = struct {
...
+        if (!isSequence(elem)) return error.MalformedExtension;
+        var reader = der.Reader.init(elem.encoded, self.limits.der);
+        var notice_ref = reader.readSequence() catch return error.MalformedExtension;
+        reader.expectEnd() catch return error.MalformedExtension;
+        const organization = notice_ref.readElement() catch return error.MalformedExtension;
+        try self.validateDisplayText(organization);
+        var numbers = notice_ref.readSequence() catch return error.MalformedExtension;
+        while (numbers.remaining() > 0) {
+            const number = numbers.readInteger() catch return error.MalformedExtension;
+            if (number.isNegative()) return error.MalformedExtension;
```
Jaruso: **[merge-blocking] `NoticeReference.noticeNumbers` is `SEQUENCE OF INTEGER`, not non-negative INTEGER.**

RFC 5280 constrains `SkipCerts` with `(0..MAX)`, but it places no sign constraint on the INTEGERs inside `NoticeReference.noticeNumbers`. This check therefore rejects a syntactically valid User Notice qualifier. Since this implementation deliberately validates and then omits qualifiers, there is no downstream indexing operation that requires the value to be non-negative.

The parser should only require canonical DER INTEGER encodings here:

```zig
while (numbers.remaining() > 0) {
    _ = numbers.readInteger() catch return error.MalformedExtension;
}
```

Please add a User Notice fixture containing a `NoticeReference` with a canonical negative INTEGER such as `02 01 ff` and verify that it parses (including when `certificatePolicies` is critical). Malformed/non-minimal INTEGER encodings should continue to fail through `readInteger()`.

src/pki/testdata/policy/generate.sh
```diff
@@ -0,0 +1,118 @@
...
+
+policy_a="1.3.6.1.4.1.55555.1"
+policy_b="1.3.6.1.4.1.55555.2"
+
+openssl genpkey -algorithm ED25519 -out "$tmp/root.key"
+openssl req -new -x509 -key "$tmp/root.key" -days 3650 -set_serial 100 \
+  -subj "/CN=Tardigrade Policy Root" \
+  -addext "basicConstraints=critical,CA:TRUE" \
+  -addext "keyUsage=critical,keyCertSign,cRLSign" \
+  -out "$script_dir/root.crt"
```
Jaruso: **[merge-blocking] Verify a staged fixture set before replacing the checked-in certificates.**

The README says regeneration verifies every expected decision before replacing the fixed fixture set, but this script writes `root.crt`, each intermediate, each leaf, and the combined chain directly into `src/pki/testdata/policy` before any of the final assertions run. If generation fails midway—or an expected-rejection case unexpectedly succeeds—the command exits with a partially replaced, internally inconsistent tracked fixture set.

Generate the complete public fixture set under the temporary directory, run every OpenSSL assertion against that staged set, and publish only after all assertions pass. A suitable shape is:

```bash
out="$tmp/fixtures"
mkdir -p "$out"

# All `-out`, `-CAfile`, `-CA`, and `-untrusted` certificate paths use `$out`.
# Private keys remain elsewhere under `$tmp`.

verify_policy ...
# ...all remaining positive and negative assertions...

fixtures=(
  root.crt direct-intermediate.crt direct-leaf.crt
  # ...the complete explicit manifest...
)
for file in "${fixtures[@]}"; do
  install -m 0644 "$out/$file" "$script_dir/$file.new"
done
for file in "${fixtures[@]}"; do
  mv "$script_dir/$file.new" "$script_dir/$file"
done
```

An explicit manifest also prevents stale certificates from silently surviving when the matrix changes. At minimum, no tracked fixture should be touched until all staged OpenSSL decisions have passed.

Jaruso (commented):
I reviewed the full 27-file change against issue #345, including the later distinguished-name normalization compatibility gate, the RFC 5280 extension syntax, the RFC 9618 graph-processing steps, integration/ownership paths, synthetic and OpenSSL fixtures, allocation behavior, and the current CI run.

The core design is strong and most of the acceptance contract is covered:

- the trust anchor is excluded from prospective-path policy processing;
- graph construction, anyPolicy propagation, mapping preparation, self-issued counter handling, pruning, and final authority/user constrained sets follow the RFC 9618 order;
- the graph is iterative and structurally bounded, with shared parents rather than tree duplication;
- policy failures are isolated per candidate and accepted-result ownership is cleaned up correctly on later identity/allocation failures;
- existing path-validation coverage plus the new signed policy chains cover the original #345 pass/fail, alternate-path, trust-anchor, no-network, and bounded-complexity criteria;
- current main's RFC 4518 chaining implementation and fixtures address the normalization gate's supported comparison classes and fail-closed behavior, while the broader independent differential corpus remains tracked by #348 as stated.

The connected account is the PR author, so GitHub does not permit a `REQUEST_CHANGES` review. Treat the three inline findings below as **merge-blocking requested changes**. The first two reject values that are valid under the RFC 5280 ASN.1 schema, so they can change a conforming path's validation result. The third can leave the checked-in differential fixture set partially replaced after a failed regeneration.

One additional documentation cleanup should be made while updating this: the `validatePath` comment still says that the owned element slice is “the only allocation.” Policy validation now allocates bounded arena state and two owned policy-output slices, so that statement is no longer true.

After these findings are corrected with the suggested regression cases, I would expect this PR to satisfy #345's acceptance criteria.

Jaruso (commented):
Re-review complete against head `ba8a90579af5a18e766ddbf6df9735973c828909`. **No remaining findings; this is good to merge.**

I re-reviewed the complete follow-up commit and the affected surrounding paths rather than only checking whether the old lines disappeared.

### Previous merge blockers

1. **Unbounded `SkipCerts` domain — resolved.** `PolicyConstraints` and `inhibitAnyPolicy` now expose `usize`, canonical non-negative DER integers are converted with saturation, and values wider than the platform saturate to `maxInt(usize)`. This preserves the only operation the policy engine needs (`min` against bounded path counters) without narrowing RFC 5280's `INTEGER (0..MAX)` domain to `u32`.

   The new path-level regression cases are effective: oversized `requireExplicitPolicy`, `inhibitPolicyMapping`, and `inhibitAnyPolicy` values each flow through a chain whose decision would change if the value were rejected or truncated to zero. The helper also has a direct wider-than-`usize` saturation test.

2. **Negative User Notice numbers — resolved.** `NoticeReference.noticeNumbers` now requires canonical DER INTEGER syntax without imposing a sign restriction. A critical `certificatePolicies` fixture with canonical `INTEGER -1` is accepted, while a non-minimal negative encoding remains rejected.

3. **Fixture publication before verification — resolved.** All certificates and the combined untrusted chain are generated into a temporary staged directory; every positive and expected-negative OpenSSL decision is evaluated against that staged set before any tracked fixture is touched. Publication uses an explicit manifest, cleans temporary `.new` files on failure, and rejects stale checked-in certificates not represented by the matrix.

### Additional follow-up checks

- The stale `validatePath` allocation comment now accurately documents temporary policy graph state and owned policy output slices.
- Both OpenSSL differential call sites derive `-attime` from the shared `validation_time` constant.
- The VisibleString tag is named and documented rather than remaining a magic number.
- The `usize` type change is consistent through the parsed extension model, accessors, and policy-counter `@min` operations.
- The fixture staging changes do not alter key placement: private keys remain under the temporary directory only.
- The latest CI run for this head completed successfully across formatting, lint/shellcheck, security and dependency checks, unit tests on Ubuntu, macOS and ARM, build profiles, packaging, performance smoke, and TLS dependency auditing.

The original #345 acceptance criteria and its distinguished-name normalization compatibility gate remain satisfied after these changes. Because the connected account owns the PR, GitHub does not permit me to submit an `APPROVE` review, but my review conclusion is **approved / ready to merge**.

PR #446, opened 2026-07-17
pki: add trust-store abstraction and root bundle loaders

Jaruso:
## Summary
- add a provider-neutral `pki.trust_store` module with immutable trust-anchor snapshots and a PEM/DER-backed `BundleStore`
- enforce exact-DER deduplication, typed malformed/non-CA anchor rejection, and reload-safe snapshot ownership
- cover mixed/file-backed loads plus reload-safe validation flows in new `test-pki` coverage

## Testing
- `git diff --check`
- `zig build test-pki --summary all --error-style verbose`

Closes #346

### DIFF
--- a/src/pki/trust_store.zig
+++ b/src/pki/trust_store.zig
@@ -56,17 +56,8 @@ pub const Snapshot = struct {
         for (inputs) |input| {
             switch (input) {
                 .pem => |pem_text| {
-                    const chain = try pem.loadChainPem(allocator, pem_text, limits.loader);
-                    defer allocator.free(chain.certificates);
-                    for (chain.certificates) |certificate| {
-                        try appendOwnedCertificate(
-                            allocator,
-                            &owned_certs,
-                            &parsed_anchors,
-                            certificate,
-                            limits,
-                        );
-                    }
+                    var chain = try pem.loadChainPem(allocator, pem_text, limits.loader);
+                    try appendPemChain(allocator, &owned_certs, &parsed_anchors, &chain, limits);
                 },
                 .der => |der_bytes| {
                     const certificate = try pem.loadCertificateDer(allocator, der_bytes, limits.loader);
@@ -100,17 +91,8 @@ pub const Snapshot = struct {
         for (inputs) |input| {
             switch (input) {
                 .pem => |sub_path| {
-                    const chain = try pem.loadChainPemFile(allocator, io, dir, sub_path, limits.loader);
-                    defer allocator.free(chain.certificates);
-                    for (chain.certificates) |certificate| {
-                        try appendOwnedCertificate(
-                            allocator,
-                            &owned_certs,
-                            &parsed_anchors,
-                            certificate,
-                            limits,
-                        );
-                    }
+                    var chain = try pem.loadChainPemFile(allocator, io, dir, sub_path, limits.loader);
+                    try appendPemChain(allocator, &owned_certs, &parsed_anchors, &chain, limits);
                 },
                 .der => |sub_path| {
                     const certificate = try pem.loadCertificateDerFile(allocator, io, dir, sub_path, limits.loader);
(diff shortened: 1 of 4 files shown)


src/pki/trust_store.zig
```diff
@@ -0,0 +1,332 @@
...
+        var owned_certs: std.ArrayList(pem.Certificate) = .empty;
+        errdefer deinitOwnedArrayList(&owned_certs, allocator);
+        var parsed_anchors: std.ArrayList(x509.Certificate) = .empty;
+        errdefer deinitParsedArrayList(&parsed_anchors, allocator);
+
+        for (inputs) |input| {
+            switch (input) {
+                .pem => |pem_text| {
+                    const chain = try pem.loadChainPem(allocator, pem_text, limits.loader);
+                    defer allocator.free(chain.certificates);
```
Jaruso: **Blocking — unprocessed certificates leak when a multi-certificate PEM fails partway through.**

`pem.loadChainPem` returns a `CertificateChain` that owns the DER allocation for *every* element. This defer releases only the outer slice. Each entry handed to `appendOwnedCertificate` is consumed, but when one entry returns `NonCaAnchor`, an X.509 parse error, `TooManyAnchors`, or OOM, every certificate after that entry remains in `chain.certificates` and its DER is never freed. The file-backed branch repeats the same ownership pattern.

For example, a bundle ordered as `[valid CA, non-CA leaf, valid CA]` does this:

1. The first DER moves into `owned_certs` and is cleaned by the outer `errdefer`.
2. The second DER is freed by `appendOwnedCertificate` before returning `error.NonCaAnchor`.
3. The third DER is never visited, and this defer frees only its `pem.Certificate` container—not `certificate.der`.

A shared helper with an ownership cursor would make the transfer explicit and fix both callers. The cursor must advance **before** calling `appendOwnedCertificate`, because that function consumes the current certificate on both success and error:

```zig
fn appendPemChain(
    allocator: std.mem.Allocator,
    chain: pem.CertificateChain,
    owned_certs: *std.ArrayList(pem.Certificate),
    parsed_anchors: *std.ArrayList(x509.Certificate),
    limits: Limits,
) Error!void {
    defer allocator.free(chain.certificates);

    var next: usize = 0;
    errdefer {
        for (chain.certificates[next..]) |*certificate| {
            certificate.deinit(allocator);
        }
    }

    while (next < chain.certificates.len) {
        const certificate = chain.certificates[next];
        next += 1; // appendOwnedCertificate now owns this entry on every path
        try appendOwnedCertificate(
            allocator,
            owned_certs,
            parsed_anchors,
            certificate,
            limits,
        );
    }
}
```

Then both `.pem` branches can load the chain and pass it to this helper. Please add a regression using one concatenated PEM input with a valid root, a non-CA certificate, and a trailing root; `expectError(error.NonCaAnchor, ...)` under `testing.allocator` should return without reporting a leak. A `max_anchors` failure with a trailing entry would be useful additional coverage.

Jaruso (commented):
I reviewed all four changed files against issue #346 and traced the new ownership model through `pem.CertificateChain`, `x509.Certificate`, `path_builder`, and `path_validator`.

The main acceptance criteria are represented correctly: configured snapshots validate through the existing builder/validator, exact-DER duplicates use first-occurrence wins, malformed/non-CA/empty inputs are typed, successful reloads preserve previously acquired snapshots, and the exported API remains provider-neutral with no OpenSSL object leakage. CI is also green across the repository matrix.

There is one blocking failure-path ownership bug in multi-certificate PEM processing. Both `loadBuffers` and `loadFiles` free only the `CertificateChain.certificates` container after transferring entries one at a time. If processing aborts on one certificate, later entries still own DER allocations but are never transferred or deinitialized. This is especially relevant to reload paths, where repeated attempts against an invalid configured bundle would accumulate leaked memory.

Please fix both PEM paths using explicit transfer accounting (or a shared helper) and add a regression where a single PEM bundle contains a valid CA, then a non-CA/malformed entry, then another certificate. The expected typed error should be returned with no leak under `std.testing.allocator`. A limit-error case with a trailing certificate would also cover the same ownership boundary.

GitHub does not allow me to submit a formal `REQUEST_CHANGES` review on my own PR, but I consider this finding merge-blocking.

Jaruso (commented):
Re-review complete at `6943e1944f5b8df08838b97c2bac3ad651a56dbe`.

The blocking PEM ownership defect from my first review is fixed correctly. `appendPemChain` now treats each certificate as consumed before calling `appendOwnedCertificate`, cleans every unvisited DER allocation on failure, and frees the outer chain slice on both success and error. Because the same helper is used by `loadBuffers` and `loadFiles`, the fix covers both affected loaders. The added `NonCaAnchor` and `TooManyAnchors` regressions exercise failure in the middle of a multi-certificate bundle with a trailing owned certificate, and `std.testing.allocator` will detect the leak if this regresses.

I repeated the full acceptance-criteria and surrounding-code review rather than limiting this pass to the patch:

- configured roots validate representative chains through the existing path builder and RFC 5280 validator;
- exact-DER deduplication remains deterministic and first-occurrence-wins;
- malformed roots, non-CA inputs, empty stores, and anchor-limit failures remain typed;
- reload is transactional on failure and already-acquired snapshots retain independent DER and parser ownership after successful reload;
- the provider boundary exposes only pure-Zig snapshot types and leaves platform-native stores as an explicit future adapter point;
- allocator cleanup is balanced across parsing, deduplication, list growth, snapshot finishing, cloning, and partial PEM transfer failures.

I found no further code changes to request. Unit tests are already green on Ubuntu, macOS, and ARM at this head; several slower repository-wide CI jobs were still running when I completed the review. Subject to the remaining CI jobs completing successfully, this is ready to merge.

GitHub does not permit a formal approval review on a pull request owned by the authenticated account, so this clean re-review is submitted as `COMMENTED`.

PR #449, opened 2026-07-17
test(pki): add three-way hostile certificate corpus

Jaruso:
**What changed and why**

Adds the first substantial slice of the differential Web PKI story:

- a bounded 17-case executable corpus comparing Tardigrade with out-of-process OpenSSL and Go `crypto/x509`;
- deterministic, project-owned hostile fixtures covering malformed DER, duplicate and unknown critical extensions, wildcard edges, Name Constraints, cross-sign ambiguity, path limits, and corrupt signatures;
- explicit per-validator decisions and required normalization rationales for intentional policy differences;
- persisted JSON mismatch artifacts with one-case reproduction commands;
- required core CI plus a weekly/manual extended workflow; and
- promotion of the reduced malformed DER input into PKI unit and fuzz corpus coverage.

The fixture generator uses the RFC 5612 documentation PEN, writes no private keys, regenerates byte-for-byte, rejects unexpected files and PEM blocks, and verifies the corpus with both external validators before publishing fixture updates.

**Related issues**

Refs #348. This is the first slice; #348 remains open for automated mismatch minimization/promotion and broader extended-corpus growth.

**Tests**

- `tests/vectors/pki/generate.sh`
- `zig build test-pki test-pki-differential test-pki-differential-extended test-pki-openssl test-pki-policy-openssl --summary all --error-style verbose --test-timeout 2m`
- `zig build test --summary all --error-style verbose --test-timeout 2m` (1,486 passed, 1 skipped)
- `zig build test-integration --summary all --error-style verbose` (65 passed, 4 skipped)
- `zig build -Doptimize=ReleaseFast`
- Go test/vet for the oracle and generator
- `shellcheck tests/vectors/pki/generate.sh`
- actionlint v1.7.12 on both changed workflows
- staged-tree private-key/token scan

**Security impact**

Test-only PKI assurance change. It exercises TLS certificate decisions but does not change the production validation path. External validators stay out of process, fixture and oracle I/O is bounded, explicit resource exhaustion is a harness failure, and no private-key material is persisted.

**Performance impact**

None on the runtime or hot path. The new work is CI/test-only.

**Checklist**

- [x] `zig build test-integration` green


tests/pki_differential.zig
```diff
@@ -0,0 +1,437 @@
...
+        openssl,
+        "verify",
+        "-attime",
+        validation_time_argument,
+        "-purpose",
+        "sslserver",
+        "-CAfile",
+        case.root_file,
+    });
+    if (case.intermediate_file) |path| try argv.appendSlice(allocator, &.{ "-untrusted", path });
```
Jaruso, line 253: **Blocking: isolate the OpenSSL oracle from host trust configuration.**

`-CAfile` supplies this case's roots, but it does not provide the same isolation guarantee as Go's explicitly constructed `CertPool` or Tardigrade's loaded snapshot. OpenSSL may still consult a configured default CA directory/store unless those sources are disabled. That permits an acceptance through a host anchor that is absent from `case.root_file`, which would turn a real three-way disagreement into a false match. The current project-owned roots are unlikely to collide with the runner store, but future public-root, cross-sign, or alternate-path fixtures can expose this immediately.

Prefer `-trusted`, which makes the provided bundle the exclusive trust-anchor source, and apply the identical invocation policy in `tests/vectors/pki/generate.sh` (and ideally the older OpenSSL differential helper):

```zig
try argv.appendSlice(allocator, &.{
    openssl,
    "verify",
    "-attime",
    validation_time_argument,
    "-purpose",
    "sslserver",
    "-trusted",
    case.root_file,
});
```

An alternative is to keep `-CAfile` and explicitly disable every default trust source supported by the pinned OpenSSL version, but `-trusted` is harder to misconfigure and directly expresses the harness contract.

.github/workflows/ci.yml
```diff
@@ -147,6 +147,54 @@ jobs:
...
+
+      - name: Print validator versions
+        run: |
+          zig version
+          openssl version
+          go version
+
+      - name: Run stable differential corpus
+        env:
+          TARDIGRADE_PKI_DIFF_ARTIFACT_DIR: pki-differential-artifacts
```
Jaruso, line 189: **Important: make CI prove that the checked-in corpus is the generator's canonical output.**

This job validates the committed certificates, but it never runs `generate.sh`. As a result, `generate.go`, the shell verification matrix, and the checked-in fixtures can drift independently while the differential test remains green. That weakens the provenance/determinism acceptance criterion: a fixture may still produce the expected accept/reject result even though it is no longer byte-for-byte generated by the documented source.

The job already has the required Go and OpenSSL installations, so add a regeneration check before replaying the corpus:

```yaml
- name: Verify deterministic PKI fixtures
  run: |
    tests/vectors/pki/generate.sh
    git diff --exit-code -- tests/vectors/pki

- name: Run stable differential corpus
  # existing step
```

This reuses the script's two-generation diff, external-validator checks, unexpected-file checks, and private-key/PEM-block scans, then fails if publication would change any tracked fixture. I would run the same check in the weekly workflow or centralize it in a reusable script/step.

tests/pki_differential.zig
```diff
@@ -0,0 +1,437 @@
...
+    );
+    defer allocator.free(reproduce);
+    const payload = try compat.stringifyAlloc(allocator, .{
+        .schema_version = 1,
+        .case_id = case.id,
+        .profile = case.profile,
+        .category = case.category,
+        .validation_time = validation_time,
+        .root_file = case.root_file,
+        .intermediate_file = case.intermediate_file,
```
Jaruso, line 390: **Suggestion: persist the validator/build identities needed to reproduce a historical mismatch.**

The artifact contains the case, inputs, decisions, diagnostics, and command, which is a good start. The command is not fully reproducible across time, though: Go is pinned by CI, but the Ubuntu OpenSSL package and runner image float, and the artifact does not identify the tested Tardigrade commit. A mismatch caused by validator-version drift can disappear before someone downloads and replays it.

Please consider adding bounded fields such as:

```zig
.git_sha = compat.getEnvVarOwned(..., "GITHUB_SHA") catch "unknown",
.openssl_version = ... // `openssl version`
.go_version = ...      // `go version`
.zig_version = builtin.zig_version_string,
```

Query the external versions once per corpus rather than once per case, then pass them into `writeArtifact`. The workflow log currently prints versions, but keeping them in the JSON makes the artifact self-contained after log retention expires.

Jaruso (commented):
Reviewed the full patch against #348's acceptance criteria and the surrounding PKI implementation.

This is a strong first slice: the 13-case core/4-case extended split is explicit, Tardigrade/OpenSSL/Go decisions are independently recorded, intentional differences require normalization, fixture and output sizes are bounded, failures persist one-case reproduction artifacts, CI is green, and the reduced DER input is wired into both DER and X.509 fuzz regression coverage. Keeping #348 open for automated minimization/promotion and broader corpus growth is appropriate.

**I would not merge this as-is because of the OpenSSL trust-store isolation issue below.** GitHub does not allow the connected account to submit a formal “Request changes” review on its own PR, so this is posted as a comment review instead. I also left two hardening comments around enforcing fixture provenance in CI and making artifacts version-reproducible.

Additional non-blocking cleanup: please include `tests/vectors/pki/generate.sh` in the repository's CI shellcheck command. The current lint job only scans `scripts/` and `packaging/`, so this security-sensitive publication wrapper is only linted manually. Longer term, per-child process deadlines would make an OpenSSL/Go hang produce a case-specific `tool_failure` artifact rather than relying solely on the coarse Zig test/job timeout.

tests/pki_differential.zig
Jaruso: Verified in follow-up commit `95786f4`: the hostile-corpus harness, fixture generator checks, and the existing OpenSSL differential helper now consistently use `-trusted`. OpenSSL documents that this implies `-no-CAfile`, `-no-CApath`, and `-no-CAstore`, so only the case-provided bundle can act as trust anchors. The updated core differential job passes. This blocker is fully resolved.

.github/workflows/ci.yml
Jaruso: Verified in `95786f4`: both the required core workflow and weekly/manual extended workflow now run `generate.sh` and then `git diff --exit-code -- tests/vectors/pki`. The latest core CI run completed the new deterministic-fixture step successfully, proving the committed fixtures match the generator output. This is resolved.

tests/pki_differential.zig
Jaruso: Verified in `95786f4`: schema v2 artifacts now persist the tested Git SHA plus OpenSSL, Go, and Zig identities. The external commands are queried once per corpus, their output is bounded, failures degrade to bounded `unavailable`/termination diagnostics, and all owned strings are released by `RuntimeIdentity.deinit`. This suggestion is resolved.

Jaruso (commented):
Follow-up review complete on `95786f4`.

I re-reviewed the five files changed after the first pass, verified each prior finding against the implementation rather than only the diff description, checked the latest CI run, and re-checked the affected #348 acceptance criteria.

All prior findings are resolved:

- OpenSSL trust is now isolated with `-trusted` consistently across the three-way harness, fixture publication checks, and the older OpenSSL differential helper.
- Core and extended CI both regenerate the corpus and fail on any tracked drift.
- The security-sensitive fixture script is covered by the repository shellcheck job.
- Mismatch artifact schema v2 records the tested Git SHA and OpenSSL, Go, and Zig identities, gathered once per corpus with bounded diagnostics and correct ownership cleanup.
- The latest required CI run is green, including deterministic fixture verification and the core differential corpus.

I found no new correctness, security, resource-handling, or acceptance-criteria blocker in the follow-up changes. **This PR is good to merge.**

GitHub does not permit the connected account to formally approve its own PR, so this is submitted as a comment review.

PR #453, opened 2026-07-17
test(pki): add automated mismatch minimization and reduced-corpus promotion

Jaruso:
**What changed and why**

Second slice of the differential Web PKI story (#348), covering reduced mismatch artifacts and promotion-ready regression seeds.

- `tests/pki_reduce.zig`: bounded deterministic greedy delta debugging over raw bytes. Identical inputs reduce to identical bytes, and a hard oracle-call budget keeps hostile inputs bounded.
- Every unexplained differential mismatch now runs component-aware minimization across the leaf, intermediate bundle members, and trust anchors. The three component groups share one 512-call mismatch budget through fair per-component allowances, so an early leaf attempt cannot starve later candidates.
- The minimizer preserves Tardigrade's exact in-process classification, then externally verifies retained candidates against Tardigrade, OpenSSL, and Go `crypto/x509`. Selection prefers promotion-ready one-minimal reproductions before larger partial reductions.
- Reduced artifacts are persisted only after verification selects the winning candidate. Candidate probe DER/PEM/bundle files are deleted, the selected bytes are republished under canonical `.reduced.*` paths, and the JSON SHA describes those canonical files.
- If every reduced candidate diverges externally, the reducer falls back to original component bytes only when those bytes still reproduce the mismatch. That fallback records the rejected candidate observations, marks the artifact non-promotable, and fails instead of returning metadata if the original bytes no longer reproduce.
- Mismatch artifacts use schema v3 with reduction metadata: component, original/reduced/candidate sizes, selected candidate budget, total 512-call budget, reduced SHA-256, canonical reduced DER/PEM paths, substituted validator inputs, observed reduced decisions, optional `candidate_observed`, and promotability flags.
- `tests/vectors/pki/reduced/manifest.zig`: promotion registry embedded into the PKI unit-test module. Registered seeds join the DER and X.509 fuzz corpora and get parse/path outcome regression coverage plus registry hygiene checks. Current seeds: `duplicate-critical-extension` and `corrupt-certificate-signature`.
- `tests/vectors/pki/generate.sh` treats `reduced/` as a closed directory and extends the private-key scan across promoted reduced seeds.

**Related issues**

Refs #348. Remaining for later slices: broader extended-corpus growth such as additional algorithm-confusion and encoding-edge cases.

**Tests**

Current local verification after the review fixes:

- `git diff --check`
- `zig build test-pki-reduce --summary all --error-style verbose` (17/17)
- `zig build test-pki --summary all --error-style verbose` (201/201)
- `zig build test-pki-differential --summary all --error-style verbose` (1/1)
- `zig build test --summary all --error-style verbose` (1412/1413 passed, 1 skipped)

Earlier validation for this PR also covered `test-pki-differential-extended`, OpenSSL PKI targets, integration tests, release build, `tests/vectors/pki/generate.sh`, shellcheck, actionlint, and an end-to-end forced mismatch drill.

**Security impact**

Test-only PKI assurance change; production validation is untouched. Minimization is bounded to 512 total in-process oracle calls per mismatch, out-of-process validators only run during differential test artifact verification, candidate probe files are removed, and the promoted reduced corpus remains provenance-documented and private-key scanned.

**Performance impact**

None on runtime or hot paths. Minimization only runs when a differential mismatch already failed the build.

**Checklist**


tests/pki_differential.zig
```diff
@@ -364,6 +561,61 @@ fn writeArtifact(
...
+            .target_class = r.target_class,
+            .reduced_der_file = reduced_der_path.?,
+            .reduced_pem_file = reduced_pem_path.?,
+            .reduced_sha256 = &reduced_sha256_hex,
+            .observed_reduced = .{
+                .tardigrade = .of(reduced_tardigrade.?),
+                .openssl = .of(reduced_openssl.?),
+                .go = .of(reduced_go.?),
+            },
+            .preserves_observed_statuses = reduced_tardigrade.?.status == tardigrade.status and
```
Jaruso: **[P1] Do not emit a “reduced mismatch” that no longer reproduces the observed validator tuple.**

The minimization oracle preserves only Tardigrade’s exact class. OpenSSL and Go are checked once afterward, but this boolean is merely recorded; a candidate with `preserves_observed_statuses = false` is still written and advertised with a promotion registry and regression target. Such a candidate is not a minimized reproduction of the disagreement required by #348.

At minimum, fall back to the original leaf bytes whenever the reduced candidate changes any original status, so every emitted fixture is reproducible. Prefer a separately bounded external-aware second stage:

```zig
const DifferentialTarget = struct {
    tardigrade_class: []const u8,
    openssl_status: Status,
    go_status: Status,
};

fn preservesDifferential(
    ctx: *const DifferentialOracle,
    candidate: []const u8,
) !bool {
    if (!try ctx.tardigrade_keeps(candidate)) return false;

    const observed = try ctx.verifyExternal(candidate);
    defer observed.deinit(ctx.allocator);
    return observed.openssl.status == ctx.target.openssl_status and
        observed.go.status == ctx.target.go_status;
}
```

The external stage can have a much smaller independent budget. Regardless of the chosen algorithm, promotion should require the original tuple to survive; an informational `false` is not enough.

tests/pki_differential.zig
```diff
@@ -349,13 +391,168 @@ fn expectedStatus(decision: manifest.Decision) Status {
...
+/// Bounded automated minimization of a mismatching case's leaf input. Returns
+/// null when the supporting inputs cannot be loaded (the artifact still
+/// records the unreduced disagreement); intermediates and the trust anchor
+/// are kept fixed while only the leaf shrinks.
+fn minimizeLeaf(
+    allocator: std.mem.Allocator,
+    case: manifest.Case,
+    tardigrade_obs: Observation,
+) error{OutOfMemory}!?Reduction {
+    var leaf_chain = pki.pem.loadChainPemFile(allocator, compat.io(), .cwd(), case.leaf_file, .{}) catch |err|
```
Jaruso: **[P1] The reducer must support the certificate component that actually causes the mismatch, not always the leaf.**

Issue #348 covers whole chains, and the existing manifest already contains cases whose relevant behavior lives in an intermediate or trust input: Name Constraints, path-length handling, and cross-sign/path selection. Keeping roots and intermediates fixed while reducing only `case.leaf_file` can produce an unrelated leaf seed while leaving the offending bytes untouched. That cannot become the focused regression fixture described by the acceptance criterion.

One workable model is to record/try an explicit component target:

```zig
const ReductionTarget = union(enum) {
    leaf,
    intermediate: usize,
    root: usize,
};

const Reduction = struct {
    target: ReductionTarget,
    reduced_der: []u8,
    // ...
};
```

For each candidate component, rebuild the same case with only that DER replaced, preserve the complete observed result, and choose the smallest valid reproduction. The artifact must record the component and index so promotion can reconstruct the same chain context.

tests/pki_differential.zig
```diff
@@ -500,6 +763,129 @@ test "pki differential manifest is bounded and fully normalized" {
...
+    for (reduced_corpus.entries) |entry| {
+        const source_pem = for (sources) |source| {
+            if (std.mem.eql(u8, source.name, entry.name)) break source.source_pem;
+        } else return error.TestUnexpectedResult;
+
+        var source_chain = try pki.pem.loadChainPem(allocator, source_pem, .{});
+        defer source_chain.deinit(allocator);
+        try testing.expectEqual(@as(usize, 1), source_chain.certificates.len);
+
+        const expected_error = entry.expected_parse_error orelse return error.TestUnexpectedResult;
```
Jaruso: **[P1] The documented promotion API cannot handle parse-success or non-parser mismatches.**

`Entry.expected_parse_error` explicitly permits `null`, and most #348 categories—identity, path construction, signature verification, Name Constraints, path length—require the reduced certificate to parse successfully. This `orelse return` makes every such promoted entry fail `test-pki-reduce`. Simply allowing `null` would still be insufficient because the current regression would assert only “some certificate parses,” not the exact classification that made the input interesting.

Please store and replay the same oracle kind/class and chain context used by minimization. For example:

```zig
pub const Expected = union(enum) {
    parse_error: []const u8,
    tardigrade_class: struct {
        class: []const u8,
        source_case: []const u8,
    },
};
```

The regeneration test can resolve `source_case` from `pki_differential_manifest.zig`, rebuild anchors/intermediates/DNS context, and run `LeafOracle` (or the generalized component oracle) against both the source and promoted seed. The focused regression should assert that exact class, not merely parse success.

tests/pki_reduce.zig
```diff
@@ -0,0 +1,192 @@
...
+};
+
+pub const Outcome = struct {
+    /// Reduced bytes. The oracle classified exactly these bytes as
+    /// interesting. Caller owns the memory.
+    data: []u8,
+    /// Oracle invocations consumed, never above `Options.max_oracle_calls`.
+    oracle_calls: usize,
+
+    pub fn deinit(self: *Outcome, allocator: std.mem.Allocator) void {
```
Jaruso, line 40: **[P2] Return whether the budget was exhausted and whether 1-minimality was actually reached.**

The reducer intentionally returns its best partial result on budget exhaustion, but `Outcome` provides no way to distinguish that result from one that completed the final `chunk == 1` no-progress sweep. The promotion test later treats byte equality under a fixed `4096` calls as proof of 1-minimality. That proves reproducibility at that budget, not that no single deletion preserves the class—especially as the registry permits seeds up to 64 KiB and future sources can be larger.

Make the invariant explicit:

```zig
pub const Outcome = struct {
    data: []u8,
    oracle_calls: usize,
    budget_exhausted: bool,
    one_minimal: bool,
};
```

Set `one_minimal` only when reduction exits after a complete `chunk == 1` sweep with no accepted deletion (or reaches empty input), assert it for every promoted seed, and include the flags in schema-v3 metadata. If the artifact reducer exhausts its budget, that is fine, but the artifact and promotion path must not claim 1-minimality.

tests/vectors/pki/generate.sh
```diff
@@ -133,6 +133,33 @@ for certificate_file in "$out"/*.crt; do
...
+        exit 1
+      fi
+    fi
+    if grep -aEq -- '-----BEGIN ([A-Z0-9 ]+ )?PRIVATE KEY-----' "$seed"; then
+      echo "private-key material found in reduced corpus: $name" >&2
+      exit 1
+    fi
+  done < <(find "$reduced_dir" -maxdepth 1 -type f -print0)
+fi
+
```
Jaruso, line 179: **[P2] This does not make `reduced/` a closed directory or fully scan it for private keys.**

`find "$reduced_dir" -maxdepth 1 -type f` ignores subdirectories and every file beneath them. A committed path such as `reduced/stale/private-key.pem` bypasses both the allowlist and private-key scan, contradicting the code comment and PR security claim.

Iterate every entry and reject anything except an immediate regular file with an allowed name:

```bash
while IFS= read -r -d '' entry; do
  rel="${entry#"$reduced_dir/"}"
  if [[ "$rel" == */* || ! -f "$entry" ]]; then
    echo "unexpected nested/non-file entry in reduced corpus: $rel" >&2
    exit 1
  fi
  # validate manifest.zig or registered kebab-case .der seed, then scan
 done < <(find "$reduced_dir" -mindepth 1 -print0)
```

Also avoid treating a fixed-string occurrence anywhere in `manifest.zig` as authoritative registration; it can match a comment or provenance string. The Zig registry test should validate the exact filename/name mapping.

tests/pki_differential.zig
```diff
@@ -349,13 +391,168 @@ fn expectedStatus(decision: manifest.Decision) Status {
...
+};
+
 fn writeArtifact(
     allocator: std.mem.Allocator,
     runtime_identity: RuntimeIdentity,
     case: manifest.Case,
     tardigrade: Observation,
     openssl: Observation,
     go: Observation,
+    reduction: ?*const Reduction,
```
Jaruso, line 1262: **[P2] Add an automated test for the schema-v3 artifact path.**

All green tests currently stop before this branch: corpus cases match their manifest, while the offline reducer test exercises only `reduce`/`LeafOracle`. The DER/PEM writes, SHA-256 encoding, external reduced decisions, `preserves_observed_statuses`, JSON shape, and failure behavior are covered only by the PR’s manual forced-mismatch drill. A future change can break the core deliverable while CI remains green.

Refactor the destination and reduced verifier into injectable inputs so an offline test can use `std.testing.tmpDir` and precomputed observations:

```zig
fn writeArtifactTo(
    allocator: Allocator,
    dir: std.Io.Dir,
    artifact_subdir: []const u8,
    // ...
    reduced_observed: ?ObservedTriple,
) ![]u8 { ... }

test "pki reduce: schema-v3 artifact is complete and reproducible" {
    var tmp = std.testing.tmpDir(.{});
    defer tmp.cleanup();
    // write known reduced bytes, parse JSON, verify schema/version/sizes/hash,
    // DER == decoded PEM, decision tuple, and both true/false preservation cases.
}
```

The live differential test can remain responsible for obtaining the external observations; serialization and file integrity should be deterministic unit coverage.

Jaruso (commented):
## Blocking review findings

GitHub does not allow the PR author to submit a formal `REQUEST_CHANGES` review on their own pull request, so this is submitted as a comment review. I would not merge PR #453 in its current form.

I reviewed all 13 changed files against issue #348, including the existing differential manifest categories, reducer behavior, promotion registry, build wiring, artifact schema, CI uploads, fixture regeneration guard, and the completed CI run.

The reducer implementation itself is deterministic, allocation cleanup is sound, and the current GitHub CI run is green. The current promotion pipeline does not yet reliably satisfy the two relevant acceptance-criterion guarantees:

1. A file advertised as the reduced mismatch is not guaranteed to reproduce the original three-validator result; the code can knowingly emit `preserves_observed_statuses: false` and still present the file as promotable.
2. The reduction/promotion model is limited to the leaf and, in practice, to parser-error seeds. Issue #348 explicitly covers parsing, identity matching, path construction, and validation across certificate chains.
3. The claimed 1-minimal invariant is not represented or asserted when the oracle budget is exhausted.
4. The schema-v3 artifact path is only manually drilled; no automated test executes the file-writing, hash, JSON, or reduced re-verification branch.
5. The reduced-corpus “closed directory” guard can be bypassed with nested content.

The inline comments include concrete implementation options. After those are addressed, I would expect the promotion registry to preserve the exact Tardigrade class and relevant chain context, every emitted reduced artifact to remain a reproduction of the observed mismatch, and the tests to prove—not infer—the stated invariants.

Minor follow-ups while touching this code: `MarkerOracle.keeps` introduces a new `std.mem.indexOf` despite `docs/CODE_REVIEW_CHECKLIST.md` explicitly prohibiting new uses; registry hygiene should also validate that `source_case` resolves to a real differential case, reject duplicate seed bytes, and enforce actual kebab-case boundaries rather than only the allowed character set.

tests/pki_differential.zig
Jaruso: Fixed in 84ce42bb. `verifyAndPersistReduced` now re-verifies the reduced case three-way and, when any observed status diverges from the original tuple, reverts to the original component bytes and rewrites the files — an emitted fixture is always a reproduction. The disqualifying decisions are kept in `candidate_observed` with `reverted_external_divergence: true`, and `promotable` requires the tuple to survive on the emitted bytes. I implemented the fallback rather than the external-aware second reduction stage: with the revert in place the invariant holds unconditionally, and the second stage only buys extra shrink at ~2 process spawns per oracle call — noted as a possible follow-up if reduced sizes ever matter more.

tests/pki_differential.zig
Jaruso: Fixed in 84ce42bb, following the suggested model. `minimizeCase` loads every chain certificate's DER once (`CaseInputs`), then runs a class-preserving reduction pass per component — leaf, each intermediate, each root — via `ComponentOracle`, which rebuilds the chain in memory around the substituted slot (`tardigradeChainObservation` re-derives trust anchors from raw DER, so even anchor bytes are reducible). The largest shrink wins with deterministic tie-breaking; the artifact records `component` (e.g. `intermediate[0]`) and `components_tried`, and `persistReducedCase` writes a substituted bundle file so the external validators verify the same chain context. The offline test "component oracle substitutes any chain component" proves substitution is live for all three slots.

tests/pki_differential.zig
Jaruso: Fixed in 84ce42bb with the suggested union: `Expected = parse_error | tardigrade_class`. `tardigrade_class` seeds parse successfully and the regression replays the exact recorded class — path building, validation, identity — through `tardigradeLeafObservation` in the seed's chain context; a second promoted seed (`corrupt-certificate-signature`, class `reject|signature_invalid certificate_index=0`) exercises this path end to end, including byte-for-byte regeneration under the pipeline oracle. `source_case` must resolve in `pki_differential_manifest.zig`, and the embedded replay context is pinned to the resolved case's DNS identity so they can't drift. One deviation from the suggestion: chain context comes from per-seed embedded fixtures rather than reading the case's files at test time, keeping `zig build test` hermetic (no cwd dependence); the manifest resolution + DNS pin provide the consistency check.

tests/pki_reduce.zig
Jaruso: Fixed in 84ce42bb exactly as suggested: `Outcome` gains `budget_exhausted` and `one_minimal`, with `one_minimal` set only after a completed chunk==1 sweep with no accepted deletion (or empty output) and never when the budget broke the search. Both flags land in the schema-v3 artifact; the reproduction test asserts `one_minimal and !budget_exhausted` for every promoted seed under a budget generous enough to complete the sweep. The forced-mismatch drill now honestly reports `budget_exhausted: true, one_minimal: false` for the 537-byte leaf at the 512-call runtime budget — no minimality claim without a completed proof.

tests/vectors/pki/generate.sh
Jaruso: Fixed in 84ce42bb. The guard now walks `find -mindepth 1` and rejects nested paths, symlinks, and non-regular files; seed names must be strict kebab-case; registration matches the structured `entry("<name>", .{` line rather than any string occurrence, and a reverse pass requires a seed file for every registered entry. The registry itself now derives each embed path from the entry name via a comptime constructor, so an entry structurally cannot point at a differently named seed. Negative-tested all three violation classes (nested dir, unregistered seed, non-kebab name) — each fails the script. Zig-side hygiene additionally rejects duplicate seed bytes and kebab boundary violations.

tests/pki_differential.zig
Jaruso: Fixed in 84ce42bb along the suggested lines: `persistReducedCase` and `writeArtifact` take an injectable `DirCompat` + artifact subdirectory and precomputed observations (external verification lives separately in `verifyAndPersistReduced`). The new offline test "schema-v3 artifact is complete and reproducible" drives both against `std.testing.tmpDir` with fabricated observations and asserts schema version, component label, sizes (original/reduced/candidate), SHA-256, DER == decoded PEM round-trip, reduced-case input mapping, and both variants — preserved leaf reduction and reverted intermediate reduction with `candidate_observed` populated. It runs under `zig build test`, so serialization can no longer break while CI stays green.

tests/pki_differential.zig
```diff
@@ -349,21 +391,647 @@ fn expectedStatus(decision: manifest.Decision) Status {
...
+            // it came from a file-load failure); nothing to preserve.
+            error.UninterestingInput => {
+                if (best) |*candidate| candidate.outcome.deinit(allocator);
+                allocator.free(target_class);
+                inputs.deinit(allocator);
+                return null;
+            },
+        };
+        const shrink = original.len - outcome.data.len;
+        if (best == null or shrink > best.?.shrink) {
```
Jaruso: **[P1] Choose among candidates that reproduce the full validator tuple, not only the largest Tardigrade-only shrink.**

At this point every other component outcome is deinitialized. Only the selected candidate is later sent through OpenSSL and Go. If this largest in-process shrink changes either external status, `verifyAndPersistReduced` reverts it to the original bytes and never considers a slightly smaller candidate from another component that may already preserve the complete disagreement.

For example, suppose the leaf shrinks by 300 bytes but makes OpenSSL return `tool_failure`, while `intermediate[0]` shrinks by 200 bytes and preserves `(reject, reject, accept)`. This code selects the leaf, discards the intermediate outcome, and ultimately emits the unreduced leaf. The requested “choose the smallest valid reproduction” behavior is therefore still missing.

Retain each component’s final `Outcome`, externally verify those bounded final candidates, discard candidates that do not preserve Tardigrade’s exact class plus the original OpenSSL/Go statuses, and then select the greatest valid shrink:

```zig
const Candidate = struct {
    component: Component,
    outcome: reduce_mod.Outcome,
    shrink: usize,
    verified: ?ReducedVerification = null,
};

// After all in-process passes:
for (candidates.items) |*candidate| {
    candidate.verified = try verifyCandidate(...);
    if (!preservesTarget(candidate.verified.?, target)) continue;
    if (best == null or candidate.shrink > best.?.shrink) best = candidate;
}
```

A later external-aware refinement within the winning component can remain optional, but do not discard already-available valid component reductions before external verification.

tests/pki_differential.zig
```diff
@@ -349,21 +391,647 @@ fn expectedStatus(decision: manifest.Decision) Status {
...
+            .component = component,
+            .dns_name = case.dns_name,
+            .target_class = target_class,
+        };
+        var outcome = reduce_mod.reduce(
+            allocator,
+            original,
+            &oracle,
+            ComponentOracle.keeps,
+            .{ .max_oracle_calls = max_reduction_oracle_calls },
```
Jaruso: **[P2] This is a per-component budget, not the documented hard budget per mismatch, and the artifact under-reports the work.**

The loop can try one leaf, eight intermediates, and eight roots. Each gets 512 calls, so a hostile mismatch can consume as many as `17 * 512 = 8,704` full in-process chain classifications. `Reduction.oracle_calls` then records only the chosen component’s calls and schema v3 reports `max_oracle_calls: 512`, making the actual work invisible.

Either enforce one shared budget across the mismatch:

```zig
var remaining = max_reduction_oracle_calls;
for (components) |component| {
    const outcome = try reduce(..., .{ .max_oracle_calls = remaining });
    total_calls += outcome.oracle_calls;
    remaining -= outcome.oracle_calls;
    if (remaining == 0) break;
}
```

or explicitly rename the setting to `max_oracle_calls_per_component`, add a separately bounded `max_total_oracle_calls`, accumulate `total_oracle_calls`, and serialize both. Given the 1 MiB fixture limit, this needs to be an enforced bound rather than only `components_tried` metadata.

tests/pki_differential.zig
```diff
@@ -349,21 +391,647 @@ fn expectedStatus(decision: manifest.Decision) Status {
...
+                .openssl = .of(verified.openssl),
+                .go = .of(verified.go),
+            },
+            .candidate_observed = if (verified.candidate_tardigrade != null) .{
+                .tardigrade = .of(verified.candidate_tardigrade.?),
+                .openssl = .of(verified.candidate_openssl.?),
+                .go = .of(verified.candidate_go.?),
+            } else null,
+            .preserves_observed_statuses = verified.preserves,
+            .promotable = verified.preserves,
```
Jaruso: **[P1] `verified.preserves` is not sufficient to mark an artifact promotable.**

This is true after `revertToOriginal`, because the original bytes naturally recover the original statuses. It is also true for a budget-exhausted partial reduction and for a zero-byte shrink. All three cases are currently advertised as promotable even though the documented promotion test requires byte-for-byte regeneration under a completed 1-minimal search.

There is a related metadata bug: `revertToOriginal` replaces `reduced_der` but leaves `one_minimal` unchanged. If the discarded candidate completed its byte sweep, the JSON claims the emitted original bytes are 1-minimal using a proof that applied to different bytes.

Promotion should require all relevant invariants:

```zig
const emitted_is_reduced = !r.reverted_external_divergence and
    r.reduced_der.len < r.original_size;
const promotable = verified.preserves and
    emitted_is_reduced and
    r.one_minimal and
    !r.budget_exhausted and
    tardigradeClassMatches(verified.tardigrade, r.target_class);
```

On revert, either set `one_minimal = false` or keep separate `candidate_one_minimal` and `emitted_one_minimal` fields. Add tests for a completed one-minimal candidate that diverges externally, and for a status-preserving candidate that exhausted its budget; neither emitted artifact should be promotable.

tests/pki_differential.zig
```diff
@@ -500,6 +1221,517 @@ test "pki differential manifest is bounded and fully normalized" {
...
+};
+
+test "pki reduce: promoted registry seeds reproduce byte-for-byte and are 1-minimal" {
+    const allocator = testing.allocator;
+    comptime std.debug.assert(seed_sources.len == reduced_corpus.entries.len);
+
+    for (reduced_corpus.entries) |entry| {
+        // The reconstruction below rebuilds the chain around the leaf slot;
+        // extend it alongside the first intermediate- or root-placed seed.
+        try testing.expect(entry.placement == .leaf);
```
Jaruso: **[P1] The documented promotion path still cannot accept intermediate or root reductions.**

The runtime can now emit `component = intermediate[n]` or `root[n]` and schema v3 can mark it promotable, but adding the corresponding registry entry necessarily fails here. `SeedContext` always treats `source_pem` as the leaf and the replay test always calls `LeafOracle`, so `Placement` is currently metadata without an executable promotion implementation.

Generalize the embedded source context to raw root/intermediate/leaf bundles and select the source component from `entry.placement`. Regeneration and focused replay should use `ComponentOracle`:

```zig
const component = placementToComponent(entry.placement);
const source_der = context.inputs.componentDer(component);
const oracle = ComponentOracle{
    .inputs = &context.inputs,
    .component = component,
    .dns_name = context.dns_name,
    .target_class = expected_class,
};
```

For parse-error seeds, apply the parse oracle to the placed source component. Until this is implemented, intermediate/root artifacts must not be marked promotable.

.github/workflows/ci.yml
```diff
@@ -196,7 +196,10 @@ jobs:
         uses: actions/upload-artifact@ea165f8d65b6e75b540449e92b4886f43607fa02 # v4
         with:
           name: pki-differential-core-mismatches
-          path: pki-differential-artifacts/*.json
+          path: |
+            pki-differential-artifacts/*.json
+            pki-differential-artifacts/*.reduced.der
+            pki-differential-artifacts/*.reduced.crt
```
Jaruso: **[P1] Upload the substituted root/intermediate bundle required by `reduced_case`.**

`persistReducedCase` writes `<case>.reduced-intermediates.crt` or `<case>.reduced-roots.crt` when a non-leaf component wins, and the JSON’s `reduced_case` points directly to that file. These globs upload only the standalone component (`*.reduced.der` / `*.reduced.crt`), so the downloaded CI artifact is missing the actual bundle consumed by OpenSSL, Go, and Tardigrade and cannot reproduce the reduced case.

Include the full bounded artifact directory, or at least the generated bundle patterns:

```yaml
path: |
  pki-differential-artifacts/*.json
  pki-differential-artifacts/*.reduced.der
  pki-differential-artifacts/*.reduced*.crt
```

Apply the same fix to `.github/workflows/pki-differential.yml`. A test or workflow assertion should verify that every path named under `reduction.reduced_case` is included in the upload set.

tests/pki_differential.zig
```diff
@@ -500,6 +1221,517 @@ test "pki differential manifest is bounded and fully normalized" {
...
+            .components_tried = 3,
+            .budget_exhausted = false,
+            .one_minimal = true,
+            .target_class = try allocator.dupe(u8, "reject|OriginalReject"),
+        };
+        reduction.inputs.roots[0] = try allocator.dupe(u8, "ROOT-BYTES");
+        reduction.inputs.intermediates[0] = try allocator.dupe(u8, drill_intermediate_original);
+        defer reduction.deinit(allocator);
+
+        var verification = ReducedVerification{
```
Jaruso, line 2230: **[P2] Exercise the verification/fallback policy instead of constructing its output by hand.**

This test validates persistence and JSON serialization, but it manually supplies `.preserves = true` and the reverted case manually calls `revertToOriginal`. It never runs `verifyAndPersistReduced`, so incorrect promotion derivation, stale minimality state after revert, and candidate-selection behavior remain green—the issues above are examples that this test currently cannot catch.

Extract the pure policy from process execution, or inject a candidate verifier. Then drive at least these cases through the real state transition:

1. reduced candidate preserves the exact Tardigrade class and all statuses;
2. one candidate diverges but another component candidate preserves;
3. all candidates diverge and the emitted original is reproducible but non-promotable;
4. a one-minimal candidate is reverted, proving the emitted minimality flag is false;
5. a budget-exhausted candidate preserves statuses but is non-promotable.

The serialization test can continue using fabricated observations, but the policy that produces `ReducedVerification` and `promotable` also needs deterministic coverage.

Jaruso (commented):
## Second-pass blocking review

The follow-up commit substantially improves the design: the reducer now exposes budget/minimality state, chain components can be substituted in process, parse-success seeds replay an exact pipeline class, the reduced directory is genuinely closed, and deterministic artifact serialization is covered. The current CI run is green across the full matrix.

GitHub prevents the PR author from selecting `REQUEST_CHANGES` on their own pull request, so this is a comment review. I still would not merge this revision yet.

The remaining blockers are in the interaction between candidate selection, external verification, promotion metadata, and CI artifact packaging:

- `minimizeCase` throws away all but the largest Tardigrade-only shrink before OpenSSL/Go verification. If that one candidate diverges, the code reverts to the original bytes without trying another component that may already contain a valid reduced reproduction.
- `promotable` currently means only “the emitted statuses match.” It can therefore be true after reverting to the original bytes, after an incomplete 512-call search, or when no byte was removed. That contradicts the documented promotion workflow and can produce a registry entry that immediately fails the 1-minimal regeneration test.
- Reverting the bytes does not invalidate the candidate’s `one_minimal` flag, so schema v3 can claim that the emitted original certificate is 1-minimal based on a proof that applied to different bytes.
- The stated 512-call hard cap is applied independently to every chain component, while the artifact reports only the selected component’s calls. With the configured limits, a mismatch can consume up to 8,704 in-process oracle calls while metadata says 512.
- Intermediate/root substitutions require generated bundle files, but neither mismatch-upload workflow includes those files.
- The registry declares intermediate/root placement support, while the mandatory promotion test rejects every non-leaf entry unconditionally.

The inline comments include concrete implementation paths and tests. The previous six threads are mostly addressed at the architectural level, but the new logic needs these invariants tightened before the acceptance criterion can be considered complete.

Jaruso (commented):
## Third-pass blocking review

GitHub does not allow the PR author to submit a formal `REQUEST_CHANGES` review on their own pull request, so this is again recorded as a comment review. I would not merge PR #453 yet.

I reviewed the force-pushed/rebased head `8d6fefad`, the six findings from the previous pass, the current promotion registry and component replay, both artifact-upload workflows, reducer metadata, serialization tests, and CI run 715. CI is fully green.

The latest commit correctly fixes most of the previous mechanics:

- one enforced 512-call mismatch-level ceiling is now recorded honestly;
- every retained candidate is checked against Tardigrade's exact class plus the OpenSSL/Go status tuple;
- reverted, zero-shrink, budget-exhausted, and non-minimal outputs are no longer marked promotable;
- intermediate/root placement is executable in regeneration and focused regression tests;
- substituted bundles are included in CI uploads; and
- minimality is cleared on revert.

Three behavioral gaps remain:

1. The shared budget is consumed sequentially, leaf first. A normal certificate-sized leaf can use all 512 calls, so no intermediate or root is attempted; this behaviorally recreates the original leaf-only limitation.
2. Among full-tuple reproductions, selection still ranks only by byte shrink. A larger budget-exhausted/non-minimal candidate can displace a smaller candidate that is immediately promotable, causing the artifact to report `promotable: false` even though a promotable result was already found.
3. The final selected candidate is externally verified a second time, but a false result on that final run is returned rather than reverted or failed, so the claimed “every emitted fixture reproduces” invariant is not unconditional.

The policy that creates these states is still not under deterministic automated coverage: the schema test manually constructs `ReducedVerification` and manually calls `revertToOriginal` rather than executing `verifyAndPersistReduced` or an extracted pure selection policy. The inline comments include concrete fixes and tests.

There is also an allocation-error cleanup gap in the candidate-verification loop, noted inline. Finally, the PR description is now materially stale: it still says only the leaf is reduced, external validators run once, and promoted regressions only pin parse outcomes. Please update the description/security and performance notes to match the component-aware multi-candidate implementation before merge.

tests/pki_differential.zig
```diff
@@ -349,21 +391,811 @@ fn expectedStatus(decision: manifest.Decision) Status {
...
+            .component = component,
+            .dns_name = case.dns_name,
+            .target_class = target_class,
+        };
+        const outcome = reduce_mod.reduce(
+            allocator,
+            original,
+            &oracle,
+            ComponentOracle.keeps,
+            .{ .max_oracle_calls = remaining },
```
Jaruso: **[P1] Reserve budget for later components; this is still leaf-only in normal cases.**

`remaining` is handed wholesale to the current component, and components are ordered leaf first. A certificate-sized leaf that does not complete reduction within 512 calls consumes the entire mismatch budget, so the loop exits before trying any intermediate or root. The previous forced-mismatch drill already demonstrated exactly this shape for a 537-byte leaf (`budget_exhausted: true` at 512 calls).

That means Name Constraints, path-length, and cross-signing mismatches will usually spend all work proving that the leaf cannot shrink while never touching the component containing the relevant behavior.

Use a fair bounded allocation or a two-phase schedule. A simple baseline reserves a share for every remaining component:

```zig
for (components[0..component_count], 0..) |component, component_index| {
    const remaining = max_total_reduction_oracle_calls - total_oracle_calls;
    const components_left = component_count - component_index;
    if (remaining < components_left) break; // or record untried explicitly

    // Earlier components cannot monopolize work; unused calls naturally flow
    // to later components as `remaining` is recalculated.
    const allowance = @max(remaining / components_left, 1);
    const outcome = try reduce_mod.reduce(
        allocator,
        original,
        &oracle,
        ComponentOracle.keeps,
        .{ .max_oracle_calls = allowance },
    );
}
```

A more effective version can do a small coarse pass over every component and spend the remainder on candidates that actually shrink. Add a deterministic test where the leaf oracle exhausts its allocation but an intermediate candidate still runs and wins.

tests/pki_differential.zig
```diff
@@ -349,21 +391,811 @@ fn expectedStatus(decision: manifest.Decision) Status {
...
+            allocator,
+            tardigrade,
+            openssl,
+            go,
+            original_statuses,
+            r.target_class,
+        );
+        const shrink = candidate.shrink();
+        if (preserves) {
+            if (best_index == null or shrink > best_shrink) {
```
Jaruso: **[P1] Prefer a promotable reproduction over a larger partial reproduction.**

This ranking considers only `shrink`. Suppose candidate A preserves the full validator tuple and shrinks 200 bytes but exhausted its budget (`one_minimal = false`), while candidate B preserves the same tuple, shrinks 150 bytes, and completed a 1-minimal proof. A wins here, then `reductionIsPromotable` correctly returns false. The artifact therefore discards an already-available promotable regression fixture.

Keep two rankings: best promotable reproduction and best general reproduction. Select the promotable tier whenever it is non-empty:

```zig
const candidate_promotable = candidate.shrink() > 0 and
    candidate.one_minimal and
    !candidate.budget_exhausted;

if (preserves and candidate_promotable) {
    if (best_promotable == null or shrink > best_promotable_shrink) {
        best_promotable = index;
        best_promotable_shrink = shrink;
    }
} else if (preserves) {
    if (best_reproduction == null or shrink > best_reproduction_shrink) {
        best_reproduction = index;
        best_reproduction_shrink = shrink;
    }
}

const selected = best_promotable orelse best_reproduction;
```

This is necessary for the acceptance criterion: when the bounded search has already found an input ready to feed into unit/fuzz regression suites, a larger incomplete search result must not hide it.

tests/pki_differential.zig
```diff
@@ -349,21 +391,811 @@ fn expectedStatus(decision: manifest.Decision) Status {
...
+        var tardigrade, var openssl, var go = try verifyReducedCase(allocator, case, &paths);
+        errdefer tardigrade.deinit(allocator);
+        errdefer openssl.deinit(allocator);
+        errdefer go.deinit(allocator);
+        return .{
+            .paths = paths,
+            .tardigrade = tardigrade,
+            .openssl = openssl,
+            .go = go,
+            .preserves = try observationsPreserveTarget(
```
Jaruso: **[P2] Do not return a final candidate when the final verification no longer preserves the target.**

The candidate was accepted during the first verification pass, but it is persisted and externally verified again here. If the second OpenSSL/Go invocation transiently returns `tool_failure`, or otherwise differs, this expression becomes false and the function still returns the reduced bytes. The artifact is non-promotable, but it no longer satisfies the stronger documented invariant that every emitted reduced fixture reproduces the mismatch.

Either retain the winning observations from the first pass and reuse them after deterministically re-persisting the same bytes, or check the final result and run the same original-byte fallback before returning:

```zig
const final_preserves = try observationsPreserveTarget(...);
if (!final_preserves) {
    // deinit final observations/paths, revert, persist, and verify original
    return fallbackToOriginal(...);
}
```

The final state should be either a reproducing reduced fixture, a reproducing original fallback, or an artifact-persistence failure—never a successfully returned non-reproduction.

tests/pki_differential.zig
```diff
@@ -349,21 +391,811 @@ fn expectedStatus(decision: manifest.Decision) Status {
...
+    errdefer if (fallback_openssl) |*obs| obs.deinit(allocator);
+    var fallback_go: ?Observation = null;
+    errdefer if (fallback_go) |*obs| obs.deinit(allocator);
+
+    for (r.candidates, 0..) |*candidate, index| {
+        try r.selectCandidate(allocator, candidate);
+        var paths = try persistReducedCase(allocator, dir, artifact_dir, case, r);
+        defer paths.deinit(allocator);
+        var tardigrade, var openssl, var go = try verifyReducedCase(allocator, case, &paths);
+        const preserves = try observationsPreserveTarget(
```
Jaruso, line 1123: **[P2] Guard the candidate observations before the class-comparison allocation.**

`verifyReducedCase` returns three owned diagnostics. `observationsPreserveTarget` can then fail with `OutOfMemory` while allocating the Tardigrade class string, and none of these observations has an `errdefer` yet, so all three diagnostics leak on that error path.

```zig
var tardigrade, var openssl, var go = try verifyReducedCase(...);
errdefer tardigrade.deinit(allocator);
errdefer openssl.deinit(allocator);
errdefer go.deinit(allocator);

const preserves = try observationsPreserveTarget(...);
```

The existing explicit deinitialization/move logic can remain unchanged for normal control flow.

tests/pki_differential.zig
```diff
@@ -500,6 +1385,611 @@ test "pki differential manifest is bounded and fully normalized" {
...
+        .total_oracle_calls = total_oracle_calls,
+        .max_total_oracle_calls = max_total_reduction_oracle_calls,
+        .components_tried = 1,
+        .budget_exhausted = budget_exhausted,
+        .one_minimal = one_minimal,
+        .target_class = try allocator.dupe(u8, "reject|OriginalReject"),
+    };
+}
+
+test "pki reduce: schema-v3 artifact is complete and reproducible" {
```
Jaruso, line 2172: **[P2] Exercise the real selection/fallback policy, not only its serialized output.**

This test still constructs `ReducedVerification` by hand and manually invokes `revertToOriginal`; it never calls `verifyAndPersistReduced`. Consequently the shared-budget starvation, promotable-vs-partial ranking, and final re-verification behavior above all remain green.

Extract candidate ranking/fallback into a pure function, or inject a deterministic verifier into `verifyAndPersistReduced`, then cover at least:

1. the first component exhausts its fair share and a later component is still attempted;
2. a larger partial reproduction loses to a smaller one-minimal reproduction;
3. all reduced candidates diverge and original bytes are emitted non-promotably;
4. a final verification differs from the discovery verification and triggers fallback/failure;
5. allocation failure after observations are created is leak-free.

The existing tmpdir serialization assertions are useful and should remain; they test a different layer.

tests/pki_differential.zig
```diff
@@ -349,21 +391,856 @@ fn expectedStatus(decision: manifest.Decision) Status {
...
+        if (fallback_go) |*obs| {
+            obs.deinit(allocator);
+            fallback_go = null;
+        }
+        try r.selectCandidate(allocator, &r.candidates[index]);
+        const verification = best_verification.?;
+        best_verification = null;
+        return verification;
+    }
+
```
Jaruso, line 1208: **[P1] Re-persist the selected winner before returning; its canonical files may have been overwritten.**

Every candidate is persisted using the same names (`<case>.reduced.der`, `<case>.reduced.crt`, and one shared bundle name per component kind). `best_verification` retains only the path strings and observations. If an earlier candidate wins but any later candidate is verified, that later pass overwrites the files referenced by the retained paths.

For example, a promotable leaf candidate can win, then a larger partial root candidate writes its root DER into `<case>.reduced.crt`. This branch updates `r.reduced_der` back to the leaf bytes but returns the old `best_verification` without rewriting the file. `writeArtifact` then hashes the leaf bytes and serializes the leaf observations while the uploaded `.reduced.der/.crt` contain root bytes.

Restore canonical output after selection without re-running the external validators:

```zig
try r.selectCandidate(allocator, &r.candidates[index]);
var final_paths = try persistReducedCase(allocator, dir, artifact_dir, case, r);
errdefer final_paths.deinit(allocator);

var verification = best_verification.?;
best_verification = null;
verification.paths.deinit(allocator);
verification.paths = final_paths;
return verification;
```

Alternatively, verify candidates under unique temporary names and publish/copy only the winner. Add a test where candidate 0 wins and candidate 1 is persisted afterward, then assert the final DER, decoded PEM, substituted bundle, and JSON SHA all contain candidate 0.

tests/pki_differential.zig
```diff
@@ -349,21 +391,856 @@ fn expectedStatus(decision: manifest.Decision) Status {
...
+    errdefer reverted_openssl.deinit(allocator);
+    errdefer reverted_go.deinit(allocator);
+    return .{
+        .paths = reverted_paths,
+        .tardigrade = reverted_tardigrade,
+        .openssl = reverted_openssl,
+        .go = reverted_go,
+        .preserves = try observationsPreserveTarget(
+            allocator,
+            reverted_tardigrade,
```
Jaruso: **[P1] Do not successfully return an original-byte fallback that still fails reproduction.**

This comparison can return `false`, but the function still returns a `ReducedVerification` and the caller serializes/uploads it. A transient OpenSSL/Go failure—or any unexpected difference while re-reading the emitted original bundle—therefore produces a successful reduced artifact with `preserves_observed_statuses: false`, contradicting the function comment and the #348 reproducibility requirement.

Compute the value before constructing the return and treat `false` as reduction failure:

```zig
const preserves = try observationsPreserveTarget(...);
if (!preserves) {
    reverted_paths.deinit(allocator);
    reverted_tardigrade.deinit(allocator);
    reverted_openssl.deinit(allocator);
    reverted_go.deinit(allocator);
    return error.ReducedFixtureDidNotReproduce;
}
```

`runCase` can then omit the reduction record and retain the main mismatch JSON, but it should also remove any partial reduced files so the CI upload does not contain unreferenced/non-reproducing fixtures.

tests/pki_differential.zig
```diff
@@ -349,21 +391,856 @@ fn expectedStatus(decision: manifest.Decision) Status {
...
+        const verified = reduced.?;
+        component_label = try componentLabel(allocator, r.component);
+        sha256Hex(r.reduced_der, &reduced_sha256_hex);
+        reduction_json = .{
+            .component = component_label.?,
+            .original_size = r.original_size,
+            .reduced_size = r.reduced_der.len,
+            .candidate_size = r.candidate_size,
+            .oracle_calls = r.oracle_calls,
+            .max_oracle_calls = r.max_total_oracle_calls,
```
Jaruso: **[P2] Record the selected component's actual allowance, not the total mismatch cap.**

After the fair-share change, the selected candidate may have been reduced with an allowance of roughly `512 / component_count` calls. This field still reports 512, while `oracle_calls` and `budget_exhausted` describe the selected component's smaller search. That makes schema-v3 claim a candidate had substantially more minimization budget than it actually received.

Store the allowance on `ReductionCandidate` and copy it in `selectCandidate`, for example:

```zig
const ReductionCandidate = struct {
    // ...
    max_oracle_calls: usize,
};

// candidate creation
.max_oracle_calls = allowance,

// selection
self.max_oracle_calls = candidate.max_oracle_calls;
```

Then serialize that value here. Keep `max_total_oracle_calls` as the distinct whole-mismatch cap.

tests/pki_differential.zig
```diff
@@ -500,6 +1430,691 @@ test "pki differential manifest is bounded and fully normalized" {
...
+    try testing.expect(leaf_allowance + intermediate_allowance < max_total_reduction_oracle_calls);
+
+    const root_allowance = componentReductionAllowance(leaf_allowance + intermediate_allowance, 2, component_count);
+    try testing.expect(root_allowance > 0);
+    try testing.expectEqual(max_total_reduction_oracle_calls, leaf_allowance + intermediate_allowance + root_allowance);
+}
+
+test "pki reduce: candidate policy prefers promotable reproductions before larger partials" {
+    var partial_bytes: [80]u8 = undefined;
+    var promotable_bytes: [85]u8 = undefined;
```
Jaruso, line 1811: **[P2] Exercise the stateful persistence/fallback policy, not only this detached comparator.**

`selectVerifiedCandidate` is used by this test but not by production; `verifyAndPersistReduced` calls the shared comparator while also moving observations and repeatedly overwriting canonical files. The arithmetic/comparator tests therefore cannot catch the two production bugs above.

Please add deterministic coverage around the real policy, either by injecting a candidate verifier into `verifyAndPersistReduced` or by extracting a pure selection/publication state machine. At minimum cover:

1. an earlier promotable winner followed by a later non-winning candidate, verifying canonical DER/PEM/bundle bytes still belong to the winner;
2. all reduced candidates diverge and original fallback preserves, producing a non-promotable reproduction;
3. original fallback verification also diverges, producing an error/no reduction record;
4. selected component budget metadata equals its fair-share allowance.

The existing schema serialization test should remain; this is the missing layer that decides which bytes it serializes.

Jaruso (commented):
I re-reviewed head `247c47a6` against `8d6fefad`. The fair per-component allocation, promotion-ready candidate ranking, retained discovery observations, and allocation guards address the five prior findings well. The unit-test and core PKI differential jobs are green; full CI run 716 is still in progress at the time of this review.

I still would not merge this head. There are two artifact-correctness blockers and two supporting coverage/metadata issues:

1. The retained winning verification stores canonical path names, but later candidate verification overwrites those same files. Selecting the winner only updates `r.reduced_der`; it does not restore the winning DER/PEM/bundle. The JSON SHA and observations can therefore describe candidate A while the uploaded files contain candidate B.
2. The original-byte fallback still returns successfully when its final re-verification does not preserve the target. That violates the stated invariant that every emitted reduced fixture reproduces the mismatch.
3. `max_oracle_calls` now reports the total mismatch cap rather than the selected component's actual allowance, so schema-v3 can materially overstate the budget used to derive/minimality-check that candidate.
4. The new tests exercise arithmetic and the comparator, but still do not execute `verifyAndPersistReduced`; consequently the canonical-file overwrite and false fallback-success paths remain green.

I also still need the PR description updated before merge. It currently describes leaf-only minimization, one external re-verification, parser-only regression outcomes, and the original one-seed/test counts, none of which matches the current implementation.

GitHub cannot record `REQUEST_CHANGES` from the PR author's connected account, so this is submitted as a blocking comment review.

tests/pki_differential.zig
```diff
@@ -349,21 +391,901 @@ fn expectedStatus(decision: manifest.Decision) Status {
...
+    var fallback_openssl: ?Observation = null;
+    errdefer if (fallback_openssl) |*obs| obs.deinit(allocator);
+    var fallback_go: ?Observation = null;
+    errdefer if (fallback_go) |*obs| obs.deinit(allocator);
+
+    for (r.candidates, 0..) |*candidate, index| {
+        try r.selectCandidate(allocator, candidate);
+        const probe_suffix = try std.fmt.allocPrint(allocator, "candidate-{d}", .{index});
+        defer allocator.free(probe_suffix);
+        var paths = try persistReducedCase(allocator, dir, artifact_dir, case, r, probe_suffix);
```
Jaruso, line 1110: **[P2] Remove candidate probe files after each synchronous verification.**

The unique names fix canonical overwrite corruption, but every call to `persistReducedCase(..., probe_suffix)` writes DER, PEM, and possibly a substituted bundle into the real artifact directory. `paths.deinit()` only frees the path strings; it never deletes those files. A single mismatch can therefore leave files for every attempted component plus the fallback probe, even though none is referenced by the JSON or included intentionally in the upload contract. With bounded 1 MiB fixtures and up to 17 components, this can leave tens of megabytes of stale output for one failure and makes repeated local runs accumulate misleading artifacts.

Please give probe persistence an explicit lifecycle. One clean approach is to isolate probes in a temporary subtree and remove it after the verifier returns:

```zig
const probe_dir = try std.fmt.allocPrint(
    allocator,
    "{s}/.probes/{s}/candidate-{d}",
    .{ artifact_dir, case.id, index },
);
defer allocator.free(probe_dir);
defer dir.deleteTree(probe_dir) catch {};

var paths = try persistReducedCase(
    allocator,
    dir,
    probe_dir,
    case,
    r,
    null,
);
```

Alternatively, track every generated file separately and delete exactly those paths after synchronous verification; do not delete unchanged source paths referenced by `ReducedPaths`. Apply the same cleanup to `fallback-original`, and add a tmpdir assertion that successful and failed policy runs leave only the final canonical files (or no reduced files on failure).

tests/pki_differential.zig
```diff
@@ -500,6 +1475,917 @@ test "pki differential manifest is bounded and fully normalized" {
...
+    allocator: std.mem.Allocator,
+    dir: compat.DirCompat,
+    case: manifest.Case,
+    paths: *const ReducedPaths,
+) !struct { Observation, Observation, Observation } {
+    const raw = try dir.readFileAlloc(allocator, paths.der_path, 1024);
+    defer allocator.free(raw);
+    const preserves = !std.mem.eql(u8, case.id, "artifact-policy-fallback-fails") and
+        std.mem.eql(u8, raw, drill_leaf_reduced);
+    const diagnostic = if (preserves) "OriginalReject" else "DifferentReject";
```
Jaruso, line 1777: **[P2] Add the missing stateful test for a successful original-byte fallback.**

This verifier can preserve only `drill_leaf_reduced`; original `drill_leaf_original` bytes always produce `DifferentReject`. Consequently the new tests cover a reduced winner and a fallback that fails, but they still never execute the successful fallback branch in `verifyAndPersistReducedWithVerifier`.

That branch owns several important state transitions that can regress independently: selecting the recorded divergent candidate, reverting to the correct original component, clearing `one_minimal`, setting `reverted_external_divergence`, moving `candidate_observed`, publishing canonical original DER/PEM/bundle files, and deriving `promotable = false`.

Please add a paired case, for example:

```zig
const preserves = if (std.mem.eql(u8, case.id, "artifact-policy-fallback-succeeds"))
    std.mem.eql(u8, raw, drill_leaf_original)
else if (std.mem.eql(u8, case.id, "artifact-policy-fallback-fails"))
    false
else
    std.mem.eql(u8, raw, drill_leaf_reduced);
```

Then drive all reduced candidates to divergence and assert the real function returns successfully with:

```zig
try testing.expect(verification.preserves);
try testing.expect(reduction.reverted_external_divergence);
try testing.expect(!reduction.one_minimal);
try testing.expectEqualSlices(u8, drill_leaf_original, reduction.reduced_der);
try testing.expect(verification.candidate_tardigrade != null);
try testing.expect(!reductionIsPromotable(&reduction, &verification));
```

Also read the canonical DER and PEM (and a substituted bundle for a non-leaf variant) to prove the published files contain the original selected component.

Jaruso (commented):
I re-reviewed the new head `7d27e5c257714576e268dcc0721647a508714f78` against issue #348 and the previous review findings.

The latest commit correctly fixes the four prior blockers:

- candidates are verified under unique names and the selected winner is materialized under canonical `.reduced.*` paths;
- a non-reproducing original-byte fallback now fails and `runCase` omits the reduction record while preserving the main mismatch JSON;
- schema v3 records the selected component's real fair-share `max_oracle_calls`;
- the stateful policy is now injectable and the winner-publication and fallback-failure paths have direct tests.

I still have two P2 findings in the new implementation, detailed inline. The first is an artifact-lifecycle defect introduced by the probe-file approach; the second leaves the successful fallback state transition—the path that produces a non-promotable but reproducible original fixture—untested.

The PR description also remains materially stale. It still says only the leaf is reduced, describes parser-only/parse-outcome promotion, references only the first seed, gives old test counts, and does not describe per-component fair-share budgeting, candidate ranking, original fallback semantics, or canonical winner publication. Please update it before merge so the stated behavior matches the code and issue #348 acceptance criteria.

CI run 717 is still in progress. Format, lint, Linux/macOS/ARM unit tests, and the core PKI differential job are already green.

tests/pki_differential.zig
```diff
@@ -349,21 +391,925 @@ fn expectedStatus(decision: manifest.Decision) Status {
...
+        if (self.intermediate_file) |path| allocator.free(path);
+        allocator.free(self.leaf_file);
+        self.* = undefined;
+    }
+};
+
+fn deleteIfCandidatePath(dir: compat.DirCompat, path: []const u8) !void {
+    if (std.mem.find(u8, path, ".candidate") == null) return;
+    dir.deleteFile(path) catch |err| switch (err) {
+        error.FileNotFound => {},
```
Jaruso, line 928: **[P1] Do not identify generated files by searching arbitrary path text.**

`ReducedPaths` contains both probe-owned outputs and unchanged source inputs. For a leaf probe, `root_file` and `intermediate_file` are copied directly from the manifest; for an intermediate probe, `root_file` and `leaf_file` are source paths; and for a root probe, `intermediate_file` and `leaf_file` are source paths. Any legitimate source filename or parent directory containing `.candidate` therefore passes this test and is deleted from the working tree.

For example, a case using `tests/vectors/pki/candidate-certs/root.crt` is safe, but `tests/vectors/pki/.candidate-certs/root.crt` or `root.candidate.crt` would be removed after verification even though the cleanup never created it. Manifest validation does not prohibit those paths.

Track ownership structurally instead of inferring it from a substring. Since persistence already knows the selected component, a minimal fix is:

```zig
fn deleteFileIfPresent(dir: compat.DirCompat, path: []const u8) !void {
    dir.deleteFile(path) catch |err| switch (err) {
        error.FileNotFound => {},
        else => return err,
    };
}

fn deleteProbeFiles(
    dir: compat.DirCompat,
    component: Component,
    paths: *const ReducedPaths,
) !void {
    try deleteFileIfPresent(dir, paths.der_path);
    try deleteFileIfPresent(dir, paths.pem_path);
    switch (component) {
        .leaf => {},
        .intermediate => try deleteFileIfPresent(dir, paths.intermediate_file.?),
        .root => try deleteFileIfPresent(dir, paths.root_file),
    }
}
```

An even clearer model is for `persistReducedCase` to return a separate list of generated paths. Add a tmpdir test whose unchanged source root/leaf path includes `.candidate`; verify cleanup removes every probe output while preserving that source file.

Jaruso (commented):
The prior two findings are addressed: candidate/fallback probe files are now cleaned, the successful original-byte fallback runs through the real verification state machine, the PR description is current, and CI run 718 is green.

I found one remaining destructive cleanup bug. The helper infers ownership from a substring in the path, but `ReducedPaths` intentionally contains both generated probe paths and unchanged source fixture paths. A valid source filename or parent directory containing `.candidate` can therefore be deleted while cleaning a probe. Please delete only paths known to have been generated for the selected component, rather than inspecting arbitrary path text.

Because this can remove checked-in/source test inputs from the working tree, I do not consider the PR ready to merge until it is fixed and covered by a test with a source path containing `.candidate`.

Jaruso (commented):
## Final review: no findings

I re-reviewed the latest delta and the surrounding reduction/persistence state machine at `dd3f386`.

The last blocker is correctly fixed:

- cleanup no longer infers ownership from a `.candidate` substring;
- DER and PEM probe files are known generated outputs;
- a substituted root/intermediate bundle is tracked explicitly through `generated_bundle_file`;
- unchanged source root, intermediate, and leaf paths are never passed to deletion;
- the regression test proves that a real source path containing `.candidate` survives cleanup.

I also rechecked the alias lifetime and deinitialization behavior. `generated_bundle_file` aliases the already-owned `root_file` or `intermediate_file`, is consulted before `ReducedPaths.deinit`, and is intentionally not freed separately, so there is no double free or dangling use in the current call flow.

The previously reviewed invariants remain intact: fair bounded reduction, promotion-ready candidate preference, external tuple verification, canonical winner publication, successful and failed original-byte fallbacks, non-promotable revert metadata, SHA/file consistency, probe cleanup, and schema-v3 serialization.

**No further changes requested. This is ready to merge from a code-review perspective once the current CI run completes successfully.**

PR #454, opened 2026-07-17
tls: relocate the concrete TLS 1.3 engine out of QUIC

Jaruso:
## Summary

- relocate the concrete reusable TLS 1.3 handshake engine to `src/tls/tls13_backend.zig`
- make transport differences explicit with `.record` and opaque-extension profiles
- reduce `src/quic/tls_backend.zig` to QUIC transport parameters, CID binding, epoch/error translation, and shared-engine construction
- construct record mode directly from the TLS-owned backend with configurable TCP ALPN and no QUIC types or extensions

## Completion report

1. **Final module graph:** `tls_core` owns the concrete TLS 1.3 engine and imports no QUIC or OpenSSL module. QUIC imports `tls_core` and wraps the engine through `src/quic/tls_backend.zig`. Record mode imports only TLS-owned types and injects the concrete engine directly into `PureZigRecordStream`.
2. **Moved versus changed:** ClientHello, ServerHello, EncryptedExtensions, certificate, CertificateVerify, Finished, transcript progression, key schedule, traffic-secret export, and teardown moved from `src/quic/tls_backend.zig` to `src/tls/tls13_backend.zig`. Protocol behavior is preserved; transport selection and fail-closed ALPN/certificate errors are now explicit shared-engine behavior.
3. **Transport profile design:** `.record` carries only configured ALPN. `.extension` carries an adapter-owned opaque extension type/payload plus ALPN. The shared engine never imports or decodes QUIC transport types.
4. **Remaining QUIC adapter:** RFC 9000 transport-parameter encoding/decoding, CID-binding state, QUIC/shared epoch conversion, QUIC error mapping, `h3` configuration, and construction of the shared engine in extension mode.
5. **Native record construction:** `src/tls/tls13_backend_tests.zig` constructs the shared backend with `.record`, injects `backend()` directly into `PureZigRecordStream`, negotiates `h2`, and verifies no peer transport extension exists.
6. **Removed temporary switches:** the cross-module `RecordModeBackend` event translator, dummy QUIC transport parameters, and mutable `omit_transport_parameters` switch are gone.
7. **Lifecycle and zeroization:** transport drivers own exactly-once backend teardown; the shared engine clears key schedules, KDF intermediates, temporary application traffic secrets, Finished material, X25519 shares, peer certificates, opaque extension scratch, reassembly buffers, and transcript/core state. QUIC-local parameter scratch and both transport event sinks are also wiped.
8. **Validation matrix:**
   - `git diff --check`
   - `zig build test-tls --summary all --error-style verbose` — 160/160 passed
   - `zig build test-quic --summary all --error-style verbose` — 311/311 passed, including deterministic QUIC/H3 and UDP smoke coverage
   - `zig build test-integration --summary all --error-style verbose` — 65 passed, 4 skipped
   - `zig build test --summary all --error-style verbose` — 1,393 passed, 1 skipped
   - `zig build test -Dtls-profile=appliance --summary all --error-style verbose` — 1,378 passed, 1 skipped; no OpenSSL dependency is pulled into the TLS-owned engine

Closes #442


src/tls/tls13_backend.zig
```diff
@@ -0,0 +1,1164 @@
...
+        try w.u16_(2 + 2 + X25519.public_length); // client_shares
+        try w.u16_(group_x25519);
+        try w.u16_(X25519.public_length);
+        try w.bytes(&key_pair.public_key);
+
+        const negotiated_alpn = self.alpn();
+        try w.u16_(ext_alpn);
+        const alpn_ext_len = try w.reserve(2);
+        const alpn_list_len = try w.reserve(2);
+        try w.u8_(@intCast(negotiated_alpn.len));
```
Jaruso, line 632: **[P1] Validate the transport profile before serializing it**

The new record-mode/profile API accepts arbitrary caller slices, but this cast assumes `alpn.len <= 255`. In Debug/ReleaseSafe, a 256-byte ALPN traps here instead of returning a typed handshake/configuration error. A zero-length ALPN is also encoded even though RFC 7301 `ProtocolName` is `opaque<1..2^8-1>`.

The same problem exists for `.extension.local`: direct construction bypasses `setExtensionProfile()`'s 512-byte check, a payload over `u16` traps at its length cast, and a payload between 513 bytes and the writer capacity is sent even though the peer-side capture limit is 512—creating a profile that cannot interoperate with itself. An extension ID that collides with an engine-owned extension (for example `ext_alpn`) also makes the engine emit duplicate extensions.

Please make validation unavoidable, preferably before `core.start()` mutates the transcript/lifecycle (or make the constructors fallible). For example:

```zig
pub const ProfileError = error{InvalidTransportProfile};

fn validate(profile: TransportProfile) ProfileError!void {
    const protocol = profile.alpn();
    if (protocol.len == 0 or protocol.len > std.math.maxInt(u8))
        return error.InvalidTransportProfile;

    switch (profile) {
        .record => {},
        .extension => |ext| {
            if (ext.local.len > max_transport_extension_len)
                return error.InvalidTransportProfile;
            if (switch (ext.extension_type) {
                ext_supported_groups,
                ext_signature_algorithms,
                ext_alpn,
                ext_supported_versions,
                ext_key_share,
                => true,
                else => false,
            }) return error.InvalidTransportProfile;
        },
    }
}

fn startImpl(...) HandshakeError!void {
    const self: *Tls13Backend = ...;
    try self.profile.validate();
    // Only then transition `core` and emit bytes.
    ...
}
```

Please add deterministic tests for ALPN lengths 0 and 256, a local extension larger than `max_transport_extension_len`, and a TLS-owned extension-type collision. None of these should panic or partially advance the handshake.

src/tls/tls13_backend.zig
```diff
@@ -0,0 +1,1164 @@
...
+        self.peer_transport_extension_pending = false;
+        return self.peer_transport_extension[0..self.peer_transport_extension_len];
+    }
+
+    fn deinitImpl(ptr: *anyopaque) void {
+        const self: *Tls13Backend = @ptrCast(@alignCast(ptr));
+        self.deinit();
+    }
+
+    pub fn deinit(self: *Tls13Backend) void {
```
Jaruso, line 408: **[P1] Wipe all backend-owned private-key material on teardown**

`deinit()` wipes the generated `X25519.KeyPair`, but it leaves `entropy.key_share_seed` intact. Because the pair is created with `X25519.KeyPair.generateDeterministic(self.entropy.key_share_seed)`, that seed reproduces the exact private key; leaving the backend's copy resident defeats the key-share zeroization acceptance criterion.

The server backend also owns an `Identity` by value, so its private signing key remains in the backend after teardown. The caller may retain another copy, but that does not justify leaving this owned copy behind.

Please centralize private-material cleanup and invoke it on every success/failure/cancellation teardown path. The ephemeral key pair and seed can also be cleared immediately after the shared secret/key schedule has been derived, because neither is needed afterward. For example:

```zig
fn wipeEphemeral(self: *Tls13Backend) void {
    if (self.key_pair) |*pair| {
        std.crypto.secureZero(u8, &pair.secret_key);
        std.crypto.secureZero(u8, &pair.public_key);
    }
    self.key_pair = null;
    std.crypto.secureZero(u8, &self.entropy.key_share_seed);
}

fn wipeIdentity(self: *Tls13Backend) void {
    if (self.identity) |*identity| {
        std.crypto.secureZero(u8, std.mem.asBytes(&identity.key));
    }
    self.identity = null;
}
```

Add post-teardown tests that poison/check the deterministic seed, generated key pair, and server identity on completed, failed, and abandoned handshakes. The current teardown test only checks `expected_client_verify` and peer-certificate storage, so it would not catch this residue.

src/tls/tls13_backend.zig
```diff
@@ -0,0 +1,1164 @@
...
+//!   - signature: Ed25519 (server CertificateVerify)
+//!   - server-only authentication; client certificates are not offered
+//!   - trust: an explicitly pinned certificate (deterministic fixture /
+//!     deployment pin) or an explicit insecure mode reporting `not_checked`
+//! Session resumption, 0-RTT, HelloRetryRequest, and web-PKI chain validation
+//! are follow-ups (see docs/QUIC_TLS.md). Entropy is caller-supplied like the
+//! rest of the native transport stack — no ambient RNG.
+
+const std = @import("std");
+const encrypted_stream = @import("encrypted_stream.zig");
```
Jaruso: **[P2] Put the shared engine on a transport-neutral TLS contract**

The engine is TLS-owned now, but it imports the high-level record-stream implementation and defines its public backend/sink/error types through `encrypted_stream.Record*`. This makes the supposedly shared engine depend upward on the TCP record/socket module, and QUIC now reaches the engine through a record-specific contract. It also makes the updated documentation's statement that `tls13_backend.zig` imports no socket or record-layer module inaccurate.

Issue #442 explicitly permits the engine to depend on TLS events and a transport contract; that contract should live below both concrete carriers. Please extract the shared handshake contract/error vocabulary into a small neutral module, for example:

```zig
// src/tls/tls13_transport.zig
const events = @import("events.zig");
const transport = @import("transport.zig");

pub const Error = events.HandshakeError ||
    @import("messages.zig").ReadError ||
    @import("messages.zig").WriteError || error{
        TransportBufferOverflow,
        UnexpectedTransportEpoch,
        MissingTransportExtension,
    };

pub const Contract = transport.Contract(
    void,
    events.EncryptionEpoch,
    Error,
);
pub const Backend = Contract.Backend;
pub const EventSink = Contract.EventSink;
```

Then have `tls13_backend.zig` depend only on this neutral module, let `encrypted_stream.zig` alias/reuse it for record mode, and let the QUIC adapter translate from it. That preserves the current behavior while satisfying the requested module direction instead of merely moving the concrete type under `src/tls/`.

src/tls/tls13_backend_tests.zig
```diff
@@ -0,0 +1,516 @@
...
+
+    const opted_in = try SocketHarness.create(.{ .client_trust = .insecure_no_verification });
+    defer opted_in.destroy();
+    opted_in.client.allow_unverified_certificate = true;
+
+    try opted_in.driveUntil(SocketHarness.bothComplete);
+    try std.testing.expect(opted_in.client.bridge.handshake_complete);
+    try std.testing.expect(opted_in.server.bridge.handshake_complete);
+    try std.testing.expectEqual(events.CertificateState.not_checked, opted_in.client.certificateState());
+}
```
Jaruso, line 939: **[P1] Restore the required derivation and teardown regression coverage**

The replacement file keeps the socket-pair behavior tests, but deleting `src/quic/record_mode_handshake_test.zig` also deletes its stronger direct-driver assertions: client-write/server-read and server-write/client-read key equality for handshake/application epochs, sequence advancement, explicit initial/handshake discard observations, application-key cleanup after teardown, and verification that the driver sink's copied secret bytes were zeroed.

The new RFC 8448 `KeySchedule` unit vector is valuable, but it is not the issue's required cross-transport regression proof. The current suite also has no allocation-failure/constructor-owner test and no test that proves exactly-once teardown clears both the shared engine and the QUIC adapter's private `scratch`/transport-parameter buffers on success, failure, and cancellation.

Please migrate—not drop—the old direct-driver checks and add the required profile coverage. A practical structure would be:

```zig
const SecretSnapshot = struct {
    handshake_read: ?[hash_len]u8 = null,
    handshake_write: ?[hash_len]u8 = null,
    application_read: ?[hash_len]u8 = null,
    application_write: ?[hash_len]u8 = null,
    transcript: [hash_len]u8,
};

// Run deterministic record and QUIC-profile handshakes through their real
// owners. For each profile, compare each sender's write secret with the peer's
// read secret and compare transcript/secrets with committed pre-refactor golden
// values. The two profiles need separate goldens because ALPN/extensions differ.
```

Also retain an observable teardown helper like the removed `expectDriverSinkWiped`, and add a QUIC-adapter equivalent that poisons/uses `scratch` and `local_transport_parameters`, tears the owner down exactly once, and verifies those ranges plus the shared engine are zeroed. Exercise completed, policy-failed, and abandoned-mid-handshake cases. If the backend constructors are intentionally allocation-free, document that explicitly and run allocation-failure coverage around the allocating record/QUIC owner constructors instead.

Jaruso (commented):
## Review verdict: changes required

I completed a full review against issue #442, including the final module graph, both transport construction paths, the QUIC adapter, record-mode integration tests, lifecycle/zeroization behavior, and the current CI run.

The main architectural relocation is successful: the concrete handshake implementation is TLS-owned, `tls_core` does not import QUIC, the QUIC adapter no longer duplicates ClientHello/ServerHello/certificate/Finished/key-schedule logic, QUIC transport parameters and CID binding remain QUIC-owned, record mode directly injects the TLS-owned backend and negotiates `h2` without a QUIC extension, the temporary `RecordModeBackend`/dummy parameters/`omit_transport_parameters` path is gone, and CI—including the appliance dependency audit—is green.

Four acceptance-criteria gaps remain:

1. The new public transport-profile configuration is not validated, so invalid ALPN/extension inputs can panic at integer casts or construct self-incompatible handshakes.
2. Teardown leaves the deterministic X25519 seed—and the backend-owned server identity key copy—resident, so the zeroization completion report is stronger than the implementation.
3. The shared engine is still coupled to the high-level record-stream module rather than a transport-neutral TLS contract, contrary to the requested layering and the updated documentation.
4. The replacement test suite removes stronger secret/teardown assertions and does not implement the issue's required cross-transport vectors or constructor/teardown failure coverage.

GitHub does not permit this account to formally request changes on its own pull request, so this is submitted as a comment review. The merge recommendation is still **do not merge until the inline findings are addressed**.

src/tls/tls13_backend.zig
```diff
@@ -0,0 +1,1298 @@
...
+        try w.bytes(&KeySchedule.verifyData(schedule.server_handshake_traffic, self.core.transcriptHash()));
+        w.patch(3, finished_len);
+        const finished = buf[finished_start..w.len];
+        self.core.recordSent(finished) catch |err| return mapCoreError(err);
+        try sink.emitCrypto(.handshake, buf[0..w.len]);
+
+        // 1-RTT secrets from the transcript through server Finished; the
+        // client Finished we will require is fixed by the same hash.
+        const finished_hash = self.core.transcriptHash();
+        const app = schedule.applicationSecrets(finished_hash);
```
Jaruso: **[P1] Wipe temporary traffic-secret and KDF copies, not only owned fields**

The follow-up correctly wipes the backend's `KeySchedule`, adapter/driver sinks, deterministic key-share seed, generated key pair, and identity. However, both this server path and `onServerFinished()` still create a separate `ApplicationSecrets` value on the stack with `const app = ...`. Those two 32-byte traffic-secret copies survive every success and error return from the function because they are never passed to `secureZero`.

The same issue exists one layer down in `key_schedule.zig`: `KeySchedule.init()` leaves local copies of `handshake_secret`, `derived_handshake`, and `master_secret`, and `verifyData()` leaves its derived Finished key. Wiping the returned schedule later does not wipe those earlier stack copies. This means the explicit acceptance criterion that traffic secrets be zeroized on success, failure, and cancellation is still not fully satisfied.

Please make sensitive temporaries mutable and attach cleanup immediately, including error exits. For example:

```zig
pub const ApplicationSecrets = struct {
    client: [hash_len]u8,
    server: [hash_len]u8,

    pub fn wipe(self: *ApplicationSecrets) void {
        crypto.secureZero(u8, std.mem.asBytes(self));
    }
};

var app = schedule.applicationSecrets(finished_hash);
defer app.wipe();
try self.emitSecret(sink, .application, .read, &app.client);
try self.emitSecret(sink, .application, .write, &app.server);
```

Apply the same pattern at the client call site and to KDF intermediates/Finished keys inside `KeySchedule.init()` and `verifyData()`:

```zig
var derived_handshake = ...;
defer crypto.secureZero(u8, &derived_handshake);

var finished_key = finishedKey(traffic_secret);
defer crypto.secureZero(u8, &finished_key);
```

A small instrumented test can use a helper returning observable temporary storage, but the important property is that every production scope which materializes an additional secret owns a `defer` wipe before any subsequent fallible operation.

Jaruso (commented):
## Second review verdict: one remaining change required

I re-reviewed the complete PR at `65a2128`, not just the follow-up commits, and rechecked every acceptance criterion from #442.

The four findings from the first review are now materially addressed:

- profile validation runs before lifecycle/transcript mutation and covers invalid ALPN lengths, oversized opaque extensions, and TLS-extension collisions;
- the deterministic X25519 seed, generated key pair, backend-owned identity, transcript/reassembly state, peer buffers, adapter parameter storage, and event sinks are explicitly wiped;
- the concrete engine now depends on `tls13_transport.zig`, not `encrypted_stream.zig` or QUIC;
- the direct derivation/sequence/discard tests, record and extension secret goldens, allocation-failure coverage, and success/failure/abandonment owner teardown tests have been restored or added.

CI for the latest head is fully green across Linux, macOS, ARM, formatting, security, packaging, and the TLS dependency/appliance audit.

One security acceptance gap remains: application traffic secrets and several key-schedule intermediates are copied into local stack values and never explicitly wiped. The persistent backend/sink copies are now handled correctly, but that does not clear these additional copies on success or error return. See the inline finding.

Also update the completion report's validation counts before merge. The PR body still reports the original `152/303/1,381` counts even though the follow-up commits added a neutral module and numerous new tests. The issue explicitly requires the closing PR to contain the final validation matrix.

GitHub does not allow this account to request changes on its own PR, so this is a comment review; the merge recommendation remains **do not merge until the remaining secret-lifetime finding is addressed**.

Jaruso (commented):
## Re-review verdict: ready to merge

Reviewed head `509a362a76c05cc547212ccdbe6f19bdaaa6fec5` against issue #442 and every finding from the prior two review passes.

The remaining secret-lifecycle issue is addressed correctly:

- `KeySchedule.init()` now accepts the shared secret by pointer and attaches immediate `defer secureZero` cleanup to the handshake secret, derived-handshake secret, and local master-secret copies.
- `verifyData()` wipes its derived Finished key.
- `ApplicationSecrets` has explicit whole-struct cleanup, and both client/server handshake paths defer that cleanup before any fallible secret emission.
- Temporary expected/server/client Finished values are mutable and wiped after use, while the persistent `expected_client_verify` copy remains covered by the existing `finish()`/`deinit()` cleanup.
- Vector and OpenSSL differential call sites were updated to use the new pointer API and to wipe their derived secret material.

The PR completion report now reflects the expanded zeroization scope and current test totals. The latest CI run is fully green, including Linux/macOS/ARM tests, Debug and ReleaseSafe builds, formatting, dependency/security checks, packaging/performance smoke tests, and the appliance no-OpenSSL dependency audit.

I also rechecked the earlier findings: profile validation, private-key/identity zeroization, transport-neutral contract layering, cross-profile derivation vectors, allocation-failure coverage, and record/QUIC owner teardown coverage remain intact.

No blocking findings remain. GitHub does not allow this account to approve its own pull request, so this is submitted as a comment review, but my merge recommendation is **approve / ready to merge**.

PR #455, opened 2026-07-17
tls: provider-neutral credential authentication and verification contracts (#334)

Jaruso:
## Summary

Adds transport-neutral certificate-authentication hooks (#334) so the TLS 1.3
engine can select local credentials, sign through opaque key handles, and
delegate peer verification — with no OpenSSL, concrete X.509 validator, QUIC,
HTTP, or private-key byte types crossing the interface.

New module `src/tls/credentials.zig`:
- **Immutable certificate chains** (`CertificateChain`) as borrowed DER views, with normative ownership/lifetime rules documented in the module header (who owns the outer collection and each slice, view validity for the callback duration only, no retention, release/failure/cancellation behavior).
- **Credential selection** (`CredentialProvider` + `SelectionContext`) carrying role, preserved SNI, peer-offered signature algorithms, negotiated version/cipher/ALPN, and local auth policy — enough for future #347 production SNI selection without changing the engine.
- **Opaque credentials & signing** (`SelectedCredential`): exposes the public chain and a bounded signing callback that reports the written length and cannot exceed the caller's buffer. Private-key bytes never enter the engine.
- **Peer verification** (`PeerVerifier` + `VerificationContext`) over immutable peer DER views, supporting server auth and handshake-time client auth at the interface level; the seam #324 Web-PKI plugs into.
- **Deterministic failures**: a complete `FailureClass` taxonomy mapping each class to one alert, one origin (peer vs local), and one engine error — distinguishing peer-originated auth failure (`bad_certificate`) from local provider/config failure (`handshake_failure`/`internal_error`).
- **Fixed identity migration**: the previously hard-coded server identity and pin/insecure trust are now `FixedCredentialProvider`/`FixedVerifier`, served through the *same* production contract as any external provider — a single authentication path. Reusable mocks included.

Engine (`tls13_backend.zig`) changes:
- Server flight selects a credential, writes its chain, and signs CertificateVerify through the handle into a bounded scratch buffer; handle released exactly once (including on failure/cancellation).
- ClientHello parsing captures peer signature schemes and SNI into the selection context.
- Peer chain reassembled as bounded immutable views; trust verdict routed through the verifier, while proof-of-possession stays in the engine (transcript crypto, not PKI policy).
- Typed failures preserved across teardown; adds `NoApplicableCredential`/`CredentialProviderFailed` errors and the `handshake_failure` alert.

Post-handshake client authentication remains explicitly deferred.

## Test plan
- [x] `zig build test-tls` — contract unit tests + record-mode integration (successful handshake through mock provider **and** verifier; exact/absent SNI reaching selection; sigalg filtering; no compatible scheme; no credential; empty local chain; signing failure; oversized-signature contract violation; bad CertificateVerify signature; verifier rejection vs verifier internal failure; deterministic alert/origin per failure class; fixed-provider replacement; exactly-once handle/verifier teardown; allocation-failure cleanup). Tests assert callback invocation counts and lifetime transitions, not just final success/failure.
- [x] `zig build test-quic` — QUIC handshake regressions green.
- [x] `zig build test` — full general-profile suite green.
- [x] `zig build test -Dtls-profile=appliance` — compiles and passes with no OpenSSL analysis.

Closes #334

### DIFF
--- a/src/tls/credentials.zig
+++ b/src/tls/credentials.zig
@@ -26,27 +26,40 @@
 //! Certificate chains are surfaced as `CertificateChain`: an immutable view
 //! over a slice of DER byte slices.
 //!
-//!   - The outer `entries` collection is owned by whoever built the chain: the
-//!     engine's bounded reassembly buffer for *peer* chains, the provider's own
-//!     storage for *local* chains. `CertificateChain` never owns or frees it.
-//!   - Each inner DER byte slice is likewise borrowed, never copied by the view.
-//!   - A view handed to a callback (peer DER into `PeerVerifier.verify`, local
-//!     DER out of `SelectedCredential.chain`) is valid ONLY for the duration of
-//!     that single call. A callback MUST NOT retain any slice past its return;
-//!     it copies out anything it needs. The engine may reuse or wipe the
-//!     backing storage immediately afterward.
-//!   - A `SelectedCredential` handle is released exactly once, via `release`,
-//!     after the local flight has been signed and emitted, or immediately on
-//!     any failure encountered after selection (cancellation included). After
-//!     `release` the handle and every slice it produced are invalid.
+//! over a slice of DER byte slices. The `entries` collection and every inner
+//! DER slice are borrowed; `CertificateChain` never owns or frees them. Three
+//! distinct lifetime rules apply depending on which chain it is:
 //!
-//! ## Non-blocking compatibility
+//!   - **Selection-context inputs** (`SelectionContext.server_name`,
+//!     `.peer_signature_schemes`) are valid ONLY during the `select` call. A
+//!     provider that suspends (returns `pending`) or wants to keep any of them
+//!     MUST copy into its own storage first.
+//!   - **Peer-chain views** (`VerificationContext.chain`) are valid ONLY during
+//!     the `verify` call. A verifier MUST NOT retain a slice past its return;
+//!     the engine may reuse or wipe the reassembly buffer immediately after,
+//!     and any pending verify operation owns its own snapshot.
+//!   - **A selected local credential's chain** (`SelectedCredential.chain`) is
+//!     immutable and valid until `SelectedCredential.release()`. The engine
+//!     calls `chain()`, then serializes those slices into the flight, so the
+//!     provider MUST keep them stable for the whole selected lifetime — a
+//!     reloadable provider snapshots them at selection, not per-call.
 //!
-//! Every callback takes its inputs by argument and writes outputs into
-//! caller-owned buffers; none require global state or blocking filesystem or
-//! network access. A synchronous fixed provider and a future event-driven
-//! external signer satisfy the same signatures — the contract does not force
-//! private-key work to block.
+//! A `SelectedCredential` handle is released exactly once, via `release`, after
+//! the local flight has been signed and emitted, or immediately on any failure
+//! after selection (cancellation included). After `release` the handle and
+//! every slice it produced are invalid.
+//!
+//! ## Non-blocking / event-driven compatibility
+//!
+//! `select`, `sign`, and `verify` each return a progress value: `complete`
+//! now, or `pending` with a `PendingOperation` the engine drives via
+//! `poll`/`cancel`/`release`. An HSM, remote signer, or asynchronous verifier
+//! returns `pending`, wakes the driver later, and the engine resumes without
+//! recording any handshake message twice; a pending operation snapshots any
+//! input it needs (the transcript-derived signing bytes, the peer chain) so the
+//! caller-owned stack buffers may disappear. Synchronous providers return
+//! `complete` and never allocate a `PendingOperation`. Every operation is
+//! cancelled and released exactly once, including on handshake teardown.
 
 const std = @import("std");
 const crypto = std.crypto;
@@ -277,7 +290,14 @@ pub const FailureClass = enum {
     malformed_credential_chain,
     signing_provider_failure,
     signature_output_overflow,
+    /// The local credential provider itself failed (distinct from a *verifier*
+    /// failure, so diagnostics name the right subsystem).
+    provider_internal_failure,
     invalid_peer_certificate_chain,
+    /// The peer's CertificateVerify signature did not check out against its
+    /// presented leaf (proof-of-possession failure). Peer-originated, but a
+    /// distinct reason from a structurally invalid chain.
+    certificate_verify_invalid,
     peer_verification_rejected,
     verifier_internal_failure,
     invalid_callback_behavior,
@@ -285,6 +305,7 @@ pub const FailureClass = enum {
     pub fn origin(self: FailureClass) Origin {
         return switch (self) {
             .invalid_peer_certificate_chain,
+            .certificate_verify_invalid,
             .peer_verification_rejected,
             => .peer,
             .no_credential_available,
@@ -292,6 +313,7 @@ pub const FailureClass = enum {
             .malformed_credential_chain,
             .signing_provider_failure,
             .signature_output_overflow,
+            .provider_internal_failure,
             .verifier_internal_failure,
             .invalid_callback_behavior,
             => .local,
@@ -312,11 +334,13 @@ pub const FailureClass = enum {
             .malformed_credential_chain,
             .signing_provider_failure,
             .signature_output_overflow,
+            .provider_internal_failure,
             .verifier_internal_failure,
             .invalid_callback_behavior,
             => .internal_error,
             // Peer-originated authentication failure.
             .invalid_peer_certificate_chain,
+            .certificate_verify_invalid,
             .peer_verification_rejected,
             => .bad_certificate,
         };
@@ -328,6 +352,7 @@ pub const FailureClass = enum {
     pub fn engineError(self: FailureClass) events.HandshakeError {
         return switch (self) {
             .invalid_peer_certificate_chain,
+            .certificate_verify_invalid,
             .peer_verification_rejected,
             => error.CertificateInvalid,
             .no_credential_available,
@@ -336,6 +361,7 @@ pub const FailureClass = enum {
             .malformed_credential_chain,
             .signing_provider_failure,
             .signature_output_overflow,
+            .provider_internal_failure,
             .verifier_internal_failure,
             .invalid_callback_behavior,
             => error.CredentialProviderFailed,
@@ -349,7 +375,7 @@ pub fn classifySelectError(err: SelectError) FailureClass {
         error.NoCredentialAvailable => .no_credential_available,
         error.NoCompatibleSignatureAlgorithm => .no_compatible_signature_algorithm,
         error.MalformedCredentialChain => .malformed_credential_chain,
-        error.ProviderInternalFailure => .verifier_internal_failure,
+        error.ProviderInternalFailure => .provider_internal_failure,
         error.InvalidCallbackBehavior => .invalid_callback_behavior,
     };
 }
@@ -704,15 +730,29 @@ pub const MockCredentialProvider = struct {
     /// Return an empty certificate chain from selection, to model a malformed
     /// local credential the engine must reject before signing.
     empty_chain: bool = false,
+    /// Return the leaf repeated this many times, to model a provider whose
+    /// chain exceeds the engine's entry/size bounds.
+    chain_repeat: usize = 1,
+    /// Skip the peer-offer compatibility check and return the credential's
+    /// scheme regardless, to model a provider that hands back an algorithm the
+    /// peer never advertised.
+    ignore_offer: bool = false,
     /// Sign normally, then flip a signature byte, to model a peer whose
     /// CertificateVerify does not check out (proof-of-possession failure).
     flip_signature: bool = false,
+    chain_storage: [16][]const u8 = undefined,
 
     select_count: usize = 0,
     sign_count: usize = 0,
     release_count: usize = 0,
-    last_server_name: ?[]const u8 = null,
+    /// The last SNI selection saw, copied into mock-owned storage — the
+    /// `SelectionContext.server_name` slice is only valid during `select`, so a
+    /// mock that wants to remember it MUST NOT retain the borrowed slice.
+    last_server_name_buf: [256]u8 = undefined,
+    last_server_name_len: usize = 0,
+    last_server_name_present: bool = false,
     last_offered_scheme_count: usize = 0,
+    last_role: ?Role = null,
 
     pub fn init(identity: Identity) MockCredentialProvider {
         return .{ .identity = identity, .chain_entry = .{identity.certificate_der} };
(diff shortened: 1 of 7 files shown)


src/tls/credentials.zig
```diff
@@ -0,0 +1,1044 @@
...
+//! Certificate chains are surfaced as `CertificateChain`: an immutable view
+//! over a slice of DER byte slices.
+//!
+//!   - The outer `entries` collection is owned by whoever built the chain: the
+//!     engine's bounded reassembly buffer for *peer* chains, the provider's own
+//!     storage for *local* chains. `CertificateChain` never owns or frees it.
+//!   - Each inner DER byte slice is likewise borrowed, never copied by the view.
+//!   - A view handed to a callback (peer DER into `PeerVerifier.verify`, local
+//!     DER out of `SelectedCredential.chain`) is valid ONLY for the duration of
+//!     that single call. A callback MUST NOT retain any slice past its return;
```
Jaruso: The normative lifetime here conflicts with both the vtable comment and the engine implementation. `certificateChain()` returns, and only then does `sendServerFlight` iterate the returned local-chain slices. Under this text those slices may already be invalid. Please split the rules: peer-chain views and selection inputs are callback-scoped, but a selected local credential's chain must remain immutable and valid until `SelectedCredential.release()` (or redesign this as a visitor/copy callback). Also update the mock provider: it currently retains `selection.server_name` past `select`, violating the callback-scoped `SelectionContext` lifetime.

src/tls/credentials.zig
```diff
@@ -0,0 +1,1044 @@
...
+//!     after the local flight has been signed and emitted, or immediately on
+//!     any failure encountered after selection (cancellation included). After
+//!     `release` the handle and every slice it produced are invalid.
+//!
+//! ## Non-blocking compatibility
+//!
+//! Every callback takes its inputs by argument and writes outputs into
+//! caller-owned buffers; none require global state or blocking filesystem or
+//! network access. A synchronous fixed provider and a future event-driven
+//! external signer satisfy the same signatures — the contract does not force
```
Jaruso: These signatures do force private-key work to complete synchronously. An event-driven signer cannot return `pending`, supply an operation token, wake the TLS driver, or resume writing into a stable buffer; the stack `SelectionContext`, signing input, and verification views disappear when the call returns. Please model begin/pending/resume/cancel explicitly and add pending-completion plus cancellation tests. Merely using caller-owned buffers does not satisfy #334's nonblocking/event-driven requirement.

src/tls/credentials.zig
```diff
@@ -0,0 +1,1044 @@
...
+    }
+};
+
+/// Map a `SelectError` to its failure class.
+pub fn classifySelectError(err: SelectError) FailureClass {
+    return switch (err) {
+        error.NoCredentialAvailable => .no_credential_available,
+        error.NoCompatibleSignatureAlgorithm => .no_compatible_signature_algorithm,
+        error.MalformedCredentialChain => .malformed_credential_chain,
+        error.ProviderInternalFailure => .verifier_internal_failure,
```
Jaruso: This reports a credential-provider selection failure as `verifier_internal_failure`, so the supposedly complete typed taxonomy identifies the wrong subsystem. Please add a distinct local `provider_internal_failure` class (mapping to `internal_error` / `CredentialProviderFailed`) and test it independently from verifier failure.

src/tls/tls13_backend.zig
```diff
@@ -732,19 +691,72 @@ pub const Tls13Backend = struct {
...
         while (list.remaining() > 0) {
-            _ = try list.slice(try list.u24_());
-            _ = try list.slice(try list.u16_());
+            const entry = try list.slice(try list.u24_());
+            _ = try list.slice(try list.u16_()); // per-certificate extensions
+            // Additional chain certificates beyond our bound are framed-checked
+            // above but not retained; the leaf is what pinning verifies and a
+            // #324 verifier receives what fits.
+            self.appendPeerCertificate(entry) catch |err| switch (err) {
+                error.CertificateInvalid => {},
```
Jaruso: Do not ignore this error and verify a truncated chain. If the ninth entry or aggregate DER exceeds the bound, the verifier receives only a prefix of the peer's chain with no indication that data was dropped. That is unsafe for #324 path building. Reject the entire peer credential and latch `.invalid_peer_certificate_chain` before invoking the verifier. Also reject `entry_len == 0` for every non-leaf CertificateEntry.

For example:
```zig
const entry_len = try list.u24_();
if (entry_len == 0 or entry_len > max_certificate_len)
    return self.failCredential(.invalid_peer_certificate_chain);
const entry = try list.slice(entry_len);
_ = try list.slice(try list.u16_());
self.appendPeerCertificate(entry) catch
    return self.failCredential(.invalid_peer_certificate_chain);
```

src/tls/tls13_backend.zig
```diff
@@ -888,7 +931,27 @@ pub const Tls13Backend = struct {
                 ext_signature_algorithms => {
                     var algorithms = Reader{ .bytes = try ext.slice(try ext.u16_()) };
                     while (algorithms.remaining() > 0) {
-                        if (try algorithms.u16_() == required_sigalg) offers_our_sigalg = true;
+                        const scheme = try algorithms.u16_();
+                        // Capture the peer's offers (in order) for credential
+                        // selection; ignore any past our bounded capacity.
```
Jaruso: Silently truncating the peer's signature-algorithm offer changes credential-selection semantics. A compatible scheme in slot 17 becomes a false local `NoApplicableCredential`, and a future provider may support schemes the engine cannot pre-filter. Preserve the complete bounded vector or fail explicitly when the bound is exceeded; never ignore the suffix. Separately track whether this required extension was present and non-empty so missing/malformed input is attributed to the peer rather than to local credential configuration.

Jaruso (commented):
## Blocking review findings

I reviewed the complete seven-file diff against #334's objective, scope, acceptance criteria, and the additional contract requirements in the issue comment. The CI workflow for this exact head SHA is green, and several parts are well done:

- A real record-mode handshake succeeds through a mock credential provider and mock verifier.
- The fixed identity/trust implementations use the same provider/verifier seam as external implementations.
- Private-key bytes and concrete OpenSSL/X.509 objects do not cross the interface.
- The bounded signing buffer, exactly-once release on the current synchronous paths, deterministic alert mapping, and the existing negative tests are useful foundations.

However, important parts of the requested contract are either absent or implemented in a way that loses security-relevant information. I would not merge this as closing #334 until the following are addressed.

### 1. Handshake-time client authentication is not supported

#334 explicitly requires hooks for both server authentication and handshake-time client authentication, while only post-handshake client authentication may be deferred. This PR adds types whose `role` fields could describe client authentication, but the engine cannot exercise them:

- there is no client-side credential provider configuration;
- there is no server-side peer verifier/client-auth policy configuration;
- `certificate_request` is not accepted or emitted by the state machine;
- the client always expects `Certificate` immediately after `EncryptedExtensions`;
- the server always expects client `Finished` after its own flight;
- `onCertificateVerify` always verifies a server signature and always constructs `.role = .client`.

Please add an actual handshake-time client-auth path rather than only role fields in the interface. One reasonable configuration shape would be:

```zig
pub const ClientAuthMode = enum { disabled, optional, required };

pub const AuthenticationConfig = struct {
    local_credentials: ?credentials.CredentialProvider = null,
    peer_verifier: ?credentials.PeerVerifier = null,
    client_auth: ClientAuthMode = .disabled,
    policy: credentials.AuthPolicy = .{},
};
```

The state machine then needs to support:

```text
server: EncryptedExtensions -> [CertificateRequest] -> Certificate
        -> CertificateVerify -> Finished
client: [Certificate -> CertificateVerify] -> Finished
```

When requested, client credential selection must use `SelectionContext.role = .client`; server verification must use `VerificationContext.role = .server`. Optional auth must allow an empty client `Certificate`; required auth must fail deterministically, using `certificate_required` when appropriate. Add integration tests for optional and required client auth, no client credential, incompatible client signature schemes, client verifier rejection/internal failure, and exactly-once credential release.

### 2. The callback contract is synchronous and cannot satisfy the required nonblocking/event-driven progression

The issue comment specifically requires callback progression compatible with nonblocking/event-driven operation. Returning directly from `select`, `sign`, and `verify` with no pending state, operation token, completion callback, or resume entry point means an HSM, remote signer, external agent, or asynchronous verifier must block the event-loop worker. Caller-owned buffers do not make a callback asynchronous.

The transport-neutral contract needs an explicit pending/resume/cancel mechanism. The exact API is flexible, but it must model something equivalent to:

```zig
pub const SignProgress = union(enum) {
    complete: usize,
    pending: PendingOperation,
};

pub const VerifyProgress = union(enum) {
    complete: Verdict,
    pending: PendingOperation,
};

pub const PendingOperation = struct {
    handle: *anyopaque,
    vtable: *const VTable,

    pub const VTable = struct {
        poll: *const fn (*anyopaque, *Completion) OperationError!bool,
        cancel: *const fn (*anyopaque) void,
        release: *const fn (*anyopaque) void,
    };
};
```

The engine must retain the selected credential and exact transcript-derived signing input while pending, resume without recording any handshake message twice, and cancel/release each operation exactly once during teardown. Please add pending-then-complete and pending-then-cancel tests for selection, signing, and verification.

### 3. Peer certificate chains must be complete or rejected, never silently truncated

The current parser drops certificates after the entry-count or byte bound is reached and then calls the verifier with the prefix that happened to fit. A Web-PKI/path-building verifier must never make a trust decision over a chain different from the one the peer sent. Zero-length non-leaf entries are also currently accepted.

Please fail closed before invoking the verifier when any entry or the full chain exceeds a bound. The failure should latch the typed peer-chain failure and map to the documented peer-originated alert. Tests should cover more than `max_peer_chain_entries`, total DER over `max_peer_chain_bytes`, empty/oversized intermediate certificates, and verify that the callback is not invoked.

### 4. ClientHello authentication metadata is silently changed by local bounds

Credential selection is only correct if the provider receives the peer's actual offer:

- signature algorithms after the first 16 are silently discarded;
- a compatible scheme in position 17 therefore becomes a false `NoCompatibleSignatureAlgorithm` local failure;
- missing or empty `signature_algorithms` is also treated as a local no-compatible-credential condition instead of a malformed/missing required peer extension;
- oversized, empty, and duplicate `host_name` values can be silently treated as absent or accepted, collapsing malformed SNI into the default-certificate path.

Use a sufficiently large bounded representation or fail explicitly when the representation cannot preserve the full value. Do not truncate. Track required-extension presence separately from compatibility. Preserve the distinction among absent SNI, valid SNI, and malformed/unsupported SNI.

### 5. Validate all provider output before using it

After selection, the engine must verify that `credential.scheme` was actually offered by the peer. Otherwise a buggy provider can make the engine emit a CertificateVerify algorithm the client did not advertise. The local chain also needs explicit entry-count and aggregate-size validation so a provider contract violation maps to `CredentialProviderFailed`, not a generic writer-buffer failure.

At minimum, enforce invariants equivalent to:

```zig
defer credential.release();

const chain = credential.certificateChain();
if (chain.count() == 0 or chain.count() > credentials.max_chain_entries)
    return self.failCredential(.malformed_credential_chain);

var encoded_len: usize = 0;
for (chain.entries) |entry| {
    if (entry.len == 0 or entry.len > max_certificate_len)
        return self.failCredential(.malformed_credential_chain);
    encoded_len = std.math.add(usize, encoded_len, entry.len + 5) catch
        return self.failCredential(.malformed_credential_chain);
    if (encoded_len > max_message_len)
        return self.failCredential(.malformed_credential_chain);
}
```

Add malicious-provider tests for an unoffered returned scheme, too many chain entries, and an aggregate chain that exceeds the flight buffer. Each must avoid signing when validation fails and release the selected handle exactly once.

### 6. The verifier does not receive the identity or policy needed for Web-PKI

The client verification context always receives `server_name = null`, and the client does not have an intended-server-name configuration that is encoded into ClientHello SNI and later passed to the verifier. That makes the #324 Web-PKI adapter unable to perform hostname verification through this seam.

Also, `authPolicy()` derives policy from the legacy `trust` field. `initClientWithVerifier` leaves that field at its insecure default, so an external verifier receives `allow_unverified_peer = true` regardless of the caller's actual policy. Store explicit authentication policy in the backend and pass the exact configured server name and policy to both the wire path and verifier.

Add a handshake test proving the verifier observes the exact intended hostname, plus tests for explicitly absent hostname and strict versus allow-unverified policy.

### 7. Clarify and enforce lifetimes

The module-level normative text says the local chain returned from `SelectedCredential.chain` is valid only during that callback and may be wiped immediately after it returns, while the vtable documentation says it is valid until `release`. The engine calls `certificateChain()`, returns from the callback, and then serializes the returned slices, so it violates the stricter stated rule.

Use separate lifetime rules:

- selection-context slices: valid only during `select` unless an async operation explicitly owns a snapshot;
- peer-chain views: valid only during the verifier invocation unless an async operation owns a snapshot;
- a selected local credential's chain: immutable and valid until `SelectedCredential.release()`.

The mock provider currently retains `selection.server_name` after the selection callback, which also violates the selection-context lifetime. Copy it into mock-owned storage or assert it inside the callback. Add a reload/snapshot test proving an in-flight selected credential remains valid while the provider's active set changes.

### 8. Complete the failure taxonomy

`ProviderInternalFailure` is classified as `verifier_internal_failure`, which makes diagnostics claim the wrong component failed. Add a distinct `provider_internal_failure` class. A bad CertificateVerify signature is also currently recorded as `invalid_peer_certificate_chain`; a separate `peer_signature_invalid`/`certificate_verify_invalid` class would preserve the actual reason while retaining the same peer origin and `bad_certificate` alert.

Once these changes are made, please extend the AC tests rather than only unit-testing the new types. The critical proof is that the transport-neutral engine can suspend/resume authentication, perform optional/required handshake-time client auth, reject incomplete peer-chain views, and deliver exact identity/policy metadata without changing handshake logic for each concrete provider.

src/tls/tls13_backend.zig
```diff
@@ -888,7 +931,27 @@ pub const Tls13Backend = struct {
...
+                    }
+                },
+                ext_server_name => {
+                    // RFC 6066 §3: ServerNameList<1..>, each { name_type, name<..> }.
+                    // Preserve the first host_name (type 0) exactly as received.
+                    var names = Reader{ .bytes = try ext.slice(try ext.u16_()) };
+                    while (names.remaining() > 0) {
+                        const name_type = try names.u8_();
+                        const name = try names.slice(try names.u16_());
+                        if (name_type == 0 and !self.server_name_present and name.len <= self.server_name.len) {
```
Jaruso: This collapses several distinct inputs into the default/absent-SNI path: an oversized host_name is silently ignored, an empty host_name is accepted, and duplicate host_name entries are ignored after the first even though RFC 6066 forbids duplicate name types. Please reject malformed/duplicate/over-bound host names deterministically rather than treating them as absent. A selector must be able to distinguish `absent` from `invalid`; otherwise an invalid SNI can receive the default certificate.

src/tls/tls13_backend.zig
```diff
@@ -754,48 +766,75 @@ pub const Tls13Backend = struct {
...
+        // DER views it must not retain.
+        var views: [max_peer_chain_entries][]const u8 = undefined;
+        var verifier_storage: credentials.FixedVerifier = undefined;
+        const verifier = if (self.external_verifier) |v| v else blk: {
+            verifier_storage = credentials.FixedVerifier.init(self.trust);
+            break :blk verifier_storage.verifier();
+        };
+        const context = credentials.VerificationContext{
+            .role = .client,
+            .server_name = null,
```
Jaruso: A production Web-PKI verifier cannot perform hostname verification because the intended server name is always discarded here. Add client configuration/storage for the exact server name, emit it as ClientHello SNI, and pass the same value into `VerificationContext`. Also stop deriving external-verifier policy from the legacy `trust` default: `initClientWithVerifier` currently causes `authPolicy()` to report `allow_unverified_peer = true` even when the caller intended strict verification.

src/tls/tls13_backend.zig
```diff
@@ -983,9 +1046,34 @@ pub const Tls13Backend = struct {
...
+        // Wipe the fixed key material (stack copy and stored identity) on the
+        // way out, whatever the outcome.
+        defer if (using_fixed) {
+            fixed_provider.deinit();
+            self.wipeIdentity();
+        };
+
+        var selection = self.serverSelectionContext();
+        var credential: credentials.SelectedCredential = undefined;
+        provider.selectCredential(&selection, &credential) catch |err|
```
Jaruso: Success from an external provider is trusted without checking its output. Before signing, verify `selection.offersScheme(credential.scheme)`; otherwise a buggy provider can make the server emit a CertificateVerify scheme the client never offered. Treat mismatch as `.invalid_callback_behavior`, do not call `sign`, and still release the handle exactly once. A malicious-provider integration test should cover this.

src/tls/tls13_backend.zig
```diff
@@ -1012,47 +1100,45 @@ pub const Tls13Backend = struct {
         w.patch(3, ee_len);
         const encrypted_extensions = buf[0..w.len];
 
-        // Certificate.
+        // Certificate: the selected credential's public DER chain. The views
+        // are borrowed for this call only and never retained past it.
+        const chain = credential.certificateChain();
+        if (chain.count() == 0) return self.failCredential(.malformed_credential_chain);
```
Jaruso: The declared contract caps chains at `credentials.max_chain_entries`, but this path enforces only non-empty. It also lets aggregate encoding overflow the 8 KiB writer and surface a generic `HandshakeBufferOverflow` instead of the typed local-provider failure. Validate entry count, every entry, and total encoded size before recording any flight messages; map violations to `.malformed_credential_chain`, skip signing, and release exactly once.

PR #456, opened 2026-07-17
test(pki): bound differential validator processes

Jaruso:
## Summary

- add a reusable bounded child-process runner for PKI differential validators
- route OpenSSL and Go oracle invocations through deadline/output-limit/structured-outcome handling
- add a deterministic compiled helper and offline tests for success, rejection-style exits, launch failure, timeout/reaping, signal termination, stdout/stderr overflow, malformed output classification, temp-dir cleanup, deterministic classification, and allocation-failure cleanup
- document stable and extended per-validator deadlines

## Validation

- git diff --check
- zig build test-pki-reduce --summary all --error-style verbose
- zig build test-pki --summary all --error-style verbose
- zig build test-pki-differential --summary all --error-style verbose --test-timeout 2m
- zig build test-pki-differential-extended --summary all --error-style verbose --test-timeout 10m
- zig build test-pki-openssl --summary all --error-style verbose
- zig build test-pki-policy-openssl --summary all --error-style verbose
- zig build test --summary all --error-style verbose --test-timeout 2m
- zig build -Doptimize=ReleaseFast

Refs #348

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -456,6 +456,22 @@ pub fn build(b: *std.Build) void {
     // OpenSSL and Go crypto/x509 are invoked as independent processes. These
     // targets stay opt-in because the external validators are test tools, not
     // production dependencies.
+    const pki_process_helper_mod = b.createModule(.{
+        .root_source_file = b.path("tests/pki_process_helper.zig"),
+        .target = target,
+        .optimize = optimize,
+        .link_libc = true,
+    });
+    const pki_process_helper = b.addExecutable(.{
+        .name = "pki_process_helper",
+        .root_module = pki_process_helper_mod,
+    });
+    const pki_process_helper_install = b.addInstallArtifact(pki_process_helper, .{});
+    const pki_diff_options = b.addOptions();
+    pki_diff_options.addOption([]const u8, "process_helper_path", b.getInstallPath(.bin, "pki_process_helper"));
+    pki_diff_options.addOption(u32, "stable_validator_deadline_ms", 10_000);
+    pki_diff_options.addOption(u32, "extended_validator_deadline_ms", 30_000);
+
     const pki_differential_mod = b.createModule(.{
         .root_source_file = b.path("tests/pki_differential.zig"),
         .target = target,
@@ -464,6 +480,7 @@ pub fn build(b: *std.Build) void {
     pki_differential_mod.addImport("crypto", crypto_mod);
     pki_differential_mod.addImport("pki", pki_mod);
     pki_differential_mod.addImport("zig_compat", compat_mod);
+    pki_differential_mod.addImport("pki_diff_options", pki_diff_options.createModule());
     pki_differential_mod.addAnonymousImport("pki_root_crt", .{
         .root_source_file = b.path("tests/vectors/pki/root.crt"),
     });
@@ -501,6 +518,7 @@ pub fn build(b: *std.Build) void {
         .filters = &.{"pki reduce"},
     });
     const run_pki_reduce_tests = b.addRunArtifact(pki_reduce_tests);
+    run_pki_reduce_tests.step.dependOn(&pki_process_helper_install.step);
     const pki_reduce_step = b.step("test-pki-reduce", "Run offline PKI mismatch-minimization tests");
     pki_reduce_step.dependOn(&run_pki_reduce_tests.step);
     test_step.dependOn(&run_pki_reduce_tests.step);
--- a/build.zig
+++ b/build.zig
@@ -475,6 +475,7 @@ pub fn build(b: *std.Build) void {
     const pki_diff_options = b.addOptions();
     pki_diff_options.addOption([]const u8, "process_helper_path", b.getInstallPath(.bin, "pki_process_helper"));
     pki_diff_options.addOption([]const u8, "go_validator_path", b.getInstallPath(.bin, "pki_go_validator"));
+    pki_diff_options.addOption([]const u8, "go_bin", go_bin);
     pki_diff_options.addOption(u32, "stable_validator_deadline_ms", 10_000);
     pki_diff_options.addOption(u32, "extended_validator_deadline_ms", 30_000);
 
--- a/build.zig
+++ b/build.zig
@@ -17,6 +17,7 @@ pub fn build(b: *std.Build) void {
     const require_static_system_libs = b.option(bool, "require-static-system-libs", "Require static linking for system libraries") orelse false;
     const static_executable = b.option(bool, "static-executable", "Build the tardi executable as a static binary") orelse false;
     const app_version = b.option([]const u8, "version", "Version string embedded in the tardi binary") orelse "dev";
+    const go_bin = b.option([]const u8, "go-bin", "Go command used to build the PKI crypto/x509 oracle") orelse "go";
     const tls_profile = b.option(
         TlsProfile,
         "tls-profile",
@@ -467,8 +468,13 @@ pub fn build(b: *std.Build) void {
         .root_module = pki_process_helper_mod,
     });
     const pki_process_helper_install = b.addInstallArtifact(pki_process_helper, .{});
+    const pki_go_validator_build = b.addSystemCommand(&.{ go_bin, "build", "-trimpath", "-o" });
+    const pki_go_validator_output = pki_go_validator_build.addOutputFileArg("pki_go_validator");
+    pki_go_validator_build.addFileArg(b.path("tests/pki_go_validator.go"));
+    const pki_go_validator_install = b.addInstallBinFile(pki_go_validator_output, "pki_go_validator");
     const pki_diff_options = b.addOptions();
     pki_diff_options.addOption([]const u8, "process_helper_path", b.getInstallPath(.bin, "pki_process_helper"));
+    pki_diff_options.addOption([]const u8, "go_validator_path", b.getInstallPath(.bin, "pki_go_validator"));
     pki_diff_options.addOption(u32, "stable_validator_deadline_ms", 10_000);
     pki_diff_options.addOption(u32, "extended_validator_deadline_ms", 30_000);
 
@@ -499,6 +505,7 @@ pub fn build(b: *std.Build) void {
         .filters = &.{"pki differential core corpus"},
     });
     const run_pki_differential_core_tests = b.addRunArtifact(pki_differential_core_tests);
+    run_pki_differential_core_tests.step.dependOn(&pki_go_validator_install.step);
     const pki_differential_step = b.step("test-pki-differential", "Run stable PKI differential corpus against OpenSSL and Go");
     pki_differential_step.dependOn(&run_pki_differential_core_tests.step);
 
@@ -507,6 +514,7 @@ pub fn build(b: *std.Build) void {
         .filters = &.{"pki differential full corpus"},
     });
     const run_pki_differential_full_tests = b.addRunArtifact(pki_differential_full_tests);
+    run_pki_differential_full_tests.step.dependOn(&pki_go_validator_install.step);
     const pki_differential_extended_step = b.step("test-pki-differential-extended", "Run full PKI differential corpus against OpenSSL and Go");
     pki_differential_extended_step.dependOn(&run_pki_differential_full_tests.step);
 
--- a/tests/bounded_process.zig
+++ b/tests/bounded_process.zig
@@ -60,10 +60,13 @@ pub const Result = struct {
 
 pub fn run(allocator: std.mem.Allocator, options: Options) std.mem.Allocator.Error!Result {
     const io = compat.io();
-    const deadline_end_ms: ?i64 = if (options.deadline_ms == 0)
+    const deadline_end: ?std.Io.Clock.Timestamp = if (options.deadline_ms == 0)
         null
     else
-        compat.milliTimestamp() + @as(i64, @intCast(options.deadline_ms));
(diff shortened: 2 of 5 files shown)


Jaruso (commented):
## Blocking review findings

I audited this PR against both the original #348 acceptance criteria and the detailed final-hardening plan recorded on the issue.

This PR is correctly scoped as one remaining slice and correctly uses `Refs #348` rather than claiming closure. It does complete several important parts of the subprocess-control work: argv-based spawning, bounded captured diagnostics, distinct runner outcomes, stable/extended configured deadlines, direct-child success/rejection/signal/overflow coverage, and no changes to the production TLS files reserved for #334 / PR #455. The current CI run is green.

However, the two lifecycle gaps called out inline mean the core requirement—an enforceable wall-clock deadline with deterministic cleanup and no orphaned validator work—is not yet satisfied:

1. The deadline applies only while reading stdout/stderr. Once both pipes reach EOF, `child.wait()` can block forever without the deadline.
2. Failure cleanup kills only the directly spawned PID, while the current Go command is `go run ...`, which creates descendant processes and temporary build state that can survive direct-parent termination.

Please make the deadline cover the entire interval from successful spawn through final reap, and make timeout/output-exhaustion cleanup terminate the whole validator process tree (or remove the process tree by compiling the Go oracle once and invoking the resulting binary directly). The tests need to cover a child that closes both output pipes before hanging and a child that leaves a hanging descendant, verifying every recorded PID is gone and controlled temporary storage is empty afterward.

The remaining #348 work—semantic reason taxonomy, expected-reason manifest support, false-agreement fixture corrections, expanded hostile algorithm/DER corpus, artifact-schema additions, and the final closure audit—can remain in later PRs, but #348 must stay open until those are complete.

These are blocking findings; GitHub does not permit the PR author’s own account to submit a formal `REQUEST_CHANGES` review.

tests/bounded_process.zig
```diff
@@ -0,0 +1,277 @@
...
+            &reaped,
+            &multi_reader,
+            &reader_active,
+            .launch_failure,
+            options.stdout_limit,
+            options.stderr_limit,
+        ),
+    };
+
+    const term = child.wait(io) catch |err| {
```
Jaruso: **Blocking: the configured deadline no longer applies after both pipes close.**

`deadline` is only passed to `multi_reader.fill()`. If a validator writes some output, closes stdout and stderr, and then hangs, the reader reaches `EndOfStream` and this unconditional `child.wait(io)` can block forever. In that case the outer Zig test/GitHub Actions timeout once again becomes the primary hang protection, which is specifically what #348 requires this runner to prevent.

Please enforce one absolute deadline across the entire child lifecycle—pipe draining **and** process termination/reaping. Reaching EOF on the pipes must not imply that the process exited. On expiry, terminate the validator, reap it, and return `.timeout`.

A regression helper/test should exercise this exact path, for example:

```zig
if (std.mem.eql(u8, mode, "close-stdio-hang")) {
    const pid = std.c.getpid();
    var buf: [64]u8 = undefined;
    try writeFd(1, try std.fmt.bufPrint(&buf, "pid={d}\n", .{pid}));
    _ = std.c.close(1);
    _ = std.c.close(2);
    hangForever();
}
```

```zig
var result = try bounded_process.run(testing.allocator, .{
    .argv = &.{ helper_path, "close-stdio-hang" },
    .stdout_limit = 64,
    .stderr_limit = 64,
    .deadline_ms = 100,
});
defer result.deinit(testing.allocator);
try expectOutcomeTag(result, .timeout);
try expectNoProcess(testing.allocator, try parseHelperPid(result.stdout));
```

That test hangs on the current implementation at this line and will prove that the fix covers the full wall-clock invocation rather than only stream reads.

tests/bounded_process.zig
```diff
@@ -0,0 +1,277 @@
...
+    stdout_limit: usize,
+    stderr_limit: usize,
+) std.mem.Allocator.Error!Result {
+    const stdout = try boundedDupe(allocator, multi_reader.reader(0).buffered(), stdout_limit);
+    errdefer allocator.free(stdout);
+    const stderr = try boundedDupe(allocator, multi_reader.reader(1).buffered(), stderr_limit);
+    errdefer allocator.free(stderr);
+    reader_active.* = false;
+    multi_reader.deinit();
+    child.kill(compat.io());
```
Jaruso: **Blocking: this only terminates/reaps the directly spawned PID, not the validator process tree.**

`goDecision()` currently executes `go run tests/pki_go_validator.go ...`. The `go` process compiles and starts a separate validator executable. On timeout or output exhaustion, killing only the `go` PID can leave the compiled child running and can leave `go run` temporary/build state behind. Closing the pipes does not guarantee that descendant is gone; a descendant can redirect/close stdio and remain orphaned.

Please use one of these designs:

1. Preferably compile the Go oracle once as a differential-test build dependency and invoke that binary directly, removing `go run` from every case; and
2. Make the generic bounded runner start the validator in an isolated process group/session (or the platform-equivalent job object) and terminate the entire group on timeout/output exhaustion before reaping the direct child.

The regression helper should have a mode that spawns a grandchild which hangs, emits both parent and grandchild PIDs, and then hangs itself. After timeout, assert that **both** PIDs are absent. It should also create a marker in a controlled temporary directory so the test can verify cleanup rather than merely checking an initially empty directory.

Conceptually:

```zig
const result = try run(helper, &.{ "spawn-grandchild-and-hang", tmp.abs_path }, 100);
try expectOutcomeTag(result, .timeout);
try expectNoProcess(allocator, result.parent_pid);
try expectNoProcess(allocator, result.grandchild_pid);
try expectDirEmptyPath(tmp.rel_path);
```

Without process-tree coverage, the runner does not yet meet the issue requirement for deterministic termination with no orphaned validator work or temporary files.

tests/bounded_process.zig
```diff
@@ -0,0 +1,365 @@
...
+) std.mem.Allocator.Error!Result {
+    const stdout = try boundedDupe(allocator, multi_reader.reader(0).buffered(), stdout_limit);
+    errdefer allocator.free(stdout);
+    const stderr = try boundedDupe(allocator, multi_reader.reader(1).buffered(), stderr_limit);
+    errdefer allocator.free(stderr);
+    reader_active.* = false;
+    multi_reader.deinit();
+    terminateProcessGroup(child.id.?);
+    child.kill(compat.io());
+    reaped.* = true;
```
Jaruso: **Blocking: process-group cleanup still occurs after fallible allocations and is not installed on every unwind path.**

`killedResult()` duplicates stdout and stderr before reaching this line. If either `boundedDupe()` returns `error.OutOfMemory`, control unwinds to `run()`'s current defer, which only calls `child.kill(io)` for the direct child. A grandchild in the new process group can remain alive. The same direct-PID-only cleanup occurs when `multi_reader.checkAnyError()` returns OOM and when `waitForExit()` fails and `failureFromBuffered()` subsequently errors.

That means the new process-group guarantee does not hold for allocation failure—the exact failure mode the issue asks the harness to cover.

Please capture the process-group ID immediately after spawn and install group termination plus direct-child reaping as the unconditional fallback before any later allocation can fail. Also terminate/reap before performing diagnostic allocations in `killedResult()`.

Conceptually:

```zig
const pgid = child.id.?;
var reaped = false;
defer if (!reaped) {
    terminateProcessGroup(pgid);
    _ = child.kill(io); // reap the direct child
};
```

```zig
fn killedResult(...) Allocator.Error!Result {
    const pgid = child.id.?;

    // No fallible allocation may precede process-tree termination.
    terminateProcessGroup(pgid);
    child.kill(compat.io());
    reaped.* = true;

    const stdout = try boundedDupe(...);
    errdefer allocator.free(stdout);
    const stderr = try boundedDupe(...);
    ...
}
```

Please extend the allocation-failure test to use `spawn-grandchild-and-hang`, catch each injected OOM, and verify both recorded PIDs are gone before returning the OOM to `checkAllAllocationFailures`. The current `success` helper cannot prove descendant cleanup.

Jaruso (commented):
Blocking follow-up review.

The update substantially addresses both previous findings: the EOF-before-exit case now has a bounded wait, the Go oracle is built once rather than launched through `go run`, and validator invocations are placed in an isolated process group with direct parent/grandchild regression coverage. The current CI run is green across the required Linux, ARM, and macOS jobs.

I still do not think this is safe to merge. The new implementation has three remaining correctness gaps:

1. allocation/error unwinds can still kill only the direct PID and orphan descendants;
2. the post-EOF deadline uses the adjustable real-time clock instead of a monotonic elapsed-time clock; and
3. `-Dgo-bin` builds the oracle with one Go tool while runtime provenance still invokes `GO_BIN`/default `go`, which can fail or record the wrong compiler version.

Because GitHub does not allow this account to request changes on its own PR, this is submitted as a blocking `COMMENT` review. Please treat the inline comments as required changes. The original review threads should remain unresolved until these lifecycle and provenance paths are covered.

tests/bounded_process.zig
```diff
@@ -0,0 +1,365 @@
...
+};
+
+pub fn run(allocator: std.mem.Allocator, options: Options) std.mem.Allocator.Error!Result {
+    const io = compat.io();
+    const deadline_end_ms: ?i64 = if (options.deadline_ms == 0)
+        null
+    else
+        compat.milliTimestamp() + @as(i64, @intCast(options.deadline_ms));
+    var child = std.process.spawn(io, .{
+        .argv = options.argv,
```
Jaruso, line 71: **Blocking: this is an adjustable real-time deadline, not a hard elapsed-time deadline.**

`compat.milliTimestamp()` is backed by `std.Io.Clock.real`. NTP corrections or a manual/system clock change can move it backward, extending the validator past its configured bound, or forward, producing a premature timeout. The stream-reading half correctly uses an `.awake` timeout, so the process lifetime is currently measured against two different clocks.

A resource-control deadline must use one monotonic/awake clock from invocation start through pipe draining and final reaping.

Please use `std.Io.Clock.awake` (or a compatibility helper explicitly backed by it), for example:

```zig
fn awakeMilliseconds() i64 {
    return std.Io.Clock.awake.now(compat.io()).toMilliseconds();
}

const deadline_end_ms = if (options.deadline_ms == 0)
    null
else
    awakeMilliseconds() + @as(i64, @intCast(options.deadline_ms));
```

Then use the same awake clock in `waitForExit()`. Prefer deriving both the MultiReader timeout and wait-loop expiration from one absolute monotonic deadline so setup/read time cannot accidentally receive a second full budget.

build.zig
```diff
@@ -17,6 +17,7 @@ pub fn build(b: *std.Build) void {
     const require_static_system_libs = b.option(bool, "require-static-system-libs", "Require static linking for system libraries") orelse false;
     const static_executable = b.option(bool, "static-executable", "Build the tardi executable as a static binary") orelse false;
     const app_version = b.option([]const u8, "version", "Version string embedded in the tardi binary") orelse "dev";
+    const go_bin = b.option([]const u8, "go-bin", "Go command used to build the PKI crypto/x509 oracle") orelse "go";
     const tls_profile = b.option(
         TlsProfile,
```
Jaruso, line 22: **Blocking: `-Dgo-bin` is not propagated to runtime version discovery, so the recorded provenance can be wrong or the run can fail despite a successful build.**

This option now selects the Go tool that compiles `pki_go_validator`, but `loadRuntimeIdentity()` still reads `GO_BIN` and otherwise executes plain `go version`. Examples:

- `zig build -Dgo-bin=/opt/go1.26/bin/go test-pki-differential` can compile with Go 1.26 but record whatever unrelated `go` appears on PATH.
- If `/opt/go1.26/bin/go` exists but PATH has no `go`, the build succeeds and the differential test then fails during provenance discovery.

Please pass the selected tool through `pki_diff_options`, and ideally query the installed oracle binary so the provenance identifies the compiler that actually produced it:

```zig
pki_diff_options.addOption([]const u8, "go_bin", go_bin);
```

```zig
const go_version = try commandSummary(allocator, &.{
    pki_diff_options.go_bin,
    "version",
    pki_diff_options.go_validator_path,
});
```

Alternatively, have the Go oracle emit `runtime.Version()` in structured metadata. Add a validation case using an absolute `-Dgo-bin` with default `go` removed from PATH and assert that the run succeeds and records the selected compiler version.

Jaruso (commented):
**Re-review result: approved / good to merge.**

Re-reviewed the complete PR at `2da772b96c0feb27a79d92e38ac22bc23328375e` against issue #348's bounded subprocess-hardening slice and all prior review findings.

The previous blockers are resolved:

- one absolute `.awake` deadline now covers stream draining and final process termination;
- the Go oracle is compiled once and invoked directly rather than through `go run`;
- each validator runs in an isolated process group, and timeout/error/OOM unwinds terminate the group before any diagnostic allocation;
- direct-child reaping is installed as the unconditional post-spawn fallback;
- allocation-failure coverage now exercises and verifies a hanging grandchild;
- Go provenance is obtained with the configured `-Dgo-bin` against the actual installed oracle binary.

I also re-checked success, rejection, timeout, EOF-before-exit, signal, output overflow, malformed output, cleanup ordering, build dependencies, and artifact version discovery. The latest CI run is green, including the PKI differential core job and unit tests on Linux x86_64, Linux ARM64, and macOS.

No remaining blocking findings. This PR is good to merge as the bounded-validator-process slice of #348. It correctly uses `Refs #348`; the broader semantic taxonomy/corpus/artifact work remains outside this PR.

PR #457, opened 2026-07-18
tls: add bounded stream watermarks and backpressure

Jaruso:
## Summary

Closes #355.

Adds bounded, runtime-effective TLS stream watermarks while preserving the existing fixed-capacity, allocation-free `PureZigRecordStream` queues.

## BufferLimits model and defaults

- Adds `Watermark { low, high, hard }` and `BufferLimits` for inbound carrier ciphertext, inbound plaintext, outbound ciphertext, and handshake bytes.
- Defaults derive from existing compile-time capacities using 25% low / 75% high / capacity hard watermarks.
- Validation rejects zero/nonsensical `low < high <= hard` violations, hard limits above compile-time queue capacity, and hard limits that cannot hold one maximum legal fragment or the complete record-mode `handshake_output_reserve` batch.

## Compile-time capacity vs runtime limits

- The hot path still uses the existing fixed `ByteQueue` storage.
- Runtime hard limits sit inside those capacities; they do not allocate or grow dynamically.
- Existing constructors keep source compatibility by using defaults, with explicit `init*WithLimits` constructors for tighter policies.

## Queue ownership and accounting

| Queue | Owned bytes counted | Notes |
| --- | --- | --- |
| inbound ciphertext | `inbound_carrier.len + initial_parser.len + ciphertext_parser.len` | Stream-owned carrier bytes plus fragmented record bytes already copied into parsers |
| inbound plaintext | `inbound_plaintext.len` | Decrypted application bytes owned by the stream |
| outbound ciphertext | `outbound_ciphertext.len` | Sealed TLS records; accepted plaintext is not staged separately |
| handshake | `inbound_handshake.len` | Opened handshake bytes owned by record mode |

Pure-Zig application writes seal accepted plaintext directly into outbound ciphertext, so there is no hidden pending-plaintext queue. Queue and parser cleanup wipes vacated storage on discard, reset, terminal close, and deinit paths.

## Readiness and hysteresis

- Carrier reads latch paused at any inbound high watermark and resume only after all inbound queues drain to their lows.
- Plaintext writes latch paused at outbound ciphertext high and resume at/below outbound low.
- `wants_read` is false while read backpressure is latched unless an already-started parser-owned record needs bounded bytes to finish.
- `can_write_plaintext` is false while write backpressure is latched.
- `wants_write` remains true while draining ciphertext can make progress.
- Already-owned carrier ciphertext can still be processed while socket reads are paused.
- Carrier reads are capped to runtime inbound-ciphertext high room before starting a record and hard room while finishing an active parser-owned record.

## Hard-limit and atomicity behavior

- Runtime hard-limit failures are checked before sealing/opening records where sequence/parser state could otherwise move.
- Existing atomic guarantees remain covered: rejected plaintext writes do not advance write sequence, rejected inbound opens do not advance read sequence, and coalesced suffixes remain retry-safe.
- Borrowed direct-feed slices are capped to the remaining owned-byte hard room; unconsumed coalesced suffixes remain caller-owned and retryable.
- Hard-limit counters represent explicit rejection paths, not normal high-watermark pauses, side-effect-free capacity probes, or internal carrier retry backpressure.

## Metrics/snapshot interface

- Extends `EncryptedStream` with `bufferSnapshot()`.
- Snapshot includes current and peak bytes by queue, peak total, optional configured limits, whether limits are enforced, latched pause state, pause/resume counters, hard-limit counters, stalled-drive counts, and an accounting-boundary marker.
- Labels/cardinality remain bounded; no SNI, URL, connection id, request id, or arbitrary protocol labels are introduced.

## OpenSSL accounting boundary

- The OpenSSL adapter implements the same snapshot hook.
- It reports measurable `SSL_pending()` decrypted bytes and retry-derived pause/resume/stall state.
- It maintains `SSL_pending()` lifetime peaks at stream read/snapshot boundaries instead of relying only on snapshot polling.
- It marks the snapshot as `backend_opaque`, leaves unknown limits unset, and reports `limits_enforced = false`; opaque OpenSSL internal memory is not reported as complete stream-owned accounting.
- Borrowed pending-write caller memory is not counted as owned bytes.
- Close/shutdown/failure lifecycle transitions do not synthesize backpressure pause/resume events.

## HTTP consumer seam for #356

HTTP/1.1 and HTTP/2 consumers can use only backend-neutral APIs:

- `readiness().can_write_plaintext`
- `readiness().can_read_plaintext`
- `readiness().wants_read`
- `readiness().wants_write`
- `bufferSnapshot()` for pause reasons, usage, peaks, and limits when enforced/known

A mock HTTP consumer regression covers this without backend-specific casts.

## Validation

- `git diff --check` — passed
- `zig build test-tls --summary all --error-style verbose` — passed, 180/180 tests
- `zig build test --summary all --error-style verbose` — passed, 1447/1448 passed / 1 skipped
- `zig build test -Dtls-profile=appliance --summary all --error-style verbose` — passed, 1430/1431 passed / 1 skipped
- `zig build test-quic --summary all --error-style verbose` — passed, 311/311 tests
- `zig build test-integration --summary all --error-style verbose` — passed, 65/69 passed / 4 skipped
- `zig build test -Doptimize=ReleaseFast --summary all --error-style verbose` — passed, 1447/1448 passed / 1 skipped

## Review follow-up

- Parser-owned fragmented record bytes are included in inbound-ciphertext snapshots, peaks, watermarks, and hard-limit decisions.
- Side-effect-free capacity predicates no longer increment hard-limit counters; only actual API/append rejections do.
- Parser and queue storage is wiped on discard/reset/terminal cleanup paths, and snapshots drop to zero after teardown.
- Socket reads are capped by runtime inbound-ciphertext room before reading from the carrier, without permanently stalling active fragmented records.
- Internal carrier-to-parser transfers no longer double-count bytes or latch doubled peaks before queue discard.
- Direct coalesced borrowed feeds no longer reject solely because the borrowed suffix exceeds owned hard room.
- Internal retry backpressure behind destination hard limits no longer inflates hard-limit counters on repeated `drive()` calls.
- OpenSSL snapshots no longer claim Pure-Zig limits; they report opaque accounting, unenforced/unknown limits, retry-derived pause/resume/stall counters, lifetime pending-byte peaks, and no close-synthetic transitions.
- The failing Ubuntu pki allocation-failure cleanup path now tolerates empty/partial pid files left by a child that failed before recording its pid.

## Scope exclusions confirmed

This PR does not implement or depend on #334 credential hooks / closed PR #455, #347 SNI selection, #348 PKI corpus work, #356 HTTP dispatch, #357 KeyUpdate, or global/per-origin proxy accounting from #140.

### DIFF
--- a/src/http/tls_termination.zig
+++ b/src/http/tls_termination.zig
@@ -339,6 +339,8 @@ pub const TlsConnection = struct {
     stream_failure: ?encrypted_stream.Error = null,
     stream_buffer_peaks: encrypted_stream.QueueBytes = .{},
     stream_peak_total: usize = 0,
+    stream_pause_state: encrypted_stream.PauseState = .{},
+    stream_buffer_counters: encrypted_stream.BufferCounters = .{},
 
     pub fn deinit(self: *TlsConnection) void {
         if (opensslShouldShutdownOnDeinit(self)) {
@@ -449,6 +451,7 @@ const OpenSslPendingWrite = struct {
 fn opensslClearRetry(self: *TlsConnection) void {
     self.retry_direction = .none;
     self.pending_write = null;
+    opensslObserveBackpressure(self);
 }
 
 fn opensslShouldShutdownOnDeinit(self: *const TlsConnection) bool {
@@ -482,11 +485,40 @@ fn opensslSetRetry(self: *TlsConnection, ssl_error: c_int) void {
         c.SSL_ERROR_WANT_WRITE => .write,
         else => .none,
     };
+    opensslObserveBackpressure(self);
 }
 
 fn opensslSetWriteRetry(self: *TlsConnection, bytes: []const u8, ssl_error: c_int) void {
-    opensslSetRetry(self, ssl_error);
+    self.retry_direction = switch (ssl_error) {
+        c.SSL_ERROR_WANT_READ => .read,
+        c.SSL_ERROR_WANT_WRITE => .write,
+        else => .none,
+    };
     self.pending_write = OpenSslPendingWrite.init(bytes);
+    opensslObserveBackpressure(self);
+}
+
+fn opensslDerivedPauseState(self: *TlsConnection) encrypted_stream.PauseState {
+    const readiness = opensslStreamReadiness(self);
+    return .{
+        .inbound_read_paused = self.stream_lifecycle == .open and !self.peer_closed and !readiness.wants_read and !readiness.can_read_plaintext,
+        .plaintext_write_paused = self.stream_lifecycle == .open and !readiness.can_write_plaintext,
+    };
+}
+
+fn opensslObserveBackpressure(self: *TlsConnection) void {
+    const next = opensslDerivedPauseState(self);
+    if (!self.stream_pause_state.inbound_read_paused and next.inbound_read_paused) {
+        self.stream_buffer_counters.inbound_read_pauses += 1;
+    } else if (self.stream_pause_state.inbound_read_paused and !next.inbound_read_paused) {
+        self.stream_buffer_counters.inbound_read_resumes += 1;
+    }
+    if (!self.stream_pause_state.plaintext_write_paused and next.plaintext_write_paused) {
+        self.stream_buffer_counters.plaintext_write_pauses += 1;
+    } else if (self.stream_pause_state.plaintext_write_paused and !next.plaintext_write_paused) {
+        self.stream_buffer_counters.plaintext_write_resumes += 1;
+    }
+    self.stream_pause_state = next;
 }
 
 fn opensslStreamBackend(_: *anyopaque) encrypted_stream.BackendKind {
@@ -579,6 +611,7 @@ fn opensslAdvanceShutdown(self: *TlsConnection) encrypted_stream.Error!bool {
     }
     if (rc == 0) {
         self.retry_direction = .read;
+        opensslObserveBackpressure(self);
         return true;
     }
 
@@ -591,6 +624,7 @@ fn opensslAdvanceShutdown(self: *TlsConnection) encrypted_stream.Error!bool {
         c.SSL_ERROR_ZERO_RETURN => peer_closed: {
             self.peer_closed = true;
             self.retry_direction = .write;
+            opensslObserveBackpressure(self);
             break :peer_closed true;
         },
         else => opensslStreamFail(self, error.SocketWriteFailed),
@@ -601,6 +635,7 @@ fn opensslStreamClose(ptr: *anyopaque) void {
     const self: *TlsConnection = @ptrCast(@alignCast(ptr));
     if (self.stream_lifecycle == .closed or self.stream_lifecycle == .failed) return;
     self.close_requested = true;
+    opensslObserveBackpressure(self);
     _ = opensslAdvanceShutdown(self) catch {};
 }
 
@@ -647,12 +682,14 @@ fn opensslStreamDrive(ptr: *anyopaque) encrypted_stream.Error!encrypted_stream.D
         try opensslAdvanceShutdown(self)
     else
         false;
+    if (!made_progress and (self.stream_pause_state.inbound_read_paused or self.stream_pause_state.plaintext_write_paused)) {
+        self.stream_buffer_counters.stalled_drives += 1;
+    }
     return .{ .made_progress = made_progress, .readiness = opensslStreamReadiness(ptr) };
 }
 
 fn opensslStreamBufferSnapshot(ptr: *anyopaque) encrypted_stream.BufferSnapshot {
     const self: *TlsConnection = @ptrCast(@alignCast(ptr));
-    const readiness = opensslStreamReadiness(ptr);
     const current = encrypted_stream.QueueBytes{
         .inbound_plaintext = if (self.stream_lifecycle == .open) self.pending() else 0,
     };
@@ -662,10 +699,8 @@ fn opensslStreamBufferSnapshot(ptr: *anyopaque) encrypted_stream.BufferSnapshot 
         .current = current,
         .peak = self.stream_buffer_peaks,
         .peak_total = self.stream_peak_total,
-        .pause_state = .{
-            .inbound_read_paused = self.stream_lifecycle == .open and !self.peer_closed and !readiness.wants_read and !readiness.can_read_plaintext,
-            .plaintext_write_paused = self.stream_lifecycle == .open and !readiness.can_write_plaintext,
-        },
+        .pause_state = self.stream_pause_state,
+        .counters = self.stream_buffer_counters,
         .accounting_boundary = .backend_opaque,
     };
 }
@@ -1362,6 +1397,10 @@ test "openssl encrypted stream adapter conforms over production connection" {
 
     var scratch: [32]u8 = undefined;
     try clientWriteAll(pair.client_ssl, "client-payload");
+    const pending_snapshot = stream.bufferSnapshot();
+    try std.testing.expectEqual(encrypted_stream.AccountingBoundary.backend_opaque, pending_snapshot.accounting_boundary);
+    try std.testing.expect(!pending_snapshot.limits_enforced);
+    try std.testing.expect(pending_snapshot.limits == null);
     const partial = try stream.read(scratch[0..6]);
     try std.testing.expectEqualStrings("client", scratch[0..partial]);
     try std.testing.expect(stream.readiness().can_read_plaintext);
@@ -1394,6 +1433,20 @@ test "openssl encrypted stream preserves write retry direction under socket back
     const blocked_drive = try stream.drive();
     try std.testing.expect(!blocked_drive.made_progress);
     try std.testing.expect(blocked_drive.readiness.wants_write);
+    const blocked_snapshot = stream.bufferSnapshot();
+    try std.testing.expectEqual(encrypted_stream.AccountingBoundary.backend_opaque, blocked_snapshot.accounting_boundary);
+    try std.testing.expect(!blocked_snapshot.limits_enforced);
+    try std.testing.expect(blocked_snapshot.limits == null);
+    try std.testing.expect(blocked_snapshot.pause_state.plaintext_write_paused);
+    try std.testing.expectEqual(@as(u64, 1), blocked_snapshot.counters.plaintext_write_pauses);
+    try std.testing.expectEqual(@as(u64, 0), blocked_snapshot.counters.plaintext_write_resumes);
+    try std.testing.expectEqual(@as(u64, 1), blocked_snapshot.counters.stalled_drives);
+    const repeated_blocked_drive = try stream.drive();
+    try std.testing.expect(!repeated_blocked_drive.made_progress);
+    const repeated_blocked_snapshot = stream.bufferSnapshot();
+    try std.testing.expectEqual(@as(u64, 1), repeated_blocked_snapshot.counters.plaintext_write_pauses);
+    try std.testing.expectEqual(@as(u64, 0), repeated_blocked_snapshot.counters.plaintext_write_resumes);
+    try std.testing.expectEqual(@as(u64, 2), repeated_blocked_snapshot.counters.stalled_drives);
 
     var interleaved_read: [1]u8 = undefined;
     try std.testing.expectError(error.RetryOperationPending, stream.read(&interleaved_read));
@@ -1431,6 +1484,9 @@ test "openssl encrypted stream preserves write retry direction under socket back
     try std.testing.expectEqualSlices(u8, target, received);
     try std.testing.expect(!stream.readiness().wants_write);
     try std.testing.expect(stream.readiness().can_write_plaintext);
+    const resumed_snapshot = stream.bufferSnapshot();
+    try std.testing.expect(!resumed_snapshot.pause_state.plaintext_write_paused);
+    try std.testing.expectEqual(@as(u64, 1), resumed_snapshot.counters.plaintext_write_resumes);
 }
 
 test "openssl encrypted stream preserves close requested during pending write retry" {
--- a/src/http/tls_termination.zig
+++ b/src/http/tls_termination.zig
@@ -337,6 +337,8 @@ pub const TlsConnection = struct {
     close_requested: bool = false,
     peer_closed: bool = false,
     stream_failure: ?encrypted_stream.Error = null,
+    stream_buffer_peaks: encrypted_stream.QueueBytes = .{},
+    stream_peak_total: usize = 0,
 
     pub fn deinit(self: *TlsConnection) void {
         if (opensslShouldShutdownOnDeinit(self)) {
@@ -648,6 +650,26 @@ fn opensslStreamDrive(ptr: *anyopaque) encrypted_stream.Error!encrypted_stream.D
     return .{ .made_progress = made_progress, .readiness = opensslStreamReadiness(ptr) };
 }
 
+fn opensslStreamBufferSnapshot(ptr: *anyopaque) encrypted_stream.BufferSnapshot {
+    const self: *TlsConnection = @ptrCast(@alignCast(ptr));
+    const readiness = opensslStreamReadiness(ptr);
+    const current = encrypted_stream.QueueBytes{
+        .inbound_plaintext = if (self.stream_lifecycle == .open) self.pending() else 0,
+    };
+    self.stream_buffer_peaks.inbound_plaintext = @max(self.stream_buffer_peaks.inbound_plaintext, current.inbound_plaintext);
+    self.stream_peak_total = @max(self.stream_peak_total, current.total());
+    return .{
+        .current = current,
+        .peak = self.stream_buffer_peaks,
+        .peak_total = self.stream_peak_total,
+        .pause_state = .{
+            .inbound_read_paused = self.stream_lifecycle == .open and !self.peer_closed and !readiness.wants_read and !readiness.can_read_plaintext,
+            .plaintext_write_paused = self.stream_lifecycle == .open and !readiness.can_write_plaintext,
+        },
+        .accounting_boundary = .backend_opaque,
+    };
+}
+
 const openssl_stream_vtable = encrypted_stream.EncryptedStream.VTable{
     .backendFn = opensslStreamBackend,
     .readFn = opensslStreamRead,
@@ -655,6 +677,7 @@ const openssl_stream_vtable = encrypted_stream.EncryptedStream.VTable{
     .closeFn = opensslStreamClose,
     .readinessFn = opensslStreamReadiness,
     .driveFn = opensslStreamDrive,
+    .bufferSnapshotFn = opensslStreamBufferSnapshot,
 };
 
 pub fn lastOpenSslError(allocator: std.mem.Allocator) ?[]u8 {
--- a/src/http/tls_termination.zig
+++ b/src/http/tls_termination.zig
@@ -451,7 +451,7 @@ const OpenSslPendingWrite = struct {
 fn opensslClearRetry(self: *TlsConnection) void {
     self.retry_direction = .none;
     self.pending_write = null;
-    opensslObserveBackpressure(self);
+    if (self.stream_lifecycle == .open and !self.close_requested) opensslObserveBackpressure(self);
 }
 
 fn opensslShouldShutdownOnDeinit(self: *const TlsConnection) bool {
@@ -462,6 +462,7 @@ fn opensslStreamFail(self: *TlsConnection, err: encrypted_stream.Error) encrypte
     self.stream_lifecycle = .failed;
     opensslClearRetry(self);
     self.close_requested = false;
+    self.stream_pause_state = .{};
     self.stream_failure = err;
     return err;
 }
@@ -485,7 +486,7 @@ fn opensslSetRetry(self: *TlsConnection, ssl_error: c_int) void {
         c.SSL_ERROR_WANT_WRITE => .write,
         else => .none,
     };
-    opensslObserveBackpressure(self);
+    if (self.stream_lifecycle == .open and !self.close_requested) opensslObserveBackpressure(self);
 }
 
 fn opensslSetWriteRetry(self: *TlsConnection, bytes: []const u8, ssl_error: c_int) void {
@@ -495,7 +496,20 @@ fn opensslSetWriteRetry(self: *TlsConnection, bytes: []const u8, ssl_error: c_in
         else => .none,
     };
     self.pending_write = OpenSslPendingWrite.init(bytes);
-    opensslObserveBackpressure(self);
+    if (self.stream_lifecycle == .open and !self.close_requested) opensslObserveBackpressure(self);
+}
+
+fn opensslCurrentBufferUsage(self: *const TlsConnection) encrypted_stream.QueueBytes {
+    return .{
+        .inbound_plaintext = if (self.stream_lifecycle == .failed or self.stream_lifecycle == .closed) 0 else self.pending(),
+    };
+}
+
+fn opensslObserveBufferUsage(self: *TlsConnection) encrypted_stream.QueueBytes {
+    const current = opensslCurrentBufferUsage(self);
+    self.stream_buffer_peaks.inbound_plaintext = @max(self.stream_buffer_peaks.inbound_plaintext, current.inbound_plaintext);
+    self.stream_peak_total = @max(self.stream_peak_total, current.total());
+    return current;
 }
 
 fn opensslDerivedPauseState(self: *TlsConnection) encrypted_stream.PauseState {
@@ -531,11 +545,14 @@ fn opensslStreamRead(ptr: *anyopaque, out: []u8) encrypted_stream.Error!usize {
     if (self.stream_lifecycle != .open) return error.StreamClosed;
     if (self.pending_write != null) return error.RetryOperationPending;
     if (self.close_requested) return error.StreamClosed;
+    _ = opensslObserveBufferUsage(self);
     c.ERR_clear_error();
     const rc = c.SSL_read(self.ssl, out.ptr, @intCast(out.len));
+    _ = opensslObserveBufferUsage(self);
     if (rc > 0) {
         opensslClearRetry(self);
         opensslRefreshPeerClosed(self);
+        _ = opensslObserveBufferUsage(self);
         return @intCast(rc);
     }
     const ssl_error = c.SSL_get_error(self.ssl, rc);
@@ -600,6 +617,7 @@ fn opensslAdvanceShutdown(self: *TlsConnection) encrypted_stream.Error!bool {
     if (self.pending_write != null) return error.RetryOperationPending;
 
     self.stream_lifecycle = .closing;
+    self.stream_pause_state = .{};
     c.ERR_clear_error();
     const rc = c.SSL_shutdown(self.ssl);
     opensslRefreshPeerClosed(self);
@@ -611,7 +629,6 @@ fn opensslAdvanceShutdown(self: *TlsConnection) encrypted_stream.Error!bool {
     }
     if (rc == 0) {
         self.retry_direction = .read;
-        opensslObserveBackpressure(self);
         return true;
     }
 
@@ -624,7 +641,6 @@ fn opensslAdvanceShutdown(self: *TlsConnection) encrypted_stream.Error!bool {
         c.SSL_ERROR_ZERO_RETURN => peer_closed: {
             self.peer_closed = true;
             self.retry_direction = .write;
-            opensslObserveBackpressure(self);
             break :peer_closed true;
         },
         else => opensslStreamFail(self, error.SocketWriteFailed),
@@ -635,7 +651,6 @@ fn opensslStreamClose(ptr: *anyopaque) void {
     const self: *TlsConnection = @ptrCast(@alignCast(ptr));
     if (self.stream_lifecycle == .closed or self.stream_lifecycle == .failed) return;
     self.close_requested = true;
-    opensslObserveBackpressure(self);
     _ = opensslAdvanceShutdown(self) catch {};
 }
 
@@ -690,11 +705,7 @@ fn opensslStreamDrive(ptr: *anyopaque) encrypted_stream.Error!encrypted_stream.D
 
 fn opensslStreamBufferSnapshot(ptr: *anyopaque) encrypted_stream.BufferSnapshot {
     const self: *TlsConnection = @ptrCast(@alignCast(ptr));
-    const current = encrypted_stream.QueueBytes{
-        .inbound_plaintext = if (self.stream_lifecycle == .open) self.pending() else 0,
-    };
-    self.stream_buffer_peaks.inbound_plaintext = @max(self.stream_buffer_peaks.inbound_plaintext, current.inbound_plaintext);
-    self.stream_peak_total = @max(self.stream_peak_total, current.total());
+    const current = opensslObserveBufferUsage(self);
     return .{
         .current = current,
         .peak = self.stream_buffer_peaks,
@@ -1404,10 +1415,16 @@ test "openssl encrypted stream adapter conforms over production connection" {
     const partial = try stream.read(scratch[0..6]);
     try std.testing.expectEqualStrings("client", scratch[0..partial]);
     try std.testing.expect(stream.readiness().can_read_plaintext);
+    var buffered_snapshot = stream.bufferSnapshot();
+    try std.testing.expect(buffered_snapshot.current.inbound_plaintext > 0);
+    try std.testing.expect(buffered_snapshot.peak.inbound_plaintext >= buffered_snapshot.current.inbound_plaintext);
 
     const rest = try stream.read(scratch[0..16]);
     try std.testing.expectEqualStrings("-payload", scratch[0..rest]);
     try std.testing.expect(!stream.readiness().can_read_plaintext);
+    buffered_snapshot = stream.bufferSnapshot();
+    try std.testing.expectEqual(@as(usize, 0), buffered_snapshot.current.inbound_plaintext);
+    try std.testing.expect(buffered_snapshot.peak.inbound_plaintext > 0);
 
     const written = try stream.write("server-reply");
     try std.testing.expectEqual(@as(usize, "server-reply".len), written);
@@ -1489,6 +1506,30 @@ test "openssl encrypted stream preserves write retry direction under socket back
     try std.testing.expectEqual(@as(u64, 1), resumed_snapshot.counters.plaintext_write_resumes);
 }
 
+test "openssl encrypted stream records pending plaintext peak before first snapshot" {
+    const allocator = std.testing.allocator;
+    var pair = try makeTestTlsPair(allocator);
+    defer pair.deinit();
+
+    const stream = pair.server.stream();
+    try clientWriteAll(pair.client_ssl, "client-payload");
+
+    var scratch: [6]u8 = undefined;
+    try std.testing.expectEqual(@as(usize, 6), try stream.read(&scratch));
+    try std.testing.expectEqualStrings("client", &scratch);
+
+    var rest: [16]u8 = undefined;
+    const rest_len = try stream.read(&rest);
+    try std.testing.expectEqualStrings("-payload", rest[0..rest_len]);
+
+    const snapshot = stream.bufferSnapshot();
+    try std.testing.expectEqual(@as(usize, 0), snapshot.current.inbound_plaintext);
+    try std.testing.expect(snapshot.peak.inbound_plaintext >= "-payload".len);
+    try std.testing.expect(snapshot.peak_total >= "-payload".len);
+    try std.testing.expect(!snapshot.limits_enforced);
+    try std.testing.expect(snapshot.limits == null);
+}
+
 test "openssl encrypted stream preserves close requested during pending write retry" {
     const allocator = std.testing.allocator;
     var pair = try makeTestTlsPair(allocator);
@@ -1547,6 +1588,48 @@ test "openssl encrypted stream preserves close requested during pending write re
     try std.testing.expectEqual(OpenSslStreamLifecycle.closed, pair.server.stream_lifecycle);
     try std.testing.expectEqualSlices(u8, blocked.expected.items, read_ctx.out.items);
     try encrypted_stream.expectClosedConformance(stream);
+}
+
+test "openssl encrypted stream close does not synthesize pause transitions" {
+    const allocator = std.testing.allocator;
+
+    {
+        var pair = try makeTestTlsPair(allocator);
+        defer pair.deinit();
+        const stream = pair.server.stream();
+
+        const before = stream.bufferSnapshot();
+        try std.testing.expectEqual(@as(u64, 0), before.counters.inbound_read_pauses);
+        try std.testing.expectEqual(@as(u64, 0), before.counters.plaintext_write_pauses);
+
+        stream.close();
+        const after = stream.bufferSnapshot();
+        try std.testing.expect(!after.pause_state.inbound_read_paused);
+        try std.testing.expect(!after.pause_state.plaintext_write_paused);
+        try std.testing.expectEqual(@as(u64, 0), after.counters.inbound_read_pauses);
+        try std.testing.expectEqual(@as(u64, 0), after.counters.inbound_read_resumes);
+        try std.testing.expectEqual(@as(u64, 0), after.counters.plaintext_write_pauses);
+        try std.testing.expectEqual(@as(u64, 0), after.counters.plaintext_write_resumes);
+    }
+
+    {
+        var pair = try makeTestTlsPair(allocator);
+        defer pair.deinit();
+        const stream = pair.server.stream();
+        var blocked: ForcedWriteBlock = .{};
+        defer blocked.deinit(allocator);
+        try forceOpenSslWriteBackpressure(allocator, &pair, stream, &blocked);
+
+        const before_close = stream.bufferSnapshot();
+        try std.testing.expect(before_close.pause_state.plaintext_write_paused);
+        try std.testing.expectEqual(@as(u64, 1), before_close.counters.plaintext_write_pauses);
+
+        stream.close();
+        const after_close = stream.bufferSnapshot();
+        try std.testing.expect(after_close.pause_state.plaintext_write_paused);
+        try std.testing.expectEqual(@as(u64, 1), after_close.counters.plaintext_write_pauses);
+        try std.testing.expectEqual(@as(u64, 0), after_close.counters.plaintext_write_resumes);
+    }
 }
 
 test "openssl encrypted stream cleanup skips shutdown with pending write retry" {
--- a/src/tls/encrypted_stream.zig
+++ b/src/tls/encrypted_stream.zig
@@ -724,11 +724,16 @@ pub const PureZigRecordStream = struct {
             self.hasHardRoom(.outbound_ciphertext, self.outbound_ciphertext.len, handshake_output_reserve);
     }
 
+    fn hasActiveInboundParser(self: *const PureZigRecordStream) bool {
+        return self.initial_parser.len > 0 or self.ciphertext_parser.len > 0;
+    }
+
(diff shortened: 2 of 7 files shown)


src/tls/encrypted_stream.zig
```diff
@@ -408,13 +578,149 @@ pub const PureZigRecordStream = struct {
...
+                .plaintext_write_paused = self.write_backpressured,
+            },
+            .counters = self.buffer_counters,
+            .accounting_boundary = .complete_stream_owned,
+        };
+    }
+
+    fn currentQueueBytes(self: *const PureZigRecordStream) QueueBytes {
+        return .{
+            .inbound_ciphertext = self.inbound_carrier.len,
```
Jaruso: **Blocking: fragmented record bytes are omitted from both accounting and enforcement.**

`record_codec.Parser.feedOne()` copies incomplete headers/bodies into `initial_parser.pending` or `ciphertext_parser.pending` and increments the parser's `len`. The carrier queue then discards the consumed bytes. At that point this snapshot can report `inbound_ciphertext = 0` even though the stream still owns almost an entire TLS record in a parser.

That has three consequences:

1. `accounting_boundary = .complete_stream_owned` is not true.
2. `peak_total` and inbound-ciphertext peaks miss fragmented-record ownership.
3. The runtime inbound-ciphertext hard limit applies only to `inbound_carrier`, so queue + parser ownership can exceed the configured hard limit.

Please account for all persistent stream-owned carrier bytes and use the same aggregate for high/low and hard-limit decisions, for example:

```zig
fn inboundCiphertextOwned(self: *const PureZigRecordStream) usize {
    return self.inbound_carrier.len +
        self.initial_parser.len +
        self.ciphertext_parser.len;
}

fn currentQueueBytes(self: *const PureZigRecordStream) QueueBytes {
    return .{
        .inbound_ciphertext = self.inboundCiphertextOwned(),
        .inbound_plaintext = self.inbound_plaintext.len,
        .outbound_ciphertext = self.outbound_ciphertext.len,
        .handshake = self.inbound_handshake.len,
    };
}
```

`refreshBackpressure()` and inbound hard-room checks need to use that same aggregate. Also call the accounting/backpressure refresh after parser-only mutations in direct/manual feed paths; an incomplete fragment currently changes parser ownership without any `noteQueueMutation()` call.

Please add tests that feed only part of a TLS header and body, then assert snapshot current/peak values and hard-bound behavior before completing the record.

src/tls/encrypted_stream.zig
```diff
@@ -408,13 +578,149 @@ pub const PureZigRecordStream = struct {
...
+            .inbound_plaintext => self.buffer_limits.inbound_plaintext,
+            .outbound_ciphertext => self.buffer_limits.outbound_ciphertext,
+            .handshake => self.buffer_limits.handshake,
+        };
+    }
+
+    fn hardRoom(self: *PureZigRecordStream, queue: BufferQueue, current_len: usize, needed: usize) bool {
+        const hard = self.watermarkFor(queue).hard;
+        if (needed > hard or current_len > hard - needed) {
+            self.buffer_counters.hard_limits.increment(queue);
```
Jaruso: **Blocking: a capacity predicate must not increment the hard-limit counter.**

`hardRoom()` is called by `canReserveInbound*`, `canReserveOutboundRecord()`, and `canReserveHandshakeOutputBatch()`. Those helpers are normal backpressure/preflight probes used repeatedly by `drive()`. Consequently, waiting for queued handshake output to drain can increment `hard_limits.outbound_ciphertext` on every drive attempt even though no append was attempted and no hard-limit overflow occurred.

That directly conflicts with the issue contract: hard-limit counters represent explicit rejection/overflow paths, not normal high-watermark or atomic-batch pauses.

Please separate the pure query from the counted rejection, e.g.:

```zig
fn hasHardRoom(self: *const PureZigRecordStream, queue: BufferQueue, current: usize, needed: usize) bool {
    const hard = self.watermarkFor(queue).hard;
    return needed <= hard and current <= hard - needed;
}

fn rejectHardLimit(self: *PureZigRecordStream, queue: BufferQueue, err: Error) Error {
    self.buffer_counters.hard_limits.increment(queue);
    return err;
}
```

Use `hasHardRoom()` for readiness/drive/preflight polling. Increment only at the actual API/append rejection point, once per rejected operation. Add a test with blocked handshake output and repeated `drive()` calls proving the hard-limit counter remains stable while stall/readiness behavior is reported separately.

src/tls/encrypted_stream.zig
```diff
@@ -408,13 +578,149 @@ pub const PureZigRecordStream = struct {
...
+        }
+    }
+
+    fn noteQueueMutation(self: *PureZigRecordStream) void {
+        self.updateQueuePeaks();
+        self.refreshBackpressure();
+    }
+
+    fn clearOwnedQueues(self: *PureZigRecordStream) void {
+        self.inbound_carrier.clear();
```
Jaruso, line 673: **Blocking: terminal cleanup does not cover all owned buffers or securely remove stale plaintext.**

This helper clears the four queue lengths, but it does not reset/wipe `initial_parser` or `ciphertext_parser`. Once parser-owned bytes are included in the snapshot, close-during-handshake and other paths calling only `clearOwnedQueues()` will still retain/account parser bytes.

There is also a separate stale-data problem in `ByteQueue`: `discard()` shifts bytes and leaves the vacated tail unchanged, while `clear()` zeroes only `buf[0..len]`. If plaintext is fully read first, `len == 0` and a later failure/deinit wipes nothing; the fixed buffer can still contain the discarded plaintext/handshake bytes. This fails the maintainer requirement that teardown/failure paths not retain sensitive bytes.

Expected direction:

```zig
fn discard(self: *Self, count: usize) Error!void {
    if (count > self.len) return error.WouldBlock;
    const old_len = self.len;
    const remaining = old_len - count;
    std.mem.copyForwards(u8, self.buf[0..remaining], self.buf[count..old_len]);
    @memset(self.buf[remaining..old_len], 0);
    self.len = remaining;
}

fn clear(self: *Self) void {
    @memset(self.buf[0..], 0);
    self.len = 0;
}
```

The terminal clear operation should also reset/wipe parser storage and be invoked when a graceful local close actually reaches `.closed`; the current close-notify completion paths can leave inbound queues populated but permanently unreadable because `readPlaintext()` returns `StreamClosed` first.

Please add byte-level tests for consumed-then-deinitialized plaintext/handshake data, plus snapshot-zero tests for every terminal path.

src/tls/encrypted_stream.zig
```diff
@@ -882,7 +1199,7 @@ pub const PureZigRecordStream = struct {
                         self.carrier_eof = true;
                         made_progress = true;
                     } else {
-                        self.inbound_carrier.append(buf[0..read_len]) catch |err| return self.fail(err);
+                        self.appendInboundCarrier(buf[0..read_len]) catch |err| return self.fail(err);
```
Jaruso, line 1297: **Blocking: a valid tight runtime policy can consume socket bytes and then fail at the hard limit instead of pausing.**

The socket read size is capped only by the compile-time queue capacity and drive budget. It is not capped by the runtime inbound-ciphertext hard room. For a valid policy where `high` is close to `hard`, the stream can be below `high`, request a 4096-byte read, consume those bytes from the carrier, and then have `appendInboundCarrier()` return `CarrierInputBufferFull`; `drive()` converts that into a terminal stream failure.

Normal event-loop operation is required to pause before the hard limit. Compute the read cap from runtime-owned bytes before calling the carrier:

```zig
const owned = self.inboundCiphertextOwned();
const hard = self.buffer_limits.inbound_ciphertext.hard;
const hard_room = hard - owned;
const read_cap = @min(
    buf.len,
    @min(
        self.inbound_carrier.available(),
        @min(drive_read_budget - read_total, hard_room),
    ),
);
if (read_cap == 0) {
    self.refreshBackpressure();
    break;
}
```

You may additionally cap at the distance to `high` so the latch occurs exactly at the configured threshold. The key requirement is that no bytes are removed from the socket unless the stream can own them without exceeding the runtime hard bound.

Add a controlled carrier test with `high`/`hard` separated by less than `drive_read_chunk`; it should pause/latch deterministically, never fail, and never exceed hard.

Jaruso (commented):
## Review result: **do not merge yet**

GitHub does not permit the PR author to submit a formal `REQUEST_CHANGES` review on their own PR, so this is posted as a comment review. Treat the findings below as blocking.

I reviewed the implementation against both issue #355's acceptance criteria and the stricter 2026-07-17 maintainer implementation plan. The overall shape is good—fixed-capacity queues remain intact, the backend-neutral snapshot seam is useful, and the high/low latch mechanics are moving in the right direction—but the current implementation does not yet satisfy the accounting, hard-limit, OpenSSL, teardown, and deterministic-test requirements.

The inline comments cover the primary code defects. In addition, the following acceptance items still need explicit regression coverage before this can close #355:

1. **Large repeated transfer:** repeatedly move application data through a constrained stream and assert every owned queue, parser buffer, and total stays at or below the configured hard limit and compile-time capacity.
2. **Fragmented/coalesced records at runtime limits:** use custom limits, stop on a partial header and partial body, retry exact suffixes, and assert parser consumption/read sequence are unchanged on rejection.
3. **Permanently stalled handshake:** prove repeated `drive()` calls remain bounded, report stable readiness, do not inflate pause/hard-limit transitions, and count only genuine stalls.
4. **Partial carrier writes crossing low:** fill outbound ciphertext above high, drain it through a carrier in partial writes, and prove exactly one resume occurs only at/below low.
5. **Every terminal path:** EOF, fatal alert, peer/local `close_notify`, cancellation/close during handshake, failure, and deinit must all end with correct current accounting and no inaccessible owned bytes.
6. **Sensitive-buffer cleanup:** verify queue/parser storage does not retain discarded plaintext or handshake bytes after read, failure, close, or deinit.
7. **Mock HTTP resume behavior:** the mock consumer test currently observes only paused states. It must also drain below low and prove backend-neutral read/write registration resumes without casts.
8. **OpenSSL snapshot conformance:** exercise real retry transitions and `SSL_pending()`, validate the accounting boundary, ensure no unenforced Pure-Zig limits are exposed, and verify pause/resume/stall counters.
9. **Validation matrix:** cover zero, `low >= high`, `high > hard`, below-reserve hard limits, and over-capacity hard limits for every queue—not only a representative subset.

The current head also has a failing `Test (ubuntu-latest)` / `Run unit tests` GitHub Actions job. The ARM and macOS test jobs passed, but the failing required job still needs to be diagnosed and made green before merge.

Once the inline correctness issues and the above deterministic cases are addressed, this should receive another full review rather than only a spot-check of the fixes.

src/http/tls_termination.zig
```diff
@@ -648,13 +650,34 @@ fn opensslStreamDrive(ptr: *anyopaque) encrypted_stream.Error!encrypted_stream.D
...
+    self.stream_peak_total = @max(self.stream_peak_total, current.total());
+    return .{
+        .current = current,
+        .peak = self.stream_buffer_peaks,
+        .peak_total = self.stream_peak_total,
+        .pause_state = .{
+            .inbound_read_paused = self.stream_lifecycle == .open and !self.peer_closed and !readiness.wants_read and !readiness.can_read_plaintext,
+            .plaintext_write_paused = self.stream_lifecycle == .open and !readiness.can_write_plaintext,
+        },
+        .accounting_boundary = .backend_opaque,
```
Jaruso, line 715: **Blocking: the OpenSSL snapshot currently reports unenforced Pure-Zig limits and incomplete transition metrics.**

Because `.limits` and `.counters` are omitted here, `BufferSnapshot` silently fills them with `BufferLimits.defaults()` and zero counters. Those limits describe `PureZigRecordStream` queue capacities; this OpenSSL connection does not enforce them. A backend-neutral HTTP/metrics consumer therefore sees apparently effective low/high/hard values that are false.

The adapter also derives pause state only when a snapshot is polled and never records pause/resume or stalled-drive transitions, despite the required common snapshot contract. Its peak is likewise only the largest value observed during snapshot calls, not a transition-maintained peak.

Please make unknown/unenforced limits representable rather than substituting native defaults. One possible shape is:

```zig
pub const BufferSnapshot = struct {
    current: QueueBytes = .{},
    peak: QueueBytes = .{},
    peak_total: usize = 0,
    limits: ?BufferLimits = null,
    limits_enforced: bool = false,
    // ...
};
```

Pure Zig would set `limits` and `limits_enforced = true`; OpenSSL would leave them unknown unless real BIO/adapter limits are configured. Track retry-derived pause transitions in `TlsConnection` when retry/lifecycle state changes, and expose those counters truthfully. Hard-limit counters should remain absent/zero only when clearly marked unsupported by the opaque boundary.

Please add real OpenSSL conformance tests covering `SSL_pending()`, WANT_READ/WANT_WRITE pause and resume transitions, stable stalled drives, peaks, and the fact that no Pure-Zig limit is claimed.

src/tls/encrypted_stream.zig
```diff
@@ -408,13 +579,171 @@ pub const PureZigRecordStream = struct {
...
+    }
+
+    fn inboundCiphertextReadRoom(self: *const PureZigRecordStream) usize {
+        const owned = self.inboundCiphertextOwned();
+        const hard = self.buffer_limits.inbound_ciphertext.hard;
+        if (owned >= hard) return 0;
+        const hard_room = hard - owned;
+        const high = self.buffer_limits.inbound_ciphertext.high;
+        const high_room = if (owned < high) high - owned else 0;
+        return @min(hard_room, high_room);
```
Jaruso, line 735: **Blocking: capping every socket read at `high` can permanently stall a valid fragmented record.**

Validation only requires the inbound-ciphertext **hard** limit to hold one maximum TLS record; `high` may legally be much smaller. Once an incomplete record grows to `high`, `refreshBackpressure()` latches `read_backpressured`, `canAcceptCarrierRead()` becomes false, and the stream can never read the bytes needed to complete that record. There is no plaintext or handshake output to drain, so repeated `drive()` calls stall forever.

The new `carrier reads cap at runtime high before hard limit` test actually constructs this terminal state: it leaves `ciphertext_parser.len == high`, asserts reads are paused, and never proves that the record can finish and the connection can resume.

A record already accepted before the high watermark must be allowed to finish, while still bounded by `hard`. One possible design is to distinguish **starting another record** from **continuing the active parser**:

```zig
fn hasPartialInboundRecord(self: *const PureZigRecordStream) bool {
    return self.initial_parser.len != 0 or self.ciphertext_parser.len != 0;
}

fn inboundCiphertextReadRoom(self: *const PureZigRecordStream) usize {
    const owned = self.inboundCiphertextOwned();
    const hard = self.buffer_limits.inbound_ciphertext.hard;
    if (owned >= hard) return 0;

    const hard_room = hard - owned;
    if (self.hasPartialInboundRecord()) {
        // Ideally cap this to the exact remaining bytes of the active record.
        return hard_room;
    }

    const high = self.buffer_limits.inbound_ciphertext.high;
    return if (owned < high) @min(hard_room, high - owned) else 0;
}
```

`canAcceptCarrierRead()` / readiness must likewise permit continuation of an already-started record even if the high latch is set, without allowing arbitrary new records beyond the threshold. Alternatively, if strict pause-at-high semantics are required, validation must reject `inbound_ciphertext.high < max_ciphertext_record_len`; that is a more restrictive policy than the issue plan currently specifies.

Please add a deterministic test using a **valid maximum-sized encrypted record**, a high watermark below its encoded length, and fragmented carrier reads. It must complete the record, keep current/peak ownership at or below hard, avoid busy loops, and then resume normal reads after the destination queue drains.

src/tls/encrypted_stream.zig
```diff
@@ -663,50 +992,70 @@ pub const PureZigRecordStream = struct {
...
+        } else {
+            if (!self.canProcessCarrierInput()) return error.WouldBlock;
+        }
+        if (direct_external_feed and !self.hasHardRoom(.inbound_ciphertext, self.inboundCiphertextOwned(), bytes.len)) return self.rejectHardLimit(.inbound_ciphertext, error.WouldBlock);
+        if (self.inbound_plaintext.available() < record_codec.max_plaintext_fragment_len or self.inbound_handshake.available() < record_codec.max_plaintext_fragment_len) return error.WouldBlock;
+        if (!self.hasHardRoom(.inbound_plaintext, self.inbound_plaintext.len, record_codec.max_plaintext_fragment_len)) return self.rejectHardLimit(.inbound_plaintext, error.WouldBlock);
+        if (!self.hasHardRoom(.handshake, self.inbound_handshake.len, record_codec.max_plaintext_fragment_len)) return self.rejectHardLimit(.handshake, error.WouldBlock);
         var sink = record_codec.RecordSink(1, record_codec.max_ciphertext_fragment_len){};
         const consumed = feedUntilOneRecord(&self.ciphertext_parser, bytes, &sink) catch |err| return self.fail(err);
+        self.noteQueueMutation();
```
Jaruso: **Blocking: internal carrier-to-parser transfer temporarily counts the same ciphertext twice and can spuriously latch backpressure.**

For the production carrier path, `bytes` is a slice of `inbound_carrier`. `feedUntilOneRecord()` copies an incomplete fragment into `ciphertext_parser`, then this call records a snapshot **before** `processCarrierInputBudget()` discards the same consumed bytes from `inbound_carrier`.

For a 40-byte partial record with `{ low = 32, high = 64 }`, the sequence is:

1. queue owns 40 bytes;
2. parser copies those 40 bytes;
3. this `noteQueueMutation()` observes `40 + 40 = 80` and latches pause;
4. queue discards 40, leaving the parser at 40;
5. 40 is above low, so the latch never clears and the record cannot receive its suffix.

This also makes `peak.inbound_ciphertext` report a duplicate transfer copy rather than the stable ownership boundary intended by the queue model.

Defer parser-only accounting for internal carrier feeds until after the queue discard. Direct/manual feeds still need an immediate refresh because there is no owned source queue:

```zig
const consumed = try feedUntilOneRecord(&self.ciphertext_parser, bytes, &sink);
if (direct_external_feed) self.noteQueueMutation();
```

and keep the existing refresh after:

```zig
try self.inbound_carrier.discard(consumed);
self.noteQueueMutation();
```

The driver-owned handshake path has the same ordering issue in `feedHandshakeToDriver()`. Add a carrier regression where an incomplete fragment is larger than `low` but smaller than `high / 2`; transferring it into the parser must not create a pause or a doubled peak.

src/tls/encrypted_stream.zig
```diff
@@ -663,50 +992,70 @@ pub const PureZigRecordStream = struct {
...
         };
-        self.outbound_ciphertext.append(record) catch return;
+        self.appendOutboundCiphertext(record) catch return;
     }
 
     pub fn feedHandshakeCiphertext(self: *PureZigRecordStream, epoch: events.EncryptionEpoch, bytes: []const u8) Error!usize {
         if (self.failed) |err| return err;
         if (self.lifecycle == .closed or self.lifecycle == .failed) return error.StreamClosed;
+        if (self.read_backpressured) return error.WouldBlock;
+        if (!self.hasHardRoom(.inbound_ciphertext, self.inboundCiphertextOwned(), bytes.len)) return self.rejectHardLimit(.inbound_ciphertext, error.WouldBlock);
```
Jaruso: **Blocking: this treats the entire borrowed caller slice as stream-owned ciphertext.**

`feedOne()` consumes at most one record and returns an exact suffix boundary. A caller may therefore pass two or more coalesced records while the stream only copies/owns the first record. With a valid policy whose hard limit is one maximum record, a two-record slice is rejected here solely because `bytes.len > hard`, even though the first record fits and the unconsumed suffix remains caller-owned.

That conflicts with both the ownership rule (borrowed caller memory is not counted) and the required retry-safe coalesced-record behavior.

Cap what is offered to the parser by the actual remaining owned-byte room rather than rejecting based on the full borrowed slice:

```zig
const owned = self.inboundCiphertextOwned();
const hard = self.buffer_limits.inbound_ciphertext.hard;
if (owned >= hard) {
    return self.rejectHardLimit(.inbound_ciphertext, error.WouldBlock);
}
const feed_len = @min(bytes.len, hard - owned);
const consumed = try feedUntilOneRecord(parser, bytes[0..feed_len], &sink);
```

Apply the same rule to `feedCiphertextInternal(..., true)`. Add a test with two coalesced maximum-size records and `hard == max_ciphertext_record_len`: the first call must consume exactly the first record without counting a limit violation, and the suffix must remain retryable.

src/tls/encrypted_stream.zig
```diff
@@ -663,50 +992,70 @@ pub const PureZigRecordStream = struct {
...
         if (self.peer_closed) return bytes.len;
-        if (!self.canAcceptCarrierRead()) return error.WouldBlock;
+        if (direct_external_feed) {
+            if (!self.canAcceptCarrierRead()) return error.WouldBlock;
+        } else {
+            if (!self.canProcessCarrierInput()) return error.WouldBlock;
+        }
+        if (direct_external_feed and !self.hasHardRoom(.inbound_ciphertext, self.inboundCiphertextOwned(), bytes.len)) return self.rejectHardLimit(.inbound_ciphertext, error.WouldBlock);
+        if (self.inbound_plaintext.available() < record_codec.max_plaintext_fragment_len or self.inbound_handshake.available() < record_codec.max_plaintext_fragment_len) return error.WouldBlock;
+        if (!self.hasHardRoom(.inbound_plaintext, self.inbound_plaintext.len, record_codec.max_plaintext_fragment_len)) return self.rejectHardLimit(.inbound_plaintext, error.WouldBlock);
```
Jaruso: **Blocking: normal processing of already-owned carrier bytes can still inflate hard-limit counters on every `drive()`.**

When `direct_external_feed == false`, this is not an explicit API overflow. The bytes are already in `inbound_carrier`, and failure to reserve a destination record is ordinary backpressure. `processCarrierInputBudget()` catches `WouldBlock`, retains the same queue, and retries on the next `drive()`. Because this branch calls `rejectHardLimit()`, every retry increments `hard_limits.inbound_plaintext` or `hard_limits.handshake`.

For example, with plaintext at/above high and less than one maximum fragment of hard room remaining, one queued ciphertext record causes the counter to grow indefinitely while the consumer is simply expected to drain plaintext.

Use side-effect-free checks for internal processing and count only direct/manual API rejection or an actual append rejection:

```zig
if (!self.hasHardRoom(.inbound_plaintext, self.inbound_plaintext.len,
        record_codec.max_plaintext_fragment_len))
{
    if (direct_external_feed) {
        return self.rejectHardLimit(.inbound_plaintext, error.WouldBlock);
    }
    return error.WouldBlock;
}
```

Do the same for the handshake reserve. Add a test with owned carrier ciphertext blocked behind a nearly-full destination queue, call `drive()` repeatedly, and assert the hard-limit counters remain unchanged while `stalled_drives` increases.

src/http/tls_termination.zig
```diff
@@ -645,16 +682,37 @@ fn opensslStreamDrive(ptr: *anyopaque) encrypted_stream.Error!encrypted_stream.D
...
+    }
     return .{ .made_progress = made_progress, .readiness = opensslStreamReadiness(ptr) };
 }
 
+fn opensslStreamBufferSnapshot(ptr: *anyopaque) encrypted_stream.BufferSnapshot {
+    const self: *TlsConnection = @ptrCast(@alignCast(ptr));
+    const current = encrypted_stream.QueueBytes{
+        .inbound_plaintext = if (self.stream_lifecycle == .open) self.pending() else 0,
+    };
+    self.stream_buffer_peaks.inbound_plaintext = @max(self.stream_buffer_peaks.inbound_plaintext, current.inbound_plaintext);
```
Jaruso: **Blocking: the OpenSSL `peak` values are still snapshot-sampled, not lifetime peaks, and close can create false pause/resume events.**

`stream_buffer_peaks` is updated only inside `bufferSnapshot()`. If the application receives pending decrypted bytes, reads them, and asks for a snapshot afterward, the reported lifetime peak remains zero even though `SSL_pending()` was measurable at the stream API boundary. The conformance test takes a snapshot before reading but does not assert current or peak bytes, so it does not protect the contract.

Also, `opensslStreamClose()` calls `opensslObserveBackpressure()` while the lifecycle is still `.open` but `close_requested == true`. `opensslDerivedPauseState()` then interprets the intentional close as both an inbound-read pause and plaintext-write pause; `opensslAdvanceShutdown()` immediately moves to `.closing` and records matching resumes. Those are lifecycle transitions, not backpressure events.

Expected direction:

```zig
fn opensslObserveBufferUsage(self: *TlsConnection) encrypted_stream.QueueBytes {
    const current = encrypted_stream.QueueBytes{
        .inbound_plaintext = self.pending(),
    };
    self.stream_buffer_peaks.inbound_plaintext = @max(
        self.stream_buffer_peaks.inbound_plaintext,
        current.inbound_plaintext,
    );
    self.stream_peak_total = @max(self.stream_peak_total, current.total());
    return current;
}
```

Call this at least before/after `SSL_read()` and from `bufferSnapshot()`, so peaks do not depend on consumers polling at the right moment. Report measurable pending bytes while the `SSL` object still owns them, including during closing, rather than forcing current to zero solely from the adapter lifecycle.

For transitions, remove the pre-shutdown observation or make `opensslDerivedPauseState()` explicitly exclude `close_requested`; close/EOF/failure must not increment pause/resume counters. Add tests that (1) read pending data before the first snapshot and still observe the lifetime peak, and (2) close an idle and a backpressured connection without introducing synthetic pause/resume events.

Jaruso (commented):
Second review of the updated head. The previous cleanup issue and the false Pure-Zig limits in the OpenSSL snapshot are addressed, and CI is now green. However, this is still **not ready to merge**: the current carrier path can permanently pause in the middle of a valid fragmented record, parser transfers can double-count bytes and latch backpressure spuriously, direct feed rejects valid coalesced borrowed input, internal retries can inflate hard-limit counters, and OpenSSL peaks/close transitions are not yet truthful lifetime metrics. Detailed fixes and regression cases are inline.

src/tls/encrypted_stream.zig
```diff
@@ -408,13 +579,176 @@ pub const PureZigRecordStream = struct {
...
+    fn hasActiveInboundParser(self: *const PureZigRecordStream) bool {
+        return self.initial_parser.len > 0 or self.ciphertext_parser.len > 0;
+    }
+
+    fn inboundCiphertextReadRoom(self: *const PureZigRecordStream) usize {
+        const owned = self.inboundCiphertextOwned();
+        const hard = self.buffer_limits.inbound_ciphertext.hard;
+        if (owned >= hard) return 0;
+        const hard_room = hard - owned;
+        if (self.hasActiveInboundParser()) return hard_room;
```
Jaruso: **Blocking: the active-parser exception still permits unrelated records to bypass the high watermark.**

Once either parser has one byte, this returns the entire remaining **hard** room. The socket read is still allowed to request up to `drive_read_chunk` (4096 bytes), so a parser that needs only a few bytes to finish can pull a coalesced suffix containing one or more new records into `inbound_carrier`. Those suffix records were not needed to finish the active record, but they are now owned above `high`.

The new regression only supplies one record, so the carrier happens to return exactly the needed suffix and this behavior is not exercised.

The continuation allowance should be parser-specific and limited to the current record's exact remaining encoded bytes. Conceptually:

```zig
fn inboundCiphertextReadRoom(self: *const PureZigRecordStream) Error!usize {
    const owned = self.inboundCiphertextOwned();
    const limits = self.buffer_limits.inbound_ciphertext;
    if (owned >= limits.hard) return 0;
    const hard_room = limits.hard - owned;

    if (self.activeParserForReadEpoch()) |parser| {
        const needed = try parser.pendingRecordBytesNeeded();
        return @min(hard_room, needed);
    }

    return if (owned < limits.high)
        @min(hard_room, limits.high - owned)
    else
        0;
}
```

For an incomplete header, read only the remaining header bytes, parse its declared length, and then recompute the exact body remainder. Do not use `hasActiveInboundParser()` across both epochs to grant an unrestricted exception.

Please add a controlled carrier test with two coalesced records: fragment the first at `high`, allow only the bytes required to finish it, and prove the second record remains unread until normal low-watermark resumption. Assert current/peak ownership never includes the unrelated suffix during the continuation read.

Jaruso (commented):
Re-review of `b990fd6971c5c6c86673437eb7f90a75e3521032` complete. The previous parser-accounting, transfer double-counting, borrowed-suffix, internal-counter, and OpenSSL peak/lifecycle findings are substantially fixed, and the current CI run is green.

I still do not think this is ready to merge. Four remaining issues affect the core #355 acceptance criteria: strict high/low hysteresis under small records, bounded continuation of fragmented records, and truthful event-loop/application readiness under valid runtime policies.

The most important uncovered case is the default inbound plaintext policy. `hard = 32 KiB`, `high = 24 KiB`, and the atomic preflight reserves a full 16 KiB fragment. Repeated 1 KiB records can therefore leave 17 KiB queued, at which point processing stops because only 15 KiB of capacity remains. The high watermark has not been reached, the pause latch/counter remains clear, and reading resumes as soon as the application drains back to 16 KiB—well above the 8 KiB low watermark. That is not the advertised hysteresis and misses the required small-record/slow-consumer behavior.

There is also still an event-loop spin state: an active parser bypasses the high latch, while `canAcceptCarrierRead()` does not require positive runtime ciphertext room or runtime destination reserve. A blocked parser can therefore fill the ciphertext hard limit and continue advertising `wants_read = true` even though `drive()` computes `read_cap == 0` and cannot make progress.

Please address the four inline findings and add the specified deterministic regressions before merging. Since GitHub does not permit the PR author to request changes on their own PR, this is submitted as a blocking COMMENT review rather than `REQUEST_CHANGES`.

src/tls/encrypted_stream.zig
```diff
@@ -1025,39 +1415,45 @@ pub const PureZigRecordStream = struct {
...
     fn canAcceptCarrierRead(self: *const PureZigRecordStream) bool {
+        return self.canProcessCarrierInput() and
+            (!self.read_backpressured or self.hasActiveInboundParser()) and
+            self.inbound_carrier.available() > 0;
+    }
+
+    fn canProcessCarrierInput(self: *const PureZigRecordStream) bool {
         return self.lifecycle != .closed and self.lifecycle != .failed and self.lifecycle != .closing and !self.peer_closed and
-            self.inbound_carrier.available() > 0 and
             self.inbound_plaintext.available() >= record_codec.max_plaintext_fragment_len and
```
Jaruso: **Blocking: the full-fragment preflight causes an early, unlatched pause below `high`, then resumes above `low`.**

This is observable with the default plaintext policy:

- capacity/hard: 32 KiB;
- high: 24 KiB;
- low: 8 KiB;
- `max_plaintext_fragment_len`: 16 KiB.

After seventeen 1 KiB records, `inbound_plaintext.len == 17 KiB`. Only 15 KiB of compile-time room remains, so this predicate becomes false even though the queue is 7 KiB below `high`. `read_backpressured` is never latched and no pause counter is emitted. Draining only 1 KiB makes the predicate true again at 16 KiB, which is still twice the configured low watermark.

The runtime checks in `feedCiphertextInternal()` have the same issue for custom hard limits: they reserve a maximum fragment before the actual record size is known, so a small record that would fit can be rejected or internally blocked below `high`.

This violates the required small-record progress and high/low hysteresis contract. The reservation needs to be based on a safe upper bound for the **current encoded record**, not the global maximum. For example, once the TLS header is available, its payload length can provide a conservative per-record bound before cryptographic state advances:

```zig
const needed = current_record.payload_len; // conservative upper bound
if (!self.hasHardRoom(.inbound_plaintext, self.inbound_plaintext.len, needed) or
    !self.hasHardRoom(.handshake, self.inbound_handshake.len, needed))
{
    return error.WouldBlock;
}
```

If the design instead keeps maximum-fragment reservation, the policy/defaults must guarantee that reserve exhaustion cannot occur before `high`; the current handshake policy (`hard == one max fragment`) makes that strategy especially problematic.

Please add a deterministic slow-consumer test using many small records under the default policy. It must reach/cross `high`, latch exactly one pause, remain paused while above `low`, and resume exactly once only at/below `low`.

src/tls/encrypted_stream.zig
```diff
@@ -1025,39 +1415,45 @@ pub const PureZigRecordStream = struct {
     }
 
     fn canAcceptCarrierRead(self: *const PureZigRecordStream) bool {
+        return self.canProcessCarrierInput() and
+            (!self.read_backpressured or self.hasActiveInboundParser()) and
```
Jaruso: **Blocking: `wants_read` can remain true when the socket read has zero runtime room or cannot advance the active record.**

`canAcceptCarrierRead()` checks compile-time queue availability, but it does not require:

1. `inboundCiphertextReadRoom() > 0`; or
2. runtime hard-room in the possible destination plaintext/handshake queues.

A concrete failure state is:

- a parser is active, so the high-watermark latch is bypassed;
- the destination plaintext queue lacks runtime room for the record;
- carrier bytes accumulate until total inbound ciphertext reaches `hard`.

At that point `drive()` computes `read_cap == 0`, makes no progress, and breaks, but `readiness().wants_read` calls this predicate and still returns true. A level-triggered event loop can repeatedly wake and call `drive()` even though no socket read or record processing can succeed.

Please split carrier readiness from direct-feed rejection accounting and require actual progress capability, e.g.:

```zig
fn canReadCarrierSocket(self: *const PureZigRecordStream) bool {
    return self.lifecycleAllowsInput() and
        self.inboundCiphertextReadRoom() > 0 and
        self.hasRuntimeInboundDestinationReserve() and
        (!self.read_backpressured or self.canContinueActiveRecord()) and
        self.inbound_carrier.available() > 0;
}
```

Direct/manual feed APIs can retain their explicit counted rejection paths; they should not rely on a readiness predicate that would suppress the required hard-limit counter.

Add a test with a partial active record plus a runtime-hard-blocked destination. Repeated `drive()` calls must not consume additional carrier bytes, must return stable `made_progress = false`, and must advertise `wants_read = false` until the application drains the destination.

src/tls/encrypted_stream.zig
```diff
@@ -781,7 +1160,7 @@ pub const PureZigRecordStream = struct {
             .wants_read = !self.carrier_eof and self.canAcceptCarrierRead(),
             .wants_write = self.outbound_ciphertext.len > 0 or (self.lifecycle == .closing and !self.close_notify_queued),
             .can_read_plaintext = self.inbound_plaintext.len > 0 or pending_terminal_read_ready,
-            .can_write_plaintext = self.pending_terminal_read_error == null and self.lifecycle == .open and self.bridge.handshake_complete and self.outbound_ciphertext.available() >= record_codec.max_ciphertext_record_len,
+            .can_write_plaintext = self.pending_terminal_read_error == null and self.lifecycle == .open and self.bridge.handshake_complete and !self.write_backpressured and self.outbound_ciphertext.available() >= record_codec.max_ciphertext_record_len,
```
Jaruso: **Blocking: `can_write_plaintext` can advertise success when `writePlaintext()` is guaranteed to reject at the runtime hard limit.**

This readiness expression checks compile-time capacity and the high-watermark latch, but `writePlaintext()` additionally requires runtime room for `max_ciphertext_record_len`. A valid custom policy may have `high == hard`; with the queue below `high` but less than one maximum record of hard room remaining:

- `write_backpressured == false`;
- compile-time `available()` is sufficient;
- `can_write_plaintext == true`;
- the very next `write()` returns `WouldBlock` and increments the hard-limit counter.

That makes the backend-neutral HTTP seam unsafe: a consumer following readiness can repeatedly call an operation that cannot progress.

Make readiness use the same side-effect-free preflight as the operation:

```zig
.can_write_plaintext = self.pending_terminal_read_error == null and
    self.lifecycle == .open and
    self.bridge.handshake_complete and
    !self.write_backpressured and
    self.canReserveOutboundRecord(),
```

`canReserveOutboundRecord()` should be `*const` if necessary. Readiness polling must not increment counters.

Add a valid-policy regression with `high == hard` and `outbound_ciphertext.len > hard - max_ciphertext_record_len`. Assert `can_write_plaintext == false`, repeated readiness polls leave counters unchanged, and draining enough room makes it true again.

src/tls/encrypted_stream.zig
```diff
@@ -408,13 +579,196 @@ pub const PureZigRecordStream = struct {
...
+        };
+    }
+
+    fn inboundCiphertextReadRoom(self: *const PureZigRecordStream) Error!usize {
+        const owned = self.inboundCiphertextOwned();
+        const hard = self.buffer_limits.inbound_ciphertext.hard;
+        if (owned >= hard) return 0;
+        const hard_room = hard - owned;
+        const parser = self.activeInboundParser();
+        if (parser.len > 0) return @min(hard_room, try parser.pendingRecordBytesNeeded());
```
Jaruso: **Blocking: this is not yet an exact socket-byte allowance because it ignores bytes already sitting in `inbound_carrier`.**

A reproducible state is:

1. `parser.len == 4` (four bytes of the TLS header are parser-owned).
2. `inbound_carrier` already owns the fifth header byte.
3. That completed header declares a record whose destination queue lacks hard room, so `processCarrierInputBudget()` returns `WouldBlock` and leaves the fifth byte queued.
4. `pendingRecordBytesNeeded()` still returns `1`, because it only sees the parser.
5. `hasRuntimeInboundDestinationReserve()` passes `&.{}`, so it also fails to notice that the queued byte completes the header and incorrectly returns `true`.
6. `canAcceptCarrierRead()` therefore advertises/read-caps one additional socket byte. The parser remains unchanged because the destination is still blocked, so the loop can repeat and pull the body or a coalesced suffix toward `hard`.

The continuation calculation needs to consider **parser bytes plus the already-owned carrier prefix**, not parser bytes alone. A parser helper along these lines would make the contract explicit:

```zig
pub fn pendingRecordBytesNeededWith(
    self: *const Parser,
    queued: []const u8,
) Error!usize {
    if (self.len + queued.len < header_len)
        return header_len - (self.len + queued.len);

    const payload_len = try self.pendingRecordPayloadLenWith(queued) orelse unreachable;
    const record_len = header_len + payload_len;
    const already_owned = @min(record_len, self.len + queued.len);
    return record_len - already_owned;
}
```

Then both probes should use the same owned prefix:

```zig
const queued = self.inbound_carrier.slice();
const socket_need = try parser.pendingRecordBytesNeededWith(queued);
return @min(hard_room, socket_need);
```

```zig
return self.hasInboundDestinationReserveFor(
    self.activeInboundParser(),
    self.inbound_carrier.slice(),
);
```

Also make the direct `feedHandshakeCiphertext(epoch, ...)` continuation check use the parser selected by `epoch`; using only `activeInboundParser()` can still block the explicitly selected parser or authorize the wrong one.

Please add a controlled-carrier regression with a four-byte parser header, the completing byte already queued, a destination hard block, and more socket data available. Repeated `drive()`/readiness calls must perform **zero** carrier reads, report `wants_read = false`, leave counters stable, and resume only after the destination drains. Include a coalesced second record to prove no suffix is pulled through the continuation exception.

src/tls/encrypted_stream.zig
```diff
@@ -663,50 +1017,95 @@ pub const PureZigRecordStream = struct {
...
+        return self.feedCiphertextInternal(bytes, true);
+    }
+
+    fn feedCiphertextInternal(self: *PureZigRecordStream, bytes: []const u8, comptime direct_external_feed: bool) Error!usize {
         if (self.failed) |err| return err;
         if (self.lifecycle == .closed or self.lifecycle == .failed) return error.StreamClosed;
         if (self.peer_closed) return bytes.len;
-        if (!self.canAcceptCarrierRead()) return error.WouldBlock;
+        if (direct_external_feed) {
+            if (!self.canAcceptCarrierRead()) return error.WouldBlock;
```
Jaruso: **Blocking: the direct API still enters through the socket-readiness predicate, which hides explicit hard-limit rejection accounting.**

For example, when inbound ciphertext ownership is already at `hard`, `canAcceptCarrierRead()` returns `false` because read room is zero. This line returns an ordinary `WouldBlock` before line 1082 can call `rejectHardLimit(.inbound_ciphertext, ...)`. The same masking occurs when the active record's declared destination size is known and the destination queue lacks hard room: the side-effect-free runtime reserve probe returns `false`, so the later queue-specific counted rejection is unreachable.

Please split the two concepts rather than reusing socket readiness for a caller action. One workable shape is a side-effect-free probe that preserves the reason:

```zig
const CapacityProbe = union(enum) {
    ready,
    high_paused,
    hard: BufferQueue,
};
```

The socket path maps anything except `.ready` to `wants_read = false`; the direct/manual path maps `.high_paused` to an uncounted `WouldBlock` and `.hard` to exactly one `rejectHardLimit(queue, error.WouldBlock)`.

There is a second accounting hole in the subsequent preflights: direct paths often return on `queue.available() < reserve` **before** checking `hasHardRoom()`. Since policy guarantees `hard <= capacity`, insufficient compile-time capacity while `current <= hard` also implies insufficient runtime hard room. Check/report the hard limit first, then use `available()` only as an invariant/capacity guard. The same ordering issue exists in `feedHandshakeCiphertext()`, `writePlaintext()`, and manual `applyEvent(.handshake_bytes)`.

Required regressions:

```zig
// Direct inbound hard rejection is counted once.
state.inbound_carrier.len = limits.inbound_ciphertext.hard;
try expectError(error.WouldBlock, state.feedCiphertext(record));
try expectEqual(@as(u64, 1), snapshot().counters.hard_limits.inbound_ciphertext);

// Direct destination-hard rejection is counted, while internal retries/readiness polls are not.
// Also cover hard == compile-time capacity so `available()` and hard room fail together.

// Direct plaintext write at outbound hard is counted once; repeated readiness probes remain zero-cost.
```

All rejected operations must leave parser/sequence/queue state unchanged.

Jaruso (commented):
Re-review of `b5e08a643e19ae13a09e2b203eb6cad20f32c20a` against #355 and the previous four findings.

This commit substantially improves the implementation: `can_write_plaintext` now agrees with the write preflight, destination reservations are record-sized rather than always max-fragment-sized, the active parser is selected by the current read epoch, and the new coalesced-record test covers the simple high-watermark continuation case. CI run 736 is also fully green.

I still do not think this is safe to merge because two blocking edge cases remain:

1. **Continuation/readiness does not account for bytes already owned in `inbound_carrier`.** The socket allowance is computed from parser state alone, and the runtime destination probe passes an empty slice. If the parser has four header bytes and `inbound_carrier` already contains the fifth byte, a destination-capacity block leaves that byte queued. Readiness nevertheless says one more socket byte is useful, so each loop can read another byte while the parser remains unchanged. This can pull the current body and unrelated coalesced records all the way to the hard limit despite the intended exact-record exception.

2. **Explicit direct-API hard-limit rejections can still bypass their counters.** `feedCiphertext()` calls the side-effect-free socket-readiness predicate before its counted hard checks, so a direct call at inbound hard or with a known destination-hard failure returns an uncounted `WouldBlock`. Several direct paths also check compile-time `available()` before runtime `hasHardRoom()`; because `hard <= capacity`, an insufficient-capacity result is also necessarily a hard-limit failure when the queue is within policy, but it is returned without incrementing the queue counter.

These are both deterministic state-machine issues, not theoretical resource races. Detailed reproductions and expected fixes are attached inline. As the PR author, GitHub will not allow me to submit a formal `REQUEST_CHANGES`, so this blocking review is submitted as `COMMENT`.

Jaruso (commented):
Re-reviewed current head `353a94a849f73d86148390ceb850fb97a6cb71f7` against issue #355, the maintainer implementation plan, and all prior blocking findings.

The final two blockers are addressed:

- Fragment continuation/readiness now accounts for bytes already queued in `inbound_carrier` via `pendingRecordBytesNeededWith(...)`, and destination-reserve readiness evaluates the queued header prefix. This prevents unnecessary socket reads, coalesced-suffix over-read, and false `wants_read` while the destination is blocked.
- Direct/manual feed operations no longer depend on the side-effect-free socket-readiness predicate for rejection accounting. Runtime hard-room checks now occur before ordinary capacity checks on the explicit API paths, so readiness polling remains counter-free while real hard-limit rejections are counted exactly once.

The added regressions cover the queued-header/destination-blocked state and direct hard rejection independently of socket readiness. I also rechecked the earlier accounting, hysteresis, atomicity, parser ownership, OpenSSL boundary, cleanup, and HTTP seam fixes for regressions.

CI run 737 completed successfully on this head, and the PR is currently mergeable. I found no remaining blocking correctness or acceptance-criteria gaps. **Looks good to merge.**

PR #458, opened 2026-07-18
tls: [323-F] handshake-time client authentication and key selection (#334)

Jaruso:
## Summary

Completes issue #334 by integrating the green credential-contract PR with the
handshake-time client-authentication WIP and finishing it, in four strict
checkpoints (each committed and green before the next):

- **Checkpoint 1** — restore the merged branch: formatting fix; the tree
  already compiled and passed `test-tls`/`test-quic`/`test`/appliance.
- **Checkpoint 2** — synchronous handshake-time client authentication:
  disabled/optional/required modes; CertificateRequest parse/emit; the client
  Certificate / optional CertificateVerify / Finished flight; role-aware
  Certificate and CertificateVerify processing; the client CertificateVerify
  context string; Finished MACs recomputed over the correct transcript. New
  `certificate_required` alert path.
- **Checkpoint 3** — asynchronous client authentication: resumable client
  credential selection and signing, server-side client-certificate verification
  through the existing async machinery, receive gating past Finished while
  pending, automatic drain of buffered handshake messages after resume, and
  exactly-once cancel/release of every operation and credential.
- **Checkpoint 4** — remaining correctness fixes: reject oversized configured
  server names; validate exact client-flight sizes before transcript mutation
  or output; map a failed CertificateVerify signature to `decrypt_error`
  (RFC 8446 §4.4.3); expose pending/resume through the generic transport
  Backend, the engine Driver, and both the record and QUIC production drivers.

No OpenSSL/X.509 objects cross the credential interface; failures map to
deterministic TLS alerts. Post-handshake client authentication remains
explicitly deferred.

## Test plan
- [x] `zig fmt --check src build.zig build.zig.zon`
- [x] `zig build test-tls`
- [x] `zig build test-quic`
- [x] `zig build test`
- [x] `zig build test -Dtls-profile=appliance`
- [x] New sync success/failure, async selection/signing/verification,
  cancellation, malformed-completion, coalesced-flight, byte-by-byte
  fragmentation, oversized-server-name, decrypt_error, and production-driver
  (record stream + generic driver) tests.

Closes #334

### DIFF
--- a/src/quic/connection.zig
+++ b/src/quic/connection.zig
@@ -1131,6 +1131,34 @@ pub const Connection = struct {
         }
     }
 
+    /// True while the TLS handshake has suspended awaiting an asynchronous
+    /// authentication operation (#334). The event loop can poll this to know
+    /// the connection is waiting on an external signer/verifier/selector; each
+    /// `pollTransmit` (or an explicit `driveAuthentication`) advances it.
+    pub fn authPending(self: *const Connection) bool {
+        return self.handshake.authPending();
+    }
+
+    /// Drive a parked asynchronous authentication operation forward without
+    /// requiring another inbound CRYPTO frame: poll it, queue any handshake
+    /// output it produced, and complete the handshake if it resolved. Safe to
+    /// call unconditionally; a no-op unless the backend is suspended.
+    pub fn driveAuthentication(self: *Connection, now_us: u64) void {
+        if (self.state_ == .closed or self.state_ == .draining or self.state_ == .closing) return;
+        if (!self.handshake.authPending()) return;
+        self.handshake.resumeAuth() catch |err| {
+            self.failHandshake(err);
+            self.startClose(.{ .error_code = cryptoErrorCode(err), .is_application = false, .local = true }, @errorName(err), now_us);
+            return;
+        };
+        self.collectCryptoOutput() catch {
+            self.failHandshake(error.HandshakeBufferOverflow);
+            self.startClose(.{ .error_code = error_crypto_buffer_exceeded, .is_application = false, .local = true }, "crypto buffer", now_us);
+            return;
+        };
+        self.afterHandshakeProgress(now_us);
+    }
+
     fn afterHandshakeProgress(self: *Connection, now_us: u64) void {
         if (self.handshake_complete or !self.handshake.isComplete()) return;
         self.handshake_complete = true;
@@ -1536,6 +1564,19 @@ pub const Connection = struct {
             self.close_needs_send = false;
             self.close_resend_allowed_at_us = now_us + self.ptoDurationNow() / 2;
             return self.buildCloseDatagram(out, now_us);
+        }
+
+        // Advance any parked asynchronous authentication (#334) before building
+        // packets, so a resolved external signer/verifier/selector emits its
+        // next flight here even when no further peer CRYPTO frame will arrive.
+        self.driveAuthentication(now_us);
+        if (self.state_ == .closing or self.state_ == .closed or self.state_ == .draining) {
+            if (self.state_ == .closing and self.close_needs_send) {
+                self.close_needs_send = false;
+                self.close_resend_allowed_at_us = now_us + self.ptoDurationNow() / 2;
+                return self.buildCloseDatagram(out, now_us);
+            }
+            return null;
         }
 
         // Force the delayed app-space ACK when its timer expired.
--- a/src/quic/connection.zig
+++ b/src/quic/connection.zig
@@ -1228,6 +1228,7 @@ pub const Connection = struct {
         return switch (err) {
             error.AlpnMismatch => error_crypto_base + 120, // no_application_protocol
             error.CertificateInvalid => error_crypto_base + 42, // bad_certificate
+            error.ClientCertificateRequired => error_crypto_base + 116, // certificate_required
             error.MissingTransportParameters, error.InvalidTransportParameters => error_transport_parameter,
             error.UnexpectedHandshakeMessage => error_crypto_base + 10, // unexpected_message
             error.IllegalParameter => error_crypto_base + 47, // illegal_parameter
--- a/src/quic/connection.zig
+++ b/src/quic/connection.zig
@@ -776,6 +776,15 @@ pub const Connection = struct {
                     self.startClose(.{ .error_code = cryptoErrorCode(err), .is_application = false, .local = true }, @errorName(err), now_us);
                     return;
                 };
+                // Poll any parked asynchronous authentication (#334) so a
+                // resolved external signer/verifier/selector progresses the
+                // handshake as this packet is processed. A no-op when nothing is
+                // suspended.
+                self.handshake.resumeAuth() catch |err| {
+                    self.failHandshake(err);
+                    self.startClose(.{ .error_code = cryptoErrorCode(err), .is_application = false, .local = true }, @errorName(err), now_us);
+                    return;
+                };
                 self.collectCryptoOutput() catch {
                     self.failHandshake(error.HandshakeBufferOverflow);
                     self.startClose(.{ .error_code = error_crypto_buffer_exceeded, .is_application = false, .local = true }, "crypto buffer", now_us);
@@ -1229,6 +1238,7 @@ pub const Connection = struct {
             error.AlpnMismatch => error_crypto_base + 120, // no_application_protocol
             error.CertificateInvalid => error_crypto_base + 42, // bad_certificate
             error.ClientCertificateRequired => error_crypto_base + 116, // certificate_required
+            error.DecryptError => error_crypto_base + 51, // decrypt_error
             error.MissingTransportParameters, error.InvalidTransportParameters => error_transport_parameter,
             error.UnexpectedHandshakeMessage => error_crypto_base + 10, // unexpected_message
             error.IllegalParameter => error_crypto_base + 47, // illegal_parameter
--- a/src/tls/encrypted_stream.zig
+++ b/src/tls/encrypted_stream.zig
@@ -489,6 +489,21 @@ pub const PureZigRecordStream = struct {
         try self.applyDriverOutcome(outcome);
     }
 
+    /// Poll a parked asynchronous authentication operation (#334) once and apply
+    /// whatever the backend emits — the client's certificate flight, the
+    /// server's, or a peer-verification verdict. A no-op unless the backend has
+    /// suspended; requires output headroom so an emitted flight serializes
+    /// atomically. Returns true if it polled (progress toward resolution).
+    fn resumeAuthIfPending(self: *PureZigRecordStream) Error!bool {
+        if (!self.driverPresent()) return false;
+        const driver = &self.handshake_driver.?;
+        if (!driver.authPending()) return false;
+        if (self.outbound_ciphertext.available() < handshake_output_reserve) return false;
+        const outcome = driver.resumeAuthOutcome();
+        try self.applyDriverOutcome(outcome);
+        return true;
+    }
+
     /// Apply one borrowed driver event batch. Every payload slice is consumed,
     /// copied, or serialized here, before any subsequent driver call resets the
     /// sink. A terminal backend error is deferred (with its fatal alert queued)
@@ -826,6 +841,11 @@ pub const PureZigRecordStream = struct {
         // initial flight; server: arms the responder). The carrier write loop
         // below flushes whatever it queued.
         if (try self.startHandshakeIfNeeded()) made_progress = true;
+
+        // Poll a parked asynchronous authentication operation (#334): an
+        // external signer/verifier/selector may have suspended the handshake,
+        // and each drive tick advances it toward resolution.
+        if (try self.resumeAuthIfPending()) made_progress = true;
 
         if (self.carrier) |carrier| {
             var written_total: usize = 0;
--- a/src/tls/encrypted_stream.zig
+++ b/src/tls/encrypted_stream.zig
@@ -671,6 +671,8 @@ pub const PureZigRecordStream = struct {
             error.InvalidHandshakeState,
             error.NoApplicableCredential,
             error.CredentialProviderFailed,
+            error.ClientCertificateRequired,
+            error.DecryptError,
             => alerts.fromHandshakeError(@errorCast(err)),
             else => null,
         };
--- a/src/tls/encrypted_stream.zig
+++ b/src/tls/encrypted_stream.zig
@@ -654,6 +654,8 @@ pub const PureZigRecordStream = struct {
             error.AlpnMismatch,
             error.SecretExportFailed,
             error.InvalidHandshakeState,
+            error.NoApplicableCredential,
+            error.CredentialProviderFailed,
             => alerts.fromHandshakeError(@errorCast(err)),
             else => null,
         };
--- a/src/tls/handshake.zig
+++ b/src/tls/handshake.zig
@@ -133,7 +133,13 @@ pub const Core = struct {
         // the client is still expecting the server's Certificate; the client
         // remembers it and will authenticate after the server Finished.
         if (message.kind == .certificate_request) {
-            if (self.role != .client or self.expected_inbound != .certificate)
+            // At most one CertificateRequest may appear in this position of the
+            // server flight (RFC 8446 §4.3.2). `expected_inbound` stays
+            // `.certificate` after the first, so the remembered flag is what
+            // rejects a duplicate rather than accepting and re-hashing it.
+            if (self.role != .client or
+                self.expected_inbound != .certificate or
+                self.client_certificate_requested)
                 return error.UnexpectedHandshakeMessage;
             self.client_certificate_requested = true;
             self.transcript.update(message.raw);
@@ -332,6 +338,25 @@ test "core records both directions of a client and server flight" {
     const client_hash = client.transcriptHash();
     const server_hash = server.transcriptHash();
     try std.testing.expectEqualSlices(u8, &client_hash, &server_hash);
+}
+
+test "a duplicate CertificateRequest in the server flight is rejected" {
+    var client = Core.init(.client);
+    try client.start();
+    var bytes: [8]u8 = undefined;
+    const ch = try messages.encode(.client_hello, "", &bytes);
+    try client.recordSent(ch);
+    const sh = try messages.encode(.server_hello, "", &bytes);
+    _ = try client.acceptReceived(sh);
+    const ee = try messages.encode(.encrypted_extensions, "", &bytes);
+    _ = try client.acceptReceived(ee);
+
+    // The first CertificateRequest is accepted (client auth requested).
+    const cr = try messages.encode(.certificate_request, "", &bytes);
+    _ = try client.acceptReceived(cr);
+    try std.testing.expect(client.client_certificate_requested);
+    // A second one in the same position is illegal (RFC 8446 §4.3.2).
+    try std.testing.expectError(error.UnexpectedHandshakeMessage, client.acceptReceived(cr));
 }
 
 test "secret lifecycle tracks directions and rejects repeated discard" {
--- a/src/tls/handshake.zig
+++ b/src/tls/handshake.zig
@@ -64,9 +64,49 @@ pub const Core = struct {
     expected_inbound: ?MessageType = null,
     transcript: transcript_mod.Transcript = .{},
     secrets: SecretLifecycle = .{},
+    /// Handshake-time client authentication (#334). The server sets
+    /// `request_client_certificate` before its flight so it may emit
+    /// CertificateRequest and then expect the client's certificate flight after
+    /// its own Finished. The client sets `client_certificate_requested` when it
+    /// receives that CertificateRequest, so it sends its own certificate flight
+    /// before Finished. Both default off — the common no-client-auth path and
+    /// QUIC are unchanged. Post-handshake client authentication is deferred.
+    request_client_certificate: bool = false,
+    client_certificate_requested: bool = false,
+    /// Server inbound sub-state for the client's post-Finished certificate
+    /// flight; `inactive` unless client auth was requested.
+    client_auth_inbound: ClientAuthInbound = .inactive,
+    /// Client outbound sub-state for its own certificate flight.
+    client_auth_outbound: ClientAuthOutbound = .inactive,
+
+    pub const ClientAuthInbound = enum { inactive, expect_certificate, expect_certificate_verify, expect_finished };
+    pub const ClientAuthOutbound = enum { inactive, send_certificate, send_certificate_verify, send_finished };
 
     pub fn init(role: state.Role) Core {
         return .{ .role = role };
+    }
+
+    /// Server: request that the client authenticate. Call before the server
+    /// flight so CertificateRequest is emitted and the client certificate
+    /// flight is expected after the server Finished.
+    pub fn requestClientCertificate(self: *Core) void {
+        self.request_client_certificate = true;
+    }
+
+    /// Server: after parsing the client's Certificate, declare whether it was
+    /// empty (no client cert). An empty certificate skips CertificateVerify, so
+    /// the server expects the client Finished next; a non-empty one expects
+    /// CertificateVerify.
+    pub fn clientCertificateWasEmpty(self: *Core, empty: bool) void {
+        self.client_auth_inbound = if (empty) .expect_finished else .expect_certificate_verify;
+    }
+
+    /// Client: begin the local certificate flight after the server Finished.
+    /// The client always sends a Certificate first (possibly empty); whether a
+    /// CertificateVerify follows depends on whether that certificate was empty,
+    /// which the outbound sequence permits either way.
+    pub fn beginClientCertificateFlight(self: *Core) void {
+        self.client_auth_outbound = .send_certificate;
     }
 
     pub fn start(self: *Core) Error!void {
@@ -88,6 +128,24 @@ pub const Core = struct {
             self.transcript.update(message.raw);
             return message;
         }
+        // A CertificateRequest (server->client, #334) is an optional message the
+        // server inserts before its Certificate. Accept it transparently while
+        // the client is still expecting the server's Certificate; the client
+        // remembers it and will authenticate after the server Finished.
+        if (message.kind == .certificate_request) {
+            if (self.role != .client or self.expected_inbound != .certificate)
+                return error.UnexpectedHandshakeMessage;
+            self.client_certificate_requested = true;
+            self.transcript.update(message.raw);
+            return message;
+        }
+        // Server inbound for the client's post-Finished certificate flight.
+        if (self.client_auth_inbound != .inactive) {
+            try self.checkClientAuthInbound(message.kind);
+            self.transcript.update(message.raw);
+            self.advanceClientAuthInbound(message.kind);
+            return message;
+        }
         if (!self.isExpectedClientFinished(message.kind)) {
             if (self.expected_inbound != message.kind)
                 return error.UnexpectedHandshakeMessage;
@@ -95,6 +153,32 @@ pub const Core = struct {
         self.transcript.update(message.raw);
         self.advanceAfterReceive(message.kind);
         return message;
+    }
+
+    fn checkClientAuthInbound(self: *const Core, kind: MessageType) Error!void {
+        const ok = switch (self.client_auth_inbound) {
+            .inactive => false,
+            .expect_certificate => kind == .certificate,
+            // Set by the backend once it has parsed the client's Certificate:
+            // a non-empty cert requires CertificateVerify, an empty one skips it.
+            .expect_certificate_verify => kind == .certificate_verify,
+            .expect_finished => kind == .finished,
+        };
+        if (!ok) return error.UnexpectedHandshakeMessage;
+    }
+
+    fn advanceClientAuthInbound(self: *Core, kind: MessageType) void {
+        switch (kind) {
+            // After the Certificate the backend refines the next expectation via
+            // `clientCertificateWasEmpty`; default to expecting CertificateVerify.
+            .certificate => self.client_auth_inbound = .expect_certificate_verify,
+            .certificate_verify => self.client_auth_inbound = .expect_finished,
+            .finished => {
+                self.client_auth_inbound = .inactive;
+                self.handshake_lifecycle = .complete;
+            },
+            else => {},
+        }
     }
 
     pub fn recordSent(self: *Core, raw: []const u8) Error!void {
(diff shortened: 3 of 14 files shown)


src/tls/tls13_backend.zig
```diff
@@ -468,13 +554,30 @@ pub const Tls13Backend = struct {
             },
         };
         input.append(bytes) catch |err| return mapCoreError(err);
+        try self.drainInput(input, level, sink);
```
Jaruso, line 571: **[P1] Do not process a new handshake message while authentication is already pending.**

`drainInput` stops *after* the message that parks an operation, but this entry point calls it unconditionally on every later `receive`. A concrete bypass is:

1. The server accepts the client's `CertificateVerify`; `core.acceptReceived` advances to expect `Finished`.
2. The verifier returns `pending` and this receive stops.
3. The peer's `Finished` arrives in a separate transport read before the verifier resolves.
4. This method appends it and immediately calls `drainInput`, which processes `Finished`, emits `handshake_complete`, and wipes the schedule even though the certificate verdict is still unresolved.

The existing coalesced-flight test does not catch this because the post-message check stops the same `drainInput` invocation. Buffer new bytes while pending, but return before dispatching any message, and also make `drainInput` defensively refuse to enter its loop while `pending_op != null`.

For example:
```zig
input.append(bytes) catch |err| return mapCoreError(err);
if (self.pending_op != null) return;
try self.drainInput(input, level, sink);
```

Add a regression where `CertificateVerify` parks in one `receive`, `Finished` is supplied by a second `receive`, and no application secrets/completion occur until an accepted resume. Rejection must never process that buffered `Finished`.

src/quic/connection.zig
```diff
@@ -776,6 +776,15 @@ pub const Connection = struct {
                     self.startClose(.{ .error_code = cryptoErrorCode(err), .is_application = false, .local = true }, @errorName(err), now_us);
                     return;
                 };
+                // Poll any parked asynchronous authentication (#334) so a
+                // resolved external signer/verifier/selector progresses the
+                // handshake as this packet is processed. A no-op when nothing is
+                // suspended.
+                self.handshake.resumeAuth() catch |err| {
```
Jaruso, line 783: **[P1] QUIC needs an authentication wake/progress path that does not depend on another inbound CRYPTO frame.**

This is currently the only connection-level call to `resumeAuth`. A genuinely asynchronous server selector/signer parks after ClientHello, then the client waits for the rest of the server flight; no further CRYPTO frame exists to trigger this code. The same deadlock occurs when a client verifier needs multiple polls before the client can emit its Finished. ACKs and timer callbacks do not call `resumeAuth`, and `pollTransmit` only sends already-collected output.

Expose a connection-level method for the provider's wakeup, or integrate polling into the normal connection progress path (`pollTransmit`/an explicit `driveAuthentication(now_us)`), and on completion always run the same sequence used here:

```zig
try self.handshake.resumeAuth();
try self.collectCryptoOutput();
self.afterHandshakeProgress(now_us);
```

The event loop also needs a way to know authentication is pending or has been woken. Add a production QUIC test with an async selection/sign/verify requiring more than the immediate post-CRYPTO poll and no extra peer CRYPTO input; it must still emit the next flight and complete.

src/tls/tls13_backend.zig
```diff
@@ -753,49 +990,93 @@ pub const Tls13Backend = struct {
...
     }
 
-    fn verifyServerCertificate(self: *Tls13Backend, algorithm: u16, signature: []const u8, content: []const u8) CertificateState {
-        const leaf = self.peer_certificate[0..self.peer_certificate_len];
+    /// Apply a peer-verification verdict: emit the certificate state and fail
+    /// closed on rejection. Reachable synchronously and after resume.
+    fn applyPeerVerdict(self: *Tls13Backend, verdict: credentials.Verdict, sink: *EventSink) HandshakeError!void {
+        const state: CertificateState = switch (verdict) {
+            .accepted => .valid,
+            .not_checked => .not_checked,
```
Jaruso: **[P1] Do not allow `.not_checked` to satisfy required—or ordinarily optional—client authentication.**

This method accepts `.not_checked` unconditionally. Client-side server authentication is later protected by transport policy, but both record and QUIC completion checks only enforce that rule when the local role is `.client`; a server requesting a client certificate can therefore complete `required` authentication with no trust decision.

There is a related policy error in `requestClientAuthentication`: `.optional` sets `allow_unverified_peer = true`. “Optional” means the client may send an empty Certificate. It does not mean a non-empty Certificate may bypass verification.

For both `.optional` and `.required`, a presented chain should normally require an `.accepted` verdict; the modes should differ only in how an empty Certificate is handled. Reject `.not_checked` whenever the applicable policy requires verification, including server role. If insecure acceptance is intentionally supported, make it a separate explicit opt-in rather than deriving it from `.optional`.

Add tests for both optional and required modes with a presented certificate and a verifier returning `.not_checked`; neither should silently establish a mutually authenticated connection under the normal policy.

src/tls/tls13_backend.zig
```diff
@@ -807,31 +1088,207 @@ pub const Tls13Backend = struct {
...
+    /// provider selects against the schemes the server advertised in its
+    /// CertificateRequest; selection may complete synchronously or suspend
+    /// (`client_select`), in which case the flight resumes from `resumeAuth`.
+    fn beginClientAuthFlight(self: *Tls13Backend, sink: *EventSink) HandshakeError!void {
+        self.core.beginClientCertificateFlight();
+        const provider = self.external_provider orelse
+            return self.emitClientCertificate(null, sink);
+        var selection = self.clientSelectionContext();
+        switch (provider.selectCredential(&selection) catch |err|
+            return self.failCredential(credentials.classifySelectError(err))) {
```
Jaruso: **[P1] A client with no suitable credential must send an empty Certificate, not abort the handshake locally.**

TLS 1.3 requires the client to respond to a handshake-time CertificateRequest with an empty Certificate when it has no appropriate certificate. This is how optional authentication succeeds and how required authentication produces the peer-attributed `certificate_required` outcome.

Currently only a completely absent provider takes that path. A configured selector returning `NoCredentialAvailable` or `NoCompatibleSignatureAlgorithm` is converted into local `NoApplicableCredential`/`handshake_failure` instead.

Handle those two deterministic selection outcomes as a decline:
```zig
const progress = provider.selectCredential(&selection) catch |err| switch (err) {
    error.NoCredentialAvailable,
    error.NoCompatibleSignatureAlgorithm,
    => return self.emitClientCertificate(null, sink),
    else => return self.failCredential(credentials.classifySelectError(err)),
};
```

The pending API must also be able to complete with the same typed “no credential” result. `Completion` currently has only `credential`, `signature_len`, and `verdict`, while every pending error becomes a provider-internal failure, so an asynchronous selector cannot faithfully report this normal outcome. Prefer a stage-specific result such as `?SelectedCredential` or a typed selection completion. Test synchronous and asynchronous no-match behavior in optional and required modes.

src/tls/tls13_backend.zig
```diff
@@ -983,9 +1481,69 @@ pub const Tls13Backend = struct {
...
+            return self.failCredential(.malformed_credential_chain);
+        var encoded_len: usize = 0;
+        for (chain.entries) |entry| {
+            if (entry.len == 0 or entry.len > max_certificate_len)
+                return self.failCredential(.malformed_credential_chain);
+            // CertificateEntry framing overhead: 3-byte cert_data length +
+            // 2-byte extensions length.
+            encoded_len = std.math.add(usize, encoded_len, entry.len + 5) catch
+                return self.failCredential(.malformed_credential_chain);
+            if (encoded_len > max_message_len)
```
Jaruso: **[P2] Preflight the exact serialized flight, not only CertificateEntry bytes.**

`encoded_len` includes each DER entry and its five-byte entry framing, but the same `max_message_len` writer subsequently contains EncryptedExtensions, optional CertificateRequest, and the Certificate message's own header/context/list framing. A chain can therefore pass this check and then produce `HandshakeBufferOverflow` after `requestClientCertificate()` and/or transcript/output mutation.

For example, four entries of 2043 bytes produce exactly 8192 bytes under this calculation, yet the Certificate alone needs another eight framing bytes, before EncryptedExtensions or CertificateRequest are counted.

Compute all checked sizes before mutating `core`, calling `recordSent`, emitting output, or invoking the signer:
- full EncryptedExtensions encoding, including optional transport extension;
- optional CertificateRequest;
- Certificate handshake header, request context, list length, and every entry.

Map any provider-caused overflow to `.malformed_credential_chain` and release without signing. The client-side `emitClientCertificate` has the same issue: its preflight omits the Certificate message's eight framing bytes. Add exact boundary tests for both paths with a custom provider whose entries approach the 8 KiB limit.

src/tls/encrypted_stream.zig
```diff
@@ -654,6 +669,8 @@ pub const PureZigRecordStream = struct {
             error.AlpnMismatch,
             error.SecretExportFailed,
             error.InvalidHandshakeState,
+            error.NoApplicableCredential,
+            error.CredentialProviderFailed,
```
Jaruso, line 673: **[P1] Record mode drops the new mandatory fatal alerts.**

`alerts.fromHandshakeError` now maps `ClientCertificateRequired` to `certificate_required` and `DecryptError` to `decrypt_error`, but neither error is included here. The backend does not emit its own `fatal_alert`, so `deferHandshakeFailure` receives no description and queues no TLS alert. The local side eventually fails, while the peer sees a stall/EOF instead of the protocol-required alert.

Include both errors in this switch:
```zig
error.ClientCertificateRequired,
error.DecryptError,
=> alerts.fromHandshakeError(@errorCast(err)),
```

Add record-stream end-to-end tests that keep driving both peers and assert the remote side receives a fatal alert for (a) required client auth with an empty Certificate and (b) a bad client or server CertificateVerify signature.

src/tls/handshake.zig
```diff
@@ -88,6 +128,24 @@ pub const Core = struct {
...
             return message;
         }
+        // A CertificateRequest (server->client, #334) is an optional message the
+        // server inserts before its Certificate. Accept it transparently while
+        // the client is still expecting the server's Certificate; the client
+        // remembers it and will authenticate after the server Finished.
+        if (message.kind == .certificate_request) {
+            if (self.role != .client or self.expected_inbound != .certificate)
+                return error.UnexpectedHandshakeMessage;
+            self.client_certificate_requested = true;
```
Jaruso, line 144: **[P2] Reject a duplicate main-handshake CertificateRequest.**

After the first request, `expected_inbound` is intentionally still `.certificate`, so this condition also accepts a second (or later) CertificateRequest, hashes it, and replaces the backend's captured signature-scheme list. The TLS 1.3 server flight permits at most one CertificateRequest in this position.

Include the remembered flag in the guard:
```zig
if (self.role != .client or
    self.expected_inbound != .certificate or
    self.client_certificate_requested)
```

Add a test with two CertificateRequest messages before Certificate and assert the second is rejected without advancing to the client-auth flight.

src/tls/tls13_backend.zig
```diff
@@ -1076,14 +1673,119 @@ pub const Tls13Backend = struct {
...
+    /// exactly where the synchronous path would have continued, recording no
+    /// handshake message twice.
+    fn dispatchResume(self: *Tls13Backend, stage: PendingStage, completion: credentials.Completion, sink: *EventSink) HandshakeError!void {
+        switch (stage) {
+            .server_select => {
+                if (completion != .credential) return self.failCredential(.provider_internal_failure);
+                return self.emitServerAuthFlight(completion.credential, sink);
+            },
+            .server_sign => {
+                if (completion != .signature_len) return self.failCredential(.invalid_callback_behavior);
```
Jaruso: **[P2] Dispose of all owned resources on a wrong-kind async completion.**

The operation is released before dispatch, but tag validation can fail while the completion itself owns a `SelectedCredential`. For example, a malicious pending signer or verifier can return `.credential`; this branch returns an error without releasing that unexpected credential. In the sign stages, `pending_credential` is also left retained until eventual backend teardown because it is cleared only after the tag check.

Invalid callback behavior still has to satisfy the engine's ownership guarantees immediately. Either make completions stage-specific so impossible owned variants cannot be returned, or add a cleanup path that:
- releases `completion.credential` whenever that variant is unexpected;
- takes and releases `pending_credential` immediately on a sign-stage mismatch;
- preserves exactly-once behavior if both aliases refer to the same provider handle according to the documented contract.

Extend the malformed-completion tests: cover a sign stage and verifier stage returning an unexpected credential, and assert every returned/held handle is released exactly once before `resumeAuth` returns.

Jaruso (commented):
## Review result: blocking changes remain

This is a substantial improvement over PR #455: the provider-neutral contracts, role-aware CertificateVerify handling, handshake-time client-auth state machine, typed alert additions, generic driver resume surface, and the test breadth are all moving in the right direction. CI is fully green across formatting, Linux/macOS/ARM tests, packaging, security, and the appliance profile.

I do not think this is safe to merge yet. The remaining findings are concentrated in adversarial async progression, client-auth policy, exact serialization bounds, and terminal alert delivery:

1. A later `receive` call can process `Finished` while peer verification is still pending, bypassing the intended authentication suspension.
2. QUIC polls pending authentication only while handling an inbound CRYPTO frame, so genuinely asynchronous operations can deadlock once the peer is waiting for our next flight.
3. Required client authentication can accept a `.not_checked` verdict, while optional mode incorrectly treats “certificate may be absent” as “a presented certificate may remain unverified.”
4. A configured client provider that has no suitable credential aborts locally instead of sending the empty Certificate required by TLS 1.3; the async completion model cannot represent that deterministic outcome either.
5. Provider-chain preflight still omits message and surrounding-flight framing, so a chain can pass validation and overflow after transcript/state mutation.
6. Record mode does not synthesize the newly introduced `certificate_required` and `decrypt_error` alerts.
7. Duplicate main-handshake CertificateRequest messages are accepted.
8. Wrong-kind async completions can leak an unexpected returned credential and retain a signing credential until teardown.

I left detailed inline comments with concrete failure sequences and expected fixes. Please add negative tests for every item, especially a `CertificateVerify` and `Finished` delivered in separate receives while verification remains pending, and a QUIC async signer that requires more than the immediate post-CRYPTO poll. The current green suite does not exercise those cases.

src/tls/encrypted_stream.zig
```diff
@@ -825,6 +844,11 @@ pub const PureZigRecordStream = struct {
         // below flushes whatever it queued.
         if (try self.startHandshakeIfNeeded()) made_progress = true;
 
+        // Poll a parked asynchronous authentication operation (#334): an
+        // external signer/verifier/selector may have suspended the handshake,
+        // and each drive tick advances it toward resolution.
+        if (try self.resumeAuthIfPending()) made_progress = true;
```
Jaruso, line 850: **[P1] Do not consume the next TLS record while authentication is still pending.**

`resumeAuthIfPending()` returns `true` whenever it polls, including when `poll` reports “still pending.” `drive()` then continues into `processCarrierInputBudget` in the same tick. That breaks both halves of mutual authentication:

* while the client is waiting on an asynchronous credential selection/signature, the server is already allowed to send application records after its Finished;
* while the server is waiting on asynchronous client-certificate verification, the client's application record can follow its Finished.

In both cases this stream still has `read_epoch == .handshake`, so the next application record is opened with handshake keys and the connection fails authentication instead of applying backpressure until the callback resolves.

After polling, check `driver.authPending()` again. If it remains true, flush already-queued output but do not open another carrier record; retain the ciphertext in the bounded carrier queue. Once resume succeeds and the buffered Finished advances the read epoch, process the retained record normally.

Add real socket-pair tests in both directions: an async client signer with server application data queued after ServerFinished, and an async server verifier with client application data queued after ClientFinished. Neither may fail, expose plaintext early, or lose the record.

src/quic/connection.zig
Jaruso, line 783: **[P1] Defer 1-RTT frames while a client-auth operation remains pending.**

This immediate poll may legitimately leave the operation pending. `ingest()` then continues with the next coalesced packet. Application keys already exist after the sender's Finished, but `streams` is not initialized until `afterHandshakeProgress` observes completion. A valid coalesced 1-RTT STREAM therefore reaches the branch below and closes with `stream before handshake`; RESET_STREAM/MAX_DATA/etc. can instead be silently ignored and then acknowledged, permanently losing them.

Do not process or acknowledge application packets that arrive during this authenticated-but-not-yet-resolved gap. Either retain them in a bounded pending queue, or drop them without recording/ACKing the packet so QUIC retransmission delivers them after authentication completes. No application state may become visible before the verifier accepts.

Please test this through actual `Connection` packet ingress, not only `tls_handshake.Handshake`: cover a client-side pending signer and a server-side pending verifier with Handshake and 1-RTT packets coalesced, and verify the connection eventually establishes and delivers the STREAM data.

src/tls/encrypted_stream.zig
```diff
@@ -654,6 +669,10 @@ pub const PureZigRecordStream = struct {
             error.AlpnMismatch,
             error.SecretExportFailed,
             error.InvalidHandshakeState,
+            error.NoApplicableCredential,
+            error.CredentialProviderFailed,
+            error.ClientCertificateRequired,
```
Jaruso, line 674: **[P1] Client-auth failure alerts are still sealed with keys the peer has discarded.**

Adding the errors to `mappedFatalAlert` fixes selection of the alert description, but not its encryption epoch. A server has sent its Finished and installed its application write secret before it can reject the client's empty Certificate, CertificateVerify, or Finished. This stream nevertheless keeps `write_epoch == .handshake` until full handshake completion, so `queueFatalAlert` seals `certificate_required`/`decrypt_error` with handshake keys. The client has completed after sending its own Finished and discarded those read keys, so it sees a record-authentication failure rather than the required TLS alert. The new test exposes this by asserting only `failures.client != null` instead of `PeerFatalAlert`.

Track TLS 1.3's directional epoch transitions, or explicitly select the application write keys for a server-side failure after ServerFinished. Client-side failures before sending ClientFinished must still use handshake write keys.

Strengthen the end-to-end tests to require `PeerFatalAlert` for required-client-certificate decline, rejected client certificate, and bad client CertificateVerify—not merely any remote error.

src/tls/tls13_backend.zig
```diff
@@ -1076,14 +1734,153 @@ pub const Tls13Backend = struct {
...
+        if (self.pending_op == null and self.core.handshake_lifecycle == .running)
+            try self.drainInput(&self.handshake_input, .handshake, sink);
+    }
+
+    /// Continue the suspended stage from its completion value. Each arm resumes
+    /// exactly where the synchronous path would have continued, recording no
+    /// handshake message twice.
+    fn dispatchResume(self: *Tls13Backend, stage: PendingStage, completion: credentials.Completion, sink: *EventSink) HandshakeError!void {
+        switch (stage) {
+            .server_select => switch (completion) {
```
Jaruso, line 1789: **[P1] Preserve the normal “no server credential” result across asynchronous selection.**

`Completion.no_credential` is documented as a credential-selection outcome, but only `.client_select` handles it. A synchronous server selector returning `NoCredentialAvailable`/`NoCompatibleSignatureAlgorithm` becomes `NoApplicableCredential` and `handshake_failure`; the equivalent asynchronous selector becomes `provider_internal_failure` and `internal_error` here.

Handle `.no_credential` in `.server_select` as a no-applicable-credential failure. A stage-specific selection completion that distinguishes unavailable from incompatible would preserve the full diagnostic taxonomy, but both must at least retain the canonical `handshake_failure` wire result. Add record and QUIC tests for an asynchronous server selector resolving with no credential.

src/quic/connection.zig
```diff
@@ -1228,6 +1265,8 @@ pub const Connection = struct {
         return switch (err) {
             error.AlpnMismatch => error_crypto_base + 120, // no_application_protocol
             error.CertificateInvalid => error_crypto_base + 42, // bad_certificate
+            error.ClientCertificateRequired => error_crypto_base + 116, // certificate_required
+            error.DecryptError => error_crypto_base + 51, // decrypt_error
```
Jaruso, line 1269: **[P1] QUIC maps `NoApplicableCredential` to the wrong TLS alert.**

The canonical TLS mapper sends `handshake_failure` (40) when this side cannot produce a compatible credential, but this switch has no `NoApplicableCredential` arm, so the default produces CRYPTO_ERROR(`internal_error`, 80). That makes the wire result transport-dependent and violates the failure taxonomy introduced by this PR.

Add:
```zig
error.NoApplicableCredential => error_crypto_base + 40,
```
Prefer delegating ordinary TLS errors to `alerts.fromHandshakeError` so the record and QUIC tables cannot drift again. Exercise both synchronous and asynchronous no-server-credential failures through an actual QUIC `Connection`.

src/tls/tls13_backend.zig
```diff
@@ -1076,14 +1734,153 @@ pub const Tls13Backend = struct {
...
+        self.pending_op = null;
+        try self.dispatchResume(stage, completion, sink);
+
+        // If the resume finished the stage without re-parking or failing, drain
+        // any handshake-epoch messages that arrived coalesced with — and were
+        // buffered behind — the message that triggered the suspend (e.g. a
+        // server that suspended verifying the client CertificateVerify still has
+        // the client Finished waiting in its handshake buffer).
+        if (self.pending_op == null and self.core.handshake_lifecycle == .running)
+            try self.drainInput(&self.handshake_input, .handshake, sink);
```
Jaruso, line 1781: **[P1] Resume must drain the epoch that actually parked, not only `handshake_input`.**

Server credential selection and server signing can park while processing a ClientHello from `initial_input`. Any additional Initial CRYPTO received while pending is correctly appended and gated, but this resume path never revisits that buffer. A concrete case is ClientHello A followed by a second ClientHello B: A parks the selector, B is buffered, resume emits the server flight, and B is silently ignored for the rest of the connection instead of producing `unexpected_message`.

Track the pending source epoch, or drain bounded buffers in protocol order (`initial_input` before `handshake_input`) after a successful resume. The invariant should be that every accepted transport byte is eventually parsed or deterministically rejected. Add coalesced and separate-receive regressions for extra Initial input behind async server selection and signing.

src/tls/tls13_backend.zig
```diff
@@ -468,13 +562,39 @@ pub const Tls13Backend = struct {
...
+        input: *tls_handshake_codec.Reassembler(max_message_len + 4),
+        level: EncryptionLevel,
+        sink: *EventSink,
+    ) HandshakeError!void {
         while (input.peek() catch |err| return mapCoreError(err)) |message| {
+            // Defensive: never dispatch a message while an operation is parked,
+            // even if this loop is somehow re-entered mid-suspend.
+            if (self.pending_op != null) break;
             if (level != try expectedLevel(message.kind)) return error.UnexpectedTransportEpoch;
             const transcript_before = self.core.transcriptHash();
```
Jaruso, line 590: **[P2] Do not leave complete trailing handshake messages unexamined after Finished.**

The later completion break establishes the handshake as soon as the first terminal Finished is processed, even if another complete handshake message is already present in the same reassembly buffer. The trailing message is retained forever and the transport can expose an established connection without ever rejecting it. A later Handshake-epoch receive has related problems because `Core.acceptReceived` permits the `.complete` lifecycle for NewSessionTicket handling and can fall through for other message kinds.

Reject any trailing Handshake/Initial bytes before exposing `handshake_complete`, and make the completed core accept only legal application-epoch post-handshake messages. Do not merely remove the break after completion events have already been emitted; validate that no same-epoch suffix exists before committing terminal completion, or ensure an outcome carrying a terminal error cannot apply stale success events.

Test `Finished || second Finished`, `Finished || Certificate`, and a separate post-completion Handshake receive.

src/tls/tls13_backend.zig
```diff
@@ -1076,14 +1734,153 @@ pub const Tls13Backend = struct {
...
+
+    /// Poll a parked authentication operation. When it is still pending this is
+    /// a no-op; when it completes the handshake resumes exactly where it
+    /// suspended, recording no handshake message twice; when it fails the typed
+    /// failure is latched.
+    pub fn resumeAuth(self: *Tls13Backend, sink: *EventSink) HandshakeError!void {
+        const op = self.pending_op orelse return error.InvalidHandshakeState;
+        const stage = self.pending_stage;
+        var completion: credentials.Completion = undefined;
+        const done = op.poll(&completion) catch {
```
Jaruso, line 1766: **[P2] Preserve `OperationError.InvalidCallbackBehavior` and do not cancel an operation that has already terminated with an error.**

This catch discards the actual `OperationError`; both `OperationFailed` and `InvalidCallbackBehavior` are mapped only from the stage. It therefore defeats the typed callback-failure taxonomy. It also calls `cancelPendingAuth`, although the `PendingOperation` contract says a poll error terminates the operation and `cancel` is for abandoning an operation before resolution. A real HSM/remote-operation implementation may not permit cancel after terminal failure.

On poll error, clear and release the operation exactly once without cancelling it, release any held signing credential, then classify:
```zig
error.InvalidCallbackBehavior => .invalid_callback_behavior,
error.OperationFailed => pendingFailureClass(stage),
```
Keep cancellation for teardown while `poll` still reports pending. Update tests to assert `cancel_count == 0`, `op_release_count == 1`, and the precise failure class for terminal poll errors.

src/tls/tls13_backend.zig
```diff
@@ -1076,14 +1734,153 @@ pub const Tls13Backend = struct {
...
+        self.pending_op = op;
+        self.pending_stage = stage;
+    }
+
+    /// Poll a parked authentication operation. When it is still pending this is
+    /// a no-op; when it completes the handshake resumes exactly where it
+    /// suspended, recording no handshake message twice; when it fails the typed
+    /// failure is latched.
+    pub fn resumeAuth(self: *Tls13Backend, sink: *EventSink) HandshakeError!void {
+        const op = self.pending_op orelse return error.InvalidHandshakeState;
```
Jaruso, line 1763: **[P2] The concrete backend violates the generic resume contract's safe no-op guarantee.**

`transport.Backend.resumeAuth` is documented as safe when nothing is parked, and both production owners are encouraged to call it opportunistically. This concrete implementation instead returns `InvalidHandshakeState` when `pending_op` is null. The higher-level wrappers currently guard most calls, but a generic `engine.Driver` caller following the contract can turn an idle poll into a terminal handshake failure.

Return success when no operation is pending (or change the generic contract and every caller consistently). Add a generic-driver regression that calls `resumeAuth` before a suspend and after completion and observes an empty event batch, not failure.

src/tls/tls13_backend.zig
```diff
@@ -886,10 +1393,41 @@ pub const Tls13Backend = struct {
...
+                    try ext.expectEnd();
+                },
+                ext_server_name => {
+                    // RFC 6066 §3: ServerNameList<1..>, each { name_type, name<..> }.
+                    // Distinguish absent (no extension / no host_name) from valid
+                    // and from malformed: an empty, over-bound, or duplicated
+                    // host_name is rejected deterministically rather than being
+                    // collapsed into the default-certificate path, so a selector
+                    // never serves the default cert for an invalid SNI.
+                    var names = Reader{ .bytes = try ext.slice(try ext.u16_()) };
```
Jaruso, line 1418: **[P2] Reject an empty `ServerNameList` instead of treating it as absent SNI.**

The RFC 6066 vector is `ServerNameList<1..2^16-1>`. An extension body of `{ 0x00, 0x00 }` creates an empty reader, skips the loop, and leaves `server_name_present == false`, routing malformed SNI to the default credential—the metadata-collapse behavior this PR is intended to eliminate.

Capture the list length, reject zero before iterating, and add a raw-extension regression for the empty list in addition to the existing empty-host-name and duplicate-host-name cases.

src/tls/tls13_backend.zig
```diff
@@ -753,49 +1007,107 @@ pub const Tls13Backend = struct {
...
-        try sink.emitCertificate(state);
-        if (state == .invalid) {
-            self.core.handshake_lifecycle = .failed;
-            return error.CertificateInvalid;
+        // §4.4.3) — before this message is added. Proof of key possession is
+        // transcript crypto, not PKI policy: a failure is a `decrypt_error`, a
+        // distinct crypto failure from a trust rejection, so it does NOT emit a
+        // certificate-invalid (bad_certificate) verdict — it fails directly.
+        const content = certificateVerifyContent(signer_role, transcript_before);
+        if (!self.checkProofOfPossession(algorithm, signature, content.slice())) {
```
Jaruso, line 1023: **[P2] Do not classify every CertificateVerify validation failure as `decrypt_error`.**

`checkProofOfPossession` returns one boolean for several materially different conditions: malformed certificate DER, unsupported certificate key type, an algorithm the peer never offered, malformed signature encoding, and an actual cryptographic signature mismatch. This caller maps all of them to `certificate_verify_invalid`/`decrypt_error`.

Keep the failure typed:
- parse/structural certificate failure → `invalid_peer_certificate_chain` / `bad_certificate`;
- a CertificateVerify algorithm not present in this side's advertised list (ClientHello or CertificateRequest) → `IllegalParameter`;
- unsupported certificate/key type → the chosen deterministic certificate alert (`unsupported_certificate` or `bad_certificate`);
- only a well-formed, permitted signature that fails verification → `DecryptError`.

Add both server-auth and client-auth tests for malformed DER and unoffered algorithms, alongside the existing flipped-signature test.

src/tls/tls13_backend.zig
```diff
@@ -917,7 +1455,12 @@ pub const Tls13Backend = struct {
                 },
             }
         }
-        if (!offers_tls13 or !offers_x25519_group or !offers_our_sigalg) return error.MalformedHandshake;
+        if (!offers_tls13 or !offers_x25519_group) return error.MalformedHandshake;
+        // signature_algorithms is required whenever the server authenticates
+        // with a certificate (RFC 8446 §9.2). A missing or empty list is a
+        // malformed/missing required *peer* extension — attribute it to the
+        // peer (decode_error), not to local credential configuration.
+        if (!saw_signature_algorithms or self.peer_sig_scheme_count == 0) return error.MalformedHandshake;
```
Jaruso, line 1463: **[P2] Missing and malformed `signature_algorithms` still produce the same alert.**

The parser now tracks whether the extension was present, but both absence and an empty/malformed vector return `MalformedHandshake`/`decode_error`. RFC 8446's required-extension path is `missing_extension`; `AlertDescription.missing_extension` already exists, but the handshake error vocabulary cannot surface it.

Introduce a typed `MissingExtension` error and map it consistently in record and QUIC transports. Use it when `signature_algorithms` is absent from ClientHello or CertificateRequest; retain a decode/semantic error for a present but empty or malformed vector. Add tests that assert the two cases produce distinct wire alerts.

src/tls/tls13_backend.zig
```diff
@@ -634,6 +766,19 @@ pub const Tls13Backend = struct {
...
+        if (self.serverNameSlice()) |name| {
+            try w.u16_(ext_server_name);
+            const sni_ext_len = try w.reserve(2);
+            const sni_list_len = try w.reserve(2);
+            try w.u8_(0); // name_type host_name
+            try w.u16_(@intCast(name.len));
+            try w.bytes(name);
+            w.patch(2, sni_list_len);
+            w.patch(2, sni_ext_len);
+        }
```
Jaruso, line 780: **[P2] Preflight the complete ClientHello against a buffer large enough for all individually valid options.**

The newly added SNI can make a configuration that passes every current bound impossible to serialize here. With ALPN at 255 bytes, SNI at 256 bytes, and a permitted 512-byte transport extension, the encoded ClientHello is roughly 1.15 KiB, but this writer has only 1024 bytes. `start` has already advanced the core and generated/stored the key pair when the writer reports overflow.

Use a buffer derived from `max_message_len`, or compute the exact ClientHello size during profile/start validation and reject the configuration before lifecycle/transcript/key mutation. Add a boundary test combining maximum ALPN, SNI, and transport-extension lengths; it should either serialize successfully or fail as an explicit configuration error before `core.start`.

Jaruso (commented):
Blocking follow-up review at `d2e6581c`.

The previous eight findings were substantially addressed: later receives are gated while verification is pending, QUIC now has an explicit progress path, presented client certificates require an accepted verdict, client selectors can decline with an empty Certificate, flight sizing is improved, duplicate CertificateRequest is rejected, and malformed completions release their handles. CI is also green.

The implementation is nevertheless not ready to merge. The new asynchronous client-auth paths expose transport and epoch transitions that the current direct-backend tests do not cover. Most importantly, valid application traffic arriving after a peer's Finished can be opened with the wrong TLS record keys or can close/drop state in QUIC while an authentication callback remains pending. The record-layer `certificate_required` fix also maps the alert but still encrypts it under keys the peer has already discarded.

I found additional deterministic-error and input-draining gaps inline. Please add actual record-socket and QUIC `Connection` tests for the pending-auth/application-data cases rather than only direct backend/handshake-adapter tests.

After these are fixed, also update the stale profile/lifetime documentation (`tls13_backend.zig` still says server-only authentication and `CertificateChain` comments conflict with the normative selected-credential lifetime), rebase onto current `main`, and rerun the complete matrix.

PR #459, opened 2026-07-18
tls: [323-F] handshake-time client authentication and key selection (#334)

Jaruso:
## Summary

Completes issue #334 by integrating the green credential-contract PR with the
handshake-time client-authentication WIP and finishing it, in four strict
checkpoints (each committed and green before the next):

- **Checkpoint 1** — restore the merged branch: formatting fix; the tree
  already compiled and passed `test-tls`/`test-quic`/`test`/appliance.
- **Checkpoint 2** — synchronous handshake-time client authentication:
  disabled/optional/required modes; CertificateRequest parse/emit; the client
  Certificate / optional CertificateVerify / Finished flight; role-aware
  Certificate and CertificateVerify processing; the client CertificateVerify
  context string; Finished MACs recomputed over the correct transcript. New
  `certificate_required` alert path.
- **Checkpoint 3** — asynchronous client authentication: resumable client
  credential selection and signing, server-side client-certificate verification
  through the existing async machinery, receive gating past Finished while
  pending, automatic drain of buffered handshake messages after resume, and
  exactly-once cancel/release of every operation and credential.
- **Checkpoint 4** — remaining correctness fixes: reject oversized configured
  server names; validate exact client-flight sizes before transcript mutation
  or output; map a failed CertificateVerify signature to `decrypt_error`
  (RFC 8446 §4.4.3); expose pending/resume through the generic transport
  Backend, the engine Driver, and both the record and QUIC production drivers.

No OpenSSL/X.509 objects cross the credential interface; failures map to
deterministic TLS alerts. Post-handshake client authentication remains
explicitly deferred.

## Test plan
- [x] `zig fmt --check src build.zig build.zig.zon`
- [x] `zig build test-tls`
- [x] `zig build test-quic`
- [x] `zig build test`
- [x] `zig build test -Dtls-profile=appliance`
- [x] New sync success/failure, async selection/signing/verification,
  cancellation, malformed-completion, coalesced-flight, byte-by-byte
  fragmentation, oversized-server-name, decrypt_error, and production-driver
  (record stream + generic driver) tests.

Closes #334

### DIFF
--- a/src/quic/connection.zig
+++ b/src/quic/connection.zig
@@ -1131,6 +1131,34 @@ pub const Connection = struct {
         }
     }
 
+    /// True while the TLS handshake has suspended awaiting an asynchronous
+    /// authentication operation (#334). The event loop can poll this to know
+    /// the connection is waiting on an external signer/verifier/selector; each
+    /// `pollTransmit` (or an explicit `driveAuthentication`) advances it.
+    pub fn authPending(self: *const Connection) bool {
+        return self.handshake.authPending();
+    }
+
+    /// Drive a parked asynchronous authentication operation forward without
+    /// requiring another inbound CRYPTO frame: poll it, queue any handshake
+    /// output it produced, and complete the handshake if it resolved. Safe to
+    /// call unconditionally; a no-op unless the backend is suspended.
+    pub fn driveAuthentication(self: *Connection, now_us: u64) void {
+        if (self.state_ == .closed or self.state_ == .draining or self.state_ == .closing) return;
+        if (!self.handshake.authPending()) return;
+        self.handshake.resumeAuth() catch |err| {
+            self.failHandshake(err);
+            self.startClose(.{ .error_code = cryptoErrorCode(err), .is_application = false, .local = true }, @errorName(err), now_us);
+            return;
+        };
+        self.collectCryptoOutput() catch {
+            self.failHandshake(error.HandshakeBufferOverflow);
+            self.startClose(.{ .error_code = error_crypto_buffer_exceeded, .is_application = false, .local = true }, "crypto buffer", now_us);
+            return;
+        };
+        self.afterHandshakeProgress(now_us);
+    }
+
     fn afterHandshakeProgress(self: *Connection, now_us: u64) void {
         if (self.handshake_complete or !self.handshake.isComplete()) return;
         self.handshake_complete = true;
@@ -1536,6 +1564,19 @@ pub const Connection = struct {
             self.close_needs_send = false;
             self.close_resend_allowed_at_us = now_us + self.ptoDurationNow() / 2;
             return self.buildCloseDatagram(out, now_us);
+        }
+
+        // Advance any parked asynchronous authentication (#334) before building
+        // packets, so a resolved external signer/verifier/selector emits its
+        // next flight here even when no further peer CRYPTO frame will arrive.
+        self.driveAuthentication(now_us);
+        if (self.state_ == .closing or self.state_ == .closed or self.state_ == .draining) {
+            if (self.state_ == .closing and self.close_needs_send) {
+                self.close_needs_send = false;
+                self.close_resend_allowed_at_us = now_us + self.ptoDurationNow() / 2;
+                return self.buildCloseDatagram(out, now_us);
+            }
+            return null;
         }
 
         // Force the delayed app-space ACK when its timer expired.
--- a/src/quic/connection.zig
+++ b/src/quic/connection.zig
@@ -1228,6 +1228,7 @@ pub const Connection = struct {
         return switch (err) {
             error.AlpnMismatch => error_crypto_base + 120, // no_application_protocol
             error.CertificateInvalid => error_crypto_base + 42, // bad_certificate
+            error.ClientCertificateRequired => error_crypto_base + 116, // certificate_required
             error.MissingTransportParameters, error.InvalidTransportParameters => error_transport_parameter,
             error.UnexpectedHandshakeMessage => error_crypto_base + 10, // unexpected_message
             error.IllegalParameter => error_crypto_base + 47, // illegal_parameter
--- a/src/quic/connection.zig
+++ b/src/quic/connection.zig
@@ -107,6 +107,7 @@ pub const Event = union(enum) {
 
 pub const DropReason = enum {
     unknown_cid,
+    keys_unavailable,
     undecryptable,
     malformed,
     unsupported_version,
@@ -645,6 +646,10 @@ pub const Connection = struct {
             else => unreachable,
         };
         const space = spaceForLevel(level);
+        if (level == .application and !self.handshake_complete) {
+            self.dropPacket(.keys_unavailable, bytes.len);
+            return;
+        }
 
         // Header protection removal needs a sample 4 bytes past pn_offset.
         var work: [2048]u8 = undefined;
@@ -2347,12 +2352,33 @@ test "handshake failures map to their RFC 9001 CRYPTO_ERROR alert codes" {
     try testing.expectEqual(error_crypto_base + 50, Connection.cryptoErrorCode(error.MalformedHandshake));
     try testing.expectEqual(error_crypto_base + 120, Connection.cryptoErrorCode(error.AlpnMismatch));
     try testing.expectEqual(error_crypto_base + 42, Connection.cryptoErrorCode(error.CertificateInvalid));
+    try testing.expectEqual(error_crypto_base + 43, Connection.cryptoErrorCode(error.UnsupportedCertificate));
+    try testing.expectEqual(error_crypto_base + 109, Connection.cryptoErrorCode(error.MissingExtension));
     // #334 review: NoApplicableCredential was missing from this table and fell
     // through to the generic internal_error(80) code instead of the canonical
     // handshake_failure(40) `alerts.fromHandshakeError` maps it to.
     try testing.expectEqual(error_crypto_base + 40, Connection.cryptoErrorCode(error.NoApplicableCredential));
     try testing.expectEqual(error_crypto_base + 116, Connection.cryptoErrorCode(error.ClientCertificateRequired));
     try testing.expectEqual(error_crypto_base + 51, Connection.cryptoErrorCode(error.DecryptError));
+}
+
+test "driver: 1-RTT packets are dropped before deprotection until TLS handshake complete" {
+    const allocator = testing.allocator;
+    var pair = try TestPair.init(allocator);
+    defer pair.deinit(allocator);
+    try pair.pump();
+
+    const id = try pair.client.openStream(.bidi);
+    try testing.expectEqual(@as(usize, 5), try pair.client.writeStream(id, "hello", false));
+    var datagram: [max_datagram_size]u8 = undefined;
+    const one_rtt = pair.client.pollTransmit(&datagram, pair.now_us) orelse return error.TestExpectedEqual;
+
+    const received_before = pair.server.metrics.packets_received;
+    const dropped_before = pair.server.metrics.packets_dropped;
+    pair.server.handshake_complete = false;
+    try pair.server.ingest(one_rtt, pair.now_us);
+    try testing.expectEqual(received_before, pair.server.metrics.packets_received);
+    try testing.expectEqual(dropped_before + 1, pair.server.metrics.packets_dropped);
 }
 
 test "a real Connection closes with handshake_failure when the server has no applicable credential" {
--- a/src/quic/connection.zig
+++ b/src/quic/connection.zig
@@ -776,6 +776,15 @@ pub const Connection = struct {
                     self.startClose(.{ .error_code = cryptoErrorCode(err), .is_application = false, .local = true }, @errorName(err), now_us);
                     return;
                 };
+                // Poll any parked asynchronous authentication (#334) so a
+                // resolved external signer/verifier/selector progresses the
+                // handshake as this packet is processed. A no-op when nothing is
+                // suspended.
+                self.handshake.resumeAuth() catch |err| {
+                    self.failHandshake(err);
+                    self.startClose(.{ .error_code = cryptoErrorCode(err), .is_application = false, .local = true }, @errorName(err), now_us);
+                    return;
+                };
                 self.collectCryptoOutput() catch {
                     self.failHandshake(error.HandshakeBufferOverflow);
                     self.startClose(.{ .error_code = error_crypto_buffer_exceeded, .is_application = false, .local = true }, "crypto buffer", now_us);
@@ -1229,6 +1238,7 @@ pub const Connection = struct {
             error.AlpnMismatch => error_crypto_base + 120, // no_application_protocol
             error.CertificateInvalid => error_crypto_base + 42, // bad_certificate
             error.ClientCertificateRequired => error_crypto_base + 116, // certificate_required
+            error.DecryptError => error_crypto_base + 51, // decrypt_error
             error.MissingTransportParameters, error.InvalidTransportParameters => error_transport_parameter,
             error.UnexpectedHandshakeMessage => error_crypto_base + 10, // unexpected_message
             error.IllegalParameter => error_crypto_base + 47, // illegal_parameter
--- a/src/quic/connection.zig
+++ b/src/quic/connection.zig
@@ -2146,6 +2146,48 @@ const TestPair = struct {
         return pair;
     }
 
+    fn initClientAuth(
+        allocator: std.mem.Allocator,
+        client_provider: tls_core.credentials.CredentialProvider,
+        server_verifier: tls_core.credentials.PeerVerifier,
+    ) !*TestPair {
+        const pair = try allocator.create(TestPair);
+        pair.* = .{
+            .client_backend = tls_backend_mod.Tls13Backend.initClient(
+                .{ .hello_random = [_]u8{0xc1} ** 32, .key_share_seed = [_]u8{0x11} ** 32 },
+                .{ .pinned_certificate = tls_backend_mod.testdata.certificate_der },
+            ),
+            .server_backend = tls_backend_mod.Tls13Backend.initServer(
+                .{ .hello_random = [_]u8{0x51} ** 32, .key_share_seed = [_]u8{0x22} ** 32 },
+                try tls_backend_mod.Identity.initPkcs8(
+                    tls_backend_mod.testdata.certificate_der,
+                    tls_backend_mod.testdata.private_key_pkcs8_der,
+                ),
+            ),
+        };
+        errdefer allocator.destroy(pair);
+        pair.client_backend.engine.setLocalCredentialProvider(client_provider);
+        pair.server_backend.engine.requestClientAuthentication(.required, server_verifier);
+        pair.client = try Connection.init(allocator, .{
+            .role = .client,
+            .local_cid = &client_cid,
+            .original_dcid = &odcid,
+            .peer_cid = &odcid,
+            .tls = pair.client_backend.backend(),
+            .now_us = pair.now_us,
+        });
+        errdefer pair.client.deinit();
+        pair.server = try Connection.init(allocator, .{
+            .role = .server,
+            .local_cid = &odcid,
+            .original_dcid = &odcid,
+            .peer_cid = &client_cid,
+            .tls = pair.server_backend.backend(),
+            .now_us = pair.now_us,
+        });
+        return pair;
+    }
+
     fn deinit(self: *TestPair, allocator: std.mem.Allocator) void {
         self.client.deinit();
         self.server.deinit();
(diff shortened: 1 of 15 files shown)


Jaruso (commented):
## Blocking review — PR #459 at `33d9108e`

I reviewed the full PR against issue #334 and rechecked the findings carried over from #458. CI is green. This WIP also fixes several important items from the previous round:

- asynchronous server selection now preserves the normal no-credential outcome;
- QUIC delegates ordinary TLS failures to the canonical alert mapper;
- `resumeAuth` is a safe no-op with no parked operation;
- terminal poll errors preserve `InvalidCallbackBehavior` and no longer cancel an already-terminated operation;
- resume revisits buffered Initial input;
- empty `ServerNameList` is rejected;
- the ClientHello buffer now fits the documented maximum configuration.

The PR is **not safe to merge yet**. The remaining issues are below, ordered by severity.

---

### [P1] Record mode consumes the next record while authentication is still pending

`resumeAuthIfPending()` returns `true` whenever it polls, including when `PendingOperation.poll` returns `false`. `drive()` then continues into `processCarrierInputBudget` and the carrier read loop.

This breaks valid TLS 1.3 traffic in both directions:

- after ServerFinished, a server may send application data while the client is waiting on an asynchronous client-credential selector/signer;
- after ClientFinished, client application data may already be queued while the server is waiting on an asynchronous client-certificate verifier.

The stream may still have `read_epoch == .handshake`, so the next application record is opened with handshake keys and the connection fails instead of retaining the ciphertext until authentication resolves.

After polling, re-check the driver. It is fine to flush output already queued by the handshake, but do not open another inbound record or perform another carrier read while authentication remains pending:

```zig
if (try self.resumeAuthIfPending()) made_progress = true;

const auth_still_pending = if (self.handshake_driver) |*driver|
    driver.authPending()
else
    false;

// Run the outbound write loop if desired, then return before inbound parsing.
if (auth_still_pending) {
    return .{ .made_progress = made_progress, .readiness = self.readiness() };
}
```

Add real socket-pair tests for async client signing with server application data immediately after ServerFinished, and async server verification with client application data immediately after ClientFinished. The ciphertext must be retained, never exposed early, and delivered after acceptance.

---

### [P1] QUIC decrypts, processes, and acknowledges 1-RTT packets before authentication resolves

The opportunistic poll after a CRYPTO frame may legitimately leave the operation pending. `ingestPacket`, however, has already decrypted the packet and updated `largest_recv_pn`, recovery state, packet events, and eventual ACK state before applying frames.

A coalesced or subsequent 1-RTT packet can then:

- close on a STREAM frame because `streamManager()` is still null;
- silently lose RESET_STREAM, MAX_DATA, and similar frames because their managers do not exist yet;
- still be acknowledged, preventing retransmission of the lost frame.

RFC 9001 §5.7 requires an endpoint not to decrypt 1-RTT packets before the TLS handshake is complete. A server also cannot acknowledge such a packet because an ACK confirms the frames were processed.

Gate application packets before deprotection and packet-number/recovery bookkeeping. Until `handshake_complete`, either retain them in a bounded queue or drop them without ACK state so the peer retransmits:

```zig
if (level == .application and !self.handshake_complete) {
    // No decryption, packet_received event, recovery insertion, or ACK state.
    self.dropPacket(.keys_unavailable, bytes.len);
    return;
}
```

Use the appropriate existing/new drop reason. Test this through actual `Connection.ingest` with both a pending client signer and pending server verifier, followed immediately by 1-RTT STREAM data. The connection must eventually establish and deliver the data exactly once.

---

### [P1] Client-auth failure alerts are encrypted with the wrong record keys

Adding `ClientCertificateRequired` and `DecryptError` to `mappedFatalAlert` selects the correct description, but `queueFatalAlert` still seals using `write_epoch`. Record mode advances `write_epoch` to application only at full `handshake_complete`.

A server performing handshake-time client authentication has already sent ServerFinished and installed its application write secret before it can reject the client's empty Certificate, CertificateVerify, or Finished. Records sent by the server after its Finished—including fatal alerts responding to the client-auth flight—must use application traffic keys. Sealing `certificate_required`, `bad_certificate`, or `decrypt_error` with handshake keys makes the client observe an AEAD failure instead of the required TLS alert.

Track directional epoch transitions rather than tying both directions to overall completion:

```zig
// Server, after its Finished is emitted:
write_epoch = .application;

// Client, after authenticating ServerFinished:
read_epoch = .application;

// Client write and server read remain handshake through ClientFinished.
```

The current required-client-auth test masks this by accepting any client error. Strengthen record end-to-end tests to require `error.PeerFatalAlert` for an empty required client Certificate, rejected client chain, and bad client CertificateVerify.

---

### [P1] A terminal Finished can establish the connection while trailing handshake bytes remain ignored

`drainInput` breaks as soon as the lifecycle changes to complete, leaving any same-epoch suffix in the reassembler. A peer can send `Finished || Finished`, `Finished || Certificate`, or a partial next handshake frame; the first Finished publishes success while the suffix is retained forever and never rejected.

The async path has the same issue: a client selection/sign completion can emit ClientFinished and complete the lifecycle inside `dispatchResume`; `resumeAuth` then skips both drains because it only drains while the lifecycle is running, abandoning Handshake bytes accepted while the operation was pending.

Completion must not be published until the terminal input boundary is clean. One workable shape is:

```zig
try self.onMessage(...); // validates Finished, marks terminal completion pending
try input.discard(message.raw.len);

if (terminal_pending and input.len != 0)

try self.publishHandshakeComplete(sink);
```

A partial suffix is also invalid at this boundary; it cannot become a legal continuation after Finished. Test `Finished || Finished`, `Finished || Certificate`, a partial framed suffix, a separate Handshake receive after completion, and extra Handshake bytes buffered during async client selection/signing. No success event or application exposure may occur before rejection.

---

### [P2] CertificateVerify validation collapses unrelated failures into `decrypt_error`

`checkProofOfPossession` returns one boolean for malformed leaf DER, unsupported certificate/key type, algorithm/key mismatch, an algorithm outside the advertised set, malformed signature encoding, and an actual cryptographic signature mismatch. The caller converts every false result into `certificate_verify_invalid` / `decrypt_error`.

Keep the result typed, for example:

```zig
const ProofResult = union(enum) {
    valid,
    invalid_signature,
    invalid_certificate,
    unsupported_certificate,
    unoffered_algorithm,
};
```

Then map:

- structural certificate failure → `invalid_peer_certificate_chain` / `bad_certificate`;
- unsupported certificate or key → the selected deterministic certificate alert;
- CertificateVerify algorithm outside this side's advertised ClientHello or CertificateRequest list → `IllegalParameter`;
- actual signature-verification failure → `DecryptError`.

Malformed signature encoding may reasonably remain a signature-verification failure, but malformed certificate DER must not masquerade as a transcript-signature mismatch. Add server-auth and client-auth tests for malformed DER, an unoffered algorithm, unsupported key type, and the existing flipped-signature case.

---

### [P2] Missing `signature_algorithms` still maps to `decode_error` rather than `missing_extension`

Both ClientHello and CertificateRequest merge two distinct cases:

- required extension absent → `missing_extension`;
- extension present but empty or malformed → malformed/illegal extension data.

`AlertDescription.missing_extension` already exists, but the handshake error vocabulary cannot surface it. Add a typed error and map it consistently:

```zig
pub const HandshakeError = error{
    // ...
    MissingExtension,
};

// alerts.zig
error.MissingExtension => .missing_extension,
```

Return it only when the required extension is absent. Preserve `MalformedHandshake` or `IllegalParameter` for a present empty/invalid vector. Add record and QUIC tests asserting distinct wire results for absent versus malformed `signature_algorithms` in ClientHello and CertificateRequest.

---

### [P2] A bad Finished verify_data maps to `decode_error`, not `decrypt_error`

The Finished length check is correctly a framing error. A correctly sized Finished whose verify_data does not match is a cryptographic authentication failure. RFC 8446 §4.4.4 requires `decrypt_error` when Finished verification fails.

Fix both directions:

```zig
if (body.len != hash_len) return error.MalformedHandshake;
if (!crypto.timing_safe.eql(...)) return error.DecryptError;
```

Add record and QUIC end-to-end tests for corrupted server Finished and corrupted client Finished, asserting alert/CRYPTO_ERROR 51 instead of 50.

---

### [P2] An empty configured server name is emitted as malformed SNI

`null` correctly represents no SNI, but `server_name = ""` sets `server_name_present = true`. `sendClientHello` then emits a zero-length host_name entry, even though this engine correctly rejects that same value when parsing incoming SNI.

Reject it before lifecycle/transcript/key mutation:

```zig
if (name.len == 0 or name.len > self.server_name.len) {
    self.server_name_invalid = true;
} else {
    @memcpy(self.server_name[0..name.len], name);
    self.server_name_len = name.len;
    self.server_name_present = true;
}
```

A fallible constructor would be clearer, but the existing deferred-start validation pattern is sufficient. Add a test proving start emits no ClientHello and does not advance handshake state for an empty configured hostname.

---

### [P2] Wrong-kind async completions use inconsistent failure classes

The pending-operation contract defines a wrong-kind completion as invalid callback behavior, and the sign-stage branch follows that taxonomy. The selection branches instead report `provider_internal_failure`, while peer verification reports `verifier_internal_failure`.

A callback that completes selection with a verdict/signature length—or verification with a credential—is not an ordinary provider outage; it violated the interface contract. Use the same classification everywhere:

```zig
else => {
    releaseCompletionCredentials(completion, null);
}
```

Apply the equivalent correction to `server_select`, `client_select`, and `peer_verify`. The wire alert remains `internal_error`, but `credentialFailure()` will accurately identify the programming error. Add one malformed-completion test per stage while retaining exactly-once cleanup assertions.

---

The acceptance criteria around provider/verifier interchangeability, SNI/sigalg selection context, immutable DER views, synchronous and asynchronous client authentication, and absence of concrete OpenSSL/X.509 types are otherwise substantially implemented. The P1 transport and terminal-boundary issues above need to be fixed before merge.

Jaruso (commented):
## Follow-up review — PR #459 at `86c250f4`

CI is green, and this commit correctly addresses most of the previous nine findings: record input is gated while auth remains pending, pre-completion QUIC 1-RTT packets are dropped before deprotection/ACK bookkeeping, record epochs now advance directionally, missing extensions and bad Finished MACs have typed alerts, configured empty SNI is rejected, CertificateVerify failures are split into typed outcomes, and wrong-kind completion tags now consistently report `invalid_callback_behavior`.

The PR is **still not ready to merge**. I found one residual authentication-boundary bug and five contract/backpressure issues that are not covered by the new tests.

---

### [P1] Bytes received while client selection/signing is pending are still abandoned when resume completes the handshake

The new `Finished` suffix check only protects bytes already present in `handshake_input` when the peer Finished is first examined. A separate receive after that message can still bypass it:

1. The client authenticates ServerFinished and enters `beginClientAuthFlight`.
2. `client_select` or `client_sign` parks.
3. The peer sends another Handshake/Initial CRYPTO fragment while the operation is pending. `receiveImpl` appends it and returns without dispatching, as intended.
4. `resumeAuth` completes selection/signing. `dispatchResume` emits the client Certificate flight and Finished; `recordSent(Finished)` changes the core lifecycle to `complete`, and `handshake_complete` is emitted.
5. The post-resume drains are guarded by `handshake_lifecycle == .running`, so both buffered inputs are skipped forever.

The connection can therefore establish while an extra Finished, Certificate, or partial framed message accepted during the suspend remains unexamined. This affects both record mode and QUIC.

Validate the terminal input boundary **before** publishing the resumed client completion. For the `client_select` and `client_sign` success paths, any buffered Initial or Handshake bytes after ServerFinished are illegal. A possible shape is:

```zig
fn ensureNoPeerHandshakeSuffix(self: *Tls13Backend) HandshakeError!void {
    if (self.initial_input.len != 0 or self.handshake_input.len != 0)
}

// Before emitSelectedClientCertificate / finishClientAuthFlight can emit
// ClientFinished and handshake_complete:
try self.ensureNoPeerHandshakeSuffix();
```

Be careful to release a returned/held credential exactly once if this precheck fails. Checking only after `dispatchResume` is too late unless the success sink is reset and the core is failed before any transport consumes it.

Add regressions for both `client_select` and `client_sign`: park after a valid ServerFinished, deliver a second Finished in a separate `receive`, then resolve the operation. Resume must return `UnexpectedHandshakeMessage`, emit no client Finished/completion, and expose no application connection. Repeat with a partial framed suffix and with Initial-level bytes.

---

### [P1] Record readiness and `made_progress` can busy-spin while an async operation is still pending

`resumeAuthIfPending()` returns `true` merely because it called `poll`, even when the operation reports `false` and emits nothing. `drive()` therefore reports `made_progress = true` indefinitely for a genuinely asynchronous HSM/verifier that is not ready yet.

At the same time, `readiness().wants_read` does not account for `authStillPending()`, even though every inbound processing/read loop now explicitly refuses to read while pending. With peer bytes sitting on a level-triggered socket, the event loop is told that read readiness can make progress, repeatedly calls `drive`, leaves the fd readable, polls the unfinished operation again, and can spin at 100% CPU. A caller that loops while `made_progress` is true has the same failure without any socket event.

This contradicts the record-stream contract that readiness should register only socket interests that can actually make progress.

Return observable progress, not “a poll occurred,” and suppress read interest while authentication is parked. For example:

```zig
const outcome = driver.resumeAuthOutcome();
try self.applyDriverOutcome(outcome);
return outcome.terminal_error != null or
    outcome.sink.len != 0 or
    !driver.authPending();
```

The exact predicate can differ, but a still-pending, empty poll must not report progress. `readiness()` should use something like:

```zig
const auth_pending = self.authStillPending();
.wants_read = !auth_pending and !self.carrier_eof and self.canAcceptCarrierRead(),
```

Queued outbound handshake data should still advertise `wants_write` and be allowed to drain.

Add an event-loop regression with an operation that remains pending for many polls and a readable carrier containing peer data. Each unresolved drive must be bounded, return `made_progress == false` once no observable work occurred, advertise no read interest, and leave the ciphertext untouched until an explicit auth wake/resolution.

---

### [P2] Async resume ignores runtime outbound hard limits and can consume an operation before its event batch fits

The comments promise that an entire borrowed driver event batch is serialized atomically. `startHandshakeIfNeeded` and inbound receive paths enforce that using `canReserveHandshakeOutputBatch()`, which checks both physical capacity and the configured `BufferLimits.outbound_ciphertext.hard` watermark.

`resumeAuthIfPending`, however, checks only:

```zig
self.outbound_ciphertext.available() < handshake_output_reserve
```

With a valid custom hard limit and even one byte already queued, physical capacity can exceed the reserve while effective hard-limit room does not. The function then polls a ready operation, releases/consumes it, and starts applying its terminal event batch. `appendOutboundCiphertext` may accept early records and reject a later one with `CiphertextBufferFull`; `fail()` clears the partial flight and tears the connection down. The completed HSM/signature operation cannot be retried.

Use the existing atomic preflight:

```zig
if (!self.canReserveHandshakeOutputBatch()) return false;
```

Add a custom-limit regression with `hard == handshake_output_reserve`, one byte already queued, and a ready async signer/selector. The operation must not be polled until enough effective room exists; after the queue drains, it should be polled once and its whole flight applied without a hard-limit failure or partial output.

---

### [P2] Malformed encodings for supported public-key types are still reported as `unsupported_certificate`

The typed `ProofResult` is a good improvement, but several structural failures remain in the wrong bucket:

- an Ed25519 SPKI whose public-key bytes have the wrong length;
- `Ed25519.PublicKey.fromBytes` rejecting malformed bytes;
- `EcdsaP256.PublicKey.fromSec1` rejecting a malformed point.

Those certificates claim a key type this implementation supports; their encoded key material is corrupt. They should be `invalid_certificate` / `bad_certificate`, not `unsupported_certificate`. Reserve `unsupported_certificate` for an unsupported key algorithm or curve.

For example:

```zig
if (parsed.pub_key_algo != .curveEd25519)
    return .unsupported_certificate;
if (parsed.pubKey().len != Ed25519.PublicKey.encoded_length)
    return .invalid_certificate;
const key = Ed25519.PublicKey.fromBytes(...) catch
    return .invalid_certificate;
```

Apply the equivalent distinction to P-256. Add server-auth and client-auth tests with a syntactically parseable certificate carrying malformed Ed25519/P-256 key bytes; both should map to `bad_certificate`, while a genuinely unsupported key algorithm remains `unsupported_certificate`.

---

### [P2] Wrong-kind credential cleanup has no sound alias identity and can leak a returned handle

`releaseCompletionCredentials` suppresses release of the completion credential whenever its `handle` pointer equals the held signing credential's pointer. `SelectedCredential`, however, is a capability identified by both `handle` and `vtable`, and the public contract does not state that equal context pointers imply one shared ownership token.

A malformed signer can return:

```zig
held      = .{ .handle = p, .vtable = &held_vtable };
completion = .{ .credential = .{ .handle = p, .vtable = &other_vtable } };
```

The current helper calls `held_vtable.release(p)` and silently skips `other_vtable.release(p)`. Conversely, blindly releasing twice is unsafe if the contract intends exact capability aliasing. The generic completion union therefore leaves the engine unable to satisfy its exactly-once guarantee for every malformed callback.

The strongest fix is stage-specific completion types so a sign or verify operation cannot produce an owned credential variant at all. Otherwise, explicitly define credential identity and duplicate-reference semantics in the contract and implement them consistently; merely comparing `handle` is insufficient. At minimum, a dedupe predicate would need both `handle` and `vtable`, but even that requires the contract to say whether returning the same capability represents another owned reference.

Add a malformed-completion test using the same context pointer with two different release vtables and assert that all ownership established by the documented rule is discharged exactly once before resume returns.

---

### [P2] The public async lifecycle and borrowed-event lifetime documentation still contradict the implementation

This story's primary deliverable is a provider-neutral contract, so these are behavioral requirements rather than editorial nits:

1. The normative credential overview says every operation is “cancelled and released exactly once,” but `PendingOperation` correctly says `cancel` is only for abandonment before resolution, and the implementation correctly does **not** cancel after completion or terminal poll error. A provider author cannot satisfy both statements.
2. `transport.zig` and `docs/TLS_RECORD_TRANSPORT.md` say event payloads remain valid until the next `start`/`receive` call and enumerate four reset calls, but `resumeAuth` and `resumeAuthOutcome` also reset and overwrite the same sink. A caller following the documented lifetime can retain a borrowed slice across resume and observe overwritten/zeroed storage.
3. `PureZigRecordStream`'s epoch field comment still says both directions advance to application only at full completion, while this commit intentionally implements directional transitions after ServerFinished.

State the rules consistently:

- a still-pending operation abandoned by teardown is cancelled once; every operation is released once; completed or terminally errored operations are not cancelled;
- every sink payload expires before the next `start`, `receive`, **or `resumeAuth`**, including their `Outcome` variants;
- record read/write epochs transition independently as implemented.

Add a small generic-driver lifetime test showing that a payload borrowed from one call is invalidated/reset by `resumeAuthOutcome`, matching the corrected documentation.

---

The issue #334 acceptance criteria are otherwise substantially implemented, and the latest commit is a meaningful improvement. The terminal-boundary and event-loop issues above still need correction before this can safely merge.

Jaruso (commented):
Re-review at `86c250f4`: substantial improvement, but I still do not think this is safe to merge. Seven of the previous nine findings are convincingly fixed. Two fixes are incomplete, and the record-driver gating introduces one new liveness/CPU issue.

### [P1] Bytes received behind pending client selection/signing can still disappear when resume completes the local handshake

`receiveImpl` correctly appends bytes and stops dispatch while `pending_op != null`. However, `resumeAuth` calls `dispatchResume` first and drains `initial_input` / `handshake_input` only while `core.handshake_lifecycle == .running`.

For `.client_select` and `.client_sign`, a successful dispatch emits the client's Certificate/CertificateVerify/Finished and calls `completeClientHandshake`, changing the lifecycle to `.complete`. Any new Initial or Handshake bytes that arrived in a separate receive while that operation was pending are therefore skipped by both drain guards and retained forever.

Concrete sequence:

1. Client processes the server Finished and parks selecting or signing its client credential.
2. Server sends another complete Handshake message at the next QUIC CRYPTO offset.
3. `receiveImpl` accepts and buffers it because authentication is pending.
4. The operation resolves; `dispatchResume` sends ClientFinished and marks the client complete.
5. The post-resume drains are skipped because the lifecycle is no longer `.running`.
6. QUIC authenticates transport parameters and exposes an established connection without ever rejecting the extra peer handshake message.

The same-buffer Finished suffix check does not cover this separate-receive case because the legitimate server Finished was already consumed before the client operation parked.

Reject these bytes before terminal completion, or validate all buffered pre-application input before emitting `handshake_complete`. A stage-aware receive guard is one possible approach:

```zig
if (self.pending_op != null) {
    if (self.role == .client and
        (self.pending_stage == .client_select or self.pending_stage == .client_sign) and
        level != .application and bytes.len != 0)
    {
    }
    return;
}
```

Alternatively, perform a cleanup-safe buffered-input check before dispatching a client stage that can complete the handshake. Be careful not to emit success events and then return the terminal error in the same outcome.

Add direct-backend and production QUIC tests: park client selection and client signing after ServerFinished, deliver a second Finished/Certificate in a separate Handshake CRYPTO receive, then resolve the callback. The client must fail with `UnexpectedHandshakeMessage`, emit no `handshake_complete`, and release the operation/credential exactly once.

### [P1] A still-pending record authentication poll is reported as progress while read readiness advertises work the driver refuses to consume

`resumeAuthIfPending()` returns `true` after every poll, including when `poll` reports that the operation is still pending and emits no events. `drive()` therefore reports `made_progress = true` indefinitely. At the same time, `readiness()` still advertises `wants_read` because it is not pending-aware, while the new guards prevent `drive()` from reading or processing carrier input.

With a real remote verifier/HSM that remains pending until an external wake, and a peer record waiting in the socket, an event loop can spin continuously:

- socket remains readable;
- readiness continues to request reads;
- no bytes are consumed;
- the callback is polled again;
- `made_progress` is always true.

This is an externally triggerable CPU/liveness failure, not harmless backpressure.

The record owner needs an explicit pending-auth readiness state. While pending, suppress `wants_read` for work that cannot be consumed and expose `auth_pending`/a wake mechanism. `resumeAuthIfPending` should report progress only when the operation resolves, emits output, or terminates—not merely because `poll` was called.

Add a socket-pair regression with an operation that remains pending indefinitely and peer ciphertext already readable. Repeated `drive()` calls must not consume the record, must not report endless progress, and must not advertise unusable read readiness. After the test explicitly makes the operation ready and wakes/drives the stream, the buffered record must be processed normally.

### [P2] CertificateVerify still misclassifies supported-key mismatch and malformed public-key encodings as `unsupported_certificate`

The new typed proof result is a good improvement, but `checkProofOfPossession` returns `.unsupported_certificate` for several cases where the certificate type is supported:

- an Ed25519 leaf with an ECDSA CertificateVerify algorithm;
- a P-256 leaf with an Ed25519 CertificateVerify algorithm;
- an Ed25519 public key of the wrong encoded length;
- `Ed25519.PublicKey.fromBytes` failure;
- `EcdsaP256.PublicKey.fromSec1` failure.

`unsupported_certificate` is appropriate for a genuinely unsupported certificate/public-key type. A supported X.509 key paired with the wrong CertificateVerify scheme is a failed proof-of-possession; malformed public-key material is a malformed/bad certificate. Those should not be collapsed into the same alert.

Keep these paths distinct:

- certificate/SPKI parse or malformed public-key encoding -> `invalid_peer_certificate_chain` / `bad_certificate`;
- known supported key with an incompatible or non-verifying CertificateVerify signature -> `certificate_verify_invalid` / `decrypt_error`;
- genuinely unsupported public-key algorithm/curve -> `unsupported_peer_certificate` / `unsupported_certificate`;
- algorithm absent from the locally advertised list -> `IllegalParameter` (already handled).

Add server-auth and client-auth tests for Ed25519 certificate + ECDSA CertificateVerify, P-256 certificate + Ed25519 CertificateVerify, and malformed supported-key encodings, asserting the exact TLS alert for each.

CI is green, and the alert epoch, Finished verification, missing-extension handling, empty hostname rejection, QUIC pre-auth 1-RTT dropping, and wrong-kind completion cleanup all look corrected. These three remaining cases should be addressed before merge.

Jaruso (commented):
## Follow-up review — merge-ready at `b639f213`

I re-reviewed the new commit against the three blockers from the prior round, then rescanned the affected TLS record, QUIC connection, credential-lifetime, transcript, alert, and production-driver paths. I found no remaining blocking defect.

### Prior blockers

1. **Extra peer handshake input during asynchronous client selection/signing — fixed.**

   The client now rejects any additional non-application handshake bytes while `client_select` or `client_sign` is parked, cancels/releases the pending operation exactly once, and returns `UnexpectedHandshakeMessage` before the bytes enter reassembly. This closes the lifecycle hole where resume could previously emit ClientFinished and complete before accepted trailing input was examined.

   Coverage now includes both direct-backend stages and real QUIC `Connection` behavior, including the expected local `CRYPTO_ERROR(unexpected_message)` and cleanup counts.

2. **Record-mode pending-auth busy loop — fixed.**

   `resumeAuthIfPending()` now reports progress only when the operation resolves, emits events, or fails. A poll that remains pending and emits nothing returns no progress. Record readiness also suppresses carrier reads while authentication is parked, and every inbound-processing path rechecks `authStillPending()` before consuming ciphertext.

   This preserves queued ciphertext under the correct epoch and prevents a readiness-driven event loop from spinning on empty polls. The remaining integration obligation is the one already documented by the credential contract: an actually asynchronous provider wakes its owner when the operation may have progressed, and that owner calls `drive()` again.

3. **CertificateVerify failure classification — fixed.**

   The proof-of-possession path now distinguishes:

   - a supported algorithm paired with the other supported key type → `certificate_verify_invalid` / `decrypt_error`;
   - malformed supported public-key encoding → `invalid_peer_certificate_chain` / `bad_certificate`;
   - unsupported key types or curves → `unsupported_certificate`;
   - malformed or failed signatures → `decrypt_error`;
   - an algorithm outside the locally advertised supported set → `IllegalParameter`.

   The added mismatch and malformed-key tests exercise the important boundary between certificate parsing/type support and CertificateVerify cryptographic failure.

### Issue #334 acceptance criteria

The final implementation satisfies the issue requirements:

- mock providers and verifiers complete real handshakes;
- selection receives exact SNI, role, ordered peer signature schemes, ALPN, version, cipher, and policy;
- local credentials and peer verification are provider-neutral and swappable without changing handshake logic;
- peer chains are exposed as bounded immutable DER views, with overflow rejected rather than truncated;
- no OpenSSL/X.509 concrete object crosses the new interface;
- server authentication plus optional/required handshake-time client authentication work synchronously and asynchronously;
- post-handshake client authentication is explicitly deferred;
- callback, credential, verification, CertificateVerify, Finished, missing-extension, and client-certificate failures have deterministic TLS/QUIC alert mappings;
- negative and lifetime tests cover absent/incompatible credentials, rejected verification, bad signatures, malformed callback completions, cancellation, fragmentation, buffered input, and production record/QUIC drivers.

CI run 741 is green across formatting, linting, Linux/macOS/ARM tests and builds, appliance dependency audit, PKI differential checks, packaging, and performance smoke.

**Verdict: good to merge.**

PR #461, opened 2026-07-18
pki: add immutable SNI credential selection and reload

Jaruso:
Claims #347 — exclusive Codex implementation

## Summary

- Adds `src/tls/sni_provider.zig`, a reloadable in-memory SNI credential provider behind the merged #334 `CredentialProvider` / `SelectedCredential` seam.
- Publishes immutable credential snapshots with mutex-protected retain/swap, monotonic generations, selected-handle snapshot pinning, and exactly-once signer/snapshot release.
- Implements deterministic exact, longest single-label wildcard, and default precedence; absent and unknown SNI policies; ASCII hostname canonicalization; duplicate/invalid pattern rejection; and no fallback from an exact class to wildcard/default on signature incompatibility.
- Validates snapshot inputs before publication: chain bounds, DER preflight, leaf key kind, signer/key/scheme compatibility, duplicate default, duplicate patterns, default usability, allocation cleanup, and external owner release.
- Supports opaque signer adapters for snapshot-owned software `Identity`, external signer handles, and crypto-provider `SigningKey`; no OpenSSL/X.509 object or private-key bytes cross the TLS provider interface.
- Adds record-mode and QUIC coverage through the existing shared backend. QUIC uses the same provider and a pass-through client-options constructor for SNI; no QUIC-specific selector was added.

## Ownership and reload notes

- Selection retains the current snapshot under a small pointer mutex, releases the mutex before matching/signing, and transfers that retained reference to the selected handle on success.
- Reload builds the replacement off-path, publishes only after complete validation, and releases retired snapshots only after the provider and every selected handle release their references.
- Selected handles return chains from their pinned generation and reject signing attempts with any scheme other than the chosen peer-offered compatible scheme.

## Policy and exclusions

- Wildcards are limited to a complete left-most `*.` label and match exactly one DNS label.
- Unknown SNI defaults to fail; absent SNI defaults to use the configured default bundle.
- This does not implement #356 HTTP dispatch, #392 appliance provisioning, #386 negotiation-registry migration, filesystem watching, remote/HSM async signing, RSA, public trust validation, or new ClientHello parsing.

## Validation

- `git diff --check` — passed
- `zig fmt --check build.zig src/ tests/` — passed
- `zig build test-tls --summary all --error-style verbose` — passed, 278/278 tests
- `zig build test-quic --summary all --error-style verbose` — passed, 381/381 tests
- `zig build test-integration --summary all --error-style verbose` — passed, 65/69 passed / 4 skipped on clean rerun; two immediately prior runs failed only the unrelated HTTP reload/static stress test at `tests/integration.zig:5228` after it had passed earlier
- `zig build test --summary all --error-style verbose` — passed, 1551/1552 passed / 1 skipped
- `zig build test -Dtls-profile=appliance --summary all --error-style verbose` — passed, 1534/1535 passed / 1 skipped
- `zig build test -Doptimize=ReleaseFast --summary all --error-style verbose` — passed, 1551/1552 passed / 1 skipped

Closes #347

### DIFF
--- a/src/quic/tls_backend.zig
+++ b/src/quic/tls_backend.zig
@@ -22,6 +22,7 @@ pub const Identity = shared.Identity;
 pub const Trust = shared.Trust;
 pub const Entropy = shared.Entropy;
 pub const ClientOptions = shared.Tls13Backend.ClientOptions;
+pub const CredentialProvider = tls_core.credentials.CredentialProvider;
 pub const KeySchedule = shared.KeySchedule;
 pub const hash_len = shared.hash_len;
 pub const max_message_len = shared.max_message_len;
@@ -213,6 +214,14 @@ pub const Tls13Backend = struct {
 
     pub fn initServer(entropy: Entropy, identity: Identity) Tls13Backend {
         return .{ .engine = shared.Tls13Backend.initServer(entropy, identity, .{ .extension = .{
+            .alpn = "h3",
+            .extension_type = ext_quic_transport_parameters,
+            .local = "",
+        } }) };
+    }
+
+    pub fn initServerWithProvider(entropy: Entropy, provider: CredentialProvider) Tls13Backend {
+        return .{ .engine = shared.Tls13Backend.initServerWithProvider(entropy, provider, .{ .extension = .{
             .alpn = "h3",
             .extension_type = ext_quic_transport_parameters,
             .local = "",
@@ -440,18 +449,17 @@ const RealHandshakeHarness = struct {
     }
 
     fn initWithSniProvider(server_name: []const u8, provider: tls_core.credentials.CredentialProvider) !RealHandshakeHarness {
-        var harness = RealHandshakeHarness{
+        const harness = RealHandshakeHarness{
             .client_backend = Tls13Backend.initClientWithOptions(
                 .{ .hello_random = [_]u8{0xc1} ** 32, .key_share_seed = [_]u8{0x11} ** 32 },
                 .{ .pinned_certificate = testdata.certificate_der },
                 .{ .server_name = server_name },
             ),
-            .server_backend = Tls13Backend.initServer(
+            .server_backend = Tls13Backend.initServerWithProvider(
                 .{ .hello_random = [_]u8{0x51} ** 32, .key_share_seed = [_]u8{0x22} ** 32 },
-                try Identity.initPkcs8(testdata.certificate_der, testdata.private_key_pkcs8_der),
+                provider,
             ),
         };
-        harness.server_backend.engine.external_provider = provider;
         return harness;
     }
 
--- a/src/quic/tls_backend.zig
+++ b/src/quic/tls_backend.zig
@@ -21,6 +21,7 @@ const EncryptionLevel = tls_adapter.EncryptionLevel;
 pub const Identity = shared.Identity;
 pub const Trust = shared.Trust;
 pub const Entropy = shared.Entropy;
+pub const ClientOptions = shared.Tls13Backend.ClientOptions;
 pub const KeySchedule = shared.KeySchedule;
 pub const hash_len = shared.hash_len;
 pub const max_message_len = shared.max_message_len;
@@ -199,11 +200,15 @@ pub const Tls13Backend = struct {
     scratch: RecordSink = .{},
 
     pub fn initClient(entropy: Entropy, trust: Trust) Tls13Backend {
-        return .{ .engine = shared.Tls13Backend.initClient(entropy, trust, .{ .extension = .{
+        return initClientWithOptions(entropy, trust, .{});
+    }
+
+    pub fn initClientWithOptions(entropy: Entropy, trust: Trust, options: ClientOptions) Tls13Backend {
+        return .{ .engine = shared.Tls13Backend.initClientWithOptions(entropy, trust, .{ .extension = .{
             .alpn = "h3",
             .extension_type = ext_quic_transport_parameters,
             .local = "",
-        } }) };
+        } }, options) };
     }
 
     pub fn initServer(entropy: Entropy, identity: Identity) Tls13Backend {
@@ -434,6 +439,22 @@ const RealHandshakeHarness = struct {
         };
     }
 
+    fn initWithSniProvider(server_name: []const u8, provider: tls_core.credentials.CredentialProvider) !RealHandshakeHarness {
+        var harness = RealHandshakeHarness{
+            .client_backend = Tls13Backend.initClientWithOptions(
+                .{ .hello_random = [_]u8{0xc1} ** 32, .key_share_seed = [_]u8{0x11} ** 32 },
+                .{ .pinned_certificate = testdata.certificate_der },
+                .{ .server_name = server_name },
+            ),
+            .server_backend = Tls13Backend.initServer(
+                .{ .hello_random = [_]u8{0x51} ** 32, .key_share_seed = [_]u8{0x22} ** 32 },
+                try Identity.initPkcs8(testdata.certificate_der, testdata.private_key_pkcs8_der),
+            ),
+        };
+        harness.server_backend.engine.external_provider = provider;
+        return harness;
+    }
+
     fn wire(self: *RealHandshakeHarness) !void {
         self.client = tls_handshake.Handshake.initClient(&self.client_adapter, self.client_backend.backend());
         self.server = tls_handshake.Handshake.initServer(&self.server_adapter, self.server_backend.backend());
@@ -498,6 +519,44 @@ test "QUIC handshake owner tears down shared and adapter storage on success" {
     harness.deinit();
     try expectQuicBackendWiped(&harness.client_backend);
     try expectQuicBackendWiped(&harness.server_backend);
+}
+
+fn quicSniConfig(patterns: []const []const u8, chain: []const []const u8) tls_core.sni_provider.CredentialBundleConfig {
+    return .{
+        .chain = chain,
+        .patterns = patterns,
+        .signer = tls_core.sni_provider.SignAdapter.fromIdentity(Identity.initPkcs8(testdata.certificate_der, testdata.private_key_pkcs8_der) catch unreachable),
+        .key_kind = .ed25519,
+        .is_default = true,
+    };
+}
+
+test "QUIC handshake uses the shared reloadable SNI provider" {
+    var provider = tls_core.sni_provider.ReloadableProvider.init(std.testing.allocator);
+    defer provider.deinit();
+
+    const chain = [_][]const u8{testdata.certificate_der};
+    const first_patterns = [_][]const u8{"quic-one.example.test"};
+    try provider.reload(&.{quicSniConfig(first_patterns[0..], chain[0..])}, .{ .unknown_sni_policy = .fail_handshake });
+    {
+        var harness = try RealHandshakeHarness.initWithSniProvider("quic-one.example.test", provider.provider());
+        defer harness.deinit();
+        try harness.wire();
+        try harness.run();
+        try std.testing.expect(harness.client.isComplete());
+        try std.testing.expect(harness.server.isComplete());
+    }
+
+    const second_patterns = [_][]const u8{"quic-two.example.test"};
+    try provider.reload(&.{quicSniConfig(second_patterns[0..], chain[0..])}, .{ .unknown_sni_policy = .fail_handshake });
+    {
+        var harness = try RealHandshakeHarness.initWithSniProvider("quic-two.example.test", provider.provider());
+        defer harness.deinit();
+        try harness.wire();
+        try harness.run();
+        try std.testing.expect(harness.client.isComplete());
+        try std.testing.expect(harness.server.isComplete());
+    }
 }
 
 test "QUIC handshake owner tears down shared and adapter storage on failure" {
--- a/src/tls/negotiation.zig
+++ b/src/tls/negotiation.zig
@@ -227,12 +227,15 @@ fn parseKeyShares(bytes: []const u8, offers: *ClientHelloOffers) Error!void {
 
 fn parseServerName(bytes: []const u8, offers: *ClientHelloOffers) Error!void {
     var r = messages.Reader{ .bytes = bytes };
-    var names = messages.Reader{ .bytes = try r.slice(try r.u16_()) };
+    const list_len = try r.u16_();
+    if (list_len == 0) return error.MalformedExtension;
+    var names = messages.Reader{ .bytes = try r.slice(list_len) };
     try r.expectEnd();
     while (names.remaining() > 0) {
         const name_type = try names.u8_();
         const name = try names.slice(try names.u16_());
-        if (name_type == 0 and offers.server_name == null) {
+        if (name_type == 0) {
+            if (offers.server_name != null) return error.MalformedExtension;
             dns_name.validateHostName(name) catch return error.MalformedExtension;
             offers.server_name = name;
         }
@@ -403,6 +406,71 @@ test "ClientHello parser feeds policy negotiation with ALPN and SNI offers" {
     try testing.expect(selected.alpn.eql(algorithms.alpn.h2));
     try testing.expectEqualStrings("example.test", selected.server_name.?);
     try testing.expectEqualStrings("share", selected.key_share);
+}
+
+fn clientHelloWithServerNameExtension(buf: []u8, server_name_payload: []const u8) ![]const u8 {
+    var w = messages.Writer{ .buf = buf };
+    try w.u16_(algorithms.legacy_version);
+    try w.bytes(&([_]u8{0} ** 32));
+    try w.u8_(0);
+    try w.u16_(2);
+    try w.u16_(@intFromEnum(algorithms.CipherSuite.tls_aes_128_gcm_sha256));
+    try w.u8_(1);
+    try w.u8_(0);
+
+    const extensions_len = try w.reserve(2);
+    try w.u16_(@intFromEnum(algorithms.ExtensionType.supported_versions));
+    try w.u16_(3);
+    try w.u8_(2);
+    try w.u16_(@intFromEnum(algorithms.ProtocolVersion.tls13));
+
+    try w.u16_(@intFromEnum(algorithms.ExtensionType.supported_groups));
+    try w.u16_(4);
+    try w.u16_(2);
+    try w.u16_(@intFromEnum(algorithms.NamedGroup.x25519));
+
+    try w.u16_(@intFromEnum(algorithms.ExtensionType.signature_algorithms));
+    try w.u16_(4);
+    try w.u16_(2);
+    try w.u16_(@intFromEnum(algorithms.SignatureScheme.ed25519));
+
+    try w.u16_(@intFromEnum(algorithms.ExtensionType.key_share));
+    const key_share_ext = try w.reserve(2);
+    const key_shares = try w.reserve(2);
+    try w.u16_(@intFromEnum(algorithms.NamedGroup.x25519));
+    try w.u16_(5);
+    try w.bytes("share");
+    w.patch(2, key_shares);
+    w.patch(2, key_share_ext);
+
+    try w.u16_(@intFromEnum(algorithms.ExtensionType.server_name));
+    try w.u16_(@intCast(server_name_payload.len));
+    try w.bytes(server_name_payload);
+    w.patch(2, extensions_len);
+    return w.written();
+}
+
+test "ClientHello parser rejects empty SNI ServerNameList" {
+    var body: [192]u8 = undefined;
+    const empty_server_name_list = [_]u8{ 0, 0 };
+    const hello = try clientHelloWithServerNameExtension(&body, &empty_server_name_list);
+    try testing.expectError(error.MalformedExtension, parseClientHello(hello));
+}
+
+test "ClientHello parser rejects duplicate SNI host_name entries" {
+    var sni_payload: [2 + 2 * (1 + 2 + 12)]u8 = undefined;
+    var sni = messages.Writer{ .buf = &sni_payload };
+    const list_len = try sni.reserve(2);
+    inline for (0..2) |_| {
+        try sni.u8_(0);
+        try sni.u16_(12);
+        try sni.bytes("example.test");
+    }
+    sni.patch(2, list_len);
+
+    var body: [256]u8 = undefined;
+    const hello = try clientHelloWithServerNameExtension(&body, sni.written());
+    try testing.expectError(error.MalformedExtension, parseClientHello(hello));
 }
 
 test "recognized offer vectors allow general-purpose client sizes" {
--- a/src/tls/negotiation.zig
+++ b/src/tls/negotiation.zig
@@ -2,6 +2,7 @@
 
 const std = @import("std");
 const algorithms = @import("algorithms.zig");
+const dns_name = @import("dns_name.zig");
 const messages = @import("messages.zig");
 const policy_mod = @import("policy.zig");
 
@@ -231,7 +232,10 @@ fn parseServerName(bytes: []const u8, offers: *ClientHelloOffers) Error!void {
     while (names.remaining() > 0) {
         const name_type = try names.u8_();
         const name = try names.slice(try names.u16_());
-        if (name_type == 0 and offers.server_name == null) offers.server_name = name;
+        if (name_type == 0 and offers.server_name == null) {
+            dns_name.validateHostName(name) catch return error.MalformedExtension;
+            offers.server_name = name;
+        }
     }
 }
 
--- a/src/tls/sni_provider.zig
+++ b/src/tls/sni_provider.zig
@@ -168,6 +168,13 @@ pub const CredentialBundleConfig = struct {
     is_default: bool = false,
 };
 
+fn releaseConfiguredSigners(configs: []const CredentialBundleConfig) void {
+    for (configs) |config| {
+        var signer = config.signer;
+        signer.release();
+    }
+}
+
 const CredentialBundle = struct {
     chain: []const []const u8,
     patterns: []const HostPattern,
@@ -229,10 +236,7 @@ pub const Snapshot = struct {
     ) BuildError!*Snapshot {
         var consumed: usize = 0;
         errdefer {
-            for (configs[consumed..]) |config| {
-                var signer = config.signer;
-                signer.release();
-            }
+            releaseConfiguredSigners(configs[consumed..]);
         }
 
         if (configs.len == 0) return error.EmptyCredentialSet;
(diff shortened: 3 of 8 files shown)


src/tls/sni_provider.zig
```diff
@@ -0,0 +1,1137 @@
...
+    var patterns = allocator.alloc(HostPattern, config.patterns.len) catch return error.OutOfMemory;
+    var copied_patterns: usize = 0;
+    errdefer {
+        for (patterns[0..copied_patterns]) |pattern| freePattern(allocator, pattern);
+        allocator.free(patterns);
+    }
+    for (config.patterns, 0..) |raw, i| {
+        patterns[i] = try parseHostPattern(allocator, raw);
+        copied_patterns += 1;
+        if (hasDuplicatePattern(previous, patterns[i]) or hasDuplicatePatternIn(patterns[0..i], patterns[i]))
```
Jaruso: **P1 — This prevents signature algorithms from selecting among credentials for the same SNI.**

The authoritative #347 plan says that after resolving the hostname class, the provider should inspect matching bundles in stable installation order and choose the first one compatible with the peer's offered signature schemes. With one chain/key per bundle, supporting both Ed25519 and P-256 for `api.example.test` necessarily requires two bundles carrying the same exact pattern (and the same is true for a wildcard class). This cross-bundle duplicate check rejects that configuration with `DuplicateHostPattern`, so the implementation cannot satisfy the issue's core signature-aware selection requirement.

The selection loop already has the correct deterministic behavior. I would keep duplicate rejection within a single bundle, but allow the same normalized pattern across bundles:

```zig
for (config.patterns, 0..) |raw, i| {
    patterns[i] = try parseHostPattern(allocator, raw);
    copied_patterns += 1;
    if (hasDuplicatePatternIn(patterns[0..i], patterns[i]))
        return error.DuplicateHostPattern;
}
```

Then add a regression with two `api.example.test` bundles, one Ed25519 and one P-256, proving peer offer order selects the compatible bundle in stable install order. Also strengthen the no-fallback test: its wildcard bundle is currently Ed25519 too, so offering only ECDSA makes both exact and wildcard incompatible and does not prove that a *compatible* wildcard is refused when an exact class exists.

src/tls/sni_provider.zig
```diff
@@ -0,0 +1,1137 @@
...
+        self.allocator.free(self.bundles);
+        self.bundles = &.{};
+    }
+
+    fn select(self: *Snapshot, allocator: std.mem.Allocator, selection: *const credentials.SelectionContext) credentials.SelectError!credentials.SelectedCredential {
+        const class = self.resolveClass(selection.server_name) orelse return error.NoCredentialAvailable;
+        for (self.bundles, 0..) |*bundle, index| {
+            if (!bundle.matchesPattern(class)) continue;
+            const scheme = chooseCompatibleScheme(bundle, selection.peer_signature_schemes) orelse continue;
+            const handle = allocator.create(SelectedHandle) catch return error.OutOfMemory;
```
Jaruso: **P1 — The selected handle can be freed through a different allocator than the one that allocated it.**

This line allocates with the allocator passed by `ReloadableProvider.select` (`self.allocator`), but `SelectedHandle.release` later calls `snapshot.allocator.destroy(self)`. Normally `reload()` builds with the provider allocator, but the public API also permits:

```zig
const snapshot = try Snapshot.build(allocator_b, configs, options, 1);
provider_using_allocator_a.install(snapshot);
```

Selection allocates the handle from allocator A and release destroys it through allocator B, which is allocator corruption/undefined behavior.

The simplest fix is to make allocator provenance intrinsic to the snapshot and remove the extra allocator parameter:

```zig
fn select(
    self: *Snapshot,
    selection: *const credentials.SelectionContext,
) credentials.SelectError!credentials.SelectedCredential {
    // ...
    const handle = self.allocator.create(SelectedHandle)
        catch return error.OutOfMemory;
    // ...
}
```

Alternatively store the actual handle allocator in `SelectedHandle` and use that during release. Please add a test that builds a snapshot with one allocator, installs it into a provider constructed with another, selects, and releases the credential.

Jaruso (commented):
**Review verdict: not merge-ready.**

I found two merge-blocking correctness issues in the new provider.

1. The snapshot builder rejects the configuration required for signature-aware credential selection. Because every bundle has exactly one chain/key kind and the current supported key kinds each map to one TLS scheme, serving both Ed25519 and ECDSA P-256 for the same SNI requires two bundles with the same normalized host pattern. The cross-bundle duplicate rejection makes that impossible, so peer signature algorithms can only accept/reject a host's sole credential rather than influence which credential is selected. This conflicts with #347's objective and the authoritative implementation plan's rule to inspect bundles in stable installation order within the resolved hostname class.

2. `SelectedHandle` allocation and destruction can use different allocators. `Snapshot.select` allocates with the provider allocator supplied by `ReloadableProvider.select`, while `SelectedHandle.release` destroys with `snapshot.allocator`. Since both `Snapshot.build` and `ReloadableProvider.install` are public, a snapshot built with allocator B can be installed into a provider using allocator A, producing an invalid free on release.

CI is green and the exact/wildcard/default and reload mechanics are otherwise thoughtfully structured, but these two issues need correction before merge. Please add focused regression tests for multi-key selection on one exact/wildcard hostname and for installing a snapshot built with a different allocator.

src/tls/sni_provider.zig
```diff
@@ -0,0 +1,1205 @@
...
+        errdefer {
+            for (bundles[0..initialized]) |*bundle| bundle.deinit(allocator);
+            allocator.free(bundles);
+        }
+
+        var default_count: usize = 0;
+        for (configs, 0..) |config, index| {
+            bundles[index] = try buildBundle(allocator, config, index);
+            initialized += 1;
+            if (bundles[index].is_default) default_count += 1;
```
Jaruso, line 269: **P1 — A failed multi-bundle build has indeterminate signer ownership and can leak unreached signers.**

`buildBundle()` consumes/releases the signer for the bundle it attempts, and the `bundles[0..initialized]` cleanup releases earlier successful bundles. But any configs after the first failing bundle are never visited or released. Failures before the loop—`configs.len > max_bundles`, snapshot allocation OOM, or bundle-array allocation OOM—release no signer at all. The existing single-config test therefore proves only one failure position.

This makes exactly-once ownership impossible for callers: after `Snapshot.build()` returns an error, they cannot know which external handles were already released, especially under allocation-failure injection, so retrying or cleaning up can either leak or double-release.

Choose one consistent contract. The current tests imply that the call consumes **all** configured signers on both success and failure. In that model, install an outer cleanup before the first possible error and track the first signer not already transferred to/consumed by `buildBundle()`:

```zig
var consumed: usize = 0;
errdefer {
    for (configs[consumed..]) |config| {
        var signer = config.signer;
        signer.release();
    }
}

// ...all validation/allocation...
for (configs, 0..) |config, index| {
    consumed = index + 1; // buildBundle consumes this signer even on error
    bundles[index] = try buildBundle(allocator, config, index);
    initialized += 1;
}
```

That sketch also requires `release()` to be safe for every accepted config shape (see the SigningKey comment). Add regressions with three independently counted external signers where the first and middle bundles fail, plus allocation-failure coverage asserting every callback runs exactly once.

src/tls/sni_provider.zig
```diff
@@ -0,0 +1,1205 @@
...
+            const allocator = self.allocator;
+            self.deinit();
+            allocator.destroy(self);
+        }
+    }
+
+    fn deinit(self: *Snapshot) void {
+        self.deinit_count.* += 1;
+        for (self.bundles) |*bundle| bundle.deinit(self.allocator);
+        self.allocator.free(self.bundles);
```
Jaruso, line 298: **P1 — This test hook creates a production data race during concurrent retirement.**

Every normal snapshot points `deinit_count` at the same mutable global `noop_deinit_count`, and `deinit()` performs a non-atomic `+= 1`. The provider otherwise supports concurrent generations with atomic refcounts: selected handles from two retired snapshots can drop their final references on different threads at the same time, causing concurrent writes to this global and undefined behavior.

Please remove test-only destruction accounting from the production snapshot. A safe minimal alternative is an optional atomic hook used only by tests:

```zig
deinit_count: ?*std.atomic.Value(usize) = null,

fn deinit(self: *Snapshot) void {
    if (self.deinit_count) |counter| {
        _ = counter.fetchAdd(1, .monotonic);
    }
    // ...
}
```

The cleaner test is to use per-snapshot external-signer release counters or debug allocators, avoiding a mutable test field on the supposedly immutable published snapshot entirely. Add a concurrent retirement regression or at least make the production path contain no shared non-atomic state.

src/tls/sni_provider.zig
```diff
@@ -0,0 +1,1205 @@
...
+            return switch (self.options.absent_sni_policy) {
+                .use_default => .default,
+                .fail_handshake => null,
+            };
+        };
+        if (raw_name.len == 0 or raw_name.len > max_host_pattern_len) return null;
+        var name_buf: [max_host_pattern_len]u8 = undefined;
+        const name = lowerHostInto(raw_name, &name_buf) catch return null;
+
+        for (self.bundles) |bundle| {
```
Jaruso, line 329: **P1 — Malformed DNS SNI is still collapsed into `NoCredentialAvailable`, contrary to the #347 contract.**

`lowerHostInto()` rejects invalid DNS syntax, but this `catch return null` makes malformed input indistinguishable from a valid unknown hostname. The ClientHello parser currently checks only SNI framing, duplication, emptiness, and its 256-byte buffer bound; it copies names such as `bad..example`, `bad host`, embedded NULs, over-253-byte DNS names, or labels with invalid boundaries and still calls the provider. `select()` then reports `NoCredentialAvailable` rather than the parser producing the distinct malformed-SNI alert required by the issue.

The authoritative plan says malformed SNI must fail during ClientHello parsing **before the provider is called**. Move one shared ASCII DNS-host validator to a neutral TLS module and use it for both configured patterns and parsed `host_name`. Once that invariant is real, this path should not silently reinterpret validation failure as an unknown host.

Add engine tests asserting invalid DNS names fail with the chosen malformed/illegal-parameter outcome and that a mock provider's `select_count` remains zero. Include empty labels, spaces/NUL, labels over 63 bytes, total names over 253 bytes, and leading/trailing hyphens.

src/tls/sni_provider.zig
```diff
@@ -0,0 +1,1205 @@
...
+
+    pub fn reload(self: *ReloadableProvider, configs: []const CredentialBundleConfig, options: SnapshotOptions) BuildError!void {
+        const replacement = try self.buildSnapshot(configs, options);
+        self.install(replacement);
+    }
+
+    pub fn install(self: *ReloadableProvider, replacement: *Snapshot) void {
+        self.mutex.lock();
+        const retired = self.current;
+        self.current = replacement;
```
Jaruso, line 424: **P1 — Concurrent reloads can publish an older credential generation after a newer one.**

Generation assignment and publication are separate critical sections. A slow reload A can reserve generation 1, reload B can reserve/build/install generation 2, and then A can finish and overwrite `current` with generation 1. New handshakes then roll back to stale certificates even though the implementation claims monotonic generations.

Do not hold the tiny selection-pointer mutex while building, but serialize reload operations with a separate reload mutex, or reject stale replacements under this mutex:

```zig
self.mutex.lock();
if (self.current) |current| {
    if (replacement.generation <= current.generation) {
        self.mutex.unlock();
        replacement.release();
        return; // or typed StaleSnapshot
    }
}
const retired = self.current;
self.current = replacement;
self.mutex.unlock();
if (retired) |snapshot| snapshot.release();
```

Also tighten the public ownership surface: reinstalling the exact current pointer currently assigns it and then releases it, leaving `current` dangling. If external installation is not required, make `Snapshot`, `buildSnapshot`, and `install` private implementation details. Add a deterministic regression that constructs generations 1 and 2, publishes 2 first, attempts to publish 1, and proves new selections remain on 2.

src/tls/sni_provider.zig
```diff
@@ -0,0 +1,1205 @@
...
+    errdefer allocator.free(schemes);
+    if (schemes.len == 0) return error.NoSupportedSignatureScheme;
+    for (schemes) |scheme| {
+        if (!schemeLegalForKeyKind(config.key_kind, scheme)) return error.KeySignerMismatch;
+    }
+
+    switch (signer) {
+        .external => {},
+        .signing_key => |adapter| {
+            if (adapter.release != null and adapter.release_context == null) return error.InvalidOwnershipContract;
```
Jaruso: **P1 — `SigningKeyAdapter` build validation is incomplete, and the invalid-ownership path can panic instead of returning `InvalidOwnershipContract`.**

Two problems:

1. When `supported_schemes` is explicitly supplied, nothing compares `adapter.key.scheme()` with the declared scheme/key kind. An ECDSA `SigningKey` can be installed behind an Ed25519 leaf with `supported_schemes = &.{.ed25519}`; build succeeds, selection returns Ed25519, and signing fails later with `InvalidCallbackBehavior`. That violates the required build-time signer/key/scheme compatibility check.
2. A config with `release != null` and `release_context == null` reaches this return, but the function-level `errdefer signer.release()` then executes `adapter.release_context.?` and traps. The advertised typed validation error is therefore unreachable without a panic.

Model ownership so the invalid state is unrepresentable, for example:

```zig
ownership: union(enum) {
    borrowed,
    owned: struct {
        context: *anyopaque,
        release: *const fn (*anyopaque) void,
    },
},
```

Then validate the actual key scheme:

```zig
const actual = providerToTlsScheme(adapter.key.scheme())
    orelse return error.NoSupportedSignatureScheme;
if (!schemeLegalForKeyKind(config.key_kind, actual) or
    !containsScheme(schemes, actual))
    return error.KeySignerMismatch;
```

Because the provider `SigningKey` exposes one concrete scheme, explicit metadata must not be allowed to contradict it. Add a fake-key regression for ECDSA-key/Ed25519-metadata mismatch and a release-function/null-context regression proving build returns a typed error without invoking an invalid callback or panicking.

Jaruso (commented):
Re-review of `575433a0`:

The two findings from the previous review are correctly fixed: same-pattern multi-key selection now works in stable bundle order, the no-fallback regression uses a genuinely compatible wildcard, and selected handles now allocate/free through the snapshot allocator.

However, this is still **not merge-ready**. The inline comments identify five additional P1 defects in ownership, concurrency, signer validation, and malformed-SNI classification.

There is also one architecture/AC blocker that cannot be anchored cleanly to a changed line: **credential selection still happens after ServerHello has already been recorded and emitted.** `onClientHello()` emits ServerHello, derives/emits handshake secrets, and discards Initial keys before calling `sendServerFlight()`, where `selectCredential()` finally runs. The #347 contract explicitly requires selection during ClientHello processing *before ServerHello/certificate-flight emission*. Unknown SNI, incompatible signature algorithms, provider OOM/failure, or a pending selector therefore occur only after the server has committed a ServerHello.

The server path should be reordered so the parsed immutable selection context is resolved first. A synchronous selection can then proceed to ServerHello and the auth flight; a pending selection must park before any server bytes or key-discard events are emitted, and resume by emitting ServerHello exactly once with the pinned credential. Please add an integration assertion that an unknown/malformed SNI or pending selector emits no ServerHello before selection resolves.

The issue discussion also explicitly asks for an integration test where a handshake has selected/pinned generation N, reload publishes N+1, that in-flight handshake completes using N, and a later handshake observes N+1. The direct selected-handle unit test proves the refcount mechanism, but not the complete engine/record or QUIC lifecycle requested by the AC.

src/tls/sni_provider.zig
```diff
@@ -0,0 +1,1359 @@
...
+        self.mutex.unlock();
+        return Snapshot.build(self.allocator, configs, options, generation);
+    }
+
+    pub fn reload(self: *ReloadableProvider, configs: []const CredentialBundleConfig, options: SnapshotOptions) BuildError!void {
+        const replacement = try self.buildSnapshot(configs, options);
+        self.install(replacement);
+    }
+
+    pub fn install(self: *ReloadableProvider, replacement: *Snapshot) void {
```
Jaruso: **P1 — A public externally generated snapshot can permanently turn later `reload()` calls into silent no-ops.**

`Snapshot.build` is public and accepts an arbitrary `generation`, while this provider's `next_generation` is not synchronized when `install` accepts such a snapshot. For example:

```zig
var provider = ReloadableProvider.init(allocator);
const external = try Snapshot.build(allocator, configs_a, .{}, 100);
provider.install(external); // current generation = 100; next_generation is still 1
try provider.reload(configs_b, .{}); // builds generation 1, drops it as stale, returns success
```

Every subsequent `reload()` reports success while leaving generation 100 active until the internal counter eventually passes it. That violates the public reload contract and can leave old credentials serving indefinitely.

The cleanest fix is to make raw `Snapshot.build`, `buildSnapshot`, and `install` non-public implementation details so all production generations originate from this provider. If external installation is required, update `next_generation` under the same lock to at least `replacement.generation + 1` (handling overflow), and make publication failure observable rather than silently returning success. Add a regression that installs an externally built high-generation snapshot and then verifies the next successful `reload()` is actually selected.

src/quic/tls_backend.zig
```diff
@@ -434,6 +439,22 @@ const RealHandshakeHarness = struct {
...
+                .{ .hello_random = [_]u8{0xc1} ** 32, .key_share_seed = [_]u8{0x11} ** 32 },
+                .{ .pinned_certificate = testdata.certificate_der },
+                .{ .server_name = server_name },
+            ),
+            .server_backend = Tls13Backend.initServer(
+                .{ .hello_random = [_]u8{0x51} ** 32, .key_share_seed = [_]u8{0x22} ** 32 },
+                try Identity.initPkcs8(testdata.certificate_der, testdata.private_key_pkcs8_der),
+            ),
+        };
+        harness.server_backend.engine.external_provider = provider;
```
Jaruso: **P1 — This test bypasses the QUIC server's public construction API instead of proving production integration.**

The wrapper exposes only `initServer(entropy, identity)`, so this helper creates and retains an unrelated fixed private key and then mutates the nested shared engine's `external_provider` field. A production QUIC caller has no supported constructor that expresses “this server is backed by a `CredentialProvider`.”

Please add the QUIC equivalent of the shared constructor and use it here:

```zig
pub fn initServerWithProvider(
    entropy: Entropy,
    provider: tls_core.credentials.CredentialProvider,
) Tls13Backend {
    return .{ .engine = shared.Tls13Backend.initServerWithProvider(
        entropy,
        provider,
        .{ .extension = .{
            .alpn = "h3",
            .local = "",
        } },
    ) };
}
```

That removes the dummy identity/key and makes the passing QUIC handshake test exercise the API real users will call.

src/tls/negotiation.zig
```diff
@@ -231,7 +232,10 @@ fn parseServerName(bytes: []const u8, offers: *ClientHelloOffers) Error!void {
     while (names.remaining() > 0) {
         const name_type = try names.u8_();
         const name = try names.slice(try names.u16_());
-        if (name_type == 0 and offers.server_name == null) offers.server_name = name;
+        if (name_type == 0 and offers.server_name == null) {
+            dns_name.validateHostName(name) catch return error.MalformedExtension;
```
Jaruso, line 239: **P2 — This public ClientHello parser still accepts structurally malformed SNI that the production backend rejects.**

DNS validation covers the first `host_name`, but:

- `ServerNameList` length zero falls through as an absent SNI;
- a second `host_name` is silently ignored because `offers.server_name != null`.

The production backend correctly rejects both cases, and RFC 6066 permits each `name_type` at most once. Please reject an empty list and return `MalformedExtension` when another type-0 entry is encountered, then add direct `parseClientHello` regressions so the two public parsing paths cannot drift again.

src/tls/tls13_backend_tests.zig
```diff
@@ -871,6 +872,207 @@ test "record stream completes a real TLS 1.3 handshake over a nonblocking socket
...
+    var server_sink = DirectSink{};
+    defer server_sink.deinit();
+
+    try client.backend().start(.client, {}, &client_sink);
+    var client_hello_buf: [1024]u8 = undefined;
+    const client_hello = firstInitialCrypto(&client_sink, &client_hello_buf);
+    try server.backend().start(.server, {}, &server_sink);
+    try server.backend().receive(.initial, client_hello, &server_sink);
+
+    try provider.reload(&.{sniIdentityConfig(&.{"pin.example.test"}, chain_two[0..], true)}, .{});
```
Jaruso, line 1034: **P1 — This does not actually reload while generation N is pinned by an in-flight handshake.**

`server.backend().receive(...)` is synchronous here. Before it returns, selection, Certificate emission, signing, `serverFinishFlight`, and `SelectedCredential.release()` have already completed. The reload on this line therefore happens only after the old snapshot handle has been released. Feeding the already-serialized old flight afterward proves only that copied output bytes remain unchanged.

The test also never sends the client's Finished back to the first server, so it does not prove that the first handshake completes after the reload. The issue contract explicitly asks for generation N to remain valid for the in-flight handshake while N+1 becomes active for later handshakes.

Please create a deterministic suspend point while the SNI selected handle is still live—for example, a synchronized blocking external signer running the server receive on a worker thread—then:

1. wait until generation N has selected and entered signing;
2. publish generation N+1;
3. release the signer and drive both sides to full handshake completion using N;
4. start and fully complete a later handshake and assert it receives N+1;
5. assert the retired N snapshot is released only after the first selected handle is done.

Jaruso (commented):
Re-review of `a46277e35c6daf118db13a9d95c5e34c02ca5bc9`.

The previous six blockers are materially addressed: failed-build signer ownership is now deterministic, the test counter is atomic and opt-in, malformed DNS SNI is rejected before selection in the shared backend, stale concurrent generations cannot roll the provider backward, `SigningKeyAdapter` validates its actual scheme without a null-context trap, and credential selection now precedes ServerHello for both synchronous and pending selectors.

I still do not consider this merge-ready because four areas remain:

1. The public `Snapshot.build` / `ReloadableProvider.install` generation surface can make later `reload()` calls silently succeed without publishing anything; see the inline comment.
2. Outbound `ClientOptions.server_name` still validates only empty/256-byte overflow. It can emit `bad..example`, leading/trailing-hyphen labels, embedded invalid bytes, or a 254–256 byte hostname that the newly shared server validator rejects. Please run `dns_name.validateHostName` during client option application and fail at `start` before emitting ClientHello. Expand the existing invalid-configured-name test beyond only empty and 257-byte values. The current “maximum SNI” serializer test also directly injects a 256-byte non-DNS name; retain a serializer-bound test if useful, but the public constructor should use the semantic 253-byte DNS bound.
3. The QUIC integration test reaches the provider only by constructing a server with an unrelated fixed private key and mutating `engine.external_provider`; see the inline comment. This needs a supported QUIC `initServerWithProvider` path.
4. The required in-flight reload integration test still does not reload while the generation-N selected handle is live or complete the first server handshake; see the inline comment.

There is also one parser-consistency issue: the public `negotiation.parseClientHello` path now performs DNS validation but still accepts an empty `ServerNameList` and silently ignores a second `host_name`, unlike the production backend. That should be made consistent and regression-tested.

CI jobs visible at review time are green, but the workflow run is still reported in progress.

src/tls/sni_provider.zig
```diff
@@ -0,0 +1,1399 @@
...
+        self.mutex.lock();
+        const retired = self.current;
+        self.current = null;
+        self.mutex.unlock();
+        if (retired) |snapshot| snapshot.release();
+    }
+
+    pub fn buildSnapshot(self: *ReloadableProvider, configs: []const CredentialBundleConfig, options: SnapshotOptions) ReloadError!*Snapshot {
+        self.mutex.lock();
+        if (self.next_generation == std.math.maxInt(u64)) {
```
Jaruso, line 384: **P2 — Generation exhaustion bypasses the failed-reload signer cleanup contract.**

Every other `reload()` build failure enters `Snapshot.build()`, whose outer `errdefer` consumes/releases every configured signer. This newly added early return happens before `Snapshot.build()`, so the same API has different ownership behavior for `GenerationOverflow`.

The path is reachable through the supported public external-install API:

```zig
const terminal = try Snapshot.build(
    allocator,
    existing_configs,
    .{},
    std.math.maxInt(u64) - 1,
);
try provider.install(terminal); // next_generation becomes maxInt(u64)

var owner = CountingExternal{};
const replacement = countedExternalConfig(
    &owner,
    &.{"replacement.example.test"},
    true,
);
try expectError(
    error.GenerationOverflow,
    provider.reload(&.{replacement}, .{}),
);
// owner.release_count is currently 0, unlike every other failed reload.
```

For an HSM/session-backed external signer this leaks the owned handle, and callers cannot safely compensate because all other `reload` errors already release it.

Please keep one ownership rule by releasing all configured signers before returning this pre-build error, ideally through a shared helper used by `Snapshot.build` and this branch:

```zig
fn releaseConfiguredSigners(configs: []const CredentialBundleConfig) void {
    for (configs) |config| {
        var signer = config.signer;
        signer.release();
    }
}
```

Add the max-generation regression above and assert the callback runs exactly once. Alternatively, make generation allocation unable to fail before `Snapshot.build` has taken ownership, but the important invariant is that every `reload()` outcome has deterministic signer ownership.

Jaruso (commented):
Re-review of `6d5199f`:

The five findings from the previous round are correctly addressed:

- externally installed generations now advance `next_generation`, stale publication is observable, and the regression proves the next reload becomes active;
- configured client SNI uses the shared DNS validator and fails before ClientHello emission;
- QUIC exposes and tests a real `initServerWithProvider` constructor;
- the record integration test now reloads while generation N is genuinely pinned inside signing, completes that handshake with N, and proves a later handshake observes N+1;
- the standalone negotiation parser now rejects empty and duplicate `host_name` lists.

The fresh CI run is green. I found one remaining P2 ownership edge in the new generation-exhaustion path, described inline. It is narrowly scoped, but it violates the otherwise consistent rule that a failed reload releases every configured owned signer exactly once. I would close that edge before merging, then I do not see another blocker in the reviewed scope.

Jaruso (commented):
Re-reviewed head `163f816a1b5fc1256efb24dc72ad5342e804faaf` against #347 and all findings from the prior rounds.

The generation-exhaustion ownership issue is fixed correctly. `buildSnapshot()` now releases every configured signer after dropping the provider mutex when no further generation can be reserved, and the shared helper preserves the same all-or-nothing ownership contract used by `Snapshot.build()`. The regression reaches exhaustion through the public external-install/reload path and verifies exactly one release.

I also re-checked the previously corrected areas: signature-aware same-host selection, allocator provenance, build-failure signer ownership, concurrent snapshot retirement, malformed-SNI parsing, monotonic publication, signer/key validation, selection before ServerHello, in-flight reload pinning, outbound SNI validation, QUIC provider construction, and the standalone negotiation parser. I did not find a remaining correctness or acceptance-criteria gap.

No findings. This is merge-ready from my review once the current CI run finishes successfully. At the time of this review, the unit tests, PKI differential, format, lint, builds, security/dependency checks, and performance jobs shown were green; packaging and the final dependency-audit steps were still in progress.

PR #470, opened 2026-07-18
tls: dispatch negotiated ALPN to HTTP/1.1 and HTTP/2

Jaruso:
Claims #356 — exclusive Codex implementation

Draft opened after the first meaningful commit per issue instructions.

Current implementation:
- adds a transport-neutral ALPN mapping and immutable listener protocol-policy snapshot helper;
- adds an EncryptedStream-backed HTTP connection adapter using drive(), readiness(), and bufferSnapshot();
- pins listener protocol policy per accepted TLS connection and uses that pinned snapshot for final dispatch validation;
- rejects absent ALPN during ClientHello unless explicit HTTP/1.1 fallback is enabled, and rejects malformed/no-overlap ALPN with no_application_protocol;
- plumbs h1-only, h2-only, and dual-protocol listener policy from config, rejecting both disabled;
- updates TLS listener protocol policy coherently on hot reload while preserving in-flight pinned snapshots;
- rejects plaintext h2-only configs and guards the plaintext HTTP/1.1 parser path when HTTP/1.1 is disabled;
- routes OpenSSL and pure-Zig downstream TLS through the shared negotiated HTTP runtime dispatch helper;
- flushes EncryptedStream-backed HTTP writes through the carrier before parking synchronous HTTP handlers;
- adds production native-listener integration coverage for HTTP/1.1 keepalive, HTTP/2 SETTINGS/HEADERS/response frames, and unsupported ALPN rejection before HTTP dispatch;
- keeps appliance-profile coverage OpenSSL-free with a P-256 native TLS fixture;
- adds upstream ALPN policy coverage for required h1, required h2, auto-to-h2, auto-to-h1, absent ALPN, and wrong-selection failures.

Validation:
- git diff --check
- zig fmt --check build.zig src/ tests/
- zig build test-tls --summary all --error-style verbose
- zig build test-quic --summary all --error-style verbose
- zig build test-integration --summary all --error-style verbose
- zig build test-integration-native-tls -Dtls-profile=appliance --summary all --error-style verbose
- zig build test-integration -Dtls-profile=appliance --summary all --error-style verbose
- zig build test --summary all --error-style verbose
- zig build test -Dtls-profile=appliance --summary all --error-style verbose
- zig build -Doptimize=ReleaseFast --summary all --error-style verbose
- zig build -Doptimize=ReleaseFast -Dtls-profile=appliance --summary all --error-style verbose

Build-step note:
- zig build test-http --summary all --error-style verbose is not available in the current build graph (no step named test-http).

Refs #355

PR #472, opened 2026-07-19
tls(bare): implement fixed-profile appliance credential provider

Jaruso:
Implements the complete fixed-profile Bare Systems appliance credential provider per the maintainer implementation plan on #392: a strict provisioning loader and lifecycle owner feeding the **existing** `sni_provider.ReloadableProvider` / `credentials.CredentialProvider` contract, shared by native TCP TLS and native HTTP/3, with full startup/`tardi check`/`tardi run` preflight, certificate-chain coherence validation, and deterministic failure classes.

Closes #392

## 1. Acceptance-criteria map

| #392 criterion | Implementation | Tests |
| --- | --- | --- |
| Pure-Zig TLS 1.3 handshake with the provisioned Ed25519 chain/key | `src/tls/appliance_credentials.zig` publishes one bundle into `sni_provider.ReloadableProvider`; `edge_gateway` borrows the provider for `NativeTlsConnection.create` | `tests/integration.zig` "native TLS listener …" (appliance profile, Ed25519 fixture) |
| Client validates the sent chain and accepts Ed25519 CertificateVerify | Pure-Zig TLS client (record mode) and pinned-leaf QUIC client complete handshakes against the provider; an out-of-process `openssl s_client` trusting only a generated root CA validates a transmitted leaf+intermediate chain and hostname | native h1 keepalive + h2 dispatch tests; `quic_h3_udp_smoke` "appliance credential provider authenticates native QUIC/H3"; "an independent client trusting only the root validates the transmitted leaf+intermediate chain and hostname" |
| Missing/malformed/mismatched/unsupported/incoherent inputs fail before HTTP dispatch | Full validation (including certificate-chain coherence and SAN binding, see §4) before any listener binds; typed errors propagate unwrapped through `edge_gateway.run`; unknown SNI / no-Ed25519 peer fail at selection | ~90 loader unit tests; startup-failure integration tests (mismatched key, unsupported key algorithm, direct/daemon/master-mode); unknown-SNI test asserts zero upstream requests |
| Private key never logged, zeroized per #372 | `BoundedSecret`/`FixedSecret` for file bytes, DER, seed; `SoftwareSigningKey.fromSeedSecret` bridges the seed through one wiped scope and clears the caller's secret before key derivation; snapshot signer release wipes the long-lived key; logs carry only error class names | secret-lifecycle tests (success and every error path, including the file-API oversized-key probe buffer and the `fromSeedSecret` call-boundary clear), allocation-failure sweep, `sni_provider` release-path tests |
| Appliance artifact links no OpenSSL/libcrypto | Unchanged `-Dtls-profile=appliance` build graph; new code links nothing foreign | `otool -L` evidence below; `tardi version` self-reports `tls-profile=appliance, tls-backend=native`; CI now runs a dedicated 3-OS "Test appliance profile" job |
| ALPN `h2` + `http/1.1` integration; same identity usable by native HTTP/3 | Listener protocol-policy snapshot from #470 unchanged; H3 runtime borrows the same provider instance; the Certificate-flight size preflight reuses the handshake writer's own exported framing constants, proven writer-identical by tests driving the real writer at that exact boundary for both transports | native h1/h2/ALPN-reject tests on the Ed25519 appliance identity; QUIC/H3 shared-provider tests; writer-boundary tests in `tls13_backend_tests.zig` |

## 2. Exact provisioning format

- **Chain file**: strict RFC 7468 PEM; one or more `CERTIFICATE` blocks, leaf first; ≤ 8 entries; only certificate blocks + ASCII whitespace (no prose, comments, or unrelated labels); strict base64; each block exactly one fully consumed DER certificate; the leaf must carry canonical RFC 8410 Ed25519 SPKI (no AlgorithmIdentifier parameters, 0 unused BIT STRING bits, 32 bytes). No raw-DER auto-detection.
- **Key file**: exactly one `PRIVATE KEY` block, whitespace-only outside; unencrypted PKCS#8/RFC 5958 with the RFC 8410 Ed25519 AlgorithmIdentifier (absent parameters) wrapping exactly one 32-byte seed (`OCTET STRING { OCTET STRING(32) }`); attributes/public-key/trailing DER rejected. `EC/RSA/ENCRYPTED/OPENSSH PRIVATE KEY`, raw DER, multiple blocks, and prose are rejected. File and byte APIs share one parser (file API only does bounded reads into typed secrets).

Documented in `docs/BARE_APPLIANCE_TLS.md`.

## 3. Error taxonomy

```
MissingCertificateChain, MissingPrivateKey, CertificateFileTooLarge, PrivateKeyFileTooLarge,
EmptyCertificateChain, TooManyCertificates, MalformedCertificatePem, AmbiguousCertificateInput,
CertificateTooLarge, MalformedCertificateDer, MalformedPrivateKeyPem, AmbiguousPrivateKeyInput,
MalformedPrivateKeyDer, UnsupportedPrivateKeyAlgorithm, UnsupportedPrivateKeyParameters,
InvalidPrivateKeySize, InvalidPrivateKey, UnsupportedLeafKeyAlgorithm, UnsupportedLeafKeyParameters,
KeyCertificateMismatch, CertificateFlightTooLarge, InvalidServerName, CertificateNameMismatch,
IntermediateNotCa, InvalidLeafCertificate, CertificateKeyUsageViolation,
CertificateExtendedKeyUsageViolation, UnhandledCriticalCertificateExtension,
DuplicateCertificateEntry, InvalidCertificateChainOrder, CertificateSignatureInvalid,
UnsupportedApplianceConfiguration, ProviderPublicationFailed, FileNotFound, AccessDenied, OutOfMemory
```

Every one of these is a distinct, operator-actionable class — nothing collapses into `InvalidPrivateKey` or a generic bootstrap error. `main.isConfigValidationError` classifies every class above except `OutOfMemory`/`ProviderPublicationFailed` (true internal/resource failures) as a configuration error, so `tardi check` and `tardi run` report "configuration validation failed" with exit code 2 for every operator-input mistake, including the certificate-chain-coherence and SAN/SPKI classes.

`tardi run`'s classification is now phase-scoped rather than name-scoped (`main.zig`: `prepareRunPlan` / `runPlan`): `edge_config.validate` + the appliance credential preflight run first and their failures use the configuration classifier; the daemon fork, PID file, master/worker startup, and the long-running `edge_gateway.run` itself run under a *separate* catch that always reports exit code 1 regardless of the error's name, so a post-start runtime `FileNotFound`/`AccessDenied` (e.g. unrelated filesystem work) is never mislabeled as a configuration failure. A test drives this with an unwritable PID-file directory.

## 4. Key/certificate match and chain-coherence procedure

1. Parse **every** chain entry with the existing pure-Zig `pki.x509.Certificate.parse` (no second X.509/SPKI parser).
2. Leaf: require `subject_public_key_info.key_type == .ed25519`, canonical AlgorithmIdentifier (`algorithm.parameters_raw == null`), BIT STRING `unused_bits == 0`, exactly 32 key bytes; must not assert `basicConstraints CA:TRUE`; if present, `keyUsage.digitalSignature` and `extendedKeyUsage` allowing `serverAuth` are required.
3. Leaf SAN is bound to the configured `tls_server_name` via `pki.identity.verifyHost` (SAN-only, RFC 9525, no Common Name fallback; wildcard SAN honored under the same one-label rule).
4. Every non-leaf entry must assert `basicConstraints CA:TRUE` and, if `keyUsage` is present, `keyCertSign`.
5. Adjacent entries must RFC 5280 §7.1 name-chain (issuer → next subject) and the earlier certificate's signature must verify under the later certificate's public key (`pki.verify.verifyCertificateSignature`) — no root/trust-store decision is made, only that the transmitted bytes form a coherent, ordered, verifiable chain. Byte-identical duplicate entries are rejected.
6. Derive the Ed25519 public key from the parsed private seed (`SoftwareSigningKey.publicKey()`), constant-time compare against the leaf's 32 bytes (`crypto.secrets.constantTimeEqual`) → `KeyCertificateMismatch` on difference.
7. Defense-in-depth: fixed sign-and-verify probe supplements, never replaces, the exact comparison.

Proven end to end by an out-of-process `openssl s_client` test that trusts only a generated root CA and validates the transmitted leaf+intermediate chain and hostname against a live listener.

## 5. Secret ownership and zeroization lifecycle

- Key-file bytes → `BoundedSecret` (file API), wiped by `deinit` on every path, including the oversized-file probe-read buffer (own regression test, driven through the file API specifically since the in-memory API never reaches that branch).
- Compacted base64 and decoded PKCS#8 DER → `BoundedSecret` scratch, wiped.
- Seed → `FixedSecret(32)`, bridged into the long-lived key via `SoftwareSigningKey.fromSeedSecret(*FixedSecret)`: copies the seed exactly once into a scope it wipes itself and clears the caller's `FixedSecret` before key derivation, rather than a by-value array parameter that left an un-wiped copy at the call boundary. A test observes the source secret cleared immediately after the call.
- Long-lived key: one heap `OwnedSigner{SoftwareSigningKey}` transferred into the snapshot via `SignAdapter.fromSigningKey(..., owned release)`; `sni_provider` releases it exactly once on every build-failure, failed-reload, retirement, and deinit path (`secureZero` before free).
- No raw key DER, seed, or PEM bytes are retained in the long-lived owner; nothing key-derived is logged.

## 6. TCP + H3 shared-provider ownership

```
edge_gateway.run (appliance profile)
  ApplianceCredentials owner            ← constructed BEFORE listen()/UDP bind
    └ ReloadableProvider snapshot (chain copies + OwnedSigner)
        │  borrow: credentials.CredentialProvider
        ├── NativeTlsConnection.create(...)   TCP: TLS 1.3 → ALPN h2 / http/1.1
        └── http3_runtime.Runtime             QUIC: Tls13Backend.initServerWithProvider
```
Owner is declared before (so deinitialized after) the HTTP/3 runtime, worker pool, and active-connection registries. `http3_runtime` no longer has `identity`/`cert_der`/`key_der`/`loadIdentity` (comptime-asserted in its tests); the general profile also initializes H3 through a provider (from the generic `NativeCredentialStore`), so there is no duplicate H3 credential architecture in any profile.

## 7. Startup / reload publication behavior

- Startup (`edge_gateway.run`): load + fully validate before binding TCP/UDP; any failure aborts before HTTP dispatch is reachable, propagating the real typed error.
- `tardi run` (`main.zig`): the same credential preflight runs before any daemon fork, PID file write, or master/worker startup — a `run --daemon` invocation with invalid credentials no longer prints a false "started" message, and a master process never writes a PID file or enters its worker-respawn loop for credentials that can never load. Verified for direct, daemon, and master-mode invocations.
- `tardi check`: identical preflight (files → strict parse → chain coherence → key/cert match → SAN → server-name policy → flight bounds → snapshot build + clean teardown) with **no** socket.
- Hot reload (preferred minimal behavior): rejected when certificate path, key path, `tls_server_name`, or SNI credential configuration changes — including turning TLS on or off relative to how the process started, gated on the build-time profile rather than whether a credential owner happened to be constructed at startup, so a server that started plaintext correctly rejects a reload that would turn TLS on. Previous configuration and provider stay active (`gateway_shutdown.applianceCredentialConfigChanged`). `reloadFromBytes/Files` remain for future rotation: parse off-path, publish atomically, failed preparation publishes nothing, pinned selections outlive retirement (unit- and integration-tested in both directions).
- `validateApplianceTlsProfile` (`edge_config.zig`) rejects, deterministically, active configuration the engine cannot honor before any of the above runs: non-1.3 TLS versions, explicit cipher selection, client-certificate verification, OCSP stapling/auto-refresh, CRL checking, ACME, session cache/tickets, PROXY protocol with TLS active, a nonzero dynamic-credential-reload interval, HTTP/3 0-RTT/connection migration, HTTP/3 enabled without a complete identity, and a configured `tls_server_name` with no credential files (which would otherwise silently start plaintext with an inert setting). `tls_min_version`, the session cache/ticket flags, and the dynamic-reload interval default to appliance-appropriate values in appliance builds so a config that never touches these settings still passes.

## 8. No second architecture

No new `CredentialProvider`, signer abstraction, SNI selector, fixed-provider path, X.509 parser, or TLS authentication path was added. The engine sees only `credentials.CredentialProvider`; the appliance module composes `pki.pem`, `pki.x509`, `pki.identity`, `pki.verify`, `crypto.pure_zig.SoftwareSigningKey`, `crypto.secrets`, and `sni_provider`.

## 9. #348 isolation

No #348-owned file was touched: `git diff origin/main...HEAD --name-only` contains no `tests/pki_differential*`, `tests/vectors/pki/**`, oracle, or reduction files. The only `build.zig` changes are additive `addImport` lines.

## 10. Validation results (all local, macOS arm64, Zig 0.16.0; also green on CI's 3-OS matrix)

| Command | Result |
| --- | --- |
| `git diff --check` | clean |
| `zig fmt --check build.zig src/ tests/` | clean |
| `zig build test-tls` | 325/325 |
| `zig build test-quic` | 392/392 |
| `zig build test-crypto` | 44/44 |
| `zig build test` | 1656/1657 (1 pre-existing unrelated skip) |
| `zig build test -Dtls-profile=appliance` | 1628/1629 (1 pre-existing unrelated skip) |
| `zig build test-integration-native-tls -Dtls-profile=appliance` | 8/8 |
| `zig build test-integration` | 72 pass, 17 skip (89 total) |
| `zig build test-integration -Dtls-profile=appliance` | 81 pass, 8 skip (89 total) |
| `zig build -Doptimize=ReleaseFast` | success |
| `zig build -Doptimize=ReleaseFast -Dtls-profile=appliance` | success |

CI: a dedicated "Test appliance profile" job (ubuntu-latest, ubuntu-24.04-arm, macos-14) runs the appliance unit suite and native-TLS-listener integration tests on all three OSes; the full integration suite runs on Linux only there, matching this repo's pre-existing `integration-tests` job's own Linux-only scope for the same class of timing-sensitive assertions.

## 11. Binary linkage audit

ReleaseFast appliance artifact:

```
$ otool -L zig-out/bin/tardi
zig-out/bin/tardi:
        /usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1351.0.0)
$ strings zig-out/bin/tardi | grep -ci 'OPENSSL_\|libssl\|libcrypto'
0
$ ./zig-out/bin/tardi version
dev (tls-profile=appliance, tls-backend=native)
```

No libssl/libcrypto or other foreign TLS/crypto/QUIC/HTTP-3 linkage. The appliance runtime path has no OpenSSL imports, no fallback through `identity_loader.zig` (structurally unreachable in the appliance branch), no retained raw private-key DER, and no private-key logging.

## 12. Limitations / notes

- `scripts/audit-dependencies.sh` requires bash ≥ 4 (associative arrays) and cannot run under macOS system bash 3.2; it runs in CI as before. The binary-level audit above was performed directly, and the new code adds no `@cImport`/OpenSSL references for it to flag.
- The pre-existing `credentials.testdata` certificate (used elsewhere in the TLS test suite) is mis-encoded X.509 (issuer Name serialized as BIT STRING); the strict `pki.x509` parser rejects it, so the appliance tests use new well-formed Ed25519 fixtures (`tests/fixtures/tls/native_ed25519*`, including a real self-signed CA and a leaf it actually issued). Existing consumers of the old fixture (via the lenient std parser) are unchanged; regenerating it is out of scope here.
- Certificate-chain coherence proves the transmitted bytes form a verifiable, ordered signing chain (issuer/subject linkage, signature verification, CA/KeyUsage constraints) but deliberately does not add public-PKI trust-store policy: no root pinning, no revocation checking, no path-length/name-constraint enforcement. That remains out of #392's scope by the maintainer's own guidance.
- Credential rotation remains restart-only in v0.5 by design (reload of changed credential inputs is explicitly rejected, not ignored).


src/edge_config.zig
```diff
@@ -2553,6 +2562,37 @@ pub fn hasTlsFiles(cfg: *const EdgeConfig) bool {
...
+/// mechanism is rejected.
+fn validateTlsServerNamePolicy(cfg: *const EdgeConfig) !void {
+    const tls_core = @import("tls_core");
+    if (cfg.tls_server_name.len > 0) {
+        tls_core.appliance_credentials.validateServerName(cfg.tls_server_name) catch {
+            std.log.err("config validation failed: TARDIGRADE_TLS_SERVER_NAME must be a single non-wildcard DNS host name", .{});
+            return error.InvalidConfigValue;
+        };
+    }
+    if (is_appliance_tls_profile and hasTlsFiles(cfg)) {
```
Jaruso: **Blocking fail-open configuration path.** `hasTlsFiles()` is true only when *both* top-level paths are present, so cert-only or key-only configurations skip this entire appliance policy and skip the strict credential preflight. They can then start without native TLS instead of returning `MissingPrivateKey` / `MissingCertificateChain`. This block also rejects only `tls_sni_certs`; appliance `server_blocks` with their own cert/key pairs are still accepted below, contradicting the exact-one-identity contract.

Validate the pair independently before using `hasTlsFiles()`:

```zig
const has_cert = cfg.tls_cert_path.len != 0;
const has_key = cfg.tls_key_path.len != 0;
if (is_appliance_tls_profile) {
    if (has_cert != has_key) return error.InvalidConfigPath;
    if (has_cert) {
        if (cfg.tls_server_name.len == 0) return error.InvalidConfigValue;
        if (cfg.tls_sni_certs.len != 0) return error.InvalidConfigValue;
        for (cfg.server_blocks) |block| {
            if (block.tls_cert_path.len != 0 or block.tls_key_path.len != 0)
        }
    }
}
```

Add `check` and startup tests for cert-only, key-only, and a server-block identity.

src/gateway_shutdown.zig
```diff
@@ -96,6 +96,28 @@ pub fn hotReloadConfig(
             return;
         };
     }
+    if (worker_ctx.appliance_credentials != null) {
```
Jaruso: **Blocking generation-coherence bug.** This check runs only when the process started with an appliance credential owner. If the appliance starts plaintext, `appliance_credentials` is null; a hot reload that adds `tls_cert_path`, `tls_key_path`, and `tls_server_name` skips this guard, publishes the new config, but never creates `native_tls_provider`. Future connections therefore continue as plaintext under a TLS-marked config.

The restart-required comparison must be profile-driven and run for every appliance reload, including transitions into or out of TLS. Compare current/proposed credential fields whenever `edge_config.is_appliance_tls_profile`, reject the entire transition before publishing, and add barrier/integration tests for no-TLS→TLS and TLS→no-TLS reloads.

src/tls/appliance_credentials.zig
```diff
@@ -0,0 +1,1234 @@
...
+        error.UnterminatedPemBlock,
+        error.InvalidPemBase64,
+        error.EmptyPemBlock,
+        => error.MalformedCertificatePem,
+        error.MalformedCertificateDer => error.MalformedCertificateDer,
+        error.OutOfMemory => error.OutOfMemory,
+    };
+}
+
+fn preflightChain(chain: *const pki.pem.CertificateChain, limits: Limits) Error!void {
```
Jaruso: **Blocking: this is only a size preflight, not the documented leaf-first certificate-chain validation.** Every non-leaf entry is accepted after the PEM layer proves only that it is one DER SEQUENCE. The test named `leaf plus intermediate chain preserves transmission order` concatenates the leaf twice and treats that as valid, demonstrating that duplicate/reordered/unrelated entries are not rejected.

Parse every entry with the existing `pki.x509.Certificate.parse`. For the constrained appliance profile, require the first certificate to be the end entity, subsequent entries to be CA certificates (and `keyCertSign` when KeyUsage is present), and each preceding issuer to match the next subject through the existing RFC 5280 name-chaining comparison. Use the existing verifier for signature linkage if its current surface supports that without adding public trust policy. At minimum, malformed X.509 in any entry must return `MalformedCertificateDer`, not fall through provider construction and become `ProviderPublicationFailed`. Add negative tests for duplicate leaf, reversed order, unrelated issuer, non-CA intermediate, and malformed second/third entries.

src/tls/appliance_credentials.zig
```diff
@@ -0,0 +1,1234 @@
...
+
+        var seed = secrets.FixedSecret(Ed25519.KeyPair.seed_length){};
+        defer seed.deinit();
+        try parseEd25519Pkcs8(key_der.slice(), &seed);
+
+        const owner = self.allocator.create(OwnedSigner) catch return error.OutOfMemory;
+        errdefer self.allocator.destroy(owner);
+        owner.* = .{
+            .allocator = self.allocator,
+            .key = crypto.pure_zig.SoftwareSigningKey.fromSeed(
```
Jaruso: **Blocking secret-lifecycle escape.** `seed.bytes[0..32].*` creates an ordinary `[32]u8` copy outside `FixedSecret`, and `SoftwareSigningKey.fromSeed` accepts that array by value without wiping its input parameter. Depending on ABI/inlining this can create one or more unmanaged stack copies of the private seed. The FixedBufferAllocator scan cannot detect them.

Add a pointer/slice/typed-secret constructor (for example `SoftwareSigningKey.fromSeedSecret(*const FixedSecret(32))`) or explicitly bridge through a scoped local array with `defer secureZero` and avoid a second by-value copy. Tests should instrument the constructor/cleanup path rather than only scanning allocator-owned memory.

src/tls/appliance_credentials.zig
```diff
@@ -0,0 +1,1234 @@
...
+) Error!void {
+    const fd = try openBounded(path);
+    defer _ = std.c.close(fd);
+
+    var used: usize = 0;
+    while (true) {
+        if (used == secret.bytes.len) {
+            // Probe for one extra byte: a file exactly at the limit is fine,
+            // anything longer is rejected without buffering it.
+            var probe: [1]u8 = undefined;
```
Jaruso, line 805: This one-byte probe contains private-key material when the file is oversized, then returns immediately without wiping it. That violates the stated rule that every secret-bearing temporary is erased on every error path.

```zig
var probe: [1]u8 = undefined;
defer secrets.secureZero(&probe);
const extra = std.posix.read(fd, &probe) catch return error.FileNotFound;
```

Please also add an oversized **file API** test; the current input-size test exercises only the in-memory API and will not cover this branch.

Jaruso (commented):
**Blocking review — not ready to merge or close #392.**

The architectural direction is strong: this reuses the existing reloadable credential provider, performs exact Ed25519 leaf/key matching, moves HTTP/3 onto the provider contract, and gives TCP and QUIC one composition-root owner. I reviewed the full change against the authoritative #392 guidance and found several release-blocking gaps.

1. **The appliance configuration still fails open around identity cardinality.** Because `hasTlsFiles()` means both paths are present, a cert-only or key-only appliance configuration bypasses the appliance credential policy and preflight and can start as plaintext. Per-server-block TLS identities are also still accepted even though the documented profile requires exactly one identity and says those configurations are rejected.
2. **A plaintext appliance can hot-reload into a TLS-marked config without constructing a provider.** The restart-required comparison runs only when `worker_ctx.appliance_credentials != null`. Starting without TLS leaves that null; a reload adding cert/key/name can therefore publish the new configuration while `native_tls_provider` remains null, so subsequent connections continue down the plaintext path under a TLS configuration.
3. **The chain contract is not actually validated.** Only the leaf is parsed through `pki.x509`; the remaining entries receive size/DER-sequence checks. The positive “intermediate” test concatenates the leaf certificate twice and expects success. This permits duplicate, unrelated, reordered, non-CA, or structurally invalid X.509 intermediates, and a malformed later entry can collapse to `ProviderPublicationFailed` instead of the documented `MalformedCertificateDer` class.
4. **Temporary private-key bytes escape the typed secret lifecycle.** Dereferencing `FixedSecret` creates an ordinary `[32]u8` seed value passed by value to `SoftwareSigningKey.fromSeed`, whose input parameter is not wiped. The allocator-backing scan does not observe stack copies. The oversized-file probe also reads one private-key byte into an ordinary stack buffer and returns without wiping it.
5. **`run` does not preserve the advertised deterministic error contract.** The composition root catches every loader error and returns `ApplianceCredentialInvalid`; that error and `ProviderPublicationFailed` are absent from `isConfigValidationError`. The PR description claims both `check` and `run` report the typed configuration-invalid class/exit behavior, but only `check` currently propagates the underlying loader errors.
6. **Current CI is red.** Run 797 fails the Linux x86_64 and ARM unit-test jobs. The connector output does not expose enough of the final logs to assign a safe root cause, but this must be green before merge.

Please add focused negative/integration coverage for cert-only, key-only, per-server-block credentials, plaintext→TLS and TLS→plaintext reloads, duplicate/reordered/unrelated/malformed intermediates, file-API oversized-key cleanup, and `run` exit/error classification. Keep `Closes #392` only after those paths and the required CI matrix pass.

src/tls/appliance_credentials.zig
```diff
@@ -0,0 +1,1322 @@
...
+    }
+
+    // Every entry — not only the leaf — must be a structurally well-formed
+    // X.509 certificate object, validated with the same pure-Zig parser used
+    // for the leaf (no second SPKI/X.509 parser). This deterministically
+    // rejects a malformed intermediate as `MalformedCertificateDer` here,
+    // before `sni_provider.reload` ever sees it, rather than surfacing as an
+    // undifferentiated `ProviderPublicationFailed` from its own (leaf-only)
+    // DER check. Chain-of-trust policy — issuer/subject linkage, CA-bit
+    // enforcement, duplicate/reorder detection — remains outside #392's
```
Jaruso: **Blocking: this still accepts an invalid ordered chain.** Parsing each entry as X.509 fixes malformed-intermediate classification, but the code explicitly excludes issuer/subject linkage, CA constraints, duplicate detection, and signature linkage. The new positive fixture is a self-signed CA with a different subject/key that did not issue the leaf, so an external client trusting that CA cannot validate the transmitted `leaf + CA` sequence.

This is chain-coherence validation, not public trust-store policy. Reuse the existing PKI surfaces:

```zig
// Parse every entry and retain the parsed views for this preflight.
for (parsed, 0..) |*cert, i| {
    if (cert.hasUnhandledCriticalExtension())
        return error.UnhandledCriticalCertificateExtension;

    if (i == 0) {
        if (cert.basicConstraints()) |bc| if (bc.is_ca)
            return error.InvalidLeafCertificate;
        if (cert.keyUsage()) |ku| if (!ku.digital_signature)
            return error.CertificateKeyUsageViolation;
        if (cert.extendedKeyUsage()) |eku| if (!eku.allowsServerAuth())
            return error.CertificateExtendedKeyUsageViolation;
    } else {
        const bc = cert.basicConstraints() orelse
            return error.IntermediateNotCa;
        if (!bc.is_ca) return error.IntermediateNotCa;
        if (cert.keyUsage()) |ku| if (!ku.key_cert_sign)
            return error.CertificateKeyUsageViolation;
    }
}

for (parsed[0 .. parsed.len - 1], parsed[1..]) |*child, *issuer| {
    if (!child.issuer.eqlForChaining(&issuer.subject))
        return error.InvalidCertificateChainOrder;
    pki.verify.verifyCertificateSignature(
        crypto_provider,
        child,
        &issuer.subject_public_key_info,
    ) catch return error.CertificateSignatureInvalid;
}
```

Also reject byte-identical duplicate entries. Do **not** require the final transmitted certificate to be self-signed or trusted; only prove that each preceding certificate was issued by the next one. Replace the current unrelated-CA fixture with a real root → intermediate → leaf fixture, transmit leaf + intermediate, and add an out-of-process client test that trusts only the root and validates the hostname.

Jaruso (commented):
## Blocking re-review — current head `5ffad359`

The new commit materially improves the PR. I verified that it now rejects partial cert/key pairs and per-server identities, gates appliance reloads by build profile, preserves typed startup errors, wipes the oversized-file probe, parses every non-leaf entry as X.509, and passes the current CI run.

The PR is still **not ready to merge**. The remaining blockers are:

1. A structurally valid but unrelated or duplicated certificate is still accepted as an “intermediate”; the positive test deliberately uses a self-signed CA that did not issue the leaf. This does not prove an ordered chain a client can validate.
2. `tls_server_name` is not checked against the leaf SAN, canonical Ed25519 SPKI parameters are not enforced, and ordinary TLS leaf suitability checks are absent.
3. The private seed still crosses a by-value API and is then copied again; wiping only the new local copy does not wipe the caller temporary or parameter storage.
4. `run` still performs credential loading only inside `edge_gateway.run`: daemonization happens before config loading, while master mode writes its PID and starts workers first. Invalid credentials can therefore be reported as a successfully started daemon or produce a master respawn loop.
5. Certificate-flight preflight still uses an approximate 512-byte reserve and accepts caller limits above the engine’s hard bounds. HTTP/3 EncryptedExtensions can exceed that reserve.
6. Appliance builds still accept active TLS features the native fixed profile cannot honor. Some fail only per connection or leave HTTP/3 bound but unbootstrapped instead of failing configuration validation.
7. The green CI run does **not** execute the appliance integration suite: the matrix runs only `zig build test` under the default/general profile, where `requireNativeTlsProfile()` skips the new listener tests. Add an appliance-profile test job running at least `zig build test -Dtls-profile=appliance`, `test-integration-native-tls`, and the appliance integration suite.

The probe wipe itself is fixed, but the requested oversized **file API** regression test is still missing; the remaining size test exercises only `initFromBytes`.

After the code is corrected, reconcile the PR body and `docs/BARE_APPLIANCE_TLS.md`: they currently claim valid client chain verification, identical `check`/startup preflight, canonical SPKI, and writer-identical flight arithmetic that the implementation does not yet provide.

src/tls/appliance_credentials.zig
```diff
@@ -0,0 +1,1322 @@
...
+    allocator: std.mem.Allocator,
+    leaf_der: []const u8,
+    out: *[Ed25519.PublicKey.encoded_length]u8,
+) Error!void {
+    var leaf = pki.x509.Certificate.parse(allocator, leaf_der, .{}) catch |err| switch (err) {
+        error.OutOfMemory => return error.OutOfMemory,
+        else => return error.MalformedCertificateDer,
+    };
+    defer leaf.deinit(allocator);
+    if (leaf.subject_public_key_info.key_type != .ed25519)
```
Jaruso, line 474: **Blocking leaf-policy gap.** `key_type == .ed25519` is not enough to enforce the documented canonical RFC 8410 SPKI: the generic parser can classify the Ed25519 OID even when illegal AlgorithmIdentifier parameters are present. Require `leaf.subject_public_key_info.algorithm.parameters_raw == null` in addition to the BIT STRING checks.

This parsed leaf must also be bound to the configured identity before publication:

```zig
const verdict = pki.identity.verifyHost(&leaf, options.server_name) catch
    return error.InvalidServerName;
if (!verdict.isMatch()) return error.CertificateNameMismatch;
```

Keep this SAN-only—no Common Name fallback. Add tests for absent/wrong SAN, case-insensitive exact SAN, a matching presented wildcard SAN, and an Ed25519 SPKI carrying illegal NULL parameters. Without this, `tls_server_name = other.example` can select and serve a certificate for `tardigrade.test`, and startup succeeds even though normal clients reject it.

src/tls/appliance_credentials.zig
```diff
@@ -0,0 +1,1322 @@
...
+        var seed = secrets.FixedSecret(Ed25519.KeyPair.seed_length){};
+        defer seed.deinit();
+        try parseEd25519Pkcs8(key_der.slice(), &seed);
+
+        const owner = self.allocator.create(OwnedSigner) catch return error.OutOfMemory;
+        errdefer self.allocator.destroy(owner);
+        owner.* = .{
+            .allocator = self.allocator,
+            .key = crypto.pure_zig.SoftwareSigningKey.fromSeed(
+                seed.bytes[0..Ed25519.KeyPair.seed_length].*,
```
Jaruso: **Blocking: the seed-copy issue remains.** This dereference creates an ordinary `[32]u8` caller temporary. `SoftwareSigningKey.fromSeed` then receives another by-value parameter and copies it again into `local_seed`; the new defer wipes only that final local. It does not wipe the caller temporary or guarantee erasure of parameter/ABI storage.

Change the application boundary to accept the typed secret or a mutable pointer, for example:

```zig
pub fn fromSeedSecret(
    seed: *secrets.FixedSecret(Ed25519.KeyPair.seed_length),
) provider.SignError!SoftwareSigningKey {
    var bridge: [Ed25519.KeyPair.seed_length]u8 = undefined;
    defer secrets.secureZero(&bridge);
    @memcpy(&bridge, seed.slice());
    seed.deinit();
    const pair = Ed25519.KeyPair.generateDeterministic(bridge) catch
        return error.ProviderFailure;
    return .{ .key_pair = pair };
}
```

Use the narrowest unavoidable bridge at the std.crypto call, wipe it in the same scope, and clear the source `FixedSecret` before returning. Add a direct test that observes the source container cleared immediately after construction; the arena scan cannot detect stack/parameter copies.

src/main.zig
```diff
@@ -749,6 +786,24 @@ fn executeStatusCommand(allocator: std.mem.Allocator, options: SignalOptions) !v
...
 
+/// Classify a `run` startup failure exactly like `tardi check` does — a
+/// deterministic configuration error (including every
+/// `appliance_credentials.Error` class the composition root can now
+/// propagate unwrapped) reports the same message shape and
+/// `EXIT_CONFIG_INVALID` exit code, rather than an opaque generic failure.
+/// Post-startup errors (the server crashing after already accepting
+/// connections) are comparatively rare in practice and still fall through to
+/// the internal-error exit code with `@errorName`.
+fn executeRunCommandOrExit(allocator: std.mem.Allocator, args: []const []const u8, options: RunOptions) !void {
```
Jaruso: **Blocking: wrapping the whole runtime does not make `run` preflight equivalent to `check`.** `executeRunCommand` still daemonizes before loading the config, then writes the PID and enters master mode before `edge_gateway.run` constructs `ApplianceCredentials`. Consequences:

- `run --daemon` can print that it started successfully while the child later rejects the credentials;
- master mode writes a PID and starts workers that fail credential loading;
- `runMaster` respawns every exited worker, creating a permanent failure loop;
- unrelated post-start runtime errors whose names overlap configuration errors can now be misreported as “configuration validation failed.”

Extract one shared appliance preflight and call it immediately after `edge_config.validate`, before daemon spawn, error-log/PID setup, or master/worker startup:

```zig
try edge_config.validate(&cfg);
try preflightApplianceCredentials(allocator, &cfg);

if (options.daemon and !options.daemonized and !worker_mode) { ... }
try configureErrorLog(&cfg);
try writePidFile(&cfg);
if (cfg.master_process_enabled and !worker_mode) { ... }
```

The worker should still construct the long-lived owner in `edge_gateway.run`. Add direct, daemon, and master-mode tests with a mismatched key; all must fail deterministically without a listener, stale PID, success message, or worker respawn loop.

src/tls/appliance_credentials.zig
```diff
@@ -0,0 +1,1322 @@
...
+const certificate_message_overhead = 1 + 3 + 1 + 3;
+/// Per-CertificateEntry framing: 3-byte cert_data length + 2-byte extensions
+/// length. Mirrors `tls13_backend`.
+const certificate_entry_overhead = 3 + 2;
+/// Headroom reserved for the other messages sharing the server's one flight
+/// buffer (EncryptedExtensions with ALPN and the optional CertificateRequest
+/// are each well under this).
+const flight_headroom = 512;
+
+pub const default_max_certificate_flight_bytes =
```
Jaruso, line 43: **Blocking boundary mismatch with the actual writer.** A fixed 512-byte reserve is not writer-identical. HTTP/3 EncryptedExtensions may contain the QUIC transport extension (up to 512 bytes) **plus** extension framing, the handshake/extensions vectors, and ALPN, so the non-certificate portion can exceed this reserve. A chain can pass startup and then fail inside `emitServerAuthFlight`.

Expose a shared sizing helper/constant from `tls13_backend.zig` or compute the exact conservative profile-specific bound from the same components the writer uses. Also ensure custom `Limits` can only tighten hard bounds:

```zig
const max_entries = @min(limits.max_chain_entries, credentials.max_chain_entries);
const max_der = @min(limits.max_certificate_der_bytes, tls13_backend.max_certificate_len);
const max_flight = @min(limits.max_certificate_flight_bytes, tls13_backend.max_message_len);
```

Add boundary tests for both record ALPN (`http/1.1`/`h2`) and the maximum HTTP/3 transport-extension profile, proving every chain accepted by preflight can be serialized by the live writer.

src/edge_config.zig
```diff
@@ -2553,6 +2562,58 @@ pub fn hasTlsFiles(cfg: *const EdgeConfig) bool {
...
+        return error.InvalidConfigValue;
+    }
+    for (cfg.server_blocks) |block| {
+        if (block.tls_cert_path.len > 0 or block.tls_key_path.len > 0) {
+            std.log.err("config validation failed: the appliance TLS profile does not support per-server-block TLS certificates; configure TARDIGRADE_TLS_CERT_PATH/TARDIGRADE_TLS_KEY_PATH directly", .{});
+            return error.InvalidConfigValue;
+        }
+    }
+
+    if (hasTlsFiles(cfg) and cfg.tls_server_name.len == 0) {
```
Jaruso, line 2626: **Blocking fixed-profile validation is still incomplete.** The exact-one-identity checks are now correct, but appliance builds continue to accept active settings the native profile ignores or cannot implement. Examples:

- the default minimum remains TLS 1.2 although the engine is TLS 1.3-only;
- `tls_client_verify` is accepted, then every native connection is dropped at runtime;
- OCSP/CRL/ACME, session cache/tickets, custom cipher settings, and dynamic credential watching are accepted but not honored by this owner;
- HTTP/3 can be enabled with no complete identity, causing the UDP socket to bind with `server_bootstrapped = false`;
- unsupported 0-RTT/migration settings only warn and continue.

Add a focused `validateApplianceTlsProfile` and profile-specific defaults. Reject active unsupported behavior with `UnsupportedApplianceConfiguration`; at minimum require exact TLS 1.3, a complete identity for HTTP/3, no downstream mTLS/CRL/OCSP/ACME, no unsupported credential watcher, and no enabled 0-RTT/migration. Harmless inactive values such as the default ACME directory URL need not be rejected. Add one negative test per setting so an appliance never starts in a configuration it cannot actually serve.

src/main.zig
```diff
@@ -491,6 +514,30 @@ fn isConfigValidationError(err: anyerror) bool {
...
+        error.CertificateTooLarge,
+        error.MalformedCertificateDer,
+        error.MalformedPrivateKeyPem,
+        error.AmbiguousPrivateKeyInput,
+        error.MalformedPrivateKeyDer,
+        error.UnsupportedPrivateKeyAlgorithm,
+        error.UnsupportedPrivateKeyParameters,
+        error.InvalidPrivateKeySize,
+        error.InvalidPrivateKey,
+        error.UnsupportedLeafKeyAlgorithm,
```
Jaruso, line 536: **Blocking error-taxonomy regression.** The appliance loader now returns additional deterministic operator-input failures, but this switch still stops at the old set. At minimum it omits:

```zig
error.UnsupportedLeafKeyParameters,
error.CertificateNameMismatch,
error.IntermediateNotCa,
error.InvalidLeafCertificate,
error.CertificateKeyUsageViolation,
error.CertificateExtendedKeyUsageViolation,
error.UnhandledCriticalCertificateExtension,
error.DuplicateCertificateEntry,
error.InvalidCertificateChainOrder,
error.CertificateSignatureInvalid,
```

As written, `tardi check` or `tardi run` with a wrong SAN, illegal Ed25519 SPKI parameters, or an incoherent chain prints `unexpected internal error` and exits with the internal-error code instead of configuration-invalid exit 2. That contradicts the issue AC and the updated provisioning documentation.

Add every **input-derived** class above to `isConfigValidationError`. Keep true resource/invariant failures such as `OutOfMemory` and `ProviderPublicationFailed` internal. Add process-level tests for both `check` and `run` using at least a wrong-SAN certificate and an unrelated/reordered chain, asserting exit 2 and the exact error class.

src/edge_config.zig
```diff
@@ -2553,6 +2572,121 @@ pub fn hasTlsFiles(cfg: *const EdgeConfig) bool {
...
+    }
+    if (cfg.tls_crl_check) {
+        logConfigDiagnostic("config validation failed: the appliance TLS profile does not support TARDIGRADE_TLS_CRL_CHECK", .{});
+        return error.UnsupportedApplianceConfiguration;
+    }
+    if (cfg.tls_acme_enabled) {
+        logConfigDiagnostic("config validation failed: the appliance TLS profile does not support TARDIGRADE_TLS_ACME_ENABLED", .{});
+        return error.UnsupportedApplianceConfiguration;
+    }
+}
```
Jaruso, line 2713: **Blocking fixed-profile fail-open gaps remain.** This validator now rejects several unsupported features, but it still accepts configurations the native appliance cannot serve correctly:

- `proxy_protocol_mode != .off` with TLS: `startNewConnection` explicitly logs that the native TLS path does not support the PROXY preface and closes every connection before the handshake.
- `tls_dynamic_reload_interval_ms`: it still defaults to `5000`, is passed only to the OpenSSL terminator, and is silently ignored by the appliance owner even though the documented profile says filesystem watchers are rejected.
- `tls_server_name` without either credential file: syntax validation succeeds and the appliance silently starts plaintext with an inert TLS identity setting.
- `tls_ocsp_auto_refresh = true` is another active OpenSSL-only request that is not rejected when stapling is off.

Suggested shape:

```zig
// loadFromEnv
const tls_dynamic_reload_interval_ms = parseIntEnv(
    u64,
    allocator,
    "TARDIGRADE_TLS_DYNAMIC_RELOAD_INTERVAL_MS",
    if (is_appliance_tls_profile) 0 else 5000,
);

// appliance policy
const has_cert = cfg.tls_cert_path.len != 0;
const has_key = cfg.tls_key_path.len != 0;
if (cfg.tls_server_name.len != 0 and !has_cert and !has_key)
    return error.UnsupportedApplianceConfiguration;
if (hasTlsFiles(cfg) and cfg.proxy_protocol_mode != .off)
    return error.UnsupportedApplianceConfiguration;
if (hasTlsFiles(cfg) and cfg.tls_dynamic_reload_interval_ms != 0)
    return error.UnsupportedApplianceConfiguration;
if (hasTlsFiles(cfg) and cfg.tls_ocsp_auto_refresh)
    return error.UnsupportedApplianceConfiguration;
```

Keep the runtime guards as defense in depth, but reject these before PID/listener creation. Add one appliance-profile unit/CLI test per setting.

src/main.zig
```diff
@@ -749,6 +796,24 @@ fn executeStatusCommand(allocator: std.mem.Allocator, options: SignalOptions) !v
...
+/// deterministic configuration error (including every
+/// `appliance_credentials.Error` class the composition root can now
+/// propagate unwrapped) reports the same message shape and
+/// `EXIT_CONFIG_INVALID` exit code, rather than an opaque generic failure.
+/// Post-startup errors (the server crashing after already accepting
+/// connections) are comparatively rare in practice and still fall through to
+/// the internal-error exit code with `@errorName`.
+fn executeRunCommandOrExit(allocator: std.mem.Allocator, args: []const []const u8, options: RunOptions) !void {
+    executeRunCommand(allocator, args, options) catch |err| {
+        var stderr_buf: [2048]u8 = undefined;
```
Jaruso: The credential preflight is now correctly before daemonization/PID/master startup, but this outer catch still covers `edge_gateway.run` and `runMaster`. Therefore the statement above that post-start errors “fall through to the internal-error exit code” is not true: any runtime error whose global name appears in `isConfigValidationError`—notably `FileNotFound` or `AccessDenied` from stores/filesystem work—will be printed as a configuration validation failure and exit 2.

Split stage-aware handling so only config discovery, `edge_config.validate`, and `preflightApplianceCredentials` use the configuration classifier. Once execution enters daemon/master/worker runtime, report returned failures as internal/runtime errors regardless of coincident error names. For example, prepare/validate a `RunPlan` first under the config-error catch, then execute the selected daemon/master/gateway action under a runtime-error catch. Add a focused test proving a synthetic runtime `FileNotFound` is not classified as invalid configuration.

Jaruso (commented):
**Blocking re-review — major progress, but not quite merge-ready.**

The two new commits correctly address the previously reported chain-coherence, hostname/SPKI, typed-seed, pre-daemon/master preflight, writer-bound, and appliance-CI gaps. I rechecked those implementations and the associated tests; the architecture remains sound and CI run 808 is green across the new appliance matrix.

Three correctness gaps remain:

1. The loader added ten new operator-input error classes, but `main.isConfigValidationError` does not classify them. Wrong SANs and incoherent chains therefore report an “unexpected internal error” and exit with the internal-error code instead of the documented configuration-invalid code 2.
2. The appliance profile still starts with active configuration it cannot honor. In particular, PROXY protocol makes the native TLS path close every connection before handshake, and the credential watcher interval remains 5000 ms by default even though the appliance never constructs that watcher. A name-only TLS configuration is also silently treated as plaintext.
3. The `run` error wrapper still classifies the *entire long-running gateway* by error name. Moving credential preflight before daemon/master startup was correct, but post-start/runtime errors such as `FileNotFound` or `AccessDenied` can still be mislabeled as configuration failures.

Please also reconcile the PR body before merge: it still lists the old error taxonomy and says startup collapses to `ApplianceCredentialInvalid`, while the implementation now intentionally preserves typed errors.

Once the inline items are fixed and the appliance matrix is green again, I do not see a need for another architectural phase or follow-up issue.

Jaruso (commented):
## Re-review complete — no blocking findings

Reviewed current head `3d453157d3b655fb4eeefd3e38fe73654403a4c2` against issue #392, the authoritative implementation guidance, the full PR diff, and every finding from the previous review rounds.

The final three blockers are now correctly addressed:

1. All deterministic operator-input credential errors—including SAN/SPKI, chain-order, CA/KU/EKU, duplicate-entry, critical-extension, and signature-linkage failures—are classified as configuration-invalid errors. True resource/invariant failures remain internal.
2. The appliance profile now fails closed for unsupported active configuration: PROXY protocol with native TLS, credential watching, OCSP auto-refresh, a server name without credentials, unsupported TLS/cipher/client-auth/resumption/ACME/CRL settings, and unsupported HTTP/3 modes. Appliance-specific defaults no longer trigger those checks accidentally.
3. `tardi run` now has phase-scoped error handling: `prepareRunPlan` owns config discovery/validation/credential preflight, while daemon/PID/master/worker/gateway execution is handled separately as runtime work. Runtime `FileNotFound`/`AccessDenied` can no longer be mislabeled as configuration errors, and the integration test exercises that behavior.

I also rechecked the previously fixed areas rather than relying on outdated thread state: complete certificate-chain coherence, SAN and canonical Ed25519 SPKI binding, typed-secret seed ownership and zeroization, pre-daemon/master credential validation, exact TLS writer bounds, one shared TCP/H3 credential provider, reload generation coherence, independent client chain/hostname verification, and no-OpenSSL appliance linkage.

CI run 809 completed successfully, including the general suites, dedicated appliance-profile suites on Linux x86_64, Linux ARM64, and macOS, native TLS listener/full appliance integration tests, formatting, builds, packaging smoke, security checks, and TLS dependency/linkage audit.

**Verdict: PR #472 satisfies #392 and is ready to merge.**

PR #473, opened 2026-07-19
test(pki): complete semantic differential validation and hostile corpus

Jaruso:
## Summary

This PR is the closing implementation for the remaining issue #348 semantic-differential work on top of PRs #449, #453, and #456. It keeps the branch test/tooling-only while adding semantic observations, reason-aware manifest expectations, schema-v4 mismatch artifacts, corrected Name Constraints/identity fixtures, a sound hostile algorithm/DER fixture subset, and a newly promoted reduced DER regression seed.

## Ownership boundary

This branch is test/tooling-only. It does not modify production TLS, HTTP, QUIC, credential, SNI, signing, appliance configuration, startup, runtime certificate-validation behavior, or GitHub workflows.

## Acceptance-criteria audit

| Criterion | Prior completion | Closing-branch status |
|---|---|---|
| Three-way deterministic policy, out-of-process OpenSSL/Go | PR #449 | Preserved |
| Stable core plus weekly/manual extended CI | PR #449 | Preserved; no duplicate workflow |
| Deterministic project-owned corpus and provenance | PR #449 | Expanded and documented in `tests/vectors/pki/README.md` |
| Reproducible mismatch artifacts | PR #449 | Upgraded to semantic schema v4 |
| Bounded reduction, external re-verification, promoted regression seeds | PR #453 | Reduction now preserves status/reason/index; promoted `non-minimal-long-length` DER seed |
| Deadlines, output caps, kill/reap, structured process failures | PR #456 | Preserved and mapped to distinct tool reasons |
| Compare semantic reason, not only accept/reject | - | Implemented with validator/status/reason observations and manifest expectations |
| Correct false-agreement Name Constraints/identity fixtures | - | DNS-only, IP-only, and isolated identity fixtures added; directoryName divergence split into extended cases |
| Hostile algorithm/DER corpus | - | Sound algorithm/DER fixture subset added; duplicate and stale-signature-confounded fixtures removed |
| Final resource/closure audit and validation | - | Local validation completed below; GitHub CI run 804 passed at head `fe566e34` |

## Review-blocker response

First review blockers:

- OpenSSL numeric verification-code mappings were corrected and regression-tested against stable `X509_V_ERR_*` meanings.
- OpenSSL code parsing now accepts only verifier-owned `error N at M depth lookup:` lines from stdout/stderr, so certificate-controlled subject text cannot spoof the semantic reason.
- Tardigrade fixture/filesystem load failures now become `tool_failure/fixture_load_failure`, not certificate rejections.
- Tool failures and unclassified observations cannot enter minimization or become promotion-ready reduced seeds.
- The OpenSSL malformed-DER exception is restricted to committed malformed parser-boundary failures; missing or unreadable inputs remain harness failures.

Follow-up review blockers:

- OpenSSL STORE parse failures no longer map through `case.id`; the classifier no longer accepts manifest metadata and records only the parser-boundary `malformed_der` reason that OpenSSL output supports.
- The byte-identical `algorithm-malformed-signature-bits` / `der-invalid-bit-string-unused` duplication was removed.
- TBS-mutating generated fixtures are re-signed when they remain certificate-shaped, so they no longer depend on stale-signature rejection.
- Algorithm cases that only proved an incompatible Ed25519 issuer or were accepted by all validators after re-signing were removed instead of being pinned as misleading negative fixtures.
- OpenSSL verification code `24` is now mapped as `X509_V_ERR_INVALID_CA` to `key_usage_failure`, not issuer/SPKI decode failure.
- The Go generator rejects byte-identical hostile DER payloads across the algorithm and DER fixture maps before writing files.
- `tests/pki_differential.zig` pins hostile DER byte-uniqueness, case-to-file metadata invariants, and case-metadata-independent OpenSSL STORE classification.

## Corpus and registry updates

- Added DNS-only and IP-only Name Constraints intermediates/leaves so DNS/IP negative cases prove the named rule in all three validators.
- Added an identity-mismatch chain with no unrelated unsupported critical constraints.
- Split legacy directoryName Name Constraints behavior into explicit extended divergence cases for Go.
- Added deterministic algorithm fixtures for outer/inner AlgorithmIdentifier mismatch, unsupported/malformed signature OIDs, illegal Ed25519 parameters, and malformed SPKI OID encoding.
- Added deterministic DER hostility fixtures for non-minimal/indefinite/truncated lengths, malformed INTEGER/BIT STRING encodings, constructed primitive encoding, trailing bytes, and malformed nested extension length.
- Promoted `tests/vectors/pki/reduced/non-minimal-long-length.der` into the reduced registry with exact DER-reader `NonMinimalLength` coverage and fuzz/X.509 parser participation.

## Validation completed locally

- `git diff --check`
- `tests/vectors/pki/generate.sh`
- `go test tests/pki_go_validator.go tests/pki_go_validator_test.go`
- `go vet tests/pki_go_validator.go`
- `shellcheck tests/vectors/pki/generate.sh`
- `zig build test-pki --summary all --error-style verbose` - 202/202 passed
- `zig build test-pki-reduce --summary all --error-style verbose` - 42/42 passed
- `zig build test-pki-differential --summary all --error-style verbose --test-timeout 2m` - core corpus passed
- `zig build test-pki-differential-extended --summary all --error-style verbose --test-timeout 10m` - full corpus passed
- `zig build test-pki-openssl --summary all --error-style verbose` - 2/2 passed
- `zig build test-pki-policy-openssl --summary all --error-style verbose` - 1/1 passed
- `zig build test --summary all --error-style verbose --test-timeout 2m` - 1637/1638 passed, 1 skipped
- `zig build test-integration --summary all --error-style verbose` - passed on rerun, 71/78 passed and 7 skipped; first run hit the existing reload short-static flake in `tests/integration.zig`

Closes #348


tests/pki_differential.zig
```diff
@@ -315,26 +435,94 @@ fn opensslDecision(allocator: std.mem.Allocator, case: manifest.Case, deadline_m
...
+        2, 18, 19, 20, 21, 27, 28, 30, 31, 56 => .untrusted_or_incomplete_path,
+        4, 7 => .signature_invalid,
+        6, 76 => .issuer_key_or_spki_invalid,
+        9, 10 => .validity_failure,
+        22, 25 => .path_length_violation,
+        24, 32, 35 => .key_usage_failure,
+        26 => .extended_key_usage_failure,
+        34 => .unknown_critical_extension,
+        37, 38, 39, 49 => .policy_failure,
+        41, 42, 57, 58, 59, 60, 61, 75, 77, 78 => .signature_algorithm_invalid,
```
Jaruso: **Blocking: several numeric OpenSSL verification codes are mapped to unrelated semantic reasons.** The official `X509_V_ERR_*` definitions do not match this table. Examples:

- `37` is `INVALID_NON_CA`, not a generic policy failure;
- `38` is `PROXY_PATH_LENGTH_EXCEEDED`, not policy;
- `39` is `KEYUSAGE_NO_DIGITAL_SIGNATURE`, not policy;
- `41` is `INVALID_EXTENSION`, not a signature-algorithm failure;
- `42` is `INVALID_POLICY_EXTENSION`, which belongs under policy;
- `49` is `SUBTREE_MINMAX`, which is a Name Constraints failure;
- `50` is generic `APPLICATION_VERIFICATION`, not Name Constraints;
- `54` is `CRL_PATH_VALIDATION_ERROR`, not Name Constraints;
- `56` is `SUITE_B_INVALID_VERSION`, not an incomplete path;
- `58` is `SUITE_B_INVALID_CURVE`, which is a key/SPKI class rather than signature algorithm;
- `75` is `OCSP_CERT_UNKNOWN`, not signature algorithm;
- `76` is `UNSUPPORTED_SIGNATURE_ALGORITHM`, not issuer-key/SPKI.

The focused test currently only repeats these incorrect assignments, so it cannot catch the problem. Please map only codes with an exact neutral semantic equivalent and leave generic/out-of-scope codes as `unclassified_rejection` rather than forcing agreement. Add the symbolic `X509_V_ERR_*` name beside every numeric entry so the mapping is auditable. At minimum, `39 -> key_usage_failure`, `42 -> policy_failure`, `38 -> path_length_violation`, `49 -> name_constraints_violation`, `58 -> issuer_key_or_spki_invalid`, and `76/77/78 -> signature_algorithm_invalid`; generic application/CRL/OCSP codes should not be classified as PKI rules they do not represent.

tests/pki_differential.zig
```diff
@@ -315,26 +435,94 @@ fn opensslDecision(allocator: std.mem.Allocator, case: manifest.Case, deadline_m
...
+        .reason = classified.reason,
+        .diagnostic = diagnostic,
     };
 }
 
+const OpenSSLClassification = struct { status: Status, reason: Reason };
+
+fn parseOpenSSLVerifyCode(diagnostic: []const u8) ?u16 {
+    var rest = diagnostic;
+    while (std.mem.indexOf(u8, rest, "error ")) |offset| {
```
Jaruso: **Blocking: the numeric-code parser is spoofable by certificate-controlled output.** `openssl verify` prints certificate/subject information before its actual error line. A hostile subject containing text such as `error 62 at 0 depth lookup` will be found before the real verification error and can force an `identity_mismatch` classification.

Parse line-by-line and accept only the exact OpenSSL verification-error grammar, for example a trimmed line beginning with `error ` followed by digits and ` at <depth> depth lookup:`. Parse the raw bounded stdout and stderr independently instead of first discarding one stream in `externalDiagnostic`.

Add a regression test like:

```zig
const diagnostic =
    "CN = attacker error 62 at 0 depth lookup\n" ++
    "error 47 at 1 depth lookup: permitted subtree violation\n";
try expectEqual(@as(?u16, 47), parseOpenSSLVerifyCode(diagnostic));
```

The hostile corpus must not be able to choose its own semantic reason through display text.

tests/pki_differential.zig
```diff
@@ -99,27 +104,107 @@ const ParsedChain = struct {
...
+        error.MalformedExtension,
+        error.NamePreparationFailed,
+        error.NoCertificates,
+        error.MalformedPemBoundary,
+        error.MismatchedPemLabel,
+        error.UnterminatedPemBlock,
+        error.InvalidPemBase64,
+        error.EmptyPemBlock,
+        => .malformed_certificate,
+        else => if (isResourceLimitError(err)) .resource_limit else .unclassified_rejection,
```
Jaruso, line 199: **Blocking: fixture/filesystem failures are currently classified as certificate rejections.** `rejectionFromError` is used around `trust_store.Snapshot.loadFiles`, `loadChainPemFile`, and intermediate loading. Therefore `FileNotFound`, `AccessDenied`, `NotDir`, read failures, and similar harness/setup errors reach this fallback and become `.reject/.unclassified_rejection`. The new test explicitly codifies `error.NotDir` as a rejection.

These are not validator decisions. They must be `.tool_failure` with a dedicated bounded reason such as `harness_io_failure`/`fixture_load_failure` (or another clearly named tool reason). Reserve `unclassified_rejection` for a completed parsing/path/validation operation whose certificate failure cannot yet be classified.

Please add focused missing-root, missing-leaf, unreadable-input, and non-file-path tests and assert they cannot enter reduction or satisfy a manifest expectation.

tests/pki_differential.zig
```diff
@@ -1374,9 +1607,9 @@ fn runCase(allocator: std.mem.Allocator, runtime_identity: RuntimeIdentity, case
     var go = try goDecision(allocator, case, deadline_ms);
     defer go.deinit(allocator);
 
-    const matches = tardigrade.status == expectedStatus(case.expected.tardigrade) and
-        openssl.status == expectedStatus(case.expected.openssl) and
-        go.status == expectedStatus(case.expected.go);
+    const matches = observationMatchesExpected(tardigrade, case.expected.tardigrade) and
```
Jaruso, line 1731: **Blocking: a tool failure can currently produce a `promotable` reduced regression seed.** After any mismatch, this path always calls `minimizeCase`. If OpenSSL is missing/times out or Go emits malformed JSON, every reduced candidate can preserve that same tool-failure tuple. `observationsPreserveTarget` then returns true and `reductionIsPromotable` has no guard against tool failures or `unclassified_rejection`.

That means an environment outage can manufacture a checked-in fuzz/unit seed even though no certificate-semantic disagreement occurred.

Gate minimization on all three original observations being classified certificate decisions, and defensively make promotion require the original and reduced observations to contain no `.tool_failure` and no `.unclassified_rejection`. Tool failures should still write one bounded reproduction artifact, but with `reduction = null`/`promotable = false`.

Add drills for OpenSSL timeout, Go malformed output, and an unclassified rejection proving no reducer/oracle loop runs and no seed is marked promotable.

tests/pki_differential.zig
```diff
@@ -315,26 +435,94 @@ fn opensslDecision(allocator: std.mem.Allocator, case: manifest.Case, deadline_m
...
+fn classifyOpenSSLRejection(case: manifest.Case, diagnostic: []const u8) OpenSSLClassification {
+    if (parseOpenSSLVerifyCode(diagnostic)) |code| return .{
+        .status = .reject,
+        .reason = opensslReasonForCode(code),
+    };
+    // OpenSSL's STORE layer reports a committed, present certificate whose
+    // DER cannot be decoded as "Could not find certificate file". Restrict
+    // this exception to manifest-declared malformed inputs; missing files,
+    // command usage failures, and setup errors remain harness failures.
+    if (case.category == .malformed_der and
```
Jaruso: **Blocking: the malformed-category exception can mask a setup/file failure.** This branch runs before the generic `Could not find certificate file` tool-failure check and relies only on `case.category == .malformed_der`. A missing, moved, unreadable, or wrong-type root/intermediate/leaf in a malformed case can emit the same OpenSSL STORE message and be accepted as `.reject/.malformed_der`.

The category is an expectation, not proof that OpenSSL successfully opened the intended leaf and rejected its DER. Please distinguish the failing path and prove it is the present target certificate, or use a dedicated bounded parse probe for the committed leaf before classifying this diagnostic. Add a malformed-category test with a missing root and a missing leaf; both must remain tool failures, while the committed malformed leaf is the only case classified `malformed_der`.

Jaruso (commented):
**Blocking first-slice review — the semantic architecture is promising, but it is not sound enough to build the hostile corpus on yet.**

The structured `Observation`, reason-aware manifest expectations, Go JSON contract, schema-v4 artifacts, and removal of diagnostic prose from reduction identity are all the right direction. I reviewed this as the first logical #348 slice rather than expecting the still-deferred fixture expansion.

The blockers below are semantic correctness issues in the foundation itself:

1. The OpenSSL numeric table assigns several official `X509_V_ERR_*` codes to unrelated neutral reasons, so a later hostile case can falsely appear to agree.
2. The OpenSSL error-code parser can be spoofed by certificate-controlled diagnostic text because it scans for the first `error N at` substring anywhere.
3. Tardigrade fixture/filesystem failures are represented as certificate rejections instead of harness failures.
4. Tool failures and unclassified observations are still minimized and can be marked promotion-ready, allowing an OpenSSL/Go outage to manufacture a regression seed.
5. The malformed-DER OpenSSL exception can turn a missing/unreadable input into an expected certificate rejection.

Please fix these before adding the larger DER and algorithm corpus; otherwise the new fixtures will encode incorrect semantic expectations. Also note that CI run 795 was cancelled before the PKI differential core job executed, so the new end-to-end oracle path has not yet been demonstrated by GitHub Actions at this head.

tests/pki_differential.zig
```diff
@@ -315,26 +481,175 @@ fn opensslDecision(allocator: std.mem.Allocator, case: manifest.Case, deadline_m
...
+        };
+    }
+    // OpenSSL's STORE layer reports a committed, present certificate whose
+    // DER cannot be decoded as "Could not find certificate file". Restrict
+    // this exception to manifest-declared malformed inputs; missing files,
+    // command usage failures, and setup errors remain harness failures.
+    if (std.mem.indexOf(u8, diagnostic, "Could not find certificate file from ") != null or
+        std.mem.indexOf(u8, diagnostic, "Could not read certificate file from ") != null)
+    {
+        if (opensslCommittedParseFailureReason(case.id)) |reason| {
```
Jaruso: **Blocking: OpenSSL's semantic result is still being inferred from manifest metadata instead of validator output.**

`opensslCommittedParseFailureReason(case.id)` makes the same OpenSSL stdout/stderr classify differently depending on the test case name. This is already happening in the checked-in corpus: `algorithm-malformed-signature-bits.der` and `der-invalid-bit-string-unused.der` have the exact same Git blob SHA (`db33d6fcd4595785eb64eadae848a73a27ad8e88`), because both generators set the signature BIT STRING unused-bits byte to `8`. Yet the manifest expects OpenSSL `signature_algorithm_invalid` for one and `malformed_der` for the other. The case-ID lookup is the only thing manufacturing that distinction.

This also does not prove which input OpenSSL failed to parse. During reduced-chain verification, the STORE diagnostic may refer to a generated root or intermediate, while this code assigns the original leaf case's reason.

Please remove case-name-based classification. Parse and verify the failing path from the STORE diagnostic and classify only from output-derived evidence. If OpenSSL exposes no finer distinction, use one neutral parse reason such as `malformed_der`/`malformed_certificate` consistently for all equivalent STORE parse failures. Add invariants proving:

```zig
// Identical oracle output must have identical semantics regardless of case metadata.
try expectEqual(
    classifyOpenSSLRejection(case_a, stdout, stderr, diagnostic),
    classifyOpenSSLRejection(case_b, stdout, stderr, diagnostic),
);
```

Also reject duplicate DER payloads in the generated corpus unless they are an explicitly documented alias with the same expected observation.

tests/vectors/pki/generate.go
```diff
@@ -390,3 +494,312 @@ func encodeBundle(certificates ...[]byte) []byte {
...
+func mutateSignatureAlgorithms(der, algorithm []byte) ([]byte, error) {
+	parts, err := splitCertificate(der)
+	if err != nil {
+		return nil, err
+	}
+	tbs, err := replaceNthInSequence(parts.tbs, ed25519AlgorithmIdentifier, algorithm, 0)
+	if err != nil {
+		return nil, err
+	}
+	return joinCertificate(tbs, algorithm, parts.signatureValue), nil
```
Jaruso: **Blocking: the algorithm corpus does not isolate several of the rules it claims to test.**

This helper changes the inner signature `AlgorithmIdentifier` inside the TBS certificate and then reuses the original `signatureValue`. `mutateSPKIAlgorithm` and `mutateSPKIKeyEncoding` do the same for SPKI fields. Every such TBS mutation invalidates the issuer signature before considering the intended rule.

That produces false coverage. For example, `algorithm-incompatible-spki-key` is expected to fail as `signature_invalid` in Tardigrade/OpenSSL, so it is effectively another corrupt-certificate-signature vector rather than proof of an SPKI/key-encoding failure. The RSA-PSS and RSA-NULL cases also use the existing Ed25519 issuer, so they primarily exercise algorithm/key incompatibility and cannot establish the claimed RSA parameter behavior.

Please generate one-variable fixtures:

- Re-sign modified Ed25519 TBS certificates with the deterministic intermediate key for illegal-parameter and SPKI mutations.
- Generate a deterministic RSA issuer/leaf chain for RSA-PSS and RSA `NULL`-versus-absent cases, then mutate only the parameter encoding while retaining a compatible issuer/key and a valid signature for the control case.
- Assert the unmutated control validates and that the intended mutation changes the targeted semantic reason.
- Rename or remove any case that intentionally tests only issuer-key mismatch or stale-signature rejection instead of the advertised parameter/SPKI rule.

Until the fixtures isolate their intended condition, the PR cannot claim the algorithm-confusion corpus is complete.

Jaruso (commented):
Blocking follow-up review on the completed #348 corpus.

The five original review areas were rechecked at the new head. Numeric OpenSSL mappings, line-oriented error parsing, fixture/filesystem tool-failure classification, minimization gating, and promotion defenses are materially improved, and CI—including the core differential job—is green.

Two semantic-integrity blockers remain in the newly added corpus. Both can produce green tests without proving the certificate rule that the manifest says was exercised:

1. OpenSSL STORE parse failures are classified from `case.id`, not from the validator output. The corpus currently contains two byte-identical DER fixtures with different expected OpenSSL reasons, and the only reason both pass is this case-name lookup.
2. Several algorithm/SPKI mutations alter the signed TBS certificate but retain the original signature. Consequently, rejection may be caused by the stale signature or incompatible issuer key rather than the intended parameter/SPKI rule. Some manifest expectations explicitly accept that confounded result.

Please address both before treating the hostile algorithm/DER corpus and #348 closure audit as complete. GitHub does not permit a formal request-changes review on my own PR, so this is submitted as a blocking comment review.

tests/pki_differential.zig
```diff
@@ -315,26 +481,156 @@ fn opensslDecision(allocator: std.mem.Allocator, case: manifest.Case, deadline_m
...
+        28, // X509_V_ERR_CERT_REJECTED
+        29, // X509_V_ERR_SUBJECT_ISSUER_MISMATCH
+        30, // X509_V_ERR_AKID_SKID_MISMATCH
+        31, // X509_V_ERR_AKID_ISSUER_SERIAL_MISMATCH
+        => .untrusted_or_incomplete_path,
+        4, // X509_V_ERR_UNABLE_TO_DECRYPT_CERT_SIGNATURE
+        7, // X509_V_ERR_CERT_SIGNATURE_FAILURE
+        => .signature_invalid,
+        6, // X509_V_ERR_UNABLE_TO_DECODE_ISSUER_PUBLIC_KEY
+        24, // X509_V_ERR_NO_ISSUER_PUBLIC_KEY
```
Jaruso: **Blocking: OpenSSL verification code 24 is still misidentified.** OpenSSL's stable `verify` output and current documentation use `error 24 ... invalid CA certificate`; code 24 is `X509_V_ERR_INVALID_CA`, not `X509_V_ERR_NO_ISSUER_PUBLIC_KEY`. Mapping it to `issuer_key_or_spki_invalid` can manufacture the wrong semantic agreement, and the focused table test currently locks that mistake in.

Please remove `24` from this SPKI group. The safest narrow fix is to leave it `unclassified_rejection` until the taxonomy has an exact CA/basic-constraints class; alternatively map it only to a deliberately documented existing class that also matches Tardigrade's `issuer_is_not_ca` treatment. Update the test so `24` can never regress back to an issuer-key/SPKI result.

tests/vectors/pki/generate.go
```diff
@@ -229,6 +297,15 @@ func generate(out string) error {
 		return fmt.Errorf("truncated DER seed unexpectedly parsed in Go")
 	}
 
+	algorithmFixtures, err := hostileAlgorithmFixtures(valid.der, intermediateKey.private)
+	if err != nil {
+		return fmt.Errorf("create algorithm fixtures: %w", err)
+	}
+	derFixtures, err := hostileDERFixtures(valid.der, intermediateKey.private)
+	if err != nil {
+		return fmt.Errorf("create DER fixtures: %w", err)
```
Jaruso, line 307: **Blocking: the duplicate-hostile-payload invariant requested in the prior review is still missing.** Removing the one known duplicate fixes today's corpus, but `algorithmFixtures` and `derFixtures` can again emit identical DER under different case names and the generator will happily publish both with independently declared expectations.

Please add a deterministic raw-DER uniqueness check before writing files. A platform-independent approach is to hash each entry in the two hostile maps with SHA-256, retain the first name for each digest, and fail with both names when a duplicate appears. Only an explicitly documented alias with identical expectations should bypass it.

Also add the focused OpenSSL classification invariant requested previously: the same STORE stdout/stderr must produce the same `OpenSSLClassification` for two different `manifest.Case` values. The implementation currently ignores `case`, which is good, but the unused parameter/helper shape and absent regression test leave the original failure mode easy to reintroduce.

Jaruso (commented):
Re-review of `d1bd372a`.

The previous structural blockers are substantively addressed: fixture I/O failures are tool failures, OpenSSL output parsing is line-anchored and stream-aware, non-certificate outcomes cannot enter reduction/promotion, the false-agreement Name Constraints cases are replaced with isolated fixtures, modified TBS certificates are re-signed, and the confounded algorithm cases were removed.

Two narrow blockers remain:

1. OpenSSL numeric code 24 is still assigned the wrong symbolic meaning and semantic class.
2. The known duplicate hostile fixture was removed, but the requested generator invariant preventing future byte-identical hostile DER cases is still absent. The same-output/case-metadata invariance should also be pinned with a focused test while this seam is being closed.

The PKI differential core CI job passes at this head. The overall CI run is currently red only in the Security job's GitHub Actions hardening step; this PR does not modify workflows, but the run still needs to be rerun or diagnosed before merge.

After the two code/test corrections and a green CI run, I do not see another implementation phase required for this PR.

Jaruso (commented):
Re-review complete at `fe566e346ee4c437c7dcea1ae688f3d9645406f8`: **no blocking findings remain**.

The final correction set addresses both outstanding issues:

- OpenSSL code 24 is now correctly identified as `X509_V_ERR_INVALID_CA` and placed in the existing issuer-authorization bucket used consistently with Tardigrade's `issuer_is_not_ca`/Go `NotAuthorizedToSign` treatment; it can no longer regress to issuer-key/SPKI classification.
- Hostile DER payload uniqueness is enforced before publication in the deterministic Go generator, independently checked by the shell regeneration workflow, and pinned against manifest case/file metadata in the Zig suite.
- OpenSSL STORE classification no longer accepts manifest metadata at all, and the focused invariant proves identical output has identical semantics.
- The process-cleanup probe now allows bounded reaping latency without weakening the assertion that parent and descendant processes are gone.

I also rechecked the earlier semantic-model, fixture-loading, reduction/promotion, Name Constraints, algorithm-fixture, DER-corpus, promoted-seed, and artifact changes. The latest GitHub CI run is fully green, including the PKI differential core job, unit-test platforms, lint/format, Security, builds, packaging, and dependency checks.

Because the authenticated account is the PR author, GitHub does not allow a formal self-approval. Treat this as a clean approval-equivalent review. The PR is technically merge-ready after it is marked ready for review. If this is intended to close the completed story, change `Refs #348` to `Closes #348` before merging.

PR #474, opened 2026-07-20
tls: complete native backpressure production wiring

Jaruso:
## Summary

Refs #355.

This PR completes the remaining production wiring after merged PRs #457 and #470. It does not reimplement the record-layer queue model, ALPN dispatch, credential/provider work, global #140 proxy accounting, or any new TLS crypto behavior.

## BufferLimits model and defaults

- Native pure-Zig TLS listeners now receive an immutable `encrypted_stream.BufferLimits` policy from `EdgeConfig`.
- Defaults come from `BufferLimits.defaults()` so operator config and stream construction share the TLS core defaults.
- Existing `NativeTlsConnection.create()` remains source-compatible and delegates to `createWithOptions(..., .{})`.

## Compile-time capacity vs runtime limits

- Runtime low/high/hard watermarks are validated by TLS core before stream use.
- Invalid ordering, over-capacity limits, or reserve violations reject startup/reload before replacing the active config.
- OpenSSL is not presented as enforcing these native pure-Zig limits.

## Queue ownership/accounting

| Queue | Ownership source | Runtime metric label |
| --- | --- | --- |
| inbound carrier ciphertext | stream-owned pure-Zig queue/parser bytes | `queue="inbound_ciphertext"` |
| decrypted plaintext | stream-owned pure-Zig plaintext queue, or measurable OpenSSL pending plaintext | `queue="inbound_plaintext"` |
| outbound ciphertext | stream-owned pure-Zig ciphertext queue | `queue="outbound_ciphertext"` |
| handshake bytes | stream-owned pure-Zig handshake queue | `queue="handshake"` |

Pure-Zig writes continue sealing accepted plaintext directly into outbound ciphertext; no hidden pending-plaintext queue was added.

## Readiness and hysteresis state machine

- Encrypted raw fd registration now comes only from `Readiness.wants_read` and `Readiness.wants_write`.
- `can_read_plaintext` and `can_write_plaintext` are treated as immediate application work, not socket interest.
- The event-loop rearm path and synchronous `poll()` path share the same `EncryptedWait` mapping and no longer OR/fallback to nominal application directions.
- Empty-interest states are represented as stalled/no registration rather than passed to the event loop.

## Hard-limit and atomicity behavior

This PR preserves the #457 record-layer hard-limit and atomicity behavior. The new production config path validates policies before constructing native streams and passes the validated policy into `initWithCarrierBackendAndLimits()`.

## Metrics and snapshot interface

- Runtime metrics now aggregate `EncryptedStream.bufferSnapshot()` through per-connection last-observed state.
- Current byte gauges are replaced by subtract/add observation and released exactly once on managed connection teardown.
- Counter deltas are monotonic and are not double-counted on repeated observation.
- Prometheus series added:
  - `tardigrade_tls_buffered_bytes_current{backend,queue}`
  - `tardigrade_tls_buffer_pause_events_total{backend,direction}`
  - `tardigrade_tls_buffer_resume_events_total{backend,direction}`
  - `tardigrade_tls_buffer_limit_exceeded_total{backend,queue}`
  - `tardigrade_tls_buffer_stalled_drives_total{backend}`
  - `tardigrade_tls_buffer_config_limit_bytes{queue,limit}`

Labels are fixed to bounded backend, queue, direction, and limit enums. No SNI, URL, IP, request ID, connection ID, or stream ID labels were added.

## OpenSSL accounting boundary

OpenSSL snapshots remain limited to adapter/BIO-measurable state and `backend_opaque` accounting. Native TLS config-limit metrics describe the pure-Zig listener policy only and are not reported as OpenSSL-enforced limits.

## #356 HTTP consumer seam

HTTP/1.1 and HTTP/2 encrypted consumers can use the backend-neutral readiness mapping without backend-specific casts. Blocked plaintext reads/writes register only the TLS raw direction that can make progress.

## Validation

- `git diff --check`
- `zig build test-tls --summary all --error-style verbose` (325/325 passed)
- `zig build test-quic --summary all --error-style verbose` (392/392 passed)
- `zig build test-integration --summary all --error-style verbose` (72/89 passed, 17 skipped)
- `zig build test-integration-native-tls -Dtls-profile=appliance --summary all --error-style verbose` (8/8 passed)
- `zig build test -Dtls-profile=appliance --summary all --error-style verbose` (1644/1645 passed, 1 skipped)
- `zig build test --summary all --error-style verbose` (1672/1673 passed, 1 skipped)
- `zig build -Doptimize=ReleaseFast -Dtls-profile=appliance --summary all --error-style verbose`

## Explicit exclusions

This PR does not implement #334, #347, #348, #356 ALPN/runtime expansion, global/per-origin #140 proxy accounting, KeyUpdate, new TLS queues, dynamic hot-path growth, or new cryptographic algorithms.

### DIFF
--- a/src/edge_config.zig
+++ b/src/edge_config.zig
@@ -1182,7 +1182,7 @@ pub fn loadFromEnv(allocator: std.mem.Allocator) !EdgeConfig {
 
     const request_total_timeout_ms = parseIntEnv(u32, allocator, "TARDIGRADE_REQUEST_TOTAL_TIMEOUT_MS", 0);
     const tls_handshake_timeout_ms = parseIntEnv(u32, allocator, "TARDIGRADE_TLS_HANDSHAKE_TIMEOUT_MS", 5_000);
-    const tls_buffer_limits = tlsBufferLimitsFromEnv(allocator);
+    const tls_buffer_limits = try tlsBufferLimitsFromEnv(allocator);
     const downstream_write_timeout_ms = parseIntEnv(u32, allocator, "TARDIGRADE_DOWNSTREAM_WRITE_TIMEOUT_MS", 30_000);
 
     const max_req_conn_str = envOrDefault(allocator, "TARDIGRADE_MAX_REQUESTS_PER_CONNECTION", "100") catch unreachable;
@@ -2886,21 +2886,37 @@ fn validateBufferedUpstreamResponseLimit(limit: usize) !void {
     if (limit == 0) return error.InvalidConfigValue;
 }
 
-fn tlsBufferLimitsFromEnv(allocator: std.mem.Allocator) encrypted_stream.BufferLimits {
+fn tlsBufferLimitsFromEnv(allocator: std.mem.Allocator) !encrypted_stream.BufferLimits {
     const defaults = encrypted_stream.BufferLimits.defaults();
+    const limits = encrypted_stream.BufferLimits{
+        .inbound_ciphertext = try tlsWatermarkFromEnv(allocator, "TARDIGRADE_TLS_INBOUND_CIPHERTEXT", defaults.inbound_ciphertext),
+        .inbound_plaintext = try tlsWatermarkFromEnv(allocator, "TARDIGRADE_TLS_INBOUND_PLAINTEXT", defaults.inbound_plaintext),
+        .outbound_ciphertext = try tlsWatermarkFromEnv(allocator, "TARDIGRADE_TLS_OUTBOUND_CIPHERTEXT", defaults.outbound_ciphertext),
+        .handshake = try tlsWatermarkFromEnv(allocator, "TARDIGRADE_TLS_HANDSHAKE", defaults.handshake),
+    };
+    limits.validate() catch |err| {
+        logConfigDiagnostic("config validation failed: TARDIGRADE_TLS_* buffer limits are invalid: {}", .{err});
+        return error.InvalidConfigValue;
+    };
+    return limits;
+}
+
+fn tlsWatermarkFromEnv(allocator: std.mem.Allocator, comptime prefix: []const u8, defaults: encrypted_stream.Watermark) !encrypted_stream.Watermark {
     return .{
-        .inbound_ciphertext = tlsWatermarkFromEnv(allocator, "TARDIGRADE_TLS_INBOUND_CIPHERTEXT", defaults.inbound_ciphertext),
-        .inbound_plaintext = tlsWatermarkFromEnv(allocator, "TARDIGRADE_TLS_INBOUND_PLAINTEXT", defaults.inbound_plaintext),
-        .outbound_ciphertext = tlsWatermarkFromEnv(allocator, "TARDIGRADE_TLS_OUTBOUND_CIPHERTEXT", defaults.outbound_ciphertext),
-        .handshake = tlsWatermarkFromEnv(allocator, "TARDIGRADE_TLS_HANDSHAKE", defaults.handshake),
+        .low = try parseTlsLimitEnv(allocator, prefix ++ "_LOW_WATERMARK_BYTES", defaults.low),
+        .high = try parseTlsLimitEnv(allocator, prefix ++ "_HIGH_WATERMARK_BYTES", defaults.high),
+        .hard = try parseTlsLimitEnv(allocator, prefix ++ "_HARD_LIMIT_BYTES", defaults.hard),
     };
 }
 
-fn tlsWatermarkFromEnv(allocator: std.mem.Allocator, comptime prefix: []const u8, defaults: encrypted_stream.Watermark) encrypted_stream.Watermark {
-    return .{
-        .low = parseIntEnv(usize, allocator, prefix ++ "_LOW_WATERMARK_BYTES", defaults.low),
-        .high = parseIntEnv(usize, allocator, prefix ++ "_HIGH_WATERMARK_BYTES", defaults.high),
-        .hard = parseIntEnv(usize, allocator, prefix ++ "_HARD_LIMIT_BYTES", defaults.hard),
+fn parseTlsLimitEnv(allocator: std.mem.Allocator, key: []const u8, default_value: usize) !usize {
+    const raw = envOrDefault(allocator, key, "") catch return default_value;
+    defer allocator.free(raw);
+    const value = std.mem.trim(u8, raw, " \t\r\n");
+    if (value.len == 0) return default_value;
+    return std.fmt.parseInt(usize, value, 10) catch {
+        logConfigDiagnostic("config validation failed: {s} must be a non-negative integer byte count", .{key});
+        return error.InvalidConfigValue;
     };
 }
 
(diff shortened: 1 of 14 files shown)


Jaruso (commented):
## Blocking review — not ready to merge

The core direction is correct and CI is green, but this does not yet satisfy the reopened #355 pickup contract. I found three independent production correctness gaps:

1. A native handshake parked on an asynchronous credential/sign/verify operation can enter `.stalled` with no fd registration and no non-socket wake path. The operation may complete, but `drive()` is never called again; the connection remains inert until timeout. The existing handshake code also refreshes the deadline on every no-progress drive, so a naïve periodic poll would make the timeout sliding rather than absolute.
2. The new environment values use `parseIntEnv`, which silently substitutes the default on malformed, negative, or overflowing input. A bad startup/reload value therefore does not fail validation; a reload can successfully replace a previously valid custom policy with defaults.
3. TLS metrics are sampled only from `ManagedConnection.updateInterestFor()` and teardown. Long-lived native HTTP/2 and synchronous HTTP/1.1 wait/write paths do not rearm through that method, so `*_bytes_current` is stale or permanently zero while buffering actually occurs. The OpenSSL serving/parking path never owns this per-connection metric state at all, so every `backend="openssl"` series is synthetic zero rather than an aggregation of its measurable snapshot.

The required acceptance evidence is also missing. No integration test file changed, and the PR does not prove the reopened issue's required cases: invalid startup/reload preserving the previous snapshot, async-auth progress without readiness spin, slow HTTP/1.1 and HTTP/2 current-byte accounting, and exactly-once zeroing across timeout, peer abort, cancellation, handshake failure, and graceful close.

I left detailed expected fixes and test shapes inline. The production `BufferLimits` construction, immutable per-connection copy, bounded label sets, config-limit export, and raw-direction mapping itself otherwise look sound.

src/edge_gateway.zig
```diff
@@ -1251,12 +1255,20 @@ fn reapActiveConnections(
 
 fn rearmActiveConnection(ctx: *WorkerContext, managed: *http.downstream_connection.ManagedConnection, requested: http.event_loop.Interest) void {
     const fd = managed.fd;
+    if (managed.encryptedWaitFor(requested)) |wait| switch (wait) {
+        .ready_now => {
+            advanceActiveConnection(ctx, managed);
+            return;
+        },
+        .socket, .stalled => {},
```
Jaruso: **Blocking: `.stalled` currently strands asynchronous authentication until the handshake timeout.**

While `PureZigRecordStream` has a pending credential/sign/verify operation, `readiness()` intentionally suppresses raw reads and writes and `drive()` is the method that polls `resumeAuthIfPending()`. This branch treats `.stalled` like a socket wait, inserts the connection back into `ActiveRegistry`, and then (because the computed interest is empty) registers no event. The timer path only calls `reapActiveConnections()`; it never redispatches a non-expired stalled connection. If the provider completes one millisecond later, nothing calls `drive()` again.

This is specifically contrary to #355's stalled-handshake/async-auth acceptance case. Please add an explicit non-socket progression source rather than restoring nominal `POLLIN`/`POLLOUT`. For example, expose a backend-neutral `needs_drive_poll`/timer-wait state, retain the connection in the active registry with a bounded next-poll deadline, and have the existing 250 ms timer dispatch due connections back to a worker:

```zig
pub const EncryptedWait = union(enum) {
    ready_now,
    socket: event_loop.Interest,
    retry_at_ms: u64,
    stalled,
};

// Timer tick: submit only connections whose retry_at_ms is due.
// Never subscribe to a raw direction that Readiness.wants_* suppressed.
```

Also preserve an **absolute** handshake deadline. `advanceNativeHandshake()` currently assigns `now + handshake_timeout_ms` after every no-progress drive; once timer polling exists, that would extend a never-completing auth operation forever. Initialize the deadline once and cancel/release the pending operation when that original deadline expires.

Add a deterministic fake pending operation that returns pending for multiple polls, completes without any fd event, and proves: bounded poll count/no busy loop, eventual handshake progress, unchanged absolute deadline, and exactly-once cancel/release when it never completes.

src/edge_config.zig
```diff
@@ -2875,6 +2886,24 @@ fn validateBufferedUpstreamResponseLimit(limit: usize) !void {
...
+        .inbound_ciphertext = tlsWatermarkFromEnv(allocator, "TARDIGRADE_TLS_INBOUND_CIPHERTEXT", defaults.inbound_ciphertext),
+        .inbound_plaintext = tlsWatermarkFromEnv(allocator, "TARDIGRADE_TLS_INBOUND_PLAINTEXT", defaults.inbound_plaintext),
+        .outbound_ciphertext = tlsWatermarkFromEnv(allocator, "TARDIGRADE_TLS_OUTBOUND_CIPHERTEXT", defaults.outbound_ciphertext),
+        .handshake = tlsWatermarkFromEnv(allocator, "TARDIGRADE_TLS_HANDSHAKE", defaults.handshake),
+    };
+}
+
+fn tlsWatermarkFromEnv(allocator: std.mem.Allocator, comptime prefix: []const u8, defaults: encrypted_stream.Watermark) encrypted_stream.Watermark {
+    return .{
+        .low = parseIntEnv(usize, allocator, prefix ++ "_LOW_WATERMARK_BYTES", defaults.low),
```
Jaruso: **Blocking: malformed TLS limit values silently become defaults instead of rejecting startup/reload.**

`parseIntEnv()` catches every `parseInt` failure and returns `default_value`. Consequently values such as `garbage`, `-1`, or an overflowing integer produce a valid default `BufferLimits`; `validate()` cannot tell that configuration was invalid. During hot reload this can successfully replace a previously configured policy with defaults, violating the issue requirement that invalid input reject the reload and preserve the active snapshot.

Make this path fallible and distinguish absent/empty from present-but-invalid input, for example:

```zig
fn parseTlsLimit(
    allocator: std.mem.Allocator,
    key: []const u8,
    default_value: usize,
) !usize {
    const raw = try envOrDefault(allocator, key, "");
    defer allocator.free(raw);
    if (raw.len == 0) return default_value;
    return std.fmt.parseInt(usize, std.mem.trim(u8, raw, " \t\r\n"), 10) catch {
        logConfigDiagnostic("config validation failed: {s} must be a non-negative integer byte count", .{key});
    };
}
```

Then make `tlsBufferLimitsFromEnv()` / `tlsWatermarkFromEnv()` return errors and use `try` in `loadFromEnv()`.

Please test explicit custom values, non-numeric input, negative input, overflow, each `low < high <= hard` violation, each capacity/reserve violation, and a rejected hot reload that leaves both the active generation and exported TLS limit snapshot unchanged.

src/http/downstream_connection.zig
```diff
@@ -286,19 +302,51 @@ pub const ManagedConnection = struct {
     pub fn updateInterestFor(self: *ManagedConnection, requested: event_loop.Interest) event_loop.Interest {
         self.interest = switch (self.transport) {
             .plaintext => requested,
-            else => combinedInterest(requested, self.transport.readiness()),
+            else => switch (encryptedWaitForInterest(self.transport.readiness(), requested)) {
+                .socket => |interest| interest,
+                .ready_now, .stalled => .{},
+            },
         };
+        self.observeTlsBufferMetrics();
```
Jaruso, line 328: **Blocking: sampling only on event-loop rearm/teardown does not implement a truthful `*_bytes_current` gauge, and OpenSSL never contributes.**

Native HTTP/2 enters `handleHttp2Connection()` through `WaitingEncryptedHttpConnection` and remains in synchronous read/write/poll loops until the connection closes. Slow HTTP/1.1 writes use the same waiting adapter. Neither path calls `ManagedConnection.updateInterestFor()` while TLS queues fill and drain. At teardown this method observes and immediately releases under the same mutex, so a scrape can never see those transient bytes. A prior rearm observation can also remain stale after the worker drains the queue.

The OpenSSL path is separate: it dispatches/parks a `TlsConnection` and never constructs a `ManagedConnection` carrying `tls_buffer_metrics`. Thus all exported `backend="openssl"` values remain zero even when `SSL_pending()` or retry-derived counters are measurable.

Please move observation to the actual encrypted-I/O progression boundary—after every relevant `read`, `write`, and `drive`/flush result—or introduce a stable runtime registry that the metrics scrape can snapshot. It must cover both native and OpenSSL connection ownership paths. Repeated observations should apply counter deltas once and replace each connection's byte contribution; teardown should release it exactly once.

Also avoid silently saturating aggregate subtraction to zero: with multiple connections, an inconsistent release would erase another connection's contribution. Report/assert `BufferAccountingUnderflow` (as the existing proxy accounting does) and add a two-connection regression.

Required integration shape: block a native HTTP/1.1 writer and a long-lived HTTP/2 writer with small limits, scrape a nonzero current gauge and one pause transition while blocked, drain below low and observe zero/resume once, then repeat teardown via success, timeout, peer abort, cancellation, handshake failure, and graceful close. Exercise an OpenSSL retry/pending case and assert only adapter-measurable fields are exported under `backend="openssl"`.

src/edge_gateway.zig
```diff
@@ -1249,14 +1258,51 @@ fn reapActiveConnections(
...
+) void {
+    var fds: [32]std.posix.fd_t = undefined;
+    while (true) {
+        const count = active.dueDrivePollFds(http.event_loop.monotonicMs(), fds[0..]);
+        if (count == 0) break;
+        for (fds[0..count]) |fd| {
+            event_loop.remove(fd) catch {};
+            worker_pool.submit(fd) catch |err| {
+                state.logger.warn(null, "active connection drive-poll submit failed: {}", .{err});
+                if (active.checkout(fd)) |taken| {
```
Jaruso, line 1326: **Blocking: queuing only this fd leaves two independent owners and permits stale-fd dispatch.**

`ActiveRegistry.dueDrivePollFds()` only clears `drive_poll_due_ms`; it does not remove or mark the connection in-flight. After this `submit(fd)` succeeds, the same `ManagedConnection` remains eligible for `reapExpired()` immediately below and for shutdown/close paths while the worker item may sit in the queue.

A failing sequence is:

1. timer queues fd 42 for an auth retry;
2. before a worker dequeues it, the absolute handshake deadline passes;
3. `reapActiveConnections()` removes/deinitializes the registry entry and closes fd 42;
4. the queued task runs later; `handleAcceptedClient(42)` finds no active/parked connection and calls `startNewConnection(42)`;
5. fd 42 is now closed, or worse, has been reused for an unrelated socket.

The generic worker shutdown path likewise force-closes queued fds even though the registry still owns this connection, so later registry teardown can close the same descriptor again.

Please make work-item identity and ownership explicit. One acceptable shape is a tagged task:

```zig
const WorkItem = union(enum) {
    accepted: fd_t,
    parked_ready: fd_t,
    active_socket_ready: fd_t,
    active_drive_poll: ActiveToken,
};
```

Before submitting `.active_drive_poll`, atomically transition the registry entry to an in-flight/reserved state (or remove and transfer the `ManagedConnection` into a dedicated active-work queue). The reaper and shutdown code must skip or own that state consistently. The worker must never fall through from a missing active-drive item to `startNewConnection`, and it must re-check the connection's absolute deadline after dequeue before calling `drive()`.

Add a deterministic test that holds the worker queue past the handshake deadline, runs the reaper/shutdown path, then releases the queued retry. Assert that it neither starts a new connection nor double-closes a descriptor; include an fd-reuse regression so this cannot silently act on a different socket.

Jaruso (commented):
## Second review — one new blocking ownership race remains

The latest commit substantively addresses the three findings from my first pass:

- malformed/negative/overflowing TLS buffer values now fail rather than silently becoming defaults;
- async-auth stalls now receive a bounded timer-driven retry and the handshake deadline is no longer refreshed on every no-progress drive;
- native and OpenSSL metrics are observed at real I/O progression points, aggregate underflow is reported, and multi-connection accounting is covered.

However, the timer retry implementation introduces a new production race that must be fixed before merge.

`dueDrivePollFds()` marks a retry as dispatched but leaves the `ManagedConnection` in `ActiveRegistry`, while `dispatchDueActiveDrivePolls()` places only its numeric fd in the generic worker queue. The registry can subsequently expire, shut down, or otherwise destroy that connection before the queued worker starts. When the stale task eventually runs, `handleAcceptedClient()` finds no active or parked entry and falls through to `startNewConnection(fd)`, treating a closed—or already reused—descriptor as a fresh accepted socket. The generic worker shutdown path can also close the queued fd while `ActiveRegistry` still owns the same transport, creating a double-close/fd-reuse hazard.

There is a second deadline consequence: if the task dequeues after the absolute handshake deadline but before the reaper removes it, the worker checks it out and drives it without checking `expired(now)`, so handshake work may proceed after its deadline.

The retry dispatch must transfer or reserve connection ownership atomically, not enqueue an untyped bare fd while leaving a reapable owner behind. I left a concrete expected design and regression test inline.

The issue's acceptance evidence is also still incomplete. The new registry unit test manually sets `drive_poll_due_ms` on a plaintext connection; it does not exercise a pending credential/sign/verify operation. Please add a deterministic fake async-auth test proving completion without an fd event, bounded polling, unchanged absolute deadline, and exactly-once cancellation/release. The PR still needs production-path coverage for slow HTTP/1.1 and HTTP/2 live gauges plus teardown-to-zero behavior across timeout, peer abort, handshake failure, cancellation, and graceful close.

CI is green on the updated head, but it does not exercise this queued-retry ownership race.

Jaruso (commented):
## Third review — ownership race fixed; required acceptance coverage still missing

The latest commit correctly addresses the concrete ownership defect from my second review:

- work items are now tagged (`accepted`, `parked_ready`, `active_socket_ready`, `active_drive_poll`), so a missing active retry can no longer fall through to `startNewConnection()`;
- `ActiveRegistry` reserves queued active work and excludes it from deadline reaping while it is waiting in the worker queue;
- both socket-ready and timer-driven active work use that reservation;
- the worker checks the original absolute protocol deadline immediately after checkout and tears the connection down instead of driving an expired handshake;
- shutdown no longer closes registry-owned queued descriptors from the generic worker queue.

I do not see a remaining code-level defect in that repaired ownership path.

However, this still does **not** satisfy the test/acceptance contract written into reopened #355. The PR changes twelve documentation/runtime source files and no integration test file. The new tests validate registry flags and worker-item ownership independently, but they do not exercise the production composition that previously failed.

Before merge, add deterministic coverage for at least these remaining cases:

1. **Queued active retry across deadline:** hold the worker busy, queue a real `.active_drive_poll`, advance beyond the absolute handshake deadline, then release the worker. Assert the active item is checked out by tag, closed once, never routed to fresh accept handling, and cannot act on an fd reused by another socket.
2. **Actual asynchronous credential progression:** use a fake credential/sign/verify operation that remains pending for multiple timer polls and completes without any socket event. Prove bounded polling, eventual handshake progression, no deadline extension, and exactly-once cancellation/release on timeout.
3. **Rejected reload preservation:** start with a valid custom TLS policy, attempt a malformed/order/capacity-invalid reload through the real reload path, and verify the active generation plus exported config-limit metrics remain unchanged.
4. **Production HTTP metrics:** run slow native HTTP/1.1 and HTTP/2 readers/writers with small watermarks; scrape a nonzero current gauge and one pause transition while blocked, drain below low, and verify the gauge/resume transition. Exercise success, timeout, peer abort, local cancellation, handshake failure, and graceful close so per-connection contributions return to zero exactly once.
5. **OpenSSL boundary:** exercise a measurable OpenSSL pending/retry case and verify only adapter-measurable state is exported, with no native limit claims.

The aggregate CI suite is green, but those scenarios are not demonstrated by the changed tests. Once that acceptance coverage is present and green, the implementation itself looks suitable to merge.

src/tls/tls13_backend_tests.zig
```diff
@@ -2824,13 +2824,33 @@ test "the record stream production driver resumes async client authentication en
...
+    while (!h.client_engine.authPending()) {
+        const c = try h.driveClient();
+        const s = try h.driveServer();
+        if (!c.made_progress and !s.made_progress) return error.Stalled;
+    }
+    try std.testing.expectEqual(@as(usize, 0), client_credential.poll_count);
+
+    _ = try h.driveClient();
+    try std.testing.expect(h.client_engine.authPending());
+    try std.testing.expectEqual(@as(usize, 1), client_credential.poll_count);
```
Jaruso, line 2836: **Blocking coverage gap: this manually calls `driveClient()` and bypasses the gateway scheduler whose correctness this PR depends on.**

This proves that the lower-level record driver polls and completes the pending signer, which is useful, but it would still pass if `dispatchDueActiveDrivePolls()` never queued a retry, if `active_work_queued` stayed stuck, if the tagged worker item were lost, or if a queued retry were dispatched after its absolute deadline.

Please add a deterministic gateway-level test (or extract an injectable scheduler helper) that starts from a `ManagedConnection` whose real record engine has `authPending()`, observes no raw fd interest, queues `.active_drive_poll` only when `retry_at_ms` is due, completes the operation without a socket event, and proves the worker advances the same connection. A second case should hold the item past the original deadline and assert exactly-once cancellation/release with no fresh-accept fallthrough or fd-reuse action.

Jaruso (commented):
## Fourth review — useful coverage added, but production-path acceptance is still incomplete

The new commit improves the PR in two real ways:

- the reload integration test now proves malformed, ordering-invalid, and over-capacity TLS limit reloads fail while the previously exported active limit remains in place;
- the record-stream test now proves a mock asynchronous signer remains pending for two polls, completes on the third, releases once, and is not cancelled on success.

I found no defect in those two tests and no production code changed in this revision.

However, the async test drives `SocketHarness` directly. It therefore does not exercise the code introduced by this PR to make async authentication progress in production: `EncryptedWait.retry_at_ms`, `ActiveRegistry.dueDrivePollFds()`, active-work reservation, the tagged `active_drive_poll` worker item, queue delay, post-dequeue deadline enforcement, or the guarantee that no socket readiness event is required. A defect in any of those pieces would still leave this test green.

The remaining acceptance evidence from the prior review is also still absent:

1. **Gateway timer/worker async-auth path.** Hold a real `.active_drive_poll` work item in the worker queue while the operation is pending, complete it without an fd event, then prove bounded polling, eventual progress before the original deadline, and exact cancellation/release when the original deadline expires. Include the queued-past-deadline and fd-reuse assertions requested in the previous review.
2. **Live native HTTP metrics.** With small limits and a deliberately blocked downstream, exercise both HTTP/1.1 and HTTP/2 through `WaitingEncryptedHttpConnection`; scrape a nonzero `tardigrade_tls_buffered_bytes_current` and one pause while blocked, then drain below low and observe the gauge/resume transition exactly once.
3. **Teardown and OpenSSL boundaries.** Prove per-connection gauge contributions return to zero on normal completion, timeout, peer abort, cancellation, and handshake failure, and add an OpenSSL pending/retry case showing only adapter-measurable state contributes under `backend="openssl"`.

The reload test should also assert the active config generation is unchanged, not only the exported limit and failure status, if that generation is exposed by the existing status surface.

CI for this head is still running. Even after it turns green, the current suite does not execute the production scheduler and live-metrics scenarios above, so this is still not ready to merge against the explicit #355 pickup contract.

src/edge_gateway.zig
```diff
@@ -2579,6 +2730,296 @@ test "connection timeout helper resets read and write timeouts" {
...
+    try std.testing.expect(registry.contains(pending.fd));
+
+    var expired_out: [1]http.downstream_connection.ManagedConnection = undefined;
+    try std.testing.expectEqual(@as(usize, 0), registry.reapExpired(1_001, expired_out[0..]));
+
+    var ready = switch (checkoutActiveForWorkerAt(&registry, pending.fd, 600)) {
+        .ready => |conn| conn,
+        .missing, .expired => return error.TestExpectedReadyActiveConnection,
+    };
+    _ = try ready.transport.native.drive();
```
Jaruso: **Blocking coverage gap: this proves only the first pending poll, not successful gateway-scheduled completion.**

After checkout, this calls `native.drive()` directly once, asserts the operation remains pending, and then `ready.deinit()` cancels it. The test never re-enters `rearmActiveConnection()`, never resets the registry reservation through `ActiveRegistry.rearm()`, and never dispatches the second/third `retry_at_ms` items that should complete the mock operation.

Please continue this same `ManagedConnection` through the production rearm/dispatch sequence until `pending_polls = 2` completes on poll 3. Assert the original `deadline_ms` remains `1_000`, `authPending()` becomes false, `poll_count == 3`, `op_release_count == 1`, and `cancel_count == 0`. That is the missing proof that the timer scheduler can make an async handshake finish without any fd event, rather than merely poll it once.

Jaruso (commented):
## Fifth review — scheduler race coverage added; two acceptance families remain

This commit materially improves the production-path coverage. The new gateway test now exercises a real native TLS connection with a pending credential operation and proves:

- no work is submitted before `retry_at_ms`;
- due work is tagged as `active_drive_poll` and remains reserved in `ActiveRegistry`;
- the reaper cannot destroy queued in-flight work;
- the worker-side checkout enforces the original absolute deadline;
- timeout cancels and releases the pending operation exactly once;
- a later socket reusing the same numeric fd is not touched by the stale work item.

I do not see a defect in the scheduler helper extraction or in the timeout/fd-reuse branch.

Two blocking acceptance areas remain:

1. **Successful repeated gateway polling is still not demonstrated.** The non-expired branch dequeues once, calls `native.drive()` directly, confirms the operation is still pending after poll 1, and immediately deinitializes the connection. It never exercises `rearmActiveConnection()` / `ActiveRegistry.rearm()`, a second and third timer dispatch, or eventual authentication completion through the gateway scheduler. A bug that leaves `active_work_queued` stuck after the first checkout, fails to schedule the next retry, or extends/replaces the deadline would still pass. Drive this same connection through repeated production rearm/dispatch cycles until the configured third poll completes, then assert `authPending() == false`, `poll_count == 3`, the original deadline is unchanged, release is exactly once, and cancellation is zero.

2. **Live HTTP/OpenSSL metrics acceptance remains absent.** No new coverage exercises a blocked native HTTP/1.1 writer, long-lived HTTP/2 writer, or OpenSSL pending/retry state through the production adapters. The PR still needs observable nonzero current gauges while blocked, one pause/resume transition as queues cross high/low, and exact return-to-zero on success, timeout, peer abort, cancellation, handshake failure, and graceful close. The OpenSSL case must assert only adapter-measurable state and no native-limit claim.

The earlier rejected-reload test also preserves the exported active limit, but still does not explicitly compare the active configuration generation if that value is available through the status surface.

CI run 819 is currently in progress. Once the successful repeated scheduler path and the live metrics/teardown boundaries are demonstrated, I would consider the implementation ready to merge.

Jaruso (commented):
## Sixth review — code and acceptance coverage are now sufficient

The latest commit closes the two remaining review gaps.

### Gateway async-auth scheduler

The gateway test now drives the same pending native TLS connection through all three production retry cycles:

- checkout of the tagged `active_drive_poll` work item;
- one bounded credential poll;
- rearm into `ActiveRegistry` with no raw socket interest;
- no submission before the next absolute `retry_at_ms`;
- timer redispatch at the due time;
- eventual completion on poll three without any socket event.

It also proves the original handshake deadline remains `1000` throughout, successful completion releases the operation once without cancellation, and the separate expired path still cancels/releases exactly once and cannot act on a reused fd.

### Runtime TLS metrics

The new HTTP/1.1 and HTTP/2 tests exercise the real `WaitingEncryptedHttpConnection` observation hooks, including the HTTP/2 frame writer, and verify live current-byte plus pause/resume Prometheus updates. Those adapter tests compose with the existing TLS-core slow-reader/slow-writer/high-low tests and the metrics-layer per-connection replacement, multi-connection aggregation, teardown release, and underflow tests. The issue explicitly permits fake or controlled carriers, so this is sufficient deterministic coverage without adding sleep-dependent socket tests.

The OpenSSL test uses a real TLS pair rather than a fabricated snapshot. It verifies measurable pending plaintext, forced write backpressure, pause/stalled-drive counters under `backend="openssl"`, isolation from the pure-Zig backend, and gauge release during connection teardown. Native configured-limit claims remain separate from OpenSSL accounting.

### Code review

I rechecked the new helper extraction and metric ownership paths:

- `rearmActiveConnectionInRegistryAt()` preserves the existing production behavior and gives the test the same registry transition used at runtime;
- native metric state stays owned by `ManagedConnection.lifecycle` across checkout/rearm and is released once by `ManagedConnection.deinit()`;
- OpenSSL metric state moves with `TlsConnection` into the parked slot and every close/reap/shutdown path reaches `TlsConnection.deinit()` once;
- observations occur after relevant read/write/flush/drive operations, and aggregate subtraction remains checked rather than saturating;
- the updated base adds unrelated TLS session work and does not conflict with these changes.

I found no remaining correctness defect or unmet #355 acceptance requirement. The PR is ready to merge once the currently running CI workflow completes successfully.

PR #475, opened 2026-07-20
tls: define transport-neutral resumable session and ticket model

Jaruso:
## Summary

- Add `src/tls/session.zig`: the owned data model for TLS 1.3 session resumption (#326-A), shared by record-mode TLS and QUIC/H3.
  - `ResumableSessionCommon`: canonical `CipherSuite`-derived resumption PSK (32 or 48 bytes, held in `crypto.secrets.FixedSecret`), optional canonical ASCII-lowercase SNI, optional ALPN, a stable SHA-256 leaf-certificate auth binding, issue time (signed Unix milliseconds)/lifetime, resume-only vs early-data-capable policy, and optional allocator-backed opaque transport/application compatibility snapshots (no QUIC or HTTP/3 types imported).
  - `ClientTicketState` and `ServerRecoverableState`: distinct owned types, the former holding the opaque ticket (as a `crypto.secrets.BoundedSecret` bearer credential), `ticket_age_add`, ticket nonce, and receipt timing.
  - A caller-tightenable `Limits` contract (field count, ticket length, transport/application compatibility blob lengths, total serialized size) validated against hard module/protocol caps, threaded through construction, encode, and decode.
  - A bounded, versioned, deterministic internal state codec (`encodeClient`/`encodeServer`/`decode` + `clientEncodedLenWithLimits`/`serverEncodedLenWithLimits`) with a real header (magic/version/kind/exact section length), a TLV body with critical/optional field ranges, global duplicate-field rejection, unknown-optional-field skipping, and unsupported-major-version rejection. This is not the RFC 8446 `NewSessionTicket` wire format.
  - `evaluateCompatibility`: typed compatibility decisions distinguishing expiry, not-yet-valid, cipher/SNI/ALPN/auth-binding/transport/application mismatches (with symmetric optional-presence semantics), with resumption eligibility kept separate from early-data eligibility.
  - Every owning type (`CompatSnapshot`, `ResumableSessionCommon`, `ClientTicketState`, `ServerRecoverableState`) uses pointer-based `init`/`deinit`/`cloneInto`/`moveFrom` with build-then-move construction, so self-cloning, reinitializing from a value's own borrowed storage, and moving into a live destination are all safe, and a failed `init` never consumes the caller's inputs.
- Add `CipherSuite.transcriptHash()` to `src/tls/algorithms.zig`, returning the existing provider-neutral `crypto.provider.Hash` (via `.digestLength()`) so the PSK length is always derived from the canonical cipher suite instead of a second hash vocabulary or independent persistence.
- Register the new module in `src/tls/root.zig`.

## Non-goals (see follow-up issues)

- `NewSessionTicket` message generation/parsing or resumption-master-secret derivation (#361).
- `pre_shared_key`/binder handling, ticket-age validation, full-handshake fallback (#362).
- Ticket encryption/authenticated envelopes (#363).
- Session cache storage/eviction (#364).
- Transport-level resumed-handshake integration (#365).

## Test plan

- [x] `zig build test-tls --summary all` — all TLS-core unit tests pass (411 in this module), including: SHA-256/SHA-384 PSK-length construction and decode boundaries (31/32/33, 47/48/49 bytes); deterministic client/server round-trips and order-independent decode; checked-in literal byte fixtures (fully populated, including both compatibility snapshots and a non-default early-data policy) with a full duplicate/missing-field matrix across every known field; millisecond expiry/overflow boundaries; every compatibility mismatch reason with symmetric optional-presence semantics; encode/decode `Limits` symmetry (ticket, transport/application compatibility, field count, total size); a malformed-input matrix (both fixtures truncated at every byte boundary, zero/under/over TLV length mutations, invalid cipher-suite enum values, inconsistent PSK/suite combinations, unknown-optional decode-then-canonical-re-encode); ownership/lifecycle tests (self-clone, live-destination clone/move, reinitializing from a value's own borrowed storage, a failed `ClientTicketState.init` leaving `common` and `self` untouched); and zeroization verification against a real inspectable backing buffer (ticket/compat-blob deinit, PSK reinitialization, a decode allocation-failure sweep, and failed clone/replace paths).
- [x] `zig build test --summary all` — full suite: 1755/1756 passed (1 skipped; pre-existing environmental UDP-socket skip unrelated to this change).
- [x] `zig fmt --check build.zig src/ tests/` — clean.

Closes #360

### DIFF
--- a/src/tls/algorithms.zig
+++ b/src/tls/algorithms.zig
@@ -18,6 +18,31 @@ pub const CipherSuite = enum(u16) {
     tls_chacha20_poly1305_sha256 = 0x1303,
 };
 
+/// Transcript/HKDF hash identifiers a `CipherSuite` can bind to. Callers must
+/// derive this from the cipher suite (`transcriptHash`) rather than
+/// persisting or selecting it independently, so the two can never disagree.
+pub const TranscriptHash = enum {
+    sha256,
+    sha384,
+
+    pub fn digestLen(self: TranscriptHash) usize {
+        return switch (self) {
+            .sha256 => std.crypto.hash.sha2.Sha256.digest_length,
+            .sha384 => std.crypto.hash.sha2.Sha384.digest_length,
+        };
+    }
+};
+
+/// The canonical transcript/HKDF hash for a cipher suite. This is the single
+/// source of truth other modules (key schedule, resumable-session model)
+/// must use instead of maintaining their own suite-to-hash switch.
+pub fn transcriptHash(suite: CipherSuite) TranscriptHash {
+    return switch (suite) {
+        .tls_aes_128_gcm_sha256, .tls_chacha20_poly1305_sha256 => .sha256,
+        .tls_aes_256_gcm_sha384 => .sha384,
+    };
+}
+
 pub const NamedGroup = enum(u16) {
     x25519 = 0x001d,
     secp256r1 = 0x0017,
@@ -64,4 +89,13 @@ test "registry exposes current TLS identifiers" {
     try std.testing.expectEqual(@as(u16, 0x001d), @intFromEnum(NamedGroup.x25519));
     try std.testing.expectEqual(@as(u16, 0x0807), @intFromEnum(SignatureScheme.ed25519));
     try std.testing.expect(alpn.h3.eql(.{ .bytes = "h3" }));
+}
+
+test "transcriptHash derives the correct hash and digest length per suite" {
+    try std.testing.expectEqual(TranscriptHash.sha256, transcriptHash(.tls_aes_128_gcm_sha256));
+    try std.testing.expectEqual(TranscriptHash.sha256, transcriptHash(.tls_chacha20_poly1305_sha256));
+    try std.testing.expectEqual(TranscriptHash.sha384, transcriptHash(.tls_aes_256_gcm_sha384));
+
+    try std.testing.expectEqual(@as(usize, 32), TranscriptHash.sha256.digestLen());
+    try std.testing.expectEqual(@as(usize, 48), TranscriptHash.sha384.digestLen());
 }
--- a/src/tls/algorithms.zig
+++ b/src/tls/algorithms.zig
@@ -5,6 +5,8 @@
 //! particular transport or provider.
 
 const std = @import("std");
+const crypto = @import("crypto");
+const crypto_provider = crypto.provider;
 
 pub const ProtocolVersion = enum(u16) {
     tls13 = 0x0304,
@@ -18,25 +20,12 @@ pub const CipherSuite = enum(u16) {
     tls_chacha20_poly1305_sha256 = 0x1303,
 };
 
-/// Transcript/HKDF hash identifiers a `CipherSuite` can bind to. Callers must
-/// derive this from the cipher suite (`transcriptHash`) rather than
-/// persisting or selecting it independently, so the two can never disagree.
-pub const TranscriptHash = enum {
-    sha256,
-    sha384,
-
-    pub fn digestLen(self: TranscriptHash) usize {
-        return switch (self) {
-            .sha256 => std.crypto.hash.sha2.Sha256.digest_length,
-            .sha384 => std.crypto.hash.sha2.Sha384.digest_length,
-        };
-    }
-};
-
-/// The canonical transcript/HKDF hash for a cipher suite. This is the single
-/// source of truth other modules (key schedule, resumable-session model)
-/// must use instead of maintaining their own suite-to-hash switch.
-pub fn transcriptHash(suite: CipherSuite) TranscriptHash {
+/// The canonical transcript/HKDF hash for a cipher suite, expressed as the
+/// existing provider-neutral `crypto.provider.Hash` identity (not a second
+/// TLS-local hash vocabulary). Callers must derive this from the cipher
+/// suite rather than persisting or selecting a hash independently, so the
+/// two can never disagree.
+pub fn transcriptHash(suite: CipherSuite) crypto_provider.Hash {
     return switch (suite) {
         .tls_aes_128_gcm_sha256, .tls_chacha20_poly1305_sha256 => .sha256,
         .tls_aes_256_gcm_sha384 => .sha384,
@@ -91,11 +80,11 @@ test "registry exposes current TLS identifiers" {
     try std.testing.expect(alpn.h3.eql(.{ .bytes = "h3" }));
 }
 
-test "transcriptHash derives the correct hash and digest length per suite" {
-    try std.testing.expectEqual(TranscriptHash.sha256, transcriptHash(.tls_aes_128_gcm_sha256));
-    try std.testing.expectEqual(TranscriptHash.sha256, transcriptHash(.tls_chacha20_poly1305_sha256));
-    try std.testing.expectEqual(TranscriptHash.sha384, transcriptHash(.tls_aes_256_gcm_sha384));
+test "transcriptHash derives the correct provider Hash and digest length per suite" {
+    try std.testing.expectEqual(crypto_provider.Hash.sha256, transcriptHash(.tls_aes_128_gcm_sha256));
+    try std.testing.expectEqual(crypto_provider.Hash.sha256, transcriptHash(.tls_chacha20_poly1305_sha256));
+    try std.testing.expectEqual(crypto_provider.Hash.sha384, transcriptHash(.tls_aes_256_gcm_sha384));
 
-    try std.testing.expectEqual(@as(usize, 32), TranscriptHash.sha256.digestLen());
-    try std.testing.expectEqual(@as(usize, 48), TranscriptHash.sha384.digestLen());
+    try std.testing.expectEqual(@as(usize, 32), transcriptHash(.tls_aes_128_gcm_sha256).digestLength());
+    try std.testing.expectEqual(@as(usize, 48), transcriptHash(.tls_aes_256_gcm_sha384).digestLength());
 }
--- a/src/tls/session.zig
+++ b/src/tls/session.zig
@@ -1,0 +1,1316 @@
+//! Transport-neutral resumable TLS 1.3 session and ticket model (#360).
+//!
+//! This module defines the owned value types for TLS 1.3 session resumption
+//! shared by record-mode TLS and QUIC/H3: common resumable-session metadata,
+//! client-held ticket state, server-recoverable state, a bounded internal
+//! serialization codec for that state, and typed compatibility decisions.
+//!
+//! It intentionally does not implement the RFC 8446 `NewSessionTicket`
+//! handshake-message codec or resumed-handshake behavior (see
+//! `messages.zig` / `tls13_backend.zig` for the existing framing check, and
+//! issue #361), `pre_shared_key` / binder handling (#362), ticket encryption
+//! (#363), session-cache storage (#364), or transport integration (#365). It
+//! also does not import QUIC transport-parameter or HTTP/3 SETTINGS types:
+//! transport/application compatibility is represented as bounded opaque
+//! blobs so those adapters can be built on top without this module knowing
+//! about them.
+//!
+//! The internal serialization defined here (`encodeClient`/`decode`/etc.) is
+//! a caller-facing persistence format, not the wire format. Encoded state
+//! contains bearer secrets (the resumption PSK and the raw ticket) and must
+//! be treated as sensitive plaintext: it must be encrypted and authenticated
+//! (#363) before leaving a trusted process or persistence boundary.
+
+const std = @import("std");
+const crypto = @import("crypto");
+const algorithms = @import("algorithms.zig");
+const dns_name = @import("dns_name.zig");
+const sni_provider = @import("sni_provider.zig");
+
+const secrets = crypto.secrets;
+
+pub const max_sni_len = dns_name.max_name_len;
+pub const max_alpn_len = 255;
+pub const max_ticket_len = 65535;
+pub const max_ticket_nonce_len = 255;
+pub const max_compat_blob_len = 512;
+pub const auth_binding_len = 32;
+pub const max_psk_len = 48;
+
+/// Overall bound on one encoded internal-state record, chosen generously
+/// above the sum of every field's individual maximum so a corrupt or hostile
+/// buffer is rejected before any attacker-controlled allocation is made.
+pub const max_encoded_state_len: usize = 128 * 1024;
+
+pub const ResumptionPsk = secrets.FixedSecret(max_psk_len);
+
+/// Resume-only vs early-data-capable policy for a resumable session.
+pub const EarlyDataPolicy = union(enum) {
+    resume_only,
+    /// Maximum early-data byte count the server is willing to accept; must
+    /// be non-zero (a zero-byte allowance is expressed as `.resume_only`).
+    early_data_capable: u32,
+
+    pub fn maxEarlyData(self: EarlyDataPolicy) u32 {
+        return switch (self) {
+            .resume_only => 0,
+            .early_data_capable => |max| max,
+        };
+    }
+};
+
+/// A bounded, opaque compatibility snapshot for a transport or application
+/// layer, identified by a caller-defined format id/version. Equality is
+/// exact: any byte difference is a mismatch.
+pub const CompatSnapshot = struct {
+    format_id: u16,
+    format_version: u16,
+    len: u16 = 0,
+    bytes: [max_compat_blob_len]u8 = undefined,
+
+    pub const Error = error{CompatSnapshotTooLarge};
+
+    pub fn init(format_id: u16, snapshot_version: u16, data: []const u8) Error!CompatSnapshot {
+        if (data.len > max_compat_blob_len) return error.CompatSnapshotTooLarge;
+        var self = CompatSnapshot{ .format_id = format_id, .format_version = snapshot_version };
+        @memcpy(self.bytes[0..data.len], data);
+        self.len = @intCast(data.len);
+        return self;
+    }
+
+    pub fn slice(self: *const CompatSnapshot) []const u8 {
+        return self.bytes[0..self.len];
+    }
+
+    pub fn eql(self: *const CompatSnapshot, other: *const CompatSnapshot) bool {
+        return self.format_id == other.format_id and
+            self.format_version == other.format_version and
+            std.mem.eql(u8, self.slice(), other.slice());
+    }
+};
+
+/// Canonical ASCII-lowercase SNI, validated with the shared TLS DNS-name
+/// rules at construction time.
+pub const SniName = struct {
+    len: u8 = 0,
+    bytes: [max_sni_len]u8 = undefined,
+
+    pub fn init(raw: []const u8) dns_name.Error!SniName {
+        try dns_name.validateHostName(raw);
+        var self = SniName{};
+        for (raw, 0..) |ch, i| self.bytes[i] = sni_provider.asciiLower(ch);
+        self.len = @intCast(raw.len);
+        return self;
+    }
+
+    pub fn slice(self: *const SniName) []const u8 {
+        return self.bytes[0..self.len];
+    }
+
+    pub fn eqlIgnoreCase(self: *const SniName, candidate: []const u8) bool {
+        return sni_provider.asciiEqlIgnoreCase(self.slice(), candidate);
+    }
+};
+
+/// Owned, bounded ALPN protocol name (RFC 7301 protocol names are
+/// length-prefixed by a single byte, so 255 bytes is the wire maximum).
+pub const AlpnProtocol = struct {
+    len: u8 = 0,
+    bytes: [max_alpn_len]u8 = undefined,
+
+    pub const Error = error{AlpnProtocolTooLarge};
+
+    pub fn init(raw: []const u8) Error!AlpnProtocol {
+        if (raw.len == 0 or raw.len > max_alpn_len) return error.AlpnProtocolTooLarge;
+        var self = AlpnProtocol{};
+        @memcpy(self.bytes[0..raw.len], raw);
+        self.len = @intCast(raw.len);
+        return self;
+    }
+
+    pub fn slice(self: *const AlpnProtocol) []const u8 {
+        return self.bytes[0..self.len];
+    }
+
+    pub fn eql(self: *const AlpnProtocol, other: []const u8) bool {
+        return std.mem.eql(u8, self.slice(), other);
+    }
+};
+
+/// Stable authentication binding for the peer identity a session was
+/// negotiated with. For format version 1 this is a SHA-256 digest of the
+/// leaf certificate DER (not `sni_provider.Snapshot.generation`, which is
+/// process-local reload-ordering state, not a portable identity).
+pub const AuthBinding = struct {
+    bytes: [auth_binding_len]u8,
+
+    pub fn fromLeafCertificateDer(der: []const u8) AuthBinding {
+        var digest: [auth_binding_len]u8 = undefined;
+        std.crypto.hash.sha2.Sha256.hash(der, &digest, .{});
+        return .{ .bytes = digest };
+    }
+
+    pub fn eql(self: AuthBinding, other: AuthBinding) bool {
+        return std.mem.eql(u8, &self.bytes, &other.bytes);
+    }
+};
+
+/// Common resumable-session metadata shared by client-ticket state and
+/// server-recoverable state.
+///
+/// Owns a resumption PSK (a secret). Fields are conceptually private:
+/// callers should use `init`/`deinit`/`clone` rather than constructing or
+/// copying this value directly, since a bitwise copy of the PSK container is
+/// only safe because `FixedSecret` is inline storage — treat that as an
+/// implementation detail, not a supported API.
+pub const ResumableSessionCommon = struct {
+    resumption_psk: ResumptionPsk = .{},
+    cipher_suite: algorithms.CipherSuite = .tls_aes_128_gcm_sha256,
+    sni: SniName = .{},
+    alpn: AlpnProtocol = .{},
+    auth_binding: AuthBinding = .{ .bytes = [_]u8{0} ** auth_binding_len },
+    issued_at: u64 = 0,
+    lifetime_seconds: u32 = 0,
+    early_data: EarlyDataPolicy = .resume_only,
+    transport_compat: ?CompatSnapshot = null,
+    application_compat: ?CompatSnapshot = null,
+
+    pub const InitParams = struct {
+        cipher_suite: algorithms.CipherSuite,
+        resumption_psk: []const u8,
+        sni: []const u8,
+        alpn: []const u8,
+        auth_binding: AuthBinding,
+        issued_at: u64,
+        lifetime_seconds: u32,
+        early_data: EarlyDataPolicy = .resume_only,
+        transport_compat: ?CompatSnapshot = null,
+        application_compat: ?CompatSnapshot = null,
+    };
+
+    pub const InitError = error{
+        InvalidDnsName,
+        AlpnProtocolTooLarge,
+        InvalidPskLength,
+        InvalidEarlyDataPolicy,
+    };
+
+    pub fn init(params: InitParams) InitError!ResumableSessionCommon {
+        const expected_psk_len = algorithms.transcriptHash(params.cipher_suite).digestLen();
+        if (params.resumption_psk.len != expected_psk_len) return error.InvalidPskLength;
+        switch (params.early_data) {
+            .resume_only => {},
+            .early_data_capable => |max| if (max == 0) return error.InvalidEarlyDataPolicy,
+        }
+
+        var self = ResumableSessionCommon{
+            .cipher_suite = params.cipher_suite,
+            .sni = try SniName.init(params.sni),
+            .alpn = try AlpnProtocol.init(params.alpn),
+            .auth_binding = params.auth_binding,
+            .issued_at = params.issued_at,
+            .lifetime_seconds = params.lifetime_seconds,
+            .early_data = params.early_data,
+            .transport_compat = params.transport_compat,
+            .application_compat = params.application_compat,
+        };
+        errdefer self.deinit();
+        self.resumption_psk.replace(params.resumption_psk) catch unreachable;
+        return self;
+    }
+
+    pub fn deinit(self: *ResumableSessionCommon) void {
+        self.resumption_psk.deinit();
+    }
+
+    /// Explicit deep copy. Does not alias the source's secret storage.
+    pub fn clone(self: *const ResumableSessionCommon) ResumableSessionCommon {
+        var out = self.*;
+        out.resumption_psk = self.resumption_psk.copy();
+        return out;
+    }
+
+    pub fn isExpired(self: *const ResumableSessionCommon, now: u64) bool {
+        const expiry = std.math.add(u64, self.issued_at, self.lifetime_seconds) catch std.math.maxInt(u64);
+        return now >= expiry;
+    }
+
+    pub fn isNotYetValid(self: *const ResumableSessionCommon, now: u64) bool {
+        return now < self.issued_at;
+    }
+
+    pub fn format(
+        _: ResumableSessionCommon,
+        comptime _: []const u8,
+        _: std.fmt.FormatOptions,
+        _: anytype,
+    ) !void {
+        @compileError("resumable session state must not be formatted or logged");
+    }
+};
+
+/// Client-owned ticket state: the common session metadata plus the opaque
+/// ticket identity and timing needed to offer resumption. The ticket bytes
+/// are a bearer credential transmitted on the wire, but are held under the
+/// same secret ownership/redaction rules as the PSK.
+pub const ClientTicketState = struct {
+    common: ResumableSessionCommon = .{},
+    ticket: secrets.BoundedSecret = .{},
+    ticket_age_add: u32 = 0,
+    ticket_nonce: secrets.FixedSecret(max_ticket_nonce_len) = .{},
+    /// Unix seconds when this ticket was received from the server.
+    received_at: u64 = 0,
+
+    pub const InitParams = struct {
+        /// Ownership of `common` is transferred in; on success it is owned
+        /// by the returned `ClientTicketState`, on failure it is deinited.
+        common: ResumableSessionCommon,
+        ticket: []const u8,
+        ticket_age_add: u32,
+        ticket_nonce: []const u8,
+        received_at: u64,
+    };
+
+    pub const InitError = error{ OutOfMemory, TicketTooLarge, NonceTooLarge };
+
+    pub fn init(allocator: std.mem.Allocator, params: InitParams) InitError!ClientTicketState {
+        var common = params.common;
+        errdefer common.deinit();
+
+        if (params.ticket.len == 0 or params.ticket.len > max_ticket_len) return error.TicketTooLarge;
+        if (params.ticket_nonce.len > max_ticket_nonce_len) return error.NonceTooLarge;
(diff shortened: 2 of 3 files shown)


src/tls/algorithms.zig
```diff
@@ -18,6 +18,31 @@ pub const CipherSuite = enum(u16) {
     tls_chacha20_poly1305_sha256 = 0x1303,
 };
 
+/// Transcript/HKDF hash identifiers a `CipherSuite` can bind to. Callers must
+/// derive this from the cipher suite (`transcriptHash`) rather than
+/// persisting or selecting it independently, so the two can never disagree.
+pub const TranscriptHash = enum {
```
Jaruso: **Blocking architecture issue:** this creates a second SHA-256/SHA-384 vocabulary even though `crypto.provider.Hash` is already the provider-neutral canonical type. The authoritative plan explicitly called out that existing owner and asked for a suite-to-`provider.Hash` mapping. Keeping `TranscriptHash` here means #361/#362 and the provider seam will need another mapping that can drift.

Please return the existing type instead, for example:

```zig
const crypto_provider = @import("crypto").provider;

pub fn transcriptHash(suite: CipherSuite) crypto_provider.Hash {
    return switch (suite) {
        .tls_aes_128_gcm_sha256,
        .tls_chacha20_poly1305_sha256,
        => .sha256,
        .tls_aes_256_gcm_sha384 => .sha384,
    };
}
```

Callers can then use `transcriptHash(suite).digestLength()` and there remains one provider-neutral hash identity.

src/tls/session.zig
```diff
@@ -0,0 +1,1316 @@
...
+pub const max_ticket_len = 65535;
+pub const max_ticket_nonce_len = 255;
+pub const max_compat_blob_len = 512;
+pub const auth_binding_len = 32;
+pub const max_psk_len = 48;
+
+/// Overall bound on one encoded internal-state record, chosen generously
+/// above the sum of every field's individual maximum so a corrupt or hostile
+/// buffer is rejected before any attacker-controlled allocation is made.
+pub const max_encoded_state_len: usize = 128 * 1024;
```
Jaruso: **Blocking bounds issue:** `max_ticket_len = 65535` plus a single 128 KiB state limit does not implement the required caller-tightenable `Limits` contract. The maintainer plan requires a small default state (roughly 4 KiB), a field-count limit (default 32), separate ticket/transport/application limits, and validation that caller limits never exceed hard module/protocol caps. The 65,535-byte value should be an **absolute TLS wire cap**, not the default amount every internal decoder/cache accepts.

Please add something along these lines and thread it through construction, encoded-length calculation, encode, and decode (with convenience wrappers using `Limits.default`):

```zig
pub const Limits = struct {
    max_fields: usize = 32,
    max_ticket_len: usize = 4096,
    max_transport_compat_len: usize = 1024,
    max_application_compat_len: usize = 1024,
    max_serialized_len: usize = 4096,

    pub fn validate(self: Limits) error{InvalidLimits}!void {
        if (self.max_fields > hard_max_fields or
            self.max_ticket_len > absolute_ticket_wire_max or
            self.max_transport_compat_len > hard_max_compat_len or
            self.max_application_compat_len > hard_max_compat_len or
            self.max_serialized_len > hard_max_serialized_len)
            return error.InvalidLimits;
    }
};
```

Add exact boundary and one-over tests for every limit, including 32/33 TLVs.

src/tls/session.zig
```diff
@@ -0,0 +1,1316 @@
...
+            .resume_only => 0,
+            .early_data_capable => |max| max,
+        };
+    }
+};
+
+/// A bounded, opaque compatibility snapshot for a transport or application
+/// layer, identified by a caller-defined format id/version. Equality is
+/// exact: any byte difference is a mismatch.
+pub const CompatSnapshot = struct {
```
Jaruso, line 117: **Blocking ownership/logging issue:** compatibility blobs were required to be owned bounded values with explicit lifecycle, but this embeds two maximum arrays in every session and remains ordinarily formattable. `std.debug.print("{any}", .{snapshot})` will expose the raw compatibility bytes (and potentially the undefined unused tail), contrary to the ticket's explicit no-logging rule.

Please make this an allocator-backed owned type (a wrapped `BoundedSecret` is reasonable even if the blob is not cryptographic key material), with pointer-based `init`/`deinit`, explicit deep clone/move, exact borrowed `slice`, and a compile-error `format` implementation. This also lets `Limits` control transport and application blobs independently and gives the allocation-failure suite the public allocations the issue requires it to exercise.

src/tls/session.zig
```diff
@@ -0,0 +1,1316 @@
...
+///
+/// Owns a resumption PSK (a secret). Fields are conceptually private:
+/// callers should use `init`/`deinit`/`clone` rather than constructing or
+/// copying this value directly, since a bitwise copy of the PSK container is
+/// only safe because `FixedSecret` is inline storage — treat that as an
+/// implementation detail, not a supported API.
+pub const ResumableSessionCommon = struct {
+    resumption_psk: ResumptionPsk = .{},
+    cipher_suite: algorithms.CipherSuite = .tls_aes_128_gcm_sha256,
+    sni: SniName = .{},
```
Jaruso: **Blocking model mismatch:** SNI and ALPN are explicitly optional in the authoritative plan. `null` means the original handshake had no value; an empty present value is invalid. Requiring both here and treating their codec fields as mandatory means this model cannot represent a valid no-SNI handshake or record-mode TLS where ALPN was absent/fallback was applied outside TLS.

The shape should preserve absence, for example:

```zig
server_name: ?SniName = null,
application_protocol: ?AlpnProtocol = null,

pub const InitParams = struct {
    application_protocol: ?[]const u8 = null,
    // ...
};
```

Construct only when present, omit the corresponding TLV for `null`, reject a present empty slice, make candidate values optional too, and compare `null` equal only to `null`. Add round-trip and mismatch tests for absent/absent, absent/present, present/absent, and present/present.

src/tls/session.zig
```diff
@@ -0,0 +1,1316 @@
...
+    alpn: AlpnProtocol = .{},
+    auth_binding: AuthBinding = .{ .bytes = [_]u8{0} ** auth_binding_len },
+    issued_at: u64 = 0,
```
Jaruso: **Blocking protocol-time issue:** the authoritative model uses signed Unix **milliseconds**, while this stores unsigned seconds and accepts any `u32` lifetime, including zero and values above TLS 1.3's seven-day maximum. #362 needs millisecond ticket age for `obfuscated_ticket_age`; `ageSeconds()` permanently loses that information.

Please use `issued_at_unix_ms: i64`, `received_at_unix_ms: i64`, and caller-supplied `now_unix_ms: i64`, reject `lifetime_seconds == 0` or `> 604800`, and perform the comparison in a wider checked type:

```zig
const age_ms: i128 = @as(i128, now_unix_ms) - @as(i128, issued_at_unix_ms);
if (age_ms < 0) return .not_yet_valid;
const lifetime_ms: i128 = @as(i128, lifetime_seconds) * 1000;
if (age_ms >= lifetime_ms) return .expired;
```

Add `InvalidLifetime`, encode/decode signed milliseconds canonically, and test issue time, final valid millisecond, exact expiry, future timestamps, and integer extremes.

src/tls/session.zig
```diff
@@ -0,0 +1,1316 @@
...
+        common: ResumableSessionCommon,
+        ticket: []const u8,
+        ticket_age_add: u32,
+        ticket_nonce: []const u8,
+        received_at: u64,
+    };
+
+    pub const InitError = error{ OutOfMemory, TicketTooLarge, NonceTooLarge };
+
+    pub fn init(allocator: std.mem.Allocator, params: InitParams) InitError!ClientTicketState {
```
Jaruso: **Blocking secret ownership bug:** this API does not actually transfer ownership. `params.common` is copied into a local and then copied again into `self`; the caller's source remains live and the successful path leaves additional PSK copies in stack/value storage. More critically, `ClientTicketState` is trivially shallow-copyable while containing `BoundedSecret`; `var b = a;` aliases the same allocation, so deinitializing both double-frees and deinitializing either invalidates the other.

The issue explicitly requires pointer-based initialization and an explicit move that invalidates the source. A safe pattern is:

```zig
pub fn init(
    self: *ClientTicketState,
    allocator: Allocator,
    common: *ResumableSessionCommon,
    params: TicketParams,
) InitError!void {
    self.* = .{};
    errdefer self.deinit();

    try self.ticket.init(allocator, params.ticket.len, params.ticket);
    try self.ticket_nonce.replace(params.ticket_nonce);
    self.ticket_age_add = params.ticket_age_add;
    self.received_at_unix_ms = params.received_at_unix_ms;

    // Last, non-failing ownership transfer. Source becomes safely empty.
    self.common.moveFrom(common);
}
```

Provide `moveFrom`/`moveInto` for every owning aggregate, pointer-based `cloneInto`, and make both moved-from and partially initialized values safe to deinit. Tests must prove the source PSK/ticket is cleared after a move, allocation failure does not leave duplicate ownership, and moving then deinitializing both sides is safe.

src/tls/session.zig
```diff
@@ -0,0 +1,1316 @@
...
+    if (!common.alpn.eql(candidate.alpn)) return .{ .rejected = .alpn_mismatch };
+    if (!common.auth_binding.eql(candidate.auth_binding)) return .{ .rejected = .auth_binding_mismatch };
+    if (!compatSnapshotMatches(common.transport_compat, candidate.transport_compat))
+        return .{ .rejected = .transport_mismatch };
+    if (!compatSnapshotMatches(common.application_compat, candidate.application_compat))
+        return .{ .rejected = .application_mismatch };
+    return .eligible;
+}
+
+fn compatSnapshotMatches(session_snapshot: ?CompatSnapshot, candidate_snapshot: ?CompatSnapshot) bool {
```
Jaruso: **Blocking compatibility bug:** an absent stored snapshot is currently treated as a wildcard (`return true`). The contract requires exact optional equality: absent and present are distinct. As written, an old/unbound session can be accepted in a candidate transport/application context that supplies a compatibility binding, which defeats the purpose of adding the snapshot.

Use symmetric optional equality:

```zig
fn compatSnapshotMatches(stored: ?CompatSnapshot, candidate: ?CompatSnapshot) bool {
    return switch (stored) {
        null => candidate == null,
        else => |s| if (candidate) |c| s.eql(&c) else false,
    };
}
```

Add tests for both asymmetric cases (`null`/present and present/`null`) for both transport and application snapshots.

src/tls/session.zig
```diff
@@ -0,0 +1,1316 @@
...
+// -----------------------------------------------------------------------
+// Internal state codec (not the RFC 8446 NewSessionTicket wire format)
+// -----------------------------------------------------------------------
+
+pub const RecordType = enum(u8) {
+    client = 1,
+    server = 2,
+};
+
+pub const format_version: u8 = 1;
```
Jaruso, line 807: **Blocking codec-contract gap:** format v1 was specified with a fixed magic, major version, state kind, and a declared bounded field section. This is only `[version, kind] + TLVs`; it has no magic, no section length, no field-count enforcement, and no global field-ID set. Consequently repeated unknown optional IDs are accepted even though the contract says duplicate fields fail.

Please introduce a real header and validate it before walking fields, for example:

```text
magic[4] | major:u8 | kind:u8 | field_section_len:u32 | TLVs...
```

During the validation/index pass, enforce `limits.max_fields`, track every field ID (known or unknown) and reject any duplicate, require the declared section to consume the input exactly, then allocate/copy only after the whole record validates. Add checked-in literal client/server byte fixtures plus magic, section-under/overrun, 32/33 fields, duplicate unknown-optional, duplicate-every-known-field, and missing-every-required-field tests.

src/tls/session.zig
```diff
@@ -0,0 +1,1316 @@
...
+            break :blk early_bytes[0..5];
+        },
+    };
+    try writeTlv(out, pos, field_early_data, early_slice);
+
+    if (common.transport_compat) |*snap| try writeCompatSnapshot(out, pos, field_transport_compat, snap);
+    if (common.application_compat) |*snap| try writeCompatSnapshot(out, pos, field_application_compat, snap);
+}
+
+pub fn encodeClient(state: *const ClientTicketState, out: []u8) EncodeError![]const u8 {
```
Jaruso: **Blocking secret-handling issue:** `encodeClient` checks only two bytes before it starts writing. A short buffer can therefore receive the header, cipher, PSK, SNI, etc. and then return `BufferTooSmall`, leaving a partial plaintext bearer state in caller memory. The authoritative plan explicitly requires undersized output to fail before writing partial secret state where practical, and `clientEncodedLen` already makes that easy.

Preflight the complete output and configured total limit before the first write:

```zig
const needed = try clientEncodedLenWithLimits(state, limits);
if (needed > limits.max_serialized_len) return error.StateTooLarge;
if (out.len < needed) return error.BufferTooSmall;

// All subsequent writes are now guaranteed to fit.
```

Do the same for server encoding. Add a test for every output size below `needed`, prefill the buffer with a sentinel, and assert the call either leaves it unchanged or securely zeros every touched byte; specifically verify that no PSK or ticket subsequence remains after failure.

Jaruso (commented):
## Blocking review — not merge-ready yet

The PR is well scoped and CI run 812 is fully green, but the implementation diverges from the authoritative #360 issue body/maintainer plan in several security- and compatibility-relevant ways. The main blockers are the by-value ownership API, required SNI/ALPN, second-based/unbounded lifetime handling, missing configurable codec limits/header framing, partial secret writes on encode failure, and wildcard compatibility semantics.

The inline comments include concrete implementation shapes. Please also expand the tests while addressing them. The current suite does **not** yet provide the required checked-in literal byte fixtures, move-source invalidation, absent-vs-present SNI/ALPN cases, millisecond lifetime boundaries, duplicate/missing coverage for every known field, field-count/default-limit boundaries, undersized-encode output inspection, or backing-memory verification that failed construction/decode actually wiped secret capacity.

Because this review is being submitted by the PR author account, GitHub only permits a comment review; treat the inline findings as requested changes.

src/tls/session.zig
```diff
@@ -0,0 +1,1954 @@
...
+    std.mem.writeInt(u32, out[6..10], field_section_len, .big);
+}
+
+/// Encodes `state` into `out`. The complete required length is computed and
+/// validated against `limits.max_serialized_len` and `out.len` before any
+/// byte is written, so an undersized buffer or over-limit state fails
+/// without ever leaving partial plaintext secret state in `out`.
+pub fn encodeClient(state: *const ClientTicketState, limits: Limits, out: []u8) EncodeError![]const u8 {
+    try limits.validate();
+    const needed = clientEncodedLen(state);
```
Jaruso: **Blocking limits asymmetry:** the encoder validates only `limits.max_serialized_len`. It does not enforce `max_fields`, `max_ticket_len`, `max_transport_compat_len`, or `max_application_compat_len`, even though `decode` enforces all four.

That permits this invalid round trip:

```zig
var wide = Limits.default;
wide.max_ticket_len = 8192;
wide.max_serialized_len = 16384;
// construct a state containing a 5000-byte ticket under `wide`

var tight = wide;
tight.max_ticket_len = 4096;
const encoded = try encodeClient(&state, tight, out); // currently succeeds
_ = try decode(allocator, tight, encoded);            // FieldTooLarge
```

`max_fields` has the same problem: `encodeClient` succeeds with `limits.max_fields = 1`, producing a record its matching decoder rejects as `TooManyFields`. Compatibility blobs built under wider limits are likewise not checked here.

Please add one preflight that validates **all** state fields against the supplied limits and counts the exact emitted TLVs before the first write. For example, make `clientEncodedLenWithLimits`/`serverEncodedLenWithLimits` return `EncodeError!usize`, with `TooManyFields` and `FieldTooLarge` in `EncodeError`, then have encode use that result. Add wide-construction/tight-encoding tests for every individual limit and assert encode/decode symmetry under the same `Limits`.

src/tls/session.zig
```diff
@@ -0,0 +1,1954 @@
...
+    }
+
+    pub fn moveInto(self: *ClientTicketState, dest: *ClientTicketState) void {
+        dest.moveFrom(self);
+    }
+
+    /// Overflow-safe elapsed time since receipt, saturating to zero if
+    /// `now_unix_ms` predates `received_at_unix_ms` (e.g. clock skew).
+    pub fn ageMillis(self: *const ClientTicketState, now_unix_ms: i64) i64 {
+        return @max(0, now_unix_ms - self.received_at_unix_ms);
```
Jaruso: **Blocking overflow bug:** `@max` runs only after `now_unix_ms - received_at_unix_ms` has already been evaluated in `i64`. For example, `now = maxInt(i64)` and `received_at = minInt(i64)` traps in Debug/ReleaseSafe and wraps in ReleaseFast.

Compute the delta in `i128`, as `isExpired` already does, then clamp or return a wider unsigned value:

```zig
pub fn ageMillis(self: *const ClientTicketState, now_unix_ms: i64) u64 {
    const age: i128 = @as(i128, now_unix_ms) -
        @as(i128, self.received_at_unix_ms);
    if (age <= 0) return 0;
    return @intCast(@min(age, std.math.maxInt(u64)));
}
```

Alternatively return `i128` and let #362 perform the explicit RFC 8446 modulo-`2^32` obfuscated-age calculation. Add tests for both `i64` extremes and future receipt times.

src/tls/session.zig
```diff
@@ -0,0 +1,1954 @@
...
+        self: *CompatSnapshot,
+        allocator: std.mem.Allocator,
+        format_id: u16,
+        snapshot_version: u16,
+        data: []const u8,
+        limit: usize,
+    ) InitError!void {
+        self.* = .{};
+        if (data.len > limit) return error.CompatSnapshotTooLarge;
+        self.format_id = format_id;
```
Jaruso: **Blocking public-bound/lifecycle gap:** `CompatSnapshot.init` trusts the raw `limit` argument without checking it against `hard_max_compat_len`. A caller can pass `limit = 9000`, create a 9000-byte public snapshot, and—because encode currently does not revalidate per-field limits—reach `writeCompatSnapshot`'s `[4 + hard_max_compat_len]` scratch with an oversized copy.

This public constructor must enforce the absolute module cap itself, not rely on callers having obtained the limit from a validated `Limits` value. Prefer accepting `Limits` plus a transport/application selector, or at minimum reject `limit > hard_max_compat_len` before allocation.

There is also a replacement leak: `self.* = .{}` discards an existing `BoundedSecret` allocation without wiping or freeing it. The same pattern exists in `cloneInto`/`moveFrom` when the destination is already live, and self-move clears the only owner. Either provide a separate `replace` that deinitializes live storage and make `init` explicitly require a zero-valued destination, or make these operations safely handle live/self destinations. Add hard-cap, reinitialization, live-destination move/clone, and self-move tests.

src/tls/session.zig
```diff
@@ -0,0 +1,1954 @@
...
+    // Generated once from `sampleClient`/`encodeClient` with `Limits.default`
+    // and pinned here so an accidental field-order or framing change is
+    // caught even if the round-trip test above is (incorrectly) also
+    // changed to match.
+    var client = try sampleClient(testing.allocator);
+    defer client.deinit();
+    var buf: [Limits.default.max_serialized_len]u8 = undefined;
+    const encoded = try encodeClient(&client, Limits.default, &buf);
+
+    // Pin the header shape rather than the full body (which embeds a
```
Jaruso: **Blocking acceptance-criteria gap:** this is not a checked-in literal byte fixture. It invokes the encoder under test and checks only the header, so an accidental body field-ID/order/integer-encoding change will make both producer and assertion change together without failing. The stated concern about hardcoding a secret does not apply to deterministic test-only bytes—the sample already uses a fake repeated `0xab` PSK and a fake ticket.

Please add complete literal `const` byte arrays (or literal hex decoded at comptime) for **both** a client record and a server record, decode those fixed bytes, assert every field, and separately assert that current encoding exactly equals those fixtures.

The required negative matrix is also incomplete: the tests duplicate only `server_name`, not every known field, and do not remove every required field. Use a fully populated fixture, enumerate its TLVs, and generate duplicate/missing cases for each applicable field so changes to required/optional classification cannot escape coverage.

Jaruso (commented):
## Re-review — four blockers remain

I re-reviewed current head `733f7c18` against the updated #360 body, the authoritative maintainer plan, every finding from review `4734574259`, and the full revised `session.zig` implementation.

The revision correctly addresses most of the first-round findings: it now uses `crypto.provider.Hash`, models SNI/ALPN as optional, uses signed millisecond timestamps and the seven-day lifetime bound, owns compatibility blobs, provides explicit move/clone APIs, uses symmetric compatibility matching, adds a real magic/version/kind/section header, globally rejects duplicate field IDs, preflights output length, and introduces caller-tightenable limits.

However, the remaining issues below still violate the codec and lifecycle contract:

1. `encodeClient`/`encodeServer` enforce only `max_serialized_len`; they ignore the same limits that decode enforces (`max_fields`, ticket length, and compatibility lengths). The same state can therefore encode successfully under limits that reject it on decode, and oversized public snapshots can reach the fixed scratch buffer.
2. `ageMillis` performs the subtraction in `i64`, so valid extreme timestamps can overflow before `@max` runs.
3. `CompatSnapshot.init` accepts caller-provided limits above the module hard cap and overwrites live storage without clearing it.
4. The claimed checked-in byte fixture dynamically calls the encoder and checks only the header. It does not pin the full client or server format, and the required duplicate-every-field/missing-every-required-field matrix is still absent.

CI run 813 is green, including the appliance profile and TLS dependency audit, but the current tests do not exercise these cases.

**Verdict: changes required before merge.**

src/tls/session.zig
```diff
@@ -0,0 +1,2435 @@
...
+
+    pub fn eql(self: *const CompatSnapshot, other: *const CompatSnapshot) bool {
+        return self.format_id == other.format_id and
+            self.format_version == other.format_version and
+            std.mem.eql(u8, self.slice(), other.slice());
+    }
+
+    /// `out` must be zero-valued or a previously-initialized, live
+    /// `CompatSnapshot` — never `undefined` memory (see `init`).
+    pub fn cloneInto(self: *const CompatSnapshot, allocator: std.mem.Allocator, out: *CompatSnapshot) error{OutOfMemory}!void {
```
Jaruso, line 181: **Blocking lifecycle/aliasing bug:** `cloneInto` destroys `out` before it has finished reading `self`. When `out == self`, this erases the only source and then clones an empty/freed value. The same pattern exists in `ResumableSessionCommon.cloneInto`, `ClientTicketState.cloneInto`, and `ServerRecoverableState.cloneInto`.

There is a related replacement alias bug in `init`: a supported live reinitialization such as

```zig
const old = snap.slice();
try snap.init(allocator, 2, 1, old, limit);
```

calls `self.deinit()` before copying `old`, so `old` points at wiped/freed storage. `ResumableSessionCommon.init` has the same issue if `resumption_psk` or a compatibility slice is borrowed from the object being reinitialized.

Use a temporary object and commit it only after construction succeeds:

```zig
pub fn cloneInto(self: *const CompatSnapshot, allocator: Allocator, out: *CompatSnapshot) !void {
    if (self == out) return; // cloning into itself is a semantic no-op

    var next: CompatSnapshot = .{};
    errdefer next.deinit();
    next.format_id = self.format_id;
    next.format_version = self.format_version;
    try next.blob.init(allocator, self.slice().len, self.slice());
    out.moveFrom(&next); // releases the old destination only after success
}
```

Apply the same temporary-then-move pattern to all owning aggregate clones and live replacement initializers. Add tests for self-clone, live destination clone, and reinitializing from the object's own borrowed PSK/blob slices.

src/tls/session.zig
```diff
@@ -0,0 +1,2435 @@
...
+        if (snap.slice().len > limits.max_application_compat_len) return error.FieldTooLarge;
+    }
+}
+
+/// Computes the exact encoded length for `state`, but first validates every
+/// field against `limits` (ticket length, compatibility blob lengths, and
+/// total field count) rather than only checking the final byte total. This
+/// is what prevents encoding a state built under wider limits into a record
+/// a matching decoder using tighter limits would reject.
+pub fn clientEncodedLenWithLimits(state: *const ClientTicketState, limits: Limits) EncodeError!usize {
```
Jaruso, line 986: **Blocking codec/preflight gap:** this public `...WithLimits` API is not yet the authoritative preflight promised by #360.

It does not:

1. call `limits.validate()`;
2. reject a result above `limits.max_serialized_len`;
3. validate the model's semantic invariants before encoding.

As a result, both of these currently succeed incorrectly:

```zig
var invalid_limits = Limits.default;
invalid_limits.max_fields = hard_max_fields + 1;
_ = try clientEncodedLenWithLimits(&client, invalid_limits); // should InvalidLimits
```

```zig
var empty: ClientTicketState = .{};
var out: [4096]u8 = undefined;
const bytes = try encodeClient(&empty, Limits.default, &out); // currently encodes
_ = try decodeDefault(allocator, bytes); // rejects zero-length PSK/ticket/lifetime
```

A moved-from/default object is intentionally safe to hold, but it must not serialize into a record the same codec cannot decode. Make the preflight the single source of truth:

```zig
pub fn clientEncodedLenWithLimits(state: *const ClientTicketState, limits: Limits) EncodeError!usize {
    try limits.validate();
    try validateCommonForEncoding(&state.common);
    if (state.ticket.slice().len == 0 or state.ticket.slice().len > limits.max_ticket_len)
        return error.InvalidState; // or a precise existing error
    // field/blob checks...
    const needed = clientEncodedLen(state);
    if (needed > limits.max_serialized_len) return error.StateTooLarge;
    return needed;
}
```

Validate at least suite/PSK length, lifetime, early-data policy, non-empty ticket, optional SNI/ALPN invariants, compatibility bounds, field count, and total size. `encodeClient`/`encodeServer` should then only add the output-buffer check. Add tests proving every successfully encoded record decodes under the same `Limits`, including rejection of zero/moved-from states.

src/tls/session.zig
```diff
@@ -0,0 +1,2435 @@
...
+            },
+            else => {
+                if (tlv.field_id & optional_field_mask != 0) continue;
+                return error.UnknownCriticalField;
+            },
+        }
+    }
+
+    var common: ResumableSessionCommon = .{};
+    try buildCommon(allocator, limits, common_fields, &common);
```
Jaruso, line 1349: **Blocking two-pass decode violation:** the decoder allocates/copies the common PSK and compatibility blobs here before checking whether the client-only mandatory fields (`ticket`, `ticket_age_add`, `ticket_nonce`, `received_at`) are present. A structurally invalid record missing one of those fields can therefore force attacker-sized allocations before returning `MissingField`.

The authoritative plan requires the complete validation/index pass to finish before *any* owned allocation. Resolve all mandatory borrowed fields first, then build:

```zig
const ticket_bytes = ticket orelse return error.MissingField;
const age_add = ticket_age_add orelse return error.MissingField;
const nonce = ticket_nonce orelse return error.MissingField;
const received = received_at_unix_ms orelse return error.MissingField;

var common: ResumableSessionCommon = .{};
try buildCommon(allocator, limits, common_fields, &common);
errdefer common.deinit();

var state: ClientTicketState = .{};
try state.init(allocator, limits, &common, .{
    .ticket = ticket_bytes,
    .ticket_age_add = age_add,
    .ticket_nonce = nonce,
    .received_at_unix_ms = received,
});
```

Add a counting/failing allocator test for each missing client-only mandatory field and assert the result is `MissingField` without any allocation attempt.

src/tls/session.zig
```diff
@@ -0,0 +1,2435 @@
...
+        if (tlv.field_id == field_id_to_duplicate) value_to_duplicate = tlv.value;
+    }
+    writeTlv(out, &pos, field_id_to_duplicate, value_to_duplicate.?);
+    patchSectionLen(out, pos);
+    return pos;
+}
+
+test "duplicating any known client field, mandatory or optional, is rejected" {
+    const all_client_field_ids = [_]u16{
+        field_cipher_suite, field_resumption_psk, field_server_name,      field_application_protocol,
```
Jaruso: **Blocking acceptance-test gap:** this is still not the requested “every known field” matrix. `field_transport_compat` and `field_application_compat` are known fields but are absent from both fixtures and both field-id arrays, so their duplicate handling and literal encoding are not pinned.

Make at least one literal fixture fully populated with both compatibility snapshots (and preferably a non-default early-data policy), then include both compatibility field IDs in the client/server duplicate loops. Removing those optional fields should decode to `null`; duplicating either must return `DuplicateField`.

Also split the current compatibility-limit test so `max_application_compat_len` is actually exercised with an application snapshot—the existing test constructs only `transport_compat`, despite its name claiming both.

Jaruso (commented):
Re-review of current head `56fc011d` against the authoritative #360 body/comment.

The four findings from the previous round are substantially addressed, and CI run 816 is green. This is much closer, but I still do **not** consider it merge-ready because the current ownership replacement APIs are not alias-safe, the encoder can emit internally invalid records that the matching decoder rejects, and `decodeClient` still violates the required validate-first/allocate-second pass boundary.

The inline comments below are blocking. Before merging, please also update the PR description: it still describes a TLS-local `TranscriptHash`/`digestLen()` type even though the implementation correctly changed to `crypto.provider.Hash`/`digestLength()`.

Remaining acceptance-test work after the code fixes:

- add self-clone and self-borrowed reinitialization tests for every owning aggregate;
- add encode rejection tests for default/moved-from and semantically corrupted state;
- prove malformed records missing client-only mandatory fields perform zero allocations;
- include `transport_compat` and `application_compat` in the literal fixture/duplicate-field matrix;
- exercise `max_application_compat_len` independently (the current test name mentions it, but the constructed state contains only transport compatibility);
- add at least one aggregate-level fixed-buffer test that inspects backing memory after normal deinit and after an allocation-failure path, rather than checking only leak freedom.

Once these are fixed, I expect the issue contract to be satisfied.

Jaruso (commented):
Re-review of `167949023fe881237f29435e76dcbd190203941b` against the complete #360 issue body and authoritative implementation plan.

The previous four findings are addressed, and CI run 818 is green. This is materially closer, but I still do not consider it merge-ready for the following reasons.

### 1. Blocking: `ClientTicketState.init` destroys the source session on a failed initialization

`next.common.moveFrom(common)` occurs before ticket/nonce validation and before the ticket allocation. Consequently `TicketTooLarge`, `NonceTooLarge`, or `OutOfMemory` causes `errdefer next.deinit()` to wipe and destroy the caller's `common`, while returning no initialized ticket state.

That is the opposite of the required ownership rule: **successful** transfer invalidates the source. It also contradicts the move-last pattern from the maintainer plan. The current test explicitly codifies the undesirable consume-on-failure behavior.

Please validate and construct all fallible client-only state first, then perform the non-failing common-session move last:

```zig
pub fn init(
    self: *ClientTicketState,
    allocator: Allocator,
    limits: Limits,
    common: *ResumableSessionCommon,
    params: InitParams,
) InitError!void {
    try limits.validate();
    if (params.ticket.len == 0 or params.ticket.len > limits.max_ticket_len)
        return error.TicketTooLarge;
    if (params.ticket_nonce.len > max_ticket_nonce_len)
        return error.NonceTooLarge;

    var next: ClientTicketState = .{};
    errdefer next.deinit();
    try next.ticket.init(allocator, params.ticket.len, params.ticket);
    try next.ticket_nonce.replace(params.ticket_nonce);
    next.ticket_age_add = params.ticket_age_add;
    next.received_at_unix_ms = params.received_at_unix_ms;

    // Last fallible operation has completed.
    next.common.moveFrom(common);
    self.moveFrom(&next);
}
```

Add explicit too-large, nonce-too-large, and allocation-failure tests proving both `common` and an already-live `self` remain unchanged on failure. On success, prove only then that `common` is invalidated.

### 2. Blocking: encoding preflight still does not enforce canonical/decodable SNI state

`validateCommonForEncoding` checks only that a present SNI slice is non-empty. The model's fields are publicly accessible, so this permits a state such as:

```zig
var common = try sampleCommon(...);
common.server_name.?.bytes[0] = '_';
```

`encodeServer` currently accepts that state, while `decode` rejects the resulting record through `dns_name.validateHostName`. Similarly, manually storing uppercase SNI bytes allows encoding a non-canonical record that decode lowercases, breaking encode/decode/encode byte stability. A corrupted `SniName.len > max_sni_len` can also trap in `slice()` instead of returning `InvalidState`.

The preflight claims that every accepted state decodes cleanly under the same limits, so it must validate the actual stored SNI, not only its presence:

- check `s.len` before calling `slice()`;
- call `dns_name.validateHostName` on the stored bytes;
- require the stored representation to already be ASCII-lowercase/canonical;
- return `InvalidState`, never panic, for an invalid public-field state.

Apply the same raw-length-before-slice discipline to the publicly mutable `FixedSecret`/`BoundedSecret` length fields used by preflight (`resumption_psk`, ticket, and ticket nonce), so malformed in-process state is rejected deterministically instead of reaching an out-of-bounds slice.

Add tests for invalid DNS bytes, a non-canonical uppercase stored SNI, over-capacity stored lengths, and encode/decode/encode stability.

### 3. Blocking acceptance gap: the required malformed-codec matrix is still absent

The authoritative #360 plan explicitly requires:

- truncating **both literal fixtures at every byte boundary**;
- mutating every TLV length field through zero, one-under, one-over, and relevant maximum values;
- invalid cipher-suite enum values;
- invalid decoded PSK/suite combinations;
- exact 31/32/33-byte SHA-256 and 47/48/49-byte SHA-384 construction/decode coverage;
- unknown-optional decode followed by canonical re-encoding that proves the unknown field is omitted;
- exact-boundary and one-over tests for ticket, nonce, SNI, ALPN, transport compatibility, application compatibility, and total serialized size.

The current suite has section-length tests, some broad limits tests, and literal fixtures, but it does not contain these exhaustive cases. These are especially important because this is a security-sensitive persisted-input decoder.

### 4. Blocking acceptance gap: leak checks are not zeroization verification

`checkAllAllocationFailures` proves that allocations are not leaked; it does not prove that secret-bearing capacity was wiped before free. The required tests also call for:

- replacing PSK/ticket values and directly verifying old bytes/tails are cleared;
- deinit clearing the **complete allocated capacity**, not only logical length;
- a fully populated decode allocation-failure sweep across transport snapshot, application snapshot, and ticket allocation;
- failed decode after owned allocations, with backing memory inspected to prove ticket/compatibility/secret bytes were zeroed;
- failed clone/replacement paths leaving no secret bytes and preserving the old destination until commit.

Please use a fixed/poisoning or recording allocator whose freed backing storage remains inspectable. The existing `crypto.secrets` tests are useful, but #360 requires the aggregate model/codec error paths to demonstrate that they invoke those lifecycles correctly.

### Documentation

The PR description is still stale: it says `CipherSuite.transcriptHash()` returns a TLS-local `TranscriptHash` with `digestLen()`. The implementation correctly returns `crypto.provider.Hash` and uses `digestLength()`. Please update the summary and test counts before merge.

Once the ownership failure semantics, canonical preflight, malformed-input matrix, and aggregate zeroization matrix are complete, I expect this to be ready.

Jaruso (commented):
Re-review complete on head `da6c7713`.

I rechecked the latest commit against the four remaining blockers from review 4735434249 and the full #360 acceptance contract. The implementation now:

- leaves both `common` and an existing destination untouched on every failed `ClientTicketState.init` path, moving `common` only after all fallible ticket/nonce work succeeds;
- rejects invalid, noncanonical, or over-capacity model state before any encoder slice/write operation;
- covers both literal fixtures with truncation, malformed-length, invalid-suite/PSK, canonical re-encoding, boundary, allocation-failure, and backing-memory zeroization tests;
- keeps the deterministic codec, ownership transfer, deep-clone, compatibility, and limit semantics aligned between encode and decode.

CI run 820 is green across the full test matrix, appliance profiles, ReleaseSafe/ReleaseFast builds, packaging, format, security, and TLS dependency audit.

I found no remaining blocking correctness, security, ownership, scope, or acceptance-criteria issue. **PR #475 is merge-ready.**

GitHub does not permit the connected account to formally approve its own PR, so this is posted as a comment review rather than an `APPROVE` review.

PR #476, opened 2026-07-20
packaging: add macOS (Darwin) archives to the release workflow

Jaruso:
## Summary

Reconciled against current `main` on **2026-08-22**, including the first #634 implementation slice from #641 (`-Dtls-profile=native`).

This PR is now the **release/packaging cutover slice** for the native shipping artifact and the Darwin archive lane needed before the next intentional Tardigrade version can be published.

### What this PR now does

- adds native Intel and Apple Silicon release rows:
  - `macos-15-intel` -> `tardigrade-darwin-x86_64`
  - `macos-15` -> `tardigrade-darwin-arm64`
- builds **every official release matrix artifact** (Linux and Darwin) with `-Dtls-profile=native`;
- removes OpenSSL development/runtime prerequisites from the release workflow and Darwin release smoke;
- audits each official release binary with `scripts/audit-release-binary.sh --profile native`, which rejects `libssl`, `libcrypto`, and other forbidden foreign TLS/crypto/QUIC/H3 linkage and checks the binary self-reports the native backend;
- keeps Linux-only source audit / DEB / RPM work gated to Linux while sharing archive, SBOM, dependency inventory, provenance, upload, checksum, and publication behavior across all four archive rows;
- makes DEB/RPM OpenSSL dependency metadata match the binary being packaged:
  - native binary -> no OpenSSL runtime dependency;
  - transitional `openssl-adapter` binary -> retain the dependency it genuinely needs;
  - unknown backend -> fail closed;
  - cross-compiled binaries remain supported via explicit `--tls-backend native|openssl-adapter` instead of requiring the target binary to execute on the packaging host;
- adds an explicit release-job assertion that native DEB/RPM metadata contains no OpenSSL/libssl/libcrypto dependency;
- verifies Darwin runner architecture matches the public archive name before packaging;
- runs dedicated PR-gated release smoke on Intel and Apple Silicon without installing Homebrew OpenSSL;
- validates archive extraction, `tardigrade -> tardi`, `tls-profile=native`, and `tls-backend=native`;
- starts the **extracted release artifact**, serves a real static request plus `/health`, and shuts it down cleanly on both Darwin architectures;
- exercises the checksum-verifying `scripts/install.sh` path on both Darwin runners;
- makes the quickstart use the shipping-equivalent native source build;
- preserves the release hardening added during reconciliation:
  - `workflow_run` publishes the exact `github.event.workflow_run.head_sha` whose CI succeeded;
  - prepare/build/publish all consume that same SHA;
  - the release tag is not created until every matrix artifact has built successfully;
  - an existing tag must resolve to that exact release SHA.

## Relationship to #634

#641 established the general-purpose native build/audit contract this PR needed. #476 now consumes that contract rather than using `appliance` as a packaging workaround.

The repository's plain/default source build remains the transitional `general` profile until later #634 slices change that default. That does **not** affect the official release contract introduced here: release artifacts are explicitly built and audited as `native`.

Remaining #634 work such as broader feature-parity/default-build cleanup and the Docker cutover is intentionally not folded into this packaging PR.

## Release-state rule

This PR deliberately does **not** modify `CHANGELOG.md`, choose the next semantic version, or cut a release solely to close #463.

The currently published latest release still predates these Darwin assets. Merging this PR completes the implementation lane. The next release should be intentional: add the chosen semantic-version changelog heading when the project is actually ready to publish.

## Before merge

- [x] Based on current `main`, including #641.
- [x] Darwin x86_64 uses `macos-15-intel`; arm64 uses `macos-15`.
- [x] Official Linux and Darwin release artifacts explicitly build `-Dtls-profile=native`.
- [x] Release/Darwin Tardigrade build path does not install OpenSSL.
- [x] Native release audit rejects OpenSSL and other forbidden foreign protocol/crypto linkage.
- [x] Darwin archive architecture assertions are present.
- [x] Shared archive/SBOM/inventory/provenance/checksum publication path is preserved.
- [x] Native DEB/RPM metadata cannot retain an OpenSSL runtime dependency unnoticed.
- [x] Cross-compiled DEB/RPM package builds remain supported via explicit backend metadata.
- [x] Extracted Darwin archive receives a real startup/request smoke.
- [x] Installer smoke covers Darwin platform mapping, checksums, install, alias, and version.
- [x] README / packaging / release checklist / quickstart describe the native release contract without claiming Darwin assets are already published.
- [x] Initial Darwin archives are documented as unsigned and not notarized.
- [ ] Darwin Release Smoke — `macos-15-intel` green on final head.
- [ ] Darwin Release Smoke — `macos-15` green on final head.
- [ ] Normal repository CI green on final head.

## First intentional release verification for #463

After the first release containing this PR is published:

- [ ] `tardigrade-darwin-x86_64.tar.gz` is attached.
- [ ] `tardigrade-darwin-arm64.tar.gz` is attached.
- [ ] Matching SPDX SBOMs and dependency inventories are attached for both.
- [ ] `tardigrade-checksums.txt` verifies the public bytes.
- [ ] `gh attestation verify <archive> --repo Bare-Systems/Tardigrade` succeeds for both.
- [ ] Both released binaries pass native-only linkage audit and self-report native profile/backend.
- [ ] Published `install.sh` works on Intel macOS without OpenSSL as a Tardigrade prerequisite.
- [ ] Published `install.sh` works on Apple Silicon macOS without OpenSSL as a Tardigrade prerequisite.
- [ ] Installed `tardi version`, `tardigrade` alias, and a minimal real startup/request path work on both architectures.
- [ ] Close #463 after that publication evidence is recorded.

## Scope boundaries

This PR does **not** publish/fix the Homebrew tap or formula (#466), validate launchd lifecycle (#467), add signing/notarization, create a universal binary, choose the next release version, finish the whole #634 architecture migration, or claim the currently published release already contains these assets.

Refs #463, #634, #641


.github/workflows/release.yml
```diff
@@ -99,10 +99,18 @@ jobs:
...
             rpm_arch: x86_64
+            linux_packages: true
           - os: ubuntu-24.04-arm
             archive_name: tardigrade-linux-aarch64
             deb_arch: arm64
             rpm_arch: aarch64
+            linux_packages: true
+          - os: macos-15-intel
+            archive_name: tardigrade-darwin-x86_64
+            linux_packages: false
```
Jaruso, line 110: These are the new production paths, but this workflow only runs after CI on `main`; the PR's green CI does not execute either row. That leaves `macos-15-intel`, the Homebrew OpenSSL setup, `otool` audit, archive creation, and SBOM/upload compatibility untested until a real release. #463 requires native evidence for both architectures. Please add a PR-gated smoke matrix in `ci.yml` using these same runner labels and at least run:

```yaml
- brew list openssl@3 >/dev/null 2>&1 || brew install openssl@3
- zig build -Doptimize=ReleaseFast -Dversion=0.0.0-test
- ./scripts/audit-release-binary.sh --binary zig-out/bin/tardi --profile general
- ./scripts/package-release-archive.sh --archive-name "$ARCHIVE_NAME" ...
- extract the archive and run `tardi version`
```

The architecture assertion should run there too. This prevents the first Intel execution from occurring during an actual release.

packaging/README.md
```diff
@@ -159,10 +161,11 @@ sudo systemctl reload tardigrade   # or: restart, if reload is insufficient
 
 ## Homebrew (macOS and Linux)
 
-The `on_macos` blocks in this formula cannot resolve today: the release
-workflow does not build or publish `tardigrade-darwin-*.tar.gz` archives (see
-"Current status" above). The `on_linux` blocks reference archives that are
-published, so a Linux install can work once real checksums are filled in.
+The Darwin URLs referenced by the `on_macos` blocks now resolve, since the
```
Jaruso: This says the formula's Darwin URLs now resolve, but the checked-in formula still has `version "0.50"`, which constructs a `v0.50` tag rather than the repository's `v0.5.0` tag. It also still has placeholder checksums and no `depends_on "openssl@3"`, so it remains non-installable.

Please keep the actual formula work in #466, but reword this to something like:

> The release workflow now produces the Darwin archive names expected by the formula. The checked-in formula remains non-installable until #466 corrects its release version, fills the checksums, declares the OpenSSL dependency, and publishes the tap.

packaging/README.md
```diff
@@ -10,12 +10,12 @@ versus what is a local-build-only tool.
 | Format | Status | Notes |
 | --- | --- | --- |
 | Linux release archives (`.tar.gz`, x86_64/aarch64) | **Supported, published** | Built and attached to every GitHub release by `.github/workflows/release.yml`, alongside `install.sh`, `tardigrade-checksums.txt`, and per-arch SPDX SBOMs. |
-| macOS release archives (`.tar.gz`, darwin x86_64/arm64) | **Planned, not published** | CI builds and tests Tardigrade on `macos-14`, but the release workflow's build matrix only packages Linux. `install.sh` and the Homebrew formula already expect `tardigrade-darwin-*.tar.gz` assets that do not yet exist. |
+| macOS release archives (`.tar.gz`, darwin x86_64/arm64) | **Supported, published** | Built natively on `macos-15-intel` (x86_64) and `macos-15` (arm64) runners by the same release workflow, and attached to every GitHub release with per-arch SPDX SBOMs, dependency inventories, checksum entries, and provenance attestations. Requires `brew install openssl@3` on the host at runtime — see the Homebrew section below. |
```
Jaruso, line 13: #463 explicitly calls for code signing/notarization to be recorded as a known gap. Please add a nearby note such as:

> **macOS signing status:** Darwin archives are currently unsigned and not notarized. Command-line installation is supported, but browser-downloaded artifacts may be subject to Gatekeeper/quarantine behavior. Signing and notarization are tracked separately.

Without this, changing the row to **Supported, published** overstates the current distribution experience.

Jaruso (commented):
The workflow implementation is well scoped and the ordinary CI suite is green, including actionlint. I found three remaining gaps before I would merge:

1. Neither new release matrix row runs on this PR. The release workflow only runs after successful CI on `main`, so the `macos-15-intel` path—and the complete release archive/SBOM/audit/upload chain on either architecture—has not been exercised. Issue #463 explicitly requires native evidence for both Intel and Apple Silicon. Please add a PR-gated Darwin release-archive smoke job (or an equivalent safe reusable/manual validation path) that runs the build, architecture check, binary audit, packaging, and extracted-binary smoke test on both runner labels.
2. The packaging documentation currently overstates the Homebrew formula's readiness. The archives will exist, but the checked-in formula still uses `version "0.50"` (constructing `v0.50`, not `v0.5.0`), contains checksum placeholders, and does not declare `openssl@3`. Keep those fixes in #466, but describe the formula as still non-installable until that issue lands.
3. #463 explicitly requires the unsigned/notarized macOS distribution gap to be recorded. Add a clear Gatekeeper/quarantine note to the packaging documentation.

Once these are addressed and both native smoke jobs pass, I expect this to be ready to merge.

packaging/README.md
```diff
@@ -159,12 +161,25 @@ sudo systemctl reload tardigrade   # or: restart, if reload is insufficient
 
 ## Homebrew (macOS and Linux)
 
-The `on_macos` blocks in this formula cannot resolve today: the release
-workflow does not build or publish `tardigrade-darwin-*.tar.gz` archives (see
-"Current status" above). The `on_linux` blocks reference archives that are
-published, so a Linux install can work once real checksums are filled in.
+The Darwin URLs referenced by the `on_macos` blocks now resolve, since the
```
Jaruso: This is improved, but the sentence still says the formula's Darwin URLs “now resolve.” They do not: as the bullets immediately below explain, `version "0.50"` interpolates a nonexistent `v0.50` tag. The same contradiction appears in the Current status table.

Please use wording such as:

> The release workflow now produces the Darwin **archive filenames expected by the formula**. The formula's generated URLs remain invalid until #466 corrects its version, checksums, and OpenSSL dependency.

That preserves the scope boundary without claiming the current URLs resolve.

Jaruso (commented):
Re-review of current head `0bd5d3c0`:

The new two-architecture Darwin smoke matrix is a strong fix, and the current CI run is green. The signing/notarization gap is also documented clearly. I resolved those two prior threads.

One documentation issue remains: `packaging/README.md` still says the Darwin URLs referenced by the formula “now resolve” / “resolve to real release assets,” but the same paragraph correctly explains that `version "0.50"` generates a nonexistent `v0.50` URL. Those statements contradict each other. Please change both occurrences to say the release workflow now produces the **archive filenames expected by the formula**, while the formula’s generated URLs remain invalid until #466 fixes the version/checksums/dependency.

The PR description’s Intel test checkbox is also stale now that the PR-gated `macos-15-intel` smoke leg has run successfully; updating it would make the validation record accurate.

After the contradictory URL wording is corrected, I do not see another code or workflow blocker.

README.md
```diff
@@ -74,17 +74,20 @@ The fastest way to install the latest release is the official install script:
...
 ```
 
-The installer downloads the matching Linux release archive (`x86_64` or
-`aarch64`), verifies it against `tardigrade-checksums.txt`, and installs the
-`tardi` binary (with a `tardigrade` compatibility alias) into
-`$HOME/.local/bin` by default. Published archives currently cover Linux
-only; macOS is not yet built or published by the release workflow (see
-[packaging/README.md](packaging/README.md#current-status) for exact status).
+The installer downloads the matching release archive for Linux (`x86_64` or
+`aarch64`) or macOS (`x86_64` or `arm64`), verifies it against
```
Jaruso, line 78: This currently describes behavior of `releases/latest`, but merging the PR will not publish a new release. `release-metadata.sh` skips `[Unreleased]` and still resolves the existing `0.5.0`; because `v0.5.0` points to the old release commit, `prepare-release` sets `should_publish=false` and skips all four `build-release` rows. Therefore the latest installer still has no Darwin asset after merge.

Please either:

1. land this together with the next semantic-version release so the assets actually exist, or
2. say “available starting with the first release containing #463” here and in the packaging status/changelog, remove `Closes #463`, and keep the issue open until the release is verified.

Without one of those, the PR closes the issue and marks macOS **Supported, published** while the documented `releases/latest` command still fails on macOS.

Jaruso (commented):
The previous Homebrew wording issue is fixed, and I resolved those outdated threads. One release-state blocker remains: the current branch still has only an `[Unreleased]` section above the existing `0.5.0` release. `scripts/release-metadata.sh` ignores `[Unreleased]`, resolves `0.5.0`, and `release.yml` sets `should_publish=false` when the existing `v0.5.0` tag points at a different commit. As a result, merging this PR will not run `build-release`, the latest release will still lack Darwin assets, and the documented `releases/latest` installer will still 404 on macOS.

Please either coordinate/cut a new semantic version as part of landing this work, or make the documentation explicitly future-scoped and do not close #463 until a release containing the Darwin assets is published. The current CI run for the latest head is also still in progress, so it should complete before merge.

Jaruso (commented):
Re-reviewed the latest head (`962ad7f4`). All previously reported blockers are addressed.

- `CHANGELOG.md` now introduces `0.5.1`, so `release-metadata.sh` resolves a new tag and the post-merge release workflow will actually execute the Linux and Darwin matrix rows rather than skipping publication against the existing `v0.5.0` tag.
- Both native Darwin smoke jobs are covered in PR CI, including OpenSSL setup, ReleaseFast build, architecture assertion, `otool`-backed binary audit, archive construction, extraction, symlink validation, and execution.
- The Homebrew documentation now accurately distinguishes matching archive filenames from the formula's still-invalid generated URLs and keeps the formula work scoped to #466.
- The unsigned/notarized Gatekeeper limitation is clearly documented.
- Linux-only source audit and DEB/RPM work remain correctly gated, while archive, SBOM, inventory, attestation, and uploads remain shared by all release rows.
- The latest CI run completed successfully, including actionlint/shellcheck and the Darwin matrix.

I found no remaining blocking code, workflow, or documentation issue. This is ready to merge. The only remaining validation is the intentionally post-merge first-release check that the `v0.5.1` assets, checksums, SBOMs, inventories, and attestations were published correctly.

Jaruso:
**2026-08-19 reconciliation note:** this PR branch was rebuilt from current `main` because the July head had become stale. The pre-reconciliation head is preserved at `archive/pr-476-pre-reconcile-2026-08-19`.

Current head: `14c5cf174ceb5f0437895fe7d0cd7ecfb5d5b55b`.

The old July review history remains useful for the original design, but any comments tied to the former `0.5.1` release-trigger strategy are superseded. The refreshed PR deliberately does **not** choose the next semantic version and uses `Refs #463` rather than auto-closing it. #463 now closes only after the first intentional release containing this implementation is verified end to end.

Fresh CI is authoritative for this head. In particular, both `Darwin Release Smoke` legs (`macos-15-intel` and `macos-15`) plus normal repository CI must be green before merge.

.github/workflows/darwin-release-smoke.yml
```diff
@@ -0,0 +1,93 @@
...
+          - os: macos-15
+            archive_name: tardigrade-darwin-arm64
+            expected_arch: arm64
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+        with:
+          persist-credentials: false
+
```
Jaruso: **P1 — this makes OpenSSL a production prerequisite for the Darwin artifact, which conflicts with #634.** The Darwin release smoke must prove the artifact users receive is the native Zig Tardigrade implementation and does not require Homebrew `openssl@3` merely to build/run Tardigrade. Remove this production dependency and make the smoke fail if the resulting `tardi` links `libssl`/`libcrypto`. External OpenSSL installs remain fine in separate interop/oracle jobs.

.github/workflows/darwin-release-smoke.yml
```diff
@@ -0,0 +1,93 @@
...
+      - name: Print Zig version
+        run: zig version
+
+      - name: Build release binary
+        run: zig build -Doptimize=ReleaseFast -Dversion=0.0.0-ci-smoke
+
+      - name: Verify macOS binary architecture
+        env:
+          EXPECTED_ARCH: ${{ matrix.expected_arch }}
```
Jaruso: **P1 — `--profile general` currently means the OpenSSL-backed production artifact, which is no longer an acceptable shipping target.** Reconcile this with #634 so the release audit validates the native shipping implementation and treats foreign TLS/crypto linkage as a failure. If a product-policy identifier remains useful, it must describe native policy (general vs appliance behavior), not select OpenSSL vs Zig implementation.

Jaruso (commented):
Architecture reconciliation after #634: this PR's Darwin packaging mechanics are still useful, but the shipping artifact contract changed. Tardigrade's final architecture is now that **every deployed `tardi` artifact is the native Zig implementation**; OpenSSL and other foreign TLS/crypto/protocol implementations are assurance/interop/benchmark peers only, not production backends or runtime dependencies.

Because this PR is authored by the same GitHub identity I cannot submit a formal REQUEST_CHANGES review, so treat the comments below as merge-blocking requested changes.

Please keep the good parts of this PR—dual Darwin runners, architecture assertions, shared archive/SBOM/inventory/provenance/checksum publication, extracted-archive smoke, and the separation from the eventual intentional release—but rework the artifact being built and audited:

- Darwin release/smoke must not install `openssl@3` merely to build or run Tardigrade.
- Build the same native Zig shipping implementation required by #634.
- The binary audit must reject `libssl`/`libcrypto` and other foreign TLS/crypto/QUIC/H3 linkage for the published Darwin artifact rather than bless `--profile general` as OpenSSL-backed production.
- The extracted archive smoke should execute `tardi version` and a minimal production-shaped startup/request path without OpenSSL as a Tardigrade prerequisite.
- Update README/packaging/release-checklist wording accordingly.
- Keep external OpenSSL/interop tooling in tests where useful; the restriction is on the shipped artifact, not independent validation peers.

#463 is being reconciled to consume #634 as the architecture source of truth. This PR should not merge in its current OpenSSL-backed shipping shape even if the Darwin matrix itself is green.

PR #477, opened 2026-07-20
tls: implement TLS 1.3 NewSessionTicket

Jaruso:
## Summary

- add RFC 8446 NewSessionTicket encode/decode support with bounds, extension validation, and conversion into the existing TLS session model
- derive resumption master secrets at handshake completion and per-ticket PSKs from ticket nonces
- add client ticket callback delivery and explicit server application-epoch ticket emission APIs

## Validation

- zig build test-integration --summary all --error-style verbose
- zig build -Doptimize=ReleaseFast --summary all --error-style verbose

Closes #361

### DIFF
--- a/src/edge_gateway.zig
+++ b/src/edge_gateway.zig
@@ -2848,7 +2848,7 @@ const GatewayPendingNative = struct {
             .{ .server_name = "tardigrade.test" },
         );
         self.client_carrier = .{ .fd = fds[1] };
-        self.client_stream = try tls_core.encrypted_stream.PureZigRecordStream.initWithCarrierBackendAndAllocator(
+        self.client_stream = try tls_core.encrypted_stream.PureZigRecordStream.initWithCarrierAndBackend(
             allocator,
             .client,
             self.client_provider.cryptoProvider(),
--- a/src/edge_gateway.zig
+++ b/src/edge_gateway.zig
@@ -2848,7 +2848,8 @@ const GatewayPendingNative = struct {
             .{ .server_name = "tardigrade.test" },
         );
         self.client_carrier = .{ .fd = fds[1] };
-        self.client_stream = tls_core.encrypted_stream.PureZigRecordStream.initWithCarrierAndBackend(
+        self.client_stream = try tls_core.encrypted_stream.PureZigRecordStream.initWithCarrierBackendAndAllocator(
+            allocator,
             .client,
             self.client_provider.cryptoProvider(),
             .tls_aes_128_gcm_sha256,
--- a/src/http/native_tls_connection.zig
+++ b/src/http/native_tls_connection.zig
@@ -196,7 +196,8 @@ pub const NativeTlsConnection = struct {
             .policy = policy,
         };
         self.crypto_provider_state = production_crypto.Provider.init(self.entropy_source.entropy());
-        record.* = try encrypted_stream.PureZigRecordStream.initWithCarrierBackendAndLimits(
+        record.* = try encrypted_stream.PureZigRecordStream.initWithCarrierBackendAllocatorAndLimits(
+            allocator,
             .server,
             self.crypto_provider_state.cryptoProvider(),
             .tls_aes_128_gcm_sha256,
--- a/src/http/native_tls_connection.zig
+++ b/src/http/native_tls_connection.zig
@@ -196,7 +196,7 @@ pub const NativeTlsConnection = struct {
             .policy = policy,
         };
         self.crypto_provider_state = production_crypto.Provider.init(self.entropy_source.entropy());
-        record.* = try encrypted_stream.PureZigRecordStream.initWithCarrierBackendAllocatorAndLimits(
+        record.* = try encrypted_stream.PureZigRecordStream.initWithCarrierBackendAndLimits(
             allocator,
             .server,
             self.crypto_provider_state.cryptoProvider(),
--- a/src/quic/connection.zig
+++ b/src/quic/connection.zig
@@ -57,6 +57,7 @@ pub const State = enum {
 /// until DPLPMTUD lands under #256.
 pub const max_datagram_size: usize = recovery.max_datagram_size;
 pub const max_application_crypto_outstanding: usize = 2 * tls_core.tls13_transport.max_emitted_new_session_ticket_message_len;
+const min_application_crypto_payload: usize = max_datagram_size / 2;
 /// RFC 9000 §14.1: datagrams carrying Initial packets are padded to 1200.
 pub const min_initial_datagram: usize = 1200;
 
@@ -208,6 +209,28 @@ const RangeList = struct {
         try self.items.insert(allocator, index, merged);
     }
 
+    fn ensureUnusedCapacity(self: *RangeList, allocator: std.mem.Allocator, additional_count: usize) !void {
+        try self.items.ensureUnusedCapacity(allocator, additional_count);
+    }
+
+    fn insertAssumeCapacity(self: *RangeList, incoming: Range) void {
+        if (incoming.start >= incoming.end) return;
+        var merged = incoming;
+        var index: usize = 0;
+        while (index < self.items.items.len) {
+            const current = self.items.items[index];
+            if (merged.end < current.start) break;
+            if (current.end < merged.start) {
+                index += 1;
+                continue;
+            }
+            merged.start = @min(merged.start, current.start);
+            merged.end = @max(merged.end, current.end);
+            _ = self.items.orderedRemove(index);
+        }
+        self.items.insertAssumeCapacity(index, merged);
+    }
+
     /// Remove and return up to `max_len` bytes from the lowest range.
     fn takeFirst(self: *RangeList, max_len: u64) ?Range {
         if (self.items.items.len == 0) return null;
@@ -266,11 +289,11 @@ const CryptoTx = struct {
 
     fn markAcked(self: *CryptoTx, allocator: std.mem.Allocator, range: Range) void {
         const live = self.liveRange(range) orelse return;
-        self.acked.insert(allocator, live) catch {};
-        self.compactAcked();
+        self.acked.insertAssumeCapacity(live);
+        self.compactAcked(allocator);
     }
 
-    fn compactAcked(self: *CryptoTx) void {
+    fn compactAcked(self: *CryptoTx, allocator: std.mem.Allocator) void {
         while (self.acked.items.items.len > 0 and self.acked.items.items[0].end <= self.base) {
             _ = self.acked.items.orderedRemove(0);
         }
@@ -295,6 +318,12 @@ const CryptoTx = struct {
             self.acked.items.items[0].start = self.base;
         }
         self.trimPendingBelowBase();
+        if (self.data.items.len == 0 and self.data.capacity > 0) {
+            crypto_secrets.secureZero(self.data.allocatedSlice());
+            self.data.clearAndFree(allocator);
+            self.pending.items.clearAndFree(allocator);
+            self.acked.items.clearAndFree(allocator);
+        }
     }
 
     fn trimPendingBelowBase(self: *CryptoTx) void {
@@ -313,11 +342,13 @@ const CryptoTx = struct {
         if (bytes_len > max_outstanding) return error.CryptoBufferTooLarge;
         if (self.data.items.len > max_outstanding - bytes_len) return error.CryptoBufferTooLarge;
         if (self.data.capacity == 0) {
-            try self.data.ensureTotalCapacity(allocator, max_outstanding);
+            try self.data.ensureTotalCapacityPrecise(allocator, bytes_len);
         } else if (self.data.capacity < self.data.items.len + bytes_len) {
             return error.CryptoBufferTooLarge;
         }
-        try self.pending.items.ensureUnusedCapacity(allocator, 1);
+        const range_budget = applicationCryptoRangeBudget(bytes_len);
+        try self.pending.ensureUnusedCapacity(allocator, range_budget + 1);
+        try self.acked.ensureUnusedCapacity(allocator, range_budget);
     }
 
     fn appendReserved(self: *CryptoTx, bytes: []const u8) void {
@@ -329,6 +360,10 @@ const CryptoTx = struct {
         });
     }
 };
+
+fn applicationCryptoRangeBudget(bytes_len: usize) usize {
+    return (bytes_len + min_application_crypto_payload - 1) / min_application_crypto_payload + 8;
+}
 
 /// Driver-owned transmit buffer for one stream: bytes the application handed
 /// to `writeStream` that are unsent or unacked. `base` is the stream offset
@@ -1109,7 +1144,13 @@ pub const Connection = struct {
             };
             // If the level's keys are already discarded the buffer is gone.
             if (tx.data.items.len > 0) {
-                if (tx.liveRange(range)) |live| tx.pending.insert(self.allocator, live) catch {};
+                if (tx.liveRange(range)) |live| {
+                    if (record.space == .application) {
+                        tx.pending.insertAssumeCapacity(live);
+                    } else {
+                        tx.pending.insert(self.allocator, live) catch {};
+                    }
+                }
             }
         }
         for (record.streams[0..record.stream_count]) |sr| {
@@ -1271,6 +1312,8 @@ pub const Connection = struct {
         limits: tls_core.session.Limits,
     ) tls_handshake.HandshakeError!tls_core.session.ServerRecoverableState {
         if (self.state_ != .established or self.role != .server) return error.InvalidHandshakeState;
+        limits.validate() catch return error.IllegalParameter;
+        if (params.opaque_ticket.len == 0 or params.opaque_ticket.len > limits.max_ticket_len) return error.TicketTooLarge;
         const emit_params: tls_core.new_session_ticket.EmitParams = .{
             .ticket_lifetime = params.ticket_lifetime,
             .ticket_age_add = params.ticket_age_add,
@@ -1890,13 +1933,21 @@ pub const Connection = struct {
                 range = tx.liveRange(range) orelse continue;
                 if (record.crypto) |existing| {
                     if (existing.end != range.start) {
-                        tx.pending.insert(self.allocator, range) catch {};
+                        if (space == .application) {
+                            tx.pending.insertAssumeCapacity(range);
+                        } else {
+                            tx.pending.insert(self.allocator, range) catch {};
+                        }
                         break;
                     }
                 }
                 const data = tx.slice(range);
                 const n = frame.encodeCrypto(range.start, data, plain[plain_len..plain_budget]) catch {
-                    tx.pending.insert(self.allocator, range) catch {};
+                    if (space == .application) {
+                        tx.pending.insertAssumeCapacity(range);
+                    } else {
+                        tx.pending.insert(self.allocator, range) catch {};
+                    }
                     break;
                 };
                 plain_len += n;
@@ -2266,10 +2317,8 @@ test "application CRYPTO transmit ignores late duplicate ACKs and compacts later
     var tx = CryptoTx{};
     defer tx.deinit(allocator);
 
-    try tx.reserveAppend(allocator, 8, max_application_crypto_outstanding);
-    tx.appendReserved("ticket-1");
-    try tx.reserveAppend(allocator, 8, max_application_crypto_outstanding);
-    tx.appendReserved("ticket-2");
+    try tx.reserveAppend(allocator, 16, max_application_crypto_outstanding);
+    tx.appendReserved("ticket-1ticket-2");
 
     tx.markAcked(allocator, .{ .start = 0, .end = 8 });
     tx.markAcked(allocator, .{ .start = 0, .end = 8 });
@@ -2292,6 +2341,23 @@ test "application CRYPTO transmit clamps lost duplicate ranges below base" {
 
     const live = tx.liveRange(.{ .start = 0, .end = 16 }) orelse return error.TestUnexpectedResult;
     try testing.expectEqual(Range{ .start = 8, .end = 16 }, live);
+}
+
+test "application CRYPTO transmit allocates exact storage and frees after ACK" {
+    const allocator = testing.allocator;
+    var tx = CryptoTx{};
+    defer tx.deinit(allocator);
+
+    try tx.reserveAppend(allocator, 12, max_application_crypto_outstanding);
+    tx.appendReserved("small-ticket");
+    try testing.expectEqual(@as(usize, 12), tx.data.capacity);
+    try testing.expect(tx.data.capacity < max_application_crypto_outstanding);
+
+    tx.markAcked(allocator, .{ .start = 0, .end = 12 });
+    try testing.expectEqual(@as(usize, 0), tx.data.items.len);
+    try testing.expectEqual(@as(usize, 0), tx.data.capacity);
+    try testing.expect(tx.pending.isEmpty());
+    try testing.expect(tx.acked.isEmpty());
 }
 
 const TestPair = struct {
@@ -2628,6 +2694,28 @@ test "driver: application CRYPTO ticket budget refuses one-over before queueing"
         .issued_at_unix_ms = 10,
     }, tls_core.session.Limits.default));
     try testing.expectEqual(max_application_crypto_outstanding, pair.server.crypto_tx[2].data.items.len);
+}
+
+test "driver: ticket model limit refusal happens before CRYPTO allocation" {
+    const allocator = testing.allocator;
+    var pair = try TestPair.init(allocator);
+    defer pair.deinit(allocator);
+    try pair.pump();
+
+    const too_large = try allocator.alloc(u8, tls_core.session.Limits.default.max_ticket_len + 1);
+    defer allocator.free(too_large);
+    @memset(too_large, 0x5a);
+
+    try testing.expectError(error.TicketTooLarge, pair.server.emitNewSessionTicket(.{
+        .ticket_lifetime = 60,
+        .ticket_age_add = 1,
+        .ticket_nonce = "\x01",
+        .opaque_ticket = too_large,
+        .issued_at_unix_ms = 10,
+    }, tls_core.session.Limits.default));
+    try testing.expectEqual(@as(usize, 0), pair.server.crypto_tx[2].data.items.len);
+    try testing.expectEqual(@as(usize, 0), pair.server.crypto_tx[2].data.capacity);
+    try testing.expect(pair.server.crypto_tx[2].pending.isEmpty());
 }
 
 test "driver: local close reaches the peer and both sides drain" {
--- a/src/quic/connection.zig
+++ b/src/quic/connection.zig
@@ -21,6 +21,7 @@
 //! adapter's AEAD seal.
 
 const std = @import("std");
+const crypto_secrets = @import("crypto_secrets");
 const varint = @import("quic_varint");
 const config = @import("config.zig");
 const packet = @import("packet.zig");
@@ -234,10 +235,43 @@ const RangeList = struct {
 const CryptoTx = struct {
     data: std.ArrayList(u8) = .empty,
     pending: RangeList = .{},
+    acked: RangeList = .{},
+    base: u64 = 0,
 
     fn deinit(self: *CryptoTx, allocator: std.mem.Allocator) void {
+        crypto_secrets.secureZero(self.data.items);
         self.data.deinit(allocator);
         self.pending.deinit(allocator);
+        self.acked.deinit(allocator);
+    }
+
+    fn bufferedEnd(self: *const CryptoTx) u64 {
+        return self.base + self.data.items.len;
+    }
+
+    fn slice(self: *const CryptoTx, range: Range) []const u8 {
+        const start: usize = @intCast(range.start - self.base);
+        const end: usize = @intCast(range.end - self.base);
+        return self.data.items[start..end];
+    }
+
+    fn markAcked(self: *CryptoTx, allocator: std.mem.Allocator, range: Range) void {
+        self.acked.insert(allocator, range) catch {};
+        self.compactAcked();
+    }
+
+    fn compactAcked(self: *CryptoTx) void {
+        if (self.acked.items.items.len == 0) return;
+        const first = self.acked.items.items[0];
+        if (first.start != self.base or first.end <= self.base) return;
+        const release_len: usize = @intCast(@min(first.end, self.bufferedEnd()) - self.base);
+        if (release_len == 0) return;
+        crypto_secrets.secureZero(self.data.items[0..release_len]);
+        const keep_len = self.data.items.len - release_len;
+        if (keep_len > 0) std.mem.copyForwards(u8, self.data.items[0..keep_len], self.data.items[release_len..]);
+        self.data.shrinkRetainingCapacity(keep_len);
+        self.base += release_len;
+        if (first.end <= self.base) _ = self.acked.items.orderedRemove(0);
     }
 };
 
@@ -364,7 +398,7 @@ pub const Connection = struct {
     known_streams: std.AutoHashMap(StreamId, void),
     accept_queue: std.ArrayList(StreamId) = .empty,
 
-    crypto_tx: [2]CryptoTx = .{ .{}, .{} },
+    crypto_tx: [3]CryptoTx = .{ .{}, .{}, .{} },
     sent_records: std.ArrayList(SentRecord) = .empty,
 
     next_pn: [3]u64 = .{ 0, 0, 0 },
@@ -496,6 +530,7 @@ pub const Connection = struct {
 
     fn deinitPartial(self: *Connection) void {
         self.handshake.deinit();
+        self.adapter.deinit();
         if (self.streams) |*manager| manager.deinit();
         var it = self.send_queues.valueIterator();
         while (it.next()) |queue| {
@@ -959,6 +994,9 @@ pub const Connection = struct {
 
     fn onRecordAcked(self: *Connection, record: SentRecord) void {
         if (record.carried_handshake_done) self.handshake_done_acked = true;
+        if (record.crypto) |range| {
+            if (record.space == .application) self.crypto_tx[2].markAcked(self.allocator, range);
+        }
         for (record.streams[0..record.stream_count]) |sr| {
             if (self.send_queues.get(sr.id)) |queue| {
                 queue.acked.insert(self.allocator, sr.range) catch {};
@@ -1006,7 +1044,7 @@ pub const Connection = struct {
             const tx = switch (record.space) {
                 .initial => &self.crypto_tx[0],
                 .handshake => &self.crypto_tx[1],
-                .application => return,
+                .application => &self.crypto_tx[2],
             };
             // If the level's keys are already discarded the buffer is gone.
             if (tx.data.items.len > 0) {
@@ -1123,11 +1161,11 @@ pub const Connection = struct {
 
     /// Drain queued TLS output into the per-level retransmission buffers.
     fn collectCryptoOutput(self: *Connection) !void {
-        inline for (.{ EncryptionLevel.initial, EncryptionLevel.handshake }, 0..) |level, tx_index| {
+        inline for (.{ EncryptionLevel.initial, EncryptionLevel.handshake, EncryptionLevel.application }, 0..) |level, tx_index| {
             var chunk: [2048]u8 = undefined;
             while (self.handshake.pollOutput(level, &chunk) catch null) |output| {
                 const tx = &self.crypto_tx[tx_index];
-                std.debug.assert(output.offset == tx.data.items.len);
+                std.debug.assert(output.offset == tx.bufferedEnd());
                 try tx.data.appendSlice(self.allocator, output.bytes);
                 try tx.pending.insert(self.allocator, .{
                     .start = output.offset,
@@ -1163,6 +1201,33 @@ pub const Connection = struct {
             return;
         };
         self.afterHandshakeProgress(now_us);
+    }
+
+    pub fn emitNewSessionTicket(
+        self: *Connection,
+        params: tls_handshake.EmitNewSessionTicketParams,
+        limits: tls_core.session.Limits,
+    ) tls_handshake.HandshakeError!tls_core.session.ServerRecoverableState {
+        if (self.state_ != .established or self.role != .server) return error.InvalidHandshakeState;
+        var sink = tls_handshake.EventSink{};
+        defer sink.deinit();
+        var ticket_state = try self.tls.emitNewSessionTicket(self.allocator, &sink, params, limits);
+        errdefer ticket_state.deinit();
+        try self.queueTlsEvents(&sink);
+        self.collectCryptoOutput() catch return error.HandshakeBufferOverflow;
+        return ticket_state;
+    }
+
+    fn queueTlsEvents(self: *Connection, sink: *tls_handshake.EventSink) tls_handshake.HandshakeError!void {
+        for (sink.items[0..sink.len]) |event| {
+            switch (event) {
+                .handshake_bytes => |hb| self.adapter.queueHandshakeOutput(hb.epoch, hb.data) catch |err| return switch (err) {
+                    error.InvalidCryptoLevel => error.UnexpectedCryptoLevel,
+                    error.CryptoBufferTooLarge => error.HandshakeBufferOverflow,
+                },
+                else => {},
+            }
+        }
     }
 
     fn afterHandshakeProgress(self: *Connection, now_us: u64) void {
@@ -1682,7 +1747,7 @@ pub const Connection = struct {
         const has_crypto = switch (space) {
             .initial => !self.crypto_tx[0].pending.isEmpty(),
             .handshake => !self.crypto_tx[1].pending.isEmpty(),
-            .application => false,
+            .application => !self.crypto_tx[2].pending.isEmpty(),
         };
         const has_app = space == .application and self.hasAppContent();
         if (!want_ack and !has_crypto and !(has_app and can_send_data) and !probe) return null;
@@ -1741,12 +1806,12 @@ pub const Connection = struct {
             const tx = switch (space) {
                 .initial => &self.crypto_tx[0],
                 .handshake => &self.crypto_tx[1],
-                .application => unreachable,
+                .application => &self.crypto_tx[2],
             };
             while (!tx.pending.isEmpty() and plain_len + frame.max_crypto_overhead + 16 < plain_budget) {
                 const room: u64 = @intCast(plain_budget - plain_len - frame.max_crypto_overhead);
                 const range = tx.pending.takeFirst(room) orelse break;
-                const data = tx.data.items[@intCast(range.start)..@intCast(range.end)];
+                const data = tx.slice(range);
                 const n = frame.encodeCrypto(range.start, data, plain[plain_len..plain_budget]) catch {
                     tx.pending.insert(self.allocator, range) catch {};
                     break;
@@ -2113,11 +2178,13 @@ const TestPair = struct {
     fn init(allocator: std.mem.Allocator) !*TestPair {
         const pair = try allocator.create(TestPair);
         pair.* = .{
-            .client_backend = tls_backend_mod.Tls13Backend.initClient(
+            .client_backend = try tls_backend_mod.Tls13Backend.initClientWithAllocator(
+                allocator,
                 .{ .hello_random = [_]u8{0xc1} ** 32, .key_share_seed = [_]u8{0x11} ** 32 },
                 .{ .pinned_certificate = tls_backend_mod.testdata.certificate_der },
             ),
-            .server_backend = tls_backend_mod.Tls13Backend.initServer(
+            .server_backend = tls_backend_mod.Tls13Backend.initServerWithAllocator(
+                allocator,
                 .{ .hello_random = [_]u8{0x51} ** 32, .key_share_seed = [_]u8{0x22} ** 32 },
                 try tls_backend_mod.Identity.initPkcs8(
                     tls_backend_mod.testdata.certificate_der,
@@ -2146,6 +2213,49 @@ const TestPair = struct {
         return pair;
     }
 
+    fn initWithTicketConsumer(
+        allocator: std.mem.Allocator,
+        limits: tls_core.session.Limits,
+        consumer: tls_core.tls13_backend.Tls13Backend.SessionTicketConsumer,
+    ) !*TestPair {
+        const pair = try allocator.create(TestPair);
+        pair.* = .{
+            .client_backend = try tls_backend_mod.Tls13Backend.initClientWithAllocator(
+                allocator,
+                .{ .hello_random = [_]u8{0xc1} ** 32, .key_share_seed = [_]u8{0x11} ** 32 },
+                .{ .pinned_certificate = tls_backend_mod.testdata.certificate_der },
+            ),
+            .server_backend = tls_backend_mod.Tls13Backend.initServerWithAllocator(
+                allocator,
+                .{ .hello_random = [_]u8{0x51} ** 32, .key_share_seed = [_]u8{0x22} ** 32 },
+                try tls_backend_mod.Identity.initPkcs8(
+                    tls_backend_mod.testdata.certificate_der,
+                    tls_backend_mod.testdata.private_key_pkcs8_der,
+                ),
+            ),
+        };
+        errdefer allocator.destroy(pair);
+        try pair.client_backend.engine.setSessionTicketConsumer(allocator, limits, consumer);
+        pair.client = try Connection.init(allocator, .{
+            .role = .client,
+            .local_cid = &client_cid,
+            .original_dcid = &odcid,
+            .peer_cid = &odcid,
+            .tls = pair.client_backend.backend(),
+            .now_us = pair.now_us,
+        });
+        errdefer pair.client.deinit();
+        pair.server = try Connection.init(allocator, .{
+            .role = .server,
+            .local_cid = &odcid,
+            .original_dcid = &odcid,
+            .peer_cid = &client_cid,
+            .tls = pair.server_backend.backend(),
+            .now_us = pair.now_us,
+        });
+        return pair;
+    }
+
     fn initClientAuth(
         allocator: std.mem.Allocator,
         client_provider: tls_core.credentials.CredentialProvider,
@@ -2153,11 +2263,13 @@ const TestPair = struct {
     ) !*TestPair {
         const pair = try allocator.create(TestPair);
         pair.* = .{
-            .client_backend = tls_backend_mod.Tls13Backend.initClient(
+            .client_backend = try tls_backend_mod.Tls13Backend.initClientWithAllocator(
+                allocator,
                 .{ .hello_random = [_]u8{0xc1} ** 32, .key_share_seed = [_]u8{0x11} ** 32 },
                 .{ .pinned_certificate = tls_backend_mod.testdata.certificate_der },
             ),
-            .server_backend = tls_backend_mod.Tls13Backend.initServer(
+            .server_backend = tls_backend_mod.Tls13Backend.initServerWithAllocator(
+                allocator,
                 .{ .hello_random = [_]u8{0x51} ** 32, .key_share_seed = [_]u8{0x22} ** 32 },
                 try tls_backend_mod.Identity.initPkcs8(
                     tls_backend_mod.testdata.certificate_der,
@@ -2223,7 +2335,8 @@ test "Connection.init failure before the handshake is assigned does not deinit u
     // was assigned -- deinitPartial()'s errdefer would then call
     // .deinit() on undefined storage. This must fail cleanly instead.
     const too_short_dcid = [_]u8{0xaa} ** (tls_adapter.min_initial_dcid_len - 1);
-    var backend = tls_backend_mod.Tls13Backend.initClient(
+    var backend = try tls_backend_mod.Tls13Backend.initClientWithAllocator(
+        allocator,
         .{ .hello_random = [_]u8{0xc1} ** 32, .key_share_seed = [_]u8{0x11} ** 32 },
         .{ .pinned_certificate = tls_backend_mod.testdata.certificate_der },
     );
@@ -2281,6 +2394,62 @@ test "driver: bidirectional stream data round-trips with FIN" {
 
     try testing.expectEqual(quic_stream.StreamState.closed, pair.client.streamState(id).?);
     try testing.expectEqual(quic_stream.StreamState.closed, pair.server.streamState(id).?);
+}
+
+test "driver: server NewSessionTicket uses application CRYPTO retransmission and stream traffic continues" {
+    const Capture = struct {
+        count: usize = 0,
+        ticket_len: usize = 0,
+
+        fn now(_: *anyopaque) i64 {
+            return 10;
+        }
+
+        fn onTicket(ctx: *anyopaque, ticket: *const tls_core.session.ClientTicketState) void {
+            const self: *@This() = @ptrCast(@alignCast(ctx));
+            self.count += 1;
+            self.ticket_len = ticket.ticket.slice().len;
+        }
+    };
+
+    const allocator = testing.allocator;
+    var capture = Capture{};
+    var pair = try TestPair.initWithTicketConsumer(allocator, tls_core.session.Limits.default, .{
+        .ctx = &capture,
+        .nowUnixMsFn = Capture.now,
+        .onTicketFn = Capture.onTicket,
+    });
+    defer pair.deinit(allocator);
+    try pair.pump();
+
+    var server_state = try pair.server.emitNewSessionTicket(.{
+        .ticket_lifetime = 60,
+        .ticket_age_add = 1,
+        .ticket_nonce = "\x01",
+        .opaque_ticket = "connection-ticket",
+        .issued_at_unix_ms = 10,
+    }, tls_core.session.Limits.default);
+    defer server_state.deinit();
+
+    var datagram: [2048]u8 = undefined;
+    const dropped = pair.server.pollTransmit(&datagram, pair.now_us) orelse return error.TestExpectedEqual;
+    try testing.expect(dropped.len > 0);
+    const sent = pair.server.sent_records.items[pair.server.sent_records.items.len - 1];
+    try testing.expectEqual(PacketNumberSpace.application, sent.space);
+    try testing.expect(sent.crypto != null);
+    pair.server.requeueRecord(sent);
+
+    try pair.pump();
+    try testing.expectEqual(@as(usize, 1), capture.count);
+    try testing.expectEqual(@as(usize, "connection-ticket".len), capture.ticket_len);
+
+    const id = try pair.client.openStream(.bidi);
+    try testing.expectEqual(@as(usize, 5), try pair.client.writeStream(id, "after", true));
+    try pair.pump();
+    try testing.expectEqual(@as(?StreamId, id), pair.server.acceptStream());
+    var buf: [16]u8 = undefined;
+    const request = try pair.server.readStream(id, &buf);
+    try testing.expectEqualStrings("after", buf[0..request.len]);
 }
 
 test "driver: local close reaches the peer and both sides drain" {
@@ -2483,11 +2652,13 @@ test "a real Connection closes with handshake_failure when the server has no app
 
         const pair = try allocator.create(TestPair);
         pair.* = .{
-            .client_backend = tls_backend_mod.Tls13Backend.initClient(
+            .client_backend = try tls_backend_mod.Tls13Backend.initClientWithAllocator(
+                allocator,
                 .{ .hello_random = [_]u8{0xc1} ** 32, .key_share_seed = [_]u8{0x11} ** 32 },
                 .{ .pinned_certificate = tls_backend_mod.testdata.certificate_der },
             ),
-            .server_backend = tls_backend_mod.Tls13Backend.initServer(
+            .server_backend = tls_backend_mod.Tls13Backend.initServerWithAllocator(
+                allocator,
                 .{ .hello_random = [_]u8{0x51} ** 32, .key_share_seed = [_]u8{0x22} ** 32 },
                 try tls_backend_mod.Identity.initPkcs8(tls_backend_mod.testdata.certificate_der, tls_backend_mod.testdata.private_key_pkcs8_der),
             ),
--- a/src/quic/connection.zig
+++ b/src/quic/connection.zig
@@ -341,20 +341,75 @@ const CryptoTx = struct {
     fn reserveAppend(self: *CryptoTx, allocator: std.mem.Allocator, bytes_len: usize, max_outstanding: usize) !void {
         if (bytes_len > max_outstanding) return error.CryptoBufferTooLarge;
         if (self.data.items.len > max_outstanding - bytes_len) return error.CryptoBufferTooLarge;
-        if (self.data.capacity == 0) {
-            try self.data.ensureTotalCapacityPrecise(allocator, bytes_len);
-        } else if (self.data.capacity < self.data.items.len + bytes_len) {
-            return error.CryptoBufferTooLarge;
+
+        const new_total = self.data.items.len + bytes_len;
+        const range_budget = applicationCryptoRangeBudget(new_total);
+        const pending_capacity = self.pending.items.items.len + range_budget + 1;
+        const acked_capacity = self.acked.items.items.len + range_budget;
+
+        var new_data: ?[]u8 = null;
+        errdefer if (new_data) |buf| {
+            crypto_secrets.secureZero(buf);
+            allocator.free(buf);
+        };
+        if (self.data.capacity < new_total) {
+            const replacement = try allocator.alloc(u8, new_total);
+            @memcpy(replacement[0..self.data.items.len], self.data.items);
+            new_data = replacement;
         }
-        const range_budget = applicationCryptoRangeBudget(bytes_len);
-        try self.pending.ensureUnusedCapacity(allocator, range_budget + 1);
-        try self.acked.ensureUnusedCapacity(allocator, range_budget);
+
+        var new_pending: ?[]Range = null;
+        errdefer if (new_pending) |buf| allocator.free(buf);
+        if (self.pending.items.capacity < pending_capacity) {
+            const replacement = try allocator.alloc(Range, pending_capacity);
+            @memcpy(replacement[0..self.pending.items.items.len], self.pending.items.items);
+            new_pending = replacement;
+        }
+
+        var new_acked: ?[]Range = null;
+        errdefer if (new_acked) |buf| allocator.free(buf);
+        if (self.acked.items.capacity < acked_capacity) {
+            const replacement = try allocator.alloc(Range, acked_capacity);
+            @memcpy(replacement[0..self.acked.items.items.len], self.acked.items.items);
+            new_acked = replacement;
+        }
+
+        if (new_pending) |replacement| {
+            self.replaceRangeCapacity(allocator, &self.pending, replacement);
+            new_pending = null;
+        }
+        if (new_acked) |replacement| {
+            self.replaceRangeCapacity(allocator, &self.acked, replacement);
+            new_acked = null;
+        }
+        if (new_data) |replacement| {
+            self.replaceDataCapacity(allocator, replacement);
+            new_data = null;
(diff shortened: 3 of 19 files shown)


Jaruso (commented):
**Review result: changes required before merge.** The overall architecture is pointed in the right direction, and the existing CI matrix is green. Several protocol, boundedness, secret-lifecycle, and test requirements from #361 are not yet satisfied.

## 1. Blocker: the application-epoch reassembler cannot accept the legal RFC wire range

`max_new_session_ticket_message_len` currently budgets only eight bytes for the extension vector:

```zig
pub const max_new_session_ticket_message_len =
    session.absolute_ticket_wire_max + 4 + 4 + 4 + 1 +
    session.max_ticket_nonce_len + 2 + 2 + 8;
```

A legal framed `NewSessionTicket` may contain:

```text
handshake header       4
lifetime               4
age_add                 4
nonce vector        1 + 255
ticket vector       2 + 65535
extensions vector   2 + 65534
--------------------------------
maximum total       131341 bytes
```

The current bound is only 65,815 bytes. A legal ticket with a large opaque identity and more than eight bytes of well-formed unknown extensions therefore fails in `application_input.append()` before `new_session_ticket.decode()` can apply the required “ignore unknown well-formed extensions” behavior.

Simply changing the inline array to 131,341 bytes would add roughly 123 KiB to every `Tls13Backend`, so I would not solve this by embedding the complete RFC maximum in every connection. Please introduce a dedicated bounded post-handshake reassembly strategy that can grow only for a ticket being received, or otherwise stream/copy the framed message into caller-provided bounded storage. It must still allow the full configured/RFC range and fail deterministically when the configured limit is exceeded.

Add boundary tests for:

- a 65,535-byte ticket with no extensions;
- a legal unknown extension vector larger than eight bytes;
- the exact maximum legal framed message split across several application-epoch receives;
- one byte beyond the accepted configured/RFC bound.

## 2. Blocker: server issuance is capped by the 16 KiB `EventSink`, and failure mutates the transcript

`emitNewSessionTicket()` builds a potentially ~65 KiB message and calls `sink.emitCrypto()`, but the shared TLS transport contract has only 16 KiB of byte scratch. Any legal ticket message larger than the remaining scratch capacity necessarily fails with `TransportBufferOverflow`, despite passing `encodedLen()` and the backend's own maximum check.

More importantly, the code updates the transcript before knowing whether the output was accepted:

```zig
self.core.transcript.update(message);
try sink.emitCrypto(.application, message);
```

If the sink rejects the message, the peer never receives it but the server transcript has already advanced. Retrying issuance then starts from a transcript state containing an unsent ticket.

The output operation must be atomic from the backend's perspective. At minimum, transcript mutation must occur only after successful output ownership/copy:

```zig
try sink.emitCrypto(.application, message);
self.core.transcript.update(message);
```

That ordering fix does not solve the 16 KiB capacity problem by itself. Please provide an output path capable of representing/fragmenting the configured legal ticket size without requiring one complete message to fit in the generic 16 KiB sink scratch.

Add tests proving:

- sink/output refusal leaves the transcript unchanged;
- issuance can be retried after refusal;
- tickets above 16 KiB can be emitted through the chosen bounded fragmentation/ownership mechanism;
- the receiver reconstructs the emitted fragments exactly once.

## 3. High: extension-vector validation rejects legal input and accepts one illegal wire length

`decode()` reads the extension-vector length as a `u16` but never enforces the RFC maximum of 65,534, even though `max_extensions_wire_len` declares that limit. A declared length of 65,535 can therefore be accepted when its contents parse cleanly.

Conversely, `ExtensionGuard` stores only 64 IDs. A ticket containing 65 unique, zero-length unknown extensions is small and otherwise legal, but the parser returns `MalformedHandshake`. This violates the issue requirement that unknown, well-formed extensions be ignored.

Please:

1. reject an extension-vector length greater than 65,534 with the documented malformed-input result;
2. remove the arbitrary 64-unique-extension semantic limit.

A simple bounded approach is to rescan the already-validated prefix for duplicate IDs while walking the at-most-64-KiB extension vector. That is O(n²), but remains tightly bounded and avoids allocating or rejecting legal extension counts. A full `u16` bitset is another option if the memory tradeoff is acceptable.

Tests should cover exact lengths 65,534 and 65,535, duplicate known and unknown types, and more than 64 unique unknown extensions.

## 4. High: server issuance ignores `Limits.max_ticket_len`

The server path validates `Limits`, but it only applies the absolute 65,535-byte wire cap. `buildServerRecoverableState()` does not retain the opaque ticket and consequently never checks `params.ticket.len` against `limits.max_ticket_len`.

With default limits, the server can issue a ticket larger than 4,096 bytes and return a valid `ServerRecoverableState`; a client using the same default limits then drops that ticket during `ClientTicketState.init()`. The server has emitted state that its corresponding configured client model cannot retain.

Please validate both the wire parameters and the caller-tightened model limit before deriving state or emitting bytes:

```zig
try limits.validate();
try validateEmitParams(params);
if (params.ticket.len > limits.max_ticket_len)
    return error.TicketTooLarge; // mapped deterministically by the backend
```

`buildServerRecoverableState()` is public, so it should either perform this validation itself or be made private and callable only after a validated constructor/preflight. Its error set should preserve a useful bounded-error distinction rather than mapping every construction failure to `InvalidTransportProfile`.

Add exact default-limit, custom-limit, absolute-maximum, and one-over tests.

## 5. High: SHA-384 RMS derivation required by #361 is not implemented

`resumptionPsk()` correctly supports SHA-256 and SHA-384, but `resumptionMasterSecret()` is still hardcoded to the SHA-256 `KeySchedule`: it requires 32-byte input/output and always uses `HkdfSha256`. The SHA-384 test begins from a synthetic 48-byte RMS, so it does not cover SHA-384 RMS derivation.

#361 explicitly requires both derivation paths even though the live negotiation path remains SHA-256-only. This can be done without pulling the full algorithm-negotiation refactor from #386 into this PR. For example, factor a hash-generic helper:

```zig
pub fn deriveResumptionMasterSecret(
    hash: provider.Hash,
    master_secret: []const u8,
    transcript_hash: []const u8,
    out: []u8,
) Error!void {
    const expected = hash.digestLength();
    if (master_secret.len != expected or
        transcript_hash.len != expected or
        out.len != expected)
        return error.InvalidSecretLength;

    switch (hash) {
        .sha256 => { /* HKDF-SHA256 */ },
        .sha384 => { /* HKDF-SHA384 */ },
    }
}
```

The current concrete `KeySchedule.resumptionMasterSecret()` may remain a SHA-256 wrapper around that helper. Add an independently generated exact SHA-384 RMS vector, not only a SHA-384 per-ticket expansion vector.

## 6. High: temporary opaque-ticket copies are not cleared after successful processing

`emitNewSessionTicket()` copies the opaque ticket into a large stack buffer and never wipes that buffer. Incoming application reassembly similarly discards a parsed message by shifting bytes and reducing `len`, but does not clear the vacated tail. The ticket identity is a bearer credential under #360/#361 and should not remain in scratch storage until backend teardown.

Please ensure all temporary copies are cleared on success and error:

```zig
var buf: [/* bounded size */]u8 = undefined;
defer crypto.secureZero(u8, &buf);
```

For receive reassembly, either make `Reassembler.discard()` zero the old occupied tail after shifting or add a secure-discard variant used by the post-handshake ticket buffer:

```zig
const old_len = self.len;
std.mem.copyForwards(u8, self.data[0 .. old_len - count], self.data[count..old_len]);
self.len = old_len - count;
crypto.secureZero(u8, self.data[self.len..old_len]);
```

Add assertions that the RMS, temporary encode buffer, consumed application-input range, PSK scratch, and partial owned state are wiped after success and injected failures.

## 7. High: the required backend and integration coverage is absent

This PR changes only the key schedule, codec, root export, and backend. It adds several codec tests inline, but no direct lifecycle tests in `tls13_backend_tests.zig` and no record-mode integration test. The green existing CI therefore does not exercise most of the new behavior.

Please add the #361-required coverage, including at minimum:

- RMS captured once at the transcript ending with client `Finished`;
- ticket before completion and at the wrong epoch;
- one fragmented ticket delivered exactly once;
- several complete/fragmented tickets delivered in order;
- distinct nonces produce distinct delivered PSKs;
- lifetime-zero parsed but not delivered;
- no consumer configured parses and drops safely;
- callback retention requires an explicit clone and survives callback return;
- callback/local storage refusal does not fail the TLS connection;
- server emission rejected before completion and from a client;
- server-returned state and client-ingested state contain matching PSKs/metadata;
- application data can proceed before issuance;
- output failure is atomic;
- allocation-failure sweeps for every owned construction path;
- teardown/error paths clear RMS and temporary ticket state;
- one full record-mode handshake → server ticket emission → client callback integration path;
- QUIC/application-epoch regression coverage for fragmented ticket delivery.

The codec test matrix should also include truncation at every fixture byte, nonce lengths 0/255/256, ticket lengths 0/configured max/65,535/one-over where representable, mutated vector lengths, malformed/trailing extension data, exact undersized output behavior, and the extension-count/length cases above.

Once these issues are addressed, the architecture should be in good shape for #362 to consume without inheriting hidden wire-limit or state-divergence behavior.

Jaruso (commented):
**Re-review result: changes are still required before merge.**

The latest commit correctly fixes several findings from the first review:

- the extension-vector `65534`/`65535` distinction is now enforced;
- duplicate detection no longer has an arbitrary 64-extension correctness cap;
- server state applies `Limits.max_ticket_len`;
- SHA-384 RMS derivation has an exact fixture;
- reassembler tails and the local encode buffer are wiped;
- transcript mutation now occurs after the inner sink accepts the event;
- direct tests cover retry after inner-sink refusal and a fragmented large ticket.

The remaining issues are below.

## 1. Blocker: the full-range fix adds roughly 256 KiB of ticket scratch to every QUIC TLS backend

The implementation resolves the previous undersized buffer by making the RFC maximum a compile-time array size everywhere:

```zig
pub const max_new_session_ticket_message_len = 131341;

application_input:
    Reassembler(max_new_session_ticket_message_len) = .{};

pub const Contract = transport.ContractWithOptions(
    ...,
    max_new_session_ticket_message_len,
    ...,
);
```

The QUIC wrapper then embeds both values:

```zig
pub const Tls13Backend = struct {
    engine: shared.Tls13Backend, // includes the 131341-byte application_input
    ...
    scratch: RecordSink = .{},   // includes another 131341-byte scratch array
};
```

That is at least **262,682 bytes of ticket buffering in every QUIC backend**, before its normal TLS/QUIC state. Compared with the previous 8 KiB application reassembler and 16 KiB sink, this adds roughly 232 KiB per connection. Ten thousand concurrent QUIC connections would consume more than 2.2 GiB solely for dormant ticket scratch.

`emitNewSessionTicket()` also creates another `131341`-byte local array on the call stack even when emitting a normal small ticket:

```zig
var buf: [max_new_session_ticket_message_len]u8 = undefined;
```

This is not an acceptable production representation of an optional post-handshake message. It also expands the application-epoch buffering available to non-ticket or deliberately oversized post-handshake frames.

Please replace the fixed maximum arrays with a lazy bounded design. A suitable shape would be:

- retain only a small fixed header/prefix staging buffer in every connection;
- once a complete handshake header and declared ticket body length are known, allocate or borrow exactly that bounded length;
- cap that allocation at the RFC/configured limit before retaining attacker-controlled bytes;
- wipe and free it immediately after dispatch or on every error/deinit path;
- encode server output into exact-size allocator-backed storage rather than a maximum-size stack array.

The allocator is already explicit in the server issuance API. The client receive path may need an explicit backend/post-handshake allocator or caller-provided bounded workspace rather than embedding the RFC maximum.

Add a size/regression assertion so this does not silently return, for example an explicit bound on `@sizeOf(Tls13Backend)` and the QUIC wrapper.

## 2. Blocker: tickets above 16 KiB still cannot traverse either real transport path

The new test `large emitted ticket is delivered once after fragmented application receives` passes the inner event payload directly from one shared backend into another. It bypasses both actual carrier adapters.

### QUIC remains capped at 16 KiB

`src/quic/tls_handshake.zig` still defines its transport sink with only `16 * 1024` bytes. `src/quic/tls_backend.zig::translate()` forwards the entire inner `handshake_bytes` event in one call:

```zig
.handshake_bytes => |item|
    try destination.emitHandshakeBytes(toLevel(item.epoch), item.data),
```

Therefore any inner ticket event larger than 16 KiB fails during translation with `HandshakeBufferOverflow`. The inner transcript has already advanced because the inner `RecordSink` accepted the event, but the QUIC-facing sink did not.

### Record mode still attempts one oversized TLS record

`record_epoch_bridge.applyEvent()` forwards the complete event to `sealHandshake()`, which forwards the complete slice to `WriteState.seal()`. TLS record protection rejects plaintext that does not fit one record with `RecordTooLarge`. A 20 KiB or 65 KiB ticket therefore cannot be carried as record-mode TLS even though the shared backend accepted and hashed it.

The output path needs to support a logical handshake message split across transport chunks:

- record mode: fragment the handshake bytes across as many application-epoch TLS records as required;
- QUIC: append the handshake bytes to the application CRYPTO stream in bounded chunks;
- preserve one logical transcript update and deterministic retry/abort behavior;
- do not require the entire logical message to reside in every generic event sink.

The final tests must go through the actual adapters, not directly between shared backends:

1. full record-mode handshake → emit a ticket larger than one TLS record → seal multiple records → client callback exactly once;
2. full QUIC handshake → emit a ticket larger than the QUIC driver sink chunk → CRYPTO-stream delivery → client callback exactly once;
3. an adapter/output refusal before ownership transfer leaves the transcript unchanged, or deterministically aborts without exposing a retryable-but-diverged state.

## 3. High: the public server-state builder still accepts an impossible nonce length

`buildServerRecoverableState()` is public, but unlike `emitNewSessionTicket()` it does not call the wire-parameter validator. It validates the ticket length and lifetime through the model, then passes an arbitrary-length nonce directly into HKDF:

```zig
try KeySchedule.resumptionPsk(
    hash,
    resumption_master_secret,
    params.ticket_nonce,
    ...,
);
```

A direct caller can therefore construct a valid `ServerRecoverableState` using a 256-byte or much larger nonce even though `ticket_nonce` is a one-byte-length TLS vector and can never be emitted or ingested.

Either make this helper private, or validate all `EmitParams` itself before derivation. For example:

```zig
pub fn buildServerRecoverableState(...) BuildServerError!ServerRecoverableState {
    try limits.validate();
    validateEmitParams(params) catch |err| switch (err) {
        error.IllegalParameter => return error.IllegalParameter,
        else => return error.InvalidLimits,
    };
    if (params.ticket.len > limits.max_ticket_len)
        return error.TicketTooLarge;
    ...
}
```

Add direct helper tests for nonce lengths `255` and `256`; testing only the encoder does not protect this public constructor.

## 4. High: the acceptance tests still bypass the real handshake lifecycle

The newly added backend tests manually assign:

```zig
backend.core.handshake_lifecycle = .complete;
backend.resumption_master_secret.replace(...);
```

That does not prove the most important lifecycle requirement: that both roles capture the same RMS exactly once from the transcript ending with the real client `Finished`.

No changed test file adds the required full record-mode handshake → ticket → callback path, and there is no QUIC-wrapper ticket test. The following #361 cases also remain absent or only partially covered:

- real client/server RMS capture from an actual completed handshake;
- multiple complete and fragmented tickets delivered in order;
- distinct ticket nonces producing distinct delivered PSKs at the backend seam;
- no-consumer parse-and-drop behavior;
- callback retention through an explicit `cloneInto` after callback return;
- allocator failure sweeps for client/common ticket construction;
- local cache/storage refusal remaining non-fatal;
- RMS clearing after successful and failed real handshakes;
- ordinary application data before and between ticket issuance;
- transport-level record and QUIC delivery as described above.

Please extend the existing `DirectHarness`/record integration machinery and QUIC real-handshake harness rather than treating manually forced `.complete` state as acceptance coverage.

## 5. High: duplicate-extension detection is now attacker-controlled quadratic work

The parser now rescans every prior extension for each new extension:

```zig
if (try extensionSeen(extensions[0..prefix_len], ext_type))
```

A legal 65,534-byte vector can hold 16,383 zero-length extensions. This implementation performs approximately **134 million previous-extension scans for one ticket**, and a peer may send repeated post-handshake tickets after the connection is established.

Use a fixed 65,536-bit seen set instead: it is exactly 8 KiB, has deterministic O(n) behavior, requires no allocation, and supports every `u16` extension ID without rejecting legal counts.

```zig
var seen: [std.math.maxInt(u16) / 8 + 1]u8 = [_]u8{0} ** 8192;

fn markSeen(seen: *[8192]u8, id: u16) bool {
    const index: usize = id / 8;
    const mask: u8 = @as(u8, 1) << @intCast(id & 7);
    const duplicate = (seen[index] & mask) != 0;
    seen[index] |= mask;
    return duplicate;
}
```

This still keeps parsing bounded while avoiding an inexpensive post-handshake CPU-exhaustion input.

## 6. Medium: the client path silently swallows every state-construction invariant failure

`onNewSessionTicket()` currently does:

```zig
var state = buildClientTicketState(...) catch return;
```

Best-effort cache/storage failures must not tear down the TLS connection, but that does not justify swallowing cryptographic or internal-invariant failures such as `InvalidSecretLength`, `InvalidPskLength`, invalid connection metadata, or an invalid prevalidated limit.

Handle only expected local retention failures as a drop, and map impossible/invariant failures explicitly:

```zig
const state = buildClientTicketState(...) catch |err| switch (err) {
    error.OutOfMemory,
    error.TicketTooLarge,
    error.CompatSnapshotTooLarge,
    => return, // best-effort local retention failure

    error.InvalidSecretLength,
    error.InvalidPskLength,
    => return error.SecretExportFailed,

    else => return error.InvalidTransportProfile,
};
```

The exact grouping can follow the final error vocabulary, but a corrupted RMS or impossible completed-connection context should not look identical to a cache refusing an otherwise valid ticket.

CI is green, and the protocol-codec fixes are materially improved, but the transport and memory architecture still prevent this from satisfying #361 safely at production scale.

Jaruso (commented):
**Review result: changes are still required before merge.**

This update correctly addresses several items from the previous pass:

- post-handshake receive storage is now allocated lazily instead of embedding the RFC maximum in every backend;
- duplicate-extension tracking is bounded and linear via a fixed bitset;
- `buildServerRecoverableState()` now validates nonce/wire bounds;
- client construction errors distinguish best-effort storage refusal from TLS invariant failures;
- the temporary server encoding buffer is exact-sized and wiped;
- real completed record and QUIC handshakes are now used by the new 20 KiB ticket tests.

CI is green, but the following production-path issues remain.

## 1. Blocker: the default no-consumer path cannot receive a ticket

`PostHandshakeInput.append()` requires an allocator after reading the four-byte handshake header:

```zig
const allocator = self.allocator orelse return error.InvalidHandshakeState;
self.buf = allocator.alloc(u8, frame_len) catch
    return error.CredentialProviderFailed;
```

However, ordinary `Tls13Backend.initClient*()` construction leaves that allocator null. It is only populated by either:

```zig
setSessionTicketConsumer(...)
setPostHandshakeAllocator(...)
```

The production QUIC wrapper constructors do not call either method, and there is no corresponding automatic record-mode wiring. Consequently, a normal completed client with no ticket consumer configured rejects the first valid `NewSessionTicket` with `InvalidHandshakeState` as soon as the full header arrives.

The new test named `client parses and drops NewSessionTicket when no consumer is configured` masks this by explicitly calling:

```zig
try backend.setPostHandshakeAllocator(std.testing.allocator);
```

That does not exercise the public default behavior required by #361: no consumer configured must still parse and validate the ticket, then drop it without failing the connection.

Please make post-handshake allocation an explicit, automatically wired part of backend/driver construction rather than an optional setter the transport owner has to discover. At minimum, add production-wrapper tests that construct record and QUIC clients normally, configure no consumer and no special test-only allocator, complete a real handshake, receive a valid ticket, and remain usable.

## 2. Blocker: changing the post-handshake allocator mid-frame can free through the wrong allocator

`PostHandshakeInput` stores one mutable `allocator` field and later uses its current value to free `buf`. `setAllocator()` simply replaces that field:

```zig
fn setAllocator(self: *PostHandshakeInput, allocator: std.mem.Allocator) void {
    self.allocator = allocator;
}
```

Both public configuration methods may call this at any time. This creates a concrete allocator-ownership bug:

1. configure allocator A;
2. receive a ticket header/body prefix, allocating `buf` from A;
3. call `setPostHandshakeAllocator(B)` or reconfigure the consumer with B;
4. complete/discard/deinitialize the message;
5. `B.free(self.buf)` is called for memory allocated by A.

Please make the allocator immutable after configuration, reject replacement while a frame allocation is active, or retain the actual allocation owner alongside `buf`. Add a test with two independently tracking allocators and a reconfiguration attempt between ticket fragments.

Also consider restricting both setters to the pre-start lifecycle. Mutating callback/allocator ownership during an established callback or partially received frame should not be legal unless there is a clearly transactional replacement contract.

## 3. Blocker: server issuance still rejects valid configured tickets above 32 KiB

The RFC/model path permits an opaque ticket up to 65,535 bytes, and `session.Limits.max_ticket_len` can be tightened anywhere up to that bound. `emitNewSessionTicket()` nevertheless contains:

```zig
const message_len = handshake_header_len + body_len;
if (message_len > EventSink.max_bytes)
    return error.TransportBufferOverflow;
```

`EventSink.max_bytes` is now 32 KiB. Therefore the public state builder can successfully construct server state for a 65,535-byte ticket, while the actual server issuance API rejects the corresponding legal ticket before either transport sees it. The new transport tests use a 20 KiB ticket and do not cross this remaining cap.

The shared engine output contract needs a chunked/streaming or owned-message representation that does not require one full post-handshake message to fit into fixed event scratch. Alternatively, if Tardigrade intentionally chooses a smaller supported issuance limit, that limit must be explicit in the API and applied coherently by `Limits`, the client receive path, the server state builder, and both transport adapters. The current state advertises the RFC/model maximum but silently imposes a different server-emission maximum.

Add exact boundary coverage for the chosen contract, including a 65,535-byte opaque ticket when the caller opts into that limit.

## 4. Blocker: the production record stream still cannot serialize the new multi-record result

`record_epoch_bridge.Bridge.sealHandshake()` can now return several concatenated TLS records when the handshake message exceeds one plaintext fragment. That helper is correct in isolation, but the production `PureZigRecordStream` still gives it only one-record scratch:

```zig
var record_buf: [record_codec.max_ciphertext_record_len]u8 = undefined;
const record = self.bridge.sealHandshake(hb.epoch, hb.data, &record_buf) ...;
```

The manual `applyEvent()` path has the same one-record buffer. A 20 KiB ticket therefore fails with `RecordBufferOverflow` in the actual stream, despite passing the new direct bridge test.

The new record ticket integration test bypasses `PureZigRecordStream`: it manually supplies `record_codec.max_ciphertext_record_len * 3` bytes to the bridge, parses the concatenated records, and feeds them directly to the driver. That proves the bridge helper, not the production TLS-over-TCP carrier.

Please change the production stream to seal and enqueue one record at a time under its existing queue/backpressure rules, rather than requiring a temporary buffer large enough for the complete ticket. The operation must remain retryable and must not advance the application write sequence for records that cannot be committed to the outbound queue.

Add a test through `PureZigRecordStream` or the socket harness that performs a real handshake, emits a ticket larger than one TLS plaintext fragment, pumps the real carrier, delivers exactly one callback, and then successfully exchanges application data.

## 5. Blocker: QUIC application CRYPTO storage has a 64 KiB lifetime-offset ceiling

The QUIC test demonstrates a single 20 KiB ticket, but both directions of the production CRYPTO stream remain fixed at 64 KiB:

```zig
pub const max_crypto_buffer = 64 * 1024;
```

Inbound `CryptoStream.insert()` rejects any absolute offset at or above that value:

```zig
if (offset >= max_crypto_buffer) return error.CryptoBufferTooLarge;
if (data.len > max_crypto_buffer - offset_usize)
    return error.CryptoBufferTooLarge;
```

Consumed prefixes are not rebased or compacted; `consumed_offset` continues increasing in the same fixed array. `CryptoOutput` similarly stores one fixed 64 KiB application stream.

This means:

- a maximum-size legal NewSessionTicket cannot pass through QUIC because its framed length exceeds 64 KiB;
- several ordinary tickets eventually fail once the cumulative application-CRYPTO offset reaches 64 KiB, even when all prior bytes were consumed;
- the required “one or more” and fragmented/batched ticket behavior is bounded by connection-lifetime traffic rather than outstanding buffered bytes.

Please make the CRYPTO storage bound apply to outstanding data, not absolute lifetime offsets—e.g. a sliding/ring buffer or chunked sparse storage retaining absolute stream offsets separately. Add tests for a maximum configured ticket and for multiple tickets whose cumulative application-CRYPTO bytes exceed 64 KiB while outstanding buffered bytes stay below the configured cap.

## 6. High: required lifecycle and failure coverage remains incomplete

The new tests materially improve coverage, but several #361 cases are still unproven at the public/production boundaries:

- several complete and fragmented tickets delivered in order;
- distinct nonces producing distinct callback-delivered PSKs after one real handshake;
- callback retention through an explicit deep clone surviving callback return;
- consumer/storage allocation refusal remaining nonfatal to an established connection;
- allocation-failure sweeps for lazy reassembly, common state, client state, server state, and exact-size encoding;
- cleanup of a partially received dynamically allocated ticket on malformed input and teardown;
- application data sent before ticket issuance and again after ticket delivery in the same record and QUIC production flows;
- the full codec boundary matrix from the issue: truncation at every fixture byte, nonce 0/255/256, ticket 0/default max/65,535/one-over, mutated vector lengths, trailing data, and proof that undersized encoding leaves no partially usable output.

Please add these alongside the production transport fixes rather than relying only on helper-level tests.

Once these issues are fixed, the core protocol work appears structurally sound. The remaining problems are primarily allocator ownership and ensuring the advertised ticket range actually survives the real record and QUIC carrier implementations.

Jaruso (commented):
**Fourth review result: changes are still required before merge.**

This revision correctly addresses most of the third-round findings:

- allocator replacement is now restricted to the idle lifecycle and allocations remember the allocator that created them;
- server emission no longer has to fit the 32 KiB scratch buffer;
- the record stream can fragment an emitted ticket into multiple TLS records;
- QUIC CRYPTO input storage is no longer permanently limited by the lifetime stream offset;
- real record-stream and QUIC-adapter maximum-ticket tests were added.

The remaining findings below are from a fresh review of head `cb36a1bd8dacc8bbc19eaa07a6b7f31674a14363`, not repetitions of already-fixed code.

## 1. Blocker: the implementation still uses `std.heap.page_allocator` as hidden ticket storage

Issue #361 explicitly requires that ticket processing never use `std.heap.page_allocator` as an implementation shortcut and that allocation be threaded through an explicit ownership boundary. The current implementation now has two such shortcuts:

```zig
const PostHandshakeInput = struct {
    allocator: std.mem.Allocator = std.heap.page_allocator,
    ...
};
```

and:

```zig
pub fn emitOwnedHandshakeBytesCopy(
    self: *EventSink,
    epoch: Epoch,
    data: []const u8,
) ErrorSet!void {
    ...
    const copy = std.heap.page_allocator.alloc(u8, data.len)
        catch return buffer_overflow_error;
    ...
}
```

The first means an ordinary client with no ticket consumer performs attacker-influenced allocations outside the caller's connection allocator and memory accounting. The second is exercised by the QUIC translation path for tickets larger than the 32 KiB scratch area, doubles the live ticket copy, and reports allocation failure as `HandshakeBufferOverflow` rather than a truthful local resource failure.

Please make every allocation explicit. Viable shapes include:

```zig
pub fn initClient(
    allocator: std.mem.Allocator,
    entropy: Entropy,
    trust: Trust,
    profile: TransportProfile,
) Tls13Backend
```

or a required pre-start allocator configuration that the record and QUIC production constructors always wire from their owning connection. There should be no fallback allocator.

For event translation, either pass the destination allocator explicitly:

```zig
try destination.emitOwnedHandshakeBytesCopy(
    connection_allocator,
    epoch,
    data,
);
```

or preferably add an ownership-transfer/move operation so the source sink's exact-size allocation can move into the destination sink without a second 65 KiB copy.

Add failing-allocator tests proving that:

- no ticket path reaches `page_allocator`;
- the source and destination sinks leak nothing when allocation or event insertion fails;
- an allocation failure is not mislabeled as a credential-provider failure or wire buffer overflow.

## 2. Blocker/security: plaintext bearer tickets remain in QUIC adapter buffers after consumption and teardown

`CryptoStream.consumeContiguous()` advances `consumed_offset` and drops ranges but leaves the consumed bytes in `buffer`. `compactConsumed()` copies surviving bytes down without clearing the vacated tail. `CryptoOutput.take()` similarly advances/reset indices without clearing the emitted bytes, and `compact()` leaves stale bytes in the old range.

There is also no `QuicTlsAdapter.deinit()` that wipes the reassembler and outbound arrays. `Handshake.deinit()` only deinitializes its driver, and `Connection.deinitPartial()` calls that driver teardown but never clears the adapter. Consequently, received and emitted opaque ticket identities can remain in connection memory until the allocation is released, and allocator destruction is not secure zeroization.

That violates the #361 requirement to wipe opaque ticket identities and all temporary copies on success, failure, and abandonment.

Please make consumption explicitly secure. For example:

```zig
pub fn copyNextHandshakeInput(
    self: *CryptoStream,
    out: []u8,
) !usize {
    const bytes = self.contiguous();
    const n = @min(bytes.len, out.len);
    @memcpy(out[0..n], bytes[0..n]);
    crypto_secrets.secureZero(
        self.buffer[relative_consumed_start..][0..n],
    );
    self.consumeOffset(n);
    return n;
}
```

The exact API can differ; the important rule is that the adapter wipes a consumed input range only after the driver has finished borrowing it, and wipes an output range immediately after `pollOutput()` copies it to caller storage. Compaction must clear the vacated tail.

Also add something equivalent to:

```zig
pub fn deinit(self: *QuicTlsAdapter) void {
    self.secrets.deinit();
    crypto_secrets.secureZero(std.mem.asBytes(&self.reassembler));
    crypto_secrets.secureZero(std.mem.asBytes(&self.outbound));
    self.* = .{};
}
```

and call it from every `Connection` teardown/failure/abandonment path after the handshake driver is finished with it.

Tests should poison a ticket pattern and prove it is absent after input consumption, output draining, compaction, successful teardown, failed processing, and abandonment.

## 3. High: maximum-ticket support adds an unnecessary ~66 KiB to every record-mode connection

`PureZigRecordStream.max_ciphertext_queue` was doubled from four maximum records to eight:

```zig
pub const max_ciphertext_queue =
    8 * record_codec.max_ciphertext_record_len;
```

This is inline `ByteQueue` storage, so it permanently increases every record-mode connection by four maximum ciphertext records—66,580 bytes—even when session tickets are never used.

The increase is caused by this conservative preflight:

```zig
const needed = count * record_codec.max_ciphertext_record_len;
```

A canonical maximum emitted ticket is 65,815 handshake bytes. It needs five records, but its exact protected size under the current AES-GCM profile is only:

```text
65,815 + 5 * (5-byte header + 1-byte inner type + 16-byte tag)
= 65,925 bytes
```

The previous four-record queue is 66,580 bytes, so it already has enough physical capacity. Only the `count * maximum_record_size` estimate makes it appear too small.

Please preflight the exact output length using the active write state's AEAD tag length, ideally through one canonical bridge helper:

```zig
const needed = try self.bridge.sealedHandshakeLen(epoch, bytes.len);
if (!self.hasHardRoom(.outbound_ciphertext, current, needed))
    return error.WouldBlock;
```

Then restore the four-record queue. An incremental pending-message design would also work, but increasing permanent per-connection storage to accommodate a rare post-handshake maximum is the wrong tradeoff for a proxy.

Add a regression assertion that the canonical maximum ticket fits the configured queue without increasing the queue above four maximum records.

## 4. Blocker: the production QUIC `Connection` still has no application-CRYPTO transmit/retransmit path

The new QUIC test proves the `Tls13Backend` → translated sink → `QuicTlsAdapter` → `Handshake.pollOutput(.application)` path, but it manually invokes `translate()`, `queueHandshakeOutput()`, and `pollOutput()`. It does not exercise the production `quic.Connection`.

The actual connection still has only:

```zig
crypto_tx: [2]CryptoTx = .{ .{}, .{} },
```

and `collectCryptoOutput()` drains only `.initial` and `.handshake`. Its loss requeue logic explicitly returns for `.application` CRYPTO. Therefore a ticket queued at the handshake adapter cannot currently become retransmittable CRYPTO frames in 1-RTT packets through the production connection.

This belongs to #361 rather than #365: #365 owns *using* a saved ticket in a later resumed handshake, while #361 owns delivery of the post-handshake `NewSessionTicket` bytes in both record mode and QUIC.

Please add a production application-CRYPTO ownership path. At minimum it needs:

- application-level CRYPTO transmit storage with absolute offsets;
- collection of `Handshake.pollOutput(.application)` after explicit ticket issuance;
- 1-RTT CRYPTO frame assembly;
- sent-range tracking, ACK release, loss/PTO requeue, and secure wipe after release;
- a narrow explicit issuance API reachable through `Connection` or its TLS owner without reaching into adapter internals.

Add a test through real `Connection` instances—not only `RealHandshakeHarness`—that issues a ticket, carries it in 1-RTT CRYPTO frames, drops/retransmits at least one frame, delivers the callback exactly once, and continues normal stream traffic afterward.

## 5. High: several explicit #361 acceptance tests are still missing

The new carrier tests are valuable, but the issue's required matrix is not yet complete. At minimum I still cannot find coverage for:

- several complete and fragmented tickets delivered in order;
- callback retention through an explicit `ClientTicketState.cloneInto()` and validity after callback return;
- consumer allocation/storage refusal remaining nonfatal to the established connection;
- allocation-failure sweeps for client-state construction, server-state construction, reassembly, and owned event translation;
- truncation at every byte of deterministic codec fixtures;
- every vector-length field at zero, one-under, exact, one-over, and maximum;
- nonce encode boundaries 0/255/256 and ticket encode boundaries 0/default max/65,535/one-over;
- undersized encoding proving the destination remains unusable/unchanged;
- exact maximum receive-side ticket framing with the 65,534-byte unknown-extension vector through the QUIC input path;
- the QUIC ticket-buffer zeroization cases described above.

CI is green, but it currently proves the implemented examples rather than the full malformed-input, allocation, ownership, and lifecycle contract required by #361.

Once these five areas are corrected, I expect the wire codec, RMS/PSK derivation, server/client state construction, and the newly added record carrier path to be in good shape.

Jaruso (commented):
**Re-review result: changes are still required before merge.**

The fifth commit is a substantial improvement. It fixes the prior hidden `page_allocator` copy path, restores the four-record ciphertext queue with exact sizing, adds adapter teardown/zeroization hooks, and finally carries application-level CRYPTO through the production `Connection` with retransmission storage. CI run 833 is green.

The resulting production path still has correctness and secret-lifecycle blockers, however.

## 1. Blocker: the default client constructors still turn a valid NewSessionTicket into a connection error

`PostHandshakeInput.allocator` is now correctly nullable, but receiving the first complete four-byte post-handshake header calls:

```zig
const allocator = self.allocator orelse return error.InvalidHandshakeState;
```

The original shared and QUIC client constructors still leave that allocator unset:

```zig
pub fn initClient(...) Tls13Backend {
    return initClientWithOptions(...);
}
```

Only the newly added `initClientWithAllocator*` variants install it. This means a client created through the existing/default API cannot satisfy the issue's required behavior that, with no consumer configured, a valid ticket is fully parsed and dropped without failing the established connection.

This is not only a theoretical compatibility path. `tests/h3_interop_tool.zig` still creates its real external-server client with `Tls13Backend.initClient(...)`. Real QUIC/TLS servers commonly send NewSessionTicket after the handshake, so that tool can now close an otherwise successful connection with `InvalidHandshakeState` as soon as the peer issues a ticket.

Please make post-handshake allocation an unavoidable part of client construction rather than an optional capability hidden behind a second constructor. Viable shapes include:

```zig
pub fn initClient(
    allocator: std.mem.Allocator,
    entropy: Entropy,
    trust: Trust,
) HandshakeError!Tls13Backend
```

or an allocator-install hook on `TlsBackend` that `Connection.init` must call with its own allocator before `Handshake.start`. The no-allocator constructors should either be removed/deprecated or provide an allocation-free validate-and-drop path. Update every production and interop caller, and add a regression using the ordinary no-consumer client shape rather than manually preconfiguring `setPostHandshakeAllocator`.

## 2. Blocker: Connection ticket emission consumes output before destination storage commits

`Handshake.pollOutput` now copies an adapter chunk and immediately calls `discardHandshakeOutput`. `Connection.collectCryptoOutput` then performs two fallible operations:

```zig
const output = try handshake.pollOutput(...); // source already discarded
try tx.data.appendSlice(self.allocator, output.bytes);
try tx.pending.insert(self.allocator, range);
```

If either allocation fails:

- the adapter has already erased/advanced past the chunk;
- the backend has already added the ticket to its transcript;
- `Connection.emitNewSessionTicket` merely returns `HandshakeBufferOverflow`;
- the connection remains usable, but the ticket is partially or completely absent from its CRYPTO stream.

The second failure point is worse: if `appendSlice` succeeds and `pending.insert` fails, the bytes remain in `tx.data` but are never eligible for transmission. Later chunks continue from the advanced stream offset, permanently creating a hole.

`queueTlsEvents` has the same transcript-commit problem if the adapter application output is already full: the shared backend has accepted the sink and updated its transcript before the production carrier rejects the message.

Please make this a transactional transfer. At minimum:

1. preflight the framed ticket length and the configured outstanding-CRYPTO limit;
2. reserve all destination byte and range bookkeeping before calling the backend or before consuming adapter output;
3. expose a `peek`/`commit` output API so the adapter advances only after the destination copy and range insertion are guaranteed;
4. ensure no fallible carrier operation remains after the backend commits the ticket transcript.

A two-phase `prepareNewSessionTicket` / `commitPreparedTicket` API may be cleaner if carrier capacity cannot be preflighted through the current interface.

Add failing-allocator tests for every `crypto_tx[2].data` and `pending` allocation, plus an adapter-full test. Each must prove either that the original output remains retryable with an unchanged transcript, or that the connection is deterministically failed and all partial state is cleared. Returning an error while leaving a live, transcript-diverged connection is not safe.

## 3. Blocker: application CRYPTO retransmission bookkeeping breaks under normal PTO/ACK ordering

The new `CryptoTx.base` compaction does not account for duplicate transmissions.

`firePto` requeues the oldest unacknowledged `SentRecord` without removing the original record. Therefore the same CRYPTO range can exist in `pending` and/or in multiple in-flight packet records. When either copy is ACKed, `markAcked` immediately advances `base` and removes the bytes.

There are then several failure sequences:

### ACK arrives after PTO requeue but before the probe is built

1. PTO inserts `[0, N)` into `tx.pending`.
2. The original packet is ACKed.
3. `compactAcked` advances `base` to `N`.
4. `tx.pending` still contains `[0, N)`.
5. The next packet calls `tx.slice(range)`, evaluating `range.start - self.base` and underflowing.

### A duplicate ACK arrives after one copy already compacted the range

`markAcked` inserts the stale `[0, N)` range into `acked`. Since its start no longer equals `base`, `compactAcked` returns without removing it. A later ACK for new adjacent data merges with that stale range, so the first range continues to start below `base` and all future compaction is permanently blocked.

### A stale duplicate is later declared lost

`requeueRecord` inserts its range without the below-`base` clamping already used by `SendQueue`. The next send reaches the same subtraction underflow.

Normalize every ACK, pending, and retransmitted CRYPTO range against `base`:

```zig
fn liveRange(self: *const CryptoTx, incoming: Range) ?Range {
    if (incoming.end <= self.base) return null;
    var range = incoming;
    if (range.start < self.base) range.start = self.base;
    return range;
}
```

When advancing `base`, also remove or trim any `pending` ranges below the new base. Late ACKs for already-released bytes should be ignored rather than retained in `acked`.

There is a second loss bug in `buildPacket`: it may encode multiple non-contiguous CRYPTO ranges into one packet, but `SentRecord.crypto` stores only one range. If the next range is not contiguous, it is still encoded while the record continues to describe only the first range. Loss of that packet therefore permanently loses the unrecorded range. Either stop after the first non-contiguous range and reinsert it:

```zig
if (record.crypto) |existing| {
    if (existing.end != range.start) {
        tx.pending.insert(self.allocator, range) catch {};
        break;
    }
    record.crypto = .{ .start = existing.start, .end = range.end };
}
```

or allow `SentRecord` to track multiple CRYPTO ranges.

The current test manually calls `requeueRecord` but does not exercise the ACK/PTO interleavings above. Add deterministic tests for original-ACK-before-probe, retransmission-ACK-before-original-loss, late duplicate ACK, and two non-contiguous lost ranges.

## 4. High: acknowledged ticket bytes are still left in freed QUIC transmit storage

`CryptoTx.compactAcked` currently does this:

```zig
secureZero(data.items[0..release_len]);
copyForwards(data.items[0..keep_len], data.items[release_len..]);
data.shrinkRetainingCapacity(keep_len);
```

After the copy, the old tail still contains a duplicate of the retained ticket bytes. It is outside `items.len`, so `deinit` later wipes only the shortened slice. When the complete ticket is ACKed, `keep_len` becomes zero; `deinit` wipes zero bytes and frees the allocation containing the stale ticket copy.

The vacated range must be cleared before shrinking:

```zig
const old_len = self.data.items.len;
if (keep_len > 0) {
    std.mem.copyForwards(
        u8,
        self.data.items[0..keep_len],
        self.data.items[release_len..old_len],
    );
}
crypto_secrets.secureZero(self.data.items[keep_len..old_len]);
self.data.shrinkRetainingCapacity(keep_len);
```

There are two additional copies that need explicit treatment:

- `collectCryptoOutput`'s `var chunk: [2048]u8` is never wiped;
- growing `std.ArrayList(u8)` can move and free its old backing allocation without zeroing it, so appending a ticket in 2 KiB increments is not a secret-safe growth strategy.

Use exact-size/segment ownership or a zeroizing grow helper for application CRYPTO. One good design is retaining each owned ticket message as a separately allocated CRYPTO segment until all of its ranges are ACKed, then wiping and freeing that segment. That avoids both repeated reallocations and capacity-tail ambiguity.

Add tests with a poison/observing allocator that verify acknowledged tickets, compacted tails, reallocated old buffers, temporary chunks, and teardown paths are zeroed before release.

## 5. High: outstanding application CRYPTO is now unbounded

`crypto_tx[2].data` is a growable `ArrayList` with no hard limit. Each successful `emitNewSessionTicket` drains the adapter into it, and the bytes remain until ACKed. A caller can issue tickets repeatedly while the peer withholds ACKs, growing connection-owned memory without bound.

The ticket size itself is bounded, but the queued ticket count is not. Add an explicit maximum outstanding application-CRYPTO byte budget or configured ticket-flight budget. Capacity refusal must occur before backend transcript mutation, returning a deterministic backpressure/bounded error. The tests should issue tickets up to the limit, verify one-over refusal is atomic, ACK a prefix, and verify issuance can resume.

## 6. Required #361 coverage is still incomplete

The latest commit adds a useful real-Connection ticket test, but the acceptance matrix still lacks several explicitly required cases:

- several complete/fragmented tickets delivered in order;
- callback retention through an explicit deep clone after callback return;
- local storage refusal without connection failure;
- allocation-failure sweeps for owned construction and carrier-transfer paths;
- exhaustive fixture truncation and vector-length mutation tests;
- nonce 0/255/256 and ticket 0/default-max/65535/one-over codec boundaries;
- maximum-size ticket through the actual `Connection` under packet loss/reordering;
- ACK-driven transmit-buffer zeroization;
- default production client receiving and dropping a ticket with no consumer.

`src/tls/new_session_ticket.zig` was not changed by this fifth commit and still has only the small grouped malformed/boundary cases, not the issue's exhaustive mutation matrix.

## Verdict

The record-mode sizing fix and the new production QUIC path are directionally correct, but the new application-CRYPTO buffer is not yet safe under allocation failure, PTO duplication, out-of-order ACK/loss, repeated issuance, or secret cleanup. These are production correctness/security issues, not test-only gaps, so the PR should not merge yet.

Jaruso (commented):
**Re-review result: changes are still required before merge.**

I reviewed the new head `d8115f4fb1428daf27dc55879139f55d4451030c` and rechecked the full #361 contract rather than only the latest delta. CI run 834 is green, and this commit correctly fixes several prior findings: ordinary QUIC clients now receive a post-handshake allocator, stale retransmission ranges are clamped below `base`, non-contiguous CRYPTO ranges are no longer silently coalesced into one `SentRecord`, application CRYPTO is bounded, and retained `ArrayList` capacity is wiped during teardown.

The following issues remain.

## 1. Blocker: ordinary record-mode clients still fail when a server sends a valid ticket and no consumer is configured

The QUIC `Connection` now installs the post-handshake allocator automatically, but the production record-stream construction path does not.

`SocketHarness.createWithAllocator()` accurately demonstrates the current production shape:

```zig
self.client_engine = Tls13Backend.initClient(...);
if (opts.client_post_handshake_allocator) |post_allocator| {
    try self.client_engine.setPostHandshakeAllocator(post_allocator);
}
self.client = PureZigRecordStream.initWithCarrierAndBackend(..., self.client_engine.backend());
```

The no-consumer record test only succeeds because it explicitly sets:

```zig
.client_post_handshake_allocator = std.testing.allocator,
```

Without that test-only/manual setup, `PostHandshakeInput.append()` reaches:

```zig
const allocator = self.allocator orelse return error.InvalidHandshakeState;
```

as soon as it has parsed the four-byte ticket header. The established record-mode TLS stream then fails instead of fully parsing and dropping the valid ticket, contrary to #361.

Please wire allocator ownership into the normal record transport construction path, analogous to the QUIC fix. The regression test must use the ordinary production constructor with no consumer and no explicit `setPostHandshakeAllocator()` call.

## 2. Blocker: application-CRYPTO reliability still depends on fallible allocations whose errors are silently discarded

The new `CryptoTx` solves stale-range arithmetic, but ACK and loss processing still allocate dynamically after the ticket has been accepted and committed to the TLS transcript:

```zig
fn markAcked(...) void {
    ...
    self.acked.insert(allocator, live) catch {};
    self.compactAcked();
}
```

and:

```zig
if (tx.liveRange(range)) |live|
    tx.pending.insert(self.allocator, live) catch {};
```

This creates two deterministic failure modes:

1. A maximum ticket spans many QUIC packets. After the original pending range has been sent, the first disjoint lost packet may reuse the range list's existing capacity, while a second disjoint lost packet may require growth. If that allocation fails, `detectAndRequeueLost()` still removes the corresponding `SentRecord`, but the CRYPTO range was never requeued. The peer is left with a permanent CRYPTO-stream gap and the ticket can never complete.
2. If `acked.insert()` fails, the peer has acknowledged the bytes and the sent record is removed, but the prefix cannot be compacted. Those bytes consume `max_application_crypto_outstanding` indefinitely and can permanently prevent later ticket issuance.

The ticket path must not silently lose reliability or capacity because bookkeeping allocation failed after acceptance. Before calling the TLS emission API, reserve all bounded storage needed for the accepted CRYPTO message, including worst-case pending and ACK range bookkeeping, or replace these lists with fixed/bounded storage. Once the transcript is advanced and the ticket API returns success, loss/ACK processing should be allocation-free.

Add failing-allocator tests covering:

- several non-contiguous lost ticket packets;
- out-of-order ACK ranges;
- PTO duplicate plus original ACK;
- cleanup and retry behavior after each injected allocation failure.

## 3. High: the first ticket allocates the entire two-ticket cap and retains it for the rest of the connection

`reserveAppend()` currently does this when `capacity == 0`:

```zig
try self.data.ensureTotalCapacity(allocator, max_outstanding);
```

`max_outstanding` is twice the maximum emitted ticket message: `2 * 65,815 = 131,630` bytes. Therefore even a tiny ticket causes a roughly 128.5 KiB allocation, and ACK compaction only shrinks `items.len`; it never releases the allocation. A server that normally issues one ticket on every connection retains about 1.23 GiB of ticket capacity across 10,000 live connections.

This allocation also happens before the backend validates caller-tightened model limits. For example, a wire-valid 4,097-byte ticket with `Limits.default` allocates the full 131,630-byte buffer and then returns `TicketTooLarge`, leaving the large capacity attached to a connection that never emitted a ticket.

Please:

- perform the complete model/limits preflight before allocating transport storage;
- allocate storage proportional to accepted outstanding bytes rather than the global cap;
- securely wipe and free fully acknowledged ticket chunks instead of retaining the maximum capacity forever;
- keep the aggregate outstanding-byte cap as the independent boundedness guard.

A secure chunked CRYPTO buffer is a natural fit: each accepted ticket owns an exact-sized allocation, packet ranges reference absolute offsets into the chunk list, and a fully ACKed leading chunk is wiped and freed. This avoids unsafe secret-bearing reallocations while preserving the aggregate bound.

## 4. Required #361 acceptance coverage is still incomplete

The current tests improve the carrier path, but the mandatory matrix is not yet present. In particular, I still do not see tests for:

- several complete and fragmented tickets delivered in order through the backend;
- distinct delivered PSKs for several tickets in one completed connection;
- consumer deep-clone retention surviving callback return;
- a consumer that refuses local storage without affecting the connection;
- allocation-failure sweeps for post-handshake reassembly, client-state construction, event ownership, and QUIC pending/ACK storage;
- maximum-size ticket delivery through the actual `Connection` under multi-packet loss/reordering and subsequent ACK compaction;
- codec truncation at every fixture byte and systematic zero/one-under/exact/one-over mutations of each vector length.

The new QUIC test exercises one small ticket and manually calls `requeueRecord`; it does not exercise the maximum ticket across dozens of packets or the real loss detector's multiple disjoint requeue ranges. The record no-consumer test also masks the production constructor defect by manually installing the allocator.

Once the record allocator is wired normally, application-CRYPTO bookkeeping is allocation-safe, and the required lifecycle/failure matrix is present, this should be ready for another merge review.

Jaruso (commented):
**Seventh review result: changes are still required before merge.**

This commit fixes several prior findings: ordinary record clients now have a reassembly allocator, the application-CRYPTO allocation is exact-sized and freed after full ACK, model limits are checked before CRYPTO allocation, and application ACK/loss range operations are pre-reserved instead of silently dropping `OutOfMemory`. CI run 835 is also green.

However, the resulting implementation still does not satisfy issue #361's multiple-ticket, explicit-allocation, and failure-atomicity contracts.

## 1. `CryptoTx.reserveAppend` rejects a second outstanding ticket even when the configured byte budget has room

`reserveAppend` allocates exactly `bytes_len` when the buffer is empty, then refuses to grow a non-empty allocation:

```zig
if (self.data.capacity == 0) {
    try self.data.ensureTotalCapacityPrecise(allocator, bytes_len);
} else if (self.data.capacity < self.data.items.len + bytes_len) {
    return error.CryptoBufferTooLarge;
}
```

That means this valid sequence fails:

1. Emit ticket A, message length 40. `data.len == 40`, `capacity == 40`.
2. Before A is acknowledged, emit ticket B, message length 40.
3. Total outstanding would be 80 bytes, far below `max_application_crypto_outstanding` (~131 KiB).
4. The second call returns `HandshakeBufferOverflow` solely because the first exact allocation cannot grow.

The constant is explicitly sized for two maximum emitted tickets, and #361 requires that a completed server can emit **one or more** tickets and that several tickets coexist in order. The current tests avoid this by changing the earlier two-append test into one combined `"ticket-1ticket-2"` append; that does not exercise two issuance calls.

**Required fix:** allow secure growth while preserving zeroization. For example, allocate `new_total = data.len + bytes_len`, copy the live bytes, securely wipe and free the old allocation, then commit the replacement only after every required range allocation also succeeds. An owned-chunk queue per emitted ticket would also avoid relocating secrets.

Add a real `Connection.emitNewSessionTicket` test that emits at least two tickets back-to-back before any `pollTransmit`/ACK, delivers them in order, checks distinct nonce-derived PSKs, and keeps application traffic usable.

## 2. `reserveAppend` is not failure-atomic and can poison retries

The function mutates allocation state in this order:

```zig
try self.data.ensureTotalCapacityPrecise(allocator, bytes_len);
try self.pending.ensureUnusedCapacity(allocator, range_budget + 1);
try self.acked.ensureUnusedCapacity(allocator, range_budget);
```

If the data allocation succeeds and either range allocation fails, the ticket is not emitted but an empty `data` allocation remains attached to the connection. A later larger ticket then hits the non-growing-capacity branch above and is rejected as `CryptoBufferTooLarge`, even though there are zero outstanding CRYPTO bytes. This also violates #361's requirement for no retained allocation after injected failures.

**Required fix:** make reservation transactional. Allocate all replacement storage into temporaries, and only swap it into `CryptoTx` after every allocation succeeds. On any failure, securely wipe/free all temporary storage and leave `data`, `pending`, `acked`, `base`, and the TLS transcript unchanged.

Add failing-allocator sweeps that fail each allocation in this reservation path, assert the `CryptoTx` snapshot is unchanged with zero retained capacity, then retry with a larger valid ticket and verify delivery.

## 3. Record mode still uses a hidden global allocator and can panic in a generic constructor

All backend-owning record constructors now do this:

```zig
if (role == .client)
    backend.setPostHandshakeAllocator(std.heap.smp_allocator) catch unreachable;
```

This is not the backend/driver's explicit allocator required by #361; it is another ambient global allocator shortcut. It also overwrites an allocator already selected through `setSessionTicketConsumer`, prevents deterministic OOM/accounting coverage for reassembly, and turns an ordinary backend error into a process panic in the non-fallible constructors.

The updated no-consumer test now passes because it silently allocates from `smp_allocator`, not because allocator ownership was correctly threaded through the production API.

**Required fix:** add allocator-aware record constructors (or store the caller allocator in the driver/backend owner) and pass that explicit allocator to `setPostHandshakeAllocator`. Do not override a previously configured allocator with a global heap, and do not use `catch unreachable` for a generic backend callback.

The no-consumer socket test should use a caller-owned allocator through the production constructor and verify allocations are reclaimed through that allocator.

## 4. Required #361 coverage remains incomplete

Still missing from the PR's effective test matrix:

- several complete/fragmented tickets delivered in order;
- callback retention through an explicit deep clone that remains valid after callback return;
- local storage refusal remaining non-fatal;
- allocation-failure sweeps for client build/clone, server build/emission, post-handshake reassembly, and QUIC reservation;
- no retained allocations after each injected failure;
- exhaustive fixture truncation and length-field mutation coverage;
- a maximum-size ticket through the real QUIC `Connection` loss detector, rather than one small ticket with direct/manual `requeueRecord`.

**Merge verdict:** not ready yet. The normal one-ticket paths are substantially improved, but back-to-back issuance currently fails by construction and allocation failures can leave transport state mutated.

Jaruso (commented):
**Deep re-review result: changes are still required before merge.**

I reviewed the complete resulting implementation at `a968f5ab373e2d328b4cf85d64633900dc26e141`, not only the eighth-commit delta, against the full #361 protocol, ownership, transport, failure-atomicity, and test requirements. CI 836 is green, and this commit correctly fixes the previous back-to-back-ticket, `CryptoTx.reserveAppend` transactionality, retry-poisoning, retained-full-capacity, and hidden-global-allocator findings. The RFC codec, RMS timing, SHA-256/SHA-384 helpers, callback best-effort behavior, owned event payload transfer, QUIC ACK/loss bookkeeping, and five-record maximum record-mode ticket path now look structurally sound.

There are still two code blockers and one explicit acceptance-criteria blocker.

## 1. The allocator-less record client constructors are still public production paths that fail on a valid NewSessionTicket

`PureZigRecordStream` still exposes all of these constructors without an allocator:

- `initWithBackend`
- `initWithBackendAndLimits`
- `initWithCarrierAndBackend`
- `initWithCarrierBackendAndLimits`

`initWithCarrierAndBackend` is even documented as “the production shape.” None of these installs the client post-handshake allocator. The new allocator-taking variants do, but callers are not prevented from using the old variants for `.client`.

A client constructed through one of the old APIs completes the handshake normally. When a server later sends any valid NST, `PostHandshakeInput` consumes the four-byte header and then executes:

```zig
const allocator = self.allocator orelse return error.InvalidHandshakeState;
```

The established connection therefore fails merely because a conforming TLS 1.3 server sent a ticket. The updated no-consumer integration test uses `initWithCarrierBackendAndAllocator`, so it does not protect the old public “production” constructor.

The cleanest fix is to require an allocator on every backend-owning constructor. `setPostHandshakeAllocator` is already first-set-wins, so passing the owner allocator does not overwrite the allocator installed by `setSessionTicketConsumer`:

```zig
pub fn initWithCarrierAndBackend(
    allocator: std.mem.Allocator,
    role: tls_state.Role,
    crypto_provider: provider.CryptoProvider,
    cipher_suite: algorithms.CipherSuite,
    carrier: Carrier,
    backend: RecordHandshakeBackend,
) Error!PureZigRecordStream {
    std.debug.assert(!carrier.owns_handle or carrier.closeFn != null);
    var stream_state = init(role, crypto_provider, cipher_suite);
    if (role == .client) try backend.setPostHandshakeAllocator(allocator);
    stream_state.handshake_driver = RecordHandshakeDriver.init(role, backend);
    stream_state.carrier = carrier;
    return stream_state;
}
```

Alternatively, retain explicitly named server/preconfigured-backend constructors, but the ordinary client API must not silently create an NST-incapable client. Add a test that deliberately uses whichever constructor remains the ordinary production client constructor, configures no consumer, receives a ticket, and continues application traffic.

## 2. `Connection.emitNewSessionTicket` is not failure-atomic across CRYPTO reservation and TLS state construction

The new `CryptoTx.reserveAppend` is internally transactional, which fixes the previous partial-reservation bug. However, `Connection.emitNewSessionTicket` installs that successful reservation **before** the TLS backend validates/builds the complete server state:

```zig
try tx.reserveAppend(...);
var ticket_state = try self.tls.emitNewSessionTicket(...);
```

Only the session limits and opaque-ticket length are checked before reservation. Several backend failures can still happen afterward:

- `ticket_lifetime == 0` passes `encodedLen`, reserves CRYPTO storage, and is then rejected by `buildServerRecoverableState` as `InvalidLifetime`;
- caller-tightened compatibility limits can reject the QUIC transport snapshot after reservation;
- any owned server-state/message allocation can fail after reservation.

On those paths no ticket bytes are appended, no ACK can trigger `compactAcked`, and the newly installed data/range capacities remain attached to the connection even though issuance failed. This violates #361’s required “no retained allocation after decode/build failure” and failure-injection cleanup contract. It also means the public operation is not atomic even though the helper now is.

Please make reservation a prepare/commit operation rather than installing capacity before the backend succeeds. For example:

```zig
var reservation = try tx.prepareAppend(
    self.allocator,
    message_len,
    max_application_crypto_outstanding,
);
defer reservation.deinit(); // wipes/frees uncommitted replacements

var ticket_state = try self.tls.emitNewSessionTicket(
    self.allocator,
    &sink,
    params,
    limits,
);
errdefer ticket_state.deinit();

// All fallible allocation has completed. This is infallible.
tx.commitReservation(self.allocator, &reservation);
try self.queueApplicationCryptoEventsReserved(&sink, message_len);
```

The reservation object can retain the replacement allocations without freeing the old buffers until commit. Another acceptable design is a two-phase TLS `prepareNewSessionTicket` / `commitNewSessionTicket`, but there must be no transcript mutation or installed CRYPTO capacity until all fallible preparation succeeds.

At minimum add tests for:

```zig
// Deterministic backend rejection after CRYPTO sizing.
try expectError(error.InvalidHandshakeState, server.emitNewSessionTicket(.{
    .ticket_lifetime = 0,
    .ticket_age_add = 1,
    .ticket_nonce = "\x01",
    .opaque_ticket = "ticket",
    .issued_at_unix_ms = 10,
}, limits));
try expectEqual(@as(usize, 0), server.crypto_tx[2].data.capacity);
try expect(server.crypto_tx[2].pending.isEmpty());
try expect(server.crypto_tx[2].acked.isEmpty());
```

Also inject failure in every allocation after CRYPTO sizing and prove that the transcript hash, CRYPTO offsets, retained capacities, pending/acked ranges, and subsequent successful retry are unchanged/correct.

## 3. The explicit #361 required-test matrix is still incomplete

The current tests cover many important happy paths and several boundaries, but the issue marks the following as required and they are still absent:

### Wire codec

- truncation at **every byte** of deterministic checked-in fixtures;
- every length field mutated to zero, one-under, exact, one-over, and maximum;
- an assertion that undersized-output failure leaves the destination unchanged;
- the complete lifetime/nonce/ticket/default-limit/absolute-limit mutation matrix.

### Backend callback and ownership

- several complete and differently fragmented tickets delivered in order in one application-epoch input sequence;
- callback retention via an explicit `ClientTicketState.cloneInto`, followed by use after the callback-owned source has been destroyed;
- local storage/clone refusal using a failing allocator, proving the established connection remains usable;
- allocation-failure sweeps for client state construction/clone, server state construction/emission, post-handshake reassembly, and transport-owned payload transfer;
- teardown/zeroization assertions after each injected failure path.

### QUIC integration

- a maximum allowed opaque ticket through the real `Connection` packetizer and client `Connection` reassembler;
- actual packet loss, reordering, ACK processing, and PTO/loss detection rather than directly calling `requeueRecord`;
- partial/full ACK cleanup after CRYPTO-buffer growth, proving retained capacity returns to zero and ticket bytes are wiped;
- two or more tickets where at least one is fragmented/reordered, delivered exactly once and in order while stream traffic continues.

Representative patterns:

```zig
for (0..fixture.len) |cut| {
    try std.testing.expectError(
        error.MalformedHandshake,
        new_session_ticket.decode(fixture[0..cut]),
    );
}
```

```zig
fn onTicket(ctx: *anyopaque, ticket: *const session.ClientTicketState) void {
    const capture: *Capture = @ptrCast(@alignCast(ctx));
    ticket.cloneInto(capture.allocator, &capture.retained) catch {
        capture.storage_refused = true;
        return;
    };
}
// After callback return, verify `retained` still owns the ticket/PSK and that
// refusal did not close the TLS/QUIC connection.
```

```zig
try std.testing.checkAllAllocationFailures(
    std.testing.allocator,
    exerciseTicketReceiveAndClone,
    .{},
);
```

These are not optional hardening suggestions; they are enumerated under the issue’s **Required tests** and allocation/fuzz-readiness requirements. Given how many interaction defects the prior revisions exposed, the PR should not close #361 without them.

## Verdict

This is now close, and the eighth commit resolves the previous review’s substantive retransmission and allocation defects. I did not find another protocol-derivation or packet-recovery flaw in the full pass. However, the allocator-less public client path, operation-level reservation atomicity, and unfulfilled required-test matrix still prevent a merge approval.

Jaruso (commented):
**Re-review result: the implementation defects are fixed, but #361 is still not complete because its required acceptance-test matrix is not present.**

I re-reviewed the ninth commit and the full affected paths. The latest changes correctly resolve the prior production-code blockers:

- all record constructors that own a handshake backend now require an explicit allocator, so an ordinary client can always reassemble and validate a post-handshake ticket;
- QUIC application-CRYPTO reservation is now prepared privately and committed only after the TLS backend successfully builds/emits the ticket;
- rejected issuance frees the uncommitted reservation and is retryable;
- the range-capacity assumption is consistent with the packet builder's bounded 32-range ACK frame, maximum short-header overhead, and CRYPTO-first packet ordering;
- CI run 837 is green.

I do **not** see another production-code change required in the current implementation. The remaining blocker is coverage that the issue explicitly labels **Required tests** and makes part of its acceptance contract.

## Required remaining work

### 1. Complete the wire-codec mutation matrix

`src/tls/new_session_ticket.zig` currently has representative malformed cases, but it does not include the required exhaustive fixture truncation or length-field mutation matrix.

Add deterministic fixtures and test every prefix:

```zig
const fixture = try encode(.{
    .ticket_lifetime = 60,
    .ticket_age_add = 0x11223344,
    .ticket_nonce = "\x01\x02",
    .ticket = "opaque-ticket",
    .max_early_data_size = 32,
}, &storage);

for (0..fixture.len) |cut| {
    try std.testing.expectError(
        error.MalformedHandshake,
        decode(fixture[0..cut]),
    );
}
```

Also mutate each variable-length field independently to the issue's required boundary values:

- nonce length: `0`, one-under, exact, one-over, `255`, and unrepresentable/encoder-rejected `256`;
- ticket length: `0`, one-under, exact, one-over, configured default maximum, absolute `65535`, and one-over where representable through the API;
- extensions vector length: `0`, one-under, exact, one-over, `65534`, and illegal `65535`;
- `early_data` payload length: `0`, `3`, exact `4`, `5`, and maximum;
- trailing bytes after the declared extension vector;
- lifetime `0`, exact `604800`, and `604801` through actual wire encode/decode paths.

The undersized-output test should also prove the destination is unchanged, not only that `OutputTooSmall` is returned:

```zig
var out = [_]u8{0xa5} ** 8;
const before = out;
try std.testing.expectError(error.OutputTooSmall, encode(params, &out));
try std.testing.expectEqualSlices(u8, &before, &out);
```

### 2. Prove callback ownership and best-effort refusal behavior

The implementation has the correct ownership shape, but the issue explicitly requires proof that retaining a ticket requires a deep clone and that the clone remains valid after the callback returns and the backend destroys its callback-owned value.

Add a consumer that clones during the callback:

```zig
const Capture = struct {
    allocator: std.mem.Allocator,
    retained: session.ClientTicketState = .{},
    count: usize = 0,

    fn onTicket(ctx: *anyopaque, ticket: *const session.ClientTicketState) void {
        const self: *@This() = @ptrCast(@alignCast(ctx));
        ticket.cloneInto(self.allocator, &self.retained) catch unreachable;
        self.count += 1;
    }
};
```

After `onNewSessionTicket`/record delivery returns, verify the retained ticket, nonce, PSK, SNI/ALPN, compatibility snapshots, and timing fields still match. Deinitialize it independently.

Also inject allocation refusal specifically during client-state construction and prove it does **not** fail the established connection or invoke the callback. A direct backend test can bypass post-handshake-frame allocation so the injected failure lands in `buildClientTicketState`; a record/QUIC regression should then prove later application traffic still works.

### 3. Add allocation-failure sweeps for each owned path

The current failing-allocator coverage protects `CryptoTx` reservation, but #361 separately requires sweeps for the ticket-owned paths. Add `checkAllAllocationFailures` or equivalent coverage for:

- `buildClientTicketState` including common metadata, ticket allocation, and compatibility snapshots;
- `buildServerRecoverableState`;
- server message allocation and owned `EventSink` transfer;
- fragmented post-handshake reassembly allocation;
- clone retention from the callback;
- QUIC ticket issuance with reservation prepared but backend/state/message allocation failing.

Each sweep should verify either a fully owned result or an error with no leaked/retained allocation, no callback delivery, no transcript/output mutation where issuance was rejected, and a subsequent successful retry.

### 4. Exercise a maximum ticket through real QUIC recovery

The existing QUIC test uses a small ticket and directly calls `requeueRecord`. That validates the buffer seam but does not satisfy the issue's required end-to-end loss/reordering/ACK/PTO case.

Add a test that:

1. completes a real `TestPair` handshake;
2. emits a `65535`-byte ticket under widened limits;
3. collects the actual application packets;
4. drops multiple non-contiguous packets and delivers others out of order;
5. advances `now_us` through the real loss/PTO path rather than directly invoking `requeueRecord`;
6. delivers retransmissions and peer ACKs;
7. verifies exactly one callback, exact ticket bytes/PSK, continued stream traffic, and eventual release of `crypto_tx[2]` data/capacity after ACK.

This should include multiple disjoint losses so the pre-reserved pending/ACK metadata is exercised under the scenario it was designed to make allocation-free.

## Verdict

The implementation now looks correct and CI is green, but this PR still closes an issue whose explicit required tests are materially incomplete. Please add the tests above; after that, assuming they pass without forcing implementation changes, I expect this to be ready to merge.

Jaruso (commented):
**Final re-review: merge-ready once the current CI run completes successfully.**

I re-reviewed the tenth commit (`011c4002badd9fb094ca228b556a7ee9b091ba48`) and rechecked the complete affected implementation against issue #361 rather than reviewing only the new test diff.

## What the final commit closes

- The wire codec now has deterministic truncation coverage across every prefix of a valid ticket, boundary/malformed length coverage, lifetime/nonce/ticket/extension cases, trailing-data rejection, and proof that undersized encoding leaves caller output unchanged.
- Client/server owned ticket-state construction and explicit cloning are exercised with allocation-failure sweeps.
- Server issuance allocation failures are tested for transcript/output atomicity, cleanup, and retry safety.
- Record mode now proves:
  - a maximum-size ticket crosses multiple encrypted records and is delivered exactly once;
  - a no-consumer client validates and drops the ticket while remaining usable;
  - callback-retained state is a deep clone that survives callback return;
  - consumer-side storage refusal is nonfatal and later application traffic still works.
- QUIC now exercises a maximum-size ticket through actual protected packets, reverse-order delivery, multiple non-contiguous drops, real loss/timeout recovery, retransmission, ACK-driven transmit-buffer release, exact ticket/nonce/PSK preservation, exactly-once callback delivery, and continued stream traffic afterward.

## Full implementation verdict

I found no remaining production-code blocker in:

- RFC 8446 NewSessionTicket encoding/decoding and extension semantics;
- RMS capture timing and SHA-256/SHA-384 derivation helpers;
- distinct per-nonce PSK derivation;
- callback ownership and best-effort storage behavior;
- explicit server issuance and failure atomicity;
- record-mode fragmentation and allocator wiring;
- QUIC application-CRYPTO buffering, multiple outstanding tickets, loss/PTO requeueing, out-of-order ACK handling, bounded range metadata, compaction, and zeroization;
- interaction with normal application data before and after ticket issuance.

The generated truncation fixture and boundary-oriented length matrix are sufficient to exercise the required parser behavior. A separately checked-in independent valid-wire fixture could still be useful later for #377 differential coverage, but I do not consider that a blocker for #361.

All test-bearing jobs currently reported by workflow run 838 are green across Linux, macOS, ARM, and appliance-profile integration. The overall workflow is still marked in progress because non-test audit/packaging work is finishing.

**Verdict: approve-equivalent / ready to merge after run 838 reports success.**

PR #478, opened 2026-07-20
tls: add stateless ticket protection

Jaruso:
## Summary
- add `src/tls/ticket_protection.zig` with the v1 TDTK envelope parser, stable AEAD ID mapping, provider-backed seal/open, and root export
- add immutable ticket-key snapshots, reloadable publication, refcounted retirement, key-window validation, durable nonce-lease high-water checks, and generation rollback semantics for failed reloads
- add the `ServerPskResolver` adapter for stateless ticket recovery using one captured `now_unix_ms`
- expand security coverage for deterministic fixtures, all supported AEADs, tamper/boundary cases, concurrent nonce uniqueness, observer outcomes, allocation failure, and plaintext zeroization

## Validation

Refs #363

### DIFF
--- a/src/tls/ticket_protection.zig
+++ b/src/tls/ticket_protection.zig
@@ -499,14 +499,16 @@ pub const ReloadableKeyRing = struct {
         };
 
         const retired = self.current;
+        const installed_generation = replacement.generation;
+        const retired_key_count = if (retired) |snapshot| retiredKeyCount(snapshot, replacement) else 0;
         self.current = replacement;
         self.next_generation = @max(self.next_generation, replacement.generation + 1);
         self.updateLedgerLocked(replacement);
         self.mutex.unlock();
 
-        self.record(.{ .snapshot_installed = replacement.generation });
+        self.record(.{ .snapshot_installed = installed_generation });
         if (retired) |snapshot| {
-            self.recordRetiredKeys(snapshot, replacement);
+            for (0..retired_key_count) |_| self.record(.{ .key_retired = snapshot.generation });
             snapshot.release();
         }
     }
@@ -592,10 +594,12 @@ pub const ReloadableKeyRing = struct {
         if (self.observer) |observer| observer.record(event);
     }
 
-    fn recordRetiredKeys(self: *ReloadableKeyRing, retired: *const Snapshot, replacement: *const Snapshot) void {
+    fn retiredKeyCount(retired: *const Snapshot, replacement: *const Snapshot) usize {
+        var count: usize = 0;
         for (retired.keys) |*old_key| {
-            if (replacement.findKeyConst(&old_key.id) == null) self.record(.{ .key_retired = retired.generation });
+            if (replacement.findKeyConst(&old_key.id) == null) count += 1;
         }
+        return count;
     }
 
     pub fn format(
@@ -1074,6 +1078,91 @@ const TestObserver = struct {
     }
 };
 
+const ReentrantDeinitObserver = struct {
+    keyring: *ReloadableKeyRing,
+    deinit_on_install: bool = true,
+    installed_events: usize = 0,
+    retired_events: usize = 0,
+
+    fn observer(self: *ReentrantDeinitObserver) Observer {
+        return .{ .ctx = self, .recordFn = record };
+    }
+
+    fn record(ctx: *anyopaque, event: Event) void {
+        const self: *ReentrantDeinitObserver = @ptrCast(@alignCast(ctx));
+        switch (event) {
+            .snapshot_installed => {
+                self.installed_events += 1;
+                if (self.deinit_on_install) {
+                    self.deinit_on_install = false;
+                    self.keyring.deinit();
+                }
+            },
+            .key_retired => self.retired_events += 1,
+            else => {},
+        }
+    }
+};
+
+fn installSingleKey(
+    keyring: *ReloadableKeyRing,
+    id: KeyId,
+    aead: provider.Aead,
+    lease: NonceLeaseConfig,
+    byte: u8,
+) !void {
+    const config = sampleKeyConfigWithByte(id, aead, lease, byte);
+    try keyring.install(try keyring.buildSnapshot(&.{config}, testCapabilities()));
+}
+
+fn expectResolveFalseUnchanged(
+    protector: *Protector,
+    allocator: std.mem.Allocator,
+    identity: []const u8,
+    now_unix_ms: i64,
+) !void {
+    var out: session.ServerRecoverableState = .{};
+    defer out.deinit();
+    try testing.expect(!try protector.resolve(allocator, identity, now_unix_ms, &out));
+    try testing.expectEqual(@as(u32, 0), out.ticket_age_add);
+    try testing.expectEqual(@as(usize, 0), out.common.resumption_psk.len);
+}
+
+fn expectRoundTrip(
+    protector: *Protector,
+    allocator: std.mem.Allocator,
+    state: *const session.ServerRecoverableState,
+    ticket: []const u8,
+    now_unix_ms: i64,
+) !void {
+    var recovered: session.ServerRecoverableState = .{};
+    defer recovered.deinit();
+    try testing.expect(try protector.resolve(allocator, ticket, now_unix_ms, &recovered));
+    try testing.expectEqual(state.ticket_age_add, recovered.ticket_age_add);
+    try testing.expectEqual(state.common.lifetime_seconds, recovered.common.lifetime_seconds);
+    try testing.expectEqualSlices(u8, state.common.resumption_psk.slice(), recovered.common.resumption_psk.slice());
+}
+
+const ConcurrentSealTask = struct {
+    protector: *Protector,
+    state: *const session.ServerRecoverableState,
+    start: *std.atomic.Value(bool),
+    successes: *std.atomic.Value(usize),
+    failures: *std.atomic.Value(usize),
+    nonces: *[32][provider.aead_nonce_len]u8,
+
+    fn run(self: *ConcurrentSealTask) void {
+        while (!self.start.load(.acquire)) std.atomic.spinLoopHint();
+        var ticket_buf = [_]u8{0} ** 512;
+        const ticket = self.protector.seal(testing.allocator, self.state, 2_000, &ticket_buf) catch {
+            _ = self.failures.fetchAdd(1, .monotonic);
+            return;
+        };
+        const slot = self.successes.fetchAdd(1, .monotonic);
+        if (slot < self.nonces.len) @memcpy(&self.nonces[slot], ticket[24..36]);
+    }
+};
+
 test "AEAD id mapping is stable and not enum ordinal dependent" {
     try testing.expectEqual(@as(u8, 1), encodeAeadId(.aes_128_gcm));
     try testing.expectEqual(@as(u8, 2), encodeAeadId(.aes_256_gcm));
@@ -1156,6 +1245,160 @@ test "seal and resolve round trip and authenticate header fields" {
     try testing.expect(!try protector.resolve(allocator, ticket, 2_000, &untouched));
 }
 
+test "all supported AEADs seal and resolve" {
+    const allocator = testing.allocator;
+    const cases = [_]struct {
+        aead: provider.Aead,
+        key_byte: u8,
+        prefix: [4]u8,
+    }{
+        .{ .aead = .aes_128_gcm, .key_byte = 0x11, .prefix = .{ 3, 1, 0, 0 } },
+        .{ .aead = .aes_256_gcm, .key_byte = 0x22, .prefix = .{ 3, 2, 0, 0 } },
+        .{ .aead = .chacha20_poly1305, .key_byte = 0x33, .prefix = .{ 3, 3, 0, 0 } },
+    };
+
+    for (cases, 0..) |case, i| {
+        var keyring = ReloadableKeyRing.init(allocator);
+        defer keyring.deinit();
+        try installSingleKey(&keyring, keyId(@intCast(30 + i)), case.aead, .{ .prefix = case.prefix, .start = 0, .end_exclusive = 3 }, case.key_byte);
+
+        var state = try sampleServerState(allocator);
+        defer state.deinit();
+        var protector = Protector{ .provider = testProvider(), .keyring = &keyring, .limits = session.Limits.default };
+        var ticket_buf = [_]u8{0} ** 512;
+        const ticket = try protector.seal(allocator, &state, 2_000, &ticket_buf);
+        try testing.expectEqual(case.aead, (try parseEnvelope(ticket, session.Limits.default)).aead);
+        try expectRoundTrip(&protector, allocator, &state, ticket, 2_000);
+    }
+}
+
+test "deterministic envelope fixture stays stable" {
+    const allocator = testing.allocator;
+    var keyring = ReloadableKeyRing.init(allocator);
+    defer keyring.deinit();
+    try installSingleKey(&keyring, keyId(31), .aes_128_gcm, .{ .prefix = .{ 3, 1, 0, 1 }, .start = 7, .end_exclusive = 8 }, 0x11);
+
+    var state = try sampleServerState(allocator);
+    defer state.deinit();
+    var protector = Protector{ .provider = testProvider(), .keyring = &keyring, .limits = session.Limits.default };
+    var ticket_buf = [_]u8{0} ** 512;
+    const ticket = try protector.seal(allocator, &state, 2_000, &ticket_buf);
+    const expected = [_]u8{
+        0x54, 0x44, 0x54, 0x4B, 0x01, 0x01, 0x00, 0x00, 0x1F, 0x1F, 0x1F, 0x1F, 0x1F, 0x1F, 0x1F, 0x1F,
+        0x1F, 0x1F, 0x1F, 0x1F, 0x1F, 0x1F, 0x1F, 0x1F, 0x03, 0x01, 0x00, 0x01, 0x00, 0x00, 0x00, 0x00,
+        0x00, 0x00, 0x00, 0x07, 0x66, 0x7F, 0x02, 0x37, 0x11, 0xE4, 0xF3, 0x9C, 0xF8, 0x63, 0x20, 0xB4,
+        0xEA, 0x5C, 0xE7, 0xAC, 0xEF, 0xFC, 0xB6, 0x6D, 0x78, 0xA8, 0xDC, 0x56, 0x81, 0x50, 0x21, 0x6E,
+        0x51, 0xB8, 0x54, 0xA5, 0x0D, 0x64, 0x75, 0x9A, 0x5A, 0x3A, 0x90, 0x88, 0x65, 0x6B, 0xEB, 0xAC,
+        0x49, 0x78, 0x2F, 0x6A, 0xE1, 0xD9, 0x78, 0xF2, 0x70, 0x7D, 0xB7, 0x9C, 0x28, 0x45, 0xD5, 0x94,
+        0xEB, 0x57, 0x24, 0x63, 0x18, 0xDB, 0x52, 0x48, 0xD1, 0x07, 0x0E, 0x53, 0x44, 0xE3, 0xA6, 0x74,
+        0x11, 0xC9, 0xCF, 0x11, 0x31, 0x52, 0x06, 0xC8, 0x8D, 0x41, 0x37, 0xB6, 0xC0, 0x22, 0x72, 0x8D,
+        0x80, 0x7D, 0xCE, 0x34, 0x30, 0x26, 0x12, 0x91, 0x0C, 0x23, 0x8F, 0xAA, 0xAB, 0x23, 0x8B, 0xEF,
+        0x80, 0x65, 0x4F, 0xCD, 0x12, 0x20, 0x02, 0xB6, 0xF1, 0x3B, 0x75, 0xFC, 0x87, 0x7C, 0xF4, 0x4B,
+        0xF7, 0x9E, 0x2E, 0x69, 0x68, 0x1B, 0xF5, 0x20, 0x0C, 0x57, 0xA2, 0xC5, 0x94, 0x17, 0x53, 0x1D,
+        0x48, 0xE9, 0x2B, 0x3A, 0xE0, 0xE6, 0x3A, 0x2F, 0xAF, 0x1D, 0xB7, 0xD8, 0x8E, 0xAC, 0x6B, 0xE0,
+        0x6B, 0x08, 0x15, 0xE1, 0xB6, 0xED, 0x57, 0x04, 0x8A, 0x76, 0xD9, 0x3D, 0xCE, 0xB2, 0x18, 0x05,
+        0x37, 0x5C, 0x7B, 0x25,
+    };
+    try testing.expectEqualSlices(u8, &expected, ticket);
+}
+
+test "ticket identity tampering is unusable and leaves output untouched" {
+    const allocator = testing.allocator;
+    var keyring = ReloadableKeyRing.init(allocator);
+    defer keyring.deinit();
+    try installSingleKey(&keyring, keyId(32), .aes_128_gcm, .{ .prefix = .{ 3, 2, 0, 0 }, .start = 0, .end_exclusive = 12 }, 0x11);
+
+    var state = try sampleServerState(allocator);
+    defer state.deinit();
+    var protector = Protector{ .provider = testProvider(), .keyring = &keyring, .limits = session.Limits.default };
+    var ticket_buf = [_]u8{0} ** 512;
+    const ticket = try protector.seal(allocator, &state, 2_000, &ticket_buf);
+
+    const positions = [_]usize{ 0, 4, 5, 6, 8, 24, fixed_header_len, ticket.len - 1 };
+    for (positions) |pos| {
+        var mutated = ticket_buf;
+        mutated[pos] ^= 0x01;
+        try expectResolveFalseUnchanged(&protector, allocator, mutated[0..ticket.len], 2_000);
+    }
+}
+
+test "decrypt windows are enforced at exact boundaries" {
+    const allocator = testing.allocator;
+    var keyring = ReloadableKeyRing.init(allocator);
+    defer keyring.deinit();
+    try installSingleKey(&keyring, keyId(33), .aes_128_gcm, .{ .prefix = .{ 3, 3, 0, 0 }, .start = 0, .end_exclusive = 3 }, 0x11);
+
+    var state = try sampleServerState(allocator);
+    defer state.deinit();
+    var protector = Protector{ .provider = testProvider(), .keyring = &keyring, .limits = session.Limits.default };
+    var ticket_buf = [_]u8{0} ** 512;
+    const ticket = try protector.seal(allocator, &state, 2_000, &ticket_buf);
+
+    try expectResolveFalseUnchanged(&protector, allocator, ticket, 999);
+    try expectRoundTrip(&protector, allocator, &state, ticket, 1_000);
+    try expectRoundTrip(&protector, allocator, &state, ticket, 4_999);
+    try expectRoundTrip(&protector, allocator, &state, ticket, 5_000);
+    try expectResolveFalseUnchanged(&protector, allocator, ticket, 20_000);
+}
+
+test "ticket identity size boundaries reject malformed envelopes" {
+    const allocator = testing.allocator;
+    var keyring = ReloadableKeyRing.init(allocator);
+    defer keyring.deinit();
+    try installSingleKey(&keyring, keyId(34), .aes_128_gcm, .{ .prefix = .{ 3, 4, 0, 0 }, .start = 0, .end_exclusive = 3 }, 0x11);
+
+    var state = try sampleServerState(allocator);
+    defer state.deinit();
+    var protector = Protector{ .provider = testProvider(), .keyring = &keyring, .limits = session.Limits.default };
+    var ticket_buf = [_]u8{0} ** 512;
+    const ticket = try protector.seal(allocator, &state, 2_000, &ticket_buf);
+    try expectResolveFalseUnchanged(&protector, allocator, ticket[0 .. ticket.len - 1], 2_000);
+
+    var min_identity = [_]u8{0} ** (fixed_header_len + 1 + tag_len);
+    writeHeader(min_identity[0..fixed_header_len], .aes_128_gcm, &keyId(34), &([_]u8{0} ** provider.aead_nonce_len));
+    try expectResolveFalseUnchanged(&protector, allocator, &min_identity, 2_000);
+
+    var trailing = [_]u8{0} ** 513;
+    @memcpy(trailing[0..ticket.len], ticket);
+    trailing[ticket.len] = 0;
+    try expectResolveFalseUnchanged(&protector, allocator, trailing[0 .. ticket.len + 1], 2_000);
+
+    var limits = session.Limits.default;
+    limits.max_ticket_len = ticket.len - 1;
+    try testing.expectError(error.EnvelopeTooLarge, parseEnvelope(ticket, limits));
+}
+
+test "concurrent sealing reserves unique nonces over one lease" {
+    const allocator = testing.allocator;
+    var keyring = ReloadableKeyRing.init(allocator);
+    defer keyring.deinit();
+    try installSingleKey(&keyring, keyId(35), .aes_128_gcm, .{ .prefix = .{ 3, 5, 0, 0 }, .start = 0, .end_exclusive = 32 }, 0x11);
+
+    var state = try sampleServerState(allocator);
+    defer state.deinit();
+    var protector = Protector{ .provider = testProvider(), .keyring = &keyring, .limits = session.Limits.default };
+    var start = std.atomic.Value(bool).init(false);
+    var successes = std.atomic.Value(usize).init(0);
+    var failures = std.atomic.Value(usize).init(0);
+    var nonces: [32][provider.aead_nonce_len]u8 = undefined;
+    var tasks: [40]ConcurrentSealTask = undefined;
+    var threads: [40]std.Thread = undefined;
+
+    for (&tasks, 0..) |*task, i| {
+        task.* = .{ .protector = &protector, .state = &state, .start = &start, .successes = &successes, .failures = &failures, .nonces = &nonces };
+        threads[i] = try std.Thread.spawn(.{}, ConcurrentSealTask.run, .{task});
+    }
+    start.store(true, .release);
+    for (&threads) |thread| thread.join();
+
+    try testing.expectEqual(@as(usize, 32), successes.load(.monotonic));
+    try testing.expectEqual(@as(usize, 8), failures.load(.monotonic));
+    for (nonces[0..], 0..) |nonce, i| {
+        try testing.expectEqualSlices(u8, &[_]u8{ 3, 5, 0, 0 }, nonce[0..4]);
+        for (nonces[0..i]) |prior| try testing.expect(!std.mem.eql(u8, &nonce, &prior));
+    }
+}
+
 test "nonce lease exhaustion fails closed without wraparound" {
     const allocator = testing.allocator;
     var keyring = ReloadableKeyRing.init(allocator);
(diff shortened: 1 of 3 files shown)


src/tls/ticket_protection.zig
```diff
@@ -0,0 +1,1001 @@
...
+
+        for (configs, 0..) |config, i| {
+            for (configs[0..i]) |prior| {
+                if (std.mem.eql(u8, &prior.id, &config.id)) return error.DuplicateKeyId;
+            }
+            keys[i] = try KeyRecord.build(config, caps);
+            initialized += 1;
+        }
+
+        snapshot.keys = keys;
```
Jaruso, line 342: `SnapshotError.AmbiguousEncryptionWindow` is currently unreachable. Please reject a replacement during construction/publication when two keys with nonce leases have overlapping encryption intervals (`[not_before_unix_ms, encrypt_until_unix_ms)`). Deferring this to `activeEncryptionKey()` means an invalid snapshot installs successfully and then every issuance during the overlap fails at runtime.

A bounded pairwise validation is sufficient (`max_keys == 16`):

```zig
for (keys, 0..) |*a, i| {
    if (a.nonce_lease == null) continue;
    for (keys[i + 1 ..]) |*b| {
        if (b.nonce_lease == null) continue;
        if (a.not_before_unix_ms < b.encrypt_until_unix_ms and
            b.not_before_unix_ms < a.encrypt_until_unix_ms)
            return error.AmbiguousEncryptionWindow;
    }
}
```

Add tests for exact-adjacent windows (accepted), one-millisecond overlap (rejected), and a future key whose activation overlaps the current key’s encryption window.

src/tls/ticket_protection.zig
```diff
@@ -0,0 +1,1001 @@
...
+            if (self.current) |current| {
+                if (current.findKey(&key.id)) |old| {
+                    if (old.aead != key.aead or !old.key.eql(&key.key)) return error.DuplicateKeyId;
+                }
+            }
+        }
+    }
+
+    fn rebuildLedgerLocked(self: *ReloadableKeyRing) void {
+        self.ledger_len = 0;
```
Jaruso: Resetting `ledger_len` here discards the nonce high-water history that the ledger is meant to protect. A concrete reuse path is:

```text
K, prefix P, lease [0,10) active
K retained decrypt-only (nonce_lease = null) -> ledger becomes empty
K re-enabled with P/[0,10) -> validation accepts reused nonces
```

Removing K for one generation and later reintroducing it has the same problem. Preserve the previous high-water entry for every still-known key even when the new snapshot makes it decrypt-only, update it only after successful installation, and define a tombstone/durable-history rule for removed IDs. Also reject the same `(aead, key bytes)` under a different key ID; nonce uniqueness belongs to the AEAD key material, not the public routing ID.

Please add active→decrypt-only→re-enable, removal→reintroduction, changed-ID/same-key, exact-adjacent range, prefix change, and failed-install rollback tests.

src/tls/ticket_protection.zig
```diff
@@ -0,0 +1,1001 @@
...
+                self.record(.{ .resolve_rejected = .authentication_failed });
+                return false;
+            },
+            error.UnsupportedCapability => return error.UnsupportedCapability,
+            error.InvalidInput => return error.InvalidInternalState,
+        };
+
+        var decoded = session.decode(allocator, self.limits, plaintext) catch {
+            self.record(.{ .resolve_rejected = .invalid_plaintext });
+            return false;
```
Jaruso: This catch-all incorrectly converts `session.decode`'s `error.OutOfMemory` into an ordinary resolver miss. That violates the resolver contract: malformed authenticated plaintext is a miss, but allocator/local configuration failures are fatal typed errors.

Please split the error set explicitly, for example:

```zig
var decoded = session.decode(allocator, self.limits, plaintext) catch |err| switch (err) {
    error.OutOfMemory => return error.OutOfMemory,
    error.InvalidLimits => return error.InvalidInternalState,
    else => {
        self.record(.{ .resolve_rejected = .invalid_plaintext });
        return false;
    },
};
```

Add an allocation-failure test that reaches owned-state construction after successful authentication and proves the resolver returns `OutOfMemory`, leaves `out` unchanged, and wipes/frees the plaintext buffer.

src/tls/ticket_protection.zig
```diff
@@ -0,0 +1,1001 @@
...
+        2 => .aes_256_gcm,
+        3 => .chacha20_poly1305,
+        else => error.UnsupportedAeadId,
+    };
+}
+
+pub fn parseEnvelope(identity: []const u8, limits: session.Limits) ParseError!ParsedEnvelope {
+    limits.validate() catch return error.EnvelopeTooLarge;
+    if (identity.len > limits.max_ticket_len or identity.len > session.absolute_ticket_wire_max)
+        return error.EnvelopeTooLarge;
```
Jaruso, line 180: Invalid `session.Limits` are local configuration/invariant failures, not an attacker-supplied oversized envelope. With this mapping, `Protector.resolve` turns a broken local configuration into `false`, which can silently trigger later-identity probing or full-handshake fallback.

Please either add `InvalidLimits` to the parser’s error contract and map it to `ResolveError.InvalidInternalState`, or require prevalidated limits and perform `self.limits.validate() catch return error.InvalidInternalState` before calling the allocation-free parser. Keep `EnvelopeTooLarge` only for a valid configured bound exceeded by peer bytes.

Jaruso (commented):
**Review result: changes required before merge.**

The module boundary and envelope shape are headed in the right direction, and the current CI run is green. There are security/correctness gaps that make the keyring unsafe to merge even as a first slice.

### Blocking findings

1. **Nonce-reuse protection is not durable across snapshot transitions.** `rebuildLedgerLocked` clears the high-water ledger on every install. The most direct failure sequence is: active key with lease `[0, 10)` → same key retained decrypt-only with no lease → same key re-enabled with `[0, 10)`. The second lease is accepted because the high-water record was discarded. A removed key ID can also be reintroduced later, and the current validation does not prevent the same AEAD key bytes from being assigned a different key ID. AES-GCM/ChaCha nonce uniqueness is a property of the key material, not the routing ID.

2. **Authenticated-plaintext allocation failures are downgraded to ordinary identity misses.** `session.decode` includes `OutOfMemory`; the catch-all currently returns `false`. Local allocator/configuration failures must remain fatal typed resolver errors so the server does not silently continue/fallback under resource failure.

3. **Ambiguous active encryption windows are accepted.** `AmbiguousEncryptionWindow` exists in the error taxonomy but is never produced. Two lease-bearing keys with overlapping `[not_before, encrypt_until)` windows install successfully and make every seal fail at runtime. Snapshot publication is supposed to reject the replacement atomically.

4. **Invalid local limits are reported as peer-controlled malformed/oversized identities.** `parseEnvelope` maps `Limits.validate()` failure to `EnvelopeTooLarge`, so `resolve` returns a normal miss for broken local configuration. Validate protector limits as a local invariant and return `InvalidInternalState` (or add an explicit local error) before parsing peer bytes.

### Prerequisite/schema blocker

The current base still has `ServerRecoverableState { common }` only; its server codec does not include `ticket_age_add`. #363 explicitly requires the encrypted plaintext to contain the exact per-ticket `ticket_age_add`, and #361 was supposed to land that prerequisite before the v1 protected-state contract. Please add the missing #361 follow-up here or rebase this PR after it lands, then add a round-trip assertion for it. Otherwise a stateless server cannot recover the value required to validate `obfuscated_ticket_age`.

### Coverage required before this is merge-ready

The six focused tests do not yet establish the ticket-protection contract. At minimum this slice needs:

- deterministic byte fixtures and seal/open round trips for AES-128-GCM, AES-256-GCM, and ChaCha20-Poly1305;
- exact header/AAD/ciphertext/tag mutation coverage (the current key-ID mutation exits at lookup and does not prove that header substitution fails authentication);
- unknown/future/retained/retired key cases and exact window boundaries;
- concurrent nonce reservation, exact lease exhaustion, active→decrypt-only→re-enabled lease history, failed-install rollback, and duplicate key material under another ID;
- allocation-failure sweeps proving decode OOM is fatal and plaintext/key storage is cleared;
- snapshot retirement with an in-flight retained reference and zeroization after final release;
- ambiguous-window rejection and exact adjacent-window acceptance;
- invalid-limit/local-error behavior and observer event coverage.

The PR description says “Draft because …”, but GitHub currently reports this as a non-draft PR. If the resolver adapter and remaining #363 acceptance criteria are intentionally deferred, please mark it draft or clearly define the merge-sized slice and follow-up before treating the issue as complete. The security blockers above still need correction before this core slice should merge.

src/tls/ticket_protection.zig
Jaruso: The original nonce-reuse paths are now covered: history survives decrypt-only/removal transitions, failed installs do not advance it, and same key material under a different ID is rejected. I’m leaving this thread unresolved because the replacement introduces two new lifecycle problems: each history entry retains a full `TicketKeySecret` after the retired snapshot’s last reference is released, and the fixed 64-entry history makes normal rotation permanently fail after 64 distinct keys. I’ve added targeted inline findings on those points.

src/tls/ticket_protection.zig
```diff
@@ -0,0 +1,1158 @@
...
+pub const magic = [4]u8{ 'T', 'D', 'T', 'K' };
+pub const format_version: u8 = 1;
+pub const key_id_len = 16;
+pub const fixed_header_len = 36;
+pub const tag_len = provider.aead_tag_len;
+pub const envelope_overhead = fixed_header_len + tag_len;
+pub const max_keys = 16;
+pub const max_lease_history = max_keys * 4;
+
+const aad_prefix = "tardigrade/tls-ticket/v1\x00";
```
Jaruso, line 24: This fixed append-only history makes rotation fail permanently after 64 distinct key IDs. `updateLedgerLocked` never removes entries, and `validateReplacementLocked` returns `TooManyKeys` when a new ID arrives at capacity, so even a deployment rotating one key at a time bricks on rotation 65.

`max_keys == 16` is a live-snapshot bound; it must not become a lifetime rotation bound. Use allocator-backed/non-secret history or an explicit durable history object supplied by the runtime, and add a test that rotates through more than 64 unique keys. Do not solve this by silently pruning entries if an old key ID/material can later be reintroduced.

src/tls/ticket_protection.zig
```diff
@@ -0,0 +1,1158 @@
...
+    }
+};
+
+const LeaseHighWater = struct {
+    key_id: KeyId,
+    prefix: [4]u8,
+    end_exclusive: u64,
+    has_lease: bool,
+    aead: provider.Aead,
+    key: TicketKeySecret,
```
Jaruso: Copying the full `TicketKeySecret` into the high-water ledger keeps retired AEAD key material alive until the entire keyring is deinitialized. That violates #363's lifecycle requirement that a retired key be wiped when the final snapshot reference is released; the snapshot can retire correctly while this hidden copy remains indefinitely.

The history only needs a non-secret identity for duplicate-material detection. Store a domain-separated collision-resistant fingerprint (including the AEAD identifier and exact logical key bytes), or another non-reversible lineage token appropriate to the runtime contract, and wipe temporary hashing input/output as required. Add a lifecycle test proving the retired snapshot's last release eliminates all copies of the actual key bytes.

src/tls/ticket_protection.zig
```diff
@@ -0,0 +1,1158 @@
...
+    limits: session.Limits,
+    observer: ?Observer = null,
+
+    pub fn protectedLen(
+        self: *const Protector,
+        state: *const session.ServerRecoverableState,
+    ) SealError!usize {
+        const plaintext_len = session.serverEncodedLenWithLimits(state, self.limits) catch |err| switch (err) {
+            error.StateTooLarge => return error.SerializedStateTooLarge,
+            error.InvalidLimits, error.TooManyFields, error.FieldTooLarge, error.InvalidState => return error.SerializedStateTooLarge,
```
Jaruso: `InvalidLimits` is still misclassified on the issuance path. `protectedLen` maps broken local `session.Limits` to `SerializedStateTooLarge`, even though `SealError.InvalidInternalState` exists and `encodeServer` later maps the same condition correctly.

Validate `self.limits` before length calculation, or split this switch so `error.InvalidLimits => error.InvalidInternalState`. The failure must occur before reserving a nonce. Add a seal/protectedLen test with invalid limits that asserts `InvalidInternalState` and proves the next successful seal still receives the first lease counter.

Jaruso (commented):
**Follow-up review: the original four inline findings are substantially fixed, but this is still not merge-ready.**

Verified fixes in the new commit:

- overlapping encryption windows are rejected before publication, with adjacent/overlap/future-window tests;
- resolver limits are validated as a local invariant;
- authenticated-state decode preserves `OutOfMemory` as a typed operational error;
- nonce history now survives decrypt-only and removed-snapshot transitions and rejects the same AEAD key under another ID;
- `ticket_age_add` is carried through the server-state codec and protected round trip.

The revised history implementation introduces two blocking lifecycle problems, called out inline: it retains every retired ticket key in full until keyring teardown, and the fixed append-only history permanently prevents rotation after 64 distinct keys. There is also a remaining seal-path error-classification bug for invalid limits.

There is now direct branch overlap with PR #479: both PRs modify `session.zig` and `new_session_ticket.zig` to add `ticket_age_add`, but they choose different initializer APIs. #479 should merge first; then this branch should rebase, drop those duplicated prerequisite edits, consume the final state API, and add the thin `ServerPskResolver` adapter using one captured `now_unix_ms` for both ticket recovery and compatibility evaluation.

The PR remains correctly marked draft. Before it is ready, the #363 acceptance suite still needs the resolver adapter, deterministic seal/open coverage for all three AEADs, complete public-header/authentication mutations, concurrent nonce reservation, failed-reload/snapshot-retirement lifecycle coverage, observer assertions, and broader allocation-failure/zeroization checks. CI for this head is still in progress at the time of review.

src/tls/ticket_protection.zig
```diff
@@ -0,0 +1,1242 @@
...
+            .prefix = if (record.nonce_lease) |*value| value.prefix else [_]u8{0} ** 4,
+            .end_exclusive = if (record.nonce_lease) |*value| value.currentEnd() else 0,
+            .has_lease = record.nonce_lease != null,
+            .aead = record.aead,
+            .key_fingerprint = fingerprintKey(record.aead, record.key.slice()),
+        };
+    }
+};
+
+pub const ReloadableKeyRing = struct {
```
Jaruso, line 421: Replacing the ledger's copied `TicketKeySecret` with a plain fingerprint fixes secret retention, but it also removes the nested secret type's compile-time formatting guard. `ReloadableKeyRing` is public and now structurally contains `key_id`, nonce `prefix`, `end_exclusive`, AEAD, and `key_fingerprint`; ordinary `{any}`-style formatting can expose all of that despite #363's requirement not to log key IDs or nonce ranges.

Add an explicit guard on the public aggregate, matching `Snapshot`/`KeyRecord`:

```zig
pub fn format(
    _: ReloadableKeyRing,
    comptime _: []const u8,
    _: std.fmt.FormatOptions,
    _: anytype,
) !void {
    @compileError("ticket keyrings must not be formatted or logged");
}
```

Also extend the non-formatting declaration test to assert `@hasDecl(ReloadableKeyRing, "format")`. A guard on `LeaseHighWater` as defense in depth is reasonable too, but the public keyring guard is the important boundary.

Jaruso (commented):
**Follow-up review: one new hardening change remains.**

The latest commit correctly resolves every previously reported implementation blocker:

- nonce history is allocator-backed rather than capped at 64 entries;
- retired key bytes are replaced with a domain-separated SHA-256 fingerprint;
- history allocation occurs before publication and failed allocation leaves the prior snapshot active;
- issuance-limit failures are classified as `InvalidInternalState` before nonce reservation;
- the targeted regression tests cover >64 rotations, key-byte cleanup, and nonce preservation;
- CI run 843 is green.

I found one security regression introduced by removing `TicketKeySecret` from the ledger: `ReloadableKeyRing` no longer contains a nested type whose `format` implementation rejects logging, so ordinary structural formatting can now expose the ledger's key IDs, nonce prefixes/high-water values, algorithms, and key fingerprints. Add an explicit non-formatting guard to the public keyring type and a compile-time coverage assertion.

After that, I do not see another defect in this hardening commit. The PR should remain draft until #479 lands and this branch is rebased to consume its final `ServerRecoverableState`/`ServerPskResolver` APIs, remove the duplicated `session.zig` and `new_session_ticket.zig` amendment, and add the stateless resolver adapter. Before marking #363 complete, the broader acceptance suite should also cover all three AEADs, concurrent nonce reservation, deterministic fixtures/mutations, observer outcomes, and reload/reference lifecycle paths.

Jaruso (commented):
Follow-up review of `0e936f5`: the requested non-formatting boundary is implemented correctly. `ReloadableKeyRing` now has the same compile-time `format` rejection as `KeyRecord` and `Snapshot`, and the declaration test covers all three secret-bearing aggregates. I found no additional defect in this incremental change, so the final open review thread is resolved.

The previously reviewed stateless-protector core is now clear of my outstanding code findings. This PR should remain draft until #479 merges and the branch is rebased to remove the duplicated `session.zig` / `new_session_ticket.zig` changes, consume the canonical #362 APIs, and add the `ServerPskResolver` adapter. CI run 844 is still in progress at the time of this review; completed format, unit-test, security, and several platform jobs are green.

src/tls/ticket_protection.zig
```diff
@@ -0,0 +1,1320 @@
...
+    }
+
+    pub fn buildSnapshot(self: *ReloadableKeyRing, configs: []const KeyConfig, caps: provider.Capabilities) SnapshotError!*Snapshot {
+        self.mutex.lock();
+        if (self.next_generation == std.math.maxInt(u64)) {
+            self.mutex.unlock();
+            return error.GenerationOverflow;
+        }
+        const generation = self.next_generation;
+        self.next_generation += 1;
```
Jaruso: `buildSnapshot` advances `next_generation` before the replacement has successfully built. Therefore any invalid config or allocator failure changes publication state even though no snapshot exists:

```text
next_generation = 7
Snapshot.build(...) -> OutOfMemory
next_generation = 8
```

#363 explicitly requires a failed reload not to partially change generation, and the observer seam defines typed `snapshot_rejected` reasons for these build failures, but this path records nothing.

Build and validate first with a provisional/unpublished generation, then assign and increment the real generation under the mutex only after construction succeeds. For example, build with `generation = 0`, then:

```zig
const replacement = Snapshot.build(self.allocator, configs, 0, caps) catch |err| {
    self.record(.{ .snapshot_rejected = snapshotReason(err) });
    return err;
};
errdefer replacement.release();

self.mutex.lock();
defer self.mutex.unlock();
if (self.next_generation == std.math.maxInt(u64))
    return error.GenerationOverflow;
replacement.generation = self.next_generation;
self.next_generation += 1;
return replacement;
```

Handle/observe the overflow path before returning and add tests showing invalid-config and allocation failures leave `next_generation`, `current`, and the nonce ledger unchanged.

src/tls/ticket_protection.zig
```diff
@@ -0,0 +1,1320 @@
...
+        if (self.current) |current| {
+            if (current == replacement) {
+                if (replacement.generation == std.math.maxInt(u64)) {
+                    self.mutex.unlock();
+                    self.record(.{ .snapshot_rejected = .generation_overflow });
+                    return error.GenerationOverflow;
+                }
+                self.next_generation = @max(self.next_generation, replacement.generation + 1);
+                self.mutex.unlock();
+                return;
```
Jaruso, line 467: This idempotent same-pointer branch leaks the caller's retained snapshot reference. The only legitimate way to obtain `current` is `acquireCurrent()`, which increments the refcount; `install` consumes/releases `replacement` on every other success/failure path, but here it returns without dropping that extra reference:

```zig
const current = keyring.acquireCurrent().?; // refcount 2
try keyring.install(current);               // still refcount 2
keyring.deinit();                           // drops only ring ref; snapshot/key survives
```

That prevents the final `Snapshot.deinit` and ticket-key wipe. Either remove the special case and reject/release it as stale, or explicitly consume the acquired reference before returning:

```zig
self.mutex.unlock();
replacement.release();
return;
```

Document that `install` consumes its argument and add a `deinit_count`/secret-wipe test for reinstalling `acquireCurrent().?`.

src/tls/ticket_protection.zig
```diff
@@ -0,0 +1,1320 @@
...
+
+    pub fn protectedLen(
+        self: *const Protector,
+        state: *const session.ServerRecoverableState,
+    ) SealError!usize {
+        self.limits.validate() catch return error.InvalidInternalState;
+        const plaintext_len = session.serverEncodedLenWithLimits(state, self.limits) catch |err| switch (err) {
+            error.StateTooLarge => return error.SerializedStateTooLarge,
+            error.InvalidLimits => return error.InvalidInternalState,
+            error.TooManyFields, error.FieldTooLarge, error.InvalidState => return error.SerializedStateTooLarge,
```
Jaruso: `error.InvalidState` is a semantic/internal model failure, not a size rejection. `session.serverEncodedLenWithLimits` returns it for conditions such as a zero/overlong lifetime, wrong PSK length, non-canonical SNI, or corrupted bounded-length metadata. Mapping it to `SerializedStateTooLarge` hides local corruption and makes the `InvalidInternalState` branch effectively unreachable for these cases because `protectedLen` runs before `encodeServer`.

Split this arm:

```zig
error.InvalidState => return error.InvalidInternalState,
error.TooManyFields, error.FieldTooLarge => return error.SerializedStateTooLarge,
```

Add a test using a semantically invalid state (for example `lifetime_seconds = 0`) that expects `InvalidInternalState` from both `protectedLen` and `seal` and proves the first nonce remains unconsumed.

src/tls/ticket_protection.zig
```diff
@@ -0,0 +1,1320 @@
...
+        }
+
+        if (!ticketExpiresWithinKey(state, key.decrypt_until_unix_ms)) {
+            self.record(.{ .seal_rejected = .ticket_outlives_key });
+            return error.TicketOutlivesKey;
+        }
+
+        const protected_len = try self.protectedLen(state);
+        if (out.len < protected_len) {
+            self.record(.{ .seal_rejected = .output_too_small });
```
Jaruso: The typed observer does not see failures propagated from this call. `protectedLen` can return `SerializedStateTooLarge`, `TicketTooLarge`, or `InvalidInternalState`, but `try` bypasses `seal_rejected`. The same omission occurs for `encodeServer` failures and both errors from `aeadSeal`; on resolve, invalid limits, allocator failure, provider failure, and decode-local errors similarly return without recording the available `unsupported_capability` / `invalid_internal_state` outcomes. `buildSnapshot` failures are also unobserved.

This makes #365 metrics depend on which exact branch produced the same public error. Centralize terminal reporting so every seal attempt emits exactly one `seal_succeeded` or `seal_rejected`, and every resolvable false/error path has a defined event policy. A wrapper is less error-prone than scattered calls:

```zig
pub fn seal(...) SealError![]const u8 {
    return self.sealInner(...) catch |err| {
        if (err == error.NonceLeaseExhausted) self.record(.nonce_lease_exhausted);
        self.record(.{ .seal_rejected = sealReason(err) });
        return err;
    };
}
```

Remove duplicate branch-level terminal events and add table-driven observer tests asserting exactly one terminal event and the correct reason for every typed error. Also correct `key_retired`: it currently fires once per retired *snapshot generation*, even when no key was removed or several keys were removed, so it is not presently a key-retirement metric.

src/tls/ticket_protection.zig
```diff
@@ -0,0 +1,1320 @@
...
+    const expires: i128 = @as(i128, state.common.issued_at_unix_ms) +
+        @as(i128, state.common.lifetime_seconds) * 1000;
+    return expires <= @as(i128, key_decrypt_until_unix_ms);
+}
+
+fn fingerprintKey(aead: provider.Aead, key: []const u8) KeyFingerprint {
+    var out: KeyFingerprint = undefined;
+    var hasher = std.crypto.hash.sha2.Sha256.init(.{});
+    hasher.update(key_fingerprint_domain);
+    hasher.update(&[_]u8{encodeAeadId(aead)});
```
Jaruso, line 885: The full ticket key is fed into the SHA-256 context, but the context is never explicitly wiped. For these short inputs the hasher's partial-block buffer can contain the domain plus raw 16/32-byte key material after `final`, leaving another transient key copy on the stack even though the allocator-backed `TicketKeySecret` is later zeroized.

Wipe the complete hasher state on every exit:

```zig
var hasher = std.crypto.hash.sha2.Sha256.init(.{});
defer secrets.secureZero(std.mem.asBytes(&hasher));
```

The fingerprint output is intentionally retained and need not be zeroed; the secret-bearing hash context does. The existing fixed-buffer lifecycle test cannot detect this because it only scans allocator backing, not stack state.

Jaruso (commented):
**Post-#479 review result: changes are still required before merge.**

The dependency integration is correct: the PR is rebased onto merged #479, the duplicated `session.zig` / `new_session_ticket.zig` implementation is gone from the PR diff, `ServerRecoverableState.init` now uses the merged API, and the new adapter returns the canonical `pre_shared_key.ServerPskResolver`. The adapter also uses the same captured `now_unix_ms` for `nowUnixMsFn` and stateless recovery, so the pre-merge clock mismatch is resolved. CI run 856 is green.

Reviewing the complete now-ready implementation against #363 uncovered five remaining blockers:

1. A failed `buildSnapshot` advances `next_generation` and emits no `snapshot_rejected` event, violating the failed-reload atomicity and typed-observation contracts.
2. Reinstalling an acquired current snapshot leaks its extra reference, preventing final key zeroization.
3. Semantically invalid server state is misreported as `SerializedStateTooLarge` instead of `InvalidInternalState`.
4. The observer seam does not report every terminal seal/resolve/build outcome despite defining reason variants for them.
5. `fingerprintKey` does not wipe the SHA-256 state that directly absorbs the ticket key; its internal partial-block buffer may retain raw key bytes on the stack.

The focused tests added so far are useful, but #363's exhaustive/concurrency acceptance coverage is also still incomplete. Before approval, add tests for concurrent nonce uniqueness/exhaustion, an in-flight snapshot surviving replacement until its last release, failed build/install preserving current state and generation, all supported AEAD round trips/tamper cases, and exactly-one observer outcome on each success/rejection/error path.

The PR description should also be updated: it still says the PR is draft because #362 is absent, although #479 is merged and this PR is now marked ready.

src/tls/ticket_protection.zig
```diff
@@ -0,0 +1,1541 @@
...
+
+        const retired = self.current;
+        self.current = replacement;
+        self.next_generation = @max(self.next_generation, replacement.generation + 1);
+        self.updateLedgerLocked(replacement);
+        self.mutex.unlock();
+
+        self.record(.{ .snapshot_installed = replacement.generation });
+        if (retired) |snapshot| {
+            self.recordRetiredKeys(snapshot, replacement);
```
Jaruso: This dereferences `replacement` after `snapshot_installed` has invoked an arbitrary caller-owned observer. The callback is outside the mutex and can re-enter the keyring:

```text
outer install publishes replacement B
snapshot_installed(B) callback calls install(C) or deinit()
inner operation retires/releases B; B reaches refcount 0
outer install resumes and recordRetiredKeys(old, B) dereferences freed B
```

There is a second form inside `recordRetiredKeys`: after its first `key_retired` callback, a reentrant observer can release `replacement` before the loop evaluates the next old key.

Do not carry snapshot pointers across observer calls unless you hold explicit references. The simplest fix is to compute a scalar `retired_key_count` while both snapshots are known-live (under the mutex or immediately before the first callback), copy the installed generation, then emit `snapshot_installed` and that many `key_retired` events using only scalars. Alternatively retain/release `replacement` around the complete callback sequence. Add a reentrant observer test that installs a third snapshot or calls `deinit` from `snapshot_installed` and proves there is no UAF/double release.

src/tls/ticket_protection.zig
```diff
@@ -0,0 +1,1541 @@
...
+    defer keyring.deinit();
+    var protector = Protector{ .provider = testProvider(), .keyring = &keyring, .limits = limits };
+    try testing.expectEqual(encoded_len + envelope_overhead, try protector.protectedLen(&state));
+
+    limits.max_ticket_len = encoded_len + envelope_overhead - 1;
+    protector.limits = limits;
+    try testing.expectError(error.TicketTooLarge, protector.protectedLen(&state));
+}
+
+test "seal and resolve round trip and authenticate header fields" {
```
Jaruso, line 1289: This is still the only seal/open round trip, and it covers only AES-128-GCM. Issue #363 explicitly requires a deterministic envelope fixture, a round trip for every supported AEAD/key size, and broad tamper/lifecycle testing before merge.

Please add table-driven coverage for at least:

- AES-128-GCM, AES-256-GCM, and ChaCha20-Poly1305 round trips;
- a pinned fixed-key/fixed-nonce envelope fixture, including complete header/ciphertext/tag bytes;
- mutations of magic/version/AEAD/flags/key ID/nonce/ciphertext/tag, all returning unusable identity state;
- unknown, future, active, retained, and retired decrypt windows at exact boundaries;
- truncated, minimum-size, oversized, and trailing-byte identities;
- concurrent sealing over one lease, proving every successful nonce is unique and exactly the lease cardinality succeeds;
- snapshot replacement while an acquired/in-flight old snapshot is still used, proving delayed final wipe;
- allocation-failure sweeps through snapshot build, seal plaintext allocation, resolve plaintext allocation, and authenticated decode, with output/state unchanged and secret buffers zeroized.

The existing key-ID flip, single-thread exhaustion, and first-allocation failure tests are useful, but they do not satisfy the issue's mandatory security test matrix.

Jaruso (commented):
Re-review after commit `70edaee`: the five findings from the previous pass are addressed and I resolved those threads. The generation assignment now occurs only after a successful build, same-pointer install consumes the retained reference, semantic state corruption maps to `InvalidInternalState`, terminal observer reporting is centralized, and the key-fingerprint hash context is wiped. CI 857 is green.

Two merge blockers remain:

1. The new per-key retirement reporting dereferences `replacement` after invoking arbitrary observer callbacks. A reentrant observer can install another snapshot or deinitialize the keyring during `snapshot_installed` (or during the first `key_retired` event), releasing/destroying `replacement` before the outer install continues. Compute the retirement count before any callback, or retain the required snapshots across callback execution.

2. The issue's explicit required security test matrix is still incomplete. This is now a ready-for-review PR, but the current module tests only one AEAD round trip and one authenticated header mutation. Add the deterministic fixture, all supported AEADs, comprehensive mutation/boundary cases, concurrent nonce uniqueness, in-flight retirement, and allocation/zeroization coverage required by #363.

The PR description also still says the resolver contract is unavailable and the PR is draft; that text should be updated now that #479 is merged.

src/tls/ticket_protection.zig
```diff
@@ -0,0 +1,1859 @@
...
+
+        self.mutex.lock();
+        if (self.next_generation == std.math.maxInt(u64)) {
+            self.mutex.unlock();
+            self.record(.{ .snapshot_rejected = .generation_overflow });
+            return error.GenerationOverflow;
+        }
+        replacement.generation = self.next_generation;
+        self.next_generation += 1;
+        self.mutex.unlock();
```
Jaruso: This still changes generation state before the candidate has passed publication validation. A valid snapshot can build successfully and then fail `install` because its nonce lease overlaps history, its key lineage conflicts, it is stale, or `ledger.ensureUnusedCapacity` returns OOM:

```text
current generation = 1
next_generation = 2
buildSnapshot(candidate) -> candidate.generation = 2, next_generation = 3
install(candidate) -> OverlappingNonceLease
current/ledger stay unchanged, but next_generation is still 3
```

That violates #363's requirement that a failed reload not partially change generation, and repeated invalid reloads consume generations without publication.

Keep keyring-built snapshots provisional (`generation = 0` is a usable unpublished sentinel) and assign/increment under the install mutex only **after** stale/overflow, lineage, lease, and capacity validation all succeed. Explicitly-generated nonzero snapshots can retain the existing monotonic/stale checks. Add a failed-install rollback test that asserts `next_generation`, `current`, and the ledger are unchanged, then installs an adjacent valid lease and proves it receives the next contiguous generation.

src/tls/ticket_protection.zig
```diff
@@ -0,0 +1,1859 @@
...
+
+    var state = try sampleServerState(allocator);
+    defer state.deinit();
+    var protector = Protector{ .provider = testProvider(), .keyring = &keyring, .limits = session.Limits.default };
+    var ticket_buf = [_]u8{0} ** 512;
+    const ticket = try protector.seal(allocator, &state, 2_000, &ticket_buf);
+
+    try expectResolveFalseUnchanged(&protector, allocator, ticket, 999);
+    try expectRoundTrip(&protector, allocator, &state, ticket, 1_000);
+    try expectRoundTrip(&protector, allocator, &state, ticket, 4_999);
```
Jaruso, line 1408: The expanded matrix is strong, but these assertions only test **key-window** rejection: at `999` the key is future, and at `20_000` the key is retired. They do not reach the authenticated session-state checks in `resolveInner`.

Please add distinct cases where the key remains active/retained:

- seal a state with `issued_at_unix_ms` later than the resolve time, then assert `false`/`.not_yet_valid`;
- seal a short-lived state and resolve exactly at session expiry while `now < key.decrypt_until_unix_ms`, then assert `false`/`.expired`;
- construct a correctly authenticated envelope whose plaintext is not a valid `session.ServerRecoverableState` encoding, then assert `false`/`.invalid_plaintext` and unchanged `out`.

Those are separate peer-controlled branches required by #363 and are otherwise easy to accidentally collapse behind key-window or authentication failures. The allocation tests also verify frees and unchanged state, but not the promised plaintext zeroization; a test allocator that inspects bytes at `free` would make that lifecycle guarantee regression-proof.

Jaruso (commented):
Re-review of `17975fb`: the observer reentrancy/UAF finding is fixed, and I resolved that thread. The expanded tests now cover all three AEADs, a deterministic envelope fixture, representative envelope tampering, size boundaries, concurrent nonce exhaustion/uniqueness, retained-snapshot lifetime, and the main allocation failures. CI run 859 is green across the full matrix.

One publication correctness blocker remains: `buildSnapshot` still reserves/increments `next_generation` before `install` validates and publishes the candidate. Therefore an overlapping-lease, duplicate-lineage, stale, or ledger-allocation install failure changes generation state despite the failed reload, contrary to #363's rollback requirement. Generation assignment for keyring-built candidates needs to happen inside `install`, after every fallible validation step and immediately before publication.

The security test thread is much closer but should remain open for three distinct resolver outcomes that the current boundary test does not exercise: authenticated malformed plaintext (`invalid_plaintext`), session-level `not_yet_valid`, and session-level `expired` while the key itself remains decryptable. The current `999` and `20_000` assertions are rejected by the key window first. Also add a zero-checking allocator/test hook if the issue's required zeroization behavior is meant to be regression-protected rather than only code-inspected.

The PR description still contains the obsolete statement that #362 is unavailable and the PR is draft.

src/tls/ticket_protection.zig
Jaruso: Addressed in `26113999`: `buildSnapshot` now returns a provisional generation-0 snapshot, and `install` assigns/increments `next_generation` only after stale checks and `validateReplacementLocked` succeed (`src/tls/ticket_protection.zig:445` and `:489`). Added `failed install leaves generation current and ledger unchanged` to assert a failed overlapping-lease install preserves `next_generation`, `current`, and ledger state, then installs the adjacent lease with the contiguous generation (`:1546`).

src/tls/ticket_protection.zig
Jaruso: Addressed in `26113999`: added authenticated session-state resolver coverage for `.not_yet_valid`, `.expired`, and `.invalid_plaintext` while the key remains active/decryptable (`src/tls/ticket_protection.zig:1413`). The invalid-plaintext case uses a correctly authenticated envelope with non-session plaintext. Also added `zeroAndFree` plus `ZeroCheckingAllocator` coverage so seal and invalid-plaintext resolve plaintext buffers must be zero at raw-free time (`:865`, `:1172`, `:1910`).

Jaruso (commented):
Final re-review of `26113999`: all previously reported blockers are addressed, and I found no new merge-blocking defect in the latest delta or the integrated implementation.

- keyring-built snapshots remain provisional until every fallible publication check succeeds, so failed installs preserve `next_generation`, `current`, and nonce history;
- explicit nonzero generations retain stale/overflow enforcement;
- authenticated `.not_yet_valid`, `.expired`, and `.invalid_plaintext` resolver outcomes are now tested without being masked by key-window or AEAD rejection;
- plaintext zeroization is verified by an allocator that inspects bytes at free time for seal and invalid-plaintext resolution;
- the broader #363 matrix covers every supported AEAD, deterministic fixture bytes, authenticated-field/ciphertext/tag tampering, exact window and size boundaries, concurrent nonce uniqueness/exhaustion, reload rollback, observer reentrancy, in-flight snapshot retirement, allocation failures, and secret lifecycle behavior.

CI run 860 completed successfully. All review threads from this review cycle are resolved. From this review, PR #478 is ready to merge.

PR #479, opened 2026-07-20
tls: implement TLS 1.3 PSK identities, modes, and binder verification

Jaruso:
## Summary

Implements RFC 8446 §4.2.11 resumption-PSK negotiation on top of the existing session/`NewSessionTicket` model (#360, #361, PR #475/#477).

- **New `src/tls/pre_shared_key.zig`** — the shared module the issue centers on: `psk_key_exchange_modes`/`pre_shared_key` wire encode/decode (with full structural preflight — vector/length bounds validated before the writer is ever touched), `OfferedPsks` with borrowed paired iteration, the exact ClientHello binder-vector offset calculation, resumption binder derivation/constant-time verification for SHA-256 *and* SHA-384, obfuscated-ticket-age and age-skew helpers, an owned move-oriented `ClientPskOfferSet` (bounded to 8), and the provider-neutral `ServerPskResolver` contract shared by future stateful (#364) and stateless (#363) resolvers. It imports only `messages.zig`/`session.zig`, no cache/keyring/HTTP/QUIC/record-layer/policy types.
- **`session.zig`** — adds the `ticket_age_add: u32` prerequisite field to `ServerRecoverableState` (requested in the issue as a #361 amendment) and threads it through the codec.
- **`key_schedule.zig`** — `KeySchedule.initWithPsk`, a real-PSK early-secret initializer alongside the existing zero-PSK `init`.
- **`handshake.zig`** — `Core.psk_authenticated` / `enterPskAuthenticated()`: the resumed flight skips Certificate/CertificateVerify on both sides.
- **`tls13_backend.zig`** — wires client and server end to end, including the async credential-selection suspend/resume path:
  - the client plans (and compacts `client_psk_offers` to) exactly the eligible, size-fitting, wire-ordered offer subset *before* `core.start()` mutates any lifecycle/key-pair state, then offers those tickets as the last ClientHello extension with real binders (RFC 8446 §4.2.11.2 truncation/patch ordering);
  - the server resolves each offered identity (bounded to 8 attempts), evaluates #360 compatibility against the identity it's actually authenticating with (`inspectSelectedServerCredential` — every chain entry, not just the leaf), verifies only the first compatible candidate's binder (fatal `DecryptError`, never probes a later identity), and on success completes ServerHello → EncryptedExtensions → Finished with the PSK-derived key schedule and no certificate flight;
  - the client retains the selected ticket until EncryptedExtensions confirms it against the negotiated ALPN/transport/application context, then carries its certificate binding forward for any ticket issued on the resumed connection;
  - async server credential selection (the `pending_op`/`resumeAuth` suspend path) resumes through the identical PSK-selection code the synchronous path uses, since the captured ClientHello is backend-owned state rather than tied to the transient reassembly buffer;
  - every owned PSK/ticket/binder/schedule copy is wiped via `defer`/`errdefer` on every path — success, fallback, resolver failure, binder failure, malformed later ClientHello/ServerHello fields, and async cancellation/failure.

### Explicitly out of scope for this PR
- Ticket encryption at rest (#363) and session-cache storage (#364) — non-goals of #362 itself; this PR only defines the `ServerPskResolver` contract they'll adapt to.
- 0-RTT, HelloRetryRequest, external PSKs — all explicit non-goals per the issue.

## Test plan
- [x] `zig build test` — full workspace, 1851/1852 tests pass (1 pre-existing unrelated skip)
- [x] End-to-end PSK round trips in `tls13_backend_tests.zig`: a real ticket issued → offered → resolved → binder-verified → resumed handshake with application data exchanged under the PSK-derived keys, over both the record profile and an extension/QUIC-style profile with *asymmetric* client/server transport payloads; a resolver miss and a certificate-rotation mismatch each falling back to a full handshake; handshake-time client authentication forcing full-handshake fallback even when a PSK is offered
- [x] Server-selection-algorithm coverage: unknown-then-valid identity, a compatible candidate with a wrong binder (fatal, never probes a later identity), a malformed non-leaf credential chain entry (empty and oversized) rejected before the resolver/binder ever runs, `psk_key_exchange_modes` omitted (`MissingExtension`), an ineligible earlier offer not desyncing a later valid one's wire index
- [x] Async credential-selection parity: PSK selection resumes identically to the synchronous path; an async selection failure clears the captured offer
- [x] `pre_shared_key.zig` unit tests: wire round-trips, malformed/truncated vectors, writer structural preflight (empty/over-bound/oversized), binder cross-check, SHA-384, ticket-age wraparound and age-skew, offer-set move semantics and the 8/9 bound
- [x] `zig fmt --check` clean

Closes #362

### DIFF
--- a/src/tls/algorithms.zig
+++ b/src/tls/algorithms.zig
@@ -50,7 +50,9 @@ pub const ExtensionType = enum(u16) {
     supported_groups = 10,
     signature_algorithms = 13,
     application_layer_protocol_negotiation = 16,
+    pre_shared_key = 41,
     supported_versions = 43,
+    psk_key_exchange_modes = 45,
     key_share = 51,
     quic_transport_parameters = 57,
 };
--- a/src/tls/handshake.zig
+++ b/src/tls/handshake.zig
@@ -78,12 +78,31 @@ pub const Core = struct {
     client_auth_inbound: ClientAuthInbound = .inactive,
     /// Client outbound sub-state for its own certificate flight.
     client_auth_outbound: ClientAuthOutbound = .inactive,
+    /// Set once a PSK-resumed handshake (#362) has been selected: the
+    /// server flight after EncryptedExtensions goes straight to Finished
+    /// (no CertificateRequest, Certificate, or CertificateVerify), and the
+    /// client expects Finished immediately after EncryptedExtensions. Both
+    /// sides set this — via `enterPskAuthenticated` — once they know a PSK
+    /// was selected (server: after choosing to accept it, before its own
+    /// EncryptedExtensions is sent; client: after parsing ServerHello's
+    /// selected_identity, before EncryptedExtensions arrives). Never set
+    /// together with `request_client_certificate`/a client-auth flight:
+    /// handshake-time client authentication forces full-handshake fallback.
+    psk_authenticated: bool = false,
 
     pub const ClientAuthInbound = enum { inactive, expect_certificate, expect_certificate_verify, expect_finished };
     pub const ClientAuthOutbound = enum { inactive, send_certificate, send_certificate_verify, send_finished };
 
     pub fn init(role: state.Role) Core {
         return .{ .role = role };
+    }
+
+    /// Declare that a PSK-resumed handshake was selected (#362): the
+    /// remaining flight on both sides skips straight from EncryptedExtensions
+    /// to Finished. Must be called before the EncryptedExtensions message
+    /// is sent (server) or accepted (client).
+    pub fn enterPskAuthenticated(self: *Core) void {
+        self.psk_authenticated = true;
     }
 
     /// Server: request that the client authenticate. Call before the server
@@ -250,8 +269,8 @@ pub const Core = struct {
                     self.expected_inbound = .encrypted_extensions;
                 },
                 .encrypted_extensions => {
-                    self.handshake_state = .certificate;
-                    self.expected_inbound = .certificate;
+                    self.handshake_state = if (self.psk_authenticated) .finished else .certificate;
+                    self.expected_inbound = if (self.psk_authenticated) .finished else .certificate;
                 },
                 .certificate => {
                     self.handshake_state = .certificate_verify;
@@ -287,8 +306,12 @@ pub const Core = struct {
             },
             .server => switch (kind) {
                 .server_hello => self.handshake_state = .encrypted_extensions,
-                .encrypted_extensions => self.handshake_state =
-                    if (self.request_client_certificate) .certificate_request else .certificate,
+                .encrypted_extensions => self.handshake_state = if (self.psk_authenticated)
+                    .finished
+                else if (self.request_client_certificate)
+                    .certificate_request
+                else
+                    .certificate,
                 .certificate_request => self.handshake_state = .certificate,
                 .certificate => self.handshake_state = .certificate_verify,
                 .certificate_verify => self.handshake_state = .finished,
@@ -333,6 +356,47 @@ test "core records both directions of a client and server flight" {
     const cf = try messages.encode(.finished, "", &bytes);
     try client.recordSent(cf);
     _ = try server.acceptReceived(cf);
+    try std.testing.expectEqual(.complete, client.handshake_lifecycle);
+    try std.testing.expectEqual(.complete, server.handshake_lifecycle);
+    const client_hash = client.transcriptHash();
+    const server_hash = server.transcriptHash();
+    try std.testing.expectEqualSlices(u8, &client_hash, &server_hash);
+}
+
+test "PSK-authenticated core skips Certificate/CertificateVerify and still completes both directions" {
+    var client = Core.init(.client);
+    var server = Core.init(.server);
+    try client.start();
+    try server.start();
+
+    var bytes: [8]u8 = undefined;
+    const ch = try messages.encode(.client_hello, "", &bytes);
+    try client.recordSent(ch);
+    _ = try server.acceptReceived(ch);
+    const sh = try messages.encode(.server_hello, "", &bytes);
+    try server.recordSent(sh);
+    _ = try client.acceptReceived(sh);
+
+    // Both sides learn PSK was selected before EncryptedExtensions crosses
+    // the wire (server: before sending it; client: before receiving it).
+    server.enterPskAuthenticated();
+    client.enterPskAuthenticated();
+
+    const ee = try messages.encode(.encrypted_extensions, "", &bytes);
+    try server.recordSent(ee);
+    _ = try client.acceptReceived(ee);
+
+    // Certificate/CertificateVerify are neither expected nor legal now.
+    const cert = try messages.encode(.certificate, "", &bytes);
+    try std.testing.expectError(error.UnexpectedHandshakeMessage, client.acceptReceived(cert));
+
+    const sf = try messages.encode(.finished, "", &bytes);
+    try server.recordSent(sf);
+    _ = try client.acceptReceived(sf);
+    const cf = try messages.encode(.finished, "", &bytes);
+    try client.recordSent(cf);
+    _ = try server.acceptReceived(cf);
+
     try std.testing.expectEqual(.complete, client.handshake_lifecycle);
     try std.testing.expectEqual(.complete, server.handshake_lifecycle);
     const client_hash = client.transcriptHash();
--- a/src/tls/key_schedule.zig
+++ b/src/tls/key_schedule.zig
@@ -42,8 +42,40 @@ pub const KeySchedule = struct {
     server_handshake_traffic: [hash_len]u8,
 
     pub fn init(shared: *const [shared_secret_len]u8, hello_transcript_hash: [hash_len]u8) KeySchedule {
+        return initFromEarlySecret(&derived_early_secret, shared, hello_transcript_hash);
+    }
+
+    /// PSK-resumed (`psk_dhe_ke`) handshake: the same key-schedule chain as
+    /// `init`, but starting from a real early secret derived from the
+    /// resumption PSK (RFC 8446 §7.1) instead of the comptime all-zero-PSK
+    /// early secret `init` uses. `psk` must already be exactly `hash_len`
+    /// bytes (the SHA-256 resumption PSK produced by
+    /// `resumptionPsk`/`KeySchedule.resumptionPsk`, matching this module's
+    /// concrete SHA-256 schedule). X25519 key share remains mandatory in
+    /// this profile, so `shared` is still the ECDHE shared secret.
+    pub fn initWithPsk(
+        psk: *const [hash_len]u8,
+        shared: *const [shared_secret_len]u8,
+        hello_transcript_hash: [hash_len]u8,
+    ) KeySchedule {
+        var early_secret = HkdfSha256.extract("", psk);
+        defer crypto.secureZero(u8, &early_secret);
+        var derived_early = tls.hkdfExpandLabel(HkdfSha256, early_secret, "derived", &empty_transcript_hash, hash_len);
+        defer crypto.secureZero(u8, &derived_early);
+        return initFromEarlySecret(&derived_early, shared, hello_transcript_hash);
+    }
+
+    /// Shared continuation from a "derived" early secret (RFC 8446 §7.1)
+    /// into the handshake/master secrets and handshake traffic secrets. Used
+    /// by both the zero-PSK (`init`) and real-PSK (`initWithPsk`) entry
+    /// points, which differ only in how `derived_early_secret` was produced.
+    fn initFromEarlySecret(
+        derived_early: *const [hash_len]u8,
+        shared: *const [shared_secret_len]u8,
+        hello_transcript_hash: [hash_len]u8,
+    ) KeySchedule {
         const zeros = [_]u8{0} ** hash_len;
-        var handshake_secret = HkdfSha256.extract(&derived_early_secret, shared);
+        var handshake_secret = HkdfSha256.extract(derived_early, shared);
         defer crypto.secureZero(u8, &handshake_secret);
         var derived_handshake = tls.hkdfExpandLabel(HkdfSha256, handshake_secret, "derived", &empty_transcript_hash, hash_len);
         defer crypto.secureZero(u8, &derived_handshake);
@@ -255,6 +287,43 @@ test "generic resumption master secret derivation supports SHA-384" {
     try KeySchedule.deriveResumptionMasterSecret(.sha384, &master_secret, &transcript_hash, &out);
     try std.testing.expectEqualSlices(u8, &hexBytes("4f9d68ff762f5b886f275d162b90c268db5ccc65c4e0b8fc810030429a070f8e9f12b641b209e15ae210b1153a68fc42"), &out);
     try std.testing.expectError(error.InvalidSecretLength, KeySchedule.deriveResumptionMasterSecret(.sha384, master_secret[0..hash_len], &transcript_hash, &out));
+}
+
+test "initWithPsk diverges from the zero-PSK schedule and is deterministic" {
+    const shared = [_]u8{0x42} ** shared_secret_len;
+    const transcript = [_]u8{0x24} ** hash_len;
+    const psk = [_]u8{0x99} ** hash_len;
+
+    var zero_psk_schedule = KeySchedule.init(&shared, transcript);
+    defer zero_psk_schedule.wipe();
+    var psk_schedule = KeySchedule.initWithPsk(&psk, &shared, transcript);
+    defer psk_schedule.wipe();
+    var psk_schedule_again = KeySchedule.initWithPsk(&psk, &shared, transcript);
+    defer psk_schedule_again.wipe();
+
+    try std.testing.expect(!std.mem.eql(u8, &zero_psk_schedule.handshake_secret, &psk_schedule.handshake_secret));
+    try std.testing.expect(!std.mem.eql(u8, &zero_psk_schedule.master_secret, &psk_schedule.master_secret));
+    try std.testing.expectEqualSlices(u8, &psk_schedule.handshake_secret, &psk_schedule_again.handshake_secret);
+    try std.testing.expectEqualSlices(u8, &psk_schedule.master_secret, &psk_schedule_again.master_secret);
+    try std.testing.expect(!std.mem.eql(u8, &psk_schedule.client_handshake_traffic, &psk_schedule.server_handshake_traffic));
+
+    var app = psk_schedule.applicationSecrets(transcript);
+    defer app.wipe();
+    try std.testing.expect(!std.mem.eql(u8, &app.client, &app.server));
+}
+
+test "a different resumption PSK produces a different PSK-resumed schedule" {
+    const shared = [_]u8{0x11} ** shared_secret_len;
+    const transcript = [_]u8{0x22} ** hash_len;
+    const psk_a = [_]u8{0xaa} ** hash_len;
+    const psk_b = [_]u8{0xbb} ** hash_len;
+
+    var schedule_a = KeySchedule.initWithPsk(&psk_a, &shared, transcript);
+    defer schedule_a.wipe();
+    var schedule_b = KeySchedule.initWithPsk(&psk_b, &shared, transcript);
+    defer schedule_b.wipe();
+
+    try std.testing.expect(!std.mem.eql(u8, &schedule_a.master_secret, &schedule_b.master_secret));
 }
 
 fn hexBytes(comptime hex: []const u8) [hex.len / 2]u8 {
--- a/src/tls/key_schedule.zig
+++ b/src/tls/key_schedule.zig
@@ -312,6 +312,29 @@ test "initWithPsk diverges from the zero-PSK schedule and is deterministic" {
     try std.testing.expect(!std.mem.eql(u8, &app.client, &app.server));
 }
 
+test "initWithPsk matches independently computed secrets" {
+    // Checked-in literals for the same inputs as "initWithPsk diverges from
+    // the zero-PSK schedule and is deterministic" above (psk=0x99*32,
+    // shared=0x42*32, transcript=0x24*32), computed independently of this
+    // module rather than by re-deriving with the same helpers under test.
+    const shared = [_]u8{0x42} ** shared_secret_len;
+    const transcript = [_]u8{0x24} ** hash_len;
+    const psk = [_]u8{0x99} ** hash_len;
+
+    var schedule = KeySchedule.initWithPsk(&psk, &shared, transcript);
+    defer schedule.wipe();
+
+    try std.testing.expectEqualSlices(u8, &hexBytes("ab0803d6203c8feddfe8adc74f986c9d89b817b3d4132fc55c866a3522d9ff49"), &schedule.handshake_secret);
+    try std.testing.expectEqualSlices(u8, &hexBytes("e92139285417b6a9a54a7a9153f4b6dcce44b99cdc0937b83dfea5c79805c920"), &schedule.client_handshake_traffic);
+    try std.testing.expectEqualSlices(u8, &hexBytes("739483d9d6a9508c73b4656de22fedd85a2a8d00e9a6ca1449d8cba678c94baf"), &schedule.server_handshake_traffic);
+    try std.testing.expectEqualSlices(u8, &hexBytes("abe96cce65361235f3126971c67760888b79d4c1724a6cb1e15f6d2ae128ff44"), &schedule.master_secret);
+
+    var app = schedule.applicationSecrets(transcript);
+    defer app.wipe();
+    try std.testing.expectEqualSlices(u8, &hexBytes("d1ba0b1be9862f1bd4c3bcc0d53b5a98c6a4951c4bad19243051237bc735031c"), &app.client);
+    try std.testing.expectEqualSlices(u8, &hexBytes("c7428c93109f1b656dcbf0971e5d1bad9c2d38b79420038b7e165a17c7f61fa1"), &app.server);
+}
+
 test "a different resumption PSK produces a different PSK-resumed schedule" {
     const shared = [_]u8{0x11} ** shared_secret_len;
     const transcript = [_]u8{0x22} ** hash_len;
--- a/src/tls/key_schedule.zig
+++ b/src/tls/key_schedule.zig
@@ -244,6 +244,24 @@ test "application traffic secret storage has explicit cleanup" {
     try std.testing.expect(std.mem.allEqual(u8, std.mem.asBytes(&app), 0));
 }
 
+test "KeySchedule.wipe zeroizes every derived secret" {
+    // Deliberately a plain (non-optional) local, wiped in place and
+    // inspected directly — unlike a backend-owned `?KeySchedule` that gets
+    // set to `null` right after wiping, there is no subsequent
+    // optional-invalidation step here whose own debug-safety poisoning
+    // could be mistaken for (or mask the absence of) this `wipe()` call's
+    // effect. That makes this the reliable place to prove the zeroing
+    // itself; backend-level tests should only assert that `schedule`
+    // becomes `null`, not re-inspect the bytes afterward.
+    const shared = [_]u8{0x77} ** shared_secret_len;
+    const transcript = [_]u8{0x88} ** hash_len;
+    var schedule = KeySchedule.init(&shared, transcript);
+    const bytes = std.mem.asBytes(&schedule);
+    try std.testing.expect(!std.mem.allEqual(u8, bytes, 0));
+    schedule.wipe();
+    try std.testing.expect(std.mem.allEqual(u8, bytes, 0));
+}
+
 test "resumption master secret and PSK derivation are deterministic" {
     const shared = hexBytes("8bd4054fb55b9d63fdfbacf9f04b9f0d35e6d63f537563efd46272900f89492d");
     const hello_hash = hexBytes("860c06edc07858ee8e78f0e7428c58edd6b43f2ca3e6e95f02ed063cf0e1cad8");
--- a/src/tls/new_session_ticket.zig
+++ b/src/tls/new_session_ticket.zig
@@ -228,7 +228,7 @@ pub fn buildServerRecoverableState(
     });
 
     var state: session.ServerRecoverableState = .{};
-    state.init(&common);
+    state.init(&common, params.ticket_age_add);
     return state;
 }
 
--- a/src/tls/pre_shared_key.zig
+++ b/src/tls/pre_shared_key.zig
@@ -239,12 +239,14 @@ pub fn writeOffer(w: *messages.Writer, items: []const OfferItem) WriteError!Clie
         if (item.identity.len > max_vector_len) return error.IdentityTooLarge;
         identities_total += 2 + item.identity.len + 4;
         if (identities_total > max_vector_len) return error.IdentitiesVectorTooLarge;
-        // This writer only ever produces the digest lengths the module's own
-        // binder derivation supports (32 for SHA-256, 48 for SHA-384), a
-        // stricter bound than the wire's general 32..255 `PskBinderEntry`
-        // range — and exactly what fits the fixed zero-placeholder buffer
-        // below.
-        if (item.digest_len < min_binder_len or item.digest_len > provider.max_digest_len)
+        // The full wire range (RFC 8446 §4.2.11: `PskBinderEntry<32..255>`),
+        // matching what `OfferedPsks.parse` accepts — this is a general
+        // wire codec, not specific to the two digest lengths this module's
+        // own binder derivation happens to support. A caller preparing an
+        // actual offer (see `tls13_backend.zig`) is the one that only ever
+        // constructs 32- or 48-byte entries; that is backend policy, not a
+        // codec-level restriction.
+        if (item.digest_len < min_binder_len or item.digest_len > max_binder_len)
             return error.InvalidBinderLength;
         binders_total += 1 + item.digest_len;
         if (binders_total > max_vector_len) return error.BindersVectorTooLarge;
@@ -274,7 +276,7 @@ pub fn writeOffer(w: *messages.Writer, items: []const OfferItem) WriteError!Clie
     for (items, 0..) |item, i| {
         const entry_len_idx = try w.reserve(1);
         const slot_offset = w.len;
-        const zeros = [_]u8{0} ** provider.max_digest_len;
+        const zeros = [_]u8{0} ** max_binder_len;
         try w.bytes(zeros[0..item.digest_len]);
         w.patch(1, entry_len_idx);
         result.slots[i] = .{ .offset = slot_offset, .len = item.digest_len };
@@ -562,17 +564,27 @@ test "writeOffer rejects illegal shapes without ever mutating the writer" {
         try testing.expectError(error.EmptyIdentity, writeOffer(&w, &items));
         try testing.expectEqual(@as(usize, 0), w.len);
     }
-    // Binder/digest length outside what this module can produce (32..48).
+    // Binder length boundary, over the full RFC 8446 wire range (32..255),
+    // matching `OfferedPsks.parse` exactly rather than this module's own
+    // narrower SHA-256/384 digest lengths: 31 rejected, 32/48/255 accepted,
+    // 256 unrepresentable on the wire and rejected.
     {
         var w = messages.Writer{ .buf = &buf };
-        const too_short = [_]OfferItem{.{ .identity = "id", .obfuscated_ticket_age = 0, .digest_len = min_binder_len - 1 }};
+        const too_short = [_]OfferItem{.{ .identity = "id", .obfuscated_ticket_age = 0, .digest_len = 31 }};
         try testing.expectError(error.InvalidBinderLength, writeOffer(&w, &too_short));
         try testing.expectEqual(@as(usize, 0), w.len);
 
-        var w2 = messages.Writer{ .buf = &buf };
-        const too_long = [_]OfferItem{.{ .identity = "id", .obfuscated_ticket_age = 0, .digest_len = provider.max_digest_len + 1 }};
-        try testing.expectError(error.InvalidBinderLength, writeOffer(&w2, &too_long));
-        try testing.expectEqual(@as(usize, 0), w2.len);
+        inline for (.{ 32, 48, 255 }) |ok_len| {
+            var w_ok = messages.Writer{ .buf = &buf };
+            const items = [_]OfferItem{.{ .identity = "id", .obfuscated_ticket_age = 0, .digest_len = ok_len }};
+            const offer = try writeOffer(&w_ok, &items);
+            try testing.expectEqual(@as(usize, ok_len), offer.slots[0].len);
+        }
+
+        var w_too_long = messages.Writer{ .buf = &buf };
+        const too_long = [_]OfferItem{.{ .identity = "id", .obfuscated_ticket_age = 0, .digest_len = 256 }};
+        try testing.expectError(error.InvalidBinderLength, writeOffer(&w_too_long, &too_long));
+        try testing.expectEqual(@as(usize, 0), w_too_long.len);
     }
     // Exact boundary values succeed: min/max supported digest length, and a
     // full eight-identity offer.
@@ -580,7 +592,7 @@ test "writeOffer rejects illegal shapes without ever mutating the writer" {
         var w = messages.Writer{ .buf = &buf };
         const items = [_]OfferItem{
             .{ .identity = "a", .obfuscated_ticket_age = 0, .digest_len = min_binder_len },
-            .{ .identity = "b", .obfuscated_ticket_age = 0, .digest_len = provider.max_digest_len },
+            .{ .identity = "b", .obfuscated_ticket_age = 0, .digest_len = max_binder_len },
         };
         _ = try writeOffer(&w, &items);
 
@@ -653,43 +665,37 @@ test "OfferedPsks.parse rejects truncated and oversized binder lengths" {
     }
 }
 
-test "resumption binder derivation matches independently computed HKDF chain (SHA-256)" {
-    // Vector reuses the SHA-256 RFC 8448-style resumption PSK from
-    // key_schedule.zig's own resumption test so binder derivation is
-    // checked against a value derived independently of this module.
-    const key_schedule = @import("key_schedule.zig");
+test "resumption binder derivation matches an independently computed SHA-256 binder" {
+    // Checked-in literal, computed independently of this module (Python
+    // hashlib/hmac against the same RFC 8446 §4.2.11.2/§7.1 label chain) —
+    // not derived with the same `std.crypto.tls.hkdfExpandLabel` helper the
+    // implementation under test uses, so this actually detects a wrong
+    // label, wrong empty-hash input, or wrong transcript hash rather than
+    // only cross-checking this module against itself.
     const psk = hexBytes("c1392efd98f6932d62f5ccd42c724230871638e8ad0ac9ce9b2af89f5f919fed");
     const client_hello_prefix = "pretend-truncated-clienthello-bytes";
+    const expected = hexBytes("de7c7afec445f9419e4f769b6ef8e6371e1f599405eff6ecc014499f234a008e");
 
     var binder: [32]u8 = undefined;
     try deriveBinder(.sha256, &psk, client_hello_prefix, &binder);
-
-    // Independently recompute the same chain inline to cross-check.
-    var empty_hash: [32]u8 = undefined;
-    Sha256.hash("", &empty_hash, .{});
-    var early_secret = HkdfSha256.extract("", &psk);
-    var binder_key = tls.hkdfExpandLabel(HkdfSha256, early_secret, "res binder", &empty_hash, 32);
-    var finished_key = tls.hkdfExpandLabel(HkdfSha256, binder_key, "finished", "", 32);
-    var transcript: [32]u8 = undefined;
-    Sha256.hash(client_hello_prefix, &transcript, .{});
-    var expected: [32]u8 = undefined;
-    HmacSha256.create(&expected, &transcript, &finished_key);
-    crypto.secureZero(u8, &early_secret);
-    crypto.secureZero(u8, &binder_key);
-    crypto.secureZero(u8, &finished_key);
-
     try testing.expectEqualSlices(u8, &expected, &binder);
+
     try testing.expect(try verifyBinder(.sha256, &psk, client_hello_prefix, &binder));
     try testing.expect(!try verifyBinder(.sha256, &psk, "different prefix bytes here", &binder));
-
-    _ = key_schedule;
 }
 
-test "resumption binder derivation supports SHA-384 and rejects mismatched PSK length" {
+test "resumption binder derivation matches an independently computed SHA-384 binder" {
+    // Checked-in literal, independently computed (see the SHA-256 test
+    // above); the module's own concrete SHA-384 code path is entirely
+    // separate from the SHA-256 one, so this is not redundant with it.
     const psk384 = [_]u8{0x77} ** 48;
+    const expected = hexBytes(
+        "138b767e68513f232636a0d2b2c53d7a923ff2c0d7879985d4ea916281c5134" ++
+            "d3bc2b1ae31178e736fe22f2d906cfdd3",
+    );
     var binder: [48]u8 = undefined;
     try deriveBinder(.sha384, &psk384, "prefix", &binder);
-    try testing.expect(!std.mem.allEqual(u8, &binder, 0));
+    try testing.expectEqualSlices(u8, &expected, &binder);
 
     var short_binder: [47]u8 = undefined;
     try testing.expectError(error.InvalidSecretLength, deriveBinder(.sha384, &psk384, "prefix", &short_binder));
--- a/src/tls/pre_shared_key.zig
+++ b/src/tls/pre_shared_key.zig
@@ -604,6 +604,236 @@ test "writeOffer rejects illegal shapes without ever mutating the writer" {
     }
 }
 
+test "a literal framed ClientHello proves the exact binder truncation boundary" {
+    // A hand-assembled, fully framed ClientHello (handshake header through a
+    // trailing `pre_shared_key` extension with one identity "abcd" and one
+    // 32-byte binder), constructed byte-by-byte from the RFC 8446 grammar
+    // rather than via this module's own writer — every enclosing length
+    // (message, extensions vector, `pre_shared_key` extension, identities
+    // vector) is already final, exactly as the wire requires before hashing.
+    // The embedded binder itself was computed independently (Python
+    // hashlib/hmac, not this module) over the expected truncated prefix.
+    const message = hexBytes(
+        "0100006503037777777777777777777777777777777777777777777777777777" ++
+            "77777777777700000213010100003a002b00030203040029002f000a00046162" ++
+            "6364112233440021208f6e79d089388cd1e7ca42e346e44ba217289f0450609a" ++
+            "9827c4ee59f16568e6",
+    );
+    // The `pre_shared_key` extension_data begins at a fixed, independently
+    // computed offset within `message` (4-byte header + fixed ClientHello
+    // fields + the `supported_versions` extension that precedes it).
+    const ext_data_offset = 58;
+    const ext_data = message[ext_data_offset..];
+
+    var offered = try OfferedPsks.parse(ext_data);
+    try testing.expectEqual(@as(usize, 1), offered.count);
+    // Exactly the offset of the *2-byte binders-vector length field* —
+    // not the binder payload two bytes later.
+    try testing.expectEqual(@as(usize, 12), offered.binder_vector_offset);
+
+    const truncated_len = ext_data_offset + offered.binder_vector_offset;
+    try testing.expectEqual(@as(usize, 70), truncated_len);
+    const prefix = message[0..truncated_len];
+    const expected_prefix = hexBytes(
+        "0100006503037777777777777777777777777777777777777777777777777777" ++
+            "77777777777700000213010100003a002b00030203040029002f000a00046162" ++
+            "636411223344",
+    );
+    try testing.expectEqualSlices(u8, &expected_prefix, prefix);
+
+    const psk = [_]u8{0xaa} ** 32;
+    const expected_binder = hexBytes("8f6e79d089388cd1e7ca42e346e44ba217289f0450609a9827c4ee59f16568e6");
+    var binder: [32]u8 = undefined;
+    try deriveBinder(.sha256, &psk, prefix, &binder);
+    try testing.expectEqualSlices(u8, &expected_binder, &binder);
+
+    // Confirm the embedded pair's identity/binder round-trip through the
+    // borrowed paired iterator too.
+    var it = offered.pairs();
+    const pair = (try it.next()).?;
+    try testing.expectEqualStrings("abcd", pair.identity.identity);
+    try testing.expectEqual(@as(u32, 0x11223344), pair.identity.obfuscated_ticket_age);
+    try testing.expectEqualSlices(u8, &expected_binder, pair.binder);
+    try testing.expectEqual(@as(?OfferedPsks.PairIterator.Pair, null), try it.next());
+
+    // The offset is exact, not "close enough": hashing two bytes further
+    // (as if the binders-vector length field were mistakenly included in
+    // the prefix) must *not* reproduce the same binder.
+    var wrong_binder: [32]u8 = undefined;
+    try deriveBinder(.sha256, &psk, message[0 .. truncated_len + 2], &wrong_binder);
+    try testing.expect(!std.mem.eql(u8, &expected_binder, &wrong_binder));
+}
+
+test "every strict prefix of a valid OfferedPsks fixture is rejected, none silently accepted" {
+    // The `pre_shared_key` extension_data from the literal ClientHello
+    // fixture above: identities_len(2) + one identity(10) + binders_len(2)
+    // + one 32-byte binder(33). Every vector is length-prefixed and spans
+    // the whole fixture, so there is no shorter prefix that could
+    // coincidentally look complete.
+    const ext_data = hexBytes(
+        "000a000461626364112233440021208f6e79d089388cd1e7ca42e346e44ba217" ++
+            "289f0450609a9827c4ee59f16568e6",
+    );
+    try testing.expectEqual(@as(usize, 47), ext_data.len);
+    _ = try OfferedPsks.parse(&ext_data); // the full fixture does parse
+
+    var cut: usize = 0;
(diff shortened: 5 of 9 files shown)


Jaruso (commented):
**Review result: changes required before merge.**

The core direction is good: `ticket_age_add` is now carried by the server state codec, the PSK extension is emitted last, the binder prefix is cut before the binders vector, identity resolution is bounded, and the abbreviated Finished-only authentication flow is wired through the existing secret events. CI run 841 is green.

A full review against #362 found the following correctness, security, ownership, and acceptance-coverage gaps.

## 1. Blocker/security: the ticket is not bound to the selected server credential

`emitServerHelloAndAuthFlight()` receives the credential selected for the current SNI, but `selectPsk()` does not receive or inspect it. Instead it builds the compatibility context with:

```zig
.auth_binding = self.peerAuthBinding(),
```

`peerAuthBinding()` hashes `peer_chain`. On the server, `peer_chain` is the **client's** certificate chain when client authentication is enabled, and is empty during the normal server-authenticated handshake. Therefore the current code usually compares the recovered ticket against `AuthBinding.fromLeafCertificateDer("")`, not against the selected server leaf certificate.

The same problem exists during ticket issuance: server-side `resumptionContext()` also uses `peerAuthBinding()`. The server state returned by `emitNewSessionTicket()` is consequently bound to the empty chain, while the client ticket state is bound to the server certificate it actually verified. The new round-trip test compares only the PSK, so it does not expose this mismatch.

This breaks the certificate-rotation fallback requirement and makes the selected credential effectively irrelevant to PSK acceptance. It also bypasses the normal selected-credential contract checks: the PSK path never verifies that the returned scheme was offered or that the credential has a non-empty, bounded certificate chain before releasing it.

Please derive and retain the connection's server-authentication binding from the selected credential itself, before PSK selection:

```zig
fn validateSelectedServerCredential(
    self: *Tls13Backend,
    credential: credentials.SelectedCredential,
) HandshakeError!session.AuthBinding {
    const selection = self.serverSelectionContext();

    const chain = credential.certificateChain();
    if (chain.count() == 0 or chain.count() > credentials.max_chain_entries)
        return self.failCredential(.malformed_credential_chain);

    const leaf = chain.entries[0];
    if (leaf.len == 0 or leaf.len > max_certificate_len)
        return self.failCredential(.malformed_credential_chain);

    return session.AuthBinding.fromLeafCertificateDer(leaf);
}
```

Then pass that binding into PSK compatibility evaluation and store it as the authenticated connection binding for subsequent `NewSessionTicket` construction:

```zig
const current_binding = try self.validateSelectedServerCredential(credential);
self.connection_auth_binding = current_binding;
const psk_selected = try self.selectPsk(current_binding);
```

The client must similarly carry the selected ticket's binding forward during a resumed handshake, because no new Certificate message exists from which to reconstruct it.

Required tests:

- server and client ticket states produced by one full handshake contain the same non-empty certificate binding;
- a ticket issued under credential A resumes under A;
- the same ticket falls back to a full handshake when the current SNI selects credential B;
- synchronous and asynchronous credential selection produce the same result;
- a malformed selected credential cannot be accepted merely because PSK succeeds.

## 2. Blocker: transmitted identity indexes do not map to the retained client offers

`sendClientHello()` filters the owned array while building the wire offer:

```zig
for (self.client_psk_offers.constSlice()) |*ticket| {
    if (ticket.common.isExpired(now_ms) or ticket.common.isNotYetValid(now_ms)) continue;
    if (ticket.common.cipher_suite != .tls_aes_128_gcm_sha256) continue;
    ...
    psk_items[psk_count] = ...;
    psk_count += 1;
}
```

The resulting wire identities are compacted. However, `onServerHello()` interprets `selected_identity` against the original, unfiltered array:

```zig
const psk_slice = self.client_psk_offers.tickets[idx]
    .common.resumption_psk.slice();
```

For example, if stored ticket 0 is expired and stored ticket 1 is valid, the ClientHello sends ticket 1 as wire identity 0. When the server selects index 0, the client uses ticket 0's PSK. The two sides derive different secrets.

The same bug lets a server select a ticket the client did not send at all. If every stored ticket is expired, incompatible, or omitted because no clock was configured, the ClientHello contains no PSK extension, but `client_psk_offers.len` remains nonzero. A ServerHello containing selected index 0 is currently accepted.

Please retain exactly the ordered subset that was emitted, either by compacting ownership into a second set or by retaining an explicit wire-index mapping. A compact owned set is less error-prone:

```zig
var emitted: pre_shared_key.ClientPskOfferSet = .{};
errdefer emitted.deinit();

for (self.client_psk_offers.slice()) |*ticket| {
    if (!eligible(ticket, current_context, now_ms)) continue;
    if (!fitsNextOffer(...)) break;
    try emitted.push(ticket); // move into exact wire order
}

self.client_psk_offers.deinit(); // skipped/unemitted entries
self.client_psk_offers.moveFrom(&emitted);
self.client_psk_offer_was_sent = self.client_psk_offers.len != 0;
```

On ServerHello, reject `selected_identity` unless a PSK extension was actually sent and index only this compact wire-order set. Use the existing `takeSelected()` operation rather than manually copying a PSK from one array and destroying every state.

Required tests:

- expired first / valid second selects wire index 0 and uses the second PSK;
- incompatible first / valid second behaves identically;
- unknown first / valid second selects index 1;
- all offers filtered out, followed by a forged selected index, is `IllegalParameter`;
- server-selected index equal to the emitted count is rejected.

## 3. Blocker: the selected client ticket is destroyed before its context can be validated or inherited

Immediately after ServerHello, the code copies only the PSK and calls:

```zig
self.client_psk_offers.deinit();
```

That discards the selected ticket before EncryptedExtensions supplies the negotiated ALPN and the server's current transport/application compatibility data. The server can check its stored view of the **client's** prior transport/application context, but only the client has the ticket state containing the **server's** prior context. The client must retain the selected state and independently compare the current EncryptedExtensions values.

This also breaks tickets issued after a resumed handshake. Since no Certificate message was received, `peer_chain` is empty. A later `NewSessionTicket` currently builds its new client state with an empty authentication binding instead of inheriting the selected ticket's server-authentication binding.

Please add an owned selected/resumed session field and keep it until all selected-ticket context has been applied:

```zig
selected_client_psk: session.ClientTicketState = .{},
selected_client_psk_present: bool = false,
```

After validating the wire index:

```zig
self.client_psk_offers.takeSelected(idx, &self.selected_client_psk);
self.selected_client_psk_present = true;
```

Then, during/after EncryptedExtensions:

- require the negotiated ALPN to equal the selected ticket's stored ALPN;
- compare the received transport compatibility snapshot;
- compare the configured/current application compatibility snapshot;
- carry forward `auth_binding` as the connection's authenticated server binding;
- deinitialize the selected state on success, fallback, failure, and teardown only after the necessary values have been copied into connection state.

## 4. Blocker: ClientHello preflight ignores PSK size and does not include the ordered prefix that fits

`clientHelloEncodedLen()` accounts only for the ordinary ClientHello. `startImpl()` therefore advances the core and then `sendClientHello()` attempts to encode every eligible ticket into the fixed 8 KiB buffer.

Two default-limit tickets can already exceed that buffer. Instead of emitting the ordered prefix that fits, the writer returns `HandshakeBufferOverflow` after lifecycle/key-pair mutation. This violates the issue's exact-size and “include candidates in order while they fit” requirements.

The size must be planned before `core.start()` and before generating/storing ephemeral state. The exact additional size is:

```text
modes extension:
  2 type + 2 ext_len + 1 vector_len + mode_count

PSK extension:
  2 type + 2 ext_len
  + 2 identities_len
  + Σ(2 + identity.len + 4)
  + 2 binders_len
  + Σ(1 + digest_len)
```

Build a prepared offer plan that:

1. performs every eligibility/context check;
2. uses checked arithmetic for each next candidate;
3. stops before `max_message_len` or a `u16` vector limit would be exceeded;
4. retains exactly the subset and mapping described above;
5. causes an unchanged ordinary ClientHello when no candidate fits;
6. leaves the lifecycle, offers, key pair, and sink unchanged when preflight itself fails.

Add exact-fit, one-byte-over, two-default-ticket, maximum-identity, and “none fit” tests.

## 5. High: ticket-age skew is implemented only as an unused helper

`observeAgeSkew()` exists and has a unit test, but the server selection path never consumes `obfuscated_ticket_age`, never uses recovered `ticket_age_add`, and never surfaces an observation for #366. Thus the corresponding acceptance criterion is not implemented.

After a candidate is compatible, compute elapsed time with overflow-safe arithmetic, but do not reject ordinary resumption based on skew:

```zig
fn elapsedMillis(now_ms: i64, issued_ms: i64) u64 {
    const delta: i128 = @as(i128, now_ms) - @as(i128, issued_ms);
    if (delta <= 0) return 0;
    return @intCast(@min(delta, @as(i128, std.math.maxInt(u64))));
}

const age = pre_shared_key.observeAgeSkew(
    pair.identity.obfuscated_ticket_age,
    candidate_state.ticket_age_add,
    elapsedMillis(now, candidate_state.common.issued_at_unix_ms),
);
```

Only publish/store the observation after the corresponding binder succeeds. Expose it through a bounded transport-neutral event or a one-shot backend accessor that #366 can consume. Add tests proving large positive/negative skew still permits 1-RTT resumption and that the exact observation is surfaced.

## 6. High: required compatibility inputs are missing

The server `CandidateContext` supplies transport compatibility but omits `application_compat`. Every recovered ticket containing an application snapshot therefore rejects as an application mismatch because the candidate defaults to null. There is no API through which #365/application code can supply the current snapshot.

The client-side recheck is also incomplete: it checks only time and the one cipher suite. It does not recheck SNI, whether the stored ALPN is currently offered, transport compatibility, or application compatibility as required by the issue.

Please add an explicit bounded compatibility-context seam and use it on both sides. At minimum, tests must cover SNI, ALPN, certificate binding, transport snapshot, and application snapshot mismatches independently and confirm clean full-handshake fallback without stale PSK state.

## 7. High/security: several bearer-ticket and PSK copies are not wiped on all paths

The new code introduces multiple secret-bearing copies that are only cleared on the happy path or not cleared at all:

- `psk_secrets` is zeroed by a normal statement at the end of ClientHello construction; any earlier writer/binder error skips it. Use `defer` immediately after declaration.
- the stack ClientHello `buf` contains opaque bearer ticket identities and binders and is never wiped after the sink has copied it;
- `clearClientHelloPsk()` assigns the optional to null, which does not securely clear the captured ClientHello payload;
- its defer is installed only **after** `try self.selectPsk()`, so resolver/binder errors retain the capture until later teardown;
- `PskSelected` stores the PSK in a raw array returned by value; the `psk_selected` copy remains on the stack after another copy is wiped;
- a resolver error can return after partially constructing `candidate_state`; the backend should defensively deinitialize a zero-safe candidate on every iteration;
- client offers are not immediately cleared on malformed ServerHello/error paths.

Use scoped ownership and unconditional cleanup, for example:

```zig
var psk_secrets: [max][hash_len]u8 = undefined;
defer crypto.secureZero(u8, std.mem.asBytes(&psk_secrets));

fn clearClientHelloPsk(self: *Tls13Backend) void {
    if (self.client_hello_psk) |*capture| {
        crypto.secureZero(u8, capture.message[0..capture.message_len]);
    }
    crypto.secureZero(u8, std.mem.asBytes(&self.client_hello_psk));
    self.client_hello_psk = null;
    ...
}
```

Install the capture cleanup before any fallible selection operation, and represent the selected PSK with an owning zeroizing type rather than an untracked raw-array value. Add poison/observing allocator or memory-pattern tests for success, fallback, resolver failure, binder failure, async cancellation, malformed ServerHello, and teardown.

## 8. High: the public wire writers can emit illegal structures or trap on caller input

`writeModes()` accepts an empty list and does not reject more than 255 entries. `writeOffer()` does not validate:

- non-empty identities;
- identity length <= 65535;
- identities-vector total <= 65535;
- digest/binder length in 32..255;
- binders-vector total <= 65535.

Consequently these public codec functions can create wire data that their own parser rejects, truncate a patched vector length, trigger an `@intCast` safety trap, or slice beyond the fixed zero buffer.

Please preflight the complete structure with checked arithmetic before mutating the writer and return typed errors. Tests should cover zero, one-under, exact, one-over, and maximum values for every vector/entry length, plus proof that failure leaves no partially usable output.

## 9. High/acceptance: asynchronous credential selection cannot be declared out of scope

The PR description explicitly excludes async credential-selection/PSK interleaving, but #362 explicitly requires synchronous/asynchronous parity and reparsing of retained ClientHello bytes after resume. The implementation appears to route a completed `server_select` back through `emitServerHelloAndAuthFlight()`, but this path is untested and its error/cancellation cleanup currently retains the ClientHello capture.

Please remove this exclusion and add tests for:

- async selection completing with the matching credential and resuming;
- async selection completing with a rotated/nonmatching credential and falling back;
- async operation failure/cancellation clearing all captured ticket material;
- no resolver or binder call occurring before credential selection completes.

## 10. The required #362 test matrix is largely absent

The two end-to-end tests prove one happy resumption and one resolver miss, but they do not cover most of the issue's mandatory behavior. Before merge, please add at least:

- truncation and mutation of every modes/identity/binder vector length;
- checked-in exact SHA-256 and SHA-384 binder fixtures generated independently of the implementation (the current SHA-384 test only proves “nonzero”);
- a literal full ClientHello fixture proving the exact truncation offset and all patched enclosing lengths;
- unknown first / valid second identity;
- compatible first with bad binder / valid second: fatal `DecryptError` and no second resolver/binder probe;
- filtered-offer wire-index mapping;
- certificate A/B rotation fallback;
- client-auth-required and client-auth-optional full-handshake fallback;
- age-skew observation without 1-RTT rejection;
- application/transport/SNI/ALPN mismatch fallback;
- async credential-selection parity;
- allocation/failure/zeroization sweeps;
- resumed QUIC carrier coverage in addition to the direct record harness.

Also, the test named `ClientPskOfferSet is bounded to eight` currently pushes only one item and never tests eight or the ninth rejection.

Once these issues are addressed, the module boundaries and abbreviated key-schedule/state-machine approach should be in good shape.

Jaruso (commented):
**Re-review result: no updated implementation is present on GitHub.**

I re-fetched the PR metadata and complete current diff. The head is still exactly `08978b41f88e0a19e30081520b392b4c3a734f92`, the PR still contains one commit, and the source is unchanged from the commit reviewed in review #4739417705.

Accordingly, all previously documented merge-blocking findings remain open, including:

- server tickets are not bound to the currently selected server credential;
- filtered ClientHello identities are indexed against the unfiltered client ticket array;
- the selected client ticket is destroyed before compatibility validation/inheritance;
- PSK offer sizes are absent from ClientHello preflight;
- ticket-age skew is not surfaced;
- application compatibility and required client rechecks are missing;
- bearer-ticket/PSK cleanup is incomplete on error paths;
- public PSK writers lack complete semantic and length preflight;
- async credential-selection parity remains excluded despite the issue AC;
- the required negative, multi-identity, rotation, async, zeroization, and QUIC test matrix remains absent.

Please push the remediation commit(s) to this PR before requesting another review. A subsequent review should be performed against a head SHA different from `08978b4`.

Jaruso (commented):
**Review result at `63f101e`: changes are still required before merge.**

The latest remediation is real and correct as far as it goes. I rechecked the affected paths rather than carrying forward the prior findings:

- the client no longer compares the ticket's stored server transport snapshot with its own outbound transport parameters;
- the selected server credential is now validated across every chain entry before PSK resolution/binder acceptance;
- the full-handshake and PSK paths share the same validated certificate-message size;
- the extension-profile regression test proves resumption with asymmetric client/server transport payloads;
- handler-local cleanup now covers errors raised *inside* `onServerHello`, `onEncryptedExtensions`, and the post-capture part of `onClientHello`;
- CI run 846 is green.

A full pass against #362 still found the following blockers.

## 1. Blocker/security: terminal errors outside the message handlers still retain PSK state and schedules

The new `errdefer`s are installed inside `onServerHello()` and `onEncryptedExtensions()`. They only execute after `drainInput()` has successfully decoded the message, verified its epoch, accepted it through `core.acceptReceived()`, and dispatched it to the handler.

There are several terminal paths before that point:

```zig
input.append(bytes) catch ...;

while (input.peek() catch |err| return mapCoreError(err)) |message| {
    if (level != try expectedLevel(message.kind))
        return error.UnexpectedTransportEpoch;

    if (level == .handshake and
        message.kind == .finished and
        input.len != message.raw.len)

    _ = self.core.acceptReceived(message.raw) catch |err|
        return mapCoreError(err);

    try self.onMessage(...);
}
```

Concrete failures that bypass the new cleanup:

1. A client has owned `client_psk_offers`, but receives a Certificate, Finished, malformed handshake header, or wrong-epoch message instead of ServerHello. `onServerHello()` is never entered, so the offers remain owned by the failed backend.
2. A client accepts a PSK-selected ServerHello and moves the chosen ticket into `selected_client_psk`, but the peer then sends Certificate instead of EncryptedExtensions. `Core.acceptReceived()` rejects the ordering before `onEncryptedExtensions()` runs, so the selected ticket remains.
3. After ServerHello, the PSK-derived `KeySchedule` is installed. Any later terminal decode/order/epoch error leaves that schedule resident until some external owner eventually calls `deinit()`.
4. `resumeAuth()` failures can likewise terminate outside the handler-local cleanup boundaries.

The generic `engine.Driver` only records `.failed`; it intentionally does not deinitialize the backend immediately. Therefore “external teardown will eventually wipe it” is not sufficient for #362's requirement that every error path release offer, candidate, selected-session, binder, and schedule state.

Please centralize terminal cleanup at the backend call boundary rather than trying to enumerate individual handlers. For example:

```zig
fn clearPskHandshakeState(self: *Tls13Backend) void {
    self.client_psk_offers.deinit();
    self.selected_client_psk.deinit();
    self.selected_client_psk_present = false;
    self.clearClientHelloPsk();
    self.last_psk_age_skew = null;

    if (self.schedule) |*schedule| schedule.wipe();
    self.schedule = null;
}

fn receiveImpl(...) HandshakeError!void {
    errdefer self.clearPskHandshakeState();
    // existing receive/reassembly/dispatch path
}
```

Apply the same principle to terminal `startImpl()` and `resumeAuth()` errors, while preserving state when an async operation merely parks successfully.

There is also an unwiped raw client-side PSK copy in `onServerHello()`:

```zig
var buf: [hash_len]u8 = undefined;
@memcpy(&buf, psk_slice);
psk_secret = buf;
```

`psk_secret` is later wiped, but `buf` is a second stack copy and has no `defer secureZero`. Prefer one scoped array with an unconditional defer, or derive directly from an owning zeroizing value.

Required tests should inspect the backend *before calling `deinit()`* after:

- malformed/wrong-epoch input before ServerHello;
- an unexpected handshake message before ServerHello;
- valid PSK-selected ServerHello followed by Certificate or another ordering violation;
- malformed EncryptedExtensions framing;
- async resume failure;
- a sink/secret-export failure after the PSK schedule exists.

Assert that offers, selected tickets, captured ClientHello bytes, age observation, and the schedule are already cleared.

## 2. Blocker/architecture: the accepted server session is still discarded instead of moved into backend ownership

Issue #362 explicitly requires that, after the selected binder succeeds, the server move the recovered `ServerRecoverableState` into backend ownership and retain the selected identity index.

The current implementation still reduces the accepted candidate to only this stack value:

```zig
const PskSelected = struct {
    index: usize,
    psk: [hash_len]u8,
};
```

Inside `selectPsk()`:

```zig
var candidate_state: session.ServerRecoverableState = .{};
defer candidate_state.deinit();
...
return .{ .index = attempts, .psk = psk_buf };
```

Consequently, immediately after binder success, the backend destroys all recovered state other than the copied PSK and age-skew calculation. It does not retain:

- the selected identity index beyond the current call;
- the ticket's `early_data` policy and maximum;
- the selected session's owned compatibility metadata;
- any selected-session object that #365/#366 can consume without resolving the bearer identity a second time.

This is not just a checklist mismatch. #366 needs the selected ticket's early-data capability/limit together with the age observation, and #365's integration boundary should consume the already authenticated selection rather than recreate it.

Please make the selected result owning, for example:

```zig
const SelectedServerPsk = struct {
    index: u16,
    state: session.ServerRecoverableState = .{},

    fn deinit(self: *SelectedServerPsk) void {
        self.state.deinit();
        self.* = .{};
    }
};

selected_server_psk: SelectedServerPsk = .{},
selected_server_psk_present: bool = false,
```

After the binder succeeds, move `candidate_state` into that selected owner instead of allowing the iteration defer to destroy it. The key schedule can read the PSK from the owned state. Expose a bounded one-shot accessor or typed view for the later integration stories, and clear the selected state on fallback, terminal error, and teardown.

Add tests proving that the accepted state survives the resolver callback and contains the expected early-data/compatibility metadata, while every non-selected or failed candidate is wiped.

## 3. High/correctness: the generic transport-extension profile permits collisions with TLS-owned PSK extensions

`TransportProfile.validate()` rejects several extension IDs owned by the TLS backend, but it was not updated for the two IDs introduced by this PR:

```zig
pre_shared_key = 41
psk_key_exchange_modes = 45
```

The current rejection switch contains supported_groups, signature_algorithms, ALPN, supported_versions, and key_share, but not 41 or 45.

A caller can therefore configure `.extension.extension_type = 41` or `45`:

- type 41 is emitted as an opaque transport extension but parsed as `pre_shared_key`; with a real PSK offer it also duplicates the final PSK extension and violates the last-extension rule;
- type 45 is emitted as opaque transport data but parsed as `psk_key_exchange_modes`, preventing the payload from being captured as the required transport extension;
- the backend produces or consumes an ambiguous/malformed handshake instead of rejecting the local configuration before lifecycle/transcript mutation.

Reserve the new TLS-owned IDs in `TransportProfile.validate()`:

```zig
pre_shared_key.ext_pre_shared_key,
pre_shared_key.ext_psk_key_exchange_modes,
=> return error.InvalidTransportProfile,
```

This would also be a good place to centralize the complete engine-owned extension namespace, including the pre-existing `server_name` ID, so future TLS features cannot repeat this collision.

Add client and server start tests for extension types 41 and 45 proving `InvalidTransportProfile`, idle lifecycle, unchanged transcript/key-pair state, and no sink output.

## 4. High/lifetime: `setApplicationCompat()` permits its borrowed bytes to expire before they are used

The public contract currently says:

```zig
/// `blob.bytes` must outlive the handshake.
/// Must be called before `start`.
pub fn setApplicationCompat(...)
```

However, the backend retains the borrowed slice and reads it through `resumptionContext()` during **post-handshake** ticket processing:

- the client uses it when ingesting a later NewSessionTicket;
- the server uses it when `emitNewSessionTicket()` is called after handshake completion.

A caller following the documented contract may release or reuse the bytes as soon as the handshake completes, after which ticket issuance/ingestion reads dangling or mutated memory.

Please either:

- copy the bounded compatibility snapshot into backend-owned storage when configured; or
- move the compatibility input to the ticket issuance/ingestion boundary; or, at minimum, explicitly require it to remain immutable until backend teardown and every post-handshake ticket operation is finished.

Backend ownership is the safer contract because the snapshot is already bounded and is security-relevant compatibility state.

Add a test that configures compatibility from temporary/poisonable storage, completes the handshake, invalidates the source storage, and then processes/emits a ticket without observing the changed bytes.

## 5. Acceptance blocker: the required independent wire/fixture and failure-lifecycle matrix is still incomplete

The newly added end-to-end extension-profile and malformed-chain tests are valuable, but several tests explicitly required by #362 remain absent:

- mutation/truncation coverage for every identities-vector, identity-entry, binders-vector, and binder-entry length field;
- a checked-in exact SHA-384 binder fixture generated independently of this implementation — the current SHA-384 test only asserts the output is nonzero;
- an independently fixed SHA-256 expected binder rather than recomputing it with the same `std.crypto.tls.hkdfExpandLabel` helper used by production;
- a literal full ClientHello fixture proving the truncation offset and all patched outer lengths;
- all offers filtered out followed by a forged selected index;
- selected index exactly equal to the emitted count;
- large positive and negative age skew completing 1-RTT while `takePskAgeSkew()` returns the exact one-shot observation;
- optional, not only required, client-auth configuration forcing full-handshake fallback;
- PSK-specific allocation-failure and poison/zeroization sweeps covering offer planning, resolver output, selected-state movement, binder failure, and terminal cleanup.

These tests are important here because the remaining defects are primarily at ownership and framing boundaries that ordinary happy-path CI does not exercise.

Once the four implementation issues and the mandatory test gaps above are addressed, the PSK negotiation/key-schedule/state-machine design should be ready for merge.

Jaruso (commented):
**Review result: changes still required before merge.**

I re-reviewed head `b41f0d24d088d9b0a1bcb52a188b505808291f94` against #362. CI 847 is green, and this commit correctly resolves the four findings from the previous round: TLS-owned extension collisions are rejected, the accepted server session is retained, application compatibility is copied rather than borrowed, and receive/resume cleanup is now installed at a central boundary.

The following issues remain.

## 1. Blocker/security: `handshake_lifecycle == .complete` is not proof that terminal authentication succeeded

`clearPskHandshakeState()` returns immediately when `core.handshake_lifecycle == .complete`. Unfortunately, the shared `Core` transitions to `.complete` *before* the backend has finished authenticating or committing the handshake:

- On the server, `Core.acceptReceived(client Finished)` updates the transcript and marks the lifecycle complete before `onClientFinished()` verifies the Finished MAC. A bad client Finished therefore returns `DecryptError`, but the receive-level `errdefer` calls a cleanup function that sees `.complete` and does nothing.
- On the client, `sendClientFinished()` calls `core.recordSent()` before emitting the Finished. `recordSent()` marks the client complete. If `sink.emitCrypto()`, `captureResumptionMasterSecret()`, key discard, or `emitHandshakeComplete()` then fails, cleanup again sees `.complete` and does nothing.
- The server has the same post-verification terminal window: after a valid client Finished, `captureResumptionMasterSecret()`, key discard, or completion-event emission may fail after `Core` has already transitioned.

This leaves a failed connection able to expose `selected_server_psk` and `last_psk_age_skew`; it can also retain the PSK schedule and, when failure occurs after RMS capture, the resumption master secret. `core.psk_authenticated` and the core lifecycle also continue to describe the failed handshake as successfully resumed.

Please separate **core message ordering** from **backend handshake commitment**. For example:

```zig
handshake_committed: bool = false,

fn clearFailedHandshakeState(self: *Tls13Backend) void {
    if (self.handshake_committed) return;

    self.client_psk_offers.deinit();
    self.selected_client_psk.deinit();
    self.selected_client_psk_present = false;
    self.selected_server_psk.deinit();
    self.selected_server_psk_present = false;
    self.last_psk_age_skew = null;
    self.clearClientHelloPsk();

    if (self.schedule) |*schedule| schedule.wipe();
    self.schedule = null;
    self.resumption_master_secret.deinit();

    self.core.psk_authenticated = false;
    self.core.handshake_lifecycle = .failed;
}
```

Set `handshake_committed = true` only after every terminal fallible operation has succeeded:

```zig
fn completeClientHandshake(self: *Tls13Backend, sink: *EventSink) HandshakeError!void {
    try self.captureResumptionMasterSecret();
    try self.emitDiscardKeys(sink, .handshake);
    try sink.emitHandshakeComplete();
    self.finish();
    self.handshake_committed = true;
}
```

Do the equivalent at the end of `onClientFinished()`. Post-handshake input errors can then preserve the successfully committed selection, while failures during the Finished transition cannot.

Required regression tests:

- Resume through the server flight, corrupt the client Finished, and assert `DecryptError`, `takeSelectedServerPsk() == null`, `takePskAgeSkew() == null`, `schedule == null`, RMS empty, and the core no longer reports successful PSK authentication.
- Force event-sink failure after the client has recorded its outbound Finished and assert the same cleanup on the client.
- Force failure after server RMS capture but before completion-event commitment and assert RMS/selection cleanup.
- After a genuinely committed handshake, inject a malformed post-handshake NST and prove the already-successful selection remains available until taken.

The newly added cleanup test exercises only a wrong epoch *before ServerHello*; it cannot detect this terminal-completion hole.

## 2. High: application compatibility is capped by the unrelated 512-byte transport-extension limit

The shared session model supports application compatibility up to 1,024 bytes under `session.Limits.default` and up to `session.hard_max_compat_len` (8,192 bytes). The new backend storage is instead:

```zig
application_compat_bytes: [max_transport_extension_len]u8
```

where `max_transport_extension_len == 512`.

As a result, an application snapshot that is valid under the default session model—e.g. 513–1,024 bytes—cannot be configured on the backend and can never match or be stamped into a resumed session. This couples an application/H3 compatibility contract to the QUIC transport-extension payload cap and makes the #360/#362 interfaces disagree.

Please either:

- size the owned buffer to `session.hard_max_compat_len` and enforce a caller-selected application limit separately; or
- own a `session.CompatSnapshot`/bounded allocation and accept an allocator plus the same `session.Limits` used for ticket state.

Whichever representation is chosen, replacement and `null` should wipe the previous bytes. Add boundary tests proving the default 1,024-byte application snapshot succeeds, the configured limit is honored exactly, and one byte over fails without modifying the previous value.

## 3. High: the new pre-start configuration APIs do not enforce their documented lifecycle

`setClientPskOffers()`, `setServerPskResolver()`, and `setApplicationCompat()` all say they must be called before `start`, but none checks `core.handshake_lifecycle == .idle`. `setClientPskOffers()` also returns `void`.

This is not only defensive API polish. Replacing `client_psk_offers` after ClientHello emission changes the array that `onServerHello()` uses to interpret `selected_identity`, reintroducing the exact transmitted-index/wrong-PSK failure fixed earlier. Swapping resolver or application compatibility while asynchronous credential selection is pending similarly changes the selection policy halfway through one captured ClientHello.

Please make the setters fallible and atomic:

```zig
pub fn setClientPskOffers(
    self: *Tls13Backend,
    offers: *pre_shared_key.ClientPskOfferSet,
    now_ctx: *anyopaque,
    now_fn: *const fn (*anyopaque) i64,
) HandshakeError!void {
    if (self.role != .client or self.core.handshake_lifecycle != .idle)
        return error.InvalidHandshakeState;

    self.client_psk_offers.deinit();
    self.client_psk_offers.moveFrom(offers);
    self.psk_now_ctx = now_ctx;
    self.psk_now_fn = now_fn;
}
```

Apply the same lifecycle rule to the resolver and application snapshot. Tests should attempt each mutation after `start()` and while async server selection is pending, confirm `InvalidHandshakeState`, and confirm the original offers/resolver/context remain unchanged.

## 4. Required #362 acceptance coverage is still incomplete

The latest commit adds useful regression tests, but mandatory protocol/security fixtures remain absent:

- an exact, independently generated SHA-384 binder expected value—the current SHA-384 test only asserts that output is nonzero;
- truncation/mutation coverage for every identities/binders/modes vector length field and entry length;
- a checked-in literal full ClientHello fixture proving all enclosing lengths and the exact binder truncation boundary;
- backend-level large positive/negative age-skew tests proving 1-RTT still resumes and the one-shot observation contains the exact values;
- PSK-specific allocation-failure and zeroization sweeps, including resolver partial-output/error, bad Finished, async cancellation, and terminal sink failure.

These are explicitly required by issue #362 and are important here because the remaining cleanup defect occurs in a state transition that happy-path and early-parse-error tests do not exercise.

Jaruso (commented):
**Review result: one implementation blocker remains, plus the mandatory #362 test matrix is still incomplete.**

The latest remediation correctly fixes all three findings from the previous review:

- terminal cleanup now uses a backend-owned `handshake_committed` boundary rather than trusting `Core.handshake_lifecycle`;
- application-compatibility storage now reaches the shared session model's hard limit;
- the PSK/configuration setters reject post-start mutation without taking ownership or replacing prior configuration.

CI run 848 is green.

## 1. Blocker/security: `takeSelectedServerPsk()` can move secret state out before Finished authenticates the handshake

`selected_server_psk` is populated immediately after the binder succeeds, while the server is still waiting for the client's Finished. The public accessor currently checks only:

```zig
if (!self.selected_server_psk_present) return null;
```

It does **not** require `handshake_committed`. This contradicts the field/accessor documentation saying the state is available once the resumed handshake completes.

The current test `the accepted server session survives PSK selection with its early-data and metadata intact` makes the problem explicit: `driveServerSelection()` only feeds ClientHello and emits the server flight, then the test immediately moves the recovered `ServerRecoverableState` out through `takeSelectedServerPsk()` before any client Finished has been received.

If the client then sends a bad Finished, `clearFailedHandshakeState()` cannot wipe that selected state because backend ownership has already been transferred to the caller. This violates #362's acceptance criterion that selected state is wiped/deinitialized on every failure path. It also gives #365 a secret-bearing session/PSK that appears usable before the peer has completed the resumed handshake.

Please gate the owning accessor on the committed boundary:

```zig
pub fn takeSelectedServerPsk(
    self: *Tls13Backend,
    out: *session.ServerRecoverableState,
) ?u16 {
    if (!self.handshake_committed or !self.selected_server_psk_present)
        return null;

    out.moveFrom(&self.selected_server_psk.state);
    self.selected_server_psk_present = false;
    const index = self.selected_server_psk.index;
    self.selected_server_psk.index = 0;
    return index;
}
```

Required regression tests:

1. after binder acceptance/server-flight emission but before client Finished, `takeSelectedServerPsk()` returns `null` and leaves the backend-owned state present;
2. after a valid client Finished and committed handshake, the accessor moves the complete state and remains one-shot;
3. after a bad client Finished, the accessor returns `null` and the allocator-backed PSK/compatibility state has been deinitialized;
4. the successful end-to-end PSK round trip consumes the retained state only after both drivers report completion.

If #366 needs information before Finished for a future 0-RTT decision, expose a separate copy-only/non-secret policy view rather than moving the owning `ServerRecoverableState` out. For example:

```zig
pub const PendingPskPolicy = struct {
    selected_identity: u16,
    early_data: session.EarlyDataPolicy,
    age_skew: pre_shared_key.AgeSkew,
};

pub fn pendingPskPolicy(self: *const Tls13Backend) ?PendingPskPolicy {
    if (!self.selected_server_psk_present or self.last_psk_age_skew == null)
        return null;
    return .{
        .selected_identity = self.selected_server_psk.index,
        .early_data = self.selected_server_psk.state.common.early_data,
        .age_skew = self.last_psk_age_skew.?,
    };
}
```

That permits an early-data policy decision without transferring the PSK or allocator-owned session state out of the backend before authentication finishes.

## 2. Required #362 tests remain incomplete

Issue #362 explicitly requires all tests from research Sections 8.1 through 8.7. Several security-critical groups are still absent:

### Wire/parser mutation matrix

The current unit tests cover a few representative malformed shapes, but not:

- identity entry length at zero, exact one, representable maximum, one-over/truncation;
- binder entry lengths 31, 32, 48, 255, and a non-representable/invalid 256 case;
- identities-vector and binders-vector lengths at zero, one-under, exact, one-over, maximum;
- mutation of every enclosing/entry length field;
- truncation at every byte of a checked-in complete fixture;
- duplicate modes/PSK and PSK-not-last through a literal extension fixture.

Please make this table-driven so future codec changes cannot silently reduce coverage.

### Independent binder fixtures and exact ClientHello truncation

The SHA-384 test still proves only that the output is nonzero:

```zig
try testing.expect(!std.mem.allEqual(u8, &binder, 0));
```

That does not detect a wrong label, wrong empty-hash input, wrong transcript hash, or another deterministic but incorrect result. Add checked-in independent SHA-256 and SHA-384 expected binder bytes generated outside this module.

Also add a literal framed ClientHello fixture that asserts:

- all handshake, extensions, PSK-extension, identities, and binders lengths are already final;
- `binder_vector_offset` points immediately before the **two-byte binders-vector length**, not merely before the binder payload;
- changing any patched enclosing length changes the binder;
- the implementation's binder matches independent output for the exact literal prefix.

### Backend age and failure/zeroization tests

The successful round-trip test does not assert the exact `takePskAgeSkew()` value. Add successful positive/negative skew and large-skew-with-1-RTT-success cases at the backend seam, not only the pure arithmetic helper.

The existing generic provider allocation-failure test does not exercise PSK offers, resolver-owned candidates, binder scratch, selected client/server state, or the PSK schedule. Add PSK-specific allocation-failure/teardown sweeps and verify ownership counters or observing allocator state after every injected failure.

The bad-binder test currently checks the error and resolver call count, but not the Section 8.6 requirement that a bad selected binder emits no ServerHello/secret events before failure. Retain the sink and assert it contains no output from the failed selection.

Once the owning selected-state seam is commit-gated and these explicit required fixtures/sweeps are present, I do not see another implementation blocker in the current delta.

Jaruso (commented):
**Review result: changes still required before merge.**

Reviewed head `bf9239f453beab3ca447f0f69816d9b39e2ca0ca` against #362 and the previous review. CI run 849 is green.

The previous ownership defect is fixed correctly: `takeSelectedServerPsk()` now refuses to transfer the secret-bearing recovered state until the client Finished has committed the handshake, and the bad-binder test now proves that no ServerHello/traffic-secret event is emitted before failure.

I found one remaining implementation mismatch plus the still-unmet, explicitly mandatory test matrix.

## 1. Blocker: `writeOffer()` does not implement the wire binder-length contract it claims to encode

The parser correctly accepts the RFC/#362 wire range:

```zig
if (blen < min_binder_len or blen > max_binder_len)
    return error.InvalidBinderLength;
```

where `max_binder_len == 255`.

The writer instead rejects every binder above `provider.max_digest_len` (48):

```zig
if (item.digest_len < min_binder_len or
    item.digest_len > provider.max_digest_len)
{
    return error.InvalidBinderLength;
}
```

This creates an asymmetric codec:

- a valid 255-byte `PskBinderEntry` can be parsed but cannot be encoded;
- the explicit #362 test requirement for lengths 31, 32, 48, 255, and 256 cannot pass through the writer;
- despite the comment saying only supported digest lengths 32/48 are emitted, the current condition also accepts 33 through 47, which are not digest lengths of either supported hash.

The protocol codec and the backend policy should be separated. A straightforward fix is to make the codec wire-generic and let the backend prepare only 32- or 48-byte entries:

```zig
if (item.binder_len < min_binder_len or
    item.binder_len > max_binder_len)
{
    return error.InvalidBinderLength;
}

const zeros = [_]u8{0} ** max_binder_len;
try w.bytes(zeros[0..item.binder_len]);
```

Then enforce the concrete ticket hash/digest length when constructing the backend's `OfferItem`. Alternatively, split this into a general `writeOfferedPsks()` accepting actual binder byte slices and a stricter client-placeholder helper accepting only the known hash lengths. What should not remain is a public OfferedPsks encoder whose accepted domain disagrees with its parser and the ticket's normative wire shape.

Required boundary tests:

```text
31 -> InvalidBinderLength, writer unchanged
32 -> success
48 -> success
255 -> success
256 -> unrepresentable/rejected, writer unchanged
```

## 2. Security hygiene: the new test helper copies the whole key schedule by value

`feedValidClientFinished()` currently does:

```zig
const schedule = server.schedule.?;
```

That creates another stack copy of the handshake secret, master secret, and both handshake traffic secrets and never wipes that copy. The helper only needs a borrow:

```zig
const schedule = &server.schedule.?;
var client_verify = tls_backend.KeySchedule.verifyData(
    &schedule.client_handshake_traffic,
    server.core.transcriptHash(),
);
```

This is test code, but #362 is specifically enforcing disciplined secret ownership and should not add an avoidable untracked schedule copy in its regression fixtures.

## 3. The explicit Sections 8.1–8.7 test requirement is still incomplete

This is not a request for general extra coverage. Issue #362 explicitly says to include **all** tests from those sections. The current tests still omit several named requirements:

### Wire/parser fixtures

The current codec tests cover a normal round trip, empty/truncated modes, one count mismatch, and a 31-byte binder. They do not yet cover:

- mutation of every identities/binders/entry length field to zero, one-under, exact, one-over, and maximum;
- truncation at every byte of a checked-in valid fixture;
- identity lengths 1, maximum, and truncated identity/age;
- binder lengths 32, 48, and 255 plus malformed/trailing binder-vector data;
- duplicate modes, duplicate PSK, and PSK-not-last alert mapping.

A minimal checked-in extension-data fixture for mutation tests can begin with:

```zig
// identities<7> { identity<1> "a", age=0 }
// binders<33>   { binder<32> zeros }
const valid = hexBytes(
    "0007" ++ "0001" ++ "61" ++ "00000000" ++
    "0021" ++ "20" ++ ("00" ** 32),
);
```

Every strict prefix shorter than `valid.len` should fail, and each length field should be mutated independently so one malformed field cannot hide another untested parser branch.

### Independent binder and resumed-secret goldens

The SHA-256 test recomputes the same chain with Tardigrade's own `tls.hkdfExpandLabel`, so it is a same-implementation cross-check rather than an independent fixture. The SHA-384 test still asserts only that the result is nonzero.

For the inputs already used by the current tests, independent Python `hashlib`/`hmac` calculations produce these fixed binders:

```zig
// SHA-256:
// psk = c1392efd...f919fed
// prefix = "pretend-truncated-clienthello-bytes"
const expected256 = hexBytes(
    "de7c7afec445f9419e4f769b6ef8e637" ++
    "1e1f599405eff6ecc014499f234a008e",
);

// SHA-384:
// psk = 0x77 repeated 48 times
// prefix = "prefix"
const expected384 = hexBytes(
    "138b767e68513f232636a0d2b2c53d7a" ++
    "923ff2c0d7879985d4ea916281c5134d" ++
    "3bc2b1ae31178e736fe22f2d906cfdd3",
);
```

The current `initWithPsk` test likewise proves only determinism/divergence. For its existing inputs (`psk=0x99*32`, `shared=0x42*32`, transcript=`0x24*32`), independent goldens are:

```text
handshake_secret:         ab0803d6203c8feddfe8adc74f986c9d89b817b3d4132fc55c866a3522d9ff49
client_handshake_traffic: e92139285417b6a9a54a7a9153f4b6dcce44b99cdc0937b83dfea5c79805c920
server_handshake_traffic: 739483d9d6a9508c73b4656de22fedd85a2a8d00e9a6ca1449d8cba678c94baf
master_secret:            abe96cce65361235f3126971c67760888b79d4c1724a6cb1e15f6d2ae128ff44
client_application:       d1ba0b1be9862f1bd4c3bcc0d53b5a98c6a4951c4bad19243051237bc735031c
server_application:       c7428c93109f1b656dcbf0971e5d1bad9c2d38b79420038b7e165a17c7f61fa1
```

Please freeze independent literals rather than deriving `expected` with the same helper under test.

### Backend seams and failure sweeps

Still missing:

- a literal full ClientHello fixture proving the binder prefix ends before the binders-vector length while every outer length is already final;
- a genuine client path accepting a later selected identity and rejecting selected indexes equal to/outside the emitted count;
- resolver-attempt-cap coverage with more than eight wire identities;
- successful positive and negative `takePskAgeSkew()` observations, including large skew that still completes 1-RTT;
- PSK-specific allocation-failure sweeps for offer setup/resolver-owned state;
- zeroization/ownership checks across success, fallback, resolver failure, binder failure, malformed ServerHello, async cancellation, and teardown.

The generic full-handshake allocator/bridge cleanup tests are valuable but do not exercise the allocator-backed ticket/compatibility state and PSK-specific failure branches introduced here.

Once the writer's contract is corrected and these ticket-mandated fixtures/sweeps are present, I did not identify another protocol implementation blocker in the current head.

Jaruso (commented):
**Re-review result: implementation changes are correct, but #362's mandatory acceptance matrix is still incomplete.**

I reviewed the new head `670bbecc40b0a7aabfcc95957da5653fd970b0e3` against the previous review and the current issue body, including the incorporated Sections 8.1–8.7 test plan.

The latest remediation is sound:

- `writeOffer()` and `OfferedPsks.parse()` now agree on the RFC 8446 `PskBinderEntry<32..255>` wire range. The backend remains free to generate only its supported SHA-256/SHA-384 lengths.
- `feedValidClientFinished()` now borrows the live `KeySchedule` instead of creating an unwiped value copy.
- The checked-in SHA-256 binder, SHA-384 binder, and PSK-aware handshake/application-secret values are genuine fixed independent vectors. I independently recalculated the supplied values and they match.
- CI run 850 is green.

I did not find another protocol or ownership defect in the latest implementation delta. However, the issue explicitly requires **all tests from research Sections 8.1 through 8.7**, and the following concrete cases are still absent. These are not open-ended hardening suggestions; they are named acceptance requirements in #362.

## 1. Blocker/acceptance: no literal framed ClientHello binder fixture

The independent tests currently call `deriveBinder()` with strings such as `"pretend-truncated-clienthello-bytes"` and `"prefix"`. They prove the HKDF/HMAC implementation, but not the highest-risk integration contract: that the backend hashes the exact framed ClientHello prefix after every enclosing length is patched, ending **before the two-byte binders-vector length field**.

Add a checked-in complete ClientHello fixture containing at least:

- handshake header and final handshake length;
- final extensions-vector length;
- final PSK extension length;
- final identities-vector length;
- final binders-vector length;
- one or more binder placeholders/final binders.

The test should independently assert the numeric offset of the binders-vector length field, compare the exact truncated byte slice with a checked-in expected prefix, and verify the final binder literal. It should fail if truncation moves two bytes later to the binder payload.

## 2. Blocker/acceptance: wire mutation and exhaustive truncation matrix remains incomplete

`pre_shared_key.zig` has useful representative malformed tests, but not the required matrix:

- mutate every PSK length field to zero, one-under, exact, one-over, and maximum;
- identity length zero, one, maximum representable in the fixture, and truncated identity/age;
- identities-vector and binders-vector malformed/trailing/count-mismatch variants;
- binder lengths 31, 32, 48, 255, and malformed/truncated 256 representation where applicable;
- truncate checked-in modes and OfferedPsks fixtures at every byte boundary;
- mixed/unknown modes and all entry validation, including malformed later entries after an initially valid entry.

A compact table-driven helper is appropriate here, for example:

```zig
fn expectEveryTruncationRejected(
    fixture: []const u8,
    parseFn: anytype,
) !void {
    var cut: usize = 0;
    while (cut < fixture.len) : (cut += 1) {
        try std.testing.expectError(
            error.MalformedHandshake,
            parseFn(fixture[0..cut]),
        );
    }
}
```

The expected error may differ at semantically complete shorter encodings, so use explicit exceptions rather than weakening the assertions.

## 3. Blocker/acceptance: client selected-identity consistency boundaries are not all tested

Add direct client tests for:

- server selects a valid later identity, not only index zero after compaction;
- selected index equal to the emitted offer count;
- selected index greater than the emitted offer count;
- forged selected identity when every offer was filtered and no PSK extension was sent;
- selected ticket hash/suite inconsistency;
- PSK-selected ServerHello with missing or invalid key share.

Each failure must be `IllegalParameter`, clear all offer/selected PSK state immediately, and emit no client Finished/application-secret/completion event.

## 4. Blocker/acceptance: resolver cap and operational-failure behavior need explicit coverage

Add tests proving:

- exactly eight unusable identities may invoke the resolver and then cleanly fall back;
- a ninth wire identity is structurally validated but never resolved;
- a valid ninth identity does not bypass the deterministic eight-attempt cap;
- `ResolverFailed` is fatal and distinct from an ordinary miss;
- a resolver that partially populates `out` before returning an error is still fully deinitialized/wiped.

The last case is especially important for enforcing the resolver ownership contract defensively.

## 5. Blocker/acceptance: backend ticket-age seam is not exercised

The helper-level skew test does not prove the selected wire identity's age reaches the backend accessor. Add successful resumed-handshake tests for:

- exact age zero;
- normal elapsed age;
- positive skew;
- negative skew;
- client time before receipt saturating elapsed age to zero;
- large skew still completing ordinary 1-RTT resumption;
- `takePskAgeSkew()` returning the exact observation once and then `null`;
- rejected/fallback identities not publishing an observation.

## 6. Blocker/security acceptance: no PSK-specific allocation-failure and zeroization sweep

The issue explicitly requires allocation-failure and teardown coverage for offer, candidate, selected, binder, and schedule state. The general handshake allocator test does not force failures through the new PSK ownership paths.

At minimum, use `std.testing.checkAllAllocationFailures` around:

- constructing/moving multiple client offers;
- client ClientHello planning and emission;
- resolver cloning a candidate with all optional compatibility fields populated;
- successful selection and retained selected state;
- resolver failure after partial candidate construction;
- fallback after all candidates reject;
- bad binder;
- malformed ServerHello after offers were emitted;
- async selection cancellation/failure;
- successful resumed teardown.

Where allocator observation alone cannot prove wiping, assert the zero-valued/empty state after the failure and use the existing poison/observing patterns for buffers that remain inspectable.

## Additional still-named cases

The matrix also still calls for a changed-PSK binder test, optional-client-auth PSK fallback, and differential/OpenSSL-or-BoringSSL fixture coverage where practical.

Once the finite tests above are added, I would consider the current implementation itself ready based on this review. I found no unresolved code defect in the new head beyond the missing acceptance coverage.

Jaruso (commented):
Re-reviewed at `5f3d05aefa289592d3da4eb434d1957a0b708d4a`. CI 851 is green.

This commit closes most of the remaining #362 acceptance matrix correctly: the literal independently-generated binder fixture is valid, strict-prefix and nested-vector failures are exercised, non-zero/out-of-range selected identities are covered, the eight-attempt resolver ceiling is proven, resolver operational/partial-output failures are covered, and positive/negative/large ticket-age skew reaches the backend seam without rejecting 1-RTT.

I did not find a new protocol implementation defect. However, the explicit #362 test requirements are still not fully satisfied in three areas.

## 1. Required PSK zeroization sweeps still assert state transitions, not memory erasure

The issue requires allocation-failure **and zeroization sweeps**, and the PR claims every owned PSK/ticket/binder/schedule copy is wiped. The new tests verify that optionals become null, lengths become zero, or ownership flags clear, but none inspects the backing bytes after cleanup.

For example, `ClientPskOfferSet is bounded to eight, move-oriented, and fully wiped on deinit` calls `deinit`, but never proves that the allocator-backed ticket, nonce, or PSK bytes were overwritten. Similarly, the bad-Finished test proves the schedule is null and selected state is inaccessible, but not that the prior secret memory was zeroed.

Please add observable zeroization tests using a `FixedBufferAllocator`, whose `free` is a no-op, so the original backing slices remain inspectable after `deinit`:

```zig
var backing = [_]u8{0xa5} ** 4096;
var fba = std.heap.FixedBufferAllocator.init(&backing);
const allocator = fba.allocator();

var ticket = try makeClientTicket(
    allocator,
    .tls_aes_128_gcm_sha256,
    &psk,
    "wipe-client-ticket",
);

const psk_memory = ticket.common.resumption_psk.slice();
const ticket_memory = ticket.ticket.slice();
const nonce_memory = ticket.ticket_nonce.slice();

// Move the ticket into the backend, then trigger an out-of-range
// selected_identity / malformed ServerHello cleanup path.
...

try std.testing.expect(std.mem.allEqual(u8, psk_memory, 0));
try std.testing.expect(std.mem.allEqual(u8, ticket_memory, 0));
try std.testing.expect(std.mem.allEqual(u8, nonce_memory, 0));
```

Please cover at least these PSK-owned paths:

- client offer cleanup after malformed/out-of-range ServerHello;
- resolver candidate cleanup after a bad binder;
- selected server session plus key schedule on backend teardown;
- async credential-selection failure wiping the captured ClientHello bytes, not only setting `client_hello_psk = null`.

For the schedule teardown check, capture the inline storage before teardown:

```zig
const schedule_memory = std.mem.asBytes(&server.schedule.?);
server.deinit();
try std.testing.expect(std.mem.allEqual(u8, schedule_memory, 0));
```

## 2. Client selected-identity coverage still omits selected suite/hash/key-share consistency

The new client tests cover selected identity index consistency, but the #362 acceptance criterion is broader: `selected index/hash/key-share consistency is validated by the client`.

The current test-only `ServerHelloOptions` always emits TLS 1.3, suite `0x1301`, and a valid X25519 share, so the PSK-selected path still lacks tests for:

- selected cipher suite/hash mismatch;
- missing key share;
- wrong key-share group;
- wrong key-share length;
- low-order/all-zero X25519 share.

Please extend the helper rather than hand-editing offsets in each test:

```zig
const ServerKeyShareCase = enum {
    valid,
    missing,
    wrong_group,
    wrong_length,
    low_order,
};

const ServerHelloOptions = struct {
    session_id: []const u8 = &.{},
    key_share_seed: [X25519.seed_length]u8 = [_]u8{0x55} ** X25519.seed_length,
    selected_identity: ?u16 = null,
    selected_version: u16 = 0x0304,
    cipher_suite: u16 = 0x1301,
    key_share: ServerKeyShareCase = .valid,
};
```

Then use a table-driven test with `selected_identity = 0`, expecting `IllegalParameter` or `MalformedHandshake` as appropriate, and assert after every case that:

```zig
try std.testing.expect(client.client_psk_offers.isEmpty());
try std.testing.expect(!client.selected_client_psk_present);
try std.testing.expect(!client.core.psk_authenticated);
try std.testing.expect(client.schedule == null);
```

This is important because it proves those generic ServerHello checks clean up the PSK-specific owning state when the server has attempted resumption.

## 3. The resolver allocation-failure test is bounded and can pass without proving an exhaustive sweep

The new test manually tries `fail_index < 32`, treats every error as an allocation failure, and does not assert that an injected OOM was actually observed or that the final non-failing execution succeeded. It could therefore pass if all iterations failed for an unrelated reason, and it does not automatically adapt when another allocation is added.

Please use the repository's normal exhaustive pattern:

```zig
try std.testing.checkAllAllocationFailures(
    std.testing.allocator,
    exerciseResolverCloneThroughBackend,
    .{ &stored_state, &psk },
);
```

The resolver should record whether its `cloneInto` saw `OutOfMemory`; the exercise helper should translate only that callback failure back into `error.OutOfMemory` so `checkAllAllocationFailures` can advance correctly:

```zig
server.backend().receive(.initial, hello, &sink) catch |err| {
    if (resolver.saw_oom and err == error.CredentialProviderFailed)
        return error.OutOfMemory;
    return err;
};

try std.testing.expect(server.core.psk_authenticated);
```

That gives a real success run after every allocation point has been failed and verifies that non-OOM failures are not silently accepted by the test.

### Minor fixture follow-up

The test named `OfferedPsks identity length boundaries: zero, one, max representable, truncated` does not actually exercise the maximum representable identity; either add that boundary or rename the test so it does not claim coverage it lacks.

Once the observable zeroization checks, PSK-selected ServerHello consistency matrix, and exhaustive allocation sweep are present, I do not currently see another implementation blocker.

Jaruso (commented):
**Re-review result: changes are still required before merge.**

Current head: `8cb8bf69be3bd97ce34e201c597fc6bc0f48277e`.

Two of the three findings from the previous round are now correctly addressed:

- The client-side PSK-selected ServerHello matrix now covers wrong suite, wrong selected version, missing key share, wrong group, wrong X25519 length, and a low-order point, and verifies PSK state/schedule cleanup.
- The resolver-clone allocation test now uses `std.testing.checkAllAllocationFailures`, translates only an actually observed injected OOM back to `error.OutOfMemory`, and includes the eventual successful run.

However, CI run **852** is red: `Test appliance profile (ubuntu-latest)` fails in `Run appliance profile unit tests`. The other visible jobs are green, so the PR cannot merge in its current state.

I also found that the new zeroization tests are not yet a sound proof of the required property.

## 1. The FixedBufferAllocator tests can pass because `Allocator.free` destroys the evidence

The tests say that `FixedBufferAllocator.free` is a no-op and therefore the backing memory proves that this code explicitly wiped the secret. But the value being used is the generic `std.mem.Allocator` interface:

```zig
var fba = std.heap.FixedBufferAllocator.init(&backing);
const allocator = fba.allocator();
```

The commit itself later acknowledges that Zig's `Allocator.free()` fills freed memory with `undefined`. That generic front-end behavior still applies even when the allocator's raw free operation does not reclaim the fixed buffer.

Consequently these assertions do **not** prove that `secureZero` ran:

```zig
try expect(!std.mem.eql(u8, ticket_memory, "wipe-client-ticket"));
try expect(std.mem.indexOf(u8, &backing, marker) == null);
```

Both can succeed if the production `secureZero` call is deleted and the value is merely freed/poisoned. The comments in the bad-binder test are internally contradictory: they correctly mention that the generic `Allocator.free` front-end writes an undefined pattern, but then treat the FBA backing scan as evidence of the pre-free contents.

The embedded `FixedSecret` assertions are useful because those arrays are not separately freed. The allocator-backed ticket identity and compatibility blob assertions are not.

## 2. Two tests inspect inactive optional payloads through stale slices

The async-capture test does this:

```zig
const capture = &server.client_hello_psk.?;
const message_memory = capture.message[0..capture.message_len];
try server.resumeAuth(&sink); // clearClientHelloPsk sets the optional to null
try expect(allZero(message_memory));
```

Likewise the teardown test retains:

```zig
const schedule_memory = std.mem.asBytes(&server.schedule.?);
server.deinit(); // schedule is wiped and then assigned null
try expect(allZero(schedule_memory));
```

Once `client_hello_psk` or `schedule` is assigned `null`, those slices point into a payload that is no longer the active value of the optional. Treating the address as stable and continuing to read it is not a portable observation contract and may differ across optimization/profile combinations. The selected-server-state slice is different because that state is a non-optional embedded field and remains valid to inspect after its `deinit`.

I cannot identify the exact assertion from the connector's truncated log tail, but these new tests are profile-sensitive by construction and must be fixed independently of the CI failure's exact line.

## Recommended testable structure

Separate each wipe operation from the state transition/free operation and test the wipe while the storage is still live:

```zig
fn wipeClientHelloPskBytes(self: *Tls13Backend) void {
    if (self.client_hello_psk) |*capture| {
        crypto.secureZero(u8, capture.message[0..capture.message_len]);
    }
}

fn clearClientHelloPsk(self: *Tls13Backend) void {
    self.wipeClientHelloPskBytes();
    self.client_hello_psk = null;
    self.offered_psk_modes_seen = false;
    self.offered_psk_dhe_ke = false;
}
```

A unit test in `tls13_backend.zig`, where the private helper is accessible, can call `wipeClientHelloPskBytes()`, inspect the still-active optional payload, and then call `clearClientHelloPsk()`. Integration tests should continue to assert that each failure path leaves the optional null.

Apply the same pattern to the schedule:

```zig
fn wipeSchedule(self: *Tls13Backend) void {
    if (self.schedule) |*schedule| schedule.wipe();
}

fn clearSchedule(self: *Tls13Backend) void {
    self.wipeSchedule();
    self.schedule = null;
}
```

Test `wipeSchedule()` while `schedule.?` is still active, then use path-level tests to prove every terminal path calls `clearSchedule()`.

For allocator-backed secret-bearing types, expose/test an internal wipe primitive before free rather than examining poisoned freed memory:

```zig
fn wipe(self: *BoundedSecret) void {
    crypto.secureZero(u8, self.bytes);
}

fn deinit(self: *BoundedSecret) void {
    self.wipe();
    // free allocator-backed members, then zero the descriptor
}
```

Then test `wipe()` directly while the allocation remains live, and separately test that `deinit()` leaves the descriptor empty and does not leak. The same approach applies to allocator-backed ticket identities and compatibility snapshots.

After the failing appliance test is corrected and the zeroization tests observe live storage rather than allocator poisoning/inactive optional payloads, I do not see another remaining protocol implementation blocker in this head.

Jaruso (commented):
## Final re-review: merge-ready

Reviewed head `7279980f942cba165664911daa458038287c0615` against the remaining findings from review `4745830361` and the full #362 acceptance work reviewed in the preceding rounds.

The last blockers are resolved:

- CI run 854 completes successfully.
- `ClientHelloPskCapture.wipe()` is now tested directly on a live, non-optional value, avoiding stale references into an optional after it becomes `null`.
- `KeySchedule.wipe()` is likewise tested directly on a live schedule, proving all derived-secret storage is zeroed without relying on optional-payload behavior.
- Backend async-failure and teardown tests now assert the ownership/state transitions that invoke those isolated wipe operations rather than inspecting invalidated optional payloads.
- Client offer cleanup still directly observes embedded PSK and nonce arrays being zeroed and verifies allocator-backed ticket ownership metadata is cleared.
- Resolver-candidate cleanup is validated through allocator bookkeeping rather than post-`free` memory contents.
- The ServerHello suite/version/key-share consistency matrix and exhaustive resolver allocation-failure sweep from the prior commit remain correct.

I found no remaining protocol, state-machine, ownership, cleanup, or required acceptance-test defect. The PR is ready to merge.

PR #480, opened 2026-07-21
tls: add bounded client/server session-resumption cache (#364)

Jaruso:
## Summary

Implements the cache-owned scope of #364 that can proceed independently of #363 (the sibling stateless-ticket-protector PR, #478, still draft):

- `src/tls/session_cache.zig`
  - `ClientSessionCache` — bounded, origin-keyed client ticket store: deep-clone ownership, duplicate-identity replacement, deterministic per-origin/global LRU eviction, capacity + byte accounting, expiry re-check at lookup and cleanup, deterministic offer ordering (newest sequence → newer `received_at` → entry ID). **Explicitly reusable-only for this PR**: `storeClone` rejects `.single_use` with a typed `.rejected_unsupported_usage` result rather than silently storing and re-offering it as reusable, and `restoreClones` drops any `.single_use` records found in a persisted snapshot. The client offer-lease API (pinning a selected single-use candidate between offer and commit) is deferred to #365, to be designed alongside the real ClientHello-selection wiring.
  - `StatefulServerCache` — bounded store keyed by a random 40-byte `TDSH` v1 opaque handle: bounded CSPRNG collision retry with a typed failure, an internal lease/commit/release model for single-use consumption (with generation-tagged ABA protection across reload), deterministic eviction that never touches leased entries, and the same origin/capacity/byte accounting as the client cache. Bearer handles are stored only inside individually heap-allocated, individually wiped `ServerEntry` records — never inline in general-purpose hashmap storage — indexed by a non-secret digest.
  - Deterministic sequence renumbering on overflow (never wrapping): allocation-free by construction (in-place sort for the client's array-backed store; a two-pass rank computation for the server's map-backed store), so it can never fail partway through and never mixes pre-/post-renumber values within one batch reservation.
  - Persistence save/load on the stateful cache are serialized end-to-end (including backend I/O) via a token-based begin/end guard, so overlapping saves/loads can never resurrect an already-consumed single-use ticket, and a new single-use lease is refused for the duration of an in-flight operation.
  - Canonical SHA-256 origin digest over cipher suite / SNI / ALPN / auth binding / transport+application compat (never ticket identity, PSK, nonce, issue time, lifetime, or early-data policy); a digest match is always re-verified via `session.evaluateCompatibility`.
  - Non-secret `Observer`/`CacheEvent` metrics seam, invoked only outside the cache mutex.
  - Thread safety via `zig_compat.Mutex` (the repo's `std.Io.Mutex`-backed replacement for the no-longer-present `std.Thread.Mutex`), not a spin lock.
- `src/tls/session_cache_persistence.zig`
  - Versioned snapshot framing reusing `session.encodeClient`/`encodeServer`, a `Protector` seam (seal/open — no built-in adapter writes the sensitive plaintext to disk directly) and a `Backend` seam (load/save; atomic replacement is the backend's job).
  - Load path: decode into a temporary cache enforcing current limits and discarding expired entries, then atomic in-memory swap (via `adoptFromLocked` on both caches) only after full success. Persistence failures never touch the live cache.

### Deliberately deferred (per issue #364's canonical plan, pending #363/PR #478)

- No changes to `ticket_protection.zig`, `root.zig`, or the public `pre_shared_key.ServerPskResolver` contract.
- No `TDSH`/`TDTK` composite adapter or stateful/stateless parity integration yet.
- Client single-use support: this PR ships reusable client tickets only, plus the fully-implemented stateful-server lease API; the client offer-lease/selected-index commit primitive is deferred to #365, matching the issue's fallback guidance.

Both new files are intentionally unwired from `root.zig`/`tls_core_mod` for now (same pattern PR #478 already uses for `ticket_protection.zig`), with their own standalone `build.zig` test modules/step (`test-session-cache`) so they're still fully covered by `zig build test`.

## Test plan

- [x] `zig build test-session-cache --summary all` (389/389)
- [x] `zig build test-tls --summary all`
- [x] `zig build test --summary all` (2271/2272, 1 pre-existing unrelated skip)
- [x] Exhaustive coverage: exact/one-over capacity, per-origin and key-cardinality pressure, byte-limit eviction, deterministic LRU ordering (incl. batch-reservation and renumbering-at-overflow regressions), duplicate replacement, single-use pin/commit/release (incl. concurrent-resolution miss and ABA-safe stale-generation no-ops across reload), bounded handle-generation collision exhaustion, allocation-failure sweeps (`FailingAllocator`) against populated/eviction-pressured caches proving exact preserved state, zeroization on eviction/deinit, persistence round trip/corruption/truncation/version/kind-mismatch/atomic-swap/protection-failure/overlapping-operation-rejection.

Closes part of #364 (cache-owned scope only — see deferred items above).

### DIFF
--- a/src/tls/session_cache.zig
+++ b/src/tls/session_cache.zig
@@ -585,11 +585,11 @@ pub const ClientSessionCache = struct {
         const new_bytes = clientAccountedBytes(cloned);
         if (new_bytes > self.limits.max_entry_bytes) return .rejected_capacity;
 
-        const origin_count = self.countOrigin(origin);
-        if (origin_count == 0 and self.countDistinctOrigins() >= self.limits.max_origins) {
-            return .rejected_capacity;
-        }
-
+        // `planClientInsertionLocked` folds expiry purging and the
+        // new-origin/`max_origins` admission decision into the same
+        // non-mutating dry run as ordinary capacity eviction — see its
+        // doc comment for why an origin's cardinality must be judged
+        // against a *post-purge* view, not raw current counts.
         var plan = (self.planClientInsertionLocked(origin, new_bytes, now_unix_ms) catch return .storage_failed) orelse
             return .rejected_capacity;
 
@@ -848,11 +848,6 @@ pub const ClientSessionCache = struct {
         const new_bytes = clientAccountedBytes(ticket);
         if (new_bytes > self.limits.max_entry_bytes) return .rejected_capacity;
 
-        const origin_count = self.countOrigin(origin);
-        if (origin_count == 0 and self.countDistinctOrigins() >= self.limits.max_origins) {
-            return .rejected_capacity;
-        }
-
         var plan = (self.planClientInsertionLocked(origin, new_bytes, now_unix_ms) catch return .storage_failed) orelse
             return .rejected_capacity;
 
@@ -903,18 +898,17 @@ pub const ClientSessionCache = struct {
         }
     }
 
-    fn countOrigin(self: *ClientSessionCache, origin: OriginDigest) usize {
-        var n: usize = 0;
-        for (self.entries.items) |e| {
-            if (std.mem.eql(u8, &e.origin, &origin)) n += 1;
-        }
-        return n;
-    }
-
-    fn countDistinctOrigins(self: *ClientSessionCache) usize {
+    /// Distinct origins among entries that are *not* expired as of
+    /// `now_unix_ms` — an origin whose every entry has expired doesn't
+    /// count, since `planClientInsertionLocked` always plans removal of
+    /// every expired entry regardless of whether eviction pressure
+    /// otherwise requires it (see that function's doc comment).
+    fn countDistinctLiveOrigins(self: *ClientSessionCache, now_unix_ms: i64) usize {
         var n: usize = 0;
         outer: for (self.entries.items, 0..) |e, i| {
+            if (e.ticket.common.isExpired(now_unix_ms)) continue;
             for (self.entries.items[0..i]) |prev| {
+                if (prev.ticket.common.isExpired(now_unix_ms)) continue;
                 if (std.mem.eql(u8, &prev.origin, &e.origin)) continue :outer;
             }
             n += 1;
@@ -923,16 +917,32 @@ pub const ClientSessionCache = struct {
     }
 
     /// A non-mutating dry run: computes the exact set of entries a new
-    /// `new_bytes`-sized entry in `origin` would need to evict, without
-    /// touching any cache state. Called *before* any fallible allocation
-    /// is attempted, so `storeLocked`/`restoreLocked` know precisely what
-    /// they are about to do and can reserve exactly the right capacity
-    /// before mutating anything. There is no lease/pinning concept on the
-    /// client side, so — unlike the stateful server cache — every live
-    /// entry is always a valid eviction candidate and this can only fail
-    /// to find room when `new_bytes` itself exceeds the cache's total
-    /// byte budget (`null`) or on the plan's own scratch allocation
-    /// (`error.OutOfMemory`).
+    /// `new_bytes`-sized entry in `origin` would need this cache to evict
+    /// (`null` if there is no way to make room, e.g. `new_bytes` itself
+    /// exceeds the total byte budget, or `origin` would exceed
+    /// `max_origins`), without touching any cache state. Called *before*
+    /// any fallible allocation is attempted, so `storeLocked`/
+    /// `restoreLocked` know precisely what they are about to do and can
+    /// reserve exactly the right capacity before mutating anything.
+    ///
+    /// Every already-expired entry, anywhere in the cache, is
+    /// unconditionally included in the plan — not merely *preferred* as a
+    /// victim when eviction pressure happens to need one. This is what
+    /// lets the `max_origins` admission decision below be judged against
+    /// the cache's state *as it will be immediately after this insert
+    /// commits* (i.e. with every stale entry already gone) rather than
+    /// its raw current state: without this, an origin whose only entries
+    /// have expired but were never purged (nothing since #364's
+    /// deliberate removal of eager purging — see the module doc — forces
+    /// that to happen promptly) would permanently occupy an `max_origins`
+    /// slot forever, blocking every other origin indefinitely. Because
+    /// removal is still only *planned* here, not applied, this stays
+    /// fully compatible with the atomicity guarantee: a rejected or
+    /// later-failed insert leaves every expired entry exactly as it was.
+    ///
+    /// There is no lease/pinning concept on the client side, so — unlike
+    /// the stateful server cache — every live entry is always a valid
+    /// eviction candidate once expired entries are accounted for.
     fn planClientInsertionLocked(
         self: *ClientSessionCache,
         origin: OriginDigest,
@@ -944,19 +954,33 @@ pub const ClientSessionCache = struct {
         var victims: std.ArrayListUnmanaged(*ClientEntry) = .empty;
         errdefer victims.deinit(self.allocator);
 
-        var origin_count = self.countOrigin(origin);
-        var global_count = self.entries.items.len;
-        var global_bytes = self.total_bytes;
+        var origin_count: usize = 0;
+        var global_count: usize = 0;
+        var global_bytes: usize = 0;
+        for (self.entries.items) |e| {
+            if (e.ticket.common.isExpired(now_unix_ms)) {
+                try victims.append(self.allocator, e);
+                continue;
+            }
+            global_count += 1;
+            global_bytes += e.bytes;
+            if (std.mem.eql(u8, &e.origin, &origin)) origin_count += 1;
+        }
+
+        if (origin_count == 0 and self.countDistinctLiveOrigins(now_unix_ms) >= self.limits.max_origins) {
+            victims.deinit(self.allocator);
+            return null;
+        }
 
         while (origin_count >= self.limits.max_entries_per_origin) {
-            const victim = self.findEvictionCandidateInOriginExcluding(origin, victims.items, now_unix_ms) orelse unreachable;
+            const victim = self.findOldestLiveInOriginExcluding(origin, victims.items) orelse unreachable;
             try victims.append(self.allocator, victim);
             origin_count -= 1;
             global_count -= 1;
             global_bytes -= victim.bytes;
         }
         while (global_count >= self.limits.max_entries or global_bytes + new_bytes > self.limits.max_total_bytes) {
-            const victim = self.findEvictionCandidateGlobalExcluding(victims.items, now_unix_ms) orelse unreachable;
+            const victim = self.findOldestLiveGlobalExcluding(victims.items) orelse unreachable;
             try victims.append(self.allocator, victim);
             global_count -= 1;
             global_bytes -= victim.bytes;
@@ -980,22 +1004,17 @@ pub const ClientSessionCache = struct {
         plan.victims.deinit(self.allocator);
     }
 
-    /// Prefers an already-expired entry (its `lru_sequence` is irrelevant
-    /// — it would be purged regardless) over the genuine oldest-by-LRU
-    /// live entry, so ordinary capacity-driven eviction opportunistically
-    /// reclaims stale entries first without needing a separate eager
-    /// purge pass before this plan is computed.
-    fn findEvictionCandidateInOriginExcluding(
+    /// Oldest-by-LRU search among candidates that are neither already
+    /// planned as a victim nor expired. Every expired entry is already
+    /// unconditionally included in `excluding` by the time this is
+    /// called (see `planClientInsertionLocked`), so — unlike earlier
+    /// revisions of this search — no separate expiry preference is
+    /// needed here: no expired candidate can remain.
+    fn findOldestLiveInOriginExcluding(
         self: *ClientSessionCache,
         origin: OriginDigest,
         excluding: []const *ClientEntry,
-        now_unix_ms: i64,
     ) ?*ClientEntry {
-        for (self.entries.items) |e| {
-            if (!std.mem.eql(u8, &e.origin, &origin)) continue;
-            if (containsClientEntryPtr(excluding, e)) continue;
-            if (e.ticket.common.isExpired(now_unix_ms)) return e;
-        }
         var best: ?*ClientEntry = null;
         for (self.entries.items) |e| {
             if (!std.mem.eql(u8, &e.origin, &origin)) continue;
@@ -1005,15 +1024,10 @@ pub const ClientSessionCache = struct {
         return best;
     }
 
-    fn findEvictionCandidateGlobalExcluding(
+    fn findOldestLiveGlobalExcluding(
         self: *ClientSessionCache,
         excluding: []const *ClientEntry,
-        now_unix_ms: i64,
     ) ?*ClientEntry {
-        for (self.entries.items) |e| {
-            if (containsClientEntryPtr(excluding, e)) continue;
-            if (e.ticket.common.isExpired(now_unix_ms)) return e;
-        }
         var best: ?*ClientEntry = null;
         for (self.entries.items) |e| {
             if (containsClientEntryPtr(excluding, e)) continue;
@@ -1047,18 +1061,29 @@ pub const ClientSessionCache = struct {
         return self.reserveLruSequenceBatchLocked(1);
     }
 
-    /// Reserves `count` consecutive fresh LRU sequence values atomically
-    /// (renumbering first if the remaining range is too small to fit all
-    /// of them), returning the first. Calling `nextLruSequence` in a loop
-    /// instead — one reservation per touched entry — is exactly the bug
-    /// this exists to avoid: if renumbering happened partway through such
-    /// a loop, values obtained before and after it would be on two
-    /// different numbering scales, corrupting relative order among the
-    /// entries touched in the same batch.
+    /// Reserves `n` consecutive fresh LRU sequence values atomically
+    /// (renumbering first if the remaining range is too small), returning
+    /// the first. Calling `nextLruSequence` in a loop instead — one
+    /// reservation per touched entry — is exactly the bug this exists to
+    /// avoid: if renumbering happened partway through such a loop, values
+    /// obtained before and after it would be on two different numbering
+    /// scales, corrupting relative order among the entries touched in the
+    /// same batch.
+    ///
+    /// The guard reserves room for the batch *and* the value the counter
+    /// itself will hold immediately afterward, not just the batch's last
+    /// value: reserving only through `first + (n - 1)` (the last value
+    /// actually handed out) would let `next_lru_sequence` land exactly on
+    /// `maxInt(u64)` via saturating assignment, and a *later* reservation
+    /// would then see `next_lru_sequence > maxInt(u64) - m` fail to hold
+    /// (maxInt is not `>` itself) and skip renumbering — handing out
+    /// `maxInt(u64)` a second time. Renumbering here first instead keeps
+    /// every live sequence value unique.
     fn reserveLruSequenceBatchLocked(self: *ClientSessionCache, n: u64) u64 {
-        if (self.next_lru_sequence > std.math.maxInt(u64) - (n - 1)) self.renumberLruSequencesLocked();
+        std.debug.assert(n > 0);
+        if (self.next_lru_sequence > std.math.maxInt(u64) - n) self.renumberLruSequencesLocked();
         const first = self.next_lru_sequence;
-        self.next_lru_sequence = first +| n;
+        self.next_lru_sequence = first + n; // safe: the guard above proved no overflow.
         return first;
     }
 
@@ -1533,54 +1558,36 @@ pub const StatefulServerCache = struct {
         return result;
     }
 
-    /// Whether `bytes` could possibly fit without evicting any currently
-    /// leased entry — a pure, non-mutating preflight. Leased entries are
-    /// never eviction candidates, so if the cache could never get under
-    /// its count/byte/per-origin limits using only *unleased* entries as
-    /// victims, the insert must be rejected before anything is mutated.
-    fn canFitLocked(self: *StatefulServerCache, origin: OriginDigest, new_bytes: usize) bool {
-        var leased_count: usize = 0;
-        var leased_bytes: usize = 0;
-        var it = self.entries.iterator();
+    /// Distinct origins with at least one entry that would survive a
+    /// purge as of `now_unix_ms` — leased entries always survive
+    /// (regardless of expiry, since leased entries are never purged), an
+    /// unleased entry survives only if it is not yet expired. An origin
+    /// whose every entry has expired-and-is-unleased doesn't count, since
+    /// `planInsertionLocked` always plans removal of every such entry
+    /// regardless of whether eviction pressure otherwise requires it (see
+    /// that function's doc comment).
+    fn countDistinctLiveOriginsLocked(self: *StatefulServerCache, now_unix_ms: i64) usize {
+        var n: usize = 0;
+        var it = self.origin_index.iterator();
         while (it.next()) |kv| {
-            const e = kv.value_ptr.*;
-            if (e.active_lease_epoch != null) {
-                leased_count += 1;
-                leased_bytes += e.bytes;
-            }
-        }
-        if (leased_count >= self.limits.max_entries) return false;
-        if (leased_bytes + new_bytes > self.limits.max_total_bytes) return false;
-
-        if (self.origin_index.get(origin)) |bucket| {
-            var origin_leased: usize = 0;
-            for (bucket.items) |id| {
+            for (kv.value_ptr.items) |id| {
                 const e = self.entries.get(id) orelse continue;
-                if (e.active_lease_epoch != null) origin_leased += 1;
+                if (e.active_lease_epoch != null or !e.state.common.isExpired(now_unix_ms)) {
+                    n += 1;
+                    break;
+                }
             }
-            if (origin_leased >= self.limits.max_entries_per_origin) return false;
         }
-        return true;
+        return n;
     }
 
-    fn originBucketLen(self: *StatefulServerCache, origin: OriginDigest) usize {
-        return if (self.origin_index.get(origin)) |b| b.items.len else 0;
-    }
-
-    /// Prefers an already-expired, unleased entry (its `lru_sequence` is
-    /// irrelevant — it would be purged regardless) over the genuine
-    /// oldest-by-LRU unleased entry, so ordinary capacity-driven eviction
-    /// opportunistically reclaims stale entries first without needing a
-    /// separate eager purge pass before this plan is computed (see
-    /// `insertLocked`'s doc comment).
-    fn findOldestUnleasedInOriginExcluding(self: *StatefulServerCache, origin: OriginDigest, excluding: []const u64, now_unix_ms: i64) ?u64 {
+    /// Oldest-by-LRU search among unleased candidates that are neither
+    /// already planned as a victim nor expired. Every expired-and-unleased
+    /// entry is already unconditionally included in `excluding` by the
+    /// time this is called (see `planInsertionLocked`), so no separate
+    /// expiry preference is needed here: no expired candidate can remain.
+    fn findOldestUnleasedInOriginExcluding(self: *StatefulServerCache, origin: OriginDigest, excluding: []const u64) ?u64 {
         const bucket = self.origin_index.getPtr(origin) orelse return null;
-        for (bucket.items) |id| {
-            if (std.mem.indexOfScalar(u64, excluding, id) != null) continue;
-            const e = self.entries.get(id) orelse continue;
-            if (e.active_lease_epoch != null) continue;
-            if (e.state.common.isExpired(now_unix_ms)) return id;
-        }
         var best_id: ?u64 = null;
         var best_seq: u64 = undefined;
         for (bucket.items) |id| {
@@ -1595,18 +1602,11 @@ pub const StatefulServerCache = struct {
         return best_id;
     }
 
-    fn findOldestUnleasedGlobalExcluding(self: *StatefulServerCache, excluding: []const u64, now_unix_ms: i64) ?u64 {
-        var it = self.entries.iterator();
-        while (it.next()) |kv| {
-            if (std.mem.indexOfScalar(u64, excluding, kv.key_ptr.*) != null) continue;
-            const e = kv.value_ptr.*;
-            if (e.active_lease_epoch != null) continue;
-            if (e.state.common.isExpired(now_unix_ms)) return kv.key_ptr.*;
-        }
+    fn findOldestUnleasedGlobalExcluding(self: *StatefulServerCache, excluding: []const u64) ?u64 {
         var best_id: ?u64 = null;
         var best_seq: u64 = undefined;
-        var it2 = self.entries.iterator();
-        while (it2.next()) |kv| {
+        var it = self.entries.iterator();
+        while (it.next()) |kv| {
             if (std.mem.indexOfScalar(u64, excluding, kv.key_ptr.*) != null) continue;
             const e = kv.value_ptr.*;
             if (e.active_lease_epoch != null) continue;
@@ -1619,38 +1619,95 @@ pub const StatefulServerCache = struct {
     }
 
     /// Computes, without mutating anything, the exact set of entries this
-    /// insertion will need to evict (`canFitLocked` already guarantees
-    /// this is possible using only unleased victims), plus whether the
-    /// target origin's bucket will still exist afterward. Called *before*
-    /// any fallible allocation is attempted, so `insertLocked` knows
-    /// precisely what it is about to do and can reserve exactly the right
-    /// capacity before mutating anything.
+    /// insertion will need to evict, plus whether the target origin's
+    /// bucket will still exist afterward. Called *before* any fallible
+    /// allocation is attempted, so `insertLocked` knows precisely what it
+    /// is about to do and can reserve exactly the right capacity before
+    /// mutating anything.
+    ///
+    /// Every already-expired, unleased entry anywhere in the cache is
+    /// unconditionally included in the plan — not merely *preferred* as a
+    /// victim when eviction pressure happens to need one (leased entries
+    /// are never purged, matching `purgeExpiredAllLocked`'s own
+    /// discipline). This is what lets the `max_origins` admission
+    /// decision below be judged against the cache's state *as it will be
+    /// immediately after this insert commits*, rather than its raw
+    /// current state: without this, an origin whose only entries have
+    /// expired but were never purged would permanently occupy an
+    /// `max_origins` slot forever, blocking every other origin
+    /// indefinitely. Because removal is still only *planned* here, not
+    /// applied, this stays fully compatible with the atomicity guarantee:
+    /// a rejected or later-failed insert leaves every expired entry
+    /// exactly as it was.
     fn planInsertionLocked(self: *StatefulServerCache, origin: OriginDigest, new_bytes: usize, now_unix_ms: i64) error{OutOfMemory}!?EvictionPlan {
-        if (!self.canFitLocked(origin, new_bytes)) return null;
-
         var victims: std.ArrayListUnmanaged(u64) = .empty;
         errdefer victims.deinit(self.allocator);
 
-        var origin_count = self.originBucketLen(origin);
-        var global_count = self.entries.count();
-        var global_bytes = self.total_bytes;
+        var leased_count: usize = 0;
+        var leased_bytes: usize = 0;
+        var origin_leased: usize = 0;
+        var global_count: usize = 0;
+        var global_bytes: usize = 0;
+        var origin_count: usize = 0;
+
+        var it = self.entries.iterator();
+        while (it.next()) |kv| {
+            const e = kv.value_ptr.*;
+            const in_origin = std.mem.eql(u8, &e.origin, &origin);
+            if (e.active_lease_epoch != null) {
+                leased_count += 1;
+                leased_bytes += e.bytes;
+                if (in_origin) origin_leased += 1;
+                continue;
+            }
+            if (e.state.common.isExpired(now_unix_ms)) {
+                try victims.append(self.allocator, kv.key_ptr.*);
+                continue;
+            }
+            global_count += 1;
+            global_bytes += e.bytes;
+            if (in_origin) origin_count += 1;
+        }
+
+        // Leased entries are never eviction candidates: if the cache
+        // could never get under its count/byte/per-origin limits using
+        // only unleased entries as victims (this insert's own new entry
+        // included), reject before anything is mutated.
+        if (leased_count >= self.limits.max_entries) {
+            victims.deinit(self.allocator);
+            return null;
+        }
+        if (leased_bytes + new_bytes > self.limits.max_total_bytes) {
+            victims.deinit(self.allocator);
+            return null;
+        }
+        if (origin_leased >= self.limits.max_entries_per_origin) {
+            victims.deinit(self.allocator);
+            return null;
+        }
+
+        const is_new_origin = (origin_count + origin_leased) == 0;
+        if (is_new_origin and self.countDistinctLiveOriginsLocked(now_unix_ms) >= self.limits.max_origins) {
+            victims.deinit(self.allocator);
+            return null;
+        }
 
         while (origin_count >= self.limits.max_entries_per_origin) {
-            const victim = self.findOldestUnleasedInOriginExcluding(origin, victims.items, now_unix_ms) orelse unreachable;
+            const victim = self.findOldestUnleasedInOriginExcluding(origin, victims.items) orelse unreachable;
             try victims.append(self.allocator, victim);
             origin_count -= 1;
             global_count -= 1;
             global_bytes -= self.entries.get(victim).?.bytes;
         }
         while (global_count >= self.limits.max_entries or global_bytes + new_bytes > self.limits.max_total_bytes) {
-            const victim = self.findOldestUnleasedGlobalExcluding(victims.items, now_unix_ms) orelse unreachable;
+            const victim = self.findOldestUnleasedGlobalExcluding(victims.items) orelse unreachable;
             try victims.append(self.allocator, victim);
             global_count -= 1;
             global_bytes -= self.entries.get(victim).?.bytes;
             if (std.mem.eql(u8, &self.entries.get(victim).?.origin, &origin) and origin_count > 0) origin_count -= 1;
         }
 
-        return EvictionPlan{ .victims = victims, .origin_bucket_survives = origin_count > 0 };
+        return EvictionPlan{ .victims = victims, .origin_bucket_survives = (origin_count + origin_leased) > 0 };
     }
 
     /// Mutation order: no expiry purge and no LRU-sequence reservation
@@ -1684,11 +1741,11 @@ pub const StatefulServerCache = struct {
         const handle_digest = digestHandle(&handle);
         if (self.handle_index.contains(handle_digest)) return .rejected_capacity;
 
-        const is_new_origin = !self.origin_index.contains(origin);
-        if (is_new_origin and self.origin_index.count() >= self.limits.max_origins) {
-            return .rejected_capacity;
-        }
-
+        // `planInsertionLocked` folds expiry purging and the
+        // new-origin/`max_origins` admission decision into the same
+        // non-mutating dry run as ordinary capacity eviction — see its
+        // doc comment for why an origin's cardinality must be judged
+        // against a *post-purge* view, not raw current counts.
         var plan = (self.planInsertionLocked(origin, bytes, now_unix_ms) catch return .storage_failed) orelse
             return .rejected_capacity;
 
@@ -2377,6 +2434,78 @@ test "max_origins rejects a new origin once the distinct-origin cap is reached" 
     try testing.expectEqual(StoreResult.rejected_capacity, cache.storeClone(&t2, 1, .reusable));
 }
 
+test "max_origins reclaims an already-expired origin without an explicit cleanup() call" {
+    // Round-6 review #1 regression. `storeLocked` no longer purges
+    // expired entries eagerly (see round 5), but the `max_origins`
+    // admission check must still be judged against the cache's
+    // *post-purge* state, not raw current origin counts — otherwise an
+    // origin whose only entry has expired but was never purged would
+    // permanently occupy an `max_origins` slot, blocking every other
+    // origin indefinitely until some unrelated caller happens to invoke
+    // `cleanup()` first.
+    var limits = Limits.client_default;
+    limits.max_origins = 1;
+    var cache = try ClientSessionCache.init(testing.allocator, limits);
+    defer cache.deinit();
+
+    var t1 = try makeClient(testing.allocator, "t1", "a.test");
+    defer t1.deinit();
+    try testing.expectEqual(StoreResult.stored, cache.storeClone(&t1, 0, .reusable));
+
+    // Far past t1's lifetime — a.test's only entry is now expired, but
+    // nothing has purged it yet.
+    var t2 = try makeClient(testing.allocator, "t2", "b.test");
+    defer t2.deinit();
+    try testing.expectEqual(StoreResult.stored, cache.storeClone(&t2, 2_000_000, .reusable));
+
+    try testing.expectEqual(@as(usize, 1), cache.count());
+    try testing.expectEqualStrings("t2", cache.entries.items[0].ticket.ticket.slice());
+}
+
+test "FailingAllocator sweep: a failed client store never purges an already-expired origin's entry it needed to reclaim" {
+    // Companion to the atomicity sweep above: this specifically forces
+    // the plan to select the expired entry as a victim *because* the
+    // origin-cardinality check requires reclaiming it (max_origins = 1,
+    // new origin), not because of ordinary count/byte pressure. A late
+    // allocation failure must still leave it completely untouched.
+    var backing: [1 << 16]u8 = undefined;
+    var limits = Limits.client_default;
+    limits.max_origins = 1;
+
+    var found_a_failure = false;
+    var fail_index: usize = 0;
+    while (fail_index < 32) : (fail_index += 1) {
+        var fba = std.heap.FixedBufferAllocator.init(&backing);
+        var cache = try ClientSessionCache.init(fba.allocator(), limits);
+        defer cache.deinit();
+
+        var stale = try makeClient(fba.allocator(), "already-expired", "a.test");
+        _ = cache.storeClone(&stale, 0, .reusable);
+        stale.deinit();
+
+        const snapshot_bytes = cache.total_bytes;
+        const snapshot_count = cache.entries.items.len;
+
+        var failing = std.testing.FailingAllocator.init(fba.allocator(), .{ .fail_index = fail_index });
+        cache.allocator = failing.allocator();
+        var incoming = try makeClient(std.testing.allocator, "incoming", "b.test");
+        defer incoming.deinit();
+        const result = cache.storeClone(&incoming, 2_000_000, .reusable);
+        cache.allocator = fba.allocator();
+
+        if (result != .storage_failed) {
+            cache.deinit();
+            break;
+        }
+
+        found_a_failure = true;
+        try testing.expectEqual(snapshot_count, cache.entries.items.len);
+        try testing.expectEqual(snapshot_bytes, cache.total_bytes);
+        try testing.expectEqualStrings("already-expired", cache.entries.items[0].ticket.ticket.slice());
+    }
+    try testing.expect(found_a_failure);
+}
+
 test "max_entry_bytes rejects an oversized entry" {
     var limits = Limits.client_default;
     limits.max_entry_bytes = 8;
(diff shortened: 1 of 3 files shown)


src/tls/session_cache.zig
```diff
@@ -0,0 +1,1805 @@
...
+        now_unix_ms: i64,
+        usage: UsagePolicy,
+        evicted: *usize,
+    ) StoreResult {
+        self.purgeExpiredOrigin(origin, now_unix_ms, evicted);
+
+        for (self.entries.items) |*e| {
+            if (!std.mem.eql(u8, &e.origin, &origin)) continue;
+            if (!e.ticket.ticket.eql(&cloned.ticket)) continue;
+            self.total_bytes -= e.bytes;
```
Jaruso: **Blocking: duplicate replacement bypasses all byte-limit checks and destroys the old entry before success is known.** `new_bytes` is calculated only after `e.ticket` has been deinitialized and replaced, so a repeated opaque ticket with a larger `ticket_nonce` can push either `max_entry_bytes` or `max_total_bytes` over its configured bound. There is also no way to reject atomically because the previous value is already gone.

Please preflight the replacement before mutating anything, for example:

```zig
const replacement_bytes = clientAccountedBytes(cloned);
if (replacement_bytes > self.limits.max_entry_bytes)
    return .rejected_capacity;

const projected = self.total_bytes - e.bytes + replacement_bytes;
if (projected > self.limits.max_total_bytes)
    return .rejected_capacity; // or evict other eligible entries first

var old = e.ticket;
e.ticket = .{};
e.ticket.moveFrom(cloned);
e.usage = usage;
e.insertion_sequence = self.nextInsertionSequence();
e.lru_sequence = self.nextLruSequence();
e.bytes = replacement_bytes;
self.total_bytes = projected;
old.deinit();
```

Add a regression test that stores one identity with a short nonce, then replaces the same identity/origin with a nonce large enough to exceed `max_entry_bytes` and `max_total_bytes`; the original entry and all accounting must remain unchanged.

src/tls/session_cache.zig
```diff
@@ -0,0 +1,1805 @@
...
+            e.sequence = self.nextSequence();
+            e.bytes = clientAccountedBytes(&e.ticket);
+            self.total_bytes += e.bytes;
+            return .replaced;
+        }
+
+        const new_bytes = clientAccountedBytes(cloned);
+        if (new_bytes > self.limits.max_entry_bytes) return .rejected_capacity;
+
+        const origin_count = self.countOrigin(origin);
```
Jaruso: **Blocking: expired entries in other origins can permanently consume `max_origins`.** `storeLocked` only purges expired entries from the incoming origin. With `max_origins = 1`, an expired `a.test` entry causes a valid `b.test` insertion to be rejected forever because this check runs before any global expiry maintenance. The issue explicitly requires lifetime cleanup and periodic maintenance.

Please add a global expiry pass used before cardinality/global-pressure decisions, plus a public maintenance API:

```zig
fn purgeExpiredAllLocked(self: *ClientSessionCache, now: i64, removed: *usize) void { ... }

pub fn cleanup(self: *ClientSessionCache, now: i64) usize {
    var removed: usize = 0;
    self.mutex.lock();
    self.purgeExpiredAllLocked(now, &removed);
    self.mutex.unlock();
    // notify outside the lock
    return removed;
}
```

The server cache needs the equivalent behavior while skipping actively leased entries. Test a full origin table containing only expired entries, followed by insertion for a new origin without first looking up the expired origin.

src/tls/session_cache.zig
```diff
@@ -0,0 +1,1805 @@
...
+                }
+                i += 1;
+            }
+
+            std.mem.sort(Candidate, buf[0..n], {}, Candidate.moreRecent);
+
+            const take = @min(n, pre_shared_key.max_offered_identities);
+            for (buf[0..take]) |c| {
+                var clone: session.ClientTicketState = .{};
+                self.entries.items[c.idx].ticket.cloneInto(self.allocator, &clone) catch break;
```
Jaruso: **Blocking: this method implements insertion-order/FIFO eviction rather than LRU, and allocation failure is silently converted into a partial successful lookup.** `sequence` is changed only by store/replace, never by lookup, while the eviction helpers choose the smallest `sequence`. The #364 AC explicitly requires recency updates. At the same time, `cloneInto` failure just `break`s, potentially returning one or more offers and reporting `lookup_hit` instead of the required typed storage failure.

Selection order and eviction recency need separate fields because the canonical offer order is newest insertion first, while LRU must change on access:

```zig
const ClientEntry = struct {
    insertion_sequence: u64,
    lru_sequence: u64,
    ...
};

pub const ClientLookupResult = union(enum) {
    hit: pre_shared_key.ClientPskOfferSet,
    miss,
    expired,
    incompatible,
    storage_failed,
};
```

Build all clones in a temporary offer set. If any clone fails, deinit the temporary set and return `.storage_failed` without updating recency. After all clones succeed, update `lru_sequence` for the returned entries and commit the result. Add tests proving a lookup protects the touched entry from the next eviction and that a failure on clone N returns no partial offers.

src/tls/session_cache.zig
```diff
@@ -0,0 +1,1805 @@
...
+
+/// Bounded stateful server-side ticket/session store keyed by a random
+/// opaque handle. Process-shared and thread-safe (see module doc).
+pub const StatefulServerCache = struct {
+    allocator: std.mem.Allocator,
+    limits: Limits,
+    random: RandomSource,
+    observer: Observer = .{},
+    mutex: zig_compat.Mutex = .{},
+    entries: std.ArrayListUnmanaged(ServerEntry) = .empty,
```
Jaruso: **Blocking against the canonical design: this is an array scan, not the required primary `handle -> entry` index or secondary origin index.** Every unknown stateful identity causes a constant-time comparison against every live entry; #362 permits up to eight identities per ClientHello, so an attacker can force `O(8 * max_entries)` work per handshake. `countDistinctOrigins` and per-origin operations are also repeated scans (including quadratic distinct-origin counting).

Please introduce stable indexes rather than making the resolver hot path scan the full cache, for example:

```zig
entries: std.AutoHashMapUnmanaged(u64, ServerEntry),
handle_index: std.AutoHashMapUnmanaged([stateful_identity_len]u8, u64),
origin_index: std.AutoHashMapUnmanaged(OriginDigest, OriginBucket),
lru: IntrusiveList(u64),
next_entry_id: u64,
```

The handle map may point to a stable entry ID so LRU movement and removal do not invalidate indexes. All index/list/accounting changes must commit atomically after fallible allocation succeeds. Add pressure tests at the configured maximum and consistency checks after swap/removal/eviction/allocation failure.

src/tls/session_cache.zig
```diff
@@ -0,0 +1,1805 @@
...
+        }
+
+        self.mutex.lock();
+        defer self.mutex.unlock();
+
+        for (self.entries.items, 0..) |*e, i| {
+            if (!secrets.constantTimeEqual(&e.handle, identity)) continue;
+
+            if (e.usage == .single_use and e.leased) {
+                self.observer.notify(.lookup_miss);
```
Jaruso: **Blocking deadlock: the observer is invoked while `self.mutex` is held.** This directly contradicts the module/PR contract. A valid observer that calls `cache.count()` or records through any component that re-enters the cache will deadlock. The same problem occurs for the expired, incompatible, storage-failed, hit, and final miss branches below.

Capture the result/event while locked and notify only after unlock:

```zig
var event: CacheEvent = .lookup_miss;
var result: ServerLookupResult = .miss;
{
    self.mutex.lock();
    defer self.mutex.unlock();
    // mutate/cache result only; set `event`
}
self.observer.notify(event);
return result;
```

Because `.hit` owns allocated state, use a local result that is moved out after the critical section. Please add a re-entrant observer test whose callback calls `count()`; it should complete rather than hang.

src/tls/session_cache.zig
```diff
@@ -0,0 +1,1805 @@
...
+            if (e.state.common.isExpired(now_unix_ms)) {
+                var removed = self.entries.swapRemove(i);
+                self.total_bytes -= removed.bytes;
+                removed.state.deinit();
+                secrets.secureZero(&removed.handle);
+                self.observer.notify(.lookup_expired);
+                return .expired;
+            }
+
+            const decision = session.evaluateCompatibility(&e.state.common, candidate, now_unix_ms);
```
Jaruso: **Architecture blocker: compatibility evaluation belongs in the shared #362 backend path, not in the stateful cache.** PR #479's canonical `ServerPskResolver` resolves an identity into owned `ServerRecoverableState`; `tls13_backend.zig` then performs the exact SNI/ALPN/cipher/auth/transport/application compatibility decision and binder verification for both stateless and stateful identities. Accepting `CandidateContext` here duplicates that policy and makes this primitive impossible to adapt cleanly to the existing resolver contract.

Please split storage resolution from protocol selection:

```zig
pub const ResolveLeaseResult = union(enum) {
    hit: struct { state: session.ServerRecoverableState, lease: ServerLease },
    miss,
    expired,
    storage_failed,
};

pub fn resolveLease(
    self: *StatefulServerCache,
    identity: []const u8,
    now_unix_ms: i64,
) ResolveLeaseResult;
```

Then the #362 selection path evaluates `session.evaluateCompatibility`; mismatch releases the lease, binder failure releases it, and binder success commits it before emitting ServerHello. This is necessary for stateful/stateless parity and for avoiding two subtly different compatibility implementations.

src/tls/session_cache.zig
```diff
@@ -0,0 +1,1805 @@
...
+        self.state.deinit();
+        secrets.secureZero(&self.handle);
+    }
+};
+
+/// An owned lease over a resolved single-use or reusable stateful entry.
+/// Reusable leases are a no-op on `commit`/`release`. `generation` guards
+/// against ABA if the same handle bytes were ever reinserted between
+/// resolution and commit/release.
+pub const ServerLeaseToken = struct {
```
Jaruso: **Blocking concurrency bug: `generation` identifies the entry, not this lease acquisition.** Consider A resolves a single-use entry, then releases it; B resolves the same entry; A's stale token can now call `commit` and delete B's active lease, or call `release` and make the entry concurrently resolvable. The entry generation is unchanged across release/re-lease, so the claimed ABA protection does not cover this case. Also, `ServerLookupResult.deinit()` only frees the cloned state and leaves a forgotten single-use lease pinned forever.

Use an owned, exactly-once lease with a unique epoch per acquisition:

```zig
const ServerEntry = struct {
    entry_generation: u64,
    active_lease_epoch: ?u64 = null,
    ...
};

pub const ServerLease = struct {
    cache: *StatefulServerCache,
    handle: [stateful_identity_len]u8,
    entry_generation: u64,
    lease_epoch: u64,
    active: bool = true,

    pub fn commit(self: *ServerLease) void { ... }
    pub fn release(self: *ServerLease) void { ... }
    pub fn deinit(self: *ServerLease) void { self.release(); }
};
```

`commit`/`release` must require both generations and `active_lease_epoch == lease_epoch`, clear/invalidate the epoch exactly once, and wipe the token's handle. Add stale-release-after-re-lease, stale-commit-after-re-lease, double-commit, double-release, and abandoned-result-deinit tests.

src/tls/session_cache.zig
```diff
@@ -0,0 +1,1805 @@
...
+        self.mutex.lock();
+        defer self.mutex.unlock();
+
+        var out: std.ArrayListUnmanaged(PersistedServerEntry) = .empty;
+        errdefer {
+            for (out.items) |*p| p.deinit();
+            out.deinit(allocator);
+        }
+        try out.ensureTotalCapacityPrecise(allocator, self.entries.items.len);
+        for (self.entries.items) |*e| {
```
Jaruso: **Blocking single-use persistence violation: this snapshots entries that are currently leased.** If save clones a leased single-use entry, the handshake can subsequently verify its binder and commit consumption, but the just-written snapshot still contains that entry. A restart then resurrects a ticket that was successfully consumed. Conversely, `loadServerCache` can replace the live cache while a lease is outstanding and expose the same persisted entry to another handshake.

The persistence contract needs an explicit quiescence/generation rule. The safest initial behavior is to reject save/load while any single-use lease is active:

```zig
pub fn cloneLiveForPersistence(...) error{ OutOfMemory, CacheBusy }!... {
    self.mutex.lock();
    defer self.mutex.unlock();
    for (self.entries.items) |e|
        if (e.usage == .single_use and e.leased) return error.CacheBusy;
    ...
}
```

`loadServerCache` should likewise return `.busy` before beginning the swap if there are outstanding leases (and should re-check under the final lock). A more advanced snapshot transaction is possible, but simply skipping leased entries can lose valid tickets on release, so it needs an explicit policy. Add save-during-lease→commit→restart and load-during-lease concurrency tests.

src/tls/session_cache.zig
```diff
@@ -0,0 +1,1805 @@
...
+const ClientEntry = struct {
+    ticket: session.ClientTicketState = .{},
+    origin: OriginDigest = [_]u8{0} ** origin_digest_len,
+    usage: UsagePolicy = .reusable,
+    sequence: u64 = 0,
+    entry_id: u64 = 0,
+    bytes: usize = 0,
+};
+
+pub const PersistedClientEntry = struct {
```
Jaruso, line 368: **Blocking persistence determinism gap: persisted entries omit both semantic insertion order and LRU recency.** `cloneLiveForPersistence` iterates the physical `ArrayList` order, but `swapRemove` intentionally scrambles that order. Restore then assigns new sequences in that arbitrary array order. After restart, offer ordering and the next eviction candidate can differ from the pre-save cache, violating deterministic selection/eviction and successful persistence round-trip semantics. The server format has the same issue.

Either persist the ordering metadata explicitly:

```zig
pub const PersistedClientEntry = struct {
    ticket: session.ClientTicketState,
    usage: UsagePolicy,
    insertion_sequence: u64,
    lru_sequence: u64,
};
```

and normalize sequences on load, or sort cloned records by the semantic sequence before encoding and restore in a documented direction. If both insertion and LRU order matter, both must survive. Add a test that performs lookup recency changes and `swapRemove` eviction before save, then verifies identical offer order and next-eviction behavior after load.

Jaruso (commented):
## Review decision: **not ready to merge — changes required**

I reviewed all three changed files at `d38e9e99cdd6deccdb0f183052b3ed14f87676d9` against the full #364 issue body and its canonical implementation amendment, including the now-merged PR #479 resolver/PSK contract. The CI run is green, but the new tests do not cover several important capacity, concurrency, ownership, and persistence failure modes.

The connected GitHub account owns this PR, so GitHub does not permit a formal `REQUEST_CHANGES` review. Treat this `COMMENT` review as blocking.

The cache-owned scope is not yet safe to merge:

1. Duplicate client-ticket replacement can exceed per-entry/total byte limits after destroying the old entry.
2. Expired entries outside the target origin can permanently consume origin cardinality; there is no periodic/global cleanup API.
3. Eviction is FIFO/insertion-order rather than LRU, and client lookup allocation failure is silently returned as a partial hit.
4. The stateful cache lacks the required handle/origin indexes, making attacker-controlled resolver lookup linear in the entire cache.
5. Stateful lookup calls the observer while holding the cache mutex, allowing immediate re-entrancy deadlocks.
6. Stateful compatibility evaluation is implemented in the cache instead of the shared #362 path, preventing clean stateful/stateless resolver parity.
7. The lease token identifies only the entry, not an individual lease acquisition; a stale token can commit or release a later handshake's lease. The result also cannot automatically release an abandoned lease.
8. Persistence can snapshot a leased single-use entry and later resurrect it after successful consumption; live load also does not guard outstanding leases.
9. Persistence does not preserve deterministic insertion/LRU order.
10. Snapshot decoding trusts an unbounded record count before cache limits, accepts trailing bytes, and does not validate persisted `TDSH` handle shape.
11. Restore ignores allocation/capacity/handle failures, so load can return `.loaded` and replace a valid live cache with a partial cache.

The inline comments include concrete implementation patterns and regression tests expected for each issue. The deliberately deferred composite resolver, root export, prepared issuance, and runtime wiring remain incomplete #364 work; that is acceptable for a foundational partial PR only if #364 stays open, but the cache/persistence primitives themselves need the fixes below first.

src/tls/session_cache_persistence.zig
```diff
@@ -0,0 +1,787 @@
...
+) SnapshotDecodeError!std.ArrayListUnmanaged(session_cache.PersistedClientEntry) {
+    const header = try readHeader(bytes);
+    if (header.kind != .client) return error.UnsupportedKind;
+
+    var out: std.ArrayListUnmanaged(session_cache.PersistedClientEntry) = .empty;
+    errdefer {
+        for (out.items) |*p| p.deinit();
+        out.deinit(allocator);
+    }
+    out.ensureTotalCapacityPrecise(allocator, header.count) catch return error.OutOfMemory;
```
Jaruso, line 226: **Blocking bounded-parser issue: `header.count` is trusted for allocation before any cache cardinality or snapshot-size limit is applied.** A corrupted/authenticated snapshot can set `count = 0xffffffff`, causing an enormous allocation attempt before the temporary cache enforces `max_entries`. Both decoders also return successfully with trailing bytes after the declared record count, and the server decoder accepts handles with bad magic/version/reserved bytes.

Pass explicit persistence/cache bounds into decoding and reject before allocation:

```zig
if (header.count > cache_limits.max_entries or
    header.count > session_cache.hard_max_entries)
    return error.TooManyRecords;
if (bytes.len > persistence_limits.max_plaintext_bytes)
    return error.SnapshotTooLarge;
...
if (rest.len != 0) return error.TrailingData;
```

For server records, validate the complete `TDSH` shape (`magic`, version, and zero reserved field) before appending, and wipe the local handle on every error path. Use checked arithmetic in encoders as well. Add count-bomb, zero-count-with-trailing-data, valid-record-plus-trailing-data, nonzero-reserved, and oversized-plaintext tests.

src/tls/session_cache_persistence.zig
```diff
@@ -0,0 +1,787 @@
...
+
+    var entries = decodeClientSnapshotAlloc(cache.allocator, limits, plaintext) catch |err| return mapDecodeError(err);
+    defer {
+        for (entries.items) |*p| p.deinit();
+        entries.deinit(cache.allocator);
+    }
+
+    var temp = session_cache.ClientSessionCache.init(cache.allocator, cache.limits) catch return .corrupted;
+    defer temp.deinit();
+    temp.restoreClones(entries.items, now_unix_ms);
```
Jaruso: **Blocking atomicity bug: restore is infallible at this call site only because it discards failures.** `restoreClones` ignores every `StoreResult`; `restoreEntries` similarly ignores `insertLocked` results, bypasses `max_entry_bytes`, and does not validate persisted handle shape. An injected allocation failure or over-limit record therefore creates a partial `temp`, after which this function returns `.loaded` and destroys the valid live cache. That contradicts the stated "swap only after full success" contract.

Make restore fallible and require every non-expired record to be accounted for before swapping:

```zig
const RestoreError = error{
    OutOfMemory,
    CapacityExceeded,
    InvalidHandle,
    DuplicateHandle,
};

try temp.restoreClones(entries.items, now_unix_ms);
// only then acquire the final swap lock
```

For server restore, preflight `bytes <= max_entry_bytes`, validate the handle, reject duplicates, and map `.storage_failed`/capacity outcomes to a non-`.loaded` result. If the intended policy is deterministic truncation rather than failure, it must be explicit, preserve ordering, and never misreport allocation failure as success. Add failing-allocator sweeps for every restore allocation and verify the original live entries, counters, indexes, and byte totals remain unchanged.

Jaruso (commented):
## Re-review of `e9977e7267f1ded9fdf084730a94c9615016d72b`

I rechecked the complete cache and persistence implementations against issue #364, not only the latest commit. The new commit correctly fixes the prior origin-bucket pointer invalidation, commit-time server LRU refresh, cache-derived snapshot-size ceiling, duplicate logical-key handling, and the all-leased capacity preflight.

**Verdict: changes are still required; this is not ready to merge.** Five blocking issues remain.

### 1. Sequence overflow now deliberately violates the explicit #364 contract

The canonical issue text requires:

> Sequence overflow must renumber live entries deterministically from oldest to newest under the mutex.

and includes deterministic sequence-overflow behavior in the additional acceptance checks. The new implementation instead uses wrapping counters everywhere (`+%=`) and documents wraparound as out of scope. That is not an equivalent implementation: once `next_lru_sequence` wraps, ordinary unsigned `<` comparisons immediately classify newly touched entries as the oldest entries and evict them first. `insertion_sequence` wrap likewise reverses offer preference.

Please implement renumbering without physically reordering live storage while callers hold indexes/pointers. One workable shape is:

```zig
const SequenceItem = struct {
    entry_id: u64,
    old_sequence: u64,
};

fn renumberServerLruLocked(self: *StatefulServerCache) error{OutOfMemory}!void {
    var order = try self.allocator.alloc(SequenceItem, self.entries.count());
    defer self.allocator.free(order);

    var i: usize = 0;
    var it = self.entries.iterator();
    while (it.next()) |kv| : (i += 1) {
        order[i] = .{
            .entry_id = kv.key_ptr.*,
            .old_sequence = kv.value_ptr.lru_sequence,
        };
    }

    std.mem.sort(SequenceItem, order, {}, struct {
        fn lessThan(_: void, a: SequenceItem, b: SequenceItem) bool {
            if (a.old_sequence != b.old_sequence)
                return a.old_sequence < b.old_sequence;
            return a.entry_id < b.entry_id;
        }
    }.lessThan);

    for (order, 0..) |item, sequence| {
        self.entries.getPtr(item.entry_id).?.lru_sequence = @intCast(sequence);
    }
    self.next_lru_sequence = order.len;
}
```

Callers must perform this fallible preparation before mutating cache state. For client lookup, retain stable `entry_id`s rather than physical array indexes across a possible renumber and re-find the entry before updating it. Add actual wrap tests proving offer order and the next eviction victim remain identical across renumbering.

### 2. Stateful insertion is still not failure-atomic

`insertMove` advances `next_lru_sequence` before knowing whether insertion succeeds. `insertLocked` then purges expired entries and performs per-origin/global evictions before these later fallible operations:

```zig
self.origin_index.ensureUnusedCapacity(self.allocator, 1)
bucket_ptr.?.ensureUnusedCapacity(self.allocator, 1)
```

If either allocation fails, the caller retains `state`, but existing entries may already have been evicted and the LRU counter has advanced. This directly conflicts with the required mutation order (“complete all fallible allocation work before mutation”) and the acceptance check that allocation failures leave indexes, lists, counters, and byte accounting unchanged.

The same ordering problem exists in the client store path: expired entries are purged before `entries.ensureUnusedCapacity` can fail.

Please reserve every structure needed by the transaction before purge/eviction and advance sequence counters only after commit. For the stateful origin bucket, either preserve an empty bucket for the duration of the transaction or prepare a detached bucket with capacity before mutation so a capacity-1 eviction cannot destroy the only preallocated bucket.

Expected invariant:

```zig
// No mutation above this line.
try reserveAllInsertionCapacity(...);
const new_lru = self.next_lru_sequence;

purgeExpired...;
evict...;
commitEntry...;
self.next_lru_sequence +%= 1;
```

Add `FailingAllocator` sweeps against a **populated cache that must evict** and fail each reservation point. After every failure, assert the original handles, origin buckets, entry count, total bytes, LRU values, and counters are byte-for-byte/logically unchanged.

### 3. The client persistence guard can break single-use semantics

The new guard prevents `consumeSingleUse` during a save, but it does not prevent the ticket from being offered before or during that save. This race remains:

1. `lookupOffers` returns a clone of a single-use ticket.
2. A save snapshots that ticket and sets `persistence_in_progress`.
3. The peer selects the ticket and binder verification succeeds.
4. `consumeSingleUse` returns `false` because the save is active.
5. The handshake has already selected PSK, while both the live cache and durable snapshot still contain the supposedly single-use ticket.

`false` also conflates “not found” with “persistence busy”, so the future caller cannot make a safe decision. Additionally, two concurrent saves can both set the boolean; the first completion clears it while the second save is still writing, reopening the same race.

Issue #364 already defines the safe choices. Since the selected-offer outcome is deferred to #365, either:

- make the client cache explicitly reusable-only for this PR and reject `.single_use`, **or**
- implement an owned client offer lease now: lookup pins stable cache entry IDs, persistence returns `CacheBusy` while any single-use offer lease exists, and a selected-index/no-selection/abort outcome commits or releases the lease.

The persistence operation itself should also be a uniquely owned token or reject an already-active save, rather than using an uncounted boolean.

### 4. Reusable server leases have an ABA bug across reload

Reusable leases now mutate LRU on `commit`, but `loadServerCache` only blocks outstanding single-use leases. A reusable lease can therefore survive a reload:

1. Old cache entry ID `1` resolves and returns a reusable lease.
2. Reload builds a temporary cache whose IDs start from `1` and swaps it in.
3. The old lease commits.
4. `commitLease(1, ..., false)` updates the LRU of an unrelated restored entry that happens to have ID `1`.

This does not require a `u64` wrap; reload itself reuses IDs. Add a cache generation to every lease and increment it on cache replacement:

```zig
pub const ServerLease = struct {
    cache: *StatefulServerCache,
    cache_generation: u64,
    entry_id: u64,
    // ...
};

fn commitLease(self: *StatefulServerCache, generation: u64, entry_id: u64, ...) void {
    if (generation != self.cache_generation) return;
    // ...
}
```

Do not copy the temporary cache’s generation into the live object during swap. Alternatively, reject reload while **any** lease is outstanding, including reusable leases. Add a regression test where an old reusable lease commits after reload and verify that no restored entry’s LRU changes.

### 5. Stateful handle zeroization is not guaranteed by the current hash-map representation

The 40-byte `TDSH` handle is explicitly documented as a bearer secret, but it is stored inline in two general-purpose hash-map allocations:

- `entries` values (`ServerEntry.handle`), and
- `handle_index` keys (`[stateful_identity_len]u8`).

`removeEntryLocked` wipes the fetched `ServerEntry` copy, but `AutoHashMapUnmanaged.remove/fetchRemove/deinit` are not secret-wiping storage APIs. Removed slots, rehash source allocations, and the `handle_index` backing allocation can retain raw handle bytes. `deinit` never walks and wipes `handle_index` keys.

The existing zeroization test does not establish this invariant: it creates the session state with `testing.allocator` rather than the inspected fixed-buffer allocator and does not assert that `insertMove` actually returned `.stored` before scanning the buffer.

Please avoid placing raw bearer handles or inline secret-bearing entries in reallocating general-purpose map storage. A safer design is:

```zig
// Non-secret, domain-separated digest used only as the hash-map key.
handle_index: std.AutoHashMapUnmanaged(HandleDigest, u64),

// Map values contain only stable pointers/IDs; the heap-owned ServerEntry is
// explicitly wiped before being freed.
entries: std.AutoHashMapUnmanaged(u64, *ServerEntry),
```

On lookup, fetch by digest and constant-time compare the owned entry’s real handle before accepting. Also wipe the local handle in `decodeServerSnapshotAlloc` on every return path:

```zig
var handle: [stateful_identity_len]u8 = undefined;
defer secrets.secureZero(&handle);
```

Add removal, eviction, rehash, failed-insert, failed-decode, reload-swap, and deinit tests using one inspected allocator; assert each insertion succeeds before testing zeroization, then scan the allocator’s full backing capacity for the handle and other secret bytes.

---

The current CI run for this head was still in progress when this review was submitted. Even if it passes, the cases above are not represented by the current tests and remain correctness/security blockers.

Jaruso (commented):
## Re-review decision: **not ready to merge — prior blockers remain unresolved**

I reviewed head `a77c75def67d8d43cf80369bd0bb93d67073f718` against the previous reviewed head `e9977e7267f1ded9fdf084730a94c9615016d72b`, the full #364 contract, and the existing cache/persistence implementation.

The new commit adds useful coverage for global server expiry cleanup and persistence ordering, and it wipes the stack-local decoded `TDSH` handle. However, the compare contains only additions in the two files; it does not modify any of the five implementation paths from the prior review. Those blockers therefore remain. I also found that the new persistence-order regression test mutates the LRU state it is trying to verify.

### 1. Sequence overflow still directly contradicts #364 and breaks deterministic ordering

The client counters, server LRU counter, entry IDs, and lease epochs still use wrapping `+%=` increments, and the module documentation explicitly accepts wraparound. #364 explicitly requires live entries to be renumbered deterministically from oldest to newest under the mutex.

This is not merely documentary: after wrap, a newly inserted or touched entry receives a small sequence and is treated as the oldest entry by the eviction helpers; client offer ordering can likewise place newly inserted tickets behind ancient entries. Lease-epoch wrap also reintroduces the ABA condition the epoch was intended to prevent.

Please implement deterministic renumbering without physically reordering backing storage. For example, collect stable indices/entry IDs, sort that scratch list by the existing sequence plus a deterministic tie-breaker, then rewrite sequence fields in place. Complete any scratch allocation before mutating state. Add near-`maxInt(u64)` tests for client insertion order, client LRU, server LRU, entry IDs, and lease epochs.

### 2. Stateful insertion remains non-atomic on allocation failure

`insertMove` still advances `next_lru_sequence` before `insertLocked` knows the insertion will succeed. More importantly, `insertLocked` can evict existing entries and only afterward call:

```zig
self.origin_index.ensureUnusedCapacity(...)
bucket_ptr.?.ensureUnusedCapacity(...)
```

Either allocation can fail after unrelated live entries have already been permanently removed. The caller retains its input state, but the cache, indexes, byte accounting, eviction order, observer count, and sequence counters have changed. This violates #364's required mutation order and its allocation-failure invariant.

All fallible allocation must complete before expiry removal/eviction/sequence advancement, or the mutation needs a real rollback transaction. A practical design is to prepare a new origin bucket off-cache when needed, reserve every map/list capacity, precompute the exact eviction set without removing it, and only then apply removals and commit the new entry. Assign the new sequence at final commit, not before calling `insertLocked`.

The current empty-cache `FailingAllocator` sweep does not exercise this. Add a populated cache with multiple unleased victims and sweep every failure point, asserting exact preservation of handles, count, bytes, origin buckets, LRU order, sequence counters, and caller ownership.

### 3. The client persistence guard can still violate single-use semantics

`lookupOffers` returns owned ticket clones without reserving or leasing the originating single-use entry. A save may begin after the ticket was offered/selected but before `consumeSingleUse` runs. During the save, `consumeSingleUse` now returns `false`; the snapshot contains the ticket and the live cache leaves it reusable. The caller cannot distinguish this from “entry not found” or safely retry through the current API.

The persistence guard therefore closes one resurrection race by introducing a single-use-consumption failure after successful selection. #364's canonical plan requires a client offer lease/outcome API even when runtime selected-index wiring is deferred to #365.

Please add a client offer lease/reservation that pins the selected candidates until commit/release, make save report busy while such a single-use offer is outstanding, or explicitly make this PR's client cache reusable-only and reject/defer single-use persistence until #365. Add a deterministic interleaving test: lookup single-use offer → begin save → selected-offer commit → finish save/reload; the consumed identity must not survive and commit must not silently fail.

### 4. A reusable server lease can mutate an unrelated entry after reload

`loadServerCache` blocks only outstanding single-use leases. Reusable `ServerLease` values remain active across the atomic cache swap and carry only `entry_id`. The temporary restored cache allocates entry IDs again from `1`, so an old reusable lease can later call `commit()` and refresh the LRU sequence of a completely unrelated restored entry that reused the same ID.

Add a cache-generation value to every lease and increment it on reload/reset, validating both generation and entry ID in `commitLease`/`releaseLease`; alternatively, track all outstanding leases and refuse reload while any lease is active. Add a regression test that holds a reusable lease, reloads a snapshot containing a different entry with the same internal ID, then commits the stale lease and verifies the restored entry is unchanged.

### 5. Raw stateful handles still remain in freed/reused hash-map capacity

The new `defer secrets.secureZero(&handle)` correctly wipes the temporary decoder variable, but it does not address the reported storage issue. The raw bearer handle exists in both `entries` map values and `handle_index` keys. `fetchRemove`/`remove` update map metadata and return/copy a value, but do not guarantee that the old key/value bytes in backing capacity are overwritten. `AutoHashMapUnmanaged.deinit` frees those allocations without wiping their full capacity.

Consequently, wiping the copied `entry.handle` and the decoder local still leaves bearer handles in allocator memory after removal, eviction, rehash, and deinit, contrary to #364's requirement to clear secret-bearing storage and full allocated capacity.

Use secret-aware storage whose occupied and spare backing capacity can be explicitly scrubbed before release/reallocation. One option is to index by a non-secret digest of the handle and keep raw handles in a separately managed secret entry slab whose slots and full allocation are wiped. Whichever design is chosen, test allocator backing after individual removal, eviction, map growth/rehash, failed insertion, load swap, and cache deinit—not only the local handle copy.

### 6. The new persistence-order test changes the LRU order before testing eviction

The new test says the post-save entries are `t1(lru=3)` and `t3(lru=4)` and that the next insertion should evict `t1`. But immediately before inserting `t4`, it calls `lookupOffers` on both caches to compare offer order. Those lookups refresh LRU in newest-insertion order: `t3` is touched first and `t1` second. That makes `t3` the smaller/older newly assigned sequence and changes the expected victim.

The final assertions only prove that both caches produced the same outcome; they never assert which ticket was evicted. Both implementations can therefore be identically wrong and the test still passes.

Please inspect offer-order metadata without a recency-changing lookup, or apply eviction pressure before calling `lookupOffers`, then assert the exact expected survivor/victim in both caches. Also correct the test comments to match the actual state transitions.

### CI

At review time, CI run `29866103658` is still in progress. ARM unit tests, macOS unit tests, format, lint, and dependency review have passed so far. Green CI will validate the added tests compile and pass, but it cannot resolve the unchanged semantic and failure-atomicity issues above.

Because this PR is owned by the connected GitHub account, GitHub does not permit a formal `REQUEST_CHANGES` submission. Treat this `COMMENT` review as blocking.

Jaruso (commented):
Re-review of head `153c3890873cd762ac4ba5d58811db67fabccb68`

**Verdict: changes still required; not ready to merge.**

The new design does fix several previous blockers: stateful entries now live in individually wiped allocations, `handle_index` no longer stores raw bearer handles, reload generation-gates stale leases, and insertion eviction victims are planned before the main capacity reservations. The corrected persistence-order test is also materially better.

I still found the following blocking issues.

### 1. Client multi-offer LRU reservation corrupts ordering when overflow occurs partway through the batch

`lookupOffers` calls `nextLruSequence()` once per touched offer and stores each returned value in `new_seqs`. That helper may renumber all live entries whenever `next_lru_sequence == maxInt(u64)`.

This is incorrect when the counter is *near* the boundary rather than exactly at it. For example, with two returned offers and:

```zig
cache.next_lru_sequence = std.math.maxInt(u64) - 1;
```

reservation proceeds as follows:

1. first call returns `maxInt - 1`, then advances the counter to `maxInt`;
2. second call renumbers every live entry to a compact `0..n-1` range, then returns `n`;
3. the code applies `[maxInt - 1, n]` to the touched entries.

The first touched entry is now effectively immortal: all future insert/touch sequence values restart near `n`, so they compare older than `maxInt - 1` for practically the lifetime of the process. Relative recency among the two entries is also reversed.

Reserve the entire batch atomically. Renumber **before returning any batch value** when the remaining numeric range is smaller than `touched_len`, then assign one contiguous range:

```zig
fn reserveLruBatchLocked(self: *ClientSessionCache, count: usize) !u64 {
    const remaining = std.math.maxInt(u64) - self.next_lru_sequence;
    if (count > remaining) try self.renumberLruSequencesLocked();

    const first = self.next_lru_sequence;
    self.next_lru_sequence += @intCast(count);
    return first;
}
```

The exact boundary arithmetic can vary, but no value captured before renumbering may be applied afterward.

Required regression: set `next_lru_sequence = maxInt(u64) - 1`, return at least two offers, then apply capacity pressure and assert the exact expected victim/survivor—not merely that no overflow trap occurred.

### 2. Insert/store failure atomicity is still violated by sequence mutation and expiry purge before later fallible work

The issue requires fallible allocation/handle-generation work to complete before mutation, and its allocation-failure acceptance check explicitly includes counters/indexes/lists.

Client `storeLocked` mutates ordering state before validation and before `entries.ensureUnusedCapacity`:

```zig
const new_insertion_seq = self.nextInsertionSequence() catch return .storage_failed;
const new_lru_seq = self.nextLruSequence() catch return .storage_failed;
```

Examples:

- if insertion renumber succeeds and LRU renumber allocation fails, insertion sequences and `next_insertion_sequence` have already changed;
- if both reservations succeed and `ensureUnusedCapacity` later fails, both counters have advanced;
- a rejected oversized/capacity-blocked store also consumes ordering values and may have renumbered live entries.

The stateful path has the same class of issue: `insertMove` reserves/mutates `next_lru_sequence` before `insertLocked`, and `insertLocked` calls `purgeExpiredAllLocked` before `planInsertionLocked` and every subsequent capacity reservation. A later OOM can therefore return `.storage_failed` after changing the counter and deleting expired entries.

The new FailingAllocator test does not prove atomicity. It shallow-copies the populated cache:

```zig
var failing_cache = cache;
```

so both structs alias the same hashmap storage and `*ServerEntry` values; successful reallocation updates only the copy and can leave the original with stale backing pointers. The assertions then inspect `cache`, not the exact object passed to `insertMove`, and do not assert sequence counters or both indexes.

Please stage the complete mutation first: compute victim/expiry plans and any renumber mappings in scratch storage, reserve all capacities, then commit sequence changes + expiry removals + planned evictions + insertion as one non-fallible phase. Alternatively provide a complete rollback, including exact counters and sequence metadata.

Each failure index should use a freshly constructed populated cache and assert, on the actual cache under test:

- exact entry handles/states;
- `entries`, `handle_index`, and `origin_index` membership;
- byte/count totals;
- ordering counters and every live LRU sequence;
- caller ownership unchanged.

Include a case with an expired entry present so a late OOM proves expiry cleanup is not partially committed.

### 3. The server persistence guard is not safe for overlapping save/load operations

`cloneLiveForPersistence` sets a single boolean but never rejects an already-active snapshot:

```zig
self.persistence_in_progress = true;
```

Two saves can therefore overlap:

1. save A clones ticket T and sets the flag;
2. save B also clones T and sets the same flag;
3. A finishes and `endPersistenceSnapshot()` clears the flag;
4. T is resolved and committed while B is still writing;
5. B persists its earlier clone containing T, resurrecting a consumed single-use ticket after restart.

There is also a save/load ordering race. A load can read an older backend snapshot, a concurrent save can then write a newer no-T snapshot and clear the flag, and the load can finally swap the older T-containing snapshot into memory. The final outstanding-lease check does not prevent that stale resurrection.

Persistence operations need serialization for their **entire** backend/encode/decode/swap duration without holding the cache data mutex during I/O. A dedicated persistence mutex is the simplest solution. A generation/token guard can also work, but `begin` must reject an active operation and `end` must only clear the matching token. Apply it to both save and load.

Required tests should use a blocking backend/barrier to deterministically prove:

- a second save returns `.cache_busy` while the first is in flight;
- load cannot overtake or be overtaken by a save;
- ending one operation cannot clear another operation's guard;
- a single-use ticket consumed between the attempted operations is never restored.

### 4. LRU refresh allocation failure is silently ignored, violating the typed result and true-LRU guarantees

Client lookup returns a successful `.hit` even when renumber allocation fails while reserving recency values. The function documentation says every returned entry is refreshed, and `ClientLookupResult` already has `.storage_failed`; silently returning a hit leaves a used entry eligible for eviction and hides an actual cache allocation failure.

The reusable server path similarly does:

```zig
e.lru_sequence = self.reserveFreshLruSequenceLocked() catch return;
```

so a successful binder-verified reuse can silently fail to become MRU. That means the stateful cache is not true LRU under the allocation-failure condition the issue explicitly requires testing.

Make overflow renumber allocation-free, or reserve the needed recency value at a phase where failure can be surfaced. For client lookup, deinitialize the cloned offers and return `.storage_failed` if the recency transaction cannot be committed. For the server lease, one option is to reserve/carry a recency token in the lease before returning the hit; another is an allocation-free deterministic renumber under the mutex.

Add targeted failing-allocation tests at the overflow boundary and assert both the result and exact eviction order afterward.

### 5. The client path is not actually an explicit reusable-only contract

The code removed `consumeSingleUse` and the persistence guard, but `storeClone` and restore still accept, preserve, and repeatedly return `.single_use` entries as ordinary reusable offers. That silently downgrades a caller-declared single-use ticket rather than enforcing a reusable-only API.

The canonical issue text says that, when selected-offer wiring is deferred to #365, #364 should ship reusable client tickets **plus the offer lease API**. This head has neither consumption nor an offer lease. The PR description is also now stale: it still says `consumeSingleUse` ships as the cache-side primitive.

Please choose one explicit contract:

- implement the client offer lease/selected-index commit API now, leaving only backend wiring to #365; or
- reject `.single_use` in the client cache and formally update #364/this PR to defer the client single-use API and acceptance criterion entirely.

Persisting a `.single_use` marker while intentionally behaving reusable is not a safe middle state.

### 6. The public restore helpers still claim atomicity they do not provide

Both `ClientSessionCache.restoreClones` and `StatefulServerCache.restoreEntries` insert records into `self` one at a time. If record N succeeds and record N+1 hits OOM, the method returns an error with the earlier records still inserted; on a non-empty target, prior entries may also have been evicted or expired entries purged.

The client test named `restoreClones aborts atomically` only forces the very first allocation to fail in an empty target and then checks an unrelated live cache. It does not test a mid-stream failure or rollback.

Either:

- make these helpers private and document/assert that they operate only on a fresh temporary cache that the load path discards on error; or
- make the public methods truly atomic by restoring into a temporary cache and adopting only after complete success.

Add a failure after at least one successful restored record against a prepopulated target and assert exact pre-call state.

### 7. CI is currently red

Run 867 completed with failure. `Test appliance profile (macos-14)` failed in **Run native TLS listener integration tests**. The cache-local/unit jobs shown in the run passed, so this may be unrelated or flaky, but the PR should not merge until the required check is green or the failure is diagnosed and explicitly dispositioned.

Once these are addressed, the remaining design is much closer to the #364 contract.

Jaruso (commented):
## Re-review — head `7b26efac9c7d9020061415d187d7ede0d3bf5470`

**Verdict: changes are still required before merge.** This revision is materially better. The prior batch-overflow bug, fallible renumbering, stale reusable lease after reload, raw-handle hashmap storage, restore-helper partial adoption, overlapping persistence operations, and the macOS CI failure are addressed. CI run **869** is now fully green.

I still found four concrete correctness/security defects and one scope mismatch with #364’s canonical plan.

### 1. Client cache relocation leaves inline PSK/nonce copies in unwiped memory

`ClientEntry` stores `session.ClientTicketState` inline inside `ArrayListUnmanaged(ClientEntry)`. That state contains inline secret-bearing values, including `ResumableSessionCommon.resumption_psk` and `ticket_nonce`.

Several normal operations bitwise-copy the entire `ClientEntry`:

- `entries.ensureUnusedCapacity` may move every entry to a new allocation and free the old allocation without wiping it;
- `std.mem.sort(ClientEntry, entries.items, ...)` swaps whole secret-bearing structs during sequence renumbering;
- `swapRemove` leaves a duplicate of the moved last entry in the now-unused tail slot beyond `items.len`.

Calling `removed.ticket.deinit()` wipes the returned copy, not every stale copy left in the ArrayList’s old allocation, unused capacity, or sort scratch. `ClientSessionCache.deinit()` only deinitializes active items and then frees the backing allocation without wiping the full capacity. This violates #364’s explicit requirement to clear secret-bearing storage and test zeroization of the **full allocated capacity**.

The server-side pointer representation already demonstrates the safer design. I recommend making the client list pointer-based as well:

```zig
entries: std.ArrayListUnmanaged(*ClientEntry) = .empty,

fn destroyClientEntry(allocator: std.mem.Allocator, entry: *ClientEntry) void {
    entry.ticket.deinit();
    secrets.secureZero(std.mem.asBytes(entry));
    allocator.destroy(entry);
}
```

Then sort/reorder only pointers or non-secret metadata, and destroy each entry through one wiping function.

Required regression test: use a `FixedBufferAllocator` with distinctive PSK and nonce byte patterns; force ArrayList growth, a non-last `swapRemove`, and overflow renumbering; then evict/deinit and scan the allocator’s **entire backing buffer** for both patterns. Testing only the heap-backed ticket string does not cover these inline secrets.

### 2. Store/insert failure atomicity is still incomplete

Both caches still mutate live state before later fallible work finishes:

- Client `storeLocked` advances/renumbers `next_insertion_sequence` and `next_lru_sequence`, then purges expired entries, and only afterward calls `ensureUnusedCapacity`.
- Server `insertMove` reserves a fresh LRU sequence before `insertLocked`; `insertLocked` purges expired entries before `planInsertionLocked` and the subsequent map, bucket, and entry allocations.

Therefore a late `storage_failed` result can leave counters advanced/renumbered and expired entries removed even though insertion failed. #364’s canonical mutation order says all fallible clone/allocation/handle-generation work completes before mutation, and its acceptance checks require allocation-failure sweeps to leave counters, lists, indexes, ownership, and accounting unchanged.

The current FailingAllocator tests do not prove this:

- the client test commonly fails during the initial ticket clone and may never force the later `ensureUnusedCapacity` allocation;
- the server test does not snapshot/assert `next_lru_sequence`, origin/index contents, or an expired entry that would be purged before a later allocation fails.

Plan expiry removals and eviction victims without applying them, reserve every needed allocation, and only then perform purge/evictions, reserve the committed sequence, move ownership, and update counters.

Required regression: fill the relevant container to its actual allocated capacity, include an expired live entry, fail a specifically late allocation, and assert exact equality of entry identities/order, origin and handle indexes, byte counts, all sequence counters, and the still-present expired entry.

### 3. A failed client lookup still changes LRU state

`lookupOffers` reserves the entire LRU batch before cloning any offers, then updates each entry immediately after its individual clone succeeds. If cloning a later offer fails, the method returns `.storage_failed` and destroys the partial offer set, but:

- `next_lru_sequence` has already advanced or renumbered;
- every earlier successfully cloned entry has already received a new LRU value.

Thus a lookup that returned no usable offers can change the next eviction victim.

Clone the complete selected set first while retaining stable entry IDs. Only after all clones succeed should the code reserve the infallible sequence batch and apply the LRU updates.

Required regression: force allocation failure on the second or later selected offer and assert every entry’s LRU sequence, the counter, physical/logical order, and the next eviction result are unchanged.

### 4. A pre-existing single-use lease can commit during load and then be resurrected

The token guard prevents new single-use leases and serializes persistence operations, but `beginPersistenceOperation` does not reject a single-use lease that was already acquired before the load began.

This interleaving is still possible:

1. Acquire a single-use lease.
2. Start `loadServerCache`; it acquires the persistence token and begins backend/decode work.
3. Commit the pre-existing lease during that work, consuming the live entry.
4. At the final swap, `hasOutstandingLeaseLocked()` is now false because the lease was committed.
5. Adopt the loaded snapshot, which may contain and therefore resurrect that consumed entry.

The current load test keeps the lease outstanding for the whole synchronous call, so it exercises only the final check, not commit-during-load.

For load, acquire the persistence token only if there are no outstanding single-use leases in the same critical section. Once the token is held, new leases are already blocked, making the invariant stable. An alternative is to capture a single-use consumption generation and abort if it changes before swap.

Required regression: use a backend/protector hook or controlled thread to commit a lease after load starts but before adoption; the load must abort without replacing the live cache or resurrecting the handle.

### 5. Client single-use scope still does not match #364’s canonical fallback

The new typed rejection is much safer than silently treating `.single_use` as reusable. However, #364’s canonical plan says that when runtime selected-index wiring is deferred, **#364 should still ship reusable tickets plus the client offer-lease API**, with only the runtime commit wiring deferred. This PR now defers the API itself to #365, while #365 is written as an integration story that consumes #364’s cache/selection seam.

Either implement the cache-side client offer lease in this PR, or formally amend/split #364 and #365 so the missing API has an explicit owner and #364 remains open. The current implementation should not be described as satisfying the canonical #364 fallback as written.

There is also a restore edge case within the reusable-only policy: `markDuplicateClientRecords` runs before unsupported `.single_use` records are skipped. A newer single-use duplicate can therefore eliminate an older valid reusable record, after which the winning single-use record is skipped and neither is restored. Filter/drop unsupported records before duplicate selection, or make duplicate selection consider only supported records.

### CI

GitHub Actions run **869** completed successfully, including the previously failing macOS appliance native TLS listener test. The remaining blockers are not represented by the current test suite.

Jaruso (commented):
## Re-review of `4deb9cff446a256aced7b6fc22394a06d34f3c54`

**Verdict: not ready to merge yet. Two blocking correctness issues remain.**

The fifth-round changes do correctly close most of the previous review:

- client secret-bearing entries are now individually allocated and wiped rather than copied through `ArrayList(ClientEntry)` backing storage;
- client/server insertion now generally follows a plan/reserve/commit structure;
- client offer clones are completed before LRU mutation;
- a persistence operation cannot begin while a single-use server lease is already outstanding;
- unsupported client single-use records no longer eliminate valid reusable duplicates during restore;
- I am treating the documented client offer-lease deferral as an explicit maintainer scope decision rather than a code blocker for this PR.

However, the atomicity rewrite reintroduced an earlier expiry/cardinality bug, and the batch LRU overflow guard still has an off-by-one boundary error.

### 1. Blocking: expired origins once again permanently consume `max_origins`

Both client `storeLocked` and server `insertLocked` reject a new origin before their insertion plans are built:

```zig
const origin_count = self.countOrigin(origin);
if (origin_count == 0 and self.countDistinctOrigins() >= self.limits.max_origins) {
    return .rejected_capacity;
}
```

and:

```zig
const is_new_origin = !self.origin_index.contains(origin);
if (is_new_origin and self.origin_index.count() >= self.limits.max_origins) {
    return .rejected_capacity;
}
```

The new implementation deliberately removed eager global expiry purging to preserve failure atomicity. That part is reasonable, but the planning phase does not account for expired entries before the cardinality check.

Reproduction:

1. Configure `max_origins = 1`.
2. Store one entry for `a.test`.
3. Advance `now_unix_ms` beyond that entry's lifetime.
4. Store a valid entry for `b.test` without first calling `cleanup()`.
5. The store returns `.rejected_capacity`, even though the only existing origin is entirely expired.

This is the same behavior previously identified in the first review: expired state can permanently block attacker-controlled key cardinality until an external maintenance call happens to run. The issue requires expiry cleanup and key-cardinality pressure to work as cache behavior; callers should not have to predictively call `cleanup()` before every insertion.

The fix must retain the new failure-atomicity guarantee. Do **not** simply restore an eager mutating purge before allocation. Instead, extend the dry-run plan so it:

1. identifies globally expired entries that are eligible for removal (skipping leased server entries);
2. computes distinct-origin cardinality after those planned removals;
3. rejects a genuinely over-cap new origin only after that calculation;
4. includes the selected expiry removals in the same victim plan as count/byte-pressure evictions;
5. reserves every required allocation; and
6. commits the full plan only after all fallible work succeeds.

Add client and server regressions for `max_origins = 1`, expired `a.test`, then successful `b.test` insertion with no explicit cleanup. Also retain a late-allocation-failure variant proving the expired entry remains untouched when the insertion returns `.storage_failed`.

### 2. Blocking: client batch LRU reservation can assign `maxInt(u64)` twice

`reserveLruSequenceBatchLocked` currently checks:

```zig
if (self.next_lru_sequence > std.math.maxInt(u64) - (n - 1))
    self.renumberLruSequencesLocked();
const first = self.next_lru_sequence;
self.next_lru_sequence = first +| n;
```

This ensures the last value in the current batch is representable, but it does **not** ensure the next counter value remains representable and distinct.

For `n = 2` and `next_lru_sequence = maxInt(u64) - 1`:

- the guard does not renumber because equality is allowed;
- the batch assigns `max - 1` and `max`;
- `first +| n` saturates `next_lru_sequence` to `max`;
- the next `n = 1` reservation also does not renumber (`max > max` is false);
- `max` is assigned a second time.

That creates duplicate LRU sequence values and makes later eviction depend on incidental physical order instead of the required deterministic recency order. The existing `client LRU batch reservation near u64 boundary` test reaches this exact boundary, but only checks that the newly stored entry survives, so it does not detect the duplicate sequence.

Reserve space for both the batch **and the next counter value**, for example:

```zig
std.debug.assert(n > 0);
if (self.next_lru_sequence > std.math.maxInt(u64) - n)
    self.renumberLruSequencesLocked();

const first = self.next_lru_sequence;
self.next_lru_sequence = first + n; // ordinary addition is now proven safe
return first;
```

Equivalently, renumber whenever the current batch would touch `maxInt(u64)`. Add tests for:

- `n = 1`, `next_lru_sequence = max`;
- `n = 2`, `next_lru_sequence = max - 1`;
- exact uniqueness/ordering of every live `lru_sequence` after the reservation; and
- at least two subsequent pressure insertions asserting the exact victims, rather than merely asserting that the newest entry survives.

### CI

CI run 873 was still in progress during this review. The completed unit-test and appliance-profile jobs I checked were green, including macOS, but the workflow had not yet reached a final conclusion. The two findings above are code blockers regardless of the eventual CI result.

Jaruso (commented):
## Review verdict: one blocking correctness issue remains

I re-reviewed head `26442b9a770e9ea6df0775e523c74814ccf3fe87` against the two findings from the previous round and then re-ran the surrounding insertion/lease/capacity logic.

The two targeted fixes are correct:

- expired origins are now planned for purge without sacrificing failure atomicity;
- client batch-LRU reservation now renumbers before the counter can land on and later reuse `maxInt(u64)`.

CI run **875** is also fully green.

I am not re-raising the client offer-lease scope in this review: this PR now explicitly ships reusable-only client tickets and leaves the issue open as partial work, with client single-use selection semantics deferred.

### Blocking: the revised server planner ignores leased survivors when enforcing all three capacity limits

`planInsertionLocked` now separates the cache into:

- `leased_count` / `leased_bytes` / `origin_leased`; and
- `global_count` / `global_bytes` / `origin_count`, which contain only live **unleased** entries.

That split is useful for identifying evictable entries, but the actual eviction loops compare the limits only against the unleased totals:

```zig
while (origin_count >= self.limits.max_entries_per_origin) { ... }

while (global_count >= self.limits.max_entries or
    global_bytes + new_bytes > self.limits.max_total_bytes)
{
    ...
}
```

The earlier feasibility checks only prove that the insertion *could* fit after enough unleased victims are removed. They do not enforce the final combined totals. As written, a mix of pinned and ordinary entries can commit above every configured bound.

#### Reproduction: global entry count

With `max_entries = 2`:

1. Store single-use `A` and acquire its lease.
2. Store reusable `B`.
3. Insert `C`.

The planner sees `leased_count = 1` and `global_count = 1`.

- `leased_count >= 2` is false.
- `global_count >= 2` is also false.

No victim is selected, so `C` commits and the cache now contains **3 entries despite a limit of 2**.

#### Reproduction: per-origin count

With `max_entries_per_origin = 2`, put the same three records in one origin:

- `origin_leased = 1`
- `origin_count = 1`

The loop checks only `origin_count >= 2`, so the third record commits and the bucket grows to **3 entries**.

#### Reproduction: total bytes

With `max_total_bytes = 1200`, suppose:

- leased entries account for 600 bytes;
- live unleased entries account for 600 bytes;
- the new entry accounts for 600 bytes.

Both current checks pass independently:

```text
leased_bytes + new_bytes == 1200
global_bytes + new_bytes == 1200
```

but the committed total is **1800 bytes**.

This violates #364's hard bounds and also means pinned entries can cause deterministic eviction policy to be skipped rather than forcing eviction of the oldest eligible unleased entry.

### Expected correction

Keep the split counters, but enforce limits against the combined set of entries that will survive the planned purge:

```zig
while (origin_leased + origin_count >= self.limits.max_entries_per_origin) {
    const victim = self.findOldestUnleasedInOriginExcluding(origin, victims.items) orelse unreachable;
    try victims.append(self.allocator, victim);
    origin_count -= 1;
    global_count -= 1;
    global_bytes -= self.entries.get(victim).?.bytes;
}

while (leased_count + global_count >= self.limits.max_entries or
    leased_bytes + global_bytes + new_bytes > self.limits.max_total_bytes)
{
    const victim = self.findOldestUnleasedGlobalExcluding(victims.items) orelse unreachable;
    try victims.append(self.allocator, victim);
    global_count -= 1;
    global_bytes -= self.entries.get(victim).?.bytes;
    if (std.mem.eql(u8, &self.entries.get(victim).?.origin, &origin) and origin_count > 0) {
        origin_count -= 1;
    }
}
```

Use checked/subtractive comparisons if necessary so the byte-limit arithmetic itself cannot overflow.

Please add regressions covering all three mixed cases:

1. one leased + one unleased at the global entry cap, then insert and assert the unleased LRU victim is evicted;
2. one leased + one unleased in the same origin at the per-origin cap;
3. leased bytes + unleased bytes + incoming bytes exceeding `max_total_bytes`, while enough unleased bytes exist to make room.

Each test should assert the leased entry survives, the correct unleased entry is removed, and final `count`, origin-bucket length, and `total_bytes` are at or below their configured limits.

Once this is corrected, I do not currently have another code blocker on this head.

Jaruso (commented):
## Re-review verdict: ready to merge

Reviewed exact head `85acbe4dca573a8cab909f987343d80791287e68` against the remaining mixed-lease capacity blocker and the previously reviewed #364 cache-owned scope.

The final blocker is fixed correctly:

- `max_entries` is now enforced against `leased_count + global_count` before admitting the incoming entry.
- `max_entries_per_origin` is enforced against `origin_leased + origin_count`.
- `max_total_bytes` is enforced against `leased_bytes + global_bytes + new_bytes`.
- Victim selection still excludes leased entries, so pressure evicts only eligible unleased entries or rejects when pinned state alone leaves no possible room.
- The existing preflight checks cover the no-victim cases, so the new `orelse unreachable` paths remain valid.
- Expired-entry planning, origin-bucket survival accounting, failure atomicity, and caller ownership are unchanged by this correction.

The three new regression tests independently cover global entry count, per-origin count, and byte pressure with a pinned single-use entry plus an ordinary evictable entry. They verify that the evictable entry is removed, the insertion succeeds without exceeding the configured limit, and pinned state remains intact.

CI run 877 completed successfully. I found no remaining code blocker on this head. This is merge-ready for the explicitly documented cache-owned/partial #364 scope; the reusable-only client boundary and client selected-offer lease remain deferred to #365 as documented.

PR #482, opened 2026-07-21
tls: drive native negotiation from policy registry

Jaruso:
## Summary

- Add protocol-version and absent-ALPN fields to TLS policy, plus standard QUIC/record policy constructors.
- Extend shared negotiation parsing with observer metadata, raw signature-scheme preservation, hello-level selection, absent-ALPN handling, and credential-signature validation helpers.
- Wire the shared TLS 1.3 backend to validated policy for ClientHello offers, ServerHello validation, server hello selection, ALPN checks, and CertificateRequest/signature policy.
- Source QUIC transport-parameter extension ID from the TLS registry while keeping QUIC transport-parameter payload handling in the adapter.

Closes #386 after merge.

## Validation

- `git diff --check`

### DIFF
--- a/src/quic/tls_backend.zig
+++ b/src/quic/tls_backend.zig
@@ -29,7 +29,7 @@ pub const max_message_len = shared.max_message_len;
 pub const max_certificate_len = shared.max_certificate_len;
 pub const testdata = shared.testdata;
 
-const ext_quic_transport_parameters: u16 = 57;
+const ext_quic_transport_parameters: u16 = @intFromEnum(tls_core.algorithms.ExtensionType.quic_transport_parameters);
 const tp_max_idle_timeout: u64 = 0x01;
 const tp_max_udp_payload_size: u64 = 0x03;
 const tp_initial_max_data: u64 = 0x04;
@@ -206,11 +206,13 @@ pub const Tls13Backend = struct {
     }
 
     pub fn initClientWithOptions(entropy: Entropy, trust: Trust, options: ClientOptions) Tls13Backend {
-        return .{ .engine = shared.Tls13Backend.initClientWithOptions(entropy, trust, .{ .extension = .{
-            .alpn = "h3",
-            .extension_type = ext_quic_transport_parameters,
-            .local = "",
-        } }, options) };
+        return .{ .engine = shared.Tls13Backend.initClientConfigured(entropy, trust, .{
+            .policy = tls_core.policy.Policy.quicDefault(),
+            .transport = .{ .extension = .{
+                .extension_type = ext_quic_transport_parameters,
+                .local = "",
+            } },
+        }, options) };
     }
 
     pub fn initClientWithAllocator(allocator: std.mem.Allocator, entropy: Entropy, trust: Trust) HandshakeError!Tls13Backend {
@@ -225,11 +227,13 @@ pub const Tls13Backend = struct {
     }
 
     pub fn initServer(entropy: Entropy, identity: Identity) Tls13Backend {
-        return .{ .engine = shared.Tls13Backend.initServer(entropy, identity, .{ .extension = .{
-            .alpn = "h3",
-            .extension_type = ext_quic_transport_parameters,
-            .local = "",
-        } }) };
+        return .{ .engine = shared.Tls13Backend.initServerConfigured(entropy, identity, .{
+            .policy = tls_core.policy.Policy.quicDefault(),
+            .transport = .{ .extension = .{
+                .extension_type = ext_quic_transport_parameters,
+                .local = "",
+            } },
+        }) };
     }
 
     pub fn initServerWithAllocator(allocator: std.mem.Allocator, entropy: Entropy, identity: Identity) Tls13Backend {
@@ -239,11 +243,13 @@ pub const Tls13Backend = struct {
     }
 
     pub fn initServerWithProvider(entropy: Entropy, provider: CredentialProvider) Tls13Backend {
-        return .{ .engine = shared.Tls13Backend.initServerWithProvider(entropy, provider, .{ .extension = .{
-            .alpn = "h3",
-            .extension_type = ext_quic_transport_parameters,
-            .local = "",
-        } }) };
+        return .{ .engine = shared.Tls13Backend.initServerWithProviderConfigured(entropy, provider, .{
+            .policy = tls_core.policy.Policy.quicDefault(),
+            .transport = .{ .extension = .{
+                .extension_type = ext_quic_transport_parameters,
+                .local = "",
+            } },
+        }) };
     }
 
     pub fn initServerWithAllocatorAndProvider(allocator: std.mem.Allocator, entropy: Entropy, provider: CredentialProvider) Tls13Backend {
@@ -336,7 +342,6 @@ pub const Tls13Backend = struct {
         const self: *Tls13Backend = @ptrCast(@alignCast(ptr));
         const encoded = try encodeTransportParametersBound(params, self.cid_binding, &self.local_transport_parameters);
         self.engine.profile = .{ .extension = .{
-            .alpn = self.alpn,
             .extension_type = ext_quic_transport_parameters,
             .local = encoded,
         } };
@@ -748,9 +753,9 @@ test "QUIC handshake uses the shared reloadable SNI provider" {
 test "QUIC handshake owner tears down shared and adapter storage on failure" {
     var harness = try RealHandshakeHarness.init();
     defer harness.deinit();
-    harness.server_backend.alpn = "h2";
+    harness.client_backend.engine.trust = .{ .pinned_certificate = "not the server certificate" };
     try harness.wire();
-    try std.testing.expectError(error.AlpnMismatch, harness.run());
+    try std.testing.expectError(error.CertificateInvalid, harness.run());
     harness.deinit();
     try expectQuicBackendWiped(&harness.client_backend);
     try expectQuicBackendWiped(&harness.server_backend);
--- a/src/tls/negotiation.zig
+++ b/src/tls/negotiation.zig
@@ -9,6 +9,8 @@ const policy_mod = @import("policy.zig");
 pub const Error = messages.Error || error{
     MalformedExtension,
     MissingSupportedVersions,
+    MissingExtension,
+    IllegalParameter,
     UnsupportedProtocolVersion,
     MissingCipherSuites,
     NoMutualCipherSuite,
@@ -38,6 +40,8 @@ pub const ClientHelloOffers = struct {
     key_shares_len: usize = 0,
     signature_schemes: [max_offers]algorithms.SignatureScheme = undefined,
     signature_schemes_len: usize = 0,
+    raw_signature_schemes: [max_offers]u16 = undefined,
+    raw_signature_schemes_len: usize = 0,
     alpn_protocols: [max_offers]algorithms.ProtocolName = undefined,
     alpn_protocols_len: usize = 0,
     server_name: ?[]const u8 = null,
@@ -72,6 +76,12 @@ pub const ClientHelloOffers = struct {
         self.signature_schemes_len += 1;
     }
 
+    fn appendRawSignatureScheme(self: *ClientHelloOffers, value: u16) Error!void {
+        if (self.raw_signature_schemes_len == self.raw_signature_schemes.len) return error.OfferVectorTooLarge;
+        self.raw_signature_schemes[self.raw_signature_schemes_len] = value;
+        self.raw_signature_schemes_len += 1;
+    }
+
     fn appendAlpn(self: *ClientHelloOffers, value: algorithms.ProtocolName) Error!void {
         if (self.alpn_protocols_len == self.alpn_protocols.len) return error.OfferVectorTooLarge;
         self.alpn_protocols[self.alpn_protocols_len] = value;
@@ -100,12 +110,47 @@ pub const Selection = struct {
     server_name: ?[]const u8,
 };
 
+pub const ExtensionObservation = struct {
+    id: u16,
+    data: []const u8,
+    data_offset_in_body: usize,
+    is_last: bool,
+};
+
+pub const ExtensionObserver = struct {
+    ctx: *anyopaque,
+    observeFn: *const fn (*anyopaque, ExtensionObservation) Error!void,
+
+    fn observe(self: ExtensionObserver, observation: ExtensionObservation) Error!void {
+        return self.observeFn(self.ctx, observation);
+    }
+};
+
+pub const ParsedClientHello = struct {
+    offers: ClientHelloOffers,
+    legacy_session_id: []const u8,
+    offered_null_compression: bool,
+};
+
+pub const HelloSelection = struct {
+    version: algorithms.ProtocolVersion,
+    cipher_suite: algorithms.CipherSuite,
+    named_group: algorithms.NamedGroup,
+    key_share: []const u8,
+    alpn: ?algorithms.ProtocolName,
+    server_name: ?[]const u8,
+};
+
 pub fn parseClientHello(body: []const u8) Error!ClientHelloOffers {
+    return (try parseClientHelloObserved(body, null)).offers;
+}
+
+pub fn parseClientHelloObserved(body: []const u8, observer: ?ExtensionObserver) Error!ParsedClientHello {
     var offers = ClientHelloOffers{};
     var r = messages.Reader{ .bytes = body };
     if (try r.u16_() != algorithms.legacy_version) return error.MalformedHandshake;
     _ = try r.slice(32);
-    _ = try r.slice(try r.u8_());
+    const legacy_session_id = try r.slice(try r.u8_());
 
     var cipher_reader = messages.Reader{ .bytes = try r.slice(try r.u16_()) };
     if (cipher_reader.remaining() == 0 or cipher_reader.remaining() % 2 != 0) return error.MissingCipherSuites;
@@ -122,11 +167,24 @@ pub fn parseClientHello(body: []const u8) Error!ClientHelloOffers {
     }
     if (!has_null_compression) return error.MalformedHandshake;
 
-    var extensions = messages.ExtensionIterator.init(try r.slice(try r.u16_()));
+    const extensions_bytes = try r.slice(try r.u16_());
+    var extensions = messages.ExtensionIterator.init(extensions_bytes);
+    var saw_supported_versions = false;
     try r.expectEnd();
     while (try extensions.next()) |extension| {
+        if (observer) |obs| {
+            try obs.observe(.{
+                .id = extension.id,
+                .data = extension.data,
+                .data_offset_in_body = @intFromPtr(extension.data.ptr) - @intFromPtr(body.ptr),
+                .is_last = extensions.reader.remaining() == 0,
+            });
+        }
         switch (algorithms.fromInt(algorithms.ExtensionType, extension.id) orelse continue) {
-            .supported_versions => try parseSupportedVersions(extension.data, &offers),
+            .supported_versions => {
+                saw_supported_versions = true;
+                try parseSupportedVersions(extension.data, &offers);
+            },
             .supported_groups => try parseSupportedGroups(extension.data, &offers),
             .signature_algorithms => try parseSignatureAlgorithms(extension.data, &offers),
             .key_share => try parseKeyShares(extension.data, &offers),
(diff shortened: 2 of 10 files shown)


src/tls/tls13_backend.zig
```diff
@@ -221,6 +235,23 @@ pub fn recordAlpnPolicy(comptime protocol: []const u8) AlpnPolicy {
...
 
+fn defaultPolicyForProfile(profile: TransportProfile) tls_policy.Policy {
+    return switch (profile) {
+        .record => |options| blk: {
+            const protocols = options.alpn.protocols;
+            if (protocols.len == 1 and std.mem.eql(u8, protocols[0], "h2")) break :blk tls_policy.Policy.recordH2Only();
+            if (protocols.len == 1 and std.mem.eql(u8, protocols[0], "http/1.1")) break :blk tls_policy.Policy.recordHttp1Only(options.alpn.allow_absent);
+            break :blk tls_policy.Policy.recordDefault();
+        },
+        .extension => tls_policy.Policy.quicDefault(),
```
Jaruso: This compatibility inference keeps ALPN in `TransportProfile` and creates two authorities instead of completing the separation required by #386. Canonical profiles use `self.policy`, while arbitrary record ALPNs later switch back to `profile`; a caller can also build `BackendConfig` with a policy and transport ALPN that disagree. Please remove ALPN from the transport union and migrate the actual record listener construction to pass the pinned listener `Policy` directly. A target shape is:

```zig
pub const TransportProfile = union(enum) {
    record,
    extension: struct { extension_type: u16, local: []const u8 },
};

pub const BackendConfig = struct {
    policy: tls_policy.Policy,
    transport: TransportProfile,
};
```

Then delete `defaultPolicyForProfile`, `usesLegacyProfileAlpn`, and the profile ALPN helpers rather than retaining a compatibility path.

src/tls/negotiation.zig
```diff
@@ -122,9 +166,18 @@ pub fn parseClientHello(body: []const u8) Error!ClientHelloOffers {
...
-    var extensions = messages.ExtensionIterator.init(try r.slice(try r.u16_()));
+    const extensions_bytes = try r.slice(try r.u16_());
+    var extensions = messages.ExtensionIterator.init(extensions_bytes);
     try r.expectEnd();
     while (try extensions.next()) |extension| {
+        if (observer) |obs| {
+            try obs.observe(.{
+                .id = extension.id,
+                .data = extension.data,
+                .data_offset_in_raw = @intFromPtr(extension.data.ptr) - @intFromPtr(body.ptr),
```
Jaruso: `body` begins after the four-byte handshake header, so this value is an offset within the ClientHello **body**, not within the captured framed `raw` message. Using it to slice the framed ClientHello for `pre_shared_key` binder verification points four bytes late and produces the wrong truncated transcript. Either have this parser accept the framed message and derive the body internally, or expose an explicitly body-relative offset and add the header length exactly once in the backend. Please add a regression that verifies a resumed handshake through this observer path.

src/tls/tls13_backend.zig
```diff
@@ -2424,6 +2602,13 @@ pub const Tls13Backend = struct {
                 },
             }
         }
+        const offers = tls_negotiation.parseClientHello(body) catch |err| return mapNegotiationError(err);
```
Jaruso: This is a second complete ClientHello parse after the function has already manually parsed cipher suites, compression, supported versions/groups, signatures, SNI, key shares, ALPN, PSK, and the transport extension. That violates the single-pass requirement and leaves two parsers whose results can diverge; notably the backend ignores `offers.raw_signature_schemes` and `hello_selection.server_name` in favor of the manual copies.

Please call `parseClientHelloObserved` once near the start, with an observer that handles only `psk_key_exchange_modes`, `pre_shared_key` (including last-extension and raw-offset capture), and the configured opaque transport extension. Copy the selected key share, SNI, ALPN, and raw signature list into stable backend storage before async credential selection, then remove the manual standard-extension switch.

src/tls/tls13_backend.zig
```diff
@@ -1156,6 +1221,41 @@ pub const Tls13Backend = struct {
         }
     }
 
+    fn validateNativePolicy(self: *const Tls13Backend) HandshakeError!void {
```
Jaruso, line 1113: This validates supported enum *values* but not every encoding bound that must be rejected before `core.start()`. For example, a policy containing 128 repeated `.tls13` entries passes this function and `clientHelloEncodedLen()`, then traps at `@intCast(2 * protocol_versions.len)` into `u8`. On the server, a large repeated signature list or near-maximum ALPN list can pass `start()`, emit ServerHello/install secrets, and only then overflow while serializing CertificateRequest/EncryptedExtensions.

Please preflight all vector count/byte limits and the maximum server flight here, before lifecycle/entropy/transcript mutation. This validation should also reject policy/transport-mode mismatches and all transport-extension collisions, including `server_name` (type 0), which `TransportProfile.validate()` currently permits even though the parser consumes it as SNI.

src/quic/tls_backend.zig
```diff
@@ -340,6 +341,8 @@ pub const Tls13Backend = struct {
             .extension_type = ext_quic_transport_parameters,
             .local = encoded,
         } };
+        self.policy_alpn_storage[0] = .{ .bytes = self.alpn };
+        self.engine.policy = tls_core.policy.Policy.fromCapabilities(.quic, shared.native_capabilities, self.policy_alpn_storage[0..1]);
```
Jaruso: The QUIC adapter is still rebuilding algorithm/ALPN policy during `start()` from mutable `self.alpn`. #386 requires the opposite ownership: construct the shared engine once with canonical `Policy.quicDefault()` / `h3`, and let `start()` update only the connection-specific opaque transport-parameter payload. Keeping this mutation also permits a QUIC connection to advertise a non-HTTP/3 ALPN. Please use the configured constructor at initialization and remove `policy_alpn_storage` plus this policy rewrite.

Jaruso (commented):
**Review result: changes required before merge.** The connected account owns this PR, so GitHub does not permit a formal `REQUEST_CHANGES` review; these are nevertheless blocking findings.

The current CI matrix is passing, but this does not yet satisfy #386’s implementation contract.

The main blockers are:

1. **Negotiation policy and transport configuration are still competing authorities.** `TransportProfile` still owns ALPN, the configured constructors are unused, record-mode call sites were not migrated from the pinned listener policy, and QUIC still rewrites policy at `start()`. This leaves behavior dependent on whether an ALPN happens to match the compatibility classifier rather than making `Policy` the sole authority.
2. **ClientHello is still parsed twice.** The backend manually parses every standard negotiation extension, then calls the shared parser over the same body. The issue explicitly requires one authoritative parse with an observer limited to PSK and the opaque transport extension.
3. **The new observer metadata is not safe for PSK binder capture.** Its offset is body-relative despite being named `data_offset_in_raw`, so applying it to the framed ClientHello is four bytes off.
4. **Native-policy validation is incomplete.** It validates enum values but not every wire length / full server-flight bound before lifecycle mutation. Valid-but-repeated entries can trap a cast or make the server fail only after ServerHello and secret emission.
5. **Missing-vs-unsupported `supported_versions` is still conflated.** A present extension containing no recognized/mutual version is reported as `MissingSupportedVersions`; #386 explicitly requires those cases to remain distinct.
6. **The required regression matrix was not added.** There are no changed direct-handshake, record listener, PSK/resumption, QUIC parity, or configured-policy tests. The two new negotiation unit tests do not cover the acceptance criteria or the new configured constructors.

Expected direction:

- Make `BackendConfig.policy` the only source of versions/ciphers/groups/signatures/SNI/ALPN behavior; make the transport union contain only record-vs-opaque-extension data.
- Update record listener and QUIC construction to pass immutable policy at initialization. QUIC should pin canonical `h3`; `start()` should update only connection-specific transport-parameter bytes.
- Replace the manual standard-extension switch in `onClientHello` with one `parseClientHelloObserved` call. The observer should handle only PSK modes, PSK raw offsets/final-extension enforcement, and the configured opaque transport extension.
- Add exact pre-start serialization validation for all policy vectors and the worst-case server flight.
- Add the issue’s required direct handshake and PSK/transport regression tests before closing #386.

Jaruso (commented):
**Second review result: changes still required before merge.**

The revision materially fixes the prior review: ALPN is now policy-owned, ClientHello has one authoritative parse with a PSK/transport observer, the PSK body offset is corrected, native-policy preflight is much stronger, the record listener passes its pinned policy, and QUIC no longer rebuilds policy during `start()`. I resolved all five previous inline threads. CI is also green on this head.

Four blockers remain:

1. An explicitly present but empty ALPN extension is indistinguishable from an absent extension, so an HTTP/1.1 fallback policy accepts malformed wire input.
2. Duplicate ClientHello extensions regress from `illegal_parameter` to the malformed/decode-error path.
3. The observer collapses PSK identity/binder count mismatch into malformed syntax, losing the existing `illegal_parameter` classification.
4. The typed hello selection is still discarded and the added shared validation helpers are not used by the backend; fixed constants/manual checks remain authoritative for ServerHello, credential contexts, and ticket state. That leaves a stated #386 implementation requirement incomplete even though the current native capability set has only one version/cipher/group.

Please add direct regressions for the three wire-error cases and complete the typed negotiated-state/shared-validation handoff before closing #386. As before, GitHub does not permit this account to formally request changes on its own PR, so this is submitted as a blocking commented review.

src/tls/negotiation.zig
```diff
@@ -135,28 +193,48 @@ pub fn parseClientHello(body: []const u8) Error!ClientHelloOffers {
...
     const cipher_suite = pickEnum(algorithms.CipherSuite, policy.cipher_suites, offers.cipher_suites[0..offers.cipher_suites_len]) orelse
         return error.NoMutualCipherSuite;
     const named_group = pickEnum(algorithms.NamedGroup, policy.named_groups, offers.supported_groups[0..offers.supported_groups_len]) orelse
         return error.NoMutualNamedGroup;
     const key_share = offers.keyShareFor(named_group) orelse return error.MissingKeyShare;
-    const signature_scheme = pickEnum(algorithms.SignatureScheme, policy.signature_schemes, offers.signature_schemes[0..offers.signature_schemes_len]) orelse
-        return error.NoMutualSignatureScheme;
-    const alpn = pickAlpn(policy.alpn_protocols, offers.alpn_protocols[0..offers.alpn_protocols_len]) orelse
+    const alpn = pickAlpn(policy.alpn_protocols, offers.alpn_protocols[0..offers.alpn_protocols_len]) orelse blk: {
+        if (offers.alpn_protocols_len == 0 and policy.allow_absent_alpn) break :blk null;
```
Jaruso, line 230: This treats `offers.alpn_protocols_len == 0` as proof that the ALPN extension was absent, but `parseAlpn()` currently accepts an explicitly present `ProtocolNameList` with a zero length. Under `recordHttp1Only(true)`, a malformed ClientHello containing `application_layer_protocol_negotiation = 00 00` therefore falls through this branch and is accepted as the no-ALPN HTTP/1.1 fallback.

Please either reject an empty list in `parseAlpn()` or retain an `alpn_extension_present` bit and distinguish present-empty from absent. Add a regression using `buildClientHello(... .alpn_protocols = &.{})` against the fallback policy; it should fail as malformed, while `null` remains the only accepted absent case.

src/tls/tls13_backend.zig
```diff
@@ -1287,6 +1300,34 @@ pub const Tls13Backend = struct {
         };
     }
 
+    fn mapNegotiationError(err: tls_negotiation.Error) HandshakeError {
+        return switch (err) {
+            error.MalformedHandshake,
+            error.MalformedExtension,
+            error.MissingCipherSuites,
+            error.OfferVectorTooLarge,
+            error.DuplicateExtension,
```
Jaruso: `DuplicateExtension` should not be collapsed into `MalformedHandshake` here. The pre-migration backend's `ExtensionGuard` returned `IllegalParameter` for repeated ClientHello extension types, matching RFC 8446 §4.2. The new shared iterator detects the duplicate correctly, but this mapping changes the wire alert to the decode-error path.

Move `error.DuplicateExtension` to the `IllegalParameter` arm and add a backend regression with a duplicated standard extension (for example `supported_versions` or ALPN), asserting `error.IllegalParameter` before credential selection or output.

src/tls/tls13_backend.zig
```diff
@@ -2276,174 +2357,79 @@ pub const Tls13Backend = struct {
...
+                const self_obs: *@This() = @ptrCast(@alignCast(ctx));
+                switch (observation.id) {
+                    pre_shared_key.ext_psk_key_exchange_modes => {
+                        self_obs.psk_dhe_ke_offered = pre_shared_key.hasMode(observation.data, .psk_dhe_ke) catch
+                            return error.MalformedExtension;
+                        self_obs.psk_modes_seen = true;
+                    },
+                    pre_shared_key.ext_pre_shared_key => {
+                        if (!observation.is_last) return error.IllegalParameter;
+                        _ = pre_shared_key.OfferedPsks.parse(observation.data) catch return error.MalformedExtension;
```
Jaruso: This catch collapses every `OfferedPsks.parse` failure to `MalformedExtension`, but the existing backend deliberately maps an identities/binders count mismatch to `IllegalParameter` (`mapPskReadError` still documents that behavior). The single-parser migration now changes that peer-visible classification.

Preserve the `pre_shared_key.ReadError` distinction in the observer—at minimum `CountMismatch => error.IllegalParameter`, with framing/empty/invalid-length cases remaining malformed—and add a `raw_ext_data` test with unequal identity/binder counts.

src/tls/tls13_backend.zig
```diff
@@ -2276,174 +2357,79 @@ pub const Tls13Backend = struct {
...
-        if (psk_ext != null and !psk_modes_seen) return error.MissingExtension;
+        };
+
+        var observer = ClientHelloObserver{ .transport_extension_type = self.profile.extensionType() };
+        const parsed = tls_negotiation.parseClientHelloObserved(body, .{
+            .ctx = &observer,
+            .observeFn = ClientHelloObserver.observe,
+        }) catch |err| return mapNegotiationError(err);
+        const offers = parsed.offers;
+        const hello_selection = tls_negotiation.negotiateServerHello(self.policy, &offers) catch |err| return mapNegotiationError(err);
```
Jaruso, line 2410: The shared negotiation result is still only used for its key-share/ALPN/SNI slices; `version`, `cipher_suite`, and `named_group` are discarded. Later the server emits fixed constants, selection/verification contexts and ticket state use fixed constants, client ServerHello validation remains manual, and the newly added `validateServerSelection` / `validateCredentialSignature` helpers are not authoritative anywhere.

#386 explicitly requires typed negotiated state to survive asynchronous selection and be used at those boundaries. Please store the typed hello selection before `beginServerSelection`, use it for ServerHello emission, credential contexts, resumption state and diagnostics, and route provider-selected signature/client selections through the shared validation helpers. The current one-algorithm capability makes the values coincide today, but leaving the fixed path in place means the registry integration is incomplete and will diverge as soon as native capabilities expand.

src/tls/tls13_backend.zig
```diff
@@ -2276,174 +2357,79 @@ pub const Tls13Backend = struct {
...
+        if (offers.raw_signature_schemes_len == 0) return error.MissingExtension;
+        if (offers.raw_signature_schemes_len > self.peer_sig_schemes.len) return error.MalformedHandshake;
+        @memcpy(self.peer_sig_schemes[0..offers.raw_signature_schemes_len], offers.raw_signature_schemes[0..offers.raw_signature_schemes_len]);
+        self.peer_sig_scheme_count = offers.raw_signature_schemes_len;
+        if (hello_selection.key_share.len != X25519.public_length) return error.MalformedHandshake;
+        const client_share = hello_selection.key_share[0..X25519.public_length].*;
+
+        if (hello_selection.alpn) |protocol| {
+            self.setSelectedAlpn(protocol.bytes);
+        } else if (!self.policy.allow_absent_alpn) {
```
Jaruso, line 2432: When this branch accepts an absent ALPN, `selected_alpn_present` correctly remains false, but every authentication context later calls `selectedAlpnForAuth()`, which converts that state to the non-null empty slice `""`. Both `credentials.SelectionContext.application_protocol` and `VerificationContext.application_protocol` are optional specifically so absence is represented as `null`; `some("")` is a different value and can make an SNI credential provider, client-certificate selector, or peer verifier apply the wrong policy.

Please preserve the optional value through those contexts—preferably remove `selectedAlpnForAuth()` and assign `self.selectedAlpn()` directly. Add a fallback-policy handshake test whose server selector and client verifier both assert `application_protocol == null`; a negotiated non-empty ALPN should remain `some(name)`.

src/tls/tls13_backend.zig
```diff
@@ -3255,6 +3242,10 @@ pub const Tls13Backend = struct {
         const selection = self.serverSelectionContext();
         if (!selection.offersScheme(credential.scheme))
             return self.failCredential(.invalid_callback_behavior);
+        const policy_scheme = tls_algorithms.fromInt(tls_algorithms.SignatureScheme, credential.scheme.code()) orelse
+            return self.failCredential(.invalid_callback_behavior);
+        if (!self.policy.containsSignatureScheme(policy_scheme))
```
Jaruso: This validates that the returned scheme was offered and allowed, but not that it is compatible with the selected credential's leaf public key. An external provider can return `.ecdsa_secp256r1_sha256` with an Ed25519 certificate (or the inverse); the engine then accepts the callback, emits the certificate, and signs a CertificateVerify whose scheme cannot authenticate that certificate. #386 explicitly requires selected credential/key compatibility to be validated before the authentication flight.

Parse the leaf here and enforce at least:

```zig
.ed25519 => parsed.pub_key_algo == .curveEd25519,
.ecdsa_secp256r1_sha256 => parsed.pub_key_algo == .{ .X9_62_id_ecPublicKey = .X9_62_prime256v1 },
```

(adapted to the actual union matching syntax). Treat a mismatch as `.invalid_callback_behavior`, release the handle, and emit no ServerHello/auth bytes. Cover both synchronous and resumed async selection so the same validation cannot be bypassed after `server_select` resumes.

Jaruso (commented):
Re-review of `59a33e5` — **changes are still required before merge**.

The revision materially fixes the original five structural findings: policy now owns ALPN, the transport profile is transport-only, QUIC no longer rebuilds policy at `start()`, the ClientHello is parsed once through the observer path, and local policy validation is substantially stronger. CI is green.

Four existing unresolved threads still block completion: present-but-empty ALPN being accepted as absent, duplicate-extension alert regression, PSK `CountMismatch` alert regression, and failure to persist/use typed negotiated version/cipher/group state throughout the engine.

I found two additional blockers in the revised path:

1. accepted absent ALPN is exposed to credential providers/verifiers as a present empty protocol rather than `null`;
2. a provider-selected signature scheme is not checked against the selected certificate's public-key algorithm before the server flight is emitted.

Once these six items are addressed, the next review should focus primarily on regression coverage and exact AC closure rather than architecture rework.

src/tls/tls13_backend.zig
```diff
@@ -3241,6 +3268,19 @@ pub const Tls13Backend = struct {
         certificate_message_len: usize,
     };
 
+    fn leafSupportsSignatureScheme(der: []const u8, scheme: credentials.SignatureScheme) HandshakeError!bool {
+        const parsed = (Certificate{ .buffer = der, .index = 0 }).parse() catch
+            return error.MalformedHandshake;
```
Jaruso, line 3249: This compatibility check is currently used only for server credentials. The handshake-time client-auth path (`emitSelectedClientCertificate` → `emitClientCertificate`) validates offer/policy membership and chain framing, but never calls this helper, so a client provider can return an ECDSA scheme with an Ed25519 leaf and the engine emits an invalid Certificate/CertificateVerify flight before the peer rejects it. Factor the selected-credential validation so both roles—including resumed async `client_select`—call `tls_negotiation.validateCredentialSignature(...)`, validate the chain, and apply `leafSupportsSignatureScheme()` before recording or emitting authentication bytes. Add sync and async client-auth mismatch tests asserting local `invalid_callback_behavior`, one release, and no client Certificate flight.

Jaruso (commented):
**Review result: one blocking change remains.**

The latest revision resolves all six findings from the previous pass: present-empty ALPN is rejected, duplicate-extension and PSK count-mismatch classifications are restored, negotiated version/cipher/group state is persisted and used, absent ALPN remains `null` in authentication contexts, and server credentials are checked against their leaf key before any flight is emitted.

The remaining problem is the equivalent handshake-time **client credential** path. `emitSelectedClientCertificate()` validates only that the selected scheme was offered by the server and is present in local policy. `emitClientCertificate()` then validates only chain framing and size. Unlike `inspectSelectedServerCredential()`, neither path confirms that the selected scheme is compatible with the leaf certificate key.

That permits a client credential provider to return an ECDSA scheme with an Ed25519 leaf (or the reverse). The engine will emit the client Certificate, invoke the signer, and send CertificateVerify before the server eventually rejects the mismatch. This is a local provider-contract violation and must fail locally before any client-authentication bytes are recorded or emitted. It also leaves the new shared `tls_negotiation.validateCredentialSignature()` helper non-authoritative; both server and client paths still reproduce its checks manually.

Please factor the shared selected-credential checks so both roles, including resumed async selection, perform:

1. `tls_negotiation.validateCredentialSignature(policy, peer_raw_offers, selected_scheme)`;
2. non-empty/bounded chain validation;
3. `leafSupportsSignatureScheme(chain.entries[0], credential.scheme)`;
4. exact encoded-size validation before transcript/output mutation.

Add synchronous and asynchronous client-auth regressions using an Ed25519 leaf with `scheme_override = .ecdsa_secp256r1_sha256`. Both should fail as local `invalid_callback_behavior`, release the selected handle exactly once, and emit no client Certificate flight.

CI's unit and appliance/integration jobs are passing on this head; packaging, dependency audit, and perf jobs were still running when reviewed.

Jaruso (commented):
**Re-review result: no remaining blocking code findings.**

The final client-auth asymmetry is fixed on this head:

- `inspectSelectedCredential` is now the shared validation path for both server credentials and handshake-time client credentials.
- It validates the provider-selected scheme against the peer's raw offer and local policy, validates the complete chain and encoded size, and verifies that the leaf key type supports the selected signature scheme before any Certificate flight is recorded or emitted.
- `emitSelectedClientCertificate` uses that helper in both the synchronous path and the resumed `.client_select` path.
- The new synchronous and asynchronous regressions assert `invalid_callback_behavior`, exactly one credential release, zero signing calls, and zero emitted client handshake bytes.

I also rechecked the previously reported policy/transport separation, single-pass ClientHello parsing, PSK binder offset and error classification, typed negotiated-state propagation, absent-ALPN semantics, QUIC policy immutability, and server credential compatibility. I found no new regression in those paths.

CI run 871 is still in progress at the time of this review, although the unit-test jobs, appliance listener/integration jobs, format, lint, security, dependency review, and PKI differential jobs visible so far are passing. **Code review is clear; merge once the remaining CI jobs complete successfully.**

PR #483, opened 2026-07-21
tls: add HRR core and codec groundwork

Jaruso:
## Summary
- add TLS HRR extension/error vocabulary and unsupported_extension alert mapping
- add transcript helpers for synthetic message_hash rebinding and non-mutating append hashes
- add first-class Core HRR transitions for HRR and ClientHello2
- add a transport-neutral hello_retry codec with sentinel detection, HRR encode/decode, and ClientHello2 mutation validation
- teach negotiation to distinguish usable key shares from retry decisions

## Validation
- zig build test

## Issue
- Refs #333

## Notes
- This is a draft foundation PR, not the full #333 implementation. Backend orchestration, retry-aware PSK binders, cookie provider integration, QUIC/record loopbacks, and OpenSSL interop remain for follow-up commits before closing the issue.


src/tls/algorithms.zig
```diff
@@ -50,8 +50,11 @@ pub const ExtensionType = enum(u16) {
     supported_groups = 10,
     signature_algorithms = 13,
     application_layer_protocol_negotiation = 16,
+    padding = 21,
+    early_data = 42,
     pre_shared_key = 41,
     supported_versions = 43,
+    cookie = 44,
```
Jaruso, line 57: **[P2] Reserve these new TLS-owned extension IDs from opaque transport profiles.** `TransportProfile.ExtensionOptions.validate()` currently rejects the other TLS-owned IDs, but it still permits `padding` (21), `early_data` (42), and `cookie` (44). A caller can therefore configure extension 44 as its transport extension and have the shared backend emit bytes that the HRR/ClientHello code interprets as a TLS cookie. Please add all three IDs to the reserved switch and cover them with configuration tests, for example:

```zig
switch (options.extension_type) {
    // existing TLS-owned IDs...
    @intFromEnum(tls_algorithms.ExtensionType.padding),
    @intFromEnum(tls_algorithms.ExtensionType.early_data),
    @intFromEnum(tls_algorithms.ExtensionType.cookie),
    => return error.InvalidTransportProfile,
    else => {},
}
```

src/tls/handshake.zig
```diff
@@ -214,6 +216,62 @@ pub const Core = struct {
...
 
+    pub fn acceptHelloRetryRequest(self: *Core, raw: []const u8) Error!Message {
+        if (self.role != .client or
+            self.handshake_lifecycle != .running or
+            self.expected_inbound != .server_hello or
+            self.retry_state != .none)
+            return error.UnexpectedHandshakeMessage;
+        const message = messages.decode(raw) catch return error.MalformedHandshake;
+        if (message.kind != .server_hello) return error.UnexpectedHandshakeMessage;
+        self.transcript.rebindClientHello();
```
Jaruso, line 236: **[P1] Require proof that ClientHello1 was actually recorded before rebinding.** `start()` sets the client’s `expected_inbound` to `.server_hello`, and the normal ClientHello send path currently leaves `handshake_state == .idle`. Consequently `acceptHelloRetryRequest()` succeeds immediately after `start()` and replaces the transcript with `message_hash(Hash(empty))`.

Use the existing `.client_hello` state (or an equivalent explicit flight phase) to distinguish “not sent” from “CH1 sent”:

```zig
// advanceAfterSend(.client_hello)
self.handshake_state = .client_hello;
self.expected_inbound = .server_hello;

// acceptHelloRetryRequest
if (self.handshake_state != .client_hello) {
}
```

Please add a test that `start(); acceptHelloRetryRequest(hrr)` is rejected and that the transcript remains unchanged.

src/tls/handshake.zig
```diff
@@ -214,6 +216,62 @@ pub const Core = struct {
...
+            self.handshake_state != .server_hello or
+            self.retry_state != .none)
+            return error.UnexpectedHandshakeMessage;
+        const message = messages.decode(raw) catch return error.MalformedHandshake;
+        if (message.kind != .server_hello) return error.UnexpectedHandshakeMessage;
+        self.transcript.rebindClientHello();
+        self.transcript.update(message.raw);
+        self.retry_state = .hrr_sent;
+        self.handshake_state = .idle;
+        self.expected_inbound = .client_hello;
```
Jaruso, line 255: **[P1] The generic Core APIs can bypass both dedicated ClientHello2 transitions.** With `expected_inbound = .client_hello`, `acceptReceived(ch2)` passes the ordinary expected-message check and advances the server without calling `acceptSecondClientHello()`. Symmetrically, after the client accepts HRR its state still permits `recordSent(ch2)`, bypassing `recordSecondClientHello()`.

That defeats the invariant documented in #333 that only the retry-specific methods may perform these transitions, and it is especially dangerous because `Tls13Backend.drainInput()` currently calls `acceptReceived()` generically.

Add explicit guards (or a stricter flight-state enum), for example:

```zig
// recordSent
if (message.kind == .client_hello and self.retry_state == .hrr_received)

// acceptReceived
if (message.kind == .client_hello and self.retry_state == .hrr_sent)
```

Then test that the generic methods reject CH2 while the dedicated methods accept it.

src/tls/hello_retry.zig
```diff
@@ -0,0 +1,474 @@
...
+    if ((try r.u8_()) != 0) return error.IllegalParameter;
+
+    const extensions_bytes = try r.slice(try r.u16_());
+    try r.expectEnd();
+    var extensions = messages.ExtensionIterator.init(extensions_bytes);
+    var selected_version: ?algorithms.ProtocolVersion = null;
+    var selected_group: ?algorithms.NamedGroup = null;
+    var cookie: ?[]const u8 = null;
+
+    while (try extensions.next()) |extension| {
```
Jaruso: **[P1] Normalize duplicate extensions to `IllegalParameter` at this boundary.** `ExtensionIterator.next()` reports `error.DuplicateExtension` before the later per-field duplicate checks run. If this error is passed through the current Core mapping it becomes `MalformedHandshake`, but duplicate HRR extensions are a semantic `illegal_parameter` failure. The same issue exists in `ClientHelloView.parse()` for duplicate ClientHello2 extensions.

Please catch and normalize the iterator error in both parsers:

```zig
const extension = extensions.next() catch |err| switch (err) {
    error.DuplicateExtension => return error.IllegalParameter,
    else => return err,
};
```

Add cases for duplicate `supported_versions`, `key_share`, `cookie`, and an ordinary duplicated ClientHello2 extension, asserting the exact fatal alert mapping.

src/tls/hello_retry.zig
```diff
@@ -0,0 +1,474 @@
...
+        .cipher_suite = cipher_suite,
+        .selected_group = selected_group,
+        .cookie = cookie,
+    };
+}
+
+pub fn encode(request: EncodeRequest, out: []u8) Error![]const u8 {
+    if (request.selected_group == null and request.cookie == null) return error.IllegalParameter;
+    if (request.cookie) |cookie| {
+        if (cookie.len == 0 or cookie.len > std.math.maxInt(u16)) return error.IllegalParameter;
```
Jaruso: **[P1] Validate all vector bounds before the narrowing casts.** The current cookie check allows lengths 65,534 and 65,535, but the encoded extension data is `2 + cookie.len`; the cast at the write site will trap. `legacy_session_id_echo.len` is also cast to `u8` without enforcing TLS’s 32-byte maximum, so sufficiently large caller input can trap instead of returning an error.

At minimum:

```zig
if (request.legacy_session_id_echo.len > 32)

if (request.cookie) |cookie| {
    if (cookie.len == 0 or cookie.len > std.math.maxInt(u16) - 2)
}
```

Because this encoder also uses a fixed 512-byte body buffer, either expose/document a smaller explicit cookie bound or preflight the exact encoded size so oversized cookies fail deterministically rather than halfway through writing. Add boundary tests around 32/33-byte session IDs and the maximum accepted cookie.

src/tls/hello_retry.zig
```diff
@@ -0,0 +1,474 @@
...
+    }
+
+    if (nextComparable(first, first_index, false) != first.len or
+        nextComparable(second, second_index, true) != second.len)
+        return error.IllegalParameter;
+
+    if (findExtension(second, @intFromEnum(algorithms.ExtensionType.early_data)) != null)
+        return error.IllegalParameter;
+
+    if (request.cookie) |expected_cookie| {
```
Jaruso, line 240: **[P1] Enforce that `pre_shared_key` is still the final ClientHello2 extension.** Since cookie is skipped during sequence comparison, a second ClientHello ordered as `..., pre_shared_key, cookie` currently passes. RFC 8446 requires `pre_shared_key` to be last, and #333 calls this out explicitly.

Validate both parsed extension lists before returning:

```zig
fn requirePskLast(extensions: []const ExtensionView) Error!void {
    for (extensions, 0..) |extension, index| {
        if (extension.id == @intFromEnum(algorithms.ExtensionType.pre_shared_key) and
            index + 1 != extensions.len)
    }
}
```

For an HRR cookie plus PSK, the encoder must insert `cookie` before `pre_shared_key`. Add positive and negative ordering tests.

src/tls/hello_retry.zig
```diff
@@ -0,0 +1,474 @@
...
+            (skip_cookie and id == @intFromEnum(algorithms.ExtensionType.cookie)))
+            continue;
+        return index;
+    }
+    return index;
+}
+
+fn compareExtensionPayload(lhs: ExtensionView, rhs: ExtensionView, request: Request) Error!void {
+    if (lhs.id == @intFromEnum(algorithms.ExtensionType.key_share)) {
+        const group = request.selected_group orelse return error.IllegalParameter;
```
Jaruso: **[P1] Cookie-only HRR is valid and must preserve the original key share.** `selected_group` is optional by design: a server may issue HRR solely to request a cookie. In that case ClientHello2 retains ClientHello1’s `key_share` unchanged, but this `orelse` rejects every such flow.

Handle the two cases separately:

```zig
if (lhs.id == @intFromEnum(algorithms.ExtensionType.key_share)) {
    if (request.selected_group) |group| {
        try expectSingleKeyShare(group, rhs.data);
    } else if (!std.mem.eql(u8, lhs.data, rhs.data)) {
    }
    return;
}
```

Please add a cookie-only positive test with an unchanged X25519 share and a negative test where that share changes.

src/tls/hello_retry.zig
```diff
@@ -0,0 +1,474 @@
...
+    }
+    return null;
+}
+
+fn samePskIdentityVector(first: []const u8, second: []const u8) bool {
+    var first_reader = messages.Reader{ .bytes = first };
+    var second_reader = messages.Reader{ .bytes = second };
+    const first_identities = first_reader.slice(first_reader.u16_() catch return false) catch return false;
+    const second_identities = second_reader.slice(second_reader.u16_() catch return false) catch return false;
+    return std.mem.eql(u8, first_identities, second_identities);
```
Jaruso: **[P1] Compare PSK identities semantically and validate the complete PSK extension.** The raw identities vector includes each `obfuscated_ticket_age`, which RFC 8446 explicitly permits ClientHello2 to update. This comparison therefore rejects a valid retry whenever an age changes. Conversely, it never parses or validates the binder vector, so malformed/missing binders can pass this helper.

Reuse `pre_shared_key.OfferedPsks.parse()` for both extensions, require equal counts, and compare only identity bytes in order while ignoring ages and binder contents:

```zig
const first_psks = pre_shared_key.OfferedPsks.parse(first) catch
const second_psks = pre_shared_key.OfferedPsks.parse(second) catch
if (first_psks.count != second_psks.count) return error.IllegalParameter;

var a = first_psks.pairs();
var b = second_psks.pairs();
while (try a.next()) |lhs| {
    const rhs = (try b.next()) orelse return error.IllegalParameter;
    if (!std.mem.eql(u8, lhs.identity.identity, rhs.identity.identity))
}
```

This preserves identity count/order, permits age and binder recomputation, and rejects malformed or count-mismatched binder vectors. Add tests for changed age, changed binder, changed identity, reordered identities, malformed binder length, and identity/binder count mismatch.

Jaruso (commented):
**Changes required — this draft is not merge-ready.**

GitHub does not allow the connected account to formally request changes on its own pull request, so this is submitted as a comment review. The findings below are blocking.

I found several correctness issues in the new foundation code and left inline comments with concrete fixes. The most important are:

- the client Core can accept an HRR before ClientHello1 has been recorded, producing `message_hash(Hash(empty))`;
- generic Core send/receive APIs can bypass the dedicated ClientHello2 transitions;
- the ClientHello2 validator rejects valid cookie-only HRR flows;
- PSK comparison rejects permitted ticket-age updates while failing to validate the binder vector;
- a cookie placed after `pre_shared_key` is accepted even though PSK must remain the final ClientHello extension;
- unchecked vector-length casts can trap instead of returning a protocol error;
- duplicate HRR/ClientHello2 extensions currently escape as `DuplicateExtension`, which would map to the wrong fatal alert;
- the newly registered TLS extension IDs are not reserved from opaque transport-extension configuration.

The issue-level acceptance criteria are also still incomplete on this commit, as the PR description acknowledges. Before this PR can satisfy #333 it still needs:

1. Client backend orchestration: HRR detection before generic ServerHello acceptance, retained ClientHello1 state, fresh requested X25519 share, exact cookie echo, final ServerHello tuple enforcement, and no secret installation or Initial-key discard at HRR.
2. Server backend orchestration: retry decision before credential/PSK selection, retained first-ClientHello context, ClientHello2 validation before state overwrite, and only then the normal async credential path.
3. Retry-aware PSK binder derivation and verification over `message_hash(Hash(CH1)) || HRR || Truncate(CH2)`, including async-capture lifetime and zeroization.
4. The cookie provider seam and bounded retained retry context.
5. QUIC and record-mode loopback tests proving HRR remains in the Initial epoch and that both transport profiles use the shared engine.
6. Independent transcript/traffic-secret fixtures, the full illegal-mutation and fatal-alert matrix, duplicate/no-op/repeated-HRR tests on both roles, and PSK-plus-async tests.
7. OpenSSL interoperability in both directions, including proof from the handshake trace that HRR actually occurred.
8. Cleanup and borrowed transport-extension lifetime updates through ClientHello2.

CI is green on the reviewed commit, but the passing suite does not cover the protocol cases above. Please keep the PR in draft, address the inline defects, and complete the remaining #333 acceptance criteria before marking it ready for another review.

src/tls/handshake.zig
```diff
@@ -141,6 +143,8 @@ pub const Core = struct {
         if (self.handshake_lifecycle != .running and self.handshake_lifecycle != .complete)
             return error.InvalidHandshakeState;
         const message = messages.decode(raw) catch return error.MalformedHandshake;
+        if (message.kind == .client_hello and self.retry_state == .hrr_sent)
```
Jaruso, line 146: **[P1] Reject a normal ServerHello before ClientHello1 as well.** The new `handshake_state == .client_hello` check protects `acceptHelloRetryRequest()`, but the generic `acceptReceived()` path still accepts `.server_hello` immediately after `start()`: the client starts with `expected_inbound = .server_hello`, `handshake_state = .idle`, and this new guard only handles a server-side ClientHello2. That advances the client to `encrypted_extensions` and hashes a ServerHello into a transcript that never contained CH1.

Apply the same flight proof to every inbound ServerHello, while preserving the post-CH2 case:

```zig
if (self.role == .client and
    message.kind == .server_hello and
    self.handshake_state != .client_hello)
{
}
```

The later `expected_inbound` check will still reject a final ServerHello between HRR and CH2 because HRR sets it to `null`; after `recordSecondClientHello()` it is restored to `.server_hello`. Add tests that `start(); acceptReceived(server_hello)` fails without changing the transcript, while both CH1→SH and CH1→HRR→CH2→SH remain accepted.

src/tls/hello_retry.zig
```diff
@@ -0,0 +1,843 @@
...
+        const ext_type = algorithms.fromInt(algorithms.ExtensionType, extension.id) orelse
+            return error.UnsupportedExtension;
+        switch (ext_type) {
+            .supported_versions => {
+                if (selected_version != null) return error.IllegalParameter;
+                selected_version = try decodeSelectedVersion(extension.data, original_offers);
+            },
+            .key_share => {
+                if (selected_group != null) return error.IllegalParameter;
+                selected_group = try decodeSelectedGroup(extension.data, original_offers);
```
Jaruso, line 74: **[P1] Treat HRR `key_share` as unsolicited when CH1 omitted the extension.** `decodeSelectedGroup()` checks that the group was in `supported_groups` and was not already shared, but it never checks `original_offers.key_share_seen`. As a result, an external server can return a `key_share` HRR response to a ClientHello that never offered the `key_share` extension, and this decoder accepts it. TLS extension responses must have been offered by the client, except for HRR `cookie`; this case must surface `UnsupportedExtension`, not synthesize a newly offered extension in CH2.

Guard the extension before decoding it:

```zig
.key_share => {
    if (!original_offers.key_share_seen)
        return error.UnsupportedExtension;
    if (selected_group != null)
    selected_group = try decodeSelectedGroup(extension.data, original_offers);
},
```

Add a test with X25519 in `supported_groups` but `key_share_seen == false`, asserting `error.UnsupportedExtension` and the `unsupported_extension` alert mapping. Keep the existing empty-vector case (`key_share_seen == true`, zero entries) valid so it can trigger HRR.

src/tls/hello_retry.zig
```diff
@@ -0,0 +1,843 @@
...
+    }
+}
+
+fn nextComparable(extensions: []const ExtensionView, start: usize, skip_cookie: bool) usize {
+    var index = start;
+    while (index < extensions.len) : (index += 1) {
+        const id = extensions[index].id;
+        if (id == @intFromEnum(algorithms.ExtensionType.padding) or
+            id == @intFromEnum(algorithms.ExtensionType.early_data) or
+            (skip_cookie and id == @intFromEnum(algorithms.ExtensionType.cookie)))
```
Jaruso, line 254: **[P2] Validate padding contents before ignoring padding mutations.** Skipping the entire padding extension means CH2 is accepted even when its padding payload contains arbitrary non-zero bytes. RFC 7685 requires the ClientHello padding payload to be empty or entirely zero; the normal negotiation parser also ignores this extension, so there is no later validation that will catch it.

Validate both parsed ClientHellos before `compareExtensions()` and continue ignoring only the length/placement during comparison:

```zig
fn validatePadding(extensions: []const ExtensionView) Error!void {
    for (extensions) |extension| {
        if (extension.id == @intFromEnum(algorithms.ExtensionType.padding) and
            !std.mem.allEqual(u8, extension.data, 0))
        {
        }
    }
}
```

Add positive tests for empty and resized zero-filled padding and a negative test for any non-zero CH2 padding byte.

src/tls/events.zig
```diff
@@ -50,6 +50,10 @@ pub const HandshakeError = error{
     /// whose vector is empty or malformed, which remains a decode or parameter
     /// failure. Maps to `missing_extension` (RFC 8446 §6).
     MissingExtension,
+    /// A peer sent an extension that is syntactically valid but not permitted
```
Jaruso: **[P2] Narrow this error's documented contract to unsolicited extension responses.** The current wording—“recognized but not permitted in the current handshake context”—describes the condition that RFC 8446 maps to `illegal_parameter`, and it conflicts with `hello_retry.decode()`, which correctly returns `IllegalParameter` for recognized extensions forbidden in HRR.

Please document `UnsupportedExtension` as the unsolicited-response case instead, for example:

```zig
/// The peer returned an extension for which the local endpoint did not send
/// the corresponding request, except where the protocol explicitly permits
/// an unsolicited extension (such as cookie in HelloRetryRequest).
/// Maps to `unsupported_extension`.
```

This distinction is important because future call sites will otherwise select the wrong fatal alert.

src/tls/transcript.zig
```diff
@@ -51,3 +63,45 @@ test "transcript hash matches direct SHA-256 and supports HRR-style replacement"
...
+}
+
+test "rebindClientHello produces deterministic HRR transcript fixture" {
+    const client_hello_1 = [_]u8{ 1, 0, 0, 3, 0x01, 0x02, 0x03 };
+    const hello_retry_request = [_]u8{ 2, 0, 0, 2, 0x04, 0x05 };
+    const client_hello_2 = [_]u8{ 1, 0, 0, 4, 0x06, 0x07, 0x08, 0x09 };
+
+    var transcript = Transcript{};
+    transcript.update(&client_hello_1);
+    const ch1_hash = transcript.peek();
```
Jaruso: **[P2] Make the HRR transcript fixture independent of the method under test.** The expected synthetic message currently embeds `ch1_hash` obtained from the same `Transcript.peek()` state used by `rebindClientHello()`. That weakens the fixture and does not satisfy #333's explicit requirement that the expected transcript not be constructed through the same helper path.

Use a separately computed or hard-coded known digest, ideally with the final expected hash fixed as bytes:

```zig
var independent_ch1_hash: [digest_len]u8 = undefined;
Hash.hash(&client_hello_1, &independent_ch1_hash, .{});
@memcpy(synthetic[4..], &independent_ch1_hash);
```

A stronger fixture hard-codes both `Hash(CH1)` and the final `Hash(message_hash(Hash(CH1)) || HRR || CH2)` values. The eventual issue-level test must also assert the derived traffic secrets, not only transcript equality.

Jaruso (commented):
Re-review of `b295f702121af1fba2b4a8bdcefa8e69237ab1d2` against issue #333.

The follow-up commit correctly addresses all eight findings from the first pass: the dedicated retry transitions can no longer be bypassed, CH1 is required before the HRR-specific transition, cookie-only HRR works, PSK identities are compared semantically, PSK remains last, duplicate extensions normalize to `illegal_parameter`, encoder lengths are preflighted, and the newly registered TLS extension IDs are reserved from opaque transport profiles.

This is still **not merge-ready as the implementation of #333**. The inline comments identify two remaining protocol/state correctness bugs and three validation/API-test issues in the foundation itself. In addition, the PR remains intentionally incomplete relative to the issue acceptance criteria: the shared backend still rejects the retry decision instead of orchestrating HRR; ClientHello2 generation and retained CH1 context are absent; retry-aware PSK binder derivation/verification is absent; cookie provider integration is absent; final ServerHello tuple locking, Initial-secret/discard invariants, cleanup/zeroization, QUIC and record loopbacks, deterministic traffic-secret fixtures, and OpenSSL interop are not implemented.

CI is green on this head, but the current tests do not cover the newly identified cases or the missing end-to-end acceptance criteria. Please keep the PR in draft and issue #333 open until those items are completed.

Jaruso (commented):
Re-review complete on `d7b5292`.

All five findings from the previous pass are addressed correctly:

- normal ServerHello now requires a recorded ClientHello1;
- an HRR `key_share` response is rejected as `unsupported_extension` when CH1 omitted `key_share`, while the present-empty-vector case remains valid;
- ClientHello padding is validated as empty/zero-filled before padding mutations are ignored;
- `UnsupportedExtension` now documents the unsolicited-response condition accurately;
- the HRR transcript fixture computes `Hash(CH1)` independently of `Transcript.peek()` / `rebindClientHello()`.

I also rechecked the earlier fixes and the surrounding Core state transitions, HRR codec, ClientHello2 mutation comparison, PSK identity handling, extension ordering, bounds checks, negotiation result, alert mappings, and transport-extension reservations. I found no additional code-level defects in the current foundation scope. CI run 879 is green.

There are no new inline findings on this head.

This is still not completion of #333, as the PR description acknowledges. Keep it draft until the backend client/server HRR orchestration, ClientHello2 generation, retry-aware PSK binders, cookie provider integration, final ServerHello tuple enforcement, cleanup/zeroization, QUIC and record loopbacks, deterministic traffic-secret fixtures, and external interoperability coverage are added. Within the explicitly stated groundwork scope, the current code is ready to build upon.

PR #486, opened 2026-07-22
tls: add resumption runtime resolver foundation

Jaruso:
## Summary

Refs #365.

This PR lands the shared TLS resumption resolver/runtime foundation needed by the larger TCP/H3 integration story:

- replaces the server PSK resolver boolean/out-pointer contract with an owned resolve result carrying a live exactly-once lease;
- adapts stateless ticket protection to return a no-op public lease and stateful cache leases to the public resolver contract;
- commits the resolver lease in `Tls13Backend.selectPsk` only after compatibility and binder verification, before a PSK-selected `ServerHello` is emitted;
- adds `tls.resumption_runtime` with disabled/stateful/stateless/hybrid configuration, shared client/server caches, ordered `TDSH`/`TDTK` resolver dispatch, and ephemeral AES-128-GCM stateless key installation;
- exports `session_cache`, `session_cache_persistence`, and `resumption_runtime` from `tls_core`.

This intentionally does not mark #365 closed. Remaining work includes the native TCP/H3 transport wiring, prepared-ticket issuance split, config file boundary, client offer-lease outcome plumbing, and end-to-end resumed HTTP traffic coverage called out by the issue.

## Validation

- `zig build test-session-cache --summary all`
- `zig build test-integration --summary all --error-style verbose` (passed on rerun; first run hit transient proxy restart 502)


src/tls/session_cache.zig
```diff
@@ -80,10 +77,23 @@ const builtin = @import("builtin");
 const crypto = @import("crypto");
 const session = @import("session.zig");
 const pre_shared_key = @import("pre_shared_key.zig");
-const zig_compat = @import("zig_compat");
 
 const secrets = crypto.secrets;
 
+const Mutex = struct {
```
Jaruso: **[P1] Restore the blocking cache mutex; do not use a cache-wide test-and-set spin lock.**

This is not a short, bounded critical section. Both caches hold the mutex across hash-map work, O(n) scans/eviction planning, allocations, deep session clones, persistence snapshots, and restore operations. Under concurrent handshakes or persistence, a waiter here repeatedly performs an atomic RMW on the same cache line and can burn an entire core while the holder is descheduled or doing allocator work. That is materially worse than the `zig_compat.Mutex` this file used before, and it contradicts both `docs/CONCURRENCY.md` and the still-present `build.zig` comment promising an `std.Io.Mutex`-backed lock.

This looks like a build-graph workaround caused by exporting `session_cache` through `tls_core`. Please keep the original implementation and wire the dependency into the root module instead:

```zig
// session_cache.zig
const zig_compat = @import("zig_compat");
const Mutex = zig_compat.Mutex;
```

```zig
// build.zig, after tls_core_mod is created
tls_core_mod.addImport("zig_compat", compat_mod);
```

That preserves the repository's concurrency boundary rather than silently changing the cache's runtime behavior.

src/tls/pre_shared_key.zig
```diff
@@ -481,29 +481,83 @@ pub const ClientPskOfferSet = struct {
...
-/// Provider-neutral identity resolver. `resolveFn` returns `true` with a
-/// completely owned, zero-valued-on-failure `out` when `identity` is a
-/// usable ticket/session identity; `false` (with `out` left untouched/zero)
-/// for any malformed, unknown, retired, or otherwise unusable identity —
-/// including a stateless envelope's own decode/authentication failures
-/// (#363), which are never distinguished from "unknown" at this contract.
+/// Provider-neutral, exactly-once server PSK lease. Stateful single-use
+/// providers pin cache entries until binder verification proves the identity
+/// was actually selected; reusable and stateless providers use `noop`.
+pub const ServerPskLease = struct {
```
Jaruso: **[P1] Make invalid/partially initialized owned leases unrepresentable.**

The issue explicitly requires every resolver contract violation or partially initialized result to be wiped. This public shape permits combinations such as `active = true` with a null `ctx`, a missing `releaseFn`, or a missing `deinitFn`. `release()`/`commit()` then silently clear the wrapper while potentially leaving provider state pinned or heap context leaked. The backend cannot repair that after the resolver returns it.

Please encode the state machine in the type instead of relying on optional callbacks plus an independent boolean. For example:

```zig
pub const ServerPskLease = union(enum) {
    noop,
    owned: Owned,
    finished,

    const Owned = struct {
        ctx: *anyopaque,
        commitFn: *const fn (*anyopaque) void,
        releaseFn: *const fn (*anyopaque) void,
        deinitFn: *const fn (*anyopaque) void,
    };
};
```

`noop()` and an `owned(...)` constructor would then be the only construction paths, with all owned callbacks non-optional. `commit`, `release`, and `deinit` can transition to `.finished` exactly once. Add tests for double commit/release, deinit-as-release, and malformed-provider behavior so this contract is safe for future resolver implementations, not only the two adapters in this PR.

src/tls/resumption_runtime.zig
```diff
@@ -0,0 +1,236 @@
...
+    allocator: std.mem.Allocator,
+    config: Config,
+    clock: Clock,
+    provider: provider.CryptoProvider,
+    client_cache: session_cache.ClientSessionCache,
+    server_cache: session_cache.StatefulServerCache,
+    keyring: ticket_protection.ReloadableKeyRing,
+    stateless_key: [16]u8 = [_]u8{0} ** 16,
+
+    pub fn init(
```
Jaruso, line 63: **[P2] Do not retain a second plaintext copy of the stateless ticket key for the full runtime lifetime.**

`Snapshot.build` copies `key_bytes` into its owned `TicketKeySecret`; after installation this field is no longer used. Keeping `stateless_key` here unnecessarily doubles the lifetime and number of locations containing the raw key.

Generate it as local scratch inside `installEphemeralStatelessKey` and wipe it on every path:

```zig
var key: [16]u8 = undefined;
defer secrets.secureZero(&key);
try self.provider.randomBytes(&key);
// build/install snapshot using &key
```

Then remove the field and the deinit-time wipe. This also makes the success and error-path secret lifetime explicit.

src/tls/resumption_runtime.zig
```diff
@@ -0,0 +1,236 @@
...
+        crypto_provider: provider.CryptoProvider,
+    ) InitError!Runtime {
+        try config.validate();
+        if ((config.mode == .stateless or config.mode == .hybrid) and
+            !crypto_provider.capabilities().supportsAead(.aes_128_gcm))
+            return error.ProviderUnsupported;
+
+        var runtime = Runtime{
+            .allocator = allocator,
+            .config = config,
```
Jaruso, line 76: **[P2] Align ownership with the configured mode instead of always constructing every service.**

#365 says disabled mode installs no caches/keyring/hooks, and the process owner should contain only the services selected by stateful/stateless/hybrid policy. This initializer always constructs the client cache, stateful server cache, and keyring—even for `.disabled`, and even when one server-side mechanism is irrelevant. It also validates irrelevant limits in every mode.

Please represent mode-owned resources explicitly, for example with optionals or a tagged server storage union:

```zig
client_cache: ?ClientSessionCache,
server: union(Mode) {
    disabled,
    stateful: StatefulServerCache,
    stateless: ReloadableKeyRing,
    hybrid: struct { cache: StatefulServerCache, keyring: ReloadableKeyRing },
},
```

The exact shape can differ, but initialization/deinitialization and validation should be mode-specific. Add a disabled-mode allocator test proving no cache/keyring allocation or entropy call occurs.

src/tls/resumption_runtime.zig
```diff
@@ -0,0 +1,236 @@
...
+    }
+
+    pub fn lookupClientOffers(self: *Runtime, candidate: session.CandidateContext) session_cache.ClientLookupResult {
+        if (self.config.mode == .disabled) return .miss;
+        return self.client_cache.lookupOffers(candidate, self.nowUnixMs());
+    }
+
+    pub fn storeClientTicket(self: *Runtime, ticket: *const session.ClientTicketState) session_cache.StoreResult {
+        if (self.config.mode == .disabled) return .rejected_capacity;
+        return self.client_cache.storeClone(ticket, self.nowUnixMs(), usagePolicy(self.config.usage));
```
Jaruso: **[P1] `single_use` currently validates successfully but necessarily discards every client ticket.**

`usagePolicy(.single_use)` is passed into `ClientSessionCache.storeClone`, while that cache explicitly returns `rejected_unsupported_usage` until the client offer-lease lifecycle lands. The result is a runtime configuration that starts successfully but silently cannot resume as a client, contrary to #365's requirement that unsupported/invalid selected configuration fail deterministically rather than run partially initialized.

Until `ClientOfferLease` and selected-index outcome plumbing are included, please either:

1. reject `.single_use` in `Config.validate`, with a test; or
2. split server issuance usage from client storage capability so the unsupported client path cannot be selected accidentally.

Once the client lease lands, replace that temporary restriction with tests proving only the selected single-use offer is consumed and all other offers are released.

src/tls/resumption_runtime.zig
```diff
@@ -0,0 +1,236 @@
...
+        var key_id: ticket_protection.KeyId = undefined;
+        var nonce_prefix: [4]u8 = undefined;
+        self.provider.randomBytes(&key_id) catch return error.EntropyFailure;
+        self.provider.randomBytes(&self.stateless_key) catch return error.EntropyFailure;
+        self.provider.randomBytes(&nonce_prefix) catch return error.EntropyFailure;
+
+        const now = self.nowUnixMs();
+        const decrypt_until = checkedAddMs(now, @as(i64, session.max_lifetime_seconds) * 1000) orelse std.math.maxInt(i64);
+        const config = ticket_protection.KeyConfig{
+            .id = key_id,
```
Jaruso, line 175: **[P1] The one-shot ephemeral key permanently stops ticket issuance after seven days and prematurely invalidates tickets near that boundary.**

`session.max_lifetime_seconds` is the maximum *ticket* lifetime, not an appropriate lifetime for the process's only current encryption key. With no scheduled rotation in this issue, a process that stays up longer than seven days will have no active encryption key forever. Worse, `encrypt_until == decrypt_until`, so a ticket issued one second before that boundary becomes undecryptable one second later even though its advertised lifetime may have almost a full day remaining.

The initial ephemeral policy needs one of these coherent behaviors:

- keep the process-local key active/decryptable for the runtime lifetime (for example, a supported non-expiring/max timestamp window; restart still replaces it and invalidates old tickets); or
- implement rotation before the encryption window closes and retain the old key for at least the maximum lifetime of every ticket issued under it.

Because #365 explicitly says scheduled rotation is not required here, the first option appears to match the intended policy. Add a deterministic clock test that advances beyond seven days and verifies issuance still has an active key, plus a boundary test proving a ticket remains decryptable for its full configured lifetime.

src/tls/tls13_backend.zig
```diff
@@ -2677,14 +2672,11 @@ pub const Tls13Backend = struct {
...
-                elapsedMillis(now, candidate_state.common.issued_at_unix_ms),
+                hit.state.ticket_age_add,
+                elapsedMillis(now, hit.state.common.issued_at_unix_ms),
             );
-            // Move the accepted session into backend ownership instead of
-            // letting the loop-scoped `defer candidate_state.deinit()`
-            // above destroy it — `moveFrom` zero-values `candidate_state`
-            // first, so that deferred deinit becomes a safe no-op.
-            self.selected_server_psk.state.moveFrom(&candidate_state);
+            hit.lease.commit();
```
Jaruso, line 2679: **[P1] Add direct tests for the exactly-once lease lifecycle at this decision point.**

The ordering here is correct, but the modified backend tests only return `noop()` leases. Nothing currently proves the new ownership contract that motivated this PR. Please add a counting owned lease and cover at least:

- incompatible first candidate: release/deinit exactly once, then a later candidate may be tried;
- first compatible candidate with a bad binder: release/deinit exactly once, fatal `DecryptError`, and no later identity probe;
- successful binder: commit exactly once, no release, and commit occurs before the sink sees the PSK-selected `ServerHello`;
- invalid recovered PSK length or binder-verification error: release exactly once;
- resolver operational failure remains fatal without leaking an already-returned result;
- stateful single-use integration: success consumes the cache entry, while incompatibility/binder failure makes it resolvable again.

A small test lease context with `commit_count`, `release_count`, `deinit_count`, and an optional sink-order marker would make regressions in this security-critical transition visible.

src/tls/resumption_runtime.zig
```diff
@@ -0,0 +1,236 @@
...
+
+test "config validates ticket lifetime ceiling" {
+    try std.testing.expectError(error.InvalidConfig, (Config{ .ticket_lifetime_seconds = 0 }).validate());
+    try std.testing.expectError(error.InvalidConfig, (Config{ .ticket_lifetime_seconds = session.max_lifetime_seconds + 1 }).validate());
+    try (Config{ .ticket_lifetime_seconds = session.max_lifetime_seconds }).validate();
+}
+
+test "disabled runtime exposes no server resolver and no client offers" {
+    const TestClock = struct {
+        fn now(_: *anyopaque) i64 {
```
Jaruso: **[P1] The new runtime/composite resolver needs substantive mode and dispatch coverage.**

The only runtime behavior tested here is `.disabled`. Please add deterministic tests for the foundation this PR claims:

- stateful mode resolves a valid `TDSH` handle and returns the public cache lease;
- stateless mode seals/resolves a `TDTK` identity and returns a no-op lease;
- hybrid dispatches `TDSH` only to the cache and `TDTK` only to the protector;
- malformed/unknown prefixes are ordinary misses and never cross-dispatched;
- expired/tampered/unknown identities miss, while provider/storage failures are fatal;
- stateless/hybrid initialization fails deterministically on entropy failure or missing AES-128-GCM capability;
- disabled mode exposes no hooks and initializes no mode-owned resources;
- key lifetime behavior remains valid after the clock advances beyond the ticket-lifetime ceiling.

These tests are necessary because CI currently proves compilation and the older component suites, but not the composition semantics introduced by this file.

Jaruso (commented):
## Review result: changes needed before merge

I reviewed this as the intentionally partial resolver/runtime foundation described in the PR, not as a demand that this single PR finish every remaining TCP/H3, prepared-ticket, config, client-offer, metrics, and end-to-end item in #365.

The `selectPsk` decision ordering is correct: compatibility is checked first, the first compatible binder is verified, and the lease is committed before a PSK-selected `ServerHello` can be emitted. The stateful adapter also transfers the internal lease and recovered state without an obvious ownership leak.

The issues called out inline should be addressed before merge. The two highest-risk correctness problems are the cache mutex regression and the stateless key validity window. The public lease representation and missing lifecycle tests also need strengthening before this becomes the foundation every transport will depend on.

Additional expectations for the revision:

- preserve the existing blocking/parking mutex and wire `zig_compat` into `tls_core` rather than introducing a cache-wide spin lock;
- make the ephemeral current key usable for the lifetime of the process, or add rotation before it expires, while retaining decryptability for every ticket issued during its advertised lifetime;
- make malformed owned leases impossible to construct, or reject them deterministically without leaking/pinning provider state;
- add adversarial lease tests and a real runtime mode/dispatch matrix;
- do not accept `single_use` as a valid runtime configuration while the client cache necessarily rejects every such ticket;
- avoid retaining an unnecessary second plaintext copy of the stateless ticket key in `Runtime`;
- align disabled/stateful/stateless resource ownership with #365 so disabled mode does not initialize unused cache/keyring services and other modes do not carry irrelevant owners.

The CI jobs visible during review were passing; these findings are semantic/concurrency/security issues that the current tests do not exercise.

GitHub does not permit the authenticated PR author to formally request changes on their own pull request, so this is submitted as a comment review rather than a `REQUEST_CHANGES` review.

src/tls/session_cache.zig
```diff
@@ -1363,6 +1361,87 @@ pub const ResolveLeaseResult = union(enum) {
...
+            box.* = .{ .allocator = allocator, .lease = hit.lease };
+            hit.lease.active = false;
+
+            var state: session.ServerRecoverableState = .{};
+            state.moveFrom(&hit.state);
+            return .{ .hit = .{
+                .state = state,
+                .lease = pre_shared_key.ServerPskLease.initOwned(
+                    box,
+                    PublicServerLeaseBox.commit,
```
Jaruso, line 1455: **[P1] Reusable stateful resolutions should not allocate an owned public lease.**

Issue #365 is explicit that **stateless and reusable stateful results use a no-op lease**, while only single-use stateful resolution pins an entry. This adapter currently boxes *every* cache hit and returns `ServerPskLease.initOwned(...)`, including reusable entries. Because the runtime currently only permits `.reusable`, that means every successful stateful resumption pays an extra heap allocation and can turn an otherwise valid reusable ticket into fatal `ResolverFailed` solely because the lease-box allocation failed.

That is both an issue-contract mismatch and an avoidable reliability/performance regression on the hot handshake path.

Please branch on the internal lease usage:

```zig
if (!hit.lease.single_use) {
    // Move the recovered state out and return a true no-op public lease.
    // If reusable LRU recency must only advance after binder success, expose
    // that as a separate non-owning completion hook rather than modeling the
    // reusable ticket as an owned/pinned lease.
    var state: session.ServerRecoverableState = .{};
    state.moveFrom(&hit.state);
    hit.lease.active = false;
    return .{ .hit = .{
        .state = state,
        .lease = pre_shared_key.ServerPskLease.initNoop(),
    } };
}

// Only single-use hits need PublicServerLeaseBox.
```

The exact recency mechanism can differ, but the public lease contract should preserve the distinction required by #365: reusable/stateless = no-op; single-use = owned exactly-once lease. Please add a test proving a reusable `TDSH` hit returns `.noop` and does not allocate a lease box.

src/tls/tls13_backend_tests.zig
```diff
@@ -4270,6 +4320,131 @@ test "a compatible candidate with a wrong binder is fatal and never probes a lat
...
+    try std.testing.expectEqual(@as(usize, 1), incompatible_lease.deinit_count);
+    try std.testing.expectEqual(@as(usize, 1), compatible_lease.commit_count);
+    try std.testing.expectEqual(@as(usize, 0), compatible_lease.release_count);
+    try std.testing.expectEqual(@as(usize, 1), compatible_lease.deinit_count);
+    try std.testing.expect(server.core.psk_authenticated);
+}
+
+test "resolver lease releases bad-binder candidate before fatal failure and probes no later identity" {
+    var server = tls_backend.Tls13Backend.initServer(serverEntropy(), fixtureIdentity(), .record);
+    defer server.deinit();
```
Jaruso, line 4366: **[P2] Add one real `StatefulServerCache` single-use integration test; synthetic lease probes are not enough.**

The new `LeaseProbe` tests are good and prove the backend's generic commit/release ordering, but they do not prove that the actual `StatefulServerCache -> PublicServerLeaseBox -> ServerPskLease -> Tls13Backend` chain obeys the single-use semantics required by #365.

Please add an integration-style unit test using the real cache adapter that demonstrates both outcomes:

1. insert a `.single_use` `TDSH` state, resolve it through the public resolver, complete a valid binder handshake, and prove the handle is consumed/misses afterward;
2. repeat with incompatibility or a bad binder and prove the lease is released so the same handle can be resolved again.

This is the ownership boundary most likely to regress because it crosses the internal lease, heap box, public lease, backend defer, and cache generation/epoch checks. The synthetic callback counters validate only the middle of that chain.

Jaruso (commented):
## Re-review result: substantially improved, but I still want two items tightened

The second commit addresses the major findings from the first review well. In particular:

- the cache is back on `zig_compat.Mutex` rather than a spin lock;
- the public lease is now a tagged state machine with non-optional owned callbacks;
- the stateless key is local scratch and wiped after snapshot installation;
- the ephemeral key remains active for a long-lived process instead of expiring after seven days;
- disabled/stateful/stateless/hybrid ownership is now mode-aware;
- unsupported `single_use` configuration is rejected until the client-offer lifecycle lands;
- backend tests now directly prove release/commit ordering, including commit before the PSK-selected `ServerHello`;
- runtime tests cover the principal mode/dispatch paths and entropy/capability failures.

The large `zig_compat` import rewrite is broad, but it is understandable here: once `session_cache.zig` consumes `zig_compat` through the named `tls_core` dependency, the rest of the executable graph needs to refer to the same module instance rather than importing the source file again under a second module identity. The current build/test matrix is also passing on the jobs completed so far.

I still see one contract-level problem in the stateful adapter and one integration-test gap. I would address those before merging this foundation.

src/tls/session_cache.zig
```diff
@@ -1363,6 +1361,97 @@ pub const ResolveLeaseResult = union(enum) {
...
+    identity: []const u8,
+    now_unix_ms: i64,
+) pre_shared_key.ResolveError!pre_shared_key.ServerPskResolveResult {
+    var result = cache.resolveLease(identity, now_unix_ms);
+    switch (result) {
+        .hit => |*hit| {
+            if (!hit.lease.single_use) {
+                var state: session.ServerRecoverableState = .{};
+                state.moveFrom(&hit.state);
+                hit.lease.active = false;
```
Jaruso: **[P1] Returning `.noop` here currently drops the reusable cache's confirmed-use LRU update.**

The no-allocation/public-no-op part of this fix is correct, but `hit.lease.active = false` permanently discards the internal reusable lease before the backend can confirm that this identity actually won compatibility + binder verification.

That internal lease is not only a single-use pin. Its documented reusable behavior is:

```zig
// ServerLease.commit()
self.cache.commitLease(..., self.single_use);

// commitLease(..., single_use = false)
e.lru_sequence = self.reserveFreshLruSequenceLocked();
```

and the cache explicitly says this recency update is deliberately deferred until **binder-verified use**, rather than `resolveLease`, so successfully reused sessions become MRU. With this branch, every reusable public resolution skips that call forever. A frequently and successfully resumed ticket therefore remains at its insertion-time LRU position and can be evicted as the oldest entry.

This also undercuts #364's required deterministic LRU/recency-update behavior.

Please keep the requirements achieved by this change — **no heap box and a public `.noop` lease for reusable state** — but preserve the post-binder completion signal separately. For example, add an allocation-free, non-owning reusable completion hook/token to the resolve result that carries the cache generation + entry id and is invoked by `selectPsk` after binder success, while the public ownership lease remains `.noop`. Do not simply call `commit()` here at resolve time: that would let incompatible or bad-binder attempts refresh LRU and reverses the cache's intentional confirmed-use policy.

Add a regression test with at least two reusable entries that proves:

1. resolving/rejecting one does **not** change its LRU position;
2. successfully selecting/binder-verifying one does make it MRU;
3. subsequent capacity pressure evicts the genuinely least-recently-successfully-used entry.

Once this is preserved, the previous reusable-allocation concern is resolved cleanly.

Jaruso (commented):
## Re-review: one remaining behavior regression before merge

The two requests from the previous pass are mostly addressed correctly: reusable stateful resolution no longer allocates `PublicServerLeaseBox`, and the real single-use cache/backend integration tests now cover both commit/consume and bad-binder release/re-resolve.

There is one remaining regression in the reusable branch, called out inline: disarming the internal reusable `ServerLease` while returning the required public `.noop` lease also removes the cache's post-binder LRU recency update. That conflicts with the existing cache contract and #364's required deterministic recency updates.

Please preserve the public no-op/no-allocation contract **and** provide an allocation-free way to refresh reusable recency only after compatibility + binder success. Once that is fixed with a focused regression test, I do not see another blocker in this latest delta.

CI for the new head is still running; the completed unit-test jobs I checked are passing.

Jaruso (commented):
## Re-review result: merge-ready

Re-reviewed the current head (`bacc35f`) against the remaining reusable-LRU blocker and the prior lease/runtime findings.

The reusable stateful path now preserves all required properties at once:

- public reusable hits still return a true `.noop` `ServerPskLease`;
- no `PublicServerLeaseBox` allocation is reintroduced for reusable hits;
- an allocation-free `on_selected` hook carries only cache identity/generation data;
- the hook runs only after compatibility and binder verification succeed;
- `commitLease(..., false)` therefore refreshes reusable LRU recency only on confirmed use;
- the new integration test proves rejected reuse does not refresh recency, successful binder-selected reuse does, and subsequent capacity pressure evicts the actual LRU entry.

The real single-use cache/backend commit/release integration coverage from the prior revision remains intact. I did not find a new correctness, ownership, concurrency, or security regression in this delta.

The additional CHANGELOG entry is consistent with the foundation landed by this PR and does not change the merge assessment.

CI for the current head is green. GitHub will not allow me to formally approve my own PR, so this is submitted as a comment review: **I consider this ready to merge.**

PR #489, opened 2026-07-22
tls: add client single-use PSK offer leases

Jaruso:
## Summary
- add a transport-neutral client PSK offer lease with selected/not-selected/aborted outcomes
- allow ClientSessionCache to store, restore, pin, release, and consume single-use tickets safely
- wire TLS backend selected_identity handling to finish leases exactly once, including ClientHello filtering and fatal-index cleanup
- guard client persistence against outstanding single-use leases

## Tests
- zig build test-integration

Closes #487


src/tls/session_cache.zig
```diff
@@ -154,9 +145,9 @@ pub const StoreResult = enum {
     /// Stateful-server-only: bounded CSPRNG handle-collision retries were
     /// exhausted.
     rejected_handle_generation_failed,
-    /// Client-cache-only: `.single_use` is not supported by this PR's
-    /// client cache (see module doc) — rejected explicitly rather than
-    /// silently stored and re-offered as if it were reusable.
+    /// Legacy/adapter-only result value retained for callers that still
+    /// surface unsupported usage from outside this cache. `ClientSessionCache`
+    /// itself accepts `.single_use`.
```
Jaruso, line 147: **Blocking — #487’s runtime/config acceptance criterion is still unmet.** This correctly makes `ClientSessionCache` accept `.single_use`, but `src/tls/resumption_runtime.zig` was not changed. `Config.validate()` still contains:

```zig
if (self.usage == .single_use) return error.InvalidConfig;
```

and the runtime test still asserts that `.single_use` is invalid. The issue explicitly requires removing that temporary rejection once the client lease lifecycle exists, so callers still cannot actually enable this through the process-shared `Runtime`.

Please remove the blanket rejection and invert/add runtime coverage. For example:

```zig
pub fn validate(self: Config) error{InvalidConfig}!void {
    if (self.ticket_lifetime_seconds == 0 or
        self.ticket_lifetime_seconds > session.max_lifetime_seconds)
        return error.InvalidConfig;

    if (self.mode != .disabled) {
        self.session_limits.validate() catch return error.InvalidConfig;
        self.client_cache_limits.validate() catch return error.InvalidConfig;
    }
    if (self.mode == .stateful or self.mode == .hybrid)
        self.server_cache_limits.validate() catch return error.InvalidConfig;
}
```

Then add a runtime-level regression along the lines of:

```zig
try (Config{ .mode = .stateful, .usage = .single_use }).validate();

var runtime = try Runtime.init(
    std.testing.allocator,
    .{ .mode = .stateful, .usage = .single_use },
    clock.clock(),
    provider_impl.cryptoProvider(),
);
defer runtime.deinit();

var ticket = try sampleClientTicket(...);
defer ticket.deinit();
try std.testing.expectEqual(.stored, runtime.storeClientTicket(&ticket));

var lookup = runtime.lookupClientOffers(candidate);
defer lookup.deinit();
try std.testing.expect(lookup == .hit);
try std.testing.expect(lookup.hit.active);
```

That proves the feature is exposed at the runtime boundary rather than only being reachable by direct cache construction.

src/tls/tls13_backend.zig
```diff
@@ -1782,9 +1847,12 @@ pub const Tls13Backend = struct {
...
         var psk_secret: ?[hash_len]u8 = null;
+        const active_offers = self.mutableClientPskOffers();
         if (selected_identity) |idx| {
-            if (idx >= self.client_psk_offers.len) return error.IllegalParameter;
-            self.client_psk_offers.takeSelected(idx, &self.selected_client_psk);
+            if (idx >= active_offers.len) return error.IllegalParameter;
+            if (self.client_offer_lease.active) self.client_offer_lease.finishPins(.{ .selected = idx });
+            active_offers.takeSelected(idx, &self.selected_client_psk);
+            self.client_offer_lease = .{};
             self.selected_client_psk_present = true;
```
Jaruso, line 1856: **Blocking test gap — this is the critical cache-consumption callback, but none of the backend tests exercise it with a real `ClientOfferLease`.** The current `selected_identity`, invalid-index, failure-cleanup, and filtering tests still construct `ClientPskOfferSet` and call the legacy `setClientPskOffers()`, which leaves `client_offer_lease.active == false`. They therefore verify the old raw-offer behavior but never execute this `finishPins(.selected)` path or prove that cache pins are released/consumed correctly.

#487 explicitly requires these cross-module cases. Please add cache-backed tests that obtain the lease from `ClientSessionCache.lookupOffers()` and transfer it with `setClientPskOfferLease()`.

A representative selected-index test should look roughly like:

```zig
var cache = try session_cache.ClientSessionCache.init(
    std.testing.allocator,
    session_cache.Limits.client_default,
);
defer cache.deinit();

// Store two distinct .single_use tickets whose PSKs can be distinguished.
try std.testing.expectEqual(.stored, cache.storeClone(&ticket_a, 0, .single_use));
try std.testing.expectEqual(.stored, cache.storeClone(&ticket_b, 1, .single_use));

var lookup = cache.lookupOffers(candidate, 2);
defer lookup.deinit();
try std.testing.expect(lookup == .hit);
try client.setClientPskOfferLease(&lookup.hit, &clock_dummy, Clock.now);

try client.backend().start(.client, {}, &sink);
const hello = try buildServerHello(&buf, .{ .selected_identity = 1 });
try client.backend().receive(.initial, hello, &sink);

// Exactly the selected cache entry was consumed; the other was released.
try std.testing.expectEqual(@as(usize, 1), cache.count());
var after = cache.lookupOffers(candidate, 3);
defer after.deinit();
try std.testing.expect(after == .hit);
try std.testing.expectEqual(@as(usize, 1), after.hit.offers.len);
// assert the survivor is the originally-unselected ticket
```

Please also cover the same real-lease path for:

- an out-of-range `selected_identity` => fatal **and every pin released**;
- no `pre_shared_key` extension => `.not_selected` and every pin released;
- failure/teardown before definitive ServerHello => `.aborted` exactly once;
- final eligibility/size filtering where an earlier cached ticket is dropped and wire index `0` selects a later cached ticket. That last case is essential because `dropOffer()` must shift both `offers` and `tokens` in lockstep.

src/tls/session_cache_persistence.zig
```diff
@@ -398,7 +398,13 @@ pub fn saveClientCache(
     backend: Backend,
     now_unix_ms: i64,
 ) SaveResult {
-    var snapshot = cache.cloneLiveForPersistence(cache.allocator, now_unix_ms) catch return .allocation_failed;
+    const token = cache.beginPersistenceOperation() catch return .cache_busy;
```
Jaruso, line 401: **Blocking test gap — the client persistence exclusion is implemented here, but #487’s required no-resurrection guarantees are not tested for the client cache.** The strong persistence regressions in this file currently exercise `StatefulServerCache`; the client round-trip still uses only `.reusable` tickets.

Please add client equivalents proving the new guard and generation behavior end-to-end. At minimum:

```zig
var cache = try session_cache.ClientSessionCache.init(
    std.testing.allocator,
    session_cache.Limits.client_default,
);
defer cache.deinit();

var ticket = try testClient(std.testing.allocator, "single", "example.test");
defer ticket.deinit();
try std.testing.expectEqual(.stored, cache.storeClone(&ticket, 0, .single_use));

var lease = cache.lookupOffers(testCandidate("example.test"), 1);
defer lease.deinit();
try std.testing.expect(lease == .hit);

var backend = MemoryBackend{ .allocator = std.testing.allocator };
defer backend.deinit();

try std.testing.expectEqual(
    SaveResult.cache_busy,
    saveClientCache(&cache, session.Limits.default, passthrough_protector, backend.interface(), 1),
);
try std.testing.expectEqual(
    LoadResult.cache_busy,
    loadClientCache(&cache, session.Limits.default, passthrough_protector, backend.interface(), 1),
);

lease.hit.finish(.{ .selected = 0 });
try std.testing.expectEqual(@as(usize, 0), cache.count());

// A save completed after consumption must not persist/resurrect the ticket.
try std.testing.expectEqual(
    SaveResult.saved,
    saveClientCache(&cache, session.Limits.default, passthrough_protector, backend.interface(), 2),
);
```

Then reload that snapshot into a fresh cache and assert the ticket is absent. Please also add the issue-mandated stale-generation test (a lease/token captured before an adopted replacement must be harmless) and an allocation-failure case showing indexes/accounting/pins/ticket ownership remain unchanged. Those are important here because the persistence code is the mechanism preventing a consumed single-use ticket from becoming replayable after restart.

Jaruso (commented):
## Merge recommendation: changes required

I traced PR #489 against every acceptance criterion in #487, including cache mutation/eviction, lease ABA protection, final ClientHello filtering, ServerHello outcome handling, teardown, persistence exclusion, restore/adoption, secret cleanup, and the existing tests.

The core implementation is directionally solid: the `(cache_generation, entry_id, lease_epoch)` token is ABA-safe; lookup pins only after all clones succeed; pinned entries are excluded from eviction/replacement; `dropOffer()` shifts the token and ticket together so wire indices stay aligned; invalid `selected_identity` reaches the `onServerHello` `errdefer` and abort-releases the lease; and the persistence token closes the save/load-vs-new-lease race. CI run #888 is green.

I would **not merge this yet** because #487 is not complete:

1. **The runtime/config rejection is still present.** `resumption_runtime.Config.validate()` still returns `InvalidConfig` for `.single_use`, and its test still requires that rejection. #487 explicitly says to remove that temporary rejection once this lifecycle is wired, so the feature remains impossible to enable through the process-shared runtime.
2. **The required cache-backed backend regression coverage is missing.** The existing `selected_identity`, invalid-index, failure-cleanup, and filtering tests still use the legacy `setClientPskOffers()` raw set. They do not exercise `setClientPskOfferLease()`, its finish callback, pin release, or selected cache consumption. This leaves the most security-sensitive cross-module behavior unverified.
3. **The required client persistence regressions are missing.** The new client persistence guard is structurally analogous to the server guard, but the strong no-resurrection/overlap tests currently exist only for `StatefulServerCache`. #487 explicitly requires client save/load no-resurrection, stale-generation harmlessness, and allocation-failure lease/accounting safety.

Please address the inline comments and add the missing #487 regressions, including cache-backed invalid-index release, final-filter index mapping, client persistence no-resurrection/overlap, stale generation completion, allocation failure with lease state unchanged, and consumed/removed secret zeroization. Once those are covered, the implementation should be in much better shape for merge.

src/tls/session_cache.zig
```diff
@@ -666,13 +661,21 @@ pub const ClientSessionCache = struct {
                     i += 1;
                     continue;
                 }
-                if (e.ticket.common.isExpired(now_unix_ms)) {
+                if (e.ticket.common.isExpired(now_unix_ms) and e.active_lease_epoch == null) {
```
Jaruso, line 666: **Blocking — `lookupOffers()` still mutates cache state before a fallible clone, violating #487's allocation-failure atomicity requirement.**

This branch removes and destroys an expired entry immediately. A later candidate is cloned using `self.allocator`; if that clone fails, the method returns `.storage_failed`, but the expired entry and its accounting are already gone.

A concrete failure sequence is:

```text
cache = [expired entry, live single-use entry]
lookupOffers(...)
  -> removes expired entry here
  -> cloneInto(live entry) => OutOfMemory
  -> returns .storage_failed
```

The caller was told lookup storage failed, yet `entries` and `total_bytes` changed. #487 explicitly requires allocation-failure paths to leave indexes, accounting, lease state, and ticket ownership unchanged.

Please stage expiry cleanup until after all selected clones succeed. One bounded approach is to collect expired stable IDs during the scan, then commit removals immediately before the already-infallible LRU/pin phase:

```zig
var expired_ids: [hard_max_entries_per_origin]u64 = undefined;
var expired_count: usize = 0;

// Scan only: do not mutate.
if (e.ticket.common.isExpired(now_unix_ms) and e.active_lease_epoch == null) {
    expired_ids[expired_count] = e.entry_id;
    expired_count += 1;
    had_expired_removal = true;
    i += 1;
    continue;
}

// clone selected candidates first...
if (storage_failed) {
    // no cache mutation has occurred
    ...
}

// After every clone succeeds, commit staged expiry removals.
for (expired_ids[0..expired_count]) |entry_id| {
    const idx = self.findIndexByEntryId(entry_id) orelse continue;
    const expired = self.entries.swapRemove(idx);
    self.total_bytes -= expired.bytes;
    destroyClientEntry(self.allocator, expired);
}
```

The exact implementation can differ, but the important property is **no destructive mutation before the last fallible allocation**.

Please add a `FailingAllocator` regression that stores an expired candidate plus a live candidate, snapshots `count`, `total_bytes`, ordering counters and ticket identities, forces `cloneInto` to fail, expects `.storage_failed`, and verifies every snapshot value plus lease state is unchanged.

Jaruso (commented):
## Merge recommendation: one change still required

The consumed-ticket zeroization requirement is now covered directly, and I re-checked the previously raised runtime, backend lease-completion, filtering, invalid-index, abort/no-selection, persistence no-resurrection, and stale-generation paths. Those fixes still look correct.

One #487 acceptance-criteria violation remains in `ClientSessionCache.lookupOffers()`: expired entries are removed before the fallible offer-cloning phase. If cloning a later live candidate returns OOM, the method returns `.storage_failed`, but the expired entry has already been destroyed and `total_bytes`/`entries` have changed. #487 explicitly requires allocation-failure paths to leave indexes, accounting, lease state, and ticket ownership unchanged.

Please make expiry cleanup transactional with respect to lookup cloning and add a failing-allocator regression that snapshots cache count/bytes/counters/entry identities, forces clone failure with an expired entry present, then proves the entire snapshot is unchanged. After that fix, I believe #487 is fully satisfied.

CI run #890 is currently still in progress, so this review does not yet treat CI as final.

Jaruso (commented):
## Merge recommendation: ready once CI completes

I re-reviewed PR #489 at `a12807c` against the full #487 design and required-test checklist, including the prior blockers and the newly changed lookup path.

The final functional blocker is resolved. `lookupOffers()` now stages expired entry IDs without mutating the cache, performs every fallible ticket clone first, and only commits expiry cleanup/LRU refresh/pinning after cloning succeeds. On `.storage_failed`, count, bytes, ordering counters, lease epoch/generation, entry IDs, LRU state, pin state, and ticket ownership remain unchanged. The new `FailingAllocator` regression explicitly exercises the mixed expired+live case that previously broke atomicity.

The previous zeroization gap is also closed with direct coverage showing selected single-use consumption removes the entry and wipes its inline PSK and nonce from allocator backing storage.

Across the complete PR, #487's required semantics are now covered: single-use store/restore, pinning and concurrent exclusion, selected-only consumption, release of unselected/no-PSK/aborted offers, fatal invalid-index cleanup, reusable behavior, eviction exclusion, final ClientHello index mapping, persistence no-resurrection/overlap, stale-generation safety, allocation-failure atomicity, runtime `.single_use` enablement, and consumed secret zeroization.

I do not see any remaining functional or security blocker in the reviewed code. The only cleanup I still notice is stale documentation in `ClientSessionCache.deinit()` / `planClientInsertionLocked()` that still describes client returns as clone-only / says there is no client pinning concept. That is non-blocking, but worth correcting because the implementation now has live `ClientOfferLease` callbacks and pins.

CI run #891 is still in progress at the time of this review; the relevant completed build, unit/integration, TLS audit, and appliance jobs are green. Assuming the overall run finishes green, I consider this PR ready to merge and #487 complete.

Jaruso (commented):
## Final re-review: merge-ready

I re-checked the current PR state at `a12807c`. The head has not changed since the prior full review, CI run #891 has completed successfully, and the previously identified #487 blockers remain addressed in the current code.

The transactional `lookupOffers()` fix preserves cache state on clone allocation failure; the consumed single-use zeroization regression is present; runtime `.single_use` enablement, cache-backed backend outcome tests, persistence no-resurrection/stale-generation coverage, final filtering index mapping, abort/no-selection/invalid-index release, and selected-only consumption are all still in place.

I do not see any remaining functional, security, or acceptance-criteria blocker. PR #489 is ready to merge and should close #487.

The only residual item is non-blocking stale documentation describing client cache returns as clone-only / claiming no client pinning concept. That can be cleaned up separately and does not affect correctness.

Note: the four older inline threads remain unresolved in GitHub even though their requested changes have been implemented. I have not resolved them automatically.

PR #495, opened 2026-07-22
tls: finish native 1-RTT resumption runtime integration

Jaruso:
## Summary

Completes the production integration work intentionally left after PR #486 (per the issue): the process-shared native TLS 1.3 resumption runtime (`tls.resumption_runtime.Runtime`) is now actually used end to end by native TLS-over-TCP and native QUIC/HTTP/3, rather than existing only as unused foundation.

- **Two-phase server ticket issuance** (`src/tls/tls13_backend.zig`, `src/tls/new_session_ticket.zig`, `src/tls/resumption_runtime.zig`): `prepareNewSessionTicket` → `Runtime.createIdentity` → `emitPreparedNewSessionTicket`, so the opaque bearer identity is derived from the exact prepared `ServerRecoverableState` instead of being invented by the caller. Stateful/stateless/hybrid dispatch, hybrid fallback on ordinary capacity refusal, and rollback (`rollbackIdentity`, backed by a new `StatefulServerCache.revokeHandle`) if emission fails after storage succeeded. The old one-phase `emitNewSessionTicket` is kept as a thin wrapper for existing callers.
- **Native TLS-over-TCP wiring** (`src/http/native_tls_connection.zig`): `Options.resumption_runtime` installs the shared server resolver before handshake start and triggers ticket issuance exactly once after application data opens, queued through the existing application-epoch record path (`PureZigRecordStream.applyEvent`) — no raw bytes written to the socket, no duplicate queue-limit logic.
- **Native QUIC/HTTP3 wiring** (`src/quic/tls_handshake.zig`, `src/quic/tls_backend.zig`, `src/quic/connection.zig`, `src/http/http3_runtime.zig`): thin forwarding seams (`setServerPskResolver`/`prepareNewSessionTicket`/`emitPreparedNewSessionTicket`) mirroring the TCP contract; `http3_runtime.Runtime` installs the same resolver per accepted connection and issues a ticket once established, through the existing application-CRYPTO reservation path. `enable_0rtt` stays false; no QUIC 0-RTT keys are touched.
- **Process composition root** (`src/edge_gateway.zig`, `src/gateway_state.zig`): one `resumption_runtime.Runtime` constructed per process when enabled, shared by every native TCP connection (via `WorkerContext`) and by `http3_runtime.Runtime`, torn down after its borrowers (defer ordering fires it last). When TCP uses OpenSSL, the runtime is only ever handed to HTTP/3 — no OpenSSL/native ticket interchange.
- **Native config boundary** (`src/edge_config.zig`): `TARDIGRADE_TLS_NATIVE_RESUMPTION_MODE` / `..._TICKET_LIFETIME_SECONDS` / `..._TICKET_USAGE`, distinct from the OpenSSL-facing `tls_session_*` fields, validated deterministically (invalid values fail `validate()`) and defaulting to disabled/safe.
- **Metrics** (`src/http/metrics.zig`): bounded, closed-enum counters — `tls_resumption_attempt_total`, `tls_resumption_outcome_total`, `tls_ticket_issue_total`, `tls_ticket_store_total`, `tls_ticket_resolve_total` — no secret or high-cardinality labels.
- **Bug fix surfaced by the new integration tests**: a PSK-resumed handshake never sends a Certificate message (by design), but both transports' completion policy (`encrypted_stream.zig`'s `completionPolicyError`, `tls_handshake.zig`'s `Handshake.complete`) unconditionally required a valid client-side peer certificate, which would have made every resumed connection fail closed with `CertificateInvalid` in production. Fixed by having the engine emit a synthetic `.certificate = .valid` event once the PSK binder confirms selection, since the client is inheriting trust already established at the original full handshake that issued the ticket.

## Scope notes

- Client-side native resumption wiring targets the codebase's current in-process test/deterministic-client usage; there is no production native TLS *client* connection type in this repo to wire (the issue explicitly says not to add one for this story).
- Metrics counters/schema are complete and tested, but not yet wired into every production call site (native TCP/QUIC issuance, client-side store) — a natural, low-risk follow-up.

## Test plan

- [x] `zig build test-session-cache --summary all` — 436/436
- [x] `zig build test-tls --summary all --error-style verbose` — 670/670 (includes new end-to-end resumed record-mode handshake test)
- [x] `zig build test-quic --summary all --error-style verbose` — 475/475 (includes new end-to-end resumed QUIC handshake test)
- [x] `zig build test-integration --summary all --error-style verbose`
- [x] `zig build test --summary all --error-style verbose` — 2482/2483 (1 known sandbox-only skip)
- [x] `zig build test -Dtls-profile=appliance --summary all --error-style verbose` — 2453/2454
- [x] `zig build -Doptimize=ReleaseFast --summary all --error-style verbose`
- [x] New focused integration tests prove genuine abbreviated handshakes (no Certificate flight, `psk_authenticated`, application data flows) for both record and QUIC, driven through `resumption_runtime.Runtime`'s cache/resolver composition rather than hand-rolled test stand-ins.

Closes #488

### DIFF
--- a/src/edge_gateway.zig
+++ b/src/edge_gateway.zig
@@ -253,6 +253,33 @@ pub fn run(cfg: *const edge_config.EdgeConfig) !void {
     var http3_runtime: ?http.http3_runtime.Runtime = null;
     var tls_terminator: ?http.tls_termination.TlsTerminator = null;
     var native_credentials: ?http.native_tls_connection.NativeCredentialStore = null;
+    // #488: one process-scoped native resumption runtime, shared by every
+    // native TCP connection and by the QUIC/H3 runtime — never by the
+    // OpenSSL terminator, which owns its own independent session cache
+    // (`tls_terminator`/`tls_session_*` above). Declared and deferred here,
+    // ahead of every borrower below, so its teardown runs last: after
+    // `http3_runtime`, the active/parked connection registries, and every
+    // dynamically accepted native connection have already torn down.
+    var native_resumption_entropy: tls_core.production_crypto.OsEntropy = .{};
+    var native_resumption_provider: tls_core.production_crypto.Provider = undefined;
+    var native_resumption_runtime: ?tls_core.resumption_runtime.Runtime = null;
+    if (edge_config.nativeResumptionMode(cfg) != .disabled) {
+        native_resumption_provider = tls_core.production_crypto.Provider.init(native_resumption_entropy.entropy());
+        native_resumption_runtime = tls_core.resumption_runtime.Runtime.init(
+            state_allocator,
+            .{
+                .mode = edge_config.nativeResumptionMode(cfg),
+                .ticket_lifetime_seconds = cfg.tls_native_resumption_ticket_lifetime_seconds,
+                .usage = edge_config.nativeResumptionTicketUsage(cfg),
+            },
+            .{ .ctx = undefined, .nowUnixMsFn = systemNowUnixMs },
+            native_resumption_provider.cryptoProvider(),
+        ) catch |err| blk: {
+            state.logger.warn(null, "native TLS/QUIC resumption runtime failed to initialize: {s}; continuing without native resumption", .{@errorName(err)});
+            break :blk null;
+        };
+    }
+    defer if (native_resumption_runtime) |*rt| rt.deinit();
     var http3_dispatch_ctx = ghandlers.Http3DispatchContext{
         .config_store = &config_store,
         .cfg = cfg,
@@ -353,6 +380,7 @@ pub fn run(cfg: *const edge_config.EdgeConfig) !void {
             .listen_host = cfg.listen_host,
             .quic_port = cfg.quic_port,
             .credential_provider = h3_credential_provider,
+            .resumption_runtime = if (native_resumption_runtime) |*rt| rt else null,
             .tls_min_version = "1.3",
             .tls_max_version = "1.3",
             .enable_0rtt = cfg.http3_enable_0rtt,
@@ -390,6 +418,7 @@ pub fn run(cfg: *const edge_config.EdgeConfig) !void {
         .tls = if (tls_terminator) |*tls| tls else null,
         .native_credentials = if (native_credentials) |*store| store else null,
         .native_tls_provider = native_tls_provider,
+        .resumption_runtime = if (native_resumption_runtime) |*rt| rt else null,
         .session_pool = undefined,
         .event_loop = &event_loop,
         .active = undefined,
@@ -1050,7 +1079,7 @@ fn startNewConnection(ctx: *WorkerContext, client_fd: std.posix.fd_t) void {
             client_fd,
             tls_protocol_policy,
             native_provider,
-            .{ .buffer_limits = cfg.tls_buffer_limits },
+            .{ .buffer_limits = cfg.tls_buffer_limits, .resumption_runtime = ctx.resumption_runtime },
         ) catch |err| {
             ctx.state.logger.warn(null, "native tls connection setup failed: {}", .{err});
             return;
@@ -1268,6 +1297,13 @@ fn activeConnectionCloseHook(raw_state: *anyopaque, fd: std.posix.fd_t) void {
 
 fn deadlineFromNow(now_ms: u64, timeout_ms: u32) u64 {
     return if (timeout_ms == 0) 0 else now_ms + @as(u64, timeout_ms);
+}
+
+/// Wall-clock source for the process-scoped native resumption runtime
+/// (#488): ticket lifetime/age bookkeeping needs real wall-clock time, not
+/// the monotonic clock the event loop otherwise uses.
+fn systemNowUnixMs(_: *anyopaque) i64 {
+    return compat.milliTimestamp();
 }
 
 fn reapActiveConnections(
--- a/src/edge_gateway.zig
+++ b/src/edge_gateway.zig
@@ -274,9 +274,9 @@ pub fn run(cfg: *const edge_config.EdgeConfig) !void {
             },
             .{ .ctx = undefined, .nowUnixMsFn = systemNowUnixMs },
             native_resumption_provider.cryptoProvider(),
-        ) catch |err| blk: {
-            state.logger.warn(null, "native TLS/QUIC resumption runtime failed to initialize: {s}; continuing without native resumption", .{@errorName(err)});
-            break :blk null;
+        ) catch |err| {
+            state.logger.err(null, "native TLS/QUIC resumption initialization failed ({s}); refusing to start", .{@errorName(err)});
+            return err;
         };
         if (native_resumption_runtime) |*rt| rt.setObserver(nativeResumptionMetricsObserver(&state));
     }
@@ -818,6 +818,8 @@ fn nativeResumptionOutcomeValue(outcome: tls_core.resumption_runtime.ResumptionO
     return switch (outcome) {
         .accepted => .accepted,
         .full_handshake => .full_handshake,
+        .incompatible => .incompatible,
+        .miss => .miss,
         .fatal => .fatal,
     };
 }
--- a/src/edge_gateway.zig
+++ b/src/edge_gateway.zig
@@ -278,6 +278,7 @@ pub fn run(cfg: *const edge_config.EdgeConfig) !void {
             state.logger.warn(null, "native TLS/QUIC resumption runtime failed to initialize: {s}; continuing without native resumption", .{@errorName(err)});
             break :blk null;
         };
+        if (native_resumption_runtime) |*rt| rt.setObserver(nativeResumptionMetricsObserver(&state));
     }
     defer if (native_resumption_runtime) |*rt| rt.deinit();
     var http3_dispatch_ctx = ghandlers.Http3DispatchContext{
@@ -777,6 +778,83 @@ pub fn run(cfg: *const edge_config.EdgeConfig) !void {
         state.logger.warn(null, "drain timeout elapsed; force-closed {d} queued connection(s)", .{drain_result.forced_closes});
     }
     state.logger.info(null, "Graceful shutdown complete (forced_closes={d} drain_timed_out={})", .{ drain_result.forced_closes, drain_result.timed_out });
+}
+
+fn nativeResumptionMetricsObserver(state: *GatewayState) tls_core.resumption_runtime.Observer {
+    return .{
+        .ctx = state,
+        .onTicketIssueFn = nativeResumptionTicketIssue,
+        .onTicketStoreFn = nativeResumptionTicketStore,
+        .onTicketResolveFn = nativeResumptionTicketResolve,
+        .onResumptionAttemptFn = nativeResumptionAttempt,
+        .onResumptionOutcomeFn = nativeResumptionOutcome,
+    };
+}
+
+fn nativeResumptionTransport(transport: tls_core.resumption_runtime.Transport) http.metrics.ResumptionTransport {
+    return switch (transport) {
+        .record => .record,
+        .quic => .quic,
+    };
+}
+
+fn nativeResumptionMode(mode: tls_core.resumption_runtime.Mode) http.metrics.ResumptionMode {
+    return switch (mode) {
+        .disabled, .stateful => .stateful,
+        .stateless => .stateless,
+        .hybrid => .hybrid,
+    };
+}
+
+fn nativeTicketResult(result: tls_core.resumption_runtime.TicketResult) http.metrics.TicketResult {
+    return switch (result) {
+        .success => .success,
+        .rejected => .rejected,
+        .failed => .failed,
+    };
+}
+
+fn nativeResumptionOutcomeValue(outcome: tls_core.resumption_runtime.ResumptionOutcome) http.metrics.ResumptionOutcome {
+    return switch (outcome) {
+        .accepted => .accepted,
+        .full_handshake => .full_handshake,
+        .fatal => .fatal,
+    };
+}
+
+fn nativeResumptionTicketIssue(ctx: *anyopaque, transport: tls_core.resumption_runtime.Transport, mode: tls_core.resumption_runtime.Mode, result: tls_core.resumption_runtime.TicketResult) void {
+    const state: *GatewayState = @ptrCast(@alignCast(ctx));
+    state.metrics_mutex.lock();
+    defer state.metrics_mutex.unlock();
+    state.metrics.recordTicketIssue(nativeResumptionTransport(transport), nativeResumptionMode(mode), nativeTicketResult(result));
+}
+
+fn nativeResumptionTicketStore(ctx: *anyopaque, result: tls_core.resumption_runtime.TicketResult) void {
+    const state: *GatewayState = @ptrCast(@alignCast(ctx));
+    state.metrics_mutex.lock();
+    defer state.metrics_mutex.unlock();
+    state.metrics.recordTicketStore(nativeTicketResult(result));
+}
+
+fn nativeResumptionTicketResolve(ctx: *anyopaque, mode: tls_core.resumption_runtime.Mode, result: tls_core.resumption_runtime.TicketResult) void {
+    const state: *GatewayState = @ptrCast(@alignCast(ctx));
+    state.metrics_mutex.lock();
+    defer state.metrics_mutex.unlock();
+    state.metrics.recordTicketResolve(nativeResumptionMode(mode), nativeTicketResult(result));
+}
+
+fn nativeResumptionAttempt(ctx: *anyopaque, transport: tls_core.resumption_runtime.Transport) void {
+    const state: *GatewayState = @ptrCast(@alignCast(ctx));
+    state.metrics_mutex.lock();
+    defer state.metrics_mutex.unlock();
+    state.metrics.recordResumptionAttempt(nativeResumptionTransport(transport));
+}
+
+fn nativeResumptionOutcome(ctx: *anyopaque, transport: tls_core.resumption_runtime.Transport, outcome: tls_core.resumption_runtime.ResumptionOutcome) void {
+    const state: *GatewayState = @ptrCast(@alignCast(ctx));
+    state.metrics_mutex.lock();
+    defer state.metrics_mutex.unlock();
+    state.metrics.recordResumptionOutcome(nativeResumptionTransport(transport), nativeResumptionOutcomeValue(outcome));
 }
 
 fn applyRuntimeIdentity(cfg: *const edge_config.EdgeConfig, logger: *const http.logger.Logger) !void {
--- a/src/http/metrics.zig
+++ b/src/http/metrics.zig
@@ -17,6 +17,20 @@ const tls_direction_count = 2;
 const TlsQueue = enum { inbound_ciphertext, inbound_plaintext, outbound_ciphertext, handshake };
 const TlsDirection = enum { carrier_read, plaintext_write };
 
+/// #488: native TLS/QUIC resumption metric labels. Every label here is a
+/// fixed, closed enum — never a raw ticket, PSK, identity, handle, SNI/ALPN
+/// value, certificate byte, key ID, ciphertext, decrypted state, filesystem
+/// path, or arbitrary error string.
+pub const ResumptionTransport = enum { record, quic };
+pub const ResumptionOutcome = enum { accepted, full_handshake, incompatible, miss, fatal };
+pub const ResumptionMode = enum { stateful, stateless, hybrid };
+pub const TicketResult = enum { success, rejected, failed };
+
+const resumption_transport_count = 2;
+const resumption_outcome_count = 5;
+const resumption_mode_count = 3;
+const ticket_result_count = 3;
+
 /// Server-wide metrics counters.
 ///
 /// Tracks request counts, status code distribution, and uptime.
@@ -145,6 +159,16 @@ pub const Metrics = struct {
     /// Total config hot-reloads rejected (load/validation/bookkeeping failure); the
     /// previous config stays active.
     reload_failure_total: u64,
+    /// #488: native TLS/QUIC resumption attempts by transport.
+    tls_resumption_attempt_total: [resumption_transport_count]u64,
+    /// #488: native resumption outcomes by transport and outcome.
+    tls_resumption_outcome_total: [resumption_transport_count][resumption_outcome_count]u64,
+    /// #488: server-side ticket issuance attempts by transport, mode, result.
+    tls_ticket_issue_total: [resumption_transport_count][resumption_mode_count][ticket_result_count]u64,
+    /// #488: client-side ticket storage attempts by result.
+    tls_ticket_store_total: [ticket_result_count]u64,
+    /// #488: server-side ticket resolution attempts by mode and result.
+    tls_ticket_resolve_total: [resumption_mode_count][ticket_result_count]u64,
     /// Total graceful-shutdown drains started.
     drain_total: u64,
     /// Total drains that hit the configured drain timeout before work finished.
@@ -251,6 +275,11 @@ pub const Metrics = struct {
             .reload_attempts_total = 0,
             .reload_success_total = 0,
             .reload_failure_total = 0,
+            .tls_resumption_attempt_total = .{0} ** resumption_transport_count,
+            .tls_resumption_outcome_total = .{.{0} ** resumption_outcome_count} ** resumption_transport_count,
+            .tls_ticket_issue_total = .{.{.{0} ** ticket_result_count} ** resumption_mode_count} ** resumption_transport_count,
+            .tls_ticket_store_total = .{0} ** ticket_result_count,
+            .tls_ticket_resolve_total = .{.{0} ** ticket_result_count} ** resumption_mode_count,
             .drain_total = 0,
             .drain_timeouts_total = 0,
             .drain_forced_closes_total = 0,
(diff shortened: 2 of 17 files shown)


src/tls/resumption_runtime.zig
```diff
@@ -128,6 +128,118 @@ pub const Runtime = struct {
...
+        self: *Runtime,
+        state: *session.ServerRecoverableState,
+        now_unix_ms: i64,
+    ) CreateIdentityError!Identity {
+        const cache = if (self.server_cache) |*cache| cache else return error.StorageUnavailable;
+        var handle: [session_cache.stateful_identity_len]u8 = undefined;
+        const result = cache.insertMove(state, now_unix_ms, usagePolicy(self.config.usage), &handle);
+        return switch (result) {
+            .stored => .{ .stateful = handle },
+            .rejected_capacity, .storage_failed, .rejected_handle_generation_failed => error.StorageUnavailable,
```
Jaruso: **Blocking — hybrid fallback is currently broader than #488 permits.** `StatefulServerCache.insertMove()` distinguishes ordinary capacity refusal from hard failures: `.storage_failed` is produced by allocation/storage failures in the insertion plan, and `.rejected_handle_generation_failed` is produced when fresh-handle generation fails. Collapsing all three into `StorageUnavailable` means the `.hybrid` branch above silently falls back to stateless on OOM/internal storage failure or RNG failure, even though the issue requires fallback **only for explicitly typed ordinary stateful refusal cases**.

Please preserve the distinction all the way through `createIdentity`, e.g.:

```zig
pub const CreateIdentityError = error{
    StatefulCapacityRefused,
    StatefulStorageFailed,
    HandleGenerationFailed,
    SealFailed,
};

fn createStatefulIdentity(...) CreateIdentityError!Identity {
    ...
    return switch (result) {
        .stored => .{ .stateful = handle },
        .rejected_capacity => error.StatefulCapacityRefused,
        .storage_failed => error.StatefulStorageFailed,
        .rejected_handle_generation_failed => error.HandleGenerationFailed,
        .replaced, .rejected_unsupported_usage => unreachable,
    };
}

// Hybrid only degrades on the explicitly ordinary refusal.
.hybrid => self.createStatefulIdentity(state, now_unix_ms) catch |err| switch (err) {
    error.StatefulCapacityRefused => self.createStatelessIdentity(state, now_unix_ms, scratch),
    else => return err,
},
```

Please also add tests with a failing allocator and a failing handle/random source proving those failures do **not** produce a stateless identity, while the existing capacity-refusal test still does.

src/http/native_tls_connection.zig
```diff
@@ -246,7 +265,62 @@ pub const NativeTlsConnection = struct {
...
+
+        var sink = tls.tls13_transport.EventSink{};
+        defer sink.deinit();
+        self.backend.emitPreparedNewSessionTicket(self.allocator, &sink, &prepared, identity.slice(), limits) catch |err| {
+            runtime.rollbackIdentity(&identity);
+            return err;
+        };
+
+        for (sink.items[0..sink.len]) |event| switch (event) {
+            .handshake_bytes => |hb| self.record.applyEvent(.{ .handshake_bytes = .{ .epoch = hb.epoch, .data = hb.data } }) catch |err| {
```
Jaruso: **Blocking — this does not actually make post-handshake issuance best-effort.** The catch here rolls the runtime identity back, but `PureZigRecordStream.applyEvent()` calls `emitHandshakeRecords(... ) catch |err| return self.fail(err)`. `fail()` permanently changes the established stream to `.failed`, clears its queues/keys, tears down the driver/bridge, and closes the carrier. So if the outbound queue cannot reserve this `NewSessionTicket`, the ticket error is swallowed by `maybeIssueSessionTicket`, but the usable connection has already been destroyed.

#488 explicitly requires a bounded post-handshake record path where queue/emission failure **does not tear down the established connection**. Please add a non-fatal application-epoch queue API instead of reusing the handshake-fatal `applyEvent` path. For example, the call site should look more like:

```zig
self.record.tryQueuePostHandshake(.{
    .handshake_bytes = .{ .epoch = hb.epoch, .data = hb.data },
}) catch |err| {
    runtime.rollbackIdentity(&identity);
    return err;
};
```

and the record-layer method should preflight/reserve the complete sealed batch and return `WouldBlock`/a bounded queue error **without calling `self.fail()`** when capacity is unavailable.

Please force the outbound queue to its limit in a test, attempt issuance, and prove all of the following after the failure: the stateful identity no longer resolves, `applicationDataOpen()` is still true, the carrier is still alive, and ordinary application data can still be exchanged.

src/http/native_tls_connection.zig
```diff
@@ -246,7 +265,62 @@ pub const NativeTlsConnection = struct {
...
+        var prepared = try self.backend.prepareNewSessionTicket(self.allocator, .{
+            .ticket_lifetime = runtime.config.ticket_lifetime_seconds,
+            .ticket_age_add = ticket_age_add,
+            .ticket_nonce = &ticket_nonce,
+            .issued_at_unix_ms = now_unix_ms,
+        }, limits);
+        defer prepared.deinit();
+
+        const scratch = try self.allocator.alloc(u8, runtime.maxIdentityLen());
+        defer self.allocator.free(scratch);
```
Jaruso: **Blocking — freeing this buffer is not sufficient for the zeroization requirement.** In stateless/hybrid mode `scratch` contains the `TDTK` bearer identity; in stateful mode the returned `Identity.stateful` contains the live bearer handle. #488 explicitly requires temporary ticket/identity/handle/scratch buffers to be wiped on every exit path. The same `defer allocator.free(scratch)` pattern exists in `http3_runtime.zig`.

I would make wiping an ownership property rather than rely on each transport remembering it. For example:

```zig
const scratch = try self.allocator.alloc(u8, runtime.maxIdentityLen());
defer {
    tls.crypto.secureZero(u8, scratch); // also covers createIdentity failure
    self.allocator.free(scratch);
}

var identity = try runtime.createIdentity(&prepared.state, now_unix_ms, scratch);
defer identity.deinit(); // wipes the inline stateful handle; stateless may wipe its used slice too
```

and in the runtime:

```zig
pub fn deinit(self: *Identity) void {
    switch (self.*) {
        .stateful => |*handle| crypto.secureZero(u8, handle),
        .stateless => |s| crypto.secureZero(u8, s.buf[0..s.len]),
    }
    self.* = undefined;
}
```

Also wipe the temporary local `handle` inside `createStatefulIdentity` after it is copied into the returned identity. Please cover both successful issuance and forced failure paths with a test that can observe memory before free (or an equivalent deterministic zeroization seam).

src/http/metrics.zig
```diff
@@ -304,6 +333,32 @@ pub const Metrics = struct {
         self.health_probe_runs += 1;
     }
 
+    /// #488: record one native TLS/QUIC resumption attempt (a connection
+    /// where a PSK-bearing ClientHello was offered/resolved).
+    pub fn recordResumptionAttempt(self: *Metrics, transport: ResumptionTransport) void {
```
Jaruso, line 338: **Blocking AC gap — these new methods are not driven by production resumption events.** I checked the entire PR diff: the only invocations of the new `recordResumption*` / `recordTicket*` APIs are the metrics tests themselves. Because these methods did not exist before this PR, the exported series will remain zero during real native TLS/QUIC resumption. The PR description also explicitly says production wiring is deferred, but #488 requires these outcomes to be observable in this story.

Please wire a typed observer/callback path from the runtime/cache/backend into the existing thread-safe gateway metrics boundary. Ticket issuance can be recorded directly around the typed operation, for example conceptually:

```zig
const result: TicketResult = blk: {
    issueSessionTicket(...) catch |err| {
        break :blk classifyTypedIssueError(err); // closed enum only
    };
    break :blk .success;
};
metrics.recordTicketIssue(transport, mode, result);
```

For `attempt`, `outcome`, `store`, and `resolve`, source the signal from the shared runtime/cache/backend typed results (or an observer they emit), **not** by reparsing `ServerHello` or reconstructing selected-index state in the adapters. Please add an integration assertion that a real full->ticket->resumed flow increments the expected series, including a failed/rejected issuance case.

src/tls/tls13_backend_tests.zig
```diff
@@ -603,7 +603,156 @@ test "PSK round trip: an offered, resolved, and verified ticket resumes the hand
...
+    try std.testing.expect(resumed.server_backend.core.psk_authenticated);
+    // No certificate flight: this is a genuine PSK-abbreviated handshake, not
+    // merely a second successful full handshake. `certificate_state` reports
+    // `.valid` anyway (#488): the client inherits trust from the original
+    // full handshake that issued this ticket, confirmed here by the binder,
+    // since every transport completion policy otherwise requires a valid
+    // peer certificate for the client role.
+    try std.testing.expectEqual(events.CertificateState.valid, resumed.observed.certificate_state.?);
 
     // The resumed connection is genuinely usable: application data flows
```
Jaruso, line 762: This is good TLS-level proof of a genuine abbreviated handshake, but it does **not** satisfy #488's record acceptance criterion as written: the AC requires the resumed native record connection to exchange **HTTP traffic**. Here `DirectHarness` seals arbitrary bytes (`"resumed request"`) directly at the TLS application-data layer; it never exercises `NativeTlsConnection`, HTTP parsing/dispatch, or the production post-handshake issuance path added in this PR.

Please keep this focused protocol test, but add an integration-layer reconnect test that performs the production sequence:

```text
full NativeTlsConnection handshake
 -> production server issuance
 -> client stores ticket via Runtime
 -> fresh connection transfers lookup lease before start
 -> assert abbreviated/psk_authenticated
 -> send a real HTTP request (e.g. GET /... HTTP/1.1)
 -> assert a valid HTTP response
```

That test should go through the actual connection/HTTP boundary so it also catches regressions in ticket delivery and queueing that this direct harness cannot see.

src/quic/connection.zig
```diff
@@ -2758,6 +2811,244 @@ test "driver: server NewSessionTicket uses application CRYPTO retransmission and
...
+
+    try resumed.pump();
+
+    try testing.expect(resumed.client.isEstablished());
+    try testing.expect(resumed.server.isEstablished());
+    try testing.expect(resumed.client_backend.engine.core.psk_authenticated);
+    try testing.expect(resumed.server_backend.engine.core.psk_authenticated);
+
+    // The resumed connection is genuinely usable: application data flows.
+    const id = try resumed.client.openStream(.bidi);
```
Jaruso, line 3059: Same acceptance-test gap on the QUIC side: opening a generic QUIC bidi stream and writing `"hello"` proves resumed QUIC application data, but #488 explicitly requires a resumed **QUIC/H3** connection to exchange **HTTP/3 traffic**. This test bypasses `http3_runtime` and the H3 framing/request layer entirely.

Please keep this QUIC transport test, and add an H3-level reconnect test that issues/stores the ticket, reconnects through the runtime-backed PSK path, then sends an actual HTTP/3 request (HEADERS/DATA via the existing H3 test/client helpers) and verifies the response while also asserting the underlying TLS backend is `psk_authenticated`. That is the coverage needed to close the literal H3 AC rather than inferring it from a raw QUIC stream.

Jaruso (commented):
**Blocking review — changes required before merge.**

I reviewed the full 14-file diff against #488, including the requirements inherited from #487/#489, the transport composition/lifetime rules, two-phase issuance transactionality, config boundary, metrics, and the resumed-traffic acceptance criteria. I also verified that PR #489 is merged at this PR's base and that the current CI run is green.

There is a lot here that is solid: the process-scoped runtime ownership/defer ordering looks correct; TCP and H3 share the same native runtime without crossing into OpenSSL; the QUIC TLS seams remain thin; the two-phase backend preparation derives the recoverable state once without touching the sink/transcript; stateful rollback is keyed by the exact handle; native config is distinct, disabled by default, validates the enabled lifetime ceiling, and accepts `single_use`; and this PR does not introduce 0-RTT keys/early application data.

I found five acceptance-criteria blockers:

1. **Hybrid fallback currently hides hard stateful failures.** `createStatefulIdentity` maps `.rejected_capacity`, `.storage_failed`, and `.rejected_handle_generation_failed` to the same `StorageUnavailable` error, and hybrid falls back on that error. #488 explicitly requires fallback only for the intended typed ordinary refusal cases. Allocation/storage failure and handle-generation/RNG failure must remain hard failures.

2. **TCP best-effort ticket issuance can still kill an established connection.** `NativeTlsConnection` catches the error from `record.applyEvent`, but `PureZigRecordStream.applyEvent` routes emission failures through `self.fail(...)`, which sets the stream to `.failed`, clears queues, tears down the bridge/driver, and closes the carrier. A full outbound queue therefore violates the AC that post-handshake ticket queue/emission failure must not tear down an otherwise usable connection. This needs the bounded, non-fatal post-handshake record queue seam called for by section C.

3. **Temporary bearer identities are not wiped.** Both TCP and H3 free the stateless scratch buffer without zeroing it, and the stateful `Identity` handle has no destruction/wipe path. #488 explicitly requires ticket/identity/handle/scratch buffers to be wiped on every exit path, including queue/fallback/allocation failures.

4. **The metrics AC is not implemented end to end.** `metrics.zig` defines and renders the required bounded series, but there are no production call sites driving the new counters; the PR description itself notes this wiring is deferred. As written, production resumption/issuance can occur while these counters remain zero. #488 requires issuance and resumption outcomes to be observable now, using typed runtime/cache/backend outcomes rather than adapter-side inference.

5. **The resumed-traffic tests stop below the required application layer.** The record test sends arbitrary TLS application bytes through `DirectHarness`, and the QUIC test writes `"hello"` to a generic QUIC bidi stream. Those are valuable protocol tests, but #488 explicitly requires a resumed native record connection to exchange **HTTP traffic** and a resumed QUIC connection to exchange **HTTP/3 traffic**. There is no changed integration test that exercises `NativeTlsConnection` + HTTP parsing/dispatch or the H3 request layer across the resumed reconnect.

The inline comments contain concrete implementation sketches and the failure tests I expect for each issue. In particular, please add forced allocator/RNG failure coverage for hybrid dispatch, outbound-queue-pressure coverage proving TCP remains open and the inserted handle is revoked, zeroization coverage, production metrics assertions, and actual HTTP/H3 reconnect tests.

Once these are addressed, the remaining pieces I inspected are consistent with #488's design and ownership constraints.

src/quic/connection.zig
```diff
@@ -2758,6 +2811,244 @@ test "driver: server NewSessionTicket uses application CRYPTO retransmission and
...
+    // Built from the retained ticket's own fields rather than guessed
+    // constants: QUIC connections carry their negotiated transport
+    // parameters as `transport_compat`, which is part of the origin the
+    // ticket was actually issued under (unlike the plain record profile).
+    const candidate: tls_core.session.CandidateContext = .{
+        .cipher_suite = capture.retained.common.cipher_suite,
+        .server_name = if (capture.retained.common.server_name) |*s| s.slice() else null,
+        .application_protocol = if (capture.retained.common.application_protocol) |*a| a.slice() else null,
+        .auth_binding = capture.retained.common.auth_binding,
+        .transport_compat = if (capture.retained.common.transport_compat) |*snap| .{
```
Jaruso: **Blocking — this test is encoding the opposite of #488's ordinary 1-RTT compatibility requirement.**

The reconnect candidate is populated with the *previous ticket's* `transport_compat` bytes so `lookupClientOffers()` will hit. That hides a production bug rather than proving the required behavior. #488 explicitly says to configure ordinary 1-RTT transport/application compatibility as **ignored where appropriate**, with 0-RTT remaining disabled.

Today the shared engine still exact-matches these snapshots on both sides:

- `resumptionContext()` stamps `peerTransportCompat()` into the issued ticket/state;
- server `selectPsk()` builds `CandidateContext.transport_compat = self.candidateTransportCompat()` and `session.evaluateCompatibility()` requires exact equality;
- client `onEncryptedExtensions()` exact-compares the selected ticket's stored transport snapshot against the newly received server transport extension *after ServerHello has already selected the PSK*.

QUIC transport parameters are connection-specific and may legitimately change across a resumed **1-RTT** connection. With the current code, a changed client transport parameter causes the server to reject resumption and fall back unnecessarily; a changed server parameter is worse: the server can select the PSK, then the client returns `IllegalParameter` from `onEncryptedExtensions()`, turning a perfectly valid reconnect into a fatal handshake because fallback is no longer possible after PSK selection.

Please introduce an explicit compatibility policy rather than feeding the old snapshot back as the current one. Conceptually:

```zig
pub const SnapshotResumePolicy = enum { exact, ignore };

pub const ResumeCompatibilityPolicy = struct {
    transport: SnapshotResumePolicy = .exact,
    application: SnapshotResumePolicy = .exact,
};
```

For #488's native record/QUIC **1-RTT-only** integration, configure both snapshot policies as `.ignore` before the backend starts. The policy then needs to apply consistently to all three places that currently enforce/store these values:

```zig
// Ticket state creation
.transport_compat = if (self.resume_compat.transport == .exact)
    self.peerTransportCompat()
else
    null,

// Server PSK compatibility
.transport_compat = if (self.resume_compat.transport == .exact)
    self.candidateTransportCompat()
else
    null,

// Client post-ServerHello validation
if (self.resume_compat.transport == .exact and
    !compatCompatible(stored.transport_compat, received_transport))
{
}
```

Do the equivalent for `application_compat`. Keep cipher suite, SNI, ALPN, authentication binding, ticket lifetime, and binder checks unchanged. A future 0-RTT path can opt back into `.exact`/appropriate remembered-parameter policy rather than making ordinary 1-RTT depend on stale connection settings.

Please change this test so it **does not copy the old snapshot into the candidate**, then deliberately issue the ticket under transport parameters A and reconnect with different valid client/server transport parameters B/C. The resumed connection should still reach `psk_authenticated` on both sides and exchange H3 traffic. That regression test is important because the existing version currently guarantees the faulty exact-match path passes.

Jaruso (commented):
**Second-pass blocking review — changes still required before merge.**

I re-reviewed the current head (`3d6c7c0`) against #488 rather than merely replaying the first review. The branch has not changed since that review, so the six existing unresolved threads remain valid and I have not duplicated them here. CI is green on this SHA, but the current tests still do not exercise several of the failure/compatibility conditions those threads cover.

This pass found one additional blocker: **ordinary QUIC 1-RTT resumption is incorrectly coupled to the previous connection's transport-compatibility snapshot.** #488 explicitly requires ordinary 1-RTT compatibility to ignore transport/application snapshots where appropriate while 0-RTT remains disabled. The current shared backend instead stores the peer QUIC transport extension in the ticket state and exact-matches it again on the next connection. The newly added QUIC resumption test masks that behavior by copying the ticket's old `transport_compat` snapshot back into the reconnect candidate.

That produces real production failure modes when QUIC transport parameters legitimately change between connections: the server can reject the ticket and fall back to a full handshake when the client's parameters differ, and if the server's parameters differ the server may already select the PSK in ServerHello before the client rejects the new EncryptedExtensions as `IllegalParameter`, which makes the reconnect fatal rather than a full-handshake fallback.

The new inline comment describes the expected policy seam and test. In particular, keep SNI/ALPN/cipher/auth-binding compatibility checks, but configure transport/application snapshots as ignored for this story's ordinary 1-RTT path; retain exact snapshot semantics for a future 0-RTT policy where remembered transport/application limits actually matter.

**Current verdict: do not merge.** There are now seven unresolved review findings: the original six plus this compatibility-policy blocker.

src/edge_gateway.zig
```diff
@@ -253,6 +253,34 @@ pub fn run(cfg: *const edge_config.EdgeConfig) !void {
...
+            state_allocator,
+            .{
+                .mode = edge_config.nativeResumptionMode(cfg),
+                .ticket_lifetime_seconds = cfg.tls_native_resumption_ticket_lifetime_seconds,
+                .usage = edge_config.nativeResumptionTicketUsage(cfg),
+            },
+            .{ .ctx = undefined, .nowUnixMsFn = systemNowUnixMs },
+            native_resumption_provider.cryptoProvider(),
+        ) catch |err| blk: {
+            state.logger.warn(null, "native TLS/QUIC resumption runtime failed to initialize: {s}; continuing without native resumption", .{@errorName(err)});
```
Jaruso: **Blocking — explicitly enabled native resumption must not silently downgrade to disabled.** Any `Runtime.init` failure here (for example stateless/hybrid entropy/provider failure, allocation failure, or another initialization fault) is converted into `null`, so the process continues serving while the operator's validated `stateful`/`stateless`/`hybrid` setting is simply ignored. That makes the configuration boundary nondeterministic and can leave later work such as #366 believing the prerequisite runtime exists when production actually disabled it.

Please propagate the typed initialization failure after logging instead of continuing:

```zig
native_resumption_runtime = tls_core.resumption_runtime.Runtime.init(
    state_allocator,
    runtime_config,
    .{ .ctx = undefined, .nowUnixMsFn = systemNowUnixMs },
    native_resumption_provider.cryptoProvider(),
) catch |err| {
    state.logger.err(null,
        "native TLS/QUIC resumption initialization failed ({s}); refusing to start",
        .{@errorName(err)},
    );
    return err;
};
```

`disabled` should remain the only configuration that intentionally starts without the runtime. Please add a startup/composition test proving an enabled mode cannot degrade to `null` on an injected init failure.

Jaruso (commented):
**Third-pass blocking review — current head is improved substantially, but #488 is not closed yet.**

I re-reviewed the current head (`9e54989`) against the full #488 acceptance criteria and the #487/#489 ownership contract. I did not simply carry forward the older findings: I verified that three prior blockers are genuinely fixed and resolved those stale threads — hybrid fallback now distinguishes hard failures, temporary identities/scratch are wiped, and TCP now uses a non-fatal post-handshake queue path.

The explicit shared-engine compatibility policy is also a real improvement, but its QUIC integration is currently applied at the wrong layer (inline comment below).

There are **five remaining blockers** before I would merge this and unblock #366:

1. **Explicitly enabled native resumption can silently disappear at startup.** `edge_gateway.run` catches any `Runtime.init` failure and continues with `null`, even when the operator selected stateful/stateless/hybrid. Entropy/provider/OOM/init failures must fail startup rather than silently running with resumption disabled. See inline comment.

2. **QUIC globally changes the shared compatibility default to `ignore`.** Every QUIC TLS backend constructor now overrides the shared engine's safe `.exact` default, even when no #488 runtime is configured. #488 requires the ordinary 1-RTT runtime path to opt into ignored transport/application snapshots; it must not globally weaken QUIC semantics, especially immediately before #366 adds 0-RTT remembered-parameter policy. See inline comment.

3. **Resumption metrics still only describe successful resumptions.** Both TCP and H3 call `resumptionAttempt`/`resumptionOutcome` only after `psk_authenticated == true`. A PSK-bearing attempt that misses, is incompatible, falls back to a full handshake, or fails fatally is never recorded. Consequently `full_handshake`, `incompatible`, `miss`, and `fatal` outcome series cannot reflect production behavior. The signal needs to originate from the shared PSK decision path, not adapter-side post-success inference. See inline comment.

4. **The record acceptance test still stops at TLS application bytes rather than HTTP.** The existing unresolved thread on `tls13_backend_tests.zig` remains valid. `DirectHarness` proves a genuine abbreviated TLS handshake, but #488 explicitly requires a resumed native record connection to exchange HTTP traffic through the production connection/HTTP boundary.

5. **The QUIC acceptance test still stops at a raw bidi stream rather than HTTP/3.** The existing unresolved thread on `connection.zig` remains valid. Writing `"hello"` to a QUIC stream proves 1-RTT transport usability, not resumed H3 request/response through the H3 framing/session layer.

I would also strengthen the final validation before closing #488: the full reconnect tests currently prove the stateful path, while the AC explicitly names stateful, stateless, and hybrid issuance/reconnect. Please exercise stateless and hybrid through an actual resumed handshake as well, and retain a forced post-handshake queue/emission failure test proving the established record connection survives while the exact inserted stateful handle is revoked.

**Verdict: do not merge yet.** Once these are addressed and CI is green on the resulting SHA, #488 should be in a position to close and #366 can begin on a sound 1-RTT baseline.

src/quic/tls_backend.zig
```diff
@@ -206,13 +206,15 @@ pub const Tls13Backend = struct {
...
     pub fn initClientWithOptions(entropy: Entropy, trust: Trust, options: ClientOptions) Tls13Backend {
-        return .{ .engine = shared.Tls13Backend.initClientConfigured(entropy, trust, .{
+        var self = Tls13Backend{ .engine = shared.Tls13Backend.initClientConfigured(entropy, trust, .{
             .policy = tls_core.policy.Policy.quicDefault(),
             .transport = .{ .extension = .{
                 .extension_type = ext_quic_transport_parameters,
                 .local = "",
             } },
         }, options) };
+        self.engine.setResumeCompatibilityPolicy(.{ .transport = .ignore, .application = .ignore }) catch unreachable;
```
Jaruso: **Blocking before #366 — this applies the #488 1-RTT compatibility exception globally to every QUIC backend.** The shared engine correctly defaults `ResumeCompatibilityPolicy` to `.exact`, but `initClient*` and both server constructors all force transport/application snapshots to `.ignore` even when no native resumption runtime is configured.

That is broader than #488's contract. Ordinary 1-RTT resumption should opt into ignored connection-specific QUIC/H3 snapshots; the low-level QUIC backend should retain the shared `.exact` default so another caller — especially #366's upcoming 0-RTT path — can require remembered transport/application compatibility instead of inheriting a silently weakened policy.

Please keep constructors at the shared default and expose a thin forwarding setter, then configure `.ignore` only at the #488 runtime composition boundary before `Connection.init`, e.g.:

```zig
pub fn setResumeCompatibilityPolicy(
    self: *Tls13Backend,
    policy: shared.Tls13Backend.ResumeCompatibilityPolicy,
) HandshakeError!void {
    self.engine.setResumeCompatibilityPolicy(policy) catch |err| return mapError(err);
}

// http3_runtime, only for ordinary native 1-RTT resumption:
try backend.setResumeCompatibilityPolicy(.{
    .transport = .ignore,
    .application = .ignore,
});
```

The reconnect regression should then deliberately use changed valid QUIC transport parameters and prove the **runtime-configured 1-RTT** connection still resumes, while a backend that never opted in retains `.exact` behavior.

src/http/native_tls_connection.zig
```diff
@@ -246,7 +267,84 @@ pub const NativeTlsConnection = struct {
...
+        self.maybeRecordResumptionOutcome();
+        self.maybeIssueSessionTicket();
+        return result;
+    }
+
+    fn maybeRecordResumptionOutcome(self: *NativeTlsConnection) void {
+        if (self.resumption_outcome_recorded) return;
+        const runtime = self.resumption_runtime orelse return;
+        if (!self.record.applicationDataOpen()) return;
+        if (!self.backend.core.psk_authenticated) return;
```
Jaruso: **Blocking observability gap — this only records successful PSK resumptions, not resumption attempts/outcomes.** The same pattern exists in `http3_runtime.recordResumptionOutcome`: when `psk_authenticated` is false, the function returns without emitting anything. That means a ClientHello carrying PSKs that results in resolver miss, incompatibility, ordinary full-handshake fallback, or a fatal PSK/binder error never increments `tls_resumption_attempt_total` and never reaches the `full_handshake`, `incompatible`, `miss`, or `fatal` outcome buckets.

Those labels therefore cannot represent production behavior even though the Prometheus schema exposes them. The signal needs to come from the shared backend's typed PSK decision lifecycle — where it actually knows whether identities were offered/resolved, whether compatibility rejected them, whether the handshake fell back, and whether selection became fatal — rather than being reconstructed only after transport success.

Conceptually, add a bounded observer/outcome seam around the shared PSK selection path and let the runtime map it to metrics:

```zig
pub const ResumptionDecision = enum {
    accepted,
    miss,
    incompatible,
    full_handshake,
    fatal,
};

// shared backend emits exactly once per attempted PSK handshake;
// runtime observer maps the closed enum to Gateway metrics.
```

Please cover at least accepted, resolver-miss/full-handshake fallback, incompatibility, and fatal binder/error paths in integration tests. TCP/QUIC adapters still must not reparse ServerHello or duplicate selected-index logic.

PR #497, opened 2026-07-23
tls: add 0-RTT client/server negotiation and early secret derivation

Jaruso:
## Summary

Issue #366 ("[326-G] add 0-RTT policy gate and request-safety filter") is
XL-complexity and its own description lays out 5 suggested PR slices. This
PR implements **slice 1 only** (TLS vocabulary + early secret + server
decision + EE signaling), scoped to `src/tls/tls13_backend.zig` and
`src/tls/key_schedule.zig` as the issue itself suggests for that slice.

- Adds `ClientEarlyDataIntent` / `ServerEarlyDataPolicy` / `EarlyDataDecision`
  / `EarlyDataReplayGate` vocabulary, all disabled/unavailable by default.
- Client: `setClientEarlyDataIntent`, `planEarlyDataAttempt` (attempts 0-RTT
  only against the *first surviving* PSK offer — never reordered, preserving
  #487's wire-index/lease semantics), emits the empty `early_data` ClientHello
  extension before `pre_shared_key`, and derives `client_early_traffic_secret`
  from the hash of the *complete* ClientHello (real binders, not the
  binder-truncated prefix).
- Server: `setServerEarlyDataPolicy`, `setEarlyDataReplayGate`, and a new
  `decideServerEarlyData` decision made live inside the existing `selectPsk`
  selection (identity-0 → ticket-capability → age-skew → anti-replay), after
  binder verification — so an early-data rejection never rejects an
  otherwise-valid PSK resumption. Anti-replay is only a seam (#368 owns the
  actual store) and defaults to `.unavailable`, which fails 0-RTT closed
  without affecting ordinary 1-RTT.
- `EncryptedExtensions` signals acceptance with an empty `early_data`
  extension; the client validates it (attempted, identity 0, non-resume-only
  ticket) before trusting it.
- New public accessors: `earlyDataAttempted()`, `earlyDataAccepted()`,
  `earlyDataDecision()`.

**Out of scope for this PR** (separate follow-up slices, per #366's own
breakdown): ticket `max_early_data_size` advertisement at the #495 issuance
call sites, record/QUIC 0-RTT key installation and EndOfEarlyData, and the
HTTP-level request-safety gate. This PR does not close #366.

## Test plan
- [x] `zig build test-tls` / `zig build test` — full suite green, including
      8 new tests: an accepted round trip with matching independently-derived
      client/server secrets, resume-only tickets never attempting 0-RTT,
      client opt-out, server-policy-disabled rejection, age-skew rejection,
      explicit replay rejection, default (unavailable) replay rejection,
      non-zero selected-identity rejection, and ClientHello wire-ordering
      (`early_data` before `pre_shared_key`).
- [x] `key_schedule.zig` gets 3 new `clientEarlyTrafficSecret` KAT tests
      against an independently computed (plain Python HKDF) fixture.
- [x] `zig fmt --check` clean on all changed files.

### DIFF
--- a/src/tls/new_session_ticket.zig
+++ b/src/tls/new_session_ticket.zig
@@ -176,8 +176,9 @@ pub fn buildClientTicketState(
         .issued_at_unix_ms = received_at_unix_ms,
         .lifetime_seconds = parsed.ticket_lifetime,
         .early_data = earlyDataPolicy(parsed.max_early_data_size),
-        .transport_compat = connection.transport_compat,
+        .transport_compat = if (parsed.max_early_data_size == null) connection.transport_compat else null,
         .application_compat = connection.application_compat,
+        .early_data_transport_compat = if (parsed.max_early_data_size != null) connection.transport_compat else null,
     });
 
     var state: session.ClientTicketState = .{};
@@ -242,8 +243,9 @@ pub fn buildServerRecoverableStateNoIdentity(
         .issued_at_unix_ms = issued_at_unix_ms,
         .lifetime_seconds = params.ticket_lifetime,
         .early_data = earlyDataPolicy(params.max_early_data_size),
-        .transport_compat = connection.transport_compat,
+        .transport_compat = if (params.max_early_data_size == null) connection.transport_compat else null,
         .application_compat = connection.application_compat,
+        .early_data_transport_compat = if (params.max_early_data_size != null) connection.transport_compat else null,
     });
 
     var state: session.ServerRecoverableState = .{};
@@ -822,7 +824,8 @@ fn exerciseClientTicketBuild(allocator: std.mem.Allocator) !void {
     try std.testing.expectEqualSlices(u8, "\x01\x02", state.ticket_nonce.slice());
     try std.testing.expectEqualStrings("example.com", state.common.server_name.?.slice());
     try std.testing.expectEqualStrings("h2", state.common.application_protocol.?.slice());
-    try std.testing.expect(state.common.transport_compat != null);
+    try std.testing.expect(state.common.transport_compat == null);
+    try std.testing.expect(state.common.early_data_transport_compat != null);
     try std.testing.expect(state.common.application_compat != null);
 }
 
@@ -845,7 +848,8 @@ fn exerciseServerTicketBuild(allocator: std.mem.Allocator) !void {
     defer state.deinit();
     try std.testing.expectEqualStrings("example.com", state.common.server_name.?.slice());
     try std.testing.expectEqualStrings("h2", state.common.application_protocol.?.slice());
-    try std.testing.expect(state.common.transport_compat != null);
+    try std.testing.expect(state.common.transport_compat == null);
+    try std.testing.expect(state.common.early_data_transport_compat != null);
     try std.testing.expect(state.common.application_compat != null);
 }
 
--- a/src/tls/new_session_ticket.zig
+++ b/src/tls/new_session_ticket.zig
@@ -41,6 +41,8 @@ pub const ConnectionResumptionContext = struct {
     auth_binding: session.AuthBinding,
     transport_compat: ?CompatBlob = null,
     application_compat: ?CompatBlob = null,
+    early_data_transport_compat: ?CompatBlob = null,
+    early_data_application_compat: ?CompatBlob = null,
 };
 
 pub const EncodeError = error{
@@ -176,9 +178,10 @@ pub fn buildClientTicketState(
         .issued_at_unix_ms = received_at_unix_ms,
         .lifetime_seconds = parsed.ticket_lifetime,
         .early_data = earlyDataPolicy(parsed.max_early_data_size),
-        .transport_compat = if (parsed.max_early_data_size == null) connection.transport_compat else null,
+        .transport_compat = connection.transport_compat,
         .application_compat = connection.application_compat,
-        .early_data_transport_compat = if (parsed.max_early_data_size != null) connection.transport_compat else null,
+        .early_data_transport_compat = connection.early_data_transport_compat,
+        .early_data_application_compat = connection.early_data_application_compat,
     });
 
     var state: session.ClientTicketState = .{};
@@ -243,9 +246,10 @@ pub fn buildServerRecoverableStateNoIdentity(
         .issued_at_unix_ms = issued_at_unix_ms,
         .lifetime_seconds = params.ticket_lifetime,
         .early_data = earlyDataPolicy(params.max_early_data_size),
-        .transport_compat = if (params.max_early_data_size == null) connection.transport_compat else null,
+        .transport_compat = connection.transport_compat,
         .application_compat = connection.application_compat,
-        .early_data_transport_compat = if (params.max_early_data_size != null) connection.transport_compat else null,
+        .early_data_transport_compat = connection.early_data_transport_compat,
+        .early_data_application_compat = connection.early_data_application_compat,
     });
 
     var state: session.ServerRecoverableState = .{};
@@ -800,6 +804,8 @@ fn allocationSweepContext() ConnectionResumptionContext {
         .auth_binding = session.AuthBinding.fromLeafCertificateDer("leaf"),
         .transport_compat = .{ .format_id = 1, .format_version = 1, .bytes = "transport-compat" },
         .application_compat = .{ .format_id = 2, .format_version = 1, .bytes = "application-compat" },
+        .early_data_transport_compat = .{ .format_id = 3, .format_version = 1, .bytes = "early-transport-compat" },
+        .early_data_application_compat = .{ .format_id = 4, .format_version = 1, .bytes = "early-application-compat" },
     };
 }
 
@@ -824,9 +830,10 @@ fn exerciseClientTicketBuild(allocator: std.mem.Allocator) !void {
     try std.testing.expectEqualSlices(u8, "\x01\x02", state.ticket_nonce.slice());
     try std.testing.expectEqualStrings("example.com", state.common.server_name.?.slice());
     try std.testing.expectEqualStrings("h2", state.common.application_protocol.?.slice());
-    try std.testing.expect(state.common.transport_compat == null);
+    try std.testing.expect(state.common.transport_compat != null);
     try std.testing.expect(state.common.early_data_transport_compat != null);
     try std.testing.expect(state.common.application_compat != null);
+    try std.testing.expect(state.common.early_data_application_compat != null);
 }
 
 fn exerciseServerTicketBuild(allocator: std.mem.Allocator) !void {
@@ -848,9 +855,10 @@ fn exerciseServerTicketBuild(allocator: std.mem.Allocator) !void {
     defer state.deinit();
     try std.testing.expectEqualStrings("example.com", state.common.server_name.?.slice());
     try std.testing.expectEqualStrings("h2", state.common.application_protocol.?.slice());
-    try std.testing.expect(state.common.transport_compat == null);
+    try std.testing.expect(state.common.transport_compat != null);
     try std.testing.expect(state.common.early_data_transport_compat != null);
     try std.testing.expect(state.common.application_compat != null);
+    try std.testing.expect(state.common.early_data_application_compat != null);
 }
 
 fn exerciseClientTicketClone(allocator: std.mem.Allocator) !void {
--- a/src/tls/session.zig
+++ b/src/tls/session.zig
@@ -307,6 +307,8 @@ pub const ResumableSessionCommon = struct {
     early_data: EarlyDataPolicy = .resume_only,
     transport_compat: ?CompatSnapshot = null,
     application_compat: ?CompatSnapshot = null,
+    early_data_transport_compat: ?CompatSnapshot = null,
+    early_data_application_compat: ?CompatSnapshot = null,
 
     pub const CompatBlobParams = struct {
         format_id: u16,
@@ -325,6 +327,8 @@ pub const ResumableSessionCommon = struct {
         early_data: EarlyDataPolicy = .resume_only,
         transport_compat: ?CompatBlobParams = null,
         application_compat: ?CompatBlobParams = null,
+        early_data_transport_compat: ?CompatBlobParams = null,
+        early_data_application_compat: ?CompatBlobParams = null,
     };
 
     pub const InitError = error{
@@ -394,6 +398,16 @@ pub const ResumableSessionCommon = struct {
             try snap.init(allocator, blob.format_id, blob.format_version, blob.bytes, limits.max_application_compat_len);
             next.application_compat = snap;
         }
+        if (params.early_data_transport_compat) |blob| {
+            var snap: CompatSnapshot = .{};
+            try snap.init(allocator, blob.format_id, blob.format_version, blob.bytes, limits.max_transport_compat_len);
+            next.early_data_transport_compat = snap;
+        }
+        if (params.early_data_application_compat) |blob| {
+            var snap: CompatSnapshot = .{};
+            try snap.init(allocator, blob.format_id, blob.format_version, blob.bytes, limits.max_application_compat_len);
+            next.early_data_application_compat = snap;
+        }
 
         self.moveFrom(&next);
     }
@@ -402,8 +416,12 @@ pub const ResumableSessionCommon = struct {
         self.resumption_psk.deinit();
         if (self.transport_compat) |*snap| snap.deinit();
         if (self.application_compat) |*snap| snap.deinit();
+        if (self.early_data_transport_compat) |*snap| snap.deinit();
+        if (self.early_data_application_compat) |*snap| snap.deinit();
         self.transport_compat = null;
         self.application_compat = null;
+        self.early_data_transport_compat = null;
+        self.early_data_application_compat = null;
     }
 
     /// `out` must be zero-valued or a previously-initialized, live value —
@@ -438,6 +456,16 @@ pub const ResumableSessionCommon = struct {
             var cloned: CompatSnapshot = .{};
             try snap.cloneInto(allocator, &cloned);
             next.application_compat = cloned;
+        }
+        if (self.early_data_transport_compat) |*snap| {
+            var cloned: CompatSnapshot = .{};
+            try snap.cloneInto(allocator, &cloned);
+            next.early_data_transport_compat = cloned;
+        }
+        if (self.early_data_application_compat) |*snap| {
+            var cloned: CompatSnapshot = .{};
+            try snap.cloneInto(allocator, &cloned);
+            next.early_data_application_compat = cloned;
         }
 
         out.moveFrom(&next);
--- a/src/tls/session.zig
+++ b/src/tls/session.zig
@@ -862,6 +862,8 @@ const field_lifetime_seconds: u16 = 0x0007;
 const field_early_data: u16 = 0x0008;
 const field_transport_compat: u16 = 0x8001;
 const field_application_compat: u16 = 0x8002;
+const field_early_data_transport_compat: u16 = 0x8003;
+const field_early_data_application_compat: u16 = 0x8004;
 
 const field_ticket: u16 = 0x0010;
 const field_ticket_age_add: u16 = 0x0011;
@@ -929,6 +931,8 @@ fn commonEncodedLen(common: *const ResumableSessionCommon) usize {
     total += tlvLen(earlyDataFieldLen(common.early_data));
     if (common.transport_compat) |*snap| total += tlvLen(4 + snap.slice().len);
     if (common.application_compat) |*snap| total += tlvLen(4 + snap.slice().len);
+    if (common.early_data_transport_compat) |*snap| total += tlvLen(4 + snap.slice().len);
+    if (common.early_data_application_compat) |*snap| total += tlvLen(4 + snap.slice().len);
     return total;
 }
 
@@ -953,6 +957,8 @@ fn commonFieldCount(common: *const ResumableSessionCommon) usize {
     if (common.application_protocol != null) count += 1;
     if (common.transport_compat != null) count += 1;
     if (common.application_compat != null) count += 1;
+    if (common.early_data_transport_compat != null) count += 1;
+    if (common.early_data_application_compat != null) count += 1;
     return count;
 }
 
@@ -964,6 +970,12 @@ fn checkCommonAgainstLimits(common: *const ResumableSessionCommon, limits: Limit
         if (snap.slice().len > limits.max_transport_compat_len) return error.FieldTooLarge;
     }
     if (common.application_compat) |*snap| {
+        if (snap.slice().len > limits.max_application_compat_len) return error.FieldTooLarge;
+    }
+    if (common.early_data_transport_compat) |*snap| {
+        if (snap.slice().len > limits.max_transport_compat_len) return error.FieldTooLarge;
+    }
+    if (common.early_data_application_compat) |*snap| {
         if (snap.slice().len > limits.max_application_compat_len) return error.FieldTooLarge;
     }
 }
@@ -1010,6 +1022,12 @@ fn validateCommonForEncoding(common: *const ResumableSessionCommon) EncodeError!
         if (snap.blob.len > snap.blob.bytes.len) return error.InvalidState;
     }
     if (common.application_compat) |*snap| {
+        if (snap.blob.len > snap.blob.bytes.len) return error.InvalidState;
+    }
+    if (common.early_data_transport_compat) |*snap| {
+        if (snap.blob.len > snap.blob.bytes.len) return error.InvalidState;
+    }
+    if (common.early_data_application_compat) |*snap| {
         if (snap.blob.len > snap.blob.bytes.len) return error.InvalidState;
     }
 }
@@ -1099,6 +1117,10 @@ fn writeCommon(out: []u8, pos: *usize, common: *const ResumableSessionCommon, co
     if (common.transport_compat) |*snap| writeCompatSnapshot(out, pos, field_transport_compat, snap, compat_scratch);
     if (common.application_compat) |*snap|
         writeCompatSnapshot(out, pos, field_application_compat, snap, compat_scratch);
+    if (common.early_data_transport_compat) |*snap|
+        writeCompatSnapshot(out, pos, field_early_data_transport_compat, snap, compat_scratch);
+    if (common.early_data_application_compat) |*snap|
+        writeCompatSnapshot(out, pos, field_early_data_application_compat, snap, compat_scratch);
 }
 
 fn writeHeader(out: []u8, kind: RecordType, field_section_len: u32) void {
@@ -1186,6 +1208,8 @@ const CommonFields = struct {
     early_data: ?EarlyDataPolicy = null,
     transport_compat: ?RawCompat = null,
     application_compat: ?RawCompat = null,
+    early_data_transport_compat: ?RawCompat = null,
+    early_data_application_compat: ?RawCompat = null,
 };
 
 const RawCompat = struct {
@@ -1290,6 +1314,14 @@ fn parseSharedField(builder: *CommonFields, limits: Limits, field_id: u16, value
             builder.application_compat = try decodeRawCompat(value, limits.max_application_compat_len);
             return true;
         },
+        field_early_data_transport_compat => {
+            builder.early_data_transport_compat = try decodeRawCompat(value, limits.max_transport_compat_len);
+            return true;
+        },
+        field_early_data_application_compat => {
+            builder.early_data_application_compat = try decodeRawCompat(value, limits.max_application_compat_len);
+            return true;
+        },
         else => return false,
     }
 }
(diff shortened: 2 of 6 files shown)


src/tls/tls13_backend.zig
```diff
@@ -292,6 +294,77 @@ const hello_retry_request_random = [32]u8{
...
+    age_skew_tolerance_ms: u64 = 60_000,
+};
+
+pub const EarlyDataReplayDecision = enum { allow, replay, unavailable };
+
+/// Bounded, non-secret metadata about an early-data attempt handed to the
+/// anti-replay gate — never the raw ticket, PSK, or binder. #368 owns the
+/// actual replay store; this is only the seam.
+pub const EarlyDataReplayCandidate = struct {
+    selected_identity: u16 = 0,
```
Jaruso: **Blocking the #368 handoff:** this candidate cannot identify a replay. For every 0-RTT handshake that can actually be accepted, TLS requires the selected PSK to be wire identity **0**, so every callback invocation here is effectively `.{ .selected_identity = 0 }`. #368 needs to derive/store a bounded replay key from ticket/ClientHello context; with this API it cannot distinguish two different tickets or two different early handshakes.

Please make this seam carry a fixed-size, non-secret fingerprint while `selectPsk()` still has both the opaque identity and captured complete ClientHello available. For example:

```zig
pub const EarlyDataReplayCandidate = struct {
    ticket_identity_hash: [Sha256.digest_length]u8,
    client_hello_hash: [Sha256.digest_length]u8,
    issued_at_unix_ms: i64,
};

var ticket_hash: [Sha256.digest_length]u8 = undefined;
Sha256.hash(pair.identity.identity, &ticket_hash, .{});

var hello_hash: [Sha256.digest_length]u8 = undefined;
Sha256.hash(capture.message[0..capture.message_len], &hello_hash, .{});
```

A single domain-separated `replay_key: [32]u8` derived from those inputs is also fine. The important constraints are: distinct accepted handshakes must be distinguishable, replaying the same one must reproduce the same bounded key, and the callback must never receive the raw ticket, PSK, binder, or request body. Please add a focused test proving same context => same candidate and changed ticket/ClientHello => changed candidate.

src/tls/tls13_backend.zig
```diff
@@ -814,6 +918,54 @@ pub const Tls13Backend = struct {
...
+    pub fn setEarlyDataReplayGate(self: *Tls13Backend, gate: EarlyDataReplayGate) HandshakeError!void {
+        if (self.role != .server) return error.InvalidHandshakeState;
+        if (self.core.handshake_lifecycle != .idle) return error.InvalidHandshakeState;
+        self.early_data_replay_gate = gate;
+    }
+
+    /// Client: whether the ClientHello just sent actually attempted 0-RTT.
+    /// Independent of acceptance — see `earlyDataAccepted`.
+    pub fn earlyDataAttempted(self: *const Tls13Backend) bool {
+        return self.client_early_data_attempted;
```
Jaruso: **Blocking the record rejected-0-RTT path:** this accessor is client-only, while the server's `client_hello_early_data_seen` is private. On a CH that carries `early_data` but whose identity misses/is incompatible and therefore falls back to a full handshake, `decideServerEarlyData()` is never reached; externally that server looks identical to a connection where the client never attempted early data at all.

The later record carrier needs this distinction *before handshake completion*: when the peer attempted 0-RTT but TLS rejected it, it may need to boundedly skip rejected early ciphertext until the handshake-key record arrives. It must not do that for arbitrary pre-handshake ciphertext when no early-data attempt existed, and the adapter must not reparse ClientHello to recover this bit.

Please expose the peer attempt state, or replace these booleans with a role-neutral early-data state. A minimal seam would be:

```zig
pub fn peerEarlyDataAttempted(self: *const Tls13Backend) bool {
    return self.role == .server and self.client_hello_early_data_seen;
}
```

and preserve that bit on ordinary PSK miss/incompatibility/full-handshake fallback (clearing it only on terminal failure/deinit). A stronger API would expose `not_attempted | attempted_rejected(reason) | accepted` so carriers cannot accidentally conflate the states.

src/tls/tls13_backend.zig
```diff
@@ -1556,6 +1727,15 @@ pub const Tls13Backend = struct {
             try w.bytes(payload);
         }
 
+        // early_data (#366): an empty extension announcing a 0-RTT attempt.
+        // Order relative to `pre_shared_key` doesn't matter under RFC
+        // 8446 §4.2.11 (only `pre_shared_key` itself must be last), so this
+        // is placed before the PSK block below for wire-format stability.
+        if (self.client_early_data_attempted) {
+            try w.u16_(ext_early_data);
```
Jaruso, line 1827: **Correctness blocker at the size boundary.** #366 explicitly calls for the 4-byte `early_data` extension to be included in ClientHello sizing. Right now `startImpl()` computes `base_len = clientHelloEncodedLen()`, `planPskOffer(base_len)` is allowed to pack identities/binders all the way up to `max_message_len`, and only *after that* `planEarlyDataAttempt()` decides to emit these four additional bytes. `clientHelloEncodedLen()` does not include them.

That means a boundary-size early-capable offer can be retained as fitting and then fail here with `HandshakeBufferOverflow` after `core.start()` has already advanced lifecycle state/generated the key share. Resumption is an optimization; this should be rejected/dropped during the same pre-start sizing transaction, not become a late handshake failure.

Please make early-attempt planning part of the exact preflight. Conceptually:

```zig
// While deciding the first surviving offer, before committing it:
const early_overhead: usize = if (
    is_first_surviving and
    self.client_early_data_intent.enabled and
    self.client_early_data_intent.max_bytes != 0 and
    ticket.common.early_data != .resume_only
) 4 else 0;

const candidate_total = try checkedAdd(
    try checkedAdd(total, entry_len + 1 + hash_len),
    early_overhead,
);
```

A two-pass planner is also fine; the invariant should be that `planPskOffer` + early planning know the *exact* final CH size before `core.start()`. Please add a regression with an identity sized so the old PSK calculation reaches the max boundary: the backend must either keep a correctly-sized offer or drop/fallback cleanly, never late-overflow from these four bytes.

src/tls/tls13_backend.zig
```diff
@@ -2003,6 +2239,12 @@ pub const Tls13Backend = struct {
             if (self.resume_compat.transport == .exact and !compatCompatible(stored.transport_compat, received_transport)) return error.IllegalParameter;
             if (self.resume_compat.application == .exact and !compatCompatible(stored.application_compat, self.candidateApplicationCompat())) return error.IllegalParameter;
 
+            // #366: a resume-only ticket can never be the basis for an
+            // accepted `early_data` extension, regardless of what the
+            // client attempted.
+            if (early_data_seen and stored.early_data == .resume_only) return error.IllegalParameter;
+            if (early_data_seen) self.early_data_accepted = true;
```
Jaruso: This leaves the new public state internally inconsistent on the client. The accessor directly above documents that client `earlyDataDecision()` "only ever reaches `.accepted`", but on a valid accepted EE this branch sets only `early_data_accepted`; `early_data_decision` remains `.not_attempted`.

Please update both pieces atomically:

```zig
if (early_data_seen) {
    self.early_data_accepted = true;
    self.early_data_decision = .accepted;
}
```

and extend the accepted round-trip test with:

```zig
try std.testing.expectEqual(
    tls_backend.EarlyDataDecision.accepted,
    resumed.client_backend.earlyDataDecision(),
);
```

That gives later metrics/carriers one coherent source of truth instead of a boolean saying accepted while the typed decision says not attempted.

src/tls/tls13_backend.zig
```diff
@@ -2828,6 +3114,41 @@ pub const Tls13Backend = struct {
         return null;
     }
 
+    const ServerEarlyDataInputs = struct {
+        selected_index: usize,
+        compatibility: session.EarlyDataEligibility,
```
Jaruso, line 3244: **Blocking #366's 0-RTT-specific compatibility contract.** This `EarlyDataEligibility` is produced from the same `CandidateContext`/`ResumeCompatibilityPolicy` used for ordinary 1-RTT resumption. In the #495 production integration, transport/application snapshots are intentionally ignored for ordinary 1-RTT. If a later QUIC slice flips that policy to `.exact` to enforce remembered 0-RTT transport state, a mismatch will make `decision.resumption != .eligible` *before this function runs* and will reject PSK resumption itself. That violates #366's AC that 0-RTT-specific TP/application incompatibility rejects only early data while valid 1-RTT resumption continues.

This slice does not need to implement the QUIC comparison, but it does need to leave an independent input/seam for it. The issue blueprint already suggested passing the selected session into the early decision. For example:

```zig
pub const EarlyDataCompatibility = enum {
    compatible,
    transport_incompatible,
    application_incompatible,
    unavailable,
};

const ServerEarlyDataInputs = struct {
    selected_index: usize,
    compatibility: session.EarlyDataEligibility, // ticket capability
    early_compatibility: EarlyDataCompatibility,
    age_skew: pre_shared_key.AgeSkew,
    session: *const session.ResumableSessionCommon,
};
```

The QUIC/H3 composition can later compute remembered-TP/application compatibility from `session` and current connection state without importing QUIC into the TLS engine. A transport/application mismatch must change only `EarlyDataDecision`, not `decision.resumption` or PSK selection. As written, `EarlyDataDecision.transport_incompatible` and `.application_incompatible` have no reachable 0-RTT-only path.

src/tls/tls13_backend.zig
```diff
@@ -2828,6 +3114,41 @@ pub const Tls13Backend = struct {
...
+        if (inputs.selected_index != 0) return .selected_identity_not_zero;
+        switch (inputs.compatibility) {
+            // Reached only for the selected (already resumption-eligible)
+            // candidate, so `.incompatible` (which `evaluateCompatibility`
+            // only produces when resumption itself is ineligible) cannot
+            // actually occur here; `.disabled` is the real "ticket is
+            // resume-only" case for an identity-0, early_data-requesting
+            // candidate.
+            .disabled, .incompatible => return .ticket_not_capable,
+            .allowed => {},
```
Jaruso, line 3272: Please do not discard the payload of `.allowed(max)` here. That `u32` is the ticket's advertised 0-RTT allowance, and the next carrier slice needs an authoritative limit **while early data is being processed**, before the handshake commits. On the client, the offered ticket has been moved into backend ownership; on the server, `takeSelectedServerPsk()` intentionally cannot release selected session state until commit. So after this switch there is no clean public way for record/QUIC code to recover the limit.

Store/expose the effective early-data allowance as part of the backend's early-data state. For example:

```zig
early_data_max_bytes: ?u32 = null,

pub fn earlyDataMaxBytes(self: *const Tls13Backend) ?u32 {
    return self.early_data_max_bytes;
}
```

Client planning should retain:

```zig
self.early_data_max_bytes = @min(ticket_max, self.client_early_data_intent.max_bytes);
```

and server acceptance should retain the selected ticket's `ticket_max` (the composition layer can additionally clamp against its current local resource policy). Clear it with the rest of early-data state on rejection/failure.

The accepted test is a useful regression case already: it issues `max_early_data_size = 32` but configures client intent `16384`; assert that the effective client allowance is 32 rather than losing that information.

src/tls/tls13_backend_tests.zig
```diff
@@ -620,6 +631,497 @@ test "PSK round trip: an offered, resolved, and verified ticket resumes the hand
...
+    // The ticket was issued and received at simulated t=1000 (see
+    // `issueEarlyCapableTicket`'s `TicketCapture.now`); the server's own
+    // resolver clock (`IdentityResolver.now`) reports t=2000, for a genuine
+    // 1000ms server-observed age. Reporting a much larger "now" here makes
+    // the *client's* apparent ticket age diverge sharply from that — real
+    // clock skew, not merely two different-but-consistent clocks.
+    return 6000;
+}
+
+test "0-RTT is rejected for a ticket-age skew outside the configured tolerance" {
```
Jaruso, line 963: This proves one outside-tolerance rejection, but #366's acceptance criteria explicitly require **exact skew boundaries and overflow cases**. Since the helper deliberately uses `i128` to handle `-minInt(i64)`, please lock that behavior down now instead of leaving the security boundary untested.

At minimum cover:

```zig
try expect(skewWithinTolerance(10, 10));
try expect(skewWithinTolerance(-10, 10));
try expect(!skewWithinTolerance(11, 10));
try expect(!skewWithinTolerance(-11, 10));
try expect(skewWithinTolerance(std.math.minInt(i64), @as(u64, 1) << 63));
try expect(!skewWithinTolerance(std.math.minInt(i64), (@as(u64, 1) << 63) - 1));
```

These can be direct helper tests in `tls13_backend.zig` or equivalent handshake fixtures. For the live policy path, also retain one exact `== tolerance` accepted case and one `tolerance + 1` rejected case so the comparison cannot accidentally drift from inclusive to exclusive.

Jaruso (commented):
**Blocking review — I would not merge this head yet.**

I reviewed this as **slice 1 of #366**, not as an attempted closure of the whole XL issue. The core TLS work is directionally strong: the `c e traffic` derivation is correct and independently fixture-tested; ClientHello keeps `pre_shared_key` last; the client derives from the complete ClientHello with real binders; the server makes the early-data decision only after normal PSK compatibility + binder verification; identity 0 and EE signaling are enforced; replay defaults fail closed without killing valid 1-RTT resumption; cleanup paths clear the new acceptance state; and CI is green across the current matrix.

I found **six correctness/contract blockers** for this slice, plus one explicit #366 test-AC gap:

1. ClientHello size preflight does not reserve the 4-byte `early_data` extension, so a boundary-size PSK offer can pass `planPskOffer()` and then overflow late in `sendClientHello()` after `core.start()` has mutated handshake state.
2. `EarlyDataReplayCandidate` contains only `selected_identity`; every acceptable 0-RTT handshake necessarily has identity 0, so this is not a usable anti-replay identity for #368.
3. The server has no public way for the record/QUIC carrier to distinguish **peer attempted-but-rejected early data** from **no early-data attempt**, especially on resolver miss/incompatibility/full-handshake fallback where `decideServerEarlyData()` is never reached.
4. The client sets `early_data_accepted` but never updates its documented `earlyDataDecision()` to `.accepted`.
5. 0-RTT-specific transport/application compatibility is currently coupled to ordinary `ResumeCompatibilityPolicy`; that cannot satisfy #366's requirement that remembered-TP/application incompatibility reject only 0-RTT while valid 1-RTT resumption continues.
6. `EarlyDataEligibility.allowed(max)` is discarded. The later record/QUIC carrier needs the negotiated/effective byte allowance while early data is being processed, before selected session state is otherwise recoverable.
7. #366 explicitly requires exact ticket-age-skew boundary and overflow tests; the PR currently covers only one outside-tolerance case.

I left detailed inline comments with concrete suggested shapes for each. I am *not* asking this PR to implement ticket advertisement, record/QUIC packet protection, EndOfEarlyData, provenance, or HTTP request safety; those remain correctly deferred to the later slices.

GitHub does not allow the PR author account to submit a formal `REQUEST_CHANGES` review on its own PR, so this is recorded as a COMMENT review; semantically, these are blocking findings.

src/tls/tls13_backend.zig
Jaruso: Second-pass status: this is **partially fixed but still open**. Reserving four bytes whenever intent is enabled prevents the late writer overflow, which is good, but it is not the exact preflight requested here. When intent is enabled but the first surviving ticket is `resume_only` (so `planEarlyDataAttempt()` will not emit `early_data`), those four bytes are still charged to `planPskOffer()`. At the boundary that can drop a later otherwise-valid 1-RTT PSK offer simply because 0-RTT intent was configured, even though no 0-RTT extension is sent.

Please make the final sizing transaction exact (or explicitly do a two-pass plan: determine surviving identity 0 / early eligibility, then budget the 4 bytes only when it will actually be emitted) and add the boundary regression requested above. The invariant should be: enabling early-data intent must not change the ordinary PSK offer set when the first surviving ticket cannot attempt early data.

src/tls/tls13_backend.zig
```diff
@@ -1704,6 +1989,28 @@ pub const Tls13Backend = struct {
...
+        const offers = self.constClientPskOffers();
+        if (offers.isEmpty()) return;
+
+        const max = switch (offers.constSlice()[0].common.early_data) {
+            .resume_only => return,
+            .early_data_capable => |n| n,
+        };
+        if (max == 0 or self.client_early_data_intent.max_bytes == 0) return;
+
+        self.client_early_data_attempted = true;
```
Jaruso, line 2017: **Still blocking the client carrier handoff.** The server-side fix now preserves the selected ticket's `allowed(max)`, but the client side still throws `max` away here and only records a boolean. The client record/QUIC carrier has to know how many bytes it is allowed to *send before EncryptedExtensions arrives*, after the ticket has already moved into backend ownership.

Please retain the effective client budget during planning:

```zig
const ticket_max = switch (offers.constSlice()[0].common.early_data) {
    .resume_only => return,
    .early_data_capable => |n| n,
};
const effective = @min(ticket_max, self.client_early_data_intent.max_bytes);
if (effective == 0) return;

self.client_early_data_attempted = true;
self.early_data_max_bytes = effective;
```

Then make `earlyDataMaxBytes()` role-neutral/documented as: client = maximum bytes this attempt may send; server = maximum bytes this accepted attempt may receive. Clear the client value on terminal failure and after an EE rejection once the carrier has observed rejection. The existing accepted fixture (`ticket max = 32`, intent = 16384) should assert **both** client and server report 32.

src/tls/tls13_backend.zig
```diff
@@ -1770,6 +2077,16 @@ pub const Tls13Backend = struct {
...
             len = try checkedAdd(len, 2 + 2 + payload.len);
         }
+        // #366: reserved whenever the client has opted in, regardless of
+        // whether `planEarlyDataAttempt` (which runs later, after this
+        // budget is used by `planPskOffer`'s own ticket-fitting arithmetic)
+        // ends up actually attempting 0-RTT — so a too-large PSK offer is
+        // rejected/trimmed by `planPskOffer` itself instead of the
+        // 4-byte `early_data` extension silently pushing `sendClientHello`'s
+        // fixed-capacity buffer past `max_message_len`.
+        if (self.client_early_data_intent.enabled) {
```
Jaruso: This prevents the late `HandshakeBufferOverflow`, but it is still not the exact preflight #366 calls for. Charging 4 bytes solely because intent is enabled means a first surviving `resume_only` ticket (so no `early_data` extension will be emitted) can cause a later PSK identity at the boundary to be dropped. In other words, merely enabling 0-RTT intent can change the ordinary 1-RTT offer set even when identity 0 is not early-capable.

Please fold the 4-byte overhead into the PSK planning transaction only once the first surviving offer is known to be early-capable, or do a two-pass plan that determines `client_early_data_attempted` before the final size commit. The invariant should be:

```text
final_preflight_len == bytes sendClientHello() will actually emit
```

Add the boundary regression from the first review, plus a resume-only identity-0 case proving `enabled=true` does not drop an otherwise-fitting later 1-RTT offer when no `early_data` extension is emitted.

src/tls/tls13_backend.zig
```diff
@@ -2804,18 +3165,61 @@ pub const Tls13Backend = struct {
...
+            // early-data rejection never rejects an otherwise-valid PSK
+            // resumption.
+            var identity_fingerprint: [hash_len]u8 = undefined;
+            Sha256.hash(pair.identity.identity, &identity_fingerprint, .{});
+            const early_decision = self.decideServerEarlyData(.{
+                .selected_index = attempts,
+                .compatibility = decision.early_data,
+                .age_skew = age_skew,
+                .compat_candidate = .{
+                    .stored_transport_compat = if (hit.state.common.transport_compat) |*snap| snap.slice() else null,
```
Jaruso: **The compatibility seam is decoupled now, but this transport state is the wrong direction for QUIC 0-RTT.** On the server, `hit.state.common.transport_compat` was built by `resumptionContext()` from `peerTransportCompat()`, so it represents the **client's previous** transport parameters. `self.candidateTransportCompat()` here is the **client's current** ClientHello transport parameters. Comparing those can detect a client-side change, but #366's QUIC acceptance rule needs to ensure the **server's current local limits are not lower than the limits the server previously advertised and the client used for 0-RTT**.

There is a second problem: `resumptionContext()` only stores these snapshots when `resume_compat == .exact`; #495 intentionally uses `.ignore` for ordinary 1-RTT. Populating the same field by flipping `.exact` would re-couple early compatibility to resumption and recreate the original bug.

Please make the seam capable of carrying an **early-data-specific remembered snapshot independent of ordinary resumption compatibility**, and preserve the snapshot discriminator too (`format_id` + `format_version`, not only bytes). One workable shape is:

```zig
pub const CompatView = struct {
    format_id: u16,
    format_version: u16,
    bytes: []const u8,
};

pub const EarlyDataCompatibilityCandidate = struct {
    remembered_transport: ?CompatView = null,   // values this endpoint advertised last time
    remembered_application: ?CompatView = null,
};
```

The gate's `ctx` can own the **current local** QUIC/H3 state and compare it against these remembered values. The later ticket-issuance slice can populate these early-only snapshots without changing ordinary `ResumeCompatibilityPolicy`. A test for this slice should at least prove the gate sees the exact format id/version/bytes it was given and that `.transport_incompatible` still leaves `psk_authenticated == true`.

Jaruso (commented):
Second-pass review on `439b376`.

A lot improved: the replay gate now receives a stable ticket fingerprint, server-side attempt visibility is role-aware, client acceptance state is coherent, early-specific compatibility has its own decision path, and the skew helper has solid boundary/overflow coverage. CI is green on this head.

I still would **not merge this slice yet**. Three blockers remain:

1. The client still does not retain its effective early-data byte budget. The server now exposes the ticket allowance, but the client carrier needs `min(ticket_max, intent.max_bytes)` before it sends any 0-RTT bytes.
2. ClientHello sizing is now safe from late overflow, but only by always reserving four bytes whenever intent is enabled. That can unnecessarily drop an otherwise-valid 1-RTT PSK offer when identity 0 is resume-only and no `early_data` extension will actually be emitted. The sizing transaction should be exact and needs the requested boundary regression.
3. The new early-data compatibility callback is decoupled from ordinary 1-RTT eligibility, but the transport data fed into it is the server's **peer/client** previous/current transport-parameter snapshot. QUIC 0-RTT acceptance needs the server's previously-advertised limits versus its current local limits. In addition, the flattened slices discard `format_id` / `format_version`, so the callback cannot safely distinguish snapshot formats.

These are all foundation/API issues for the next slices, not requests to pull record/QUIC/HTTP implementation into this PR. Once these are fixed, I think slice 1 is in very good shape.

src/tls/new_session_ticket.zig
```diff
@@ -176,8 +176,9 @@ pub fn buildClientTicketState(
         .issued_at_unix_ms = received_at_unix_ms,
         .lifetime_seconds = parsed.ticket_lifetime,
         .early_data = earlyDataPolicy(parsed.max_early_data_size),
-        .transport_compat = connection.transport_compat,
+        .transport_compat = if (parsed.max_early_data_size == null) connection.transport_compat else null,
```
Jaruso: **Blocking regression in ordinary 1-RTT compatibility.** For every ticket that advertises early data, this moves the one existing transport snapshot into `early_data_transport_compat` and forces the ordinary `transport_compat` to `null`. That means enabling early capability changes the semantics of the same ticket for normal PSK resumption. `session.checkResumeEligibility()` uses symmetric optional equality, so with `ResumeCompatibilityPolicy.transport = .exact`, a current non-null candidate compared against this forced-null stored value is a transport mismatch; the client has the analogous exact recheck after ServerHello. An early-capable ticket can therefore become unusable for ordinary 1-RTT resumption even when 0-RTT is disabled or rejected.

The ordinary and early-data snapshots need to coexist rather than multiplex through `ConnectionResumptionContext.transport_compat`. I would extend the context explicitly:

```zig
pub const ConnectionResumptionContext = struct {
    cipher_suite: algorithms.CipherSuite,
    application_protocol: ?[]const u8 = null,
    auth_binding: session.AuthBinding,

    // Existing 1-RTT compatibility semantics — unchanged.
    transport_compat: ?CompatBlob = null,
    application_compat: ?CompatBlob = null,

    // Independent remembered state used only for early-data admission.
    early_data_transport_compat: ?CompatBlob = null,
    early_data_application_compat: ?CompatBlob = null,
};
```

Then `resumptionContext()` should continue producing the ordinary fields exactly as before, while `resumptionContextForTicket(true)` sets only the early-data fields (client: remembered peer/server transport state; server: the local transport state it advertised). The builders can pass both sets through to `ResumableSessionCommon` without nulling the ordinary field.

Please add a regression proving an early-capable ticket with ordinary transport compatibility set to `.exact` still preserves/enforces the pre-existing 1-RTT snapshot, while the independent early-data gate receives the early-specific remembered snapshot. This is important because #366 requires early rejection to preserve otherwise-valid ordinary PSK resumption; it must not accomplish that by deleting normal resumption compatibility state.

src/tls/session.zig
```diff
@@ -307,6 +307,8 @@ pub const ResumableSessionCommon = struct {
     early_data: EarlyDataPolicy = .resume_only,
     transport_compat: ?CompatSnapshot = null,
     application_compat: ?CompatSnapshot = null,
+    early_data_transport_compat: ?CompatSnapshot = null,
+    early_data_application_compat: ?CompatSnapshot = null,
```
Jaruso, line 312: **Blocking stateless/persistence bug:** these new owned fields are not represented anywhere in the internal session codec below. `commonEncodedLen`, `commonFieldCount`, `checkCommonAgainstLimits`, `validateCommonForEncoding`, `writeCommon`, `CommonFields`, `parseSharedField`, and `buildCommon` still know only `transport_compat` / `application_compat`.

That means the new remembered early-data state survives the in-memory/stateful-cache path, but is silently dropped by every encode/decode boundary. In particular, `Runtime.createStatelessIdentity()` seals `ServerRecoverableState` through `ticket_protection.Protector`, which uses the session codec, and the resolver decodes it back later. After that round trip `early_data_transport_compat` is `null`, so the QUIC early-data gate no longer has the limits the server previously advertised. Client cache persistence has the same loss for remembered server state.

Please make both new snapshots first-class optional codec fields. The existing optional-field scheme makes this backward-compatible, for example:

```zig
const field_early_data_transport_compat: u16 = 0x8003;
const field_early_data_application_compat: u16 = 0x8004;
```

and then include them in:

```text
commonEncodedLen
commonFieldCount
checkCommonAgainstLimits
validateCommonForEncoding
writeCommon
CommonFields
parseSharedField
buildCommon
```

using the same `format_id | format_version | bytes` encoding and transport/application limits as the ordinary snapshots.

Please add at least:

```text
client encode -> decode preserves both early snapshots exactly
server encode -> decode preserves both early snapshots exactly
stateless Runtime createIdentity -> resolver preserves early_data_transport_compat
```

The stateless runtime test is the important integration regression here; the current compatibility test resolves an in-memory `ServerRecoverableState`, so it cannot catch this loss.

Jaruso (commented):
Blocking follow-up review on `0eff864` — the previous client budget and exact ClientHello sizing blockers are fixed, and I resolved those threads. The early-data compatibility direction is also improved in-memory, but the latest fix introduces two state-model problems that still make this unsafe to use as the foundation for the record/QUIC slices: (1) early-capable tickets currently replace the ordinary transport compatibility snapshot instead of carrying an independent early-data snapshot alongside it, which changes existing 1-RTT `.exact` resumption semantics; and (2) the new early-data compatibility snapshots are not part of the session codec, so stateless tickets and persisted client/server cache state silently lose them across encode/decode. I would not merge this head until both are corrected and covered by round-trip/runtime tests.

The TLS negotiation/secret derivation itself still looks good on this pass. Current unit/appliance/security jobs I inspected are green; the overall CI workflow was still finishing packaging/dependency-audit jobs at review time.

src/tls/tls13_backend.zig
```diff
@@ -3343,6 +3814,17 @@ pub const Tls13Backend = struct {
         };
     }
 
+    fn resumptionContextForTicket(self: *const Tls13Backend, early_data_capable: bool) new_session_ticket.ConnectionResumptionContext {
+        var ctx = self.resumptionContext();
+        if (early_data_capable) {
+            ctx.early_data_transport_compat = switch (self.role) {
+                .client => self.peerTransportCompat(),
+                .server => self.localTransportCompat(),
```
Jaruso, line 3822: **Non-blocking handoff note for the later H3 slice:** the state model, codec, and `EarlyDataCompatibilityCandidate` now correctly support `early_data_application_compat`, but this helper only populates the early transport snapshot. `selectPsk()` already forwards `remembered_application`, so it will remain `null` in the real backend-issued ticket path until this is wired.

I would *not* hold slice 1 for that now—the important architectural/persistence seam is complete—but before H3 0-RTT is enabled the H3 integration needs to populate the remembered application snapshot independently of `resume_compat.application`. If the existing application snapshot has the right semantics, that may be as small as:

```zig
ctx.early_data_application_compat = self.ownedApplicationCompat();
```

If H3 needs peer SETTINGS rather than the existing owned snapshot, use a role-aware early-application seam instead. The key invariant for the later slice is that `EarlyDataDecision.application_incompatible` has real remembered application state to evaluate without re-coupling it to ordinary 1-RTT compatibility.

Jaruso (commented):
Re-review complete on `b02dafa14de66fbadb9a507b153d412d86df4200`.

**No blocking findings remain for #366 slice 1. I consider this head ready to merge.**

I rechecked the full slice, not only the latest patch. The previously blocking compatibility/state issues are now addressed correctly:

- ordinary 1-RTT `transport_compat` / `application_compat` remain independent from the new early-data-only snapshots, so advertising early capability no longer weakens or breaks ordinary `.exact` resumption compatibility;
- `early_data_transport_compat` and `early_data_application_compat` are owned/cloned/deinitialized with the session state and are first-class optional fields in the internal codec;
- encoded length, field count, per-field limits, validation, write, parse, and reconstruction all include the new fields;
- client and server encode/decode regressions preserve format id/version/bytes for both ordinary and early snapshots;
- the stateless `Runtime.createIdentity -> resolver` integration test proves remembered early transport state survives ticket protection instead of only working through the in-memory stateful path;
- the early-data-specific compatibility gate still rejects only 0-RTT while the underlying PSK resumption remains authenticated;
- the prior client budget, exact ClientHello sizing, replay candidate, attempted/rejected observability, accepted-decision consistency, and skew-boundary fixes remain intact.

CI run 906 is green across the full workflow.

I left one **non-blocking** inline note for the later H3 integration: the model/codec/gate support `early_data_application_compat`, but `resumptionContextForTicket()` currently only populates the early transport snapshot. That does not block this TLS negotiation slice because the persistent seam is now complete, but the H3 slice must populate remembered application state before enabling H3 0-RTT.

Good to merge for slice 1.

PR #498, opened 2026-07-23
http: add early-data policy primitives

Jaruso:
## Summary
- add protocol-neutral HTTP early-data provenance, decision, and 425 planning primitives
- add route/config policy parsing for `early_data replay_safe` and `proxy_early_data rfc8470`
- normalize RFC 8470 proxy header handling by stripping inbound `Early-Data` and appending one canonical marker when enabled

Refs #367

## Testing

Note: this is slice 1 for #367 and intentionally does not close the umbrella issue.

### DIFF
--- a/src/http/early_data.zig
+++ b/src/http/early_data.zig
@@ -38,15 +38,17 @@ pub fn decide(inputs: Inputs) Decision {
     const exposed = inputs.replay_exposed or inputs.transport_early or inputs.inbound_marker;
     if (!exposed) return .ordinary;
 
-    if (inputs.action_class == .defer_until_handshake) return .defer_until_handshake;
+    if (inputs.action_class == .defer_until_handshake and inputs.transport_early and !inputs.inbound_marker) {
+        return .defer_until_handshake;
+    }
 
-    const replay_safe = inputs.method_safe or inputs.route_replay_safe;
-    if (!replay_safe) return .too_early;
+    if (!inputs.method_safe) return .too_early;
+    if (!inputs.route_replay_safe) return .too_early;
 
     return switch (inputs.action_class) {
         .local => .execute_local,
         .proxy => if (inputs.proxy_origin_rfc8470) .forward_rfc8470 else .too_early,
-        .defer_until_handshake => unreachable,
+        .defer_until_handshake => .too_early,
     };
 }
 
@@ -56,9 +58,7 @@ pub fn tooEarlyPlan() TooEarlyPlan {
 
 pub fn methodSafe(method: []const u8) bool {
     return std.ascii.eqlIgnoreCase(method, "GET") or
-        std.ascii.eqlIgnoreCase(method, "HEAD") or
-        std.ascii.eqlIgnoreCase(method, "OPTIONS") or
-        std.ascii.eqlIgnoreCase(method, "TRACE");
+        std.ascii.eqlIgnoreCase(method, "HEAD");
 }
 
 test "early data decision treats unexposed requests as ordinary" {
@@ -90,7 +90,7 @@ test "early data decision executes replay-safe local work" {
         .replay_exposed = false,
         .transport_early = true,
         .inbound_marker = false,
-        .method_safe = false,
+        .method_safe = true,
         .route_replay_safe = true,
         .action_class = .local,
         .proxy_origin_rfc8470 = false,
@@ -103,7 +103,7 @@ test "early data decision forwards only to RFC 8470 aware proxy origins" {
         .transport_early = false,
         .inbound_marker = true,
         .method_safe = true,
-        .route_replay_safe = false,
+        .route_replay_safe = true,
         .action_class = .proxy,
         .proxy_origin_rfc8470 = false,
     };
@@ -112,6 +112,27 @@ test "early data decision forwards only to RFC 8470 aware proxy origins" {
     var aware = base;
     aware.proxy_origin_rfc8470 = true;
     try std.testing.expectEqual(Decision.forward_rfc8470, decide(aware));
+}
+
+test "early data decision requires both method and route replay-safety gates" {
+    try std.testing.expectEqual(Decision.too_early, decide(.{
+        .replay_exposed = true,
+        .transport_early = false,
+        .inbound_marker = false,
+        .method_safe = false,
+        .route_replay_safe = true,
+        .action_class = .local,
+        .proxy_origin_rfc8470 = false,
+    }));
+    try std.testing.expectEqual(Decision.too_early, decide(.{
+        .replay_exposed = false,
+        .transport_early = false,
+        .inbound_marker = true,
+        .method_safe = true,
+        .route_replay_safe = false,
+        .action_class = .proxy,
+        .proxy_origin_rfc8470 = true,
+    }));
 }
 
 test "early data decision supports explicit handshake deferral" {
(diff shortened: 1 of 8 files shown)


src/http/early_data.zig
```diff
@@ -0,0 +1,135 @@
...
+
+pub fn decide(inputs: Inputs) Decision {
+    const exposed = inputs.replay_exposed or inputs.transport_early or inputs.inbound_marker;
+    if (!exposed) return .ordinary;
+
+    if (inputs.action_class == .defer_until_handshake) return .defer_until_handshake;
+
+    const replay_safe = inputs.method_safe or inputs.route_replay_safe;
+    if (!replay_safe) return .too_early;
+
```
Jaruso, line 47: **Blocking:** this must be an AND contract, not OR. #367 deliberately requires two independent gates: the method must be in Tardigrade's initial 0-RTT allowlist **and** the selected route must explicitly opt into `replay_safe`. As written, a GET on an `early_data off` route can pass, and a replay-safe POST can pass.

Please make the two rejections explicit so this invariant is hard to accidentally weaken:

```zig
if (!inputs.method_safe) return .too_early;
if (!inputs.route_replay_safe) return .too_early;
```

Then evaluate the action/origin capability. The current tests also need to flip: the `method_safe = false, route_replay_safe = true` local case must expect 425, and the `method_safe = true, route_replay_safe = false, proxy_origin_rfc8470 = true` case must also expect 425. Add positive cases only where both booleans are true.

src/http/early_data.zig
```diff
@@ -0,0 +1,135 @@
...
+    };
+}
+
+pub fn tooEarlyPlan() TooEarlyPlan {
+    return .{};
+}
+
+pub fn methodSafe(method: []const u8) bool {
+    return std.ascii.eqlIgnoreCase(method, "GET") or
+        std.ascii.eqlIgnoreCase(method, "HEAD") or
```
Jaruso: **Blocking:** this helper is encoding generic RFC "safe methods", but #367 intentionally narrows Tardigrade's initial 0-RTT execution policy to **GET and HEAD only**. OPTIONS must not become early-executable merely because RFC 9110 classifies it as safe, and TRACE is globally disabled in Tardigrade anyway.

Expected shape:

```zig
pub fn methodSafe(method: []const u8) bool {
    return std.ascii.eqlIgnoreCase(method, "GET") or
        std.ascii.eqlIgnoreCase(method, "HEAD");
}
```

Please add negative assertions for OPTIONS, TRACE, POST, PUT, PATCH, DELETE, and CONNECT so later protocol integrations inherit the exact policy.

src/http/request_context.zig
```diff
@@ -4,6 +4,19 @@ const Allocator = std.mem.Allocator;
...
+    inbound_marker: bool = false,
+
+    pub fn replayExposed(self: EarlyDataContext) bool {
+        return self.transport_early or self.inbound_marker;
+    }
+
+    pub fn mayRetryUpstream425(self: EarlyDataContext) bool {
+        return self.inbound_marker and !self.transport_early;
+    }
+};
```
Jaruso, line 18: **Blocking:** this is the opposite of the RFC 8470/#367 retry rule. An inbound `Early-Data` marker means replay exposure happened on a prior hop; an intermediary that receives an upstream 425 for that request must forward the 425 and **must not** locally retry it. The one local retry is for a request that was early only on this hop.

This should be:

```zig
pub fn mayRetryUpstream425(self: EarlyDataContext) bool {
    return self.transport_early and !self.inbound_marker;
}
```

Please cover all four truth-table states:

```text
transport=false marker=false -> false
transport=true  marker=false -> true
transport=false marker=true  -> false
transport=true  marker=true  -> false
```

The current test specifically asserts the unsafe prior-hop case as retryable, so that test needs to be corrected before slice 2 consumes this helper.

src/http/early_data.zig
```diff
@@ -0,0 +1,135 @@
...
+    message: []const u8 = "Request rejected because it was sent in early data and the route is not replay safe.",
+};
+
+pub fn decide(inputs: Inputs) Decision {
+    const exposed = inputs.replay_exposed or inputs.transport_early or inputs.inbound_marker;
+    if (!exposed) return .ordinary;
+
+    if (inputs.action_class == .defer_until_handshake) return .defer_until_handshake;
+
+    const replay_safe = inputs.method_safe or inputs.route_replay_safe;
```
Jaruso: **Blocking:** deferral is only a safe escape hatch for **current-hop** transport early data. An inbound `Early-Data` marker came from a previous hop; waiting for this connection's handshake cannot remove that replay exposure (explicit #367 invariant).

Do not return `.defer_until_handshake` merely from the action class. At minimum the deferrable case must exclude `inbound_marker`, e.g. conceptually:

```zig
if (inputs.action_class == .defer_until_handshake and
    inputs.transport_early and
    !inputs.inbound_marker)
{
    return .defer_until_handshake;
}
```

An inbound-marker case must continue into replay-safe route/method handling (or 425), not become executable after waiting. Add a regression test with `action_class = .defer_until_handshake`, `inbound_marker = true` proving it does **not** return `.defer_until_handshake`.

Jaruso (commented):
Blocking review — I would not merge slice 1 yet.

The overall slice boundary is good, and the config/header-normalization work looks solid. CI run 30034649149 is fully green across unit tests, builds, formatting, packaging, security, and the appliance profile. I did not find issues in the `Early-Data` presence helper, route/config parsing/defaults, raw-request-header stripping, canonical marker helper, or response-header stripping.

The blockers are concentrated in the protocol-neutral policy primitives, and they need to be corrected before later H1/H2/H3 slices build on this API:

1. `decide()` currently treats method eligibility **OR** explicit route opt-in as sufficient. #367 requires both independently: only GET/HEAD **and** an explicitly `replay_safe` route may execute/forward early. The existing tests currently encode the wrong behavior in both directions.
2. The initial #367 policy is GET/HEAD only, but `methodSafe()` also admits OPTIONS and TRACE. (TRACE is globally rejected elsewhere, but this primitive must still encode the actual early-data contract.)
3. `EarlyDataContext.mayRetryUpstream425()` is exactly reversed. Automatic retry is allowed only for current-hop transport early data with **no inbound `Early-Data` marker**. A prior-hop marker requires forwarding the upstream 425 downstream without a local retry.
4. `defer_until_handshake` is currently returned even when `inbound_marker` is true. Waiting for the current connection's handshake only addresses current-hop early data; it cannot erase replay exposure inherited from a previous hop. A prior-hop marker must continue through replay-safety/425 handling rather than being made safe by deferral.

I left inline comments with concrete corrections/tests. Once these policy semantics are fixed, this slice should be in good shape; I did not find another blocker in the other changed files.

Jaruso (commented):
Re-review of `005084215d6633646020648460f290f0980584ac`: all four blocking findings from my prior review are fixed, with regression coverage for each invariant. The decision helper now requires both GET/HEAD eligibility and explicit route replay-safety, the retry predicate is current-hop-only, and handshake deferral cannot erase prior-hop `Early-Data` exposure. I also rechecked the unchanged slice-1 plumbing (config defaults/validation, marker presence semantics, canonical request-header handling, and response stripping) and found no additional blocking issues. CI run 909 is green. No blocking findings remain for #367 slice 1; I consider this head ready to merge.

PR #499, opened 2026-07-23
http: wire H1 early-data preflight and proxy retry

Jaruso:
## Summary
- populate HTTP/1.1 early-data request context from inbound `Early-Data` header presence and the existing transport provenance seam
- add H1 early-data preflight before mirrors, auth/session work, mutable middleware, and route side effects
- emit real `425 Too Early` responses with `Cache-Control: no-store` for rejected replay-exposed H1 requests
- thread an explicit RFC 8470 forwarding marker through buffered proxy requests and force replay-exposed proxy work onto the buffered path
- add one bounded upstream-425 retry for current-hop-only early data after the downstream handshake barrier completes, while forwarding prior-hop-marker 425s

Refs #367

## Testing

Note: this is slice 2 for #367 and intentionally does not close the umbrella issue.


src/gateway_state.zig
```diff
@@ -2804,6 +2804,8 @@ pub const ConnectionSession = struct {
     proxy_protocol_checked: bool = false,
     proxy_client_ip_len: usize = 0,
     proxy_client_ip_buf: [64]u8 = undefined,
+    early_data_transport_early: bool = false,
```
Jaruso: **Blocking — this provenance is request-scoped, not connection-scoped.** `ConnectionSession` survives HTTP/1.1 keep-alive (and can carry pipelined bytes), while this bit is only cleared when the whole session is released back to the pool. If #366 sets this to `true` because request A consumed any early-data bytes, request B on the same connection will inherit `transport_early=true` even when B is entirely 1-RTT. A single boolean also cannot represent pipelined requests with different provenance or the required “request spans early + 1-RTT records” case.

Please do not establish a connection-wide sticky bit as the carrier contract. The handoff needs to be per request, conceptually something like:

```zig
const RequestTransportProvenance = struct {
    early: bool = false,
};

// #366's record reader accumulates provenance while bytes for *this request*
// are consumed, then the completed request receives the snapshot.
ctx.early_data.transport_early = request_provenance.early;
```

Because #366's real record-provenance descriptors are not on this branch yet, it is fine for production H1 to remain `false` in this slice and inject request-scoped provenance in tests. What we should avoid is a placeholder that later makes every keep-alive request early once one request was early.

Please add a regression for two sequential requests on one session (`A=early`, `B=ordinary`) proving B is not tainted, plus the eventual carrier-facing test for a request spanning early and 1-RTT bytes.

src/gateway_handlers.zig
```diff
@@ -82,6 +82,92 @@ pub fn resolveRoutePath(
...
+    return std.mem.eql(u8, path, "/transcripts") or
+        std.mem.startsWith(u8, path, "/transcripts/") or
+        std.mem.eql(u8, path, "/bearclaw/transcripts") or
+        std.mem.startsWith(u8, path, "/bearclaw/transcripts/");
+}
+
+pub fn earlyDataDecisionForRequest(
+    cfg: *const edge_config.EdgeConfig,
+    early_ctx: http.request_context.EarlyDataContext,
+    method: http.Method,
```
Jaruso, line 132: **Blocking — the preflight has no body/framing input, so body-bearing GET/HEAD is currently approved.** `http.Request.parse()` accepts a body whenever `Content-Length` or `Transfer-Encoding: chunked` is present regardless of method. Since this helper only receives `method` and `path`, an early `GET /proxy` with `Content-Length: 1` reaches `methodSafe(GET) == true` and can be forwarded as replay-safe. #367 explicitly keeps body-bearing requests unavailable for initial early execution.

Please pass a pure request-body fact into this decision rather than inferring it from `Method.hasRequestBody()`:

```zig
const has_body_framing = request.hasTransferEncoding() or
    (request.contentLength() orelse 0) != 0;

const decision = earlyDataDecisionForRequest(
    effective_cfg,
    ctx.early_data,
    request.method,
    request.uri.path,
    has_body_framing,
);
```

and fail closed for replay-exposed requests before location approval:

```zig
if (early_ctx.replayExposed() and has_body_framing) return .too_early;
```

Use framing/header facts rather than `request.body != null`, because the head-only streaming parser intentionally leaves `body = null`. Please cover early GET/HEAD with a positive Content-Length, chunked GET, and `Content-Length: 0` (which should not by itself make the request body-bearing).

src/gateway_handlers.zig
```diff
@@ -82,6 +82,92 @@ pub fn resolveRoutePath(
...
+    path: []const u8,
+) http.early_data.Decision {
+    if (!early_ctx.replayExposed()) return .ordinary;
```
Jaruso, line 136: **Blocking — matching mirror rules are still outside the safety decision.** This helper can return `.execute_local` / `.forward_rfc8470`, and `serveOneRequest()` then immediately calls `spawnMirrorRequests()`. That helper uses the configured mirror rule and sends a POST subrequest. Therefore a replay-safe early GET with a matching mirror rule can be replayed into duplicate mirror POST side effects, violating #367's explicit requirement that mirrors remain behind the preflight.

Please make mirror applicability part of the pure decision before this returns an executable outcome. Reuse the exact same matching predicates as the executor so the gate and execution cannot drift, e.g.:

```zig
fn hasMatchingMirrorRule(
    rules: []const edge_config.EdgeConfig.MirrorRule,
    method: []const u8,
    path: []const u8,
) bool {
    for (rules) |rule| {
        if (!http.rewrite.methodMatches(rule.method, method)) continue;
        if (!http.rewrite.regexMatches(rule.pattern, path)) continue;
        return true;
    }
    return false;
}

if (early_ctx.replayExposed() and
    hasMatchingMirrorRule(cfg.mirror_rules, method.toString(), path))
{
    return .too_early;
}
```

A conservative `cfg.mirror_rules.len != 0` rejection would also be safe, but matching only the applicable rule preserves more usable replay-safe routes. Add a test/probe that a rejected early request produces 425 **and zero mirror invocations**.

src/edge_gateway.zig
```diff
@@ -2460,12 +2460,21 @@ fn handleConnection(conn: anytype, session: *ConnectionSession, cfg: *const edge
     else
         connection_ip;
     const client_ip = http.request_context.extractClientIp(&request, effective_connection_ip);
+    var ctx = http.request_context.RequestContext.init(allocator, correlation_id, client_ip);
```
Jaruso: **Blocking — this context is created too late for the TRACE policy.** The global TRACE rejection immediately above this code returns 405 before `inbound_marker` or transport provenance is observed. That means `TRACE` + `Early-Data: 1` (or current-hop transport early provenance) bypasses #367's unsafe-early-method behavior and returns 405 rather than 425.

Keep the global TRACE ban for ordinary traffic, but populate provenance before making that policy decision:

```zig
const effective_connection_ip = ...;
const client_ip = extractClientIp(...);
var ctx = RequestContext.init(...);
ctx.early_data.transport_early = request_provenance;
ctx.early_data.inbound_marker = request.headers.hasEarlyDataMarker();
ctx.early_data.downstream_handshake = session.downstream_handshake;

if (request.method == .TRACE) {
    if (ctx.early_data.replayExposed()) {
        const status = try ghandlers.writeTooEarlyResponse(...);
        ghandlers.logAccessForRequest(state, &ctx, &request, status);
        return;
    }
    // existing ordinary 405 path
}
```

Missing Host can remain an earlier syntactic 400. This is specifically about TRACE being an application/method policy decision. Please test both ordinary TRACE -> 405 and replay-exposed TRACE -> 425.

Jaruso (commented):
Blocking review — I would not merge #499 at this head yet.

I reviewed the complete slice-2 diff against #367 and the slice-2 handoff, including the surrounding H1 request lifecycle, mirror/auth/middleware ordering, request parser/body semantics, buffered H1/H2 upstream paths, streaming fallback, retry accounting, response filtering, upstream health accounting, keep-alive session lifetime, and CI.

The core RFC 8470 proxy orchestration is in good shape: replay-exposed proxy requests are forced onto the buffered path; raw inbound `Early-Data` remains stripped and the outbound marker is injected explicitly; a prior-hop marker cannot trigger the local 425 retry; current-hop-only 425 retry is bounded to one and waits/drives the downstream handshake barrier; the retry and any later attempts run without `Early-Data`; and a final 425 is neither an upstream health success nor failure. The local synthesized 425 path also emits `Cache-Control: no-store`, request IDs/security headers, metrics, and access-log-compatible response accounting. CI run 30042350743 / #911 is fully green.

There are four blockers before later slices should build on this:

1. **Transport provenance is stored on `ConnectionSession`, but the contract is per request.** `ConnectionSession` survives keep-alive/pipelined requests, and this bit is only reset when the session is returned to the pool. Once a future #366 carrier sets it for one early request, later ordinary 1-RTT requests on the same connection can inherit `transport_early=true`. More importantly, one connection-wide bit cannot represent two pipelined requests with different record provenance or a request that spans early + 1-RTT records. Please make the HTTP handoff request-scoped rather than connection-sticky; until #366's descriptor carrier lands, keep production provenance false and inject request provenance in focused tests rather than establishing the wrong storage contract.

2. **The pure preflight does not account for matching mirror rules.** A replay-safe GET can be approved, after which `spawnMirrorRequests()` runs and emits a POST mirror request. A replay can therefore duplicate a side-effectful mirror even though #367 explicitly requires mirror work to remain behind the replay-safety gate. The preflight needs a pure `hasMatchingMirrorRule()`-style check using the same `methodMatches`/`regexMatches` predicates and must return 425 before spawning it. Add a regression test proving a rejected replay-exposed request performs zero mirror execution.

3. **Body-bearing GET/HEAD requests pass the early-data gate.** `methodSafe()` intentionally admits GET/HEAD, but Tardigrade's request parser accepts `Content-Length` or chunked bodies on those methods. #367 explicitly keeps body-bearing requests out of initial 0-RTT execution. Pass body-framing facts into the pure preflight and reject replay-exposed requests with a nonzero Content-Length or Transfer-Encoding before route approval. Please cover GET/HEAD body framing, `Content-Length: 0`, and chunked GET in tests.

4. **TRACE is rejected before early-data provenance is populated.** Ordinary TRACE should stay globally 405, but replay-exposed TRACE is an unsafe early request and #367's initial decision table requires 425. Build/populate the request early-data context before the TRACE policy check, then select 425 for replay-exposed TRACE and preserve the existing 405 for ordinary TRACE.

I also specifically checked for additional issues in the special 425 loop, header normalization, response `Early-Data` filtering, 5xx retry interaction, stale-connection retries, and final health accounting and did not find another blocker there.

Please add integration-level coverage around the fixed H1 preflight, not only decision-helper tests: rejected requests should demonstrably avoid mirror/auth/rate-limit/upstream side effects, and the current-hop 425 case should prove first attempt has exactly one `Early-Data: 1`, second attempt has none, and two consecutive 425 responses result in exactly two upstream deliveries and a downstream 425.

This remains a slice-2 review only; H3 remembered SETTINGS and H2 safe deferral are still later #367 work.

Jaruso (commented):
Follow-up review — the four code blockers from review #4768167790 are fixed, and I did a fresh pass over the complete slice-2 diff and surrounding H1/proxy runtime. I did not find a new code defect in the corrected behavior.

I still would **not merge this head yet**, because the implementation is missing behavioral tests that are explicit acceptance criteria in #367 and were also called out in the first review.

What is now correct:

- the incorrect connection-scoped `transport_early` carrier was removed; production H1 stays conservatively false until #366 provides the actual request-scoped provenance carrier;
- replay-exposed GET/HEAD with non-zero Content-Length or Transfer-Encoding is rejected before execution, while `Content-Length: 0` remains eligible;
- matching mirror rules participate in the pure preflight using the same method/regex predicates as mirror execution;
- replay-exposed TRACE now produces 425 while ordinary TRACE remains 405;
- replay-exposed proxy requests are forced onto the buffered path;
- canonical `Early-Data: 1` injection is explicit and the retry path drops the marker;
- prior-hop 425 is never locally retried; current-hop-only 425 is bounded to one special retry after handshake completion;
- final 425 is not counted as upstream success/failure, and ordinary 5xx behavior remains unchanged.

The remaining blocker is **AC proof**, not another logic bug:

1. #367 explicitly requires side-effect counters/stubs proving a rejected 425 path performs zero mirror/auth/rate-limit/upstream/handler calls. The new tests verify the pure decision helper, but they do not exercise the H1 orchestration boundary where those side effects are sequenced.
2. #367 explicitly requires the upstream-425 state machine to be exercised end-to-end: first eligible request carries exactly one `Early-Data: 1`; first 425 triggers exactly one post-handshake retry; retry carries no `Early-Data`; `425 -> 200` returns the 200; `425 -> 425` performs exactly two upstream deliveries and forwards the second 425; inbound-marker requests forward 425 with zero local retries; origin capability off makes zero early upstream calls. The current `shouldRetryEarlyUpstream425()` unit test only proves the predicate, not orchestration/header behavior.

Please add focused state-machine/integration harnesses around these boundaries. They do not need a live TLS 0-RTT carrier: inject `RequestContext.early_data` plus a test `DownstreamHandshakeBarrier`, and use counting/fake side-effect/upstream seams. That keeps #366 ownership intact while actually locking the #367 semantics.

CI run #912 is green. Once these required behavioral tests are present and green, I expect this slice to be mergeable; I found no additional implementation blocker in this pass.

src/edge_gateway.zig
```diff
@@ -2606,6 +2652,15 @@ fn handleConnection(conn: anytype, session: *ConnectionSession, cfg: *const edge
         effective_cfg.named_locations,
     );
 
+    switch (ghandlers.earlyDataDecisionForRequest(effective_cfg, ctx.early_data, request.method, request.uri.path, earlyDataHasReplayExposedBodyFraming(&request))) {
+        .ordinary, .execute_local, .forward_rfc8470 => {},
+        .too_early, .defer_until_handshake => {
```
Jaruso: **Blocking AC gap — this ordering is now correct in code, but #367 explicitly requires behavioral proof at this boundary.** The issue says the tests must use counters/stubs to prove a 425 path performs **zero** mirror/auth/rate-limit/upstream/handler calls. The current additions test `earlyDataDecisionForRequest()` directly, which would not catch a future refactor that accidentally moves one side effect above this switch.

Please add a focused H1 orchestration test/harness that drives a replay-exposed rejected request through this sequencing point and observes counters, e.g. conceptually:

```zig
var effects = SideEffectProbe{};
const ctx = EarlyDataContext{ .inbound_marker = true };

const status = try serveH1RequestWithProbe(..., ctx, &effects);
try expectEqual(@as(u16, 425), status);
try expectEqual(@as(usize, 0), effects.mirror_calls);
try expectEqual(@as(usize, 0), effects.auth_calls);
try expectEqual(@as(usize, 0), effects.rate_limit_mutations);
try expectEqual(@as(usize, 0), effects.upstream_calls);
try expectEqual(@as(usize, 0), effects.handler_calls);
```

The exact seam can be smaller than `handleConnection`; the important part is that the test covers the **ordering/orchestration**, not only the pure decision function. This was an explicit #367 AC and was called out in the previous review body.

src/gateway_proxy_runtime.zig
```diff
@@ -730,6 +743,29 @@ pub fn handleLocationProxyPass(
         };
         const upstream_ttfb_ms = http.event_loop.monotonicMs() - upstream_start_ms;
         state.metricsRecordProxyBufferedRequest(result.bodyLen(), upstream_ttfb_ms);
+        if (result.statusCode() == @intFromEnum(http.Status.too_early) and shouldRetryEarlyUpstream425(
+            ctx.early_data,
+            first_attempt_forward_early_data,
```
Jaruso: **Blocking AC gap — the special 425 state machine is implemented, but only the predicate is tested.** #367 requires tests of the actual orchestration here, not just `shouldRetryEarlyUpstream425()`.

Please add a deterministic fake/counting upstream executor (or extract the loop behind an injected attempt function) and prove at least:

```text
current-hop-only + RFC8470:
  attempt 1 headers -> exactly one Early-Data: 1
  attempt 1 response -> 425
  handshake barrier invoked/completes
  attempt 2 headers -> no Early-Data
  attempt 2 response -> 200
  downstream -> 200
  total upstream deliveries -> 2

same, but attempt 2 -> 425:
  downstream -> 425
  total upstream deliveries -> exactly 2 (no third request)

inbound_marker=true + upstream 425:
  total upstream deliveries -> 1
  downstream -> 425

proxy_early_data=off:
  no early upstream delivery
```

This also needs to assert the canonical marker behavior through the real request-building path, because a correct retry predicate would not catch a regression where attempt 2 accidentally reuses `Early-Data: 1`. A small injected-attempt seam is enough; no live TLS/QUIC carrier is necessary.

src/edge_gateway.zig
```diff
@@ -2228,6 +2228,60 @@ fn streamingUploadEligibilityBeforeBodyRead(
...
+
+const H1PreflightSideEffectProbe = struct {
+    mirror_calls: usize = 0,
+    auth_calls: usize = 0,
+    rate_limit_mutations: usize = 0,
+    upstream_calls: usize = 0,
+    handler_calls: usize = 0,
+};
+
+fn runH1PreflightSideEffectProbe(
```
Jaruso: **Still blocking the assurance AC — this is a second orchestration, not a probe of the production one.** The counters look right, but `runH1PreflightSideEffectProbe()` has its own `switch (earlyDataPreflightDecisionForH1(...))` and then synthetic increments. The real ordering is separately written later in `handleConnection()`. If someone accidentally moves `spawnMirrorRequests()` or `primeRequestAuthContext()` above the real preflight, this test still passes unchanged.

Please make the test share the production sequencing. Two reasonable approaches:

```zig
const H1Effects = struct {
    ctx: *anyopaque,
    mirror_fn: *const fn (*anyopaque, ...) anyerror!void,
    auth_fn: *const fn (*anyopaque, ...) anyerror!void,
    middleware_fn: *const fn (*anyopaque, ...) anyerror!bool,
    route_fn: *const fn (*anyopaque, ...) anyerror!u16,
};

fn executeH1PostParse(..., effects: H1Effects) !u16 {
    switch (earlyDataPreflightDecisionForH1(...)) {
        .too_early, .defer_until_handshake => return 425,
        else => {},
    }
    try effects.mirror_fn(...);
    try effects.auth_fn(...);
    ...
}
```

Production passes the real hooks; the test passes counting hooks. Or drive the existing H1 integration harness and use real counting mirror/auth/upstream endpoints. The key is that the test must fail if the *production* side-effect ordering regresses.

src/gateway_proxy_runtime.zig
```diff
@@ -824,6 +861,93 @@ fn shouldRetryStaleUpstreamConnection(
...
+        if (std.ascii.eqlIgnoreCase(header.name, "Early-Data") and
+            std.mem.eql(u8, header.value, "1"))
+        {
+            count += 1;
+        }
+    }
+    try counts.append(count);
+}
+
+fn runEarly425RetryHarness(
```
Jaruso, line 1224: **Still blocking the retry assurance AC — this harness duplicates the state machine instead of exercising it.** It owns separate `retry_used`, `retry_as_ordinary`, `extra_attempts`, and delivery accounting and calls `appendCanonicalEarlyDataHeader()` directly instead of going through `handleLocationProxyPass()` -> `executeBufferedDataPlaneProxyRequest()` -> the real request builder.

That is not just theoretical: the production loop currently has a configured-retry accounting bug that this harness cannot see because the harness hard-codes `1 + extra_attempts` and has no `max_attempts` behavior at all.

Please extract the retry loop/state transition behind an injected attempt executor and use that same function from production, e.g. conceptually:

```zig
const AttemptFn = *const fn (ctx: *anyopaque, forward_early_data: bool) anyerror!DataPlaneProxyResponse;

fn executeProxyAttempts(
    ...,
    attempt_fn: AttemptFn,
    attempt_ctx: *anyopaque,
) !DataPlaneProxyResponse {
    // the one real retry state machine lives here
}
```

The test executor can record the actual `forward_early_data` value and return scripted 425/500/200 responses. Production's executor calls `executeBufferedDataPlaneProxyRequest()`. That makes the tests capable of catching drift in attempt budgets and marker suppression.

src/gateway_proxy_runtime.zig
```diff
@@ -730,6 +744,29 @@ pub fn handleLocationProxyPass(
...
+        )) {
+            ctx.early_data.waitOrDriveDownstreamHandshake() catch |err| {
+                state.logger.warn(correlation_id, "could not complete downstream handshake before early-data 425 retry: {}", .{err});
+                break result;
+            };
+            if (!ctx.early_data.downstreamHandshakeComplete()) {
+                break result;
+            }
+            early_425_retry_used = true;
+            early_425_extra_attempts = 1;
```
Jaruso: **Blocking — the RFC-8470 retry is added to the loop bound but still consumes the normal configured retry index.** #367 requires the semantic 425 retry to have a separate max of one while existing stale/5xx attempt accounting remains unchanged. After this sets `early_425_extra_attempts = 1`, `attempt` still increments. Both the transport-error retry check and the 5xx retry check later use raw `attempt + 1 < max_attempts`.

Concrete failure with `upstream_retry_attempts = 2`:

```text
attempt=0 -> 425
semantic retry allowed; extra_attempts=1
attempt=1 -> 500
1 + 1 < 2 == false
```

So the configured normal retry that should still be available after the extra semantic retry is lost.

A minimal correction while preserving the existing stale-connection behavior is to compute the configured-attempt index with the semantic extra removed:

```zig
const configured_attempt_index = attempt - early_425_extra_attempts;

if (configured_attempt_index + 1 < max_attempts) {
    // ordinary error retry
}

if (result.statusCode() >= 500 and
    configured_attempt_index + 1 < max_attempts)
{
    // ordinary 5xx retry
}
```

Use that for every `max_attempts` comparison/logging that is supposed to reflect operator-configured attempts. Please add the regression `425 -> semantic retry 500 -> configured retry 200` with `max_attempts=2`; expected total upstream deliveries are 3 and only the first carries `Early-Data: 1`.

Jaruso (commented):
Blocking follow-up on head `a8cd46a36a6ce65ad684d70745f603ef32ba5875`.

I re-ran the full slice-2 review, including the new assurance commit, the real H1 ordering, the real proxy retry loop, marker construction, handshake barrier behavior, retry/health accounting, and CI.

The four original implementation bugs remain fixed, and CI #913 is green. The new tests cover the requested scenarios at a model level, but they still do not close the two assurance gaps because both new harnesses duplicate the production sequencing rather than exercising it. That distinction is now material: the production proxy loop has a retry-budget bug that the new `runEarly425RetryHarness()` cannot detect.

### Blocking findings

1. **H1 side-effect proof is still a parallel model.** `runH1PreflightSideEffectProbe()` runs the same pure decision helper and then increments synthetic counters. It does not execute the production `handleConnection()` sequencing where `spawnMirrorRequests()`, `primeRequestAuthContext()`, `runMiddlewarePipeline()`, and `routeRequest()` actually live. Moving one of those real side effects above the production preflight would leave this test green. Please either extract the real post-parse orchestration behind injectable hooks and test that shared function, or add a focused integration test that drives the real H1 path with counting mirror/auth/upstream/middleware seams.

2. **The 425 retry harness still duplicates the state machine and request-header decision.** It has its own `retry_used`, `retry_as_ordinary`, `extra_attempts`, delivery counter, and direct call to `appendCanonicalEarlyDataHeader()`. It never executes `handleLocationProxyPass()` or `executeBufferedDataPlaneProxyRequest()`. Consequently it cannot prove the production loop's attempt accounting or the real request-building path.

3. **Newly confirmed production bug: the semantic 425 retry consumes normal retry budget.** #367 says to keep existing stale/5xx accounting unchanged and give the RFC-8470 retry a separate max of one. The loop adds `early_425_extra_attempts` to its bound, but the ordinary error and 5xx retry checks still compare raw `attempt + 1` to `max_attempts`. Example with `upstream_retry_attempts=2`: first delivery -> 425, semantic retry -> 500. At that point `attempt == 1`, so `attempt + 1 < max_attempts` is false and the configured normal 5xx retry is lost. The semantic retry needs to be excluded from the configured-attempt index for both the ordinary error retry check and the 5xx retry check.

A minimal shape that preserves the pre-existing stale-connection accounting is:

```zig
const configured_attempt_index = attempt - early_425_extra_attempts;

// ordinary transport/error retry
if (configured_attempt_index + 1 < max_attempts) {
    continue;
}

// 5xx retry
if (result.statusCode() >= 500 and
    configured_attempt_index + 1 < max_attempts)
{
    ...
    continue;
}
```

Because `early_425_extra_attempts` only becomes 1 after the first 425 is discarded, it offsets exactly the one semantic retry while leaving the existing stale retry behavior untouched. A separate explicit `configured_attempts_used` counter is also fine and may be clearer.

Please add a production-shared regression such as:

```text
upstream_retry_attempts = 2
current-hop early GET
attempt 1 -> 425
semantic retry -> 500
normal configured retry -> 200
final downstream status -> 200
total delivered upstream requests -> 3
headers -> [Early-Data: 1, no Early-Data, no Early-Data]
```

and an error-path equivalent if practical. That test should run through the same retry-state function used by production rather than a second model.

No other new blocker surfaced in the code paths I rechecked. The earlier four implementation threads should remain resolved; the two assurance threads remain open until the tests share production orchestration rather than reproducing it.

src/edge_gateway.zig
```diff
@@ -2228,6 +2228,174 @@ fn streamingUploadEligibilityBeforeBodyRead(
...
+    }
+
+    fn route(
+        self: H1ProductionPostPreflightHooks,
+        conn: anytype,
+        allocator: std.mem.Allocator,
+        cfg: *const edge_config.EdgeConfig,
+        state: *GatewayState,
+        ctx: *http.request_context.RequestContext,
+        request: *const http.Request,
```
Jaruso: **Blocking / compile regression — this must remain mutable through the production route hook.** `gateway_handlers.routeRequest()` takes `request: *http.Request`, not `*const http.Request`, because route execution can mutate the request (for example a location `.rewrite` changes `request.uri.path`). This production hook currently tries to pass a const pointer into that mutable API, which Zig cannot implicitly do.

Please make the shared orchestration carry the mutable request pointer through the routing seam:

```zig
fn executeH1PostPreflightOrchestration(
    ...,
    request: *http.Request,
    ...
) !H1PostPreflightOutcome { ... }

fn route(
    self: H1ProductionPostPreflightHooks,
    ...,
    request: *http.Request,
    ...
) !u16 {
    return ghandlers.routeRequest(..., request, ...);
}
```

The read-only preflight/mirror/middleware helpers can still receive the mutable pointer and coerce it to `*const` where appropriate. Update the counting hook's route signature similarly. This also preserves the semantics of the original pre-refactor H1 path instead of accidentally making request mutation impossible. CI #914's broad build failures are consistent with this generic production-only instantiation failure.

src/gateway_proxy_runtime.zig
```diff
@@ -824,6 +857,121 @@ fn shouldRetryStaleUpstreamConnection(
...
+        if (std.ascii.eqlIgnoreCase(header.name, "Early-Data") and
+            std.mem.eql(u8, header.value, "1"))
+        {
+            count += 1;
+        }
+    }
+    try counts.append(count);
+}
+
+fn runEarly425RetryHarness(
```
Jaruso, line 1224: **Still blocking the retry assurance AC.** Sharing `BufferedProxyRetryState` fixed the real budget bug and is a good improvement, but this function still implements a second retry loop. It separately decides branch order, calls `consumeEarly425Retry()`, handles 5xx continuation, and invokes `recordCanonicalEarlyDataHeaderCount()` directly. Production separately implements all of those choices around `executeBufferedDataPlaneProxyRequest()`.

That means this test remains unable to detect production drift in the exact behavior #367 asks us to prove. For example, all tests here would still pass if production changed to:

```zig
// regression in production only
executeBufferedDataPlaneProxyRequest(..., first_attempt_forward_early_data, ...);
```

instead of `retry_state.forwardEarlyData(...)`, causing the semantic retry to incorrectly resend `Early-Data: 1`.

Please put the *one real attempt state machine* behind an injected executor, roughly:

```zig
fn runBufferedProxyAttempts(
    ...,
    attempt_executor: anytype,
) !AttemptResult {
    // production while-loop, 425 branch, 5xx branch, and retry state live here
}
```

Production's executor calls `executeBufferedDataPlaneProxyRequest()`; the test executor records `forward_early_data` and returns scripted 425/500/200 responses. Then the existing excellent scenario assertions (`425->200`, `425->425`, inbound marker, and `425->500->200`) actually protect the production loop instead of a parallel copy.

Jaruso (commented):
Blocking follow-up review on `e36be4238cf6f86b045c4fff4fba640d696d159e`.

This iteration fixes two important things from the previous review:

- The H1 side-effect assurance now shares `executeH1PostPreflightOrchestration()` with the production `handleConnection()` path. That satisfies the ordering-proof concern; I resolved both older H1 assurance threads.
- `BufferedProxyRetryState` fixes the semantic-425/configured-retry accounting bug. The new `425 -> 500 -> 200` regression with `max_attempts=2` correctly expects three deliveries and marker counts `{1, 0, 0}`. I resolved the retry-budget bug thread.

I found two remaining blockers:

1. **The new H1 orchestration extraction does not compile in the production instantiation.** `executeH1PostPreflightOrchestration()` and `H1ProductionPostPreflightHooks.route()` accept `*const http.Request`, but `gateway_handlers.routeRequest()` intentionally requires `*http.Request` because downstream routing can mutate the request (for example location rewrite handling). Zig cannot discard const here. This matches CI #914: unit tests that instantiate the counting hooks pass on multiple platforms, while the actual gateway builds fail broadly. Make the orchestration's request pointer mutable through the route hook; helper hooks that only read it may still accept/coerce to const.

2. **The proxy retry assurance is improved but still tests a second retry loop rather than the production one.** `runEarly425RetryHarness()` now shares `BufferedProxyRetryState`, so it catches state-accounting regressions, but it still owns its own `while`, retry branch ordering, scripted response handling, and direct canonical-header counting. Production separately owns the real loop and separately passes `retry_state.forwardEarlyData(...)` into `executeBufferedDataPlaneProxyRequest()`. Therefore the acceptance test still cannot catch a production regression such as failing to call `consumeEarly425Retry()`, moving the 5xx branch ahead of the 425 branch, or passing the original marker flag to the real executor on attempt two. This is the remaining part of the earlier assurance thread: extract the actual attempt loop behind an injected attempt executor, or drive `handleLocationProxyPass()` with a deterministic fake upstream.

I rechecked the actual retry arithmetic after the state change and did not find another semantic issue there. The first early attempt, one semantic retry, and remaining configured 5xx budget now compose correctly; subsequent attempts after the handshake remain ordinary/no-marker. Prior-hop marker behavior and 425 health accounting also remain correct.

CI run #914 is currently failing, so this head is not mergeable regardless of the assurance item. Once the mutable-request compile regression and the shared production retry-loop test seam are fixed, I do not currently have another slice-2 blocker.

src/gateway_proxy_runtime.zig
```diff
@@ -824,6 +778,381 @@ fn shouldRetryStaleUpstreamConnection(
...
+        var result = resp;
+        try attempt_executor.onBufferedResponse(&result);
+        if (result.statusCode() == @intFromEnum(http.Status.too_early) and shouldRetryEarlyUpstream425(
+            early_ctx.*,
+            first_attempt_forward_early_data,
+            retry_state.early_425_retry_used,
+            method,
+            matched_block,
+        )) {
+            try early_ctx.waitOrDriveDownstreamHandshake();
```
Jaruso: **Blocking — a handshake-barrier failure must forward the existing 425, not escape the proxy path.** Before the shared-loop extraction this call was deliberately wrapped in `catch`: Tardigrade logged the failure and `break result`, which forwarded the upstream 425. That matches #367's rule that if the carrier cannot safely wait/drive the downstream handshake, the gateway may simply forward 425.

With `try` here, any barrier error now bubbles out of `runBufferedProxyAttempts()` / `handleLocationProxyPass()` instead. `result` is also an owned buffered response with no defer on this path, so this error path can leak its response allocation rather than forwarding/deinitializing it.

Please keep this behavior inside the one shared state machine, e.g. add an executor hook for logging and return the response unchanged:

```zig
early_ctx.waitOrDriveDownstreamHandshake() catch |err| {
    try attempt_executor.onEarly425HandshakeFailure(err);
    return .{ .response = result };
};
```

Production can log the warning in that hook; the scripted executor can count it. Add a regression with a barrier whose `waitOrDrive` returns an error and assert: one upstream delivery, downstream status 425, no semantic retry, and normal response cleanup.

src/gateway_proxy_runtime.zig
```diff
@@ -824,6 +778,381 @@ fn shouldRetryStaleUpstreamConnection(
...
+        return resp;
+    }
+
+    fn onBufferedResponse(
+        self: *ProductionBufferedProxyAttemptExecutor,
+        result: *DataPlaneProxyResponse,
+    ) !void {
+        const upstream_ttfb_ms = http.event_loop.monotonicMs() - self.budget_start_ms;
+        self.state.metricsRecordProxyBufferedRequest(result.bodyLen(), upstream_ttfb_ms);
+    }
```
Jaruso, line 975: **Blocking regression — this changed the meaning of the buffered upstream TTFB metric.** The pre-refactor loop captured `upstream_start_ms` immediately before each `executeBufferedDataPlaneProxyRequest()` and recorded `now - upstream_start_ms`. This now subtracts `budget_start_ms`, which was captured once before streaming selection / the entire buffered retry sequence.

After a retry, the reported TTFB therefore includes the previous attempt(s), and after a semantic 425 it also includes the downstream-handshake wait. A sequence like `425 (40ms) -> handshake wait (100ms) -> 200 (30ms)` will report roughly 170ms for the second upstream attempt instead of ~30ms.

Please preserve an attempt-local timestamp in the production executor:

```zig
last_attempt_start_ms: u64 = 0,

fn execute(...) !DataPlaneProxyResponse {
    ...
    self.state.recordUpstreamAttemptStart(self.selection_base_url);
    self.last_attempt_start_ms = http.event_loop.monotonicMs();
    const resp = executeBufferedDataPlaneProxyRequest(...);
    ...
}

fn onBufferedResponse(self: *Self, result: *DataPlaneProxyResponse) !void {
    const ttfb = http.event_loop.monotonicMs() - self.last_attempt_start_ms;
    self.state.metricsRecordProxyBufferedRequest(result.bodyLen(), ttfb);
}
```

This keeps the extraction behavior-neutral for existing observability.

src/gateway_proxy_runtime.zig
```diff
@@ -824,6 +778,381 @@ fn shouldRetryStaleUpstreamConnection(
...
+                stale_conn_retries,
+                max_stale_conn_retries,
+                upstream_retry_idempotent_only,
+            )) {
+                stale_conn_retries += 1;
+                try attempt_executor.onStaleConnectionRetry(stale_conn_retries, max_stale_conn_retries);
+                continue;
+            }
+            if (err == error.UpstreamAtCapacity) return .upstream_at_capacity;
+            if (err == error.RequestCancelled) return .request_cancelled;
```
Jaruso: **Blocking — the extraction changed existing non-425 error/health accounting, and it also conflates local budget exhaustion with an upstream failure.**

Two concrete differences from the pre-extraction loop:

1. Previously `state.recordUpstreamFailure(...)` ran **before** the `RequestCancelled` branch. This now returns `.request_cancelled` before `onTerminalAttemptError()`, so the same cancellation no longer affects upstream health.
2. `ProductionBufferedProxyAttemptExecutor.execute()` now runs `perAttemptTimeoutMs()` inside the error caught by this state machine. When the *overall local retry budget* is already exhausted, `perAttemptTimeoutMs()` returns `error.Timeout` before an upstream attempt starts. This branch treats that as an origin attempt error, calls `onTerminalAttemptError()` (marking health failure), and may `continue` through remaining configured retries even though every next iteration will immediately hit the same expired local budget.

Please preserve the old semantics by separating a pre-attempt budget-expired outcome from an actual upstream attempt error, and keep `RequestCancelled` health behavior unchanged unless intentionally changed outside this story. One workable shape is:

```zig
// production executor
fn perAttemptTimeoutMs(...) error{ProxyBudgetExhausted}!u32 { ... }

// shared loop
const resp = attempt_executor.execute(...) catch |err| {
    if (err == error.ProxyBudgetExhausted)
        return .retry_budget_exhausted;

    if (shouldRetryStaleUpstreamConnection(...)) { ... }
    if (err == error.UpstreamAtCapacity) return .upstream_at_capacity;

    // Matches the old placement of recordUpstreamFailure().
    try attempt_executor.onTerminalAttemptError(err);
    if (err == error.RequestCancelled) return .request_cancelled;
    ...
};
```

Then map `.retry_budget_exhausted` back to the existing timeout behavior without marking the origin or spinning through unused retries. Please add focused regressions for both RequestCancelled accounting and an already-expired total retry budget.

Jaruso (commented):
Blocking follow-up review on the latest head `b6fbd3035f29dfb5c285c87b480ae2bdfe913b2c`.

The two blockers from review #4768980024 are fixed:

- the H1 orchestration now carries a mutable `*http.Request` through the production route hook, and the gateway builds compile again;
- production and tests now share the same `runBufferedProxyAttempts()` state machine behind injected attempt executors. This is the test seam I was asking for, and I resolved the prior retry-assurance threads.

I re-ran the full slice-2 regression pass over the extracted retry loop rather than stopping at those fixes. I found three behavior regressions introduced by the extraction that need correction before merge:

1. **Handshake-barrier failure now escapes instead of forwarding the original 425.** The old production loop explicitly caught `waitOrDriveDownstreamHandshake()` errors, logged them, and forwarded the existing 425. The shared loop now uses `try`, so a carrier/barrier error propagates out of `handleLocationProxyPass()` and also strands the owned buffered response. #367 explicitly allows forwarding 425 when the carrier cannot safely wait/drive. Please restore that fail-safe behavior and add a failing-barrier regression.

2. **Buffered upstream TTFB changed from per-attempt timing to cumulative request-budget timing.** Before this refactor, every attempt captured `upstream_start_ms` immediately before `executeBufferedDataPlaneProxyRequest()` and recorded `now - upstream_start_ms`. `ProductionBufferedProxyAttemptExecutor.onBufferedResponse()` now records `now - budget_start_ms`, so retries include all prior attempt latency plus the 425 handshake wait. That silently changes an existing metric and makes later-attempt TTFB increasingly inflated. Preserve an attempt-local start timestamp in the production executor.

3. **The extraction changed non-425 error/health semantics.** `RequestCancelled` used to pass through `recordUpstreamFailure()` before terminating; it now returns `.request_cancelled` before `onTerminalAttemptError()`. In the other direction, a locally exhausted `upstream_timeout_budget_ms` now arises inside `execute()` as `error.Timeout`, gets treated as an upstream attempt failure, can mark the origin unhealthy, and can burn through remaining configured retries without issuing any upstream request. Previously the budget check returned immediately before `recordUpstreamAttemptStart()`. Please distinguish local pre-attempt budget exhaustion from actual upstream attempt errors and preserve the existing RequestCancelled health behavior unless this PR intentionally changes it with separate justification/tests.

The early-data-specific state machine otherwise looks correct on this pass: marker suppression after the semantic retry, the one-retry bound, prior-hop no-retry behavior, and independence from configured 5xx retry budget all hold in the shared production/test loop.

CI #915 is also not fully green at the time of this review: builds, format, TLS audit, and most test jobs are passing, but `Test (macos-14)` failed in the unit-test step. Please get the head fully green (or establish that failure as an unrelated known flake) before merge.

This remains a #367 slice-2 review only; slices 3/4 are not expected here.

Jaruso (commented):
Ready to merge for #367 slice 2 at `92ac1473410bbede00180475984d39db1710c1e8`.

I re-reviewed the latest delta and then re-ran the full slice-2 invariants across the H1 preflight, RFC 8470 marker propagation, buffered proxy retry state machine, response ownership, health accounting, retry-budget accounting, and ordinary proxy behavior.

The three regressions from my previous review are fixed:

- handshake-barrier failure now logs and forwards the original upstream 425 rather than escaping the proxy path; the regression test proves one delivery / downstream 425 / no retry;
- buffered proxy TTFB is again attempt-local via `last_attempt_start_ms`, so retries and the downstream handshake wait do not contaminate the next upstream-attempt latency;
- `RequestCancelled` once again records the upstream failure before terminating, while local overall retry-budget exhaustion is now a distinct `ProxyBudgetExhausted` outcome that neither penalizes origin health nor spins through unused configured retries.

The earlier assurance concern is also fully addressed: production and tests now execute the same `runBufferedProxyAttempts()` state machine. The scripted executor only supplies attempts/observations, while the production executor threads the shared state's `forward_early_data` decision into `executeBufferedDataPlaneProxyRequest()`. The tests cover 425->200, 425->425, inbound-marker no-retry, handshake-drive failure, and 425->500->200 with the semantic retry independent from configured retry budget.

All prior blocking threads are resolved. I did not find another correctness, replay-safety, ownership, retry-accounting, health-accounting, or ordinary-H1 regression in this pass.

CI run #916 is green.

This verdict is specifically for the advertised #367 slice-2 scope. H3 remembered SETTINGS / H3 integration and the deliberately safe H2 policy remain later slices, and #366 still owns the real per-request TLS/QUIC transport-provenance carrier.

PR #500, opened 2026-07-24
http3: add remembered SETTINGS early-data compatibility

Jaruso:
## Summary
- add canonical HTTP/3 remembered SETTINGS snapshot encode/decode and directional 0-RTT compatibility checks
- stamp early-capable tickets with H3 application compatibility and expose the client-side remembered application snapshot for the planned identity-0 early-data attempt
- wire native H3 runtime SETTINGS emission/application compatibility into the QUIC TLS backend wrapper

## Validation
- zig build --summary all --error-style verbose

Part of #367 slice 3.

### DIFF
--- a/src/http3/conn.zig
+++ b/src/http3/conn.zig
@@ -78,6 +78,7 @@ pub fn Conn(comptime Transport: type) type {
         peer_qpack_encoder: ?u64 = null,
         peer_qpack_decoder: ?u64 = null,
         peer_control_view: frame.ControlStream = .{},
+        local_settings: frame.Settings = .{},
         /// Peer uni streams whose type varint has not fully arrived yet, and
         /// unknown-type streams we drain and ignore.
         pending_uni: std.AutoHashMap(u64, PendingUni),
@@ -116,9 +117,14 @@ pub fn Conn(comptime Transport: type) type {
         };
 
         pub fn init(allocator: std.mem.Allocator, role: Role) Self {
+            return initWithSettings(allocator, role, .{});
+        }
+
+        pub fn initWithSettings(allocator: std.mem.Allocator, role: Role, settings: frame.Settings) Self {
             return .{
                 .allocator = allocator,
                 .role = role,
+                .local_settings = settings,
                 .pending_uni = std.AutoHashMap(u64, PendingUni).init(allocator),
                 .requests = std.AutoHashMap(u64, *ServerRequest).init(allocator),
                 .responses = std.AutoHashMap(u64, *ClientResponse).init(allocator),
@@ -190,13 +196,37 @@ pub fn Conn(comptime Transport: type) type {
             var bytes: [64]u8 = undefined;
             var len: usize = 0;
             len += (try frame.encodeStreamType(.control, bytes[len..])).len;
-            var settings_payload: [16]u8 = undefined;
-            // Static-table-only QPACK: all our SETTINGS are the RFC defaults,
-            // so the frame is legitimately empty.
-            const settings = try frame.encodeSettings(&.{}, &settings_payload);
+            var settings_payload: [64]u8 = undefined;
+            var settings_entries: [5]frame.Setting = undefined;
+            const settings = try encodeLocalSettings(self.local_settings, &settings_entries, &settings_payload);
             len += (try frame.encodeKnownFrame(.settings, settings, bytes[len..])).len;
             _ = try transport.writeStream(control, bytes[0..len], false);
             self.control_out = control;
+        }
+
+        fn encodeLocalSettings(settings: frame.Settings, entries: *[5]frame.Setting, out: []u8) ![]u8 {
+            var count: usize = 0;
+            if (settings.qpack_max_table_capacity != 0) {
+                entries[count] = .{ .id = .qpack_max_table_capacity, .id_value = 0x01, .value = settings.qpack_max_table_capacity };
+                count += 1;
+            }
+            if (settings.max_field_section_size) |max| {
+                entries[count] = .{ .id = .max_field_section_size, .id_value = 0x06, .value = max };
+                count += 1;
+            }
+            if (settings.qpack_blocked_streams != 0) {
+                entries[count] = .{ .id = .qpack_blocked_streams, .id_value = 0x07, .value = settings.qpack_blocked_streams };
+                count += 1;
+            }
+            if (settings.enable_connect_protocol) {
+                entries[count] = .{ .id = .enable_connect_protocol, .id_value = 0x08, .value = 1 };
+                count += 1;
+            }
+            if (settings.h3_datagram) {
+                entries[count] = .{ .id = .h3_datagram, .id_value = 0x33, .value = 1 };
+                count += 1;
+            }
+            return frame.encodeSettings(entries[0..count], out);
         }
 
         /// Drain newly accepted and readable peer streams. Call after every
@@ -655,6 +685,37 @@ test "H3 conn: SETTINGS exchange and request/response over a mock transport" {
     try testing.expectEqualStrings("pong", response.body);
     try testing.expectEqualStrings("server", response.headers[0].name);
     client.releaseResponse(id);
+}
+
+test "H3 conn: start emits configured non-default SETTINGS" {
+    const allocator = testing.allocator;
+    var client_transport = MockTransport.init(allocator, true);
+    defer client_transport.deinit();
+    var server_transport = MockTransport.init(allocator, false);
+    defer server_transport.deinit();
+    client_transport.peer = &server_transport;
+    server_transport.peer = &client_transport;
+
+    const H3 = Conn(MockTransport);
+    var server = H3.initWithSettings(allocator, .server, .{
+        .qpack_max_table_capacity = 1024,
+        .qpack_blocked_streams = 2,
+        .max_field_section_size = 4096,
+        .enable_connect_protocol = true,
+    });
+    defer server.deinit();
+
+    try server.start(&server_transport);
+    try testing.expectEqual(@as(?u64, 3), server.control_out);
+    const stream = client_transport.streams.get(3) orelse return error.TestExpectedEqual;
+    var control = frame.ControlStream{};
+    defer control.deinit(allocator);
+    try testing.expectEqual(stream.data.items.len, try control.ingest(allocator, stream.data.items));
+    try testing.expect(control.saw_settings);
+    try testing.expectEqual(@as(u64, 1024), control.settings.qpack_max_table_capacity);
+    try testing.expectEqual(@as(u64, 2), control.settings.qpack_blocked_streams);
+    try testing.expectEqual(@as(?u64, 4096), control.settings.max_field_section_size);
+    try testing.expect(control.settings.enable_connect_protocol);
 }
 
 test "H3 conn: duplicate QPACK encoder stream is a stream creation error" {
--- a/src/http3/conn.zig
+++ b/src/http3/conn.zig
@@ -13,6 +13,7 @@
 //! rejected per RFC 9114 §7.2.3 (we never send MAX_PUSH_ID).
 
 const std = @import("std");
+const early_data = @import("early_data.zig");
 const frame = @import("frame.zig");
 const qpack = @import("qpack.zig");
 const session = @import("session.zig");
@@ -153,6 +154,19 @@ pub fn Conn(comptime Transport: type) type {
         pub fn closeCode(self: *const Self) u64 {
             const code = self.close_code orelse ErrorCode.general_protocol_error;
             return code.wire();
+        }
+
+        pub fn peerSettings(self: *const Self) ?frame.Settings {
+            if (!self.peer_control_view.saw_settings) return null;
+            return self.peer_control_view.settings;
+        }
+
+        /// Client-side #367 handoff: before SETTINGS arrives, H3 0-RTT uses
+        /// RFC defaults; after the peer control stream is decoded,
+        /// subsequently received tickets inherit the remembered server
+        /// SETTINGS snapshot.
+        pub fn encodePeerSettingsForEarlyDataTicket(self: *const Self, out: []u8) early_data.SnapshotError![]const u8 {
+            return early_data.encodeSettingsSnapshot(self.peerSettings() orelse .{}, out);
         }
 
         /// Record the wire close code (first failure wins) and produce the
@@ -716,6 +730,38 @@ test "H3 conn: start emits configured non-default SETTINGS" {
     try testing.expectEqual(@as(u64, 2), control.settings.qpack_blocked_streams);
     try testing.expectEqual(@as(?u64, 4096), control.settings.max_field_section_size);
     try testing.expect(control.settings.enable_connect_protocol);
+}
+
+test "H3 conn: early ticket snapshot uses defaults until peer SETTINGS arrive" {
+    const allocator = testing.allocator;
+    var client_transport = MockTransport.init(allocator, true);
+    defer client_transport.deinit();
+    var server_transport = MockTransport.init(allocator, false);
+    defer server_transport.deinit();
+    client_transport.peer = &server_transport;
+    server_transport.peer = &client_transport;
+
+    const H3 = Conn(MockTransport);
+    var client = H3.init(allocator, .client);
+    defer client.deinit();
+    var snapshot: [early_data.encoded_snapshot_len]u8 = undefined;
+    const defaults = try client.encodePeerSettingsForEarlyDataTicket(&snapshot);
+    try testing.expectEqual(frame.Settings{}, try early_data.decodeSettingsSnapshot(defaults));
+
+    var server = H3.initWithSettings(allocator, .server, .{
+        .qpack_blocked_streams = 3,
+        .max_field_section_size = 8192,
+        .enable_connect_protocol = true,
+    });
+    defer server.deinit();
+    try server.start(&server_transport);
+    try client.pump(&client_transport);
+    const remembered = try client.encodePeerSettingsForEarlyDataTicket(&snapshot);
+    try testing.expectEqual(frame.Settings{
+        .qpack_blocked_streams = 3,
+        .max_field_section_size = 8192,
+        .enable_connect_protocol = true,
+    }, try early_data.decodeSettingsSnapshot(remembered));
 }
 
 test "H3 conn: duplicate QPACK encoder stream is a stream creation error" {
--- a/src/http3/conn.zig
+++ b/src/http3/conn.zig
@@ -26,6 +26,7 @@ pub const H3Error = error{
     ProtocolError,
     ResponseTooLarge,
     UnknownStream,
+    UnsupportedH3Setting,
 };
 
 /// HTTP/3 wire error codes (RFC 9114 §8.1). Zig errors carry no payload, so
@@ -206,6 +207,7 @@ pub fn Conn(comptime Transport: type) type {
         /// connection is established.
         pub fn start(self: *Self, transport: *Transport) !void {
             if (self.control_out != null) return;
+            frame.validateLocallySupportedSettings(self.local_settings) catch return error.UnsupportedH3Setting;
             const control = try transport.openStream(.uni);
             var bytes: [64]u8 = undefined;
             var len: usize = 0;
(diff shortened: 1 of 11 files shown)


src/tls/tls13_backend.zig
```diff
@@ -3821,6 +3833,7 @@ pub const Tls13Backend = struct {
                 .client => self.peerTransportCompat(),
                 .server => self.localTransportCompat(),
             };
+            ctx.early_data_application_compat = self.ownedApplicationCompat();
```
Jaruso: **Blocking — this needs a dedicated early-ticket application snapshot, not `ownedApplicationCompat()`.** #367 deliberately separates ordinary 1-RTT application compatibility from the H3 state stamped into *early-capable* tickets. The distinction is also required for the client lifecycle: `setApplicationCompat()` is idle-only, but an H3 client must seed the early-ticket snapshot with RFC default server SETTINGS before `start()`, then update it **after handshake completion** when peer SETTINGS arrives so only subsequently received NSTs inherit the decoded values.

Please add separate owned state, conceptually:

```zig
early_application_compat_format_id: u16 = 0,
early_application_compat_format_version: u16 = 0,
early_application_compat_bytes: [session.hard_max_compat_len]u8 = undefined,
early_application_compat_len: usize = 0,
early_application_compat_present: bool = false,

pub fn setEarlyDataApplicationCompat(
    self: *Tls13Backend,
    blob: ?new_session_ticket.CompatBlob,
) HandshakeError!void {
    switch (self.core.handshake_lifecycle) {
        .idle, .complete => {},
        .running, .failed => return error.InvalidHandshakeState,
    }
    // bounded owned copy, same hygiene as setApplicationCompat
}
```

and stamp:

```zig
ctx.early_data_application_compat = self.ownedEarlyDataApplicationCompat();
```

The H3 client side then needs to initialize this to an encoded `frame.Settings{}` before the handshake and replace it with the first decoded peer SETTINGS snapshot after completion. Tests should prove: (1) NST before SETTINGS stores defaults; (2) peer SETTINGS then a later NST stores the decoded settings; (3) the earlier ticket is not retroactively changed. Server H3 ticket issuance should use the same dedicated setter with its *local* SETTINGS.

src/http3/early_data.zig
```diff
@@ -0,0 +1,176 @@
...
+    std.mem.writeInt(u64, out[17..25], settings.max_field_section_size orelse 0, .big);
+    return out[0..encoded_snapshot_len];
+}
+
+pub fn decodeSettingsSnapshot(bytes: []const u8) SnapshotError!frame.Settings {
+    if (bytes.len != encoded_snapshot_len) return error.MalformedSnapshot;
+    const flags = bytes[0];
+    if (flags & ~known_flags != 0) return error.MalformedSnapshot;
+    return .{
+        .qpack_max_table_capacity = std.mem.readInt(u64, bytes[1..9], .big),
```
Jaruso, line 62: **Blocking — the snapshot decoder accepts values the H3 SETTINGS codec cannot represent.** #367 explicitly requires every decoded value to fit the setting domain. These three fields are HTTP/3 SETTINGS values encoded as QUIC varints, whose maximum is `2^62 - 1`; this code currently accepts arbitrary `u64` values from a recovered ticket.

Please validate the semantic values on both encode and decode, e.g. with the existing `quic_varint.max_value`:

```zig
fn settingsDomainValid(settings: frame.Settings) bool {
    if (settings.qpack_max_table_capacity > quic_varint.max_value) return false;
    if (settings.qpack_blocked_streams > quic_varint.max_value) return false;
    if (settings.max_field_section_size) |v| {
        if (v > quic_varint.max_value) return false;
    }
    return true;
}
```

A malformed recovered snapshot should classify as `.malformed_state` and reject **only 0-RTT**. On the local/config encoding side, return a real validation error instead of allowing `Runtime.init()` to depend on `catch unreachable`. Add max-value / max+1 tests for all three numeric fields.

src/http3/early_data.zig
```diff
@@ -0,0 +1,176 @@
...
+    };
+}
+
+pub fn decodeCompatView(view: CompatView) SnapshotError!frame.Settings {
+    if (view.format_id != format_id or view.format_version != format_version) return error.FormatMismatch;
+    return decodeSettingsSnapshot(view.bytes);
+}
+
+pub fn rememberedSettingsCompatible(remembered: frame.Settings, current: frame.Settings) bool {
+    if (current.qpack_max_table_capacity < remembered.qpack_max_table_capacity) return false;
```
Jaruso: **Blocking protocol bug — QPACK max-table compatibility is not simply `current >= remembered`.** RFC 9204 §3.2.3 has a special 0-RTT rule: if the remembered `SETTINGS_QPACK_MAX_TABLE_CAPACITY` is `0`, the server may advertise a nonzero current value; but if the remembered value is nonzero, the server must send the **same nonzero value**.

With the current predicate, `remembered=4096, current=8192` is accepted. `Conn.start()` will then emit 8192 after accepting 0-RTT, and a compliant client must treat that as a QPACK connection error.

Please special-case it before the other directional checks:

```zig
fn qpackMaxTableCompatible(remembered: u64, current: u64) bool {
    if (remembered == 0) return true;
    return current == remembered;
}

if (!qpackMaxTableCompatible(
    remembered.qpack_max_table_capacity,
    current.qpack_max_table_capacity,
)) return false;
```

Regression matrix: `0->0` yes, `0->N` yes, `N->N` yes, `N->smaller` no, **`N->larger` no**. The generic `>=` rule in #367 should be corrected as well because RFC 9204 overrides it for this setting.

Jaruso (commented):
Blocking from my review on the currently visible head `9bd062bd9f4355065eb817177da17270f71013a9`. CI #918 is green, and the slice gets several important pieces right: the canonical fixed-size H3 snapshot, format/version discriminator, identity-0 client accessor, server-side early-data compatibility hook, and emission of configured non-default SETTINGS all fit the intended architecture. Ordinary QUIC PSK resumption also remains configured independently via the existing `.ignore` resumption compatibility policy.

I found three blockers:

1. **The H3 early-ticket application snapshot is not actually separate from ordinary `application_compat`, and the required client NST lifecycle cannot be implemented with the current idle-only setter.** #367 explicitly requires a dedicated early-data ticket snapshot that can be seeded with RFC-default server SETTINGS before client handshake start and updated after handshake completion when peer SETTINGS arrives, affecting only subsequently received NSTs. Please add the separate owned backend state/setter, wire H3 client SETTINGS reception into it, and test NST-before-SETTINGS/defaults plus SETTINGS-before-later-NST without retroactively mutating the first ticket.

2. **The canonical decoder accepts setting values outside the HTTP/3/QUIC varint domain.** The issue explicitly requires domain validation. Values above `2^62-1` can be decoded into `frame.Settings` even though `frame.encodeSettings()` cannot put them on the wire. Validate both snapshot encoding and decoding and fail closed for malformed remembered state; runtime configuration should fail cleanly rather than relying on `catch unreachable`.

3. **`SETTINGS_QPACK_MAX_TABLE_CAPACITY` has a QPACK-specific 0-RTT rule that is stricter than the generic monotonic-limit rule.** RFC 9204 §3.2.3 permits remembered `0 -> nonzero`, but when the remembered value is nonzero the server must send the *same* nonzero value. The current `current >= remembered` check accepts `4096 -> 8192`; after 0-RTT acceptance `Conn.start()` would then emit 8192, which a compliant client must treat as an error. Please special-case this field and add the boundary matrix. The generic `>=` wording in #367 should also be corrected so it does not reintroduce this later.

Before merge I also want the slice-level behavioral proof required by #367: an H3/TLS resumption case where a more restrictive/malformed remembered application snapshot yields `EarlyDataDecision.application_incompatible` while the otherwise-valid PSK still resumes at 1-RTT, plus a compatible case that reaches 0-RTT acceptance when the other #366 gates are injected as allowing. Pure snapshot unit tests alone do not prove the runtime gate is installed with the intended resumption independence.

Because this PR is authored by the authenticated account, I cannot submit a formal REQUEST_CHANGES review; this COMMENT should be treated as blocking.

src/http3/conn.zig
```diff
@@ -149,6 +156,19 @@ pub fn Conn(comptime Transport: type) type {
...
+            return self.peer_control_view.settings;
+        }
+
+        /// Client-side #367 handoff: before SETTINGS arrives, H3 0-RTT uses
+        /// RFC defaults; after the peer control stream is decoded,
+        /// subsequently received tickets inherit the remembered server
+        /// SETTINGS snapshot.
+        pub fn encodePeerSettingsForEarlyDataTicket(self: *const Self, out: []u8) early_data.SnapshotError![]const u8 {
+            return early_data.encodeSettingsSnapshot(self.peerSettings() orelse .{}, out);
+        }
```
Jaruso, line 171: **Blocking — the client ticket-lifecycle pieces are still disconnected.** This helper correctly produces RFC defaults before peer SETTINGS and the decoded peer values afterward, and the TLS backend now has the right post-handshake-mutable `setEarlyDataApplicationCompat()` state. But no changed composition path actually calls the setter from this H3 state. The H3 unit test only exercises this encoder, while the TLS lifecycle test calls the setter manually, so both can stay green even if real H3 NST ingestion never updates the ticket-stamping snapshot.

Please wire one shared handoff used by the H3 client composition, conceptually:

```zig
fn syncEarlyTicketSettings(
    h3: *H3,
    backend: *quic.tls_backend.Tls13Backend,
) !void {
    var buf: [http3.early_data.encoded_snapshot_len]u8 = undefined;
    const snapshot = try h3.encodePeerSettingsForEarlyDataTicket(&buf);
    try backend.setEarlyDataApplicationCompat(.{
        .format_id = http3.early_data.format_id,
        .format_version = http3.early_data.format_version,
        .bytes = snapshot,
    });
}
```

Seed it once with defaults before the client handshake/ticket ingestion path is armed; after `pump()` transitions peer SETTINGS to known, update the backend before later NSTs are consumed. Then add one stitched deterministic regression that captures two actual tickets: first before SETTINGS (default snapshot), second after SETTINGS (decoded snapshot), and asserts the first stored ticket was not retroactively mutated.

src/http3/conn.zig
```diff
@@ -190,15 +210,39 @@ pub fn Conn(comptime Transport: type) type {
...
             len += (try frame.encodeKnownFrame(.settings, settings, bytes[len..])).len;
             _ = try transport.writeStream(control, bytes[0..len], false);
             self.control_out = control;
         }
 
+        fn encodeLocalSettings(settings: frame.Settings, entries: *[5]frame.Setting, out: []u8) ![]u8 {
+            var count: usize = 0;
+            if (settings.qpack_max_table_capacity != 0) {
+                entries[count] = .{ .id = .qpack_max_table_capacity, .id_value = 0x01, .value = settings.qpack_max_table_capacity };
+                count += 1;
```
Jaruso, line 239: **Blocking — do not advertise local SETTINGS for capabilities this `Conn` cannot honor.** The new `local_settings` path can emit nonzero `SETTINGS_QPACK_MAX_TABLE_CAPACITY`, but this same type still documents a static-table QPACK baseline and `pumpUniStreams()` explicitly drains/ignores QPACK encoder/decoder instructions. A peer is allowed to use dynamic QPACK when we advertise nonzero capacity; its request field section will then reach the static-only `session.RequestStream` decoder and fail.

The same problem exists for the two booleans: `enable_connect_protocol = true` advertises Extended CONNECT even though `RequestStream.applyPseudoHeader()` does not accept `:protocol`, and `h3_datagram = true` advertises willingness to receive HTTP/3 Datagrams while this layer has no HTTP Datagram handling.

For this slice I would fail closed rather than pulling those protocol features in:

```zig
fn localSettingsSupported(settings: frame.Settings) bool {
    if (!early_data.settingsDomainValid(settings)) return false;
    if (settings.qpack_max_table_capacity != 0) return false; // dynamic QPACK not wired here
    if (settings.enable_connect_protocol) return false;       // :protocol not supported
    if (settings.h3_datagram) return false;                   // HTTP Datagram path not supported
    return true;
}
```

Apply that before sending SETTINGS (preferably at `Runtime.init` and/or `Conn.start`). `max_field_section_size` is independently representable, and `qpack_blocked_streams` with table capacity 0 is inert; the important invariant is that every advertised non-default capability is one the current H3 implementation can actually service. Replace the current non-default SETTINGS test with supported values plus explicit rejection tests for the unsupported capabilities.

Jaruso (commented):
Still blocking on two integration issues after the fresh pass.

What is fixed from the prior review:
- dedicated `early_application_compat` state is now independent from ordinary 1-RTT application compatibility;
- the early snapshot can be updated after a completed handshake and is wiped on teardown;
- snapshot encode/decode now validates the QUIC-varint domain and `Runtime.init` returns a real validation error;
- QPACK max-table 0-RTT compatibility now implements the RFC 9204 special rule (`0 -> N` permitted, nonzero remembered value must remain exactly equal);
- regression coverage now proves application incompatibility rejects 0-RTT without rejecting the PSK-resumed handshake;
- CI #919 is green.

The remaining blockers are:

1. **The H3 client SETTINGS -> TLS ticket-stamping handoff is still not actually wired.** `Conn.encodePeerSettingsForEarlyDataTicket()` and `Tls13Backend.setEarlyDataApplicationCompat()` now exist, but the changed code has no composition path that invokes them together. The H3 test proves the first helper returns defaults/peer SETTINGS; the TLS test separately calls the setter by hand. Both tests can pass while a real H3 client continues receiving NSTs with no/default application snapshot forever. Wire the handoff (or add a shared composition helper used by the real/deterministic client path) and add one stitched regression proving: NST before SETTINGS -> defaults, SETTINGS receipt updates the backend, later NST -> decoded peer snapshot, first ticket unchanged.

2. **`initWithSettings` can now advertise H3 capabilities this `Conn` cannot honor.** This connection implementation still declares itself static-table QPACK, drains/ignores QPACK encoder/decoder instructions, and decodes request field sections through the static-only session decoder. Advertising nonzero `SETTINGS_QPACK_MAX_TABLE_CAPACITY` therefore authorizes a compliant peer to use dynamic QPACK that this server cannot decode. The same API can advertise Extended CONNECT and H3 DATAGRAM even though this request parser does not accept `:protocol` and this layer has no HTTP Datagram handling. Either wire those features before advertising them or reject unsupported local SETTINGS; for this slice, fail-closed validation is the smaller and safer change.

I am intentionally not treating QUIC remembered-TP comparison as a new blocker in this slice because #366 owns that implementation; this PR correctly uses the single early-data compatibility gate seam rather than creating a second gate. The eventual #366 integration still needs to extend this callback so transport incompatibility is evaluated before the H3 application result.

Once these two integration gaps are closed, I expect this slice to be ready.

Jaruso (commented):
Re-reviewed the full Slice 3 diff at `ebc88ac2ecd91456744087acf9d3e8dcf35dfd63` against #367 and the prior review findings. CI #920 is green, and the previous SETTINGS-advertisement blocker is fixed correctly: unsupported local QPACK/Extended CONNECT/H3 Datagram/max-field settings now fail closed before they can be advertised. The dedicated early-ticket TLS state, snapshot validation, RFC 9204 QPACK rule, server compatibility gate, and 1-RTT-independence tests also remain sound.

There are still **two blocking integration issues**, called out inline:

1. The client default H3 snapshot is still seeded from `H3.start()`, which occurs only after the QUIC/TLS handshake is established. That does not satisfy the explicit NST-before-SETTINGS requirement if a server sends a NewSessionTicket immediately after handshake completion but before the application starts H3. The current e2e test accidentally avoids that case by calling `client_h3.start()` before issuing ticket #1.
2. `TlsBackend.setEarlyDataApplicationCompat()` treats a missing optional vtable hook as `InvalidHandshakeState`. Because `Connection` always declares the forwarding method, `http3.Conn`'s `@hasDecl` check will call it even for backends such as `TestTlsBackend` that intentionally leave the hook null, regressing ordinary non-resumption H3 clients over backend-agnostic QUIC.

So this is **not ready to merge for Slice 3 yet**. I did not find additional protocol/compatibility defects beyond these two in this pass.

src/http3/conn.zig
```diff
@@ -186,17 +207,54 @@ pub fn Conn(comptime Transport: type) type {
...
-            // Static-table-only QPACK: all our SETTINGS are the RFC defaults,
-            // so the frame is legitimately empty.
-            const settings = try frame.encodeSettings(&.{}, &settings_payload);
+            var settings_payload: [64]u8 = undefined;
+            var settings_entries: [5]frame.Setting = undefined;
+            const settings = try encodeLocalSettings(self.local_settings, &settings_entries, &settings_payload);
             len += (try frame.encodeKnownFrame(.settings, settings, bytes[len..])).len;
             _ = try transport.writeStream(control, bytes[0..len], false);
             self.control_out = control;
+            if (self.role == .client) try self.publishEarlyTicketSnapshot(transport);
```
Jaruso, line 221: **Blocking — the default remembered SETTINGS snapshot is still seeded too late for the actual NST-before-SETTINGS case.** `start()` is explicitly called only once the QUIC connection is established, while `Connection.init()` starts TLS immediately. A server is allowed to send an NST as soon as the handshake completes, so that NST can be consumed before this line runs; the ticket then gets no `early_data_application_compat` snapshot instead of the required canonical RFC-default H3 SETTINGS snapshot.

The new e2e test does not exercise this edge because it does `runUntil(bothEstablished) -> client_h3.start() -> issue ticket #1`, so defaults are already installed before ticket #1 exists.

Please seed the H3 default snapshot on the concrete client backend **before `Connection.init()` starts TLS** (or through an equivalent pre-handshake composition seam), then keep the current first-peer-SETTINGS publication for later tickets. The regression should deliver ticket #1 immediately after handshake completion *without calling `H3.start()` first*, verify defaults, then start/pump H3, receive peer SETTINGS, and verify ticket #2 gets those settings while ticket #1 remains unchanged.

src/quic/tls_handshake.zig
```diff
@@ -122,6 +126,11 @@ pub const TlsBackend = struct {
         if (self.setPostHandshakeAllocatorFn) |set| try set(self.transport.ptr, allocator);
     }
 
+    pub fn setEarlyDataApplicationCompat(self: TlsBackend, blob: ?tls_core.new_session_ticket.CompatBlob) HandshakeError!void {
+        const set = self.setEarlyDataApplicationCompatFn orelse return error.InvalidHandshakeState;
```
Jaruso: **Blocking regression — this hook is declared optional, but a missing implementation now breaks ordinary H3 clients.** `http3.Conn.publishEarlyTicketSnapshot()` checks `@hasDecl(Transport, "setEarlyDataApplicationCompat")`; `Connection` always has that method, so the check is true even when the underlying `TlsBackend` does not implement `setEarlyDataApplicationCompatFn`. `TestTlsBackend.backend()` is exactly such a backend: it leaves this hook null. A client H3 `start()` over that backend now gets `InvalidHandshakeState` despite not using resumption at all.

Missing support here should be capability absence, not a handshake failure. The simplest shape is consistent with `setPostHandshakeAllocator()`:

```zig
pub fn setEarlyDataApplicationCompat(self: TlsBackend, blob: ?CompatBlob) HandshakeError!void {
    if (self.setEarlyDataApplicationCompatFn) |set| {
        try set(self.transport.ptr, blob);
    }
}
```

Native backend lifecycle errors should still propagate when the hook exists. Please add a regression proving a backend without this optional hook can still run an ordinary H3 client path.

Jaruso (commented):
Re-reviewed the full #367 Slice 3 implementation at `bbca36cb65b631c836ddc0afc197d1485d3da75f`, including the latest fixes and all previously reviewed H3/TLS compatibility paths.

The remaining blockers are fixed:

- the canonical RFC-default H3 SETTINGS snapshot is now installed on the client before `Connection.init()` starts TLS, so a NewSessionTicket delivered immediately after handshake completion and before `H3.start()` is stamped with defaults;
- the stitched E2E now actually exercises that ordering, then proves a later ticket receives decoded peer SETTINGS while the first ticket remains unchanged;
- the optional `setEarlyDataApplicationCompat` backend hook is now a no-op when unsupported, so ordinary backend-agnostic H3 clients are not regressed while real native-backend lifecycle errors still propagate;
- unsupported local H3 SETTINGS remain fail-closed;
- dedicated early-ticket application state, QUIC-varint domain validation, RFC 9204 QPACK max-table compatibility, first-surviving-wire offer access, and 1-RTT/0-RTT compatibility independence remain intact.

CI #921 is green. I found no additional blocking correctness issues in the latest delta or the full affected slice.

**Ready to merge for #367 Slice 3.** This does not close #367; H2 safe deferral/final cross-protocol work remains in the later slice, and actual QUIC/TLS 0-RTT carrier + anti-replay remain owned by #366/#368.

PR #501, opened 2026-07-24
Issue #367 Slice 4: H2 deferral + early-data observability/docs

Jaruso:
## Summary
Implements Issue #367 Slice 4 with scoped HTTP-level behavior and observability updates:

- Add H2 early-stream safe deferral behavior with single-dispatch guardrails.
- Add bounded early-data observability fields in request context and access logs.
- Add bounded HTTP early-data metrics families:
  - `tardigrade_http_early_data_requests_total{protocol,source}`
  - `tardigrade_http_early_data_decisions_total{protocol,decision}`
  - `tardigrade_http_early_data_upstream_425_total{action}`
  - `tardigrade_http_early_data_retry_total{result}`
  - `tardigrade_http3_early_data_compat_total{decision}`
- Wire HTTP/3 compatibility outcomes into gateway metrics via callback bridge.
- Update operator docs for observability and support boundaries.

## Scope Boundaries
- Includes #367 Slice 4 behavior (H2 deferral + observability/docs).
- Does **not** implement #366 transport-carrier internals.
- Does **not** implement #368 anti-replay persistence/store.

## Validation
- `zig build --summary all --error-style verbose`

All commands passed locally.

### DIFF
--- a/docs/OBSERVABILITY.md
+++ b/docs/OBSERVABILITY.md
@@ -11,7 +11,18 @@ Runtime logs and access logs are JSON by default.
   `request_id` and `correlation_id`
 - access logs: `type`, `ts`, `request_id`, `correlation_id`, `method`, `path`,
   `status`, `latency_ms`, `client_ip`, `upstream_addr`, `upstream_status`,
-  `identity`, `bytes_sent`, `response_bytes`, and `error_category`
+  `identity`, `bytes_sent`, `response_bytes`, `error_category`,
+  `early_data_source`, `early_data_action`, `early_data_retry_result`, and
+  `early_data_replay_exposed`
+
+Early-data access-log fields are bounded enums/booleans only:
+
+- `early_data_source`: `none`, `transport`, `header`, `both`
+- `early_data_action`: `ordinary`, `accepted`, `forwarded`, `too_early`,
+  `deferred`, `retried`
+- `early_data_retry_result`: `none`, `success`, `too_early`, `failed`
+- `early_data_replay_exposed`: `true` when transport and/or header provenance
+  indicates replay exposure for this request
 
 Access logs are written through `src/http/access_log.zig`. Runtime component
 logs are written through `src/http/logger.zig`.
@@ -77,6 +88,15 @@ logs are written through `src/http/logger.zig`.
   eligibility. Upload and response eligibility are evaluated separately, so one
   request can contribute more than one fallback event.
 - reverse-proxy upstream TTFB summary: `tardigrade_proxy_ttfb_ms`
+- HTTP-level early-data counters (fixed labels only):
+  `tardigrade_http_early_data_requests_total{protocol,source}`,
+  `tardigrade_http_early_data_decisions_total{protocol,decision}`,
+  `tardigrade_http_early_data_upstream_425_total{action}`,
+  `tardigrade_http_early_data_retry_total{result}`,
+  and `tardigrade_http3_early_data_compat_total{decision}`
+
+Early-data metric label sets are intentionally bounded and never include
+high-cardinality request attributes (URL, request id, stream id, host, IP).
 
 The latency histogram is intentionally global rather than route-labeled to keep
 hot-path overhead predictable.
--- a/docs/OBSERVABILITY.md
+++ b/docs/OBSERVABILITY.md
@@ -20,7 +20,7 @@ Early-data access-log fields are bounded enums/booleans only:
 - `early_data_source`: `none`, `transport`, `header`, `both`
 - `early_data_action`: `ordinary`, `accepted`, `forwarded`, `too_early`,
   `deferred`, `retried`
-- `early_data_retry_result`: `none`, `success`, `too_early`, `failed`
+- `early_data_retry_result`: `none`, `success`, `too_early`, `failure`
 - `early_data_replay_exposed`: `true` when transport and/or header provenance
   indicates replay exposure for this request
 
--- a/docs/SUPPORT_MATRIX.md
+++ b/docs/SUPPORT_MATRIX.md
@@ -62,6 +62,7 @@ surfaces that should not be marketed as generic operator-facing capabilities.
 | --- | --- | --- | --- |
 | HTTP/2 | `tls_termination` HTTP/2 path, `hpack`, `http2_frame` | `experimental` | Present in-tree, but not documented or release-gated at the same level as the HTTP/1.1 core. |
 | HTTP/3 / QUIC | `http3_handler`, `http3_session`, `http3_runtime`, `quic`, `http3` | `experimental` | Served by the native Zig QUIC/H3 stack (no system libraries); enable at runtime with `--http3` and a QUIC-compatible TLS identity (Ed25519 or ECDSA P-256). ngtcp2/nghttp3 removed under #328; they remain out-of-process interop peers only. |
+| Early-data replay handling across H1/H2/H3 | `early_data`, `request_context`, `edge_gateway` H1/H2 preflight/deferral, `http3_runtime` compatibility gate | `experimental` | Current scope: #367 HTTP-level policy, H2 safe deferral, and bounded observability (metrics/access logs). Out of scope here: #366 transport-carrier internals and #368 anti-replay persistence/store. |
 | WebSocket, SSE, and mux realtime paths | `websocket`, `event_hub`, mux counters in `metrics` | `experimental` | Integration coverage exists, but the public operator docs are still example-scoped rather than Core v1. |
 | ACME automation | `acme_client` | `experimental` | Useful feature surface, but not part of the stable release/install promise yet. |
 | Auth and identity extensions | `auth`, `basic_auth`, `jwt`, `access_control` | `experimental` | Operators can use these paths, but they are not the defining Core v1 server contract. |
--- a/src/edge_gateway.zig
+++ b/src/edge_gateway.zig
@@ -389,6 +389,8 @@ pub fn run(cfg: *const edge_config.EdgeConfig) !void {
             .max_datagram_size = cfg.http3_max_datagram_size,
             .request_handler = ghandlers.handleHttp3Request,
             .request_handler_ctx = &http3_dispatch_ctx,
+            .early_data_compat_metrics_ctx = &state,
+            .early_data_compat_metrics_cb = recordHttp3EarlyDataCompatFromRuntime,
         }) catch |err| blk: {
             state.logger.warn(null, "HTTP/3 listener failed to initialize: {s}", .{@errorName(err)});
             break :blk null;
@@ -1591,6 +1593,8 @@ const WaitingEncryptedHttpConnection = struct {
     tls_metrics: ?*http.metrics.Metrics = null,
     tls_metrics_mutex: ?*compat.Mutex = null,
     tls_metrics_state: ?*http.metrics.TlsBufferConnectionMetrics = null,
+    last_read_transport_early: bool = false,
+    handshake_complete_override: ?bool = null,
 
     fn init(
         inner: http.encrypted_stream_connection.EncryptedStreamHttpConnection,
@@ -1641,9 +1645,26 @@ const WaitingEncryptedHttpConnection = struct {
                 },
                 else => return err,
             };
+            self.last_read_transport_early = false;
             self.observeTlsBufferMetrics();
             return n;
         }
+    }
+
+    pub fn lastReadTransportEarly(self: *const WaitingEncryptedHttpConnection) bool {
+        return self.last_read_transport_early;
+    }
+
+    pub fn setLastReadTransportEarlyForTest(self: *WaitingEncryptedHttpConnection, early: bool) void {
+        self.last_read_transport_early = early;
+    }
+
+    pub fn downstreamHandshakeComplete(self: *const WaitingEncryptedHttpConnection) bool {
+        return self.handshake_complete_override orelse true;
+    }
+
+    pub fn setDownstreamHandshakeCompleteForTest(self: *WaitingEncryptedHttpConnection, complete: bool) void {
+        self.handshake_complete_override = complete;
     }
 
     pub fn write(self: *WaitingEncryptedHttpConnection, bytes: []const u8) !usize {
@@ -1857,6 +1878,105 @@ fn parkedConnectionCloseHook(raw_state: *anyopaque, fd: std.posix.fd_t) void {
     state.releaseConnectionSlot(fd);
 }
 
+fn recordHttp3EarlyDataCompatFromRuntime(
+    raw_state: *anyopaque,
+    decision: http.metrics.H3EarlyDataCompatDecision,
+) void {
+    const state: *GatewayState = @ptrCast(@alignCast(raw_state));
+    state.metricsRecordHttp3EarlyDataCompat(decision);
+}
+
+fn h2LastReadTransportEarly(conn: anytype) bool {
+    const T = @TypeOf(conn);
+    if (comptime std.meta.activeTag(@typeInfo(T)) == .pointer) {
+        const Child = std.meta.Child(T);
+        if (comptime @hasDecl(Child, "lastReadTransportEarly")) return conn.lastReadTransportEarly();
+    } else {
+        if (comptime @hasDecl(T, "lastReadTransportEarly")) return conn.lastReadTransportEarly();
+    }
+    return false;
+}
+
+fn h2DownstreamHandshakeComplete(conn: anytype) bool {
+    const T = @TypeOf(conn);
+    if (comptime std.meta.activeTag(@typeInfo(T)) == .pointer) {
+        const Child = std.meta.Child(T);
+        if (comptime @hasDecl(Child, "downstreamHandshakeComplete")) return conn.downstreamHandshakeComplete();
+    } else {
+        if (comptime @hasDecl(T, "downstreamHandshakeComplete")) return conn.downstreamHandshakeComplete();
+    }
+    return true;
+}
+
+fn h2AppendReadyStream(ready_streams: *std.array_list.Managed(u31), sid: u31) !void {
+    for (ready_streams.items) |existing| {
+        if (existing == sid) return;
+    }
+    try ready_streams.append(sid);
+}
+
+fn h2DispatchReadyStreams(
+    conn: anytype,
+    allocator: std.mem.Allocator,
+    state: *GatewayState,
+    cfg: *const edge_config.EdgeConfig,
+    pending: *std.AutoHashMap(u31, Http2PendingStream),
+    streams: *std.AutoHashMap(u31, http.http2_stream.Stream),
+    ready_streams: *std.array_list.Managed(u31),
+    next_server_stream_id: *u31,
+    conn_send_window: *i32,
+) !void {
+    if (ready_streams.items.len == 0) return;
+
+    const initial_ready = ready_streams.items.len;
+    var processed: usize = 0;
+    while (processed < initial_ready and ready_streams.items.len > 0) : (processed += 1) {
+        var best_idx: usize = 0;
+        var best_weight: u8 = 0;
+        for (ready_streams.items, 0..) |sid, idx| {
+            const w = if (streams.get(sid)) |s| s.priority_weight else 16;
+            if (w >= best_weight) {
+                best_weight = w;
+                best_idx = idx;
+            }
+        }
+
+        const sid = ready_streams.swapRemove(best_idx);
+        const stream_send = if (streams.get(sid)) |s| s.send_window else conn_send_window.*;
+        const ps = pending.getPtr(sid) orelse continue;
+        if (ps.dispatch_count > 0) continue;
+
+        if (ps.transport_early and !h2DownstreamHandshakeComplete(conn)) {
+            if (!ps.early_request_recorded) {
+                state.metricsRecordEarlyDataRequest(.h2, .transport);
+                ps.early_request_recorded = true;
+            }
+            if (!ps.deferred_recorded) {
+                state.metricsRecordEarlyDataDecision(.h2, .deferred);
+                ps.deferred_recorded = true;
+            }
+            try h2AppendReadyStream(ready_streams, sid);
+            continue;
+        }
+
+        if (ps.transport_early and !ps.early_request_recorded) {
+            state.metricsRecordEarlyDataRequest(.h2, .transport);
+            ps.early_request_recorded = true;
+        }
+        if (ps.transport_early) {
+            state.metricsRecordEarlyDataDecision(.h2, .accepted);
+        }
+
+        try respondHttp2Stream(conn.writer(), allocator, state, cfg, sid, ps, next_server_stream_id, conn_send_window, stream_send);
+        ps.dispatch_count += 1;
+        if (pending.fetchRemove(sid)) |removed| {
+            var tmp = removed.value;
+            tmp.deinit(allocator);
+        }
+        _ = streams.remove(sid);
+    }
+}
+
 fn handleHttp2Connection(conn: anytype, session: *ConnectionSession, cfg: *const edge_config.EdgeConfig, state: *GatewayState, connection_ip: []const u8) !void {
     _ = session;
     _ = connection_ip;
@@ -1893,11 +2013,24 @@ fn handleHttp2Connection(conn: anytype, session: *ConnectionSession, cfg: *const
     }
 
     while (!http.shutdown.isShutdownRequested() and !goaway_received) {
+        try h2DispatchReadyStreams(
+            conn,
+            allocator,
+            state,
+            cfg,
+            &pending,
+            &streams,
+            &ready_streams,
+            &next_server_stream_id,
+            &conn_send_window,
+        );
+
         var frame = http.http2_frame.readFrame(conn, allocator, HTTP2_MAX_FRAME_SIZE) catch |err| switch (err) {
             error.ConnectionClosed => return,
             else => return err,
         };
         defer http.http2_frame.deinitFrame(allocator, &frame);
+        const frame_transport_early = h2LastReadTransportEarly(conn);
 
         switch (frame.typ) {
             .settings => {
@@ -1924,6 +2057,7 @@ fn handleHttp2Connection(conn: anytype, session: *ConnectionSession, cfg: *const
                 };
                 defer http.hpack.deinitDecoded(allocator, &decoded);
                 var ps = pending.get(frame.stream_id) orelse Http2PendingStream.init(allocator);
+                ps.transport_early = ps.transport_early or frame_transport_early;
                 if (streams.get(frame.stream_id)) |s| ps.priority_weight = s.priority_weight;
                 for (decoded.headers) |h| {
                     if (std.mem.eql(u8, h.name, ":method")) {
@@ -1939,7 +2073,7 @@ fn handleHttp2Connection(conn: anytype, session: *ConnectionSession, cfg: *const
                 try pending.put(frame.stream_id, ps);
                 if ((frame.flags & http.http2_frame.Flags.END_STREAM) != 0) {
                     if (streams.getPtr(frame.stream_id)) |s| s.remoteEndStream() catch {};
-                    try ready_streams.append(frame.stream_id);
+                    try h2AppendReadyStream(&ready_streams, frame.stream_id);
                 }
             },
             .data => {
@@ -1947,6 +2081,7 @@ fn handleHttp2Connection(conn: anytype, session: *ConnectionSession, cfg: *const
                 if (streams.getPtr(frame.stream_id)) |s| s.send_window -= @intCast(frame.payload.len);
                 conn_send_window -= @intCast(frame.payload.len);
                 if (pending.getPtr(frame.stream_id)) |ps| {
+                    ps.transport_early = ps.transport_early or frame_transport_early;
                     try ps.body.appendSlice(frame.payload);
                     try http.http2_frame.writeWindowUpdate(conn.writer(), frame.stream_id, @intCast(frame.payload.len));
                     try http.http2_frame.writeWindowUpdate(conn.writer(), 0, @intCast(frame.payload.len));
@@ -1954,7 +2089,7 @@ fn handleHttp2Connection(conn: anytype, session: *ConnectionSession, cfg: *const
                     conn_send_window += @intCast(frame.payload.len);
                     if ((frame.flags & http.http2_frame.Flags.END_STREAM) != 0) {
                         if (streams.getPtr(frame.stream_id)) |s| s.remoteEndStream() catch {};
-                        try ready_streams.append(frame.stream_id);
+                        try h2AppendReadyStream(&ready_streams, frame.stream_id);
                     }
                 } else {
                     try http.http2_frame.writeGoaway(conn.writer(), frame.stream_id, 1);
@@ -1980,6 +2115,14 @@ fn handleHttp2Connection(conn: anytype, session: *ConnectionSession, cfg: *const
                     tmp.deinit(allocator);
                 }
                 _ = streams.remove(frame.stream_id);
+                var idx: usize = 0;
+                while (idx < ready_streams.items.len) {
+                    if (ready_streams.items[idx] == frame.stream_id) {
+                        _ = ready_streams.swapRemove(idx);
+                        continue;
+                    }
+                    idx += 1;
+                }
             },
             .goaway => {
                 goaway_received = true;
@@ -1987,27 +2130,17 @@ fn handleHttp2Connection(conn: anytype, session: *ConnectionSession, cfg: *const
             .continuation, .push_promise => {},
         }
 
-        while (ready_streams.items.len > 0) {
-            var best_idx: usize = 0;
-            var best_weight: u8 = 0;
-            for (ready_streams.items, 0..) |sid, idx| {
-                const w = if (streams.get(sid)) |s| s.priority_weight else 16;
-                if (w >= best_weight) {
-                    best_weight = w;
-                    best_idx = idx;
-                }
-            }
-            const sid = ready_streams.swapRemove(best_idx);
-            const stream_send = if (streams.get(sid)) |s| s.send_window else conn_send_window;
-            if (pending.getPtr(sid)) |ps| {
-                try respondHttp2Stream(conn.writer(), allocator, state, cfg, sid, ps, &next_server_stream_id, &conn_send_window, stream_send);
-            }
-            if (pending.fetchRemove(sid)) |removed| {
-                var tmp = removed.value;
-                tmp.deinit(allocator);
-            }
-            _ = streams.remove(sid);
-        }
+        try h2DispatchReadyStreams(
+            conn,
+            allocator,
+            state,
+            cfg,
+            &pending,
+            &streams,
+            &ready_streams,
+            &next_server_stream_id,
+            &conn_send_window,
+        );
     }
 }
 
@@ -2256,6 +2389,15 @@ fn earlyDataPreflightDecisionForH1(
     );
 }
 
+fn metricsEarlyDataSource(early_ctx: http.request_context.EarlyDataContext) ?http.metrics.EarlyDataSource {
+    return switch (early_ctx.source()) {
+        .none => null,
+        .transport => .transport,
+        .header => .header,
+        .both => .both,
+    };
+}
+
 const H1PreflightSideEffectProbe = struct {
     mirror_calls: usize = 0,
     auth_calls: usize = 0,
@@ -2370,9 +2512,26 @@ fn executeH1PostPreflightOrchestration(
     lifecycle: *http.request_lifecycle.RequestLifecycle,
     hooks: anytype,
 ) !H1PostPreflightOutcome {
-    switch (earlyDataPreflightDecisionForH1(cfg, ctx.early_data, request)) {
-        .ordinary, .execute_local, .forward_rfc8470 => {},
+    if (metricsEarlyDataSource(ctx.early_data)) |source| {
+        state.metricsRecordEarlyDataRequest(.h1, source);
+    }
+
+    const early_decision = earlyDataPreflightDecisionForH1(cfg, ctx.early_data, request);
+    switch (early_decision) {
+        .ordinary => {
+            ctx.early_data_action = .ordinary;
+        },
+        .execute_local => {
+            ctx.early_data_action = .accepted;
+            state.metricsRecordEarlyDataDecision(.h1, .accepted);
+        },
+        .forward_rfc8470 => {
+            ctx.early_data_action = .forwarded;
+            state.metricsRecordEarlyDataDecision(.h1, .forwarded);
+        },
         .too_early, .defer_until_handshake => {
+            ctx.early_data_action = if (early_decision == .defer_until_handshake) .deferred else .too_early;
+            state.metricsRecordEarlyDataDecision(.h1, if (ctx.early_data_action == .deferred) .deferred else .too_early);
             const status = try hooks.rejectEarly(allocator, writer, state, ctx, request, correlation_id, keep_alive.*);
             return .{ .terminal_status = status };
         },
@@ -2630,6 +2789,11 @@ fn handleConnection(conn: anytype, session: *ConnectionSession, cfg: *const edge
     // Reject before routing so no location block can accidentally serve it.
     if (traceRejectionStatus(request.method, ctx.early_data)) |trace_status| {
         if (trace_status == .too_early) {
+            if (metricsEarlyDataSource(ctx.early_data)) |source| {
+                state.metricsRecordEarlyDataRequest(.h1, source);
+            }
+            ctx.early_data_action = .too_early;
+            state.metricsRecordEarlyDataDecision(.h1, .too_early);
             const status = try ghandlers.writeTooEarlyResponse(allocator, writer, state, &ctx, correlation_id, keep_alive);
             ghandlers.logAccessForRequest(state, &ctx, &request, status);
             return;
@@ -2645,6 +2809,11 @@ fn handleConnection(conn: anytype, session: *ConnectionSession, cfg: *const edge
         std.mem.startsWith(u8, request.uri.path, acme_prefix))
     {
         if (ctx.early_data.replayExposed()) {
+            if (metricsEarlyDataSource(ctx.early_data)) |source| {
+                state.metricsRecordEarlyDataRequest(.h1, source);
+            }
+            ctx.early_data_action = .too_early;
+            state.metricsRecordEarlyDataDecision(.h1, .too_early);
             const status = try ghandlers.writeTooEarlyResponse(allocator, writer, state, &ctx, correlation_id, keep_alive);
             ghandlers.logAccessForRequest(state, &ctx, &request, status);
             return;
@@ -3037,6 +3206,8 @@ test "H1 early-data 425 preflight runs before side effects" {
     defer request.request.deinit();
     var effects = H1PreflightSideEffectProbe{};
     var state: GatewayState = undefined;
+    state.metrics_mutex = .{};
+    state.metrics = http.metrics.Metrics.init();
     var ctx = http.request_context.RequestContext.init(allocator, "req-early", "127.0.0.1");
     ctx.early_data.inbound_marker = true;
     var lifecycle = http.request_lifecycle.RequestLifecycle.init("req-early", 0);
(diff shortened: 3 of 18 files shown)


src/edge_gateway.zig
```diff
@@ -1641,11 +1645,28 @@ const WaitingEncryptedHttpConnection = struct {
                 },
                 else => return err,
             };
+            self.last_read_transport_early = false;
```
Jaruso: **Blocking:** this makes the new H2 provenance seam test-only in production. Every successful plaintext read hard-resets `last_read_transport_early` to `false`, and the only setter added in this PR is `setLastReadTransportEarlyForTest()`. Likewise, `downstreamHandshakeComplete()` defaults to `true` unless a test override is installed. As a result, the real native H2 path cannot ever classify a production request as transport-early or defer it on handshake state.

Please consume the actual #366 carrier/TLS provenance and authenticated handshake-completion state here rather than synthesizing them in tests. The adapter needs a production read/provenance contract (or equivalent narrow accessors) and `downstreamHandshakeComplete()` must reflect the native TLS state. Add a regression that enters through the real adapter/parser seam rather than manually assigning `Http2PendingStream.transport_early`.

src/edge_gateway.zig
```diff
@@ -1893,11 +2013,24 @@ fn handleHttp2Connection(conn: anytype, session: *ConnectionSession, cfg: *const
...
+            &next_server_stream_id,
+            &conn_send_window,
+        );
+
         var frame = http.http2_frame.readFrame(conn, allocator, HTTP2_MAX_FRAME_SIZE) catch |err| switch (err) {
             error.ConnectionClosed => return,
             else => return err,
         };
         defer http.http2_frame.deinitFrame(allocator, &frame);
+        const frame_transport_early = h2LastReadTransportEarly(conn);
```
Jaruso, line 2105: **Blocking:** sampling `lastReadTransportEarly()` once *after* `readFrame()` cannot satisfy #367's sticky rule. `http2_frame.readFrame()` performs a separate exact read for the 9-byte frame header and another exact read for the payload, and either exact read may itself call `conn.read()` multiple times. If any earlier read was accepted 0-RTT but the final read was 1-RTT, this value is false and the request is misclassified ordinary.

The provenance needs to be OR-accumulated across every plaintext read contributing to the frame/request, with the accumulator reset only at an explicit parsing boundary. For example, expose a `takeFrameTransportEarly()`/read-scope accumulator that does roughly `scope_early |= read_result.transport_early`, then consume it after the frame is assembled. Please add a fragmented regression such as early frame-header/first payload bytes followed by 1-RTT remainder and assert the stream remains `transport_early = true`.

src/edge_gateway.zig
```diff
@@ -1857,6 +1878,105 @@ fn parkedConnectionCloseHook(raw_state: *anyopaque, fd: std.posix.fd_t) void {
...
+        if (ps.transport_early and !h2DownstreamHandshakeComplete(conn)) {
+            if (!ps.early_request_recorded) {
+                state.metricsRecordEarlyDataRequest(.h2, .transport);
+                ps.early_request_recorded = true;
+            }
+            if (!ps.deferred_recorded) {
+                state.metricsRecordEarlyDataDecision(.h2, .deferred);
+                ps.deferred_recorded = true;
+            }
+            try h2AppendReadyStream(ready_streams, sid);
```
Jaruso, line 2012: **Blocking:** re-appending the stream is not enough to make handshake completion dispatch it. After this returns, `handleHttp2Connection()` proceeds back into `readFrame()`. TLS can consume the peer Finished and transition to complete without producing another H2 application frame; if the client sends nothing else, the server waits for more application bytes and this ready stream never gets another `h2DispatchReadyStreams()` call.

Please give deferred-ready H2 work an explicit handshake-completion wake/drive path that re-enters dispatch without requiring a subsequent H2 frame. This should remain nonblocking/no-busy-loop and integrate with the native event-loop/TLS readiness machinery. The key regression is: submit a complete early request, observe dispatch count 0, complete TLS, send **no additional H2 application frame**, and observe dispatch count exactly 1.

docs/SUPPORT_MATRIX.md
```diff
@@ -62,6 +62,7 @@ surfaces that should not be marketed as generic operator-facing capabilities.
 | --- | --- | --- | --- |
 | HTTP/2 | `tls_termination` HTTP/2 path, `hpack`, `http2_frame` | `experimental` | Present in-tree, but not documented or release-gated at the same level as the HTTP/1.1 core. |
 | HTTP/3 / QUIC | `http3_handler`, `http3_session`, `http3_runtime`, `quic`, `http3` | `experimental` | Served by the native Zig QUIC/H3 stack (no system libraries); enable at runtime with `--http3` and a QUIC-compatible TLS identity (Ed25519 or ECDSA P-256). ngtcp2/nghttp3 removed under #328; they remain out-of-process interop peers only. |
+| Early-data replay handling across H1/H2/H3 | `early_data`, `request_context`, `edge_gateway` H1/H2 preflight/deferral, `http3_runtime` compatibility gate | `experimental` | Current scope: #367 HTTP-level policy, H2 safe deferral, and bounded observability (metrics/access logs). Out of scope here: #366 transport-carrier internals and #368 anti-replay persistence/store. |
```
Jaruso, line 65: **Blocking for the final #367 slice:** this row claims early-data replay handling across H1/H2/H3, but the H3 request path is still not integrated with the shared #367 HTTP policy. `http3_session.StreamRequest` has no transport-early provenance, and `routeHttp3Location()` still goes from pure route resolution directly into auth/action execution without constructing an `EarlyDataContext` or running the shared early-data decision. That also means the new `protocol="h3"` request/decision metrics are not populated by production H3 request handling.

This needs to close the H3 request-execution gap before #367 can be considered complete. At minimum: preserve/inject #366 sticky transport provenance into the H3 stream request, derive `inbound_marker` from any `Early-Data` header, run the common replay-safety decision immediately after pure route resolution and before auth/action, emit/forward 425 semantics as already defined by #367, and record the bounded H3 request/decision metrics. Please include a prior-hop `Early-Data` test that works even over ordinary 1-RTT H3, plus an injected transport-provenance test and a zero-side-effect rejection assertion.

docs/OBSERVABILITY.md
```diff
@@ -11,7 +11,18 @@ Runtime logs and access logs are JSON by default.
...
+  `identity`, `bytes_sent`, `response_bytes`, `error_category`,
+  `early_data_source`, `early_data_action`, `early_data_retry_result`, and
+  `early_data_replay_exposed`
+
+Early-data access-log fields are bounded enums/booleans only:
+
+- `early_data_source`: `none`, `transport`, `header`, `both`
+- `early_data_action`: `ordinary`, `accepted`, `forwarded`, `too_early`,
+  `deferred`, `retried`
+- `early_data_retry_result`: `none`, `success`, `too_early`, `failed`
```
Jaruso: Minor docs mismatch: the runtime enum/log label is `failure`, but this documents `failed`. Please make the operator contract match the emitted value (`none | success | too_early | failure`) and add/retain a serialization assertion so the docs cannot drift from the bounded enum again.

Jaruso (commented):
Blocking review for #367 Slice 4. CI #923 is green, and the bounded metrics/access-log machinery is generally well structured, but this is not ready to merge yet.

The blocking issues are:

1. The H2 production adapter does not currently consume real #366 transport provenance or real handshake-completion state; the new booleans are effectively test-only.
2. Even with that seam connected, sampling provenance once after `readFrame()` loses the required sticky bit when an H2 frame spans multiple TLS/plaintext reads with mixed early and 1-RTT bytes.
3. Deferred H2 streams are requeued, but handshake completion itself does not wake/re-enter dispatch. A complete early request can remain parked forever if the client sends no additional H2 application frame after Finished.
4. This is intended to be the final #367 slice, but H3 request-level RFC 8470 integration is still absent. The H3 path does not construct/consume `EarlyDataContext` before auth/action execution, so prior-hop `Early-Data` requests are not gated and the new H3 request/decision metric series are not populated by production request handling.

Please add end-to-end regression coverage at the actual parsing/dispatch seams rather than only constructing `Http2PendingStream` with `transport_early = true`. In particular, prove fragmented mixed early/1-RTT H2 input remains sticky, prove TLS handshake completion dispatches a deferred ready stream without requiring another H2 frame, and prove H3 prior-hop marker/current-hop injected provenance hits the shared policy before side effects.

Once these are fixed I’ll re-review the full Slice 4 surface against #367 again.

src/http/native_tls_connection.zig
```diff
@@ -251,7 +251,23 @@ pub const NativeTlsConnection = struct {
...
+            nativeHandshakeComplete,
+        );
+    }
+
+    pub fn readTransportEarly(self: *const NativeTlsConnection) bool {
+        // #366 server-side provenance: accepted 0-RTT contributes replay
+        // exposure until the connection reaches authenticated application data.
+        return self.backend.earlyDataAccepted() and !self.record.applicationDataOpen();
+    }
+
```
Jaruso, line 266: **Blocking:** this still infers transport provenance from handshake state, which #367 explicitly forbids. `earlyDataAccepted()` says the connection negotiated/accepted 0-RTT; it does **not** say that the plaintext returned by this specific read came from a 0-RTT record. A decrypted early record can remain buffered until after Finished completes, at which point this returns false and the early bytes are mislabeled ordinary. Conversely, this predicate is not an epoch tag for the bytes being drained.

Please consume/fix #366's actual record-carrier provenance: preserve the record epoch alongside delivered plaintext (or expose an equivalent per-read provenance result) and have `currentReadTransportEarly()` report the provenance of the bytes actually returned. The H2 read-scope OR accumulator added in this revision is the right HTTP-side shape once the underlying per-read signal is real.

src/quic/connection.zig
```diff
@@ -1823,6 +1823,10 @@ pub const Connection = struct {
         return self.accept_queue.orderedRemove(0);
     }
 
+    pub fn streamTransportEarly(self: *const Connection, _: StreamId) bool {
+        return self.tls.earlyDataAccepted() and !self.handshake_complete;
+    }
+
```
Jaruso, line 1834: **Blocking:** this has the same handshake-state/provenance conflation on QUIC. #366 requires a sticky per-stream bit set when an accepted `.zero_rtt` STREAM fragment is processed, before feeding those bytes into ordinary stream reassembly. A delayed 0-RTT packet may be processed after handshake completion, and later 1-RTT fragments must not clear the bit.

This implementation also cannot work with the current H3 runtime: `pumpH3()` returns until `conn.isEstablished()`, so H3 first calls `acceptStream()` only after `handshake_complete` is true and this function returns false by construction. Store bounded sticky provenance with the QUIC stream state and return that value by `stream_id`; do not infer it from current handshake state.

src/gateway_handlers.zig
```diff
@@ -1874,6 +1883,48 @@ fn routeHttp3Location(
...
+        request_path,
+        request.body.len != 0,
+    );
+    switch (decision) {
+        .execute_local => {
+            ctx.state.metricsRecordEarlyDataDecision(.h3, .accepted);
+        },
+        .forward_rfc8470 => {
+            ctx.state.metricsRecordEarlyDataDecision(.h3, .forwarded);
+        },
```
Jaruso, line 2430: **Blocking:** this records `.forwarded`, but the H3 proxy path does not actually forward the RFC 8470 marker. `handleHttp3LocationProxyPass()` still passes a hard-coded `false` as `forward_early_data` to `executeBufferedDataPlaneProxyRequest()`. The shared proxy header copier intentionally strips every raw inbound `Early-Data`, so both prior-hop markers and approved current-hop H3 early requests reach the origin with **no** `Early-Data: 1` despite this decision saying `forward_rfc8470`.

Thread the authoritative early-data forwarding decision/context into the H3 proxy call and pass `true` for approved RFC8470 forwarding so the shared canonicalizer emits exactly one `Early-Data: 1`. Add an H3 proxy regression proving inbound invalid/duplicate/prior-hop markers normalize to one `1`, approved current-hop forwarding adds one marker, and ordinary H3 adds none.

src/gateway_handlers.zig
```diff
@@ -1874,6 +1883,48 @@ fn routeHttp3Location(
...
+        .transport_early = request.transport_early,
+        .inbound_marker = request.headers.hasEarlyDataMarker(),
+    };
+    if (metricsEarlyDataSource(early_ctx)) |source| {
+        ctx.state.metricsRecordEarlyDataRequest(.h3, source);
+    }
+
+    const method = http.Method.parse(request.method) orelse .GET;
+    const decision = earlyDataDecisionForRequest(
+        ctx.cfg,
```
Jaruso, line 2417: **Blocking:** `orelse .GET` turns every unknown H3 method into GET for the early-data safety decision. #367's initial gate is explicitly GET/HEAD only and says unknown/unsafe replay-exposed methods must return 425. As written, `FOO /safe` + `Early-Data` on an explicitly replay-safe local route can be classified like GET and execute.

Please fail closed without fabricating a method. One clean option is to let the shared early-data decision consume the raw method / an explicit `method_safe` boolean, so `Method.parse(...) == null` maps to unsafe for replay-exposed requests while ordinary non-early H3 behavior remains unchanged. Add `FOO` (or another extension method) + replay exposure => 425 with zero side effects.

Jaruso (commented):
Re-review of head `ac4d8971de74d7f72623b8b34cf954a5aba7b8bb` against the full #367 Slice 4/final-story contract.

The previous H2 frame-scope accumulation issue is fixed: provenance is now OR-accumulated across all reads contributing to a frame, with a fragmented early→1-RTT regression. The deferred-ready H2 wake-up is also present with a no-extra-H2-frame regression, and the observability `failure` documentation mismatch is fixed.

This is still **not ready to merge**. Remaining blockers:

1. The production TCP and QUIC provenance sources still infer "early" from handshake state (`earlyDataAccepted() && !handshake_complete/applicationDataOpen()`) instead of preserving the epoch/packet provenance of the bytes actually delivered. That directly violates #367's dependency contract on #366 sticky provenance. On H3 this is especially fatal because `pumpH3()` does not run until `Connection.isEstablished()`, so the new `streamTransportEarly()` predicate is false by construction when H3 first asks for it.
2. H3 `.forward_rfc8470` currently records a `forwarded` metric but the proxy call still receives `forward_early_data = false`. Since the shared header copier intentionally strips raw inbound `Early-Data`, approved H3 proxy requests lose the required canonical `Early-Data: 1` marker.
3. H3 unknown methods are coerced to `GET` for the early-data decision. #367 explicitly requires unknown/unsafe replay-exposed methods to fail closed with 425; an extension method can currently inherit GET's early-safety result.
4. CI #927 is red on this head (format, unit/build jobs across platforms/profiles). The formatting failure is visible in the new `http3_runtime.zig` assignment indentation, and the build/test failures also need to be cleared before merge.

I left detailed inline comments on the three semantic blockers. CI must be green after the fixes.

src/quic/connection.zig
```diff
@@ -1030,6 +1029,9 @@ pub const Connection = struct {
                     self.closeOnStreamError(err, now_us);
                     return;
                 };
+                if (level == .zero_rtt) {
```
Jaruso: **Blocking:** the sticky map fixes the old handshake-state inference, but this revision also accidentally enables part of the #366 QUIC carrier without the #366 safety contract. `ingestPacket()` now accepts `.zero_rtt` through deprotection and calls `recovery.onPacketReceived()` / builds ACK state before any `tls.earlyDataAccepted()` gate; there is no acceptance check at all. That violates #366's requirement that rejected 0-RTT is neither processed nor ACKed.

There is a second runtime failure here: genuine accepted 0-RTT STREAM data arrives **before** handshake completion, but `self.streams` is only initialized in `afterHandshakeProgress()` after `handshake_complete = true`. So this branch reaches `streamManager() orelse ... "stream before handshake"` and closes the connection. The new test does not catch that because it calls `applyFrame(.zero_rtt, ...)` directly *after* `pair.pump()` has completed the handshake, bypassing packet keys, acceptance, ACK state, and the pre-handshake manager lifecycle.

The issue also requires the sticky mark to happen before ordinary stream reassembly; this currently calls `receiveStreamFrame()` first and marks afterward.

Given PR #501 explicitly says #366 carrier internals are out of scope, the expected fix is to restore the previous fail-closed `.zero_rtt` packet drop and keep only a narrow bounded `markStreamZeroRtt`/`streamTransportEarly` seam that #366 can invoke once it has actually accepted a packet. If this PR intentionally takes over the carrier instead, it needs the full #366 contract: acceptance gate before recovery/ACK bookkeeping, rejected packets never ACKed/processed, pre-handshake stream admission/flow-control from remembered transport state, mark-before-reassembly, and real packet-ingest tests for accepted/rejected 0-RTT.

src/gateway_handlers.zig
```diff
@@ -1874,6 +1890,48 @@ fn routeHttp3Location(
...
+    }
+
+    const decision = earlyDataDecisionForRawMethod(
+        ctx.cfg,
+        early_ctx,
+        request.method,
+        request_path,
+        request.body.len != 0,
+    );
+    const forward_early_data = decision == .forward_rfc8470;
```
Jaruso: **Blocking:** `forward_rfc8470` is the *policy decision that this route may be forwarded*, not the RFC 8470 condition for actually emitting `Early-Data: 1`. #367 defines the first-attempt marker separately as:

```zig
inbound_marker or (transport_early and !downstream_handshake_complete)
```

This matters immediately because `http3_runtime.pumpH3()` still returns until `conn.isEstablished()`. If a request contains sticky current-hop provenance but is dispatched only after Finished, this code now emits `Early-Data: 1` even though the downstream handshake is already complete. Inbound/prior-hop markers, by contrast, must always remain marked.

Please carry an explicit downstream-handshake completion/barrier signal into the H3 request context and compute the forwarding marker from the same rule H1 uses. Do **not** use handshake state to derive `transport_early`; use it only to decide whether the current-hop marker is required at forwarding time.

Add H3 regressions for: prior-hop marker -> exactly one outbound `Early-Data: 1`; current-hop-only + handshake incomplete -> one marker; current-hop-only + handshake complete -> no marker; ordinary request -> no marker.

Jaruso (commented):
Re-review of head `196b57bdb4e0a3ef08a39d3be3cf25e79e47da50`.

CI #928 is green, and the previous unknown-method, direct H3 marker plumbing, H2 fragmented-read provenance accumulation, H2 handshake wake-up, and observability-doc issues are fixed.

Two blockers remain before merge:

1. The QUIC fix now crosses into #366 carrier behavior by accepting `.zero_rtt` into the ordinary packet/frame path without the required acceptance/ACK/pre-handshake stream-manager semantics. This is unsafe and also cannot process genuine accepted pre-handshake STREAM data with the current `StreamManager` lifecycle. Given this PR explicitly excludes #366 carrier internals, the safest correction is to restore the fail-closed 0-RTT packet behavior here and leave a narrow sticky-provenance mark/query seam for #366.
2. H3 currently equates `forward_rfc8470` policy approval with actually emitting `Early-Data: 1`. #367 explicitly separates those: the first-attempt marker is `inbound_marker || (transport_early && !downstream_handshake_complete)`. Current H3 dispatch happens only after `isEstablished()`, so a current-hop-only request whose provenance survives until dispatch must not gain a marker merely because its route is RFC8470-capable.

Detailed fixes/tests are on the inline comments. After those are corrected, I do not currently see another Slice 4 blocker in the reviewed paths.

src/gateway_handlers.zig
```diff
@@ -1712,7 +1740,7 @@ fn handleHttp3LocationProxyPass(
         null,
         null,
         null,
-        false,
+        forward_early_data,
```
Jaruso, line 2231: **Blocking:** this now passes the correct first-attempt marker bit, but the H3 proxy path still performs only one low-level `executeBufferedDataPlaneProxyRequest()` call. That bypasses #367's upstream-425 state machine entirely.

For a current-hop-only early H3 request forwarded before the downstream handshake completes, an upstream 425 must be eligible for **exactly one** retry after the downstream handshake completes; that retry must be ordinary/1-RTT with no `Early-Data` marker. An inbound/prior-hop marker must never be retried. The current H3 path cannot do any of that, and therefore also cannot populate the new upstream-425/retry metrics for H3 behavior.

Please factor/reuse the existing `runBufferedProxyAttempts` semantics (or an equivalent protocol-neutral orchestration layer) rather than duplicating the state machine. H3 also needs a **live** downstream-handshake barrier capable of querying/driving completion; the new `StreamRequest.downstream_handshake_complete` snapshot plus `h3RequestDriveHandshake()` no-op is insufficient for the retry case.

Required H3 regressions: current-hop-only + forwarded marker + upstream 425 -> wait/drive handshake -> exactly one ordinary retry; second 425 is forwarded/no loop; inbound marker + upstream 425 is forwarded with zero retry; retry success/425/failure metrics are recorded correctly.

src/gateway_handlers.zig
```diff
@@ -1874,6 +1902,53 @@ fn routeHttp3Location(
         .location => |m| m,
         .reload_status, .metrics, .unmatched => return .not_handled,
     };
+
+    const early_ctx = http.request_context.EarlyDataContext{
```
Jaruso: **Blocking:** the H3 early-data preflight still starts too late for non-location route results. The `resolveRoutePath(...)` switch immediately above returns `.not_handled` for `.reload_status`, `.metrics`, and `.unmatched`, so replay-exposed H3 requests in those cases never construct this `EarlyDataContext`, never call `earlyDataDecisionForRawMethod()`, and fall through to the ordinary 404 path.

That contradicts #367's shared decision table: replay-exposed requests to routes that are not explicitly replay-safe (including operational/unmatched outcomes) must produce 425, not silently take the ordinary path. It also means `http_early_data_requests_total{protocol="h3"...}` / decision metrics are skipped for those requests.

Keep the pure route result first, but run the common early-data decision against **all** route outcomes before returning/dispatching. Only ordinary non-early H3 requests should preserve the existing `.not_handled`/404 behavior. Add regressions for inbound `Early-Data` and injected transport provenance against an unmatched/operational path => 425, while ordinary H3 remains 404/not-handled.

Jaruso (commented):
Re-review of head `4a061a46`. CI #929 is green, and the two blockers from the prior pass are substantially addressed: QUIC is fail-closed again with only a sticky provenance seam, and H3 now separates RFC8470 forwarding permission from the actual `Early-Data: 1` marker rule.

Two blocking #367 gaps remain before this final slice is ready to merge:

1. H3 proxy execution still bypasses the shared upstream-425 retry state machine. Current-hop-only early H3 forwarded before handshake completion must be eligible for exactly one ordinary retry after the downstream handshake completes; prior-hop/inbound-marker 425 must never retry. The current H3 path is a single `executeBufferedDataPlaneProxyRequest()` call and the H3 handshake barrier is only a bool snapshot with a no-op drive callback, so the required retry cannot happen and the H3 425/retry metrics are not produced.
2. H3 returns `.not_handled` for reload-status/metrics/unmatched before constructing `EarlyDataContext` or invoking the shared decision. That bypasses the documented `replay-exposed + route not explicitly replay_safe -> 425` rule and skips H3 early-data observability for those requests.

Please address both with protocol-level regressions. Once those are fixed, I do not see another blocker in the latest delta.

src/http/http3_runtime.zig
```diff
@@ -609,6 +623,16 @@ pub const Runtime = struct {
...
     }
 
+    fn h3DownstreamHandshakeComplete(ctx: *anyopaque) bool {
+        const conn: *Connection = @ptrCast(@alignCast(ctx));
+        return conn.isEstablished();
+    }
+
+    fn h3DownstreamWaitOrDriveHandshake(ctx: *anyopaque) anyerror!void {
+        const conn: *Connection = @ptrCast(@alignCast(ctx));
+        conn.driveAuthentication(nowUs());
```
Jaruso, line 693: **Blocking:** this does not provide the live handshake barrier that the new shared 425 retry path requires. `Connection.driveAuthentication()` is explicitly a no-op unless `handshake.authPending()` is true; it only advances parked async signer/verifier/selector work. It cannot receive or process the peer Finished/CRYPTO for an ordinary incomplete QUIC handshake.

That becomes a dead end as soon as #366 allows accepted early H3 requests to reach the handler pre-establishment: `serveRequest()` is running synchronously on the same UDP-owner thread, the first upstream attempt can return 425, this callback runs once, and no peer datagram can be ingested while the handler owns the loop. `runBufferedProxyAttempts()` then sees `downstreamHandshakeComplete() == false`, records failure, and forwards the 425 instead of performing #367's required one post-handshake ordinary retry.

Please make this an event-loop continuation rather than a synchronous one-shot drive. Park the pending current-hop-only 425 retry with the request/stream state, return to the QUIC loop, and resume it exactly once when `conn.isEstablished()` becomes true. The resumed attempt must have `forward_early_data = false`; a second 425 is forwarded with no loop. A runtime-level regression should deliver the first upstream 425 before Finished, deliver Finished later with **no additional H3 application request bytes**, and assert exactly one ordinary retry after completion.

Jaruso (commented):
Re-review of the latest #501 head. The unmatched/operational H3 preflight blocker is fixed, and the H3 proxy path now correctly reuses the shared bounded retry state machine instead of duplicating it.

One blocking issue remains before #367 can be considered complete: the production H3 downstream-handshake barrier cannot actually wait for or drive an ordinary QUIC handshake to completion. The new callback only calls `Connection.driveAuthentication()`, which advances a parked asynchronous credential operation; it cannot receive/process the peer Finished while the request handler is synchronously executing on the same UDP runtime thread. Once #366 permits H3 request execution before establishment, a current-hop early request that receives upstream 425 will therefore fail the completion check and forward the 425 instead of performing the required one ordinary retry after handshake completion.

This needs an event-loop-safe continuation/deferred retry design rather than a synchronous one-shot callback: park the pending current-hop 425 retry, return control to the QUIC runtime so it can ingest handshake datagrams, then resume exactly once after `conn.isEstablished()`. The retry must carry no `Early-Data` marker and must not loop on a second 425. Add a runtime-level regression where the first upstream attempt receives 425 before Finished, Finished arrives later with no additional H3 application request data, and exactly one ordinary upstream retry occurs afterward.

CI #930 is still running at the time of this review; the completed jobs I inspected are green.

src/http/http3_runtime.zig
```diff
@@ -609,6 +672,38 @@ pub const Runtime = struct {
...
+        entry: *ConnEntry,
+        stream_id: u64,
+    };
+
+    fn parkH3Early425Retry(ctx: *anyopaque, request: *const http3_session.StreamRequest) anyerror!void {
+        const park_ctx: *H3ParkContext = @ptrCast(@alignCast(ctx));
+        var parked_request = try request.clone(park_ctx.runtime.allocator);
+        errdefer parked_request.deinit();
+        parked_request.stream_id = park_ctx.stream_id;
+        parked_request.transport_early = false;
```
Jaruso: **Blocking:** do not clear `transport_early` here. #367 defines this as sticky request provenance: once any byte contributing to the request was accepted as early data, the request remains replay-exposed for its lifetime. Parking only changes *when/how the upstream retry is sent*; it does not turn the inbound request into an ordinary request.

This is security-significant because `resumeParkedH3Retries()` re-enters the full request handler after the original config lease has been released. With this assignment, a config reload between park and resume can change `/safe` from a replay-safe RFC8470 proxy to a non-replay-safe/local side-effectful route, and the resumed request now sees `replayExposed() == false` and can execute that new route. Preserve `transport_early = true`; use `resume_early_425_retry` only to ensure the resumed **upstream attempt** is sent ordinary (`forward_early_data = false`) and to suppress duplicate request/decision metrics if needed.

Please add a regression that parks a current-hop early request, changes the route to non-replay-safe before resume, then asserts the resumed path still returns 425/executes zero side effects. Also assert the actual post-handshake upstream retry carries no `Early-Data` marker.

src/http/http3_session.zig
```diff
@@ -34,6 +50,33 @@ pub const StreamRequest = struct {
...
+    pub fn clone(self: *const StreamRequest, allocator: std.mem.Allocator) !StreamRequest {
+        var headers = Headers.init(allocator);
+        errdefer headers.deinit();
+        for (self.headers.iterator()) |header| {
+            try headers.append(header.name, header.value);
+        }
+        return .{
+            .allocator = allocator,
+            .stream_id = self.stream_id,
+            .method = try allocator.dupe(u8, self.method),
```
Jaruso: Ownership bug in the new clone helper: these `dupe()` calls are evaluated incrementally, but only `headers` has an `errdefer`. If `path`, `authority`, or `body` allocation fails after `method` (and possibly other fields) succeeded, those earlier allocations leak because no `StreamRequest` is ever returned for the caller to deinit.

Please construct the owned fields into locals with matching `errdefer allocator.free(...)` (or otherwise make clone transactional) before returning the struct. An allocation-failure test around `StreamRequest.clone()` would be useful here.

Jaruso (commented):
Re-review of the latest #501 head. The previous H3 425 problem is substantially fixed: the request can now be parked, control returns to the QUIC owner loop, and parked work is resumed after establishment rather than synchronously trying to manufacture handshake progress. CI #931 is also green.

I found one remaining replay-safety blocker in the parked request state and one new ownership bug in the clone helper:

1. **Sticky transport provenance is erased while parking.** #367 defines `transport_early` as a sticky semantic property of the request: once any contributing byte was accepted early, that request remains replay-exposed. Setting it to false makes the resumed request look ordinary when the handler/routing stack is entered again. Because the config lease is released while parked, a reload can turn the original replay-safe proxy route into a non-replay-safe/local route and the resumed early request can then execute it without the 425 gate. Keep the inbound provenance sticky; make only the *outbound retry attempt* ordinary/no-marker. Suppress duplicate observability explicitly via the resume state rather than by changing the security fact.
2. **`StreamRequest.clone()` leaks partial allocations on OOM.** The new struct-literal duplication allocates method/path/authority/body incrementally but only `headers` has an `errdefer`. A later allocation failure leaks the earlier successful dupes. Build the clone incrementally (or add errdefers for each owned field) and ideally cover it with allocation-failure testing.

After the provenance fix, please add a regression that parks a current-hop early request, changes/re-resolves to a route that is no longer replay-safe before resume, and proves the resumed request is still treated as replay-exposed and cannot execute side effects. The actual second upstream attempt must remain ordinary with no `Early-Data` marker.

src/http/http3_runtime.zig
```diff
@@ -569,16 +591,50 @@ pub const Runtime = struct {
...
             self.logger.warn(null, "http3: request handler failed: {s}", .{@errorName(err)});
             entry.h3.sendResponse(entry.conn, incoming.stream_id, 500, &.{}, "") catch {};
             entry.h3.finishRequest(incoming.stream_id);
             return;
         };
 
+        self.sendHandlerResponse(entry, incoming.stream_id, &response);
+    }
+
+    fn resumeParkedH3Retries(self: *Runtime, entry: *ConnEntry) void {
```
Jaruso, line 621: **Blocking:** `resumeParkedH3Retries()` needs its own production `entry.conn.isEstablished()` gate. Right now it is safe only because `pumpH3()` returns before doing *any* H3 work while the handshake is incomplete. But #366 must relax/restructure that outer guard so accepted 0-RTT H3 request bytes can be parsed and dispatched pre-establishment—the exact condition under which this queue is populated. Once that happens, this unconditional loop can immediately resume the parked ordinary retry before Finished.

Make the continuation invariant local and explicit, e.g. `if (!entry.conn.isEstablished()) return;` at the top of this function (not a debug assert). Then add a runtime/seam regression: queue a continuation while incomplete, call the H3 pump/continuation wake while still incomplete and assert zero resume calls; transition to established with no additional H3 request bytes and assert exactly one resume.

src/http/http3_runtime.zig
```diff
@@ -569,16 +591,50 @@ pub const Runtime = struct {
...
+        while (entry.parked_h3_retries.items.len != 0) {
+            var parked = entry.parked_h3_retries.orderedRemove(0);
+            defer parked.deinit(self.allocator);
+
+            var response = response_mod.Response.init(self.allocator);
+            defer response.deinit();
+            parked.continuation.run(self.allocator, &response) catch |err| {
+                self.logger.warn(null, "http3: parked request retry failed: {s}", .{@errorName(err)});
+                entry.h3.sendResponse(entry.conn, parked.stream_id, 500, &.{}, "") catch {};
```
Jaruso: **Blocking lifecycle bug:** if `continuation.run()` fails and this fallback `sendResponse(..., 500)` also fails (for example `StreamBackpressure`), the error is swallowed and the request is never removed/reset. The parked continuation has already been removed/deinitialized, but the finished request remains resident in the H3 request map until connection teardown.

Use the same terminal-send discipline as `sendHandlerResponse()`: either build a 500 `Response` and send it through that helper, or on `sendResponse` failure explicitly reset the stream and call `finishRequest(stream_id)`. Add a failure-path test proving the stream is released exactly once even when the fallback response cannot be queued.

src/gateway_proxy_runtime.zig
```diff
@@ -878,17 +884,44 @@ fn runBufferedProxyAttempts(
             method,
             matched_block,
         )) {
+            if (!early_ctx.downstreamHandshakeComplete()) {
+                if (maybeParkEarly425Retry(attempt_executor, &result)) |parked| {
+                    if (parked) return .early_425_retry_parked;
+                } else |err| {
+                    attempt_executor.recordEarlyRetryResult(.failure);
+                    try attempt_executor.onEarly425HandshakeFailure(err);
+                    return .{ .response = result };
```
Jaruso, line 912: **Blocking observability gap:** every return in these parking/handshake-failure branches forwards the *original upstream 425* downstream, but they return before the normal `status == 425` block that records `EarlyDataUpstream425Action.forwarded`. The result is `retry{failure}` with no corresponding `upstream_425{action="forwarded"}` even though a 425 is actually forwarded.

Before returning the preserved `result` on park failure, handshake-drive failure, or still-incomplete/no-park failure, record `.forwarded` exactly once. Extend the parking-failure and handshake-failure regressions to assert both `retry=failure` and `upstream_425 action=forwarded`.

src/gateway_proxy_runtime.zig
```diff
@@ -878,17 +884,44 @@ fn runBufferedProxyAttempts(
...
             retry_state.consumeEarly425Retry();
             try attempt_executor.onEarly425Retry(&result);
             continue;
         }
+        if (result.statusCode() == @intFromEnum(http.Status.too_early)) {
+            attempt_executor.recordEarlyUpstream425Action(.forwarded);
+            if (retry_state.early_425_retry_used) {
+                attempt_executor.recordEarlyRetryResult(.too_early);
+            }
+        } else if (retry_state.early_425_retry_used) {
```
Jaruso: **Blocking observability/state-machine bug:** `early_425_retry_used` stays true for the rest of the configured retry loop, so this records `success` (or the 425 branch records `too_early`) on *every later configured response*, not exactly once for the automatic ordinary early-425 retry. Conversely, if that ordinary retry terminates in `ProxyBudgetExhausted`, `UpstreamAtCapacity`, cancellation, or another terminal transport error in the execute catch above, no `retry=failure` is recorded at all.

Track the result of the automatic early-425 retry separately from configured retries—e.g. an `early_425_retry_result_recorded` bit / explicit phase. Record exactly one terminal result for the first ordinary attempt caused by the 425: non-425 response => `success`; 425 => `too_early`; terminal/no-response error => `failure`. Later configured retries must not increment this metric again. Add regressions for ordinary retry transport failure and for an ordinary retry returning 5xx followed by a configured retry, asserting the early-data retry metric increments exactly once.

Jaruso (commented):
Re-review of `fd08c906` after the proxy-continuation rewrite. The core architecture is now correct: the parked H3 work resumes the already-approved proxy operation, retains the config version, preserves the original timeout budget, sends the second attempt without `Early-Data`, and the old `StreamRequest.clone()` ownership problem is gone. CI #932 is green.

I found four remaining blockers in the edge/error paths below. These are the complete issues from this pass:

1. The parked-retry runner still relies on `pumpH3()`'s current `isEstablished()` guard instead of enforcing the handshake barrier itself. That guard is exactly what #366 must relax to admit H3 early request bytes, so the continuation would otherwise become runnable too early when the carrier lands.
2. A continuation failure can leave the H3 request stream resident if writing the fallback 500 fails.
3. When parking/handshake completion fails and the original upstream 425 is forwarded, the `forwarded` action metric is skipped.
4. The shared retry-result metric is not exactly-once: terminal errors after the ordinary early-425 retry miss `failure`, while subsequent configured responses can increment `success`/`too_early` more than once.

Please fix these together and add regressions for the failure paths. I would not merge #501 until they are covered.

src/gateway_proxy_runtime.zig
```diff
@@ -847,6 +865,7 @@ fn runBufferedProxyAttempts(
     var attempt: usize = 0;
     while (attempt < retry_state.loopBound(max_attempts, stale_conn_retries)) : (attempt += 1) {
         const resp = attempt_executor.execute(attempt, retry_state.forwardEarlyData(first_attempt_forward_early_data)) catch |err| {
+            retry_state.recordEarly425RetryResultOnce(attempt_executor, .failure);
             if (err == error.ProxyBudgetExhausted) return .retry_budget_exhausted;
             if (shouldRetryStaleUpstreamConnection(
```
Jaruso, line 873: **Blocking:** the automatic 425 retry is finalized as `failure` *before* this code checks the transparent stale-connection exception. `shouldRetryStaleUpstreamConnection()` explicitly defines `HttpConnectionClosing` as a pre-delivery close—the request never reached the origin—and production gives those retries their own `max_stale_conn_retries` budget. So `425 -> stale pooled socket -> fresh 200` currently records `retry{failure}` and clears `early_425_retry_result_pending`; the successful fresh delivery can no longer record `success`.

There is a second budget bug in the same accounting model: `configuredAttemptIndex()` subtracts `early_425_extra_attempts` but not `stale_conn_retries`, even though stale retries are documented as independent/transparent. With `max_attempts=2`, `425 -> stale -> fresh 500` can therefore consume the configured retry slot and suppress the configured attempt that should still be available.

Please keep stale recovery transparent in both dimensions. At minimum, do not call `recordEarly425RetryResultOnce(.failure)` until after the stale-retry branch declines the error, and exclude stale attempts when deriving configured-attempt consumption (or track configured attempts explicitly rather than inferring them from the raw loop index). Add regressions for:

```text
425 -> HttpConnectionClosing -> 200
=> retried=1, retry success=1, failure=0, markers [1,0,0]

425 -> HttpConnectionClosing -> 500 -> configured 200
=> the configured retry remains available
```

Also add the non-early control case `stale -> fresh 500 -> configured 200` so the independent stale budget cannot regress outside #367.

src/http/http3_runtime.zig
```diff
@@ -569,16 +591,59 @@ pub const Runtime = struct {
...
+            defer parked.deinit(self.allocator);
+
+            var response = response_mod.Response.init(self.allocator);
+            defer response.deinit();
+            parked.continuation.run(self.allocator, &response) catch |err| {
+                self.logger.warn(null, "http3: parked request retry failed: {s}", .{@errorName(err)});
+                self.sendInternalErrorResponse(entry, parked.stream_id);
+                return;
+            };
+            self.sendHandlerResponse(entry, parked.stream_id, &response);
```
Jaruso, line 642: **Blocking:** after terminally handling one failed continuation, this `return` abandons every later parked H3 stream in the same connection. Multiple early request streams can be parked before Finished. If the first resumed continuation fails and the client sends no more datagrams, there may be no later `pumpH3()` call to wake the remaining queue, so those requests can hang until connection teardown.

`sendInternalErrorResponse()` already completes or resets the current stream without closing the connection, so this path should continue draining the established queue rather than return. Add a two-entry regression where continuation #1 fails and is terminally handled, continuation #2 still runs/sends exactly once, both continuations deinit exactly once, and the parked queue ends empty. Cover both successful fallback-500 send and fallback-send/reset if practical.

src/gateway_proxy_runtime.zig
```diff
@@ -918,6 +974,28 @@ const ProductionBufferedProxyAttemptExecutor = struct {
...
+    fn recordEarlyUpstream425Action(
+        self: *ProductionBufferedProxyAttemptExecutor,
+        action: http.metrics.EarlyDataUpstream425Action,
+    ) void {
+        self.state.metricsRecordEarlyDataUpstream425(action);
+    }
+
+    fn recordEarlyRetryResult(
+        self: *ProductionBufferedProxyAttemptExecutor,
+        result: http.metrics.EarlyDataRetryResult,
```
Jaruso, line 994: **Blocking for #367 observability:** `early_data_action` becomes `.retried` only when the retry **succeeds**. If the real ordinary retry is attempted and ends in a second 425 or a transport failure, the request is logged with `early_data_action=forwarded` plus `early_data_retry_result=too_early|failure`. Those fields disagree about whether the retry happened.

The action and result are intentionally separate bounded dimensions in #367: `retried` describes that the automatic retry was actually attempted; `success|too_early|failure` describes its outcome. Set `ctx.early_data_action = .retried` when `recordEarlyUpstream425Action(.retried)` fires—this callback is invoked immediately before the real ordinary retry—and let `recordEarlyRetryResult()` set only `early_data_retry_result`. That also correctly leaves park/handshake setup failures as `forwarded`, because no retry attempt occurred.

Add access-log/context assertions for retry success, second-425, and retry transport failure so all three have `action=retried` with the corresponding result, while parking/handshake failure remains `action=forwarded`.

Jaruso (commented):
Re-review of ca18cd04. The four blockers from the previous pass are fixed and I resolved those threads. The continuation now has a local establishment gate, failed fallback sends reset/finish the stream, forwarded-425 accounting is present on park/handshake failure, and the semantic retry result has an exactly-once pending bit.

The fresh pass found three remaining issues in the newly shared retry/continuation behavior. Two are correctness/state-machine bugs; one is an operator-facing access-log inconsistency. CI #933 is also currently red: the general macOS `Run unit tests` job failed while Ubuntu and the other visible jobs passed. I could not retrieve the truncated failure tail reliably, so I am not guessing at the exact macOS failure.

Please address the three inline comments and get CI green before merge. I do not see another architectural #367 gap beyond these bounded fixes.

src/gateway_handlers.zig
```diff
@@ -1680,6 +1779,396 @@ fn finalizeHttp3Response(response: *http.Response) void {
...
+        const self: *Http3Early425ProxyContinuation = @ptrCast(@alignCast(ctx));
+        std.debug.assert(self.transport_early);
+        const per_attempt_timeout_ms = self.perAttemptTimeoutMs() catch |err| {
+            self.state.metricsRecordEarlyDataRetry(.failure);
+            if (err == error.OutOfMemory) return error.OutOfMemory;
+            try rejectHttp3ProxyErrorWithState(allocator, response, self.state, .gateway_timeout, "upstream_timeout", "Upstream request timed out", self.correlation_id);
+            return;
+        };
+        self.state.metricsRecordEarlyDataUpstream425(.retried);
+        var upstream_response = self.execute(per_attempt_timeout_ms) catch |err| {
```
Jaruso: **Blocking:** the parked H3 continuation still bypasses the transparent stale-connection recovery that the same #367 retry gets on the synchronous path. `runBufferedProxyAttempts()` treats `error.HttpConnectionClosing` specially *before* recording retry failure/upstream failure because it means the pooled keep-alive died before the request was delivered; it retries on a fresh connection using the independent stale budget. Here the parked continuation performs a single `self.execute(...)` and immediately records `early_data_retry=failure` and `recordUpstreamFailure(...)` for that same error.

That makes behavior depend on handshake timing:

```text
first early attempt -> 425 -> handshake already complete
    -> shared state machine -> stale pooled socket -> fresh connection -> 200

first early attempt -> 425 -> parked until Finished
    -> continuation -> stale pooled socket -> 502 + retry=failure + upstream health penalty
```

The latter is a regression: the stale socket never delivered the ordinary retry to the origin, so it should remain a transparent connection recovery, not consume/finalize the one semantic 425 retry or mark the origin failed.

Please keep the continuation intentionally free of **configured** retries, but give its single semantic ordinary retry the same bounded stale-connection recovery (currently 2) as `Http3BufferedProxyAttemptExecutor`. Prefer factoring/reusing the existing stale predicate rather than creating different rules. Recompute `perAttemptTimeoutMs()` before each fresh-socket attempt so the original parked request budget remains authoritative; every attempt must continue to use `forward_early_data=false`.

Required regression at the actual parked-continuation seam:

```text
current-hop early request -> upstream 425 -> park
Finished -> resume
first ordinary execute -> HttpConnectionClosing (pre-delivery)
fresh execute -> 200

assert:
- downstream 200
- upstream_425{retried}=1
- retry{success}=1, retry{failure}=0
- no Early-Data marker on either ordinary execute
- stale close does not call recordUpstreamFailure
- original timeout budget is not reset
```

Also cover stale recovery exhausted -> exactly one `retry=failure` and terminal error mapping, with the health failure recorded only after the transparent stale budget is exhausted.

Jaruso (commented):
Re-review of `09605ff88c77177a294dc7964fec4a01826e00c8` against the three remaining #367 blockers plus a fresh regression pass. CI #934 is green. The stale-budget accounting in `runBufferedProxyAttempts`, H3 parked-queue draining, and `early_data_action=.retried` timing are all fixed; I resolved those three threads.

One blocking production inconsistency remains in the parked H3 continuation: its post-handshake ordinary retry bypasses the transparent stale pooled-connection recovery used by the shared proxy attempt state machine. Details and exact regression are inline. Apart from that finding, I did not find another blocker in this revision.

PR #505, opened 2026-07-24
tls: add process-local early-data anti-replay store (#368 Slice 1)

Jaruso:
## Summary

Slice 1 of #368 only: the bounded process-local anti-replay store and its pluggable contract. Does **not** wire into the existing #366/#497 `EarlyDataReplayGate` seam in `tls13_backend.zig`, does **not** touch `resumption_runtime.zig`, and does **not** add config/metrics/docs — those are Slices 2 and 3.

- New `src/tls/early_data_replay.zig`:
  - `Key` = the existing `ticket_identity_fingerprint` (`[32]u8`); no raw ticket/PSK/binder/request material is ever stored.
  - `LocalStore.claim()` performs the duplicate check and insertion atomically under a single mutex acquisition — never a separate contains-then-insert.
  - Inclusive recording-window boundary: a claim exactly at `retain_until_unix_ms` is still a duplicate.
  - Capacity is fail-closed: an unexpired entry is never evicted to admit a new key; only genuinely expired entries are swept, and if the store is still full the claim is `rejected_capacity`.
  - Startup quarantine (`quarantine_duration_ms`) rejects claims as `unavailable` until it elapses; a detected backward wall-clock jump conservatively re-arms the same quarantine window rather than trusting the regressed elapsed time.
  - Bounded closed-enum `Observer`/`Event` seam, notified only after the store's mutex is released, so a re-entrant observer can't deadlock.
  - Generic `Store` contract (`claimFn: *const fn(*anyopaque, Claim) ClaimResult`) that a future distributed backend must satisfy with the same atomicity guarantee; `LocalStore.store()` adapts to it.
  - `Limits`/`hard_max_entries` validated against a hard ceiling (1,048,576), mirroring `session_cache.zig`'s convention.
- Registered the new module in `src/tls/root.zig` so `zig build test-tls` picks up its tests.

## Test plan

- [x] `zig build test-tls` — new deterministic unit tests: first claim, duplicate before/at/after the inclusive deadline, independent keys, capacity rejection, expired-entry reclamation at capacity, unexpired entries never evicted, startup quarantine, quarantine boundary, backward-clock re-arm, saturating-arithmetic overflow fail-closed, allocation-failure fail-closed (via `FailingAllocator`), reentrant-observer deadlock check, `Store` adapter, and the closed `Event` vocabulary across a multi-call sequence.
- [x] Concurrency tests: N threads racing the same key produce exactly one `accepted` and the rest `duplicate`; N threads with distinct keys all `accepted` and stay within capacity.
- [x] `zig build test` — full suite passes.
- [x] `zig fmt --check` on changed files.

Refs #368

### DIFF
--- a/src/tls/early_data_replay.zig
+++ b/src/tls/early_data_replay.zig
@@ -126,21 +126,36 @@ pub const LocalStore = struct {
     /// Highest wall-clock `now_unix_ms` this store has ever observed, used
     /// only to detect backward clock movement.
     high_water_unix_ms: u64,
+    /// Entries removed (duplicate-then-expired, or reclaimed at capacity)
+    /// since the last tombstone compaction. `AutoHashMapUnmanaged` deletes
+    /// leave a tombstone behind, so a long-lived store with heavy
+    /// expire/reclaim churn across diverse keys would otherwise let probe
+    /// length grow without bound even though logical occupancy never
+    /// exceeds `limits.max_entries`. See `noteRemovalLocked`.
+    removed_since_rehash: usize = 0,
 
+    pub const InitError = error{ InvalidLimits, OutOfMemory };
+
+    /// Reserves table capacity for the full configured `limits.max_entries`
+    /// up front so the ordinary claim hot path — insert, remove, and the
+    /// at-capacity reclaim — never depends on further allocator growth
+    /// (see `claimLocked`/`reclaimOneExpiredLocked`).
     pub fn init(
         allocator: std.mem.Allocator,
         limits: Limits,
         quarantine_duration_ms: u64,
         now_unix_ms: u64,
-    ) error{InvalidLimits}!LocalStore {
+    ) InitError!LocalStore {
         try limits.validate();
-        return .{
+        var self: LocalStore = .{
             .allocator = allocator,
             .limits = limits,
             .quarantine_duration_ms = quarantine_duration_ms,
             .quarantine_until_unix_ms = now_unix_ms +| quarantine_duration_ms,
             .high_water_unix_ms = now_unix_ms,
         };
+        try self.entries.ensureTotalCapacity(allocator, @intCast(limits.max_entries));
+        return self;
     }
 
     pub fn deinit(self: *LocalStore) void {
@@ -220,6 +235,7 @@ pub const LocalStore = struct {
                 return .duplicate;
             }
             _ = self.entries.remove(c.key);
+            self.noteRemovalLocked();
             expired_removed.* += 1;
         }
 
@@ -227,12 +243,13 @@ pub const LocalStore = struct {
             return self.insertLocked(c, event);
         }
 
-        // At capacity: prune only genuinely expired entries — an unexpired
-        // entry must never be evicted to admit a new key (that would turn
-        // memory pressure into a replay bypass) — and retry.
-        expired_removed.* += self.sweepExpiredLocked(now_unix_ms);
-
-        if (self.entries.count() < self.limits.max_entries) {
+        // At capacity: this claim only needs one free slot, so reclaim at
+        // most one genuinely expired entry — an unexpired entry must never
+        // be evicted to admit a new key (that would turn memory pressure
+        // into a replay bypass) — and retry. Allocation-free: capacity was
+        // fully reserved in `init`.
+        if (self.reclaimOneExpiredLocked(now_unix_ms)) {
+            expired_removed.* += 1;
             return self.insertLocked(c, event);
         }
 
@@ -241,6 +258,13 @@ pub const LocalStore = struct {
     }
 
     fn insertLocked(self: *LocalStore, c: Claim, event: *Event) ClaimResult {
+        // Always a no-op in practice: `init` reserves capacity for
+        // `limits.max_entries` and this is only reached when
+        // `entries.count() < limits.max_entries`, so removals (which give
+        // capacity back — see `hash_map.zig`'s `available` bookkeeping)
+        // keep enough headroom for one more insert without growing.
+        // Kept as a defensive, fail-closed guard rather than an unchecked
+        // `putAssumeCapacity` given how security-sensitive this store is.
         self.entries.ensureUnusedCapacity(self.allocator, 1) catch {
             event.* = .unavailable;
             return .unavailable;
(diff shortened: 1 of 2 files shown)


src/tls/early_data_replay.zig
```diff
@@ -0,0 +1,557 @@
...
+/// H3 runtime in a process (Slice 2 composition) — a thread/worker-local
+/// store is not an acceptable production path.
+pub const LocalStore = struct {
+    allocator: std.mem.Allocator,
+    limits: Limits,
+    quarantine_duration_ms: u64,
+    mutex: zig_compat.Mutex = .{},
+    entries: std.AutoHashMapUnmanaged(Key, Entry) = .empty,
+    observer: Observer = .{},
+    /// Normally `start_unix_ms + quarantine_duration_ms`; re-armed to
```
Jaruso, line 119: **Blocking — bound deletion/tombstone churn, not just live entry count.** #368 explicitly calls out the stdlib-table deletion/tombstone case in its hot-path rules. This store is long-lived and repeatedly removes expired keys; leaving `AutoHashMapUnmanaged` with no compaction/rehash strategy lets tombstones accumulate under diverse-key churn and can make lookup/insert probe cost degrade over time. Because this is one process-shared mutex, that slowdown serializes every worker using the replay store.

Tardigrade supports Zig 0.16, where the hash-map API has an in-place rehash facility. A reasonable shape is to reserve the configured bounded capacity at init and compact after bulk expiry (or use a purpose-built fixed-capacity open-addressing table if that gives a cleaner structural bound):

```zig
pub const InitError = error{ InvalidLimits, OutOfMemory };

pub fn init(...) InitError!LocalStore {
    try limits.validate();
    var store = LocalStore{ ... };
    try store.entries.ensureTotalCapacity(allocator, @intCast(limits.max_entries));
    return store;
}

// After expiry removals, compact tombstones with the pinned Zig 0.16
// rehash API (exact call signature as exposed by AutoHashMapUnmanaged),
// or make the bounded table structurally tombstone-safe.
```

Please also add a deterministic churn regression: initialize once, disable/fail further allocations, then run many fill → expire → reclaim → insert cycles and prove claims continue to work within the configured bound. That gives us evidence that the production hot path remains bounded rather than only testing a fresh map.

src/tls/early_data_replay.zig
```diff
@@ -0,0 +1,557 @@
...
+    /// memory pressure just stops the sweep early (a partial sweep is
+    /// always safe: it never evicts anything live).
+    fn sweepExpiredLocked(self: *LocalStore, now_unix_ms: u64) usize {
+        var victims: std.ArrayListUnmanaged(Key) = .empty;
+        defer victims.deinit(self.allocator);
+
+        var it = self.entries.iterator();
+        while (it.next()) |kv| {
+            if (kv.value_ptr.retain_until_unix_ms < now_unix_ms) {
+                victims.append(self.allocator, kv.key_ptr.*) catch break;
```
Jaruso: **Blocking — allocator failure is being converted into `rejected_capacity`, and this sweep allocates while holding the global store mutex.** If `victims.append()` fails, we `break`; when no victim was collected the caller sees the map still full and returns `.rejected_capacity`. The real condition is store/allocator unavailability, which #368 deliberately keeps distinct for fail-closed observability. With many expired entries this also builds an O(expired-count) scratch list while all replay claims are blocked on the same mutex.

We only need **one** free slot for the current claim, so this can be allocation-free:

```zig
fn reclaimOneExpiredLocked(self: *LocalStore, now_unix_ms: u64) bool {
    var victim: ?Key = null;
    var it = self.entries.iterator();
    while (it.next()) |kv| {
        if (kv.value_ptr.retain_until_unix_ms < now_unix_ms) {
            victim = kv.key_ptr.*;
            break;
        }
    }

    if (victim) |key| {
        _ = self.entries.remove(key);
        return true;
    }
    return false;
}
```

Then, at capacity, reclaim one genuinely expired record; if none exists, return `.rejected_capacity`. If you keep an allocating sweep instead, make it return an error/typed outcome and map OOM to `.unavailable` rather than capacity. Add a test that exercises expiry reclamation with allocations disabled after initialization so this path cannot silently regress.

Jaruso (commented):
Reviewed the complete PR against the current #368 contract, scoped specifically to **Slice 1 — store + contract**. I also checked the current head (`acad0602`) and CI; the CI run is green across formatting, unit tests/builds, appliance/native TLS coverage, and the other required jobs.

The core security semantics are good: the existing `[32]u8` ticket fingerprint is the only key material retained, duplicate check + insert are atomic under one mutex, the deadline boundary is inclusive, live records are never evicted for capacity, quarantine fails closed and re-arms on wall-clock rollback, observer callbacks happen outside the mutex, and the same-key concurrency test proves exactly one successful claim.

I found **two blocking Slice-1 boundedness/failure-path issues** that I would resolve before merging:

1. The chosen `AutoHashMapUnmanaged` lifecycle does not currently satisfy the ticket's explicit deletion/tombstone-churn requirement. This is a long-lived process-wide security store with repeated expiration/removal. We need a compaction/rehash strategy (or a purpose-built bounded table) so probe cost cannot degrade indefinitely while every native worker is serialized behind this mutex. Please add a deterministic churn/no-hot-path-allocation regression test as part of the fix.
2. `sweepExpiredLocked()` allocates a victim list while holding the process-wide mutex and swallows allocation failure as an early `break`. That can report `rejected_capacity` when the actual failure was store/allocator unavailability, defeating the distinct failure vocabulary that Slice 3 metrics will consume. Prefer an allocation-free reclamation path; only one expired slot is needed to admit the current claim.

The inline comments include concrete implementation shapes. Once those are fixed, I do not see another Slice-1 correctness gap in the code reviewed here. Slices 2/3 (TLS retention-deadline wiring, process composition, config, metrics/docs, fake distributed-store proof) correctly remain out of scope for this PR.

GitHub does not permit the authenticated author to submit a formal `REQUEST_CHANGES` review on their own PR, so this is submitted as a COMMENT review; treat the two inline findings as merge blockers.

Jaruso (commented):
Re-reviewed PR #505 at head `5e701f79` against #368 **Slice 1 — store + contract**, including the follow-up commit that addresses the two prior blocking findings.

Both blockers are resolved:

- The store now reserves its full configured hash-table capacity during `init`, tracks removals, and periodically performs an in-place `rehash` to bound tombstone/probe degradation under long-lived diverse-key churn.
- At-capacity expiry reclamation is now allocation-free and removes only one genuinely expired record, so allocator failure can no longer be silently misreported as `rejected_capacity` from the old victim-list sweep.
- The new 500-cycle churn regression disables further allocations after initialization and repeatedly fills/expires/reclaims/inserts distinct keys, providing direct evidence that the production claim path stays within the pre-reserved bound.

I also re-checked the surrounding Slice-1 invariants: same-key claims remain atomic under one mutex; the recording deadline is inclusive; unexpired entries are never evicted; startup quarantine and backward-clock handling remain fail-closed; observer callbacks still run after releasing the mutex; and only the `[32]u8` ticket fingerprint plus retention deadline are retained.

I do **not** see another Slice-1 correctness/security blocker in the updated code.

Current CI for this head is still finishing, but the unit-test jobs on macOS and ARM Linux, formatting, lint, and the completed build jobs are green at the time of this review. Final merge readiness is therefore code-review clean, subject to the remaining CI jobs completing successfully.

GitHub does not allow the authenticated PR author to self-approve, so this is recorded as a COMMENT review rather than `APPROVE`.

PR #506, opened 2026-07-25
tls: wire early-data replay store into TLS/composition (#368 Slice 2)

Jaruso:
## Summary

Slice 2 of #368 — TLS/composition integration for the process-local anti-replay store landed in Slice 1 (#505).

- Extends `tls13_backend.EarlyDataReplayCandidate` with an authoritative `retain_until_unix_ms` deadline: `expected_arrival_time = ticket_issued_at + apparent_ticket_age`, `retain_until = expected_arrival_time + age_skew_tolerance`, computed via checked arithmetic from the already-validated ticket-age/freshness inputs. Overflow or impossible timing fails closed for 0-RTT (`.replay_unavailable`) *before* the replay gate is ever consulted — never a silently shortened retention window.
- Adds `early_data_replay.GateAdapter`, adapting any `Store` implementation (the bounded `LocalStore` today, a future distributed backend later) to the `EarlyDataReplayGate` seam. Maps `accepted -> allow`, `duplicate -> replay`, and `rejected_capacity`/`unavailable` -> `unavailable` — only a proven duplicate is reported as replay; everything else fails closed without touching 1-RTT resumption.
- `edge_gateway.zig` constructs exactly one process-scoped `LocalStore`/`GateAdapter` pair at the composition root (mirroring #488's `resumption_runtime` ownership pattern) and installs the resulting gate into every native TCP worker (`native_tls_connection.zig`) and the QUIC/H3 runtime (`http3_runtime.zig`, via a new `quic.tls_backend.Tls13Backend.setEarlyDataReplayGate` passthrough) — independent of whether native resumption itself is wired for a given connection, so worker-local duplicate 0-RTT acceptance is architecturally impossible.
- No config, metrics, deployment docs, or real distributed backend — that's Slice 3. No replay state added to `resumption_runtime.zig`.

## Test plan

- [x] `computeReplayRetainUntilUnixMs`: deadline arithmetic and boundary/overflow fail-closed (inline unit tests in `tls13_backend.zig`)
- [x] `GateAdapter`/`mapClaimResult`: allow/duplicate/unavailable mapping, including capacity exhaustion and startup quarantine (`early_data_replay.zig`)
- [x] Full handshake: replay gate receives the computed retention deadline; deadline overflow rejects 0-RTT without ever consulting the gate, preserving 1-RTT resumption (`tls13_backend_tests.zig`)
- [x] Two independent `Tls13Backend` "workers" sharing one real `LocalStore` reject the second worker's duplicate 0-RTT claim while both complete ordinary resumption (`tls13_backend_tests.zig`)
- [x] Native TCP (`NativeTlsConnection`) and QUIC/H3 (`http3_runtime.Runtime`) each install the shared gate independently of the resumption runtime (`native_tls_connection.zig`, `http3_runtime.zig`)
- [x] One process-scoped store is shared by native TCP and QUIC/H3 through the actual composition-root wiring; startup quarantine is exercised through the exact helper `run()` calls (`edge_gateway.zig`)
- [x] `zig fmt --check` on all changed files
- [x] Full suite: `zig build test` — 2647/2648 passing (1 pre-existing unrelated skip)

Part of #368 (Slice 2 of 3 — config/metrics/docs and the distributed seam proof remain in Slice 3).

### DIFF
--- a/src/edge_gateway.zig
+++ b/src/edge_gateway.zig
@@ -282,18 +282,19 @@ pub fn run(cfg: *const edge_config.EdgeConfig) !void {
     }
     defer if (native_resumption_runtime) |*rt| rt.deinit();
     // #368 Slice 2: one process-scoped anti-replay store, shared by every
-    // native TCP worker and the QUIC/H3 runtime — independent of whether
-    // native resumption is enabled above, since Slice 3 owns the switch
-    // that actually lets native 0-RTT be attempted. Declared ahead of
-    // every borrower below for the same teardown-ordering reason as
+    // native TCP worker and the QUIC/H3 runtime — but constructed only once
+    // a native PSK/early-data path can actually exist (below, once
+    // `native_tls_provider`/`h3_credential_provider` are known), so plain
+    // HTTP, OpenSSL-only, or resumption-disabled deployments never pay the
+    // default 65,536-entry reservation or gain a new startup-failure path
+    // for a security feature they cannot use. Declared (but not yet
+    // constructed) here, ahead of every borrower below, so its teardown —
+    // once present — runs last, for the same reason as
     // `native_resumption_runtime`.
-    var native_early_data_replay_store = initNativeEarlyDataReplayStore(state_allocator, @intCast(compat.milliTimestamp())) catch |err| {
-        state.logger.err(null, "native TLS/QUIC early-data replay store initialization failed ({s}); refusing to start", .{@errorName(err)});
-        return err;
-    };
-    defer native_early_data_replay_store.deinit();
-    var native_early_data_replay_gate_adapter = tls_core.early_data_replay.GateAdapter.init(native_early_data_replay_store.store());
-    const native_early_data_replay_gate = native_early_data_replay_gate_adapter.gate();
+    var native_early_data_replay_store: ?tls_core.early_data_replay.LocalStore = null;
+    defer if (native_early_data_replay_store) |*store| store.deinit();
+    var native_early_data_replay_gate_adapter: tls_core.early_data_replay.GateAdapter = undefined;
+    var native_early_data_replay_gate: ?tls_core.tls13_backend.EarlyDataReplayGate = null;
     var http3_dispatch_ctx = ghandlers.Http3DispatchContext{
         .config_store = &config_store,
         .cfg = cfg,
@@ -386,6 +387,21 @@ pub fn run(cfg: *const edge_config.EdgeConfig) !void {
     }
     defer if (native_credentials) |*store| store.deinit();
     defer if (tls_terminator) |*tls| tls.deinit();
+    if (nativeEarlyDataReplayStoreNeeded(
+        native_resumption_runtime != null,
+        native_tls_provider != null,
+        cfg.http3_enabled,
+        h3_credential_provider != null,
+    )) {
+        native_early_data_replay_store = initNativeEarlyDataReplayStore(state_allocator, @intCast(compat.milliTimestamp())) catch |err| {
+            state.logger.err(null, "native TLS/QUIC early-data replay store initialization failed ({s}); refusing to start", .{@errorName(err)});
+            return err;
+        };
+    }
+    if (native_early_data_replay_store) |*store| {
+        native_early_data_replay_gate_adapter = tls_core.early_data_replay.GateAdapter.init(store.store());
+        native_early_data_replay_gate = native_early_data_replay_gate_adapter.gate();
+    }
     if (cfg.http3_enabled) {
         if (!edge_config.hasTlsFiles(cfg)) {
             state.logger.warn(null, "HTTP/3 requested without TLS cert/key; QUIC bootstrap will remain incomplete", .{});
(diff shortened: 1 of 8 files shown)


src/tls/early_data_replay.zig
```diff
@@ -101,6 +102,47 @@ pub const Store = struct {
...
+
+    pub fn gate(self: *GateAdapter) tls13_backend.EarlyDataReplayGate {
+        return .{ .ctx = self, .decideFn = decide };
+    }
+
+    fn decide(ctx: *anyopaque, candidate: tls13_backend.EarlyDataReplayCandidate) tls13_backend.EarlyDataReplayDecision {
+        const self: *GateAdapter = @ptrCast(@alignCast(ctx));
+        return mapClaimResult(self.backing.claim(.{
+            .key = candidate.ticket_identity_fingerprint,
+            .retain_until_unix_ms = candidate.retain_until_unix_ms,
```
Jaruso, line 141: **Blocking — close the TLS-freshness / replay-store clock boundary.** `tls13_backend` validates skew using the resolver's `now`, computes `retain_until`, and only afterward reaches this adapter. `LocalStore.store()` then takes a *second* wall-clock reading in its trampoline. At the exact accepted negative-skew boundary, `retain_until == resolver_now`; if the clock advances before this `claim`, an absent key is currently inserted with a deadline that is already expired and returned as `.accepted`. A concurrent replay that also passed the earlier TLS freshness check can then see that entry as expired, remove it, and be accepted too.

That is the exact gap #368 calls out: an attempt accepted by TLS freshness must not simultaneously have fallen out of replay protection.

The local store should fail closed before inserting a new/reclaimed claim whose supplied deadline has already passed, while still letting an existing live record win as `.duplicate`:

```zig
// First preserve an already-recorded key exactly as today.
if (self.entries.getPtr(c.key)) |entry_ptr| {
    if (now_unix_ms <= entry_ptr.retain_until_unix_ms) {
        event.* = .duplicate;
        return .duplicate;
    }
    _ = self.entries.remove(c.key);
    self.noteRemovalLocked();
    expired_removed.* += 1;
}

// The freshness check may have passed on an earlier clock read, but the
// store can no longer establish a live recording window. Fail closed.
if (now_unix_ms > c.retain_until_unix_ms) {
    event.* = .unavailable;
    return .unavailable;
}
```

Put that check before capacity reclamation/insertion. Please add a deterministic regression for `deadline=T`, store `now=T+1` proving no claim is accepted/recorded, plus a TLS-level boundary test where the resolver validates at `T` and the backing local-store clock is `T+1`; PSK resumption must still complete at 1-RTT.

src/edge_gateway.zig
```diff
@@ -281,6 +281,19 @@ pub fn run(cfg: *const edge_config.EdgeConfig) !void {
         if (native_resumption_runtime) |*rt| rt.setObserver(nativeResumptionMetricsObserver(&state));
     }
     defer if (native_resumption_runtime) |*rt| rt.deinit();
+    // #368 Slice 2: one process-scoped anti-replay store, shared by every
+    // native TCP worker and the QUIC/H3 runtime — independent of whether
+    // native resumption is enabled above, since Slice 3 owns the switch
+    // that actually lets native 0-RTT be attempted. Declared ahead of
+    // every borrower below for the same teardown-ordering reason as
+    // `native_resumption_runtime`.
+    var native_early_data_replay_store = initNativeEarlyDataReplayStore(state_allocator, @intCast(compat.milliTimestamp())) catch |err| {
```
Jaruso: **Blocking availability regression — do not unconditionally allocate/fail startup on a replay store that cannot be used.** This now runs for every gateway invocation, before we even know whether a native TLS/H3 backend capable of PSK resumption exists. `LocalStore.init(.{})` reserves the full default **65,536-entry** table up front, and an OOM here aborts the whole gateway. That means plain HTTP, OpenSSL-only TCP with H3 disabled, or native TLS with resumption disabled all gain a new sizable allocation and startup-failure dependency for 0-RTT they cannot accept.

Slice 3 owns the final `disabled|process_local` config, but we should not leave an incorrect production default between slices. Keep the same single process lifetime, but construct the store only once a relevant native PSK/early-data path exists. For example, after the native/H3 credential providers are resolved:

```zig
var replay_store: ?tls_core.early_data_replay.LocalStore = null;
var replay_adapter: ?tls_core.early_data_replay.GateAdapter = null;
var replay_gate: ?tls_core.tls13_backend.EarlyDataReplayGate = null;

const native_psk_path_exists = native_resumption_runtime != null and
    (native_tls_provider != null or (cfg.http3_enabled and h3_credential_provider != null));

if (native_psk_path_exists) {
    replay_store = try initNativeEarlyDataReplayStore(
        state_allocator,
        @intCast(compat.milliTimestamp()),
    );
    replay_adapter = tls_core.early_data_replay.GateAdapter.init(replay_store.?.store());
    replay_gate = replay_adapter.?.gate();
}
defer if (replay_store) |*store| store.deinit();
```

The exact condition can follow the current composition model, and Slice 3 can replace it with the explicit replay mode. The important invariant here is: **one shared store when native early data can exist; zero replay-store allocation/failure path when it cannot.** Add a composition test proving a no-native-resumption/no-native-0RTT configuration does not initialize the store.

Jaruso (commented):
Reviewed the complete PR against the current #368 **Slice 2 — TLS/composition integration** contract, including all 8 changed files, the existing #366/#497 decision ordering, process ownership/lifetimes, TCP and H3 installation seams, the new deadline arithmetic, the multi-worker tests, and the current CI run. CI is fully green on head `3ebd9635`.

The core architecture is strong: the replay decision still happens only after PSK resolution/compatibility and binder verification; the authoritative retention deadline is derived once from the validated age observation; overflow fails closed before the gate; `GateAdapter` maps only a proven duplicate to `.replay`; duplicate/unavailable outcomes preserve ordinary PSK resumption; one adapter/store is borrowed by TCP workers and H3; and teardown ordering keeps the store alive until all borrowers are gone.

I found **two blockers** before merge:

1. **There is an exact freshness-window race between the TLS clock read and the replay-store clock read.** At the accepted negative-skew boundary, `retain_until == resolver_now`. `Store.claim()` then takes a second wall-clock reading. If that advances before the claim, `LocalStore` currently inserts an already-expired record and returns `accepted`; another concurrent replay can observe that record as expired, remove it, and also return `accepted`. That is precisely the boundary gap #368 says must not exist. The inline comment includes the safe placement for a fail-closed past-deadline check and the regression test I expect.
2. **The process replay table is allocated unconditionally for every gateway invocation.** `LocalStore.init(.{})` reserves the full default 65,536-entry capacity, and `run()` now returns an error if that allocation fails even for HTTP-only, OpenSSL-only, or otherwise non-resumable deployments where native 0-RTT cannot use the store. Keep one process-owned store when a native PSK/early-data path can actually exist, but do not make a disabled/unusable security feature a mandatory startup allocation/failure path. Slice 3 can add the final operator-facing mode switch, but current main should remain safe between slices.

Once those two are fixed, I do not see another Slice-2 correctness/security gap in the code reviewed here. Config/metrics/deployment docs and the fake distributed-store proof correctly remain Slice 3.

As with #505, GitHub does not allow the authenticated author to submit a formal `REQUEST_CHANGES` review on their own PR, so this is a COMMENT review; treat both inline findings as merge blockers.

PR #507, opened 2026-07-25
tls: add early-data replay config, metrics, and distributed-store contract proof (#368 Slice 3)

Jaruso:
## Summary

Completes #368 (final slice). Slices 1-2 (already merged) landed the process-local replay store (`early_data_replay.zig`) and wired it into TLS/composition behind a temporary "construct whenever a native path exists" gate. This slice adds the operator-facing configuration, metrics, and documentation the issue calls for, plus a proof that the TLS/composition layer depends only on the abstract `Store` contract rather than `LocalStore` internals.

- **Configuration** (`src/edge_config.zig`): `TARDIGRADE_TLS_NATIVE_EARLY_DATA_REPLAY_MODE` (`disabled` default / `process_local`) and `TARDIGRADE_TLS_NATIVE_EARLY_DATA_REPLAY_MAX_ENTRIES` (bounded by `early_data_replay.hard_max_entries`, independent of ordinary TLS resumption/session-cache config, no separate replay TTL). Unknown mode or invalid capacity fails config validation.
- **Composition** (`src/edge_gateway.zig`): replaces Slice 2's "construct whenever a native path exists" gate with `nativeEarlyDataReplayStoreEnabled`, which additionally requires the operator to have opted into `process_local`. `disabled` allocates no replay table and installs no gate (fail-closed default preserved). A configured `process_local` store that fails to initialize refuses to start rather than silently falling back unprotected.
- **Metrics** (`src/http/metrics.zig`): `tardigrade_tls_early_data_replay_total{outcome}` with the closed outcome set `accepted`/`duplicate`/`capacity_rejected`/`expired`/`unavailable`/`startup_quarantine`, bridged from the store's existing `Observer` vocabulary in `edge_gateway.zig` the same way native resumption metrics are wired (observer callbacks run outside the store mutex).
- **Distributed-store contract proof** (`src/tls/early_data_replay.zig`, `src/tls/tls13_backend_tests.zig`): a deterministic `FakeDistributedStore` test double (not a real networked backend) scripts commit/timeout/network-failure/ambiguous-commit outcomes and proves `GateAdapter`/`tls13_backend.zig` depend only on `Store.claim()`'s contract — never `LocalStore` internals — including that `duplicate`/`unavailable` map to `replay_rejected`/`replay_unavailable` while 1-RTT PSK resumption is always preserved.
- **Docs/changelog**: `docs/OBSERVABILITY.md` documents the exact process-local guarantee (at most one successful 0-RTT claim per replay key, per Tardigrade process, during the recording window), startup/restart quarantine, capacity behavior, and the cluster limitation — `process_local` is explicitly not cluster-safe or globally at-most-once; operators requiring that must leave 0-RTT disabled until a distributed backend exists. `docs/SUPPORT_MATRIX.md` and `CHANGELOG.md` updated accordingly.

No redesign of the replay store or TLS integration from Slices 1-2; no Redis/etcd/database implementation (out of scope per the issue).

## Test plan
- [x] `zig fmt --check` on all changed files
- [x] `zig build test` — full suite passes (2665+ tests)
- [x] New config tests: default disabled, `process_local` parses, unknown mode rejected, zero/above-hard-cap/hard-cap-boundary max_entries, disabled/process_local store-construction gating
- [x] New metrics tests: counter increments per outcome, Prometheus rendering with closed-enum labels only
- [x] New `FakeDistributedStore` contract tests: first claim accepted, repeated claim duplicate, timeout/network-failure/ambiguous-commit all unavailable, `GateAdapter` over the fake drives the same allow/replay/unavailable decisions as `LocalStore`
- [x] New TLS-level integration test proving a fake (non-`LocalStore`) backend behind `GateAdapter` drives `accepted`/`replay_rejected`/`replay_unavailable` while 1-RTT resumption stays intact throughout
- [x] Independent security-focused subagent review of the diff — no high-confidence findings

Closes #368


src/edge_config.zig
```diff
@@ -788,6 +820,15 @@ pub fn loadFromEnv(allocator: std.mem.Allocator) !EdgeConfig {
...
     errdefer allocator.free(tls_native_resumption_ticket_usage);
+    // #368 Slice 3: explicit replay mode/capacity — see `EarlyDataReplayMode`
+    // and `validateEarlyDataReplayConfig`. An unrecognized mode fails at load
+    // time (like `upstream_protocol`); the capacity bound is validated in
+    // `validate()` alongside the rest of numeric config so its error message
+    // can reference the configured value.
+    const tls_native_early_data_replay_mode_str = envOrDefault(allocator, "TARDIGRADE_TLS_NATIVE_EARLY_DATA_REPLAY_MODE", "disabled") catch unreachable;
+    defer allocator.free(tls_native_early_data_replay_mode_str);
+    const tls_native_early_data_replay_mode = try parseEarlyDataReplayModeConfig(tls_native_early_data_replay_mode_str);
+    const tls_native_early_data_replay_max_entries = parseIntEnv(usize, allocator, "TARDIGRADE_TLS_NATIVE_EARLY_DATA_REPLAY_MAX_ENTRIES", (tls_core.early_data_replay.Limits{}).max_entries);
```
Jaruso: **Blocking config-safety bug — malformed values silently become 65,536.** `parseIntEnv` is intentionally permissive repository-wide: it catches `std.fmt.parseInt` failure and returns the supplied default. For this new security control that means, for example, `TARDIGRADE_TLS_NATIVE_EARLY_DATA_REPLAY_MAX_ENTRIES=garbage`, `-1`, or a value that overflows `usize` never reaches `validateEarlyDataReplayConfig`; it silently becomes the default `65536` and startup succeeds. That contradicts this PR's operator contract that invalid replay capacity is rejected rather than silently coerced.

Please keep the legacy helper behavior unchanged and use a strict parser for this setting, e.g.:

```zig
const raw = try envOrDefault(
    allocator,
    "TARDIGRADE_TLS_NATIVE_EARLY_DATA_REPLAY_MAX_ENTRIES",
    "65536",
);
defer allocator.free(raw);
const tls_native_early_data_replay_max_entries =
    try parseEarlyDataReplayMaxEntriesConfig(raw);

fn parseEarlyDataReplayMaxEntriesConfig(raw: []const u8) !usize {
    const value = std.mem.trim(u8, raw, " \t\r\n");
    const parsed = std.fmt.parseInt(usize, value, 10) catch
    try validateEarlyDataReplayConfig(parsed);
    return parsed;
}
```

Keep the later `validate()` check as defense in depth for programmatically constructed configs. Add regressions for non-numeric input, negative input, integer overflow, zero, hard-cap+1, and a trimmed valid value.

src/edge_gateway.zig
```diff
@@ -387,16 +389,26 @@ pub fn run(cfg: *const edge_config.EdgeConfig) !void {
     }
     defer if (native_credentials) |*store| store.deinit();
     defer if (tls_terminator) |*tls| tls.deinit();
-    if (nativeEarlyDataReplayStoreNeeded(
+    if (nativeEarlyDataReplayStoreEnabled(
+        cfg.tls_native_early_data_replay_mode,
```
Jaruso, line 393: **Blocking — these process-owned security settings cannot currently be hot-reloaded safely.** This store/gate is constructed once here from the startup `cfg` and then borrowed for the lifetime of `run()`. But `gateway_shutdown.hotReloadConfig()` loads/validates a fresh `EdgeConfig` and installs it into `ReloadableConfigStore` without rebuilding/removing this store or changing `WorkerContext.early_data_replay_gate`.

That creates real config/runtime divergence:

```text
startup process_local -> reload disabled
active config says disabled, but existing gate/store remain installed

startup disabled -> reload process_local
active config says process_local, but no store/gate is ever created

process_local max_entries=65536 -> reload max_entries=1024
active config says 1024, live table is still 65536
```

For a security mode, publishing a configuration that is not the behavior actually being enforced is not acceptable. Rebuilding in place is also risky because it discards replay history and would require a fresh quarantine/lifetime handoff.

For this slice, please make both replay fields **restart-only**: compare the current leased config to the proposed config in `hotReloadConfig()` before installation, reject the reload if either `tls_native_early_data_replay_mode` or `tls_native_early_data_replay_max_entries` changed, keep the old config/runtime active, and report `restart required`. A small helper makes this easy to test:

```zig
fn earlyDataReplayConfigChanged(
    current: *const EdgeConfig,
    proposed: *const EdgeConfig,
) bool {
    return current.tls_native_early_data_replay_mode != proposed.tls_native_early_data_replay_mode or
        current.tls_native_early_data_replay_max_entries != proposed.tls_native_early_data_replay_max_entries;
}
```

Please test disabled<->process_local in both directions plus a capacity-only change, and document that these process-owned fields require restart.

Jaruso (commented):
Full first-pass review against the complete #368 final-slice contract. I read all 8 changed files, traced the new config through startup and hot reload, checked the process-owned store/gate lifetime, replay observer -> metrics locking, the abstract Store/GateAdapter boundary, the fake distributed-path tests, docs/support-matrix/changelog wording, and the current CI run. CI is fully green on `69178325`.

The core Slice-3 direction is good: safe default is disabled; process-local allocation is gated and bounded; init failure is fail-closed; one store/gate remains shared across native TCP + H3; metrics use a closed label set; the cluster limitation is documented clearly; and TLS tests preserve 1-RTT on duplicate/unavailable outcomes.

I found **two correctness/security blockers** plus two additional fixes I want in this pass so we do not discover them after the first round:

1. **Malformed replay capacity silently becomes the default.** The new security-sensitive max-entry setting uses the repository's permissive `parseIntEnv`, which returns the default on parse failure/overflow. `...MAX_ENTRIES=garbage`, `-1`, whitespace that does not parse, or a value larger than `usize` can therefore become `65536` and pass validation instead of producing `InvalidConfigValue`. The inline comment includes a narrow strict parser that avoids changing unrelated legacy settings.
2. **Replay mode/capacity are currently hot-reloadable in configuration but not in runtime state.** `run()` constructs the store/gate once from the startup config. `hotReloadConfig()` can then publish a new `EdgeConfig` without reconstructing or removing those process-owned objects. Starting with `process_local` and reloading to `disabled` leaves the old gate installed despite the active config saying disabled; starting disabled and reloading to `process_local` publishes process-local config with no store at all; changing max entries similarly does not change the active table. Because rebuilding the store would discard replay history, the safest first implementation is to make both fields restart-only and reject a reload that changes either one.
3. **Do not ship the deliberately non-atomic fake as public TLS API.** `tls/root.zig` publicly exports `early_data_replay`, so these new `pub` fake types become reachable production API even though their comments explicitly state that `claim()` is not atomic and does not implement the distributed retention contract. Keep the fake in test-only support (or local test code) so the public `Store` implementations cannot include one that knowingly violates the contract this issue is defining.
4. **Observability docs disagree with the implementation.** Prometheus rendering always emits all six replay series; when replay is disabled they are zero. Also, `duplicate` proves the replay key was presented again, not that an attacker necessarily did it. Please make the docs describe the actual emission behavior and avoid attributing intent.

After those are addressed, I do not currently see another #368 Slice-3 gap in the reviewed code. In particular I did not find a new store atomicity/lifetime problem, metric-cardinality leak, cluster-scope overclaim in the guarantee itself, or regression to 1-RTT fallback semantics.

As on the prior PRs, GitHub will not allow the authenticated author to submit a formal `REQUEST_CHANGES` review on their own PR, so this is a COMMENT review; treat items 1-2 as merge blockers and items 3-4 as requested cleanup in this same round.

src/tls/early_data_replay.zig
```diff
@@ -143,6 +143,69 @@ pub const GateAdapter = struct {
...
+    /// whether the claim committed.
+    timeout,
+    /// Simulates a network failure reaching the backend.
+    network_failure,
+    /// Simulates a partial/ambiguous commit the backend cannot prove either
+    /// way (e.g. replication uncertainty).
+    ambiguous_commit,
+};
+
+pub const FakeDistributedStore = struct {
```
Jaruso: **Please keep this deliberately non-conforming fake out of the public TLS API.** `tls/root.zig` exports `early_data_replay` publicly, so `pub const FakeDistributedStore`/`FakeDistributedOutcome` become production-facing symbols. The type's own contract says its `commit` path is `contains()` + `put()` rather than atomic CAS, it has no synchronization, ignores `retain_until_unix_ms`, and keeps entries forever. Those shortcuts are reasonable for a narrowly scoped test double, but they are exactly the behaviors #368 says a real `Store` backend must *not* have.

I do not want the final anti-replay API to publicly expose a `Store` implementation that knowingly violates the contract we're trying to make authoritative. Move this into test-only support (for example a module imported only by `early_data_replay` tests and `tls13_backend_tests.zig`), or define local scripted stores in the tests. If you choose to make the shared fake model the contract rather than merely script return values, put its duplicate-check/commit in one critical section and retain the supplied deadline; either approach is fine as long as the non-atomic fake is not part of `tls_core.early_data_replay`'s production public surface.

docs/OBSERVABILITY.md
```diff
@@ -94,13 +94,133 @@ logs are written through `src/http/logger.zig`.
   `tardigrade_http_early_data_upstream_425_total{action}`,
   `tardigrade_http_early_data_retry_total{result}`,
   and `tardigrade_http3_early_data_compat_total{decision}`
+- native TLS/QUIC 0-RTT anti-replay store outcomes (#368):
+  `tardigrade_tls_early_data_replay_total{outcome}` with fixed outcomes
+  `accepted`, `duplicate`, `capacity_rejected`, `expired`, `unavailable`, and
+  `startup_quarantine`. Only exported once
```
Jaruso: The implementation does not conditionally export this metric family: `Metrics.toPrometheus()` always calls `appendEarlyDataReplayPrometheus()`, and that function always emits all six series. With replay disabled/unconstructed they simply remain zero. Please change this wording to match the implementation.

Also, I would avoid `"an attacker replayed a 0-RTT flight"` below. `.duplicate` proves that the same replay key was presented again; it cannot distinguish malicious replay from a legitimate retransmission/retry. Something like `"the same replay key was observed again"` is precise and still operationally useful.

src/gateway_shutdown.zig
```diff
@@ -124,6 +124,35 @@ pub fn hotReloadConfig(
             return;
         }
     }
+    {
```
Jaruso: **Blocking regression in the restart-only fix — reject before any runtime mutation.** This new replay check is currently reached only *after* `worker_ctx.tls.updateProtocolPolicy(...)` above has already been allowed to mutate live TLS listener state. So a combined reload can partially apply and then be rejected:

```text
current: replay=disabled, protocol=P1
SIGHUP:  replay=process_local, protocol=P2

updateProtocolPolicy(P2)  -> succeeds / mutates live TLS state
earlyDataReplayConfigChanged(...) -> true
reload rejected, old EdgeConfig P1 remains installed
```

The log/error then says the whole reload was rejected and the previous configuration remains active, but runtime protocol behavior may already be P2. That is exactly the kind of config/runtime divergence this restart-only guard is meant to prevent.

Please move the replay mode/capacity comparison to the **pre-mutation phase**, immediately after `loadFromEnv` + `validate` (before `prepareOwned` is fine, and definitely before `updateProtocolPolicy`). Acquire the current config lease there, reject/destroy the newly loaded config directly if it changed, and only then begin preparing/applying runtime updates.

I would also add one regression around ordering: proposed config changes both replay mode (forcing restart) and a TLS protocol-policy field; assert the reload rejects before the TLS policy update hook is invoked, or otherwise prove the live policy is unchanged. This keeps SIGHUP transactional for this new restart-only security control.

Jaruso (commented):
Re-review of the follow-up commit against all four findings from my first pass, plus a regression scan of every line changed since `69178325`.

The original findings are addressed correctly:

- replay max entries now uses a strict parser and explicitly tests malformed, negative, overflow, zero, above-cap, hard-cap, and trimmed valid values;
- replay mode/capacity are now treated as restart-only and a rejected reload preserves the existing process-owned store/gate;
- the scripted distributed fakes are no longer public `tls_core.early_data_replay` API (the module fake is private and the TLS integration fake is local to the test);
- observability docs now correctly say all six Prometheus series are always emitted at zero when inactive, and `.duplicate` is described as repeated key presentation rather than proof of attacker intent.

Current-head CI is fully green on `5516db14`.

I found **one remaining blocker introduced by the hot-reload fix itself**: the new restart-only replay check currently runs after `worker_ctx.tls.updateProtocolPolicy(...)`. If a SIGHUP changes both replay mode/capacity *and* TLS listener protocol policy, the protocol update can succeed first, then the replay check rejects the reload. That leaves the old `EdgeConfig` installed but part of the proposed runtime policy already applied — i.e. the reload is not actually rejected atomically.

The inline comment has the narrow fix: perform the replay restart-only comparison immediately after load/validation (before any runtime mutation, ideally before `prepareOwned`/`updateProtocolPolicy`), then proceed with the rest of reload only if it is unchanged. Please add a regression that combines a replay change with a protocol-policy change and proves the protocol update callback is never invoked / runtime policy remains unchanged when the replay change forces restart.

Once that ordering issue is fixed, I do not see another merge blocker in #507 or another unmet #368 final-slice requirement in the code reviewed here. As before, GitHub cannot record a formal `REQUEST_CHANGES` from the authenticated PR author, so this is a COMMENT review; treat the inline finding as blocking.

Jaruso (commented):
Final re-review on head `5bbe33691628ddfaf12fbab3b055621a79e8e1e4`.

I compared this commit directly against the prior reviewed head `5516db14` and re-checked the complete #368 contract after the fix. The remaining hot-reload blocker is now addressed correctly: the replay mode/capacity restart-only guard runs before `updateProtocolPolicy` or any later runtime mutation, so a rejected reload cannot partially change live TLS behavior while leaving the old `EdgeConfig` active. The new end-to-end regression test is particularly good: it changes both replay mode and HTTP/2 policy, drives the real `hotReloadConfig`, and verifies both the live TLS policy and active config remain unchanged after rejection.

The additional test-discovery change is also correct and useful. `gateway_shutdown.zig` is now explicitly imported by the edge-gateway test aggregator, so its own test declarations actually participate in `zig build test`; the accompanying initialization fix for the pre-existing metrics test is consistent with that newly-exercised path.

I re-verified the prior findings as well:
- replay capacity uses strict parsing and rejects malformed/overflowing values;
- replay mode/capacity are restart-only and cannot diverge from the process-owned store/gate;
- the deliberately simplified distributed-store fakes are not public production API;
- Prometheus/docs wording now matches actual emission semantics and does not infer attacker intent from `.duplicate`;
- one process-local store/gate remains shared across eligible native TCP and H3 paths;
- duplicate/unavailable replay outcomes still reject only 0-RTT and preserve otherwise-valid 1-RTT resumption;
- the documented guarantee remains explicitly process-local, not cluster-wide.

I do not see another code or #368 acceptance-criteria blocker on this head. **Code review verdict: merge-ready pending the current-head CI run.** GitHub has CI run #945 queued at the time of this review, so I am not treating checks as green until that run completes.

As before, the authenticated account is the PR author, so GitHub does not permit a meaningful formal approval; this COMMENT review records the merge-ready code verdict.

PR #508, opened 2026-07-25
tls: add resumption/0-RTT scenario tests (#369 Slice 1)

Jaruso:
## Summary

Slice 1 of #369 (326-J), the resumption/0-RTT interop + restart + soak
validation epic. #369 is XL — external OpenSSL/QUIC-peer interop, restart/
rotation soak runs, and CI wiring — too large for one PR, so it's being
split into slices the way #367/#368 were.

This slice adds in-repo, deterministic Zig test coverage for what's
testable in-process against the resumption/0-RTT/anti-replay code already
landed by #365–#368:

- **Ticket expiry**: an end-to-end (real ClientHello/PSK-offer) test that an
  expired ticket falls back to a full handshake, not just the pre-existing
  pure-logic boundary tests.
- **Key rotation**: `ReloadableKeyRing`/`Protector` tests for both the
  grace-window shape (old key decrypt-only, still resolves) and the full-
  removal shape (old key gone, resolves as `unknown_key`, no crash).
- **Process restart / lost cache (in-process model only)**: two independent
  `Runtime` instances (stateless and stateful) modeling lost ephemeral
  state with no shared key or cache — the closest existing test only used a
  resolver hardcoded to miss. This proves the resolve-side fail-closed
  behavior in isolation; it is not the full operational restart assurance
  #369 owns (see deferred items below).
- **ALPN / cipher-suite mismatch**: end-to-end fallback-to-full-handshake
  tests alongside the pre-existing SNI-mismatch and cert-change tests, so
  every `session.ResumeMismatch` variant now has integration coverage.

Anti-replay rejection, 0-RTT-reject-falls-back-to-1-RTT, and no-double-
execution-on-425-retry were already extensively covered by #367/#368's own
suites (25+ tests across `early_data_replay.zig`, `tls13_backend_tests.zig`,
`gateway_proxy_runtime.zig`, `gateway_handlers.zig`, `edge_gateway.zig`), so
no new tests were added there — duplicating them would be noise, not signal.
That existing coverage is unit/integration-level, not the process-level
cross-layer assurance #369 owns (see deferred items below).

`docs/RESUMPTION_TEST_PLAN.md` records what this slice covers and what's
explicitly deferred to later slices:
- External OpenSSL/QUIC-peer interop.
- The operational restart/rotation matrix (actual restart/composition
  re-init, generation-overlap reloads, lifecycle boundaries, failed-reload
  retention, concurrent/cross-worker cases).
- The process-level 0-RTT/replay/425 matrix (real gateway/native path,
  duplicate-0RTT-rejected-while-1RTT-continues, capacity/quarantine
  rejecting only early data, a real 425 retry through an upstream
  execution counter) — including validating the anti-replay store's actual
  documented scope: all workers/transports in one process share one
  authoritative store, while separate processes are independent, so the
  guarantee is process-local, not cluster-wide. A real distributed backend
  (Redis/etcd/similar) was explicitly scoped out of #368's contract and is
  not required to close #369 either.
- Soak runs and CI wiring.
- A QUIC-transport-specific variant of the 0-RTT-fallback matrix.

## Test plan

- [x] `zig build test` — 2678/2679 passing (1 skipped: the known UDP smoke
      test that fails under socket-blocking sandboxes)
- [x] All new tests fail without the fix under test / pass with it (verified
      incrementally while writing each one — see the cipher-suite-mismatch
      test's PSK-length fixup in this branch's history)

Refs #369

### DIFF
--- a/docs/RESUMPTION_TEST_PLAN.md
+++ b/docs/RESUMPTION_TEST_PLAN.md
@@ -83,14 +83,16 @@ real OpenSSL/QUIC peer, no CI/soak changes.
   - Drive a real 425 early-retry through an upstream request-execution
     counter and prove it lands at exactly one execution, through the actual
     proxy path rather than a scripted attempt executor.
-  - Validate the store's documented scope directly: it is process-local
-    (each worker/process anti-replay state is independent), and it is
-    *not* cluster-wide — #368 explicitly scoped out a real distributed
-    backend (Redis/etcd/similar) as out of contract for this epic, so that
-    remains this repo's committed behavior, not a gap to close. This slice
-    should validate the process-local/multi-worker guarantee and the
-    documented non-cluster-wide limitation, not require implementing a
-    distributed backend.
+  - Validate the store's documented scope directly: all workers/transports
+    inside one process share one authoritative replay store (a replay
+    accepted by worker A is rejected by worker B in the same process);
+    separate Tardigrade processes are independent, so the guarantee is
+    process-local, not cluster-wide — #368 explicitly scoped out a real
+    distributed backend (Redis/etcd/similar) as out of contract for this
+    epic, so that remains this repo's committed behavior, not a gap to
+    close. This slice should validate the process-shared/multi-worker
+    guarantee and the documented non-cluster-wide limitation, not require
+    implementing a distributed backend.
 - **Soak scenarios**: repeated reconnect/rotation loops with memory/metrics
   checks over long runs.
 - **CI wiring**: a smoke subset plus an optional nightly/soak job.
--- a/docs/RESUMPTION_TEST_PLAN.md
+++ b/docs/RESUMPTION_TEST_PLAN.md
@@ -1,0 +1,67 @@
+# Resumption / 0-RTT Test Plan
+
+Tracks test coverage for the session-resumption and 0-RTT epic (#326) against
+the validation scope in issue #369 (326-J). #369 itself is XL and spans
+external interop, restart/rotation soak runs, and CI wiring — too large for
+one PR, so it is being landed in slices. This doc records what each slice
+covers so gaps are visible instead of implied.
+
+## Slice 1 (this PR): in-repo scenario tests
+
+Fast, deterministic, single-process Zig tests exercising the resumption/
+0-RTT/anti-replay code already landed by #365–#368. No external process, no
+real OpenSSL/QUIC peer, no CI/soak changes.
+
+- **Ticket expiry**: `src/tls/tls13_backend_tests.zig` — `"a ticket past its
+  lifetime is rejected and falls back to a full handshake (#369)"` drives a
+  real ClientHello/PSK-offer against a resolved-but-expired ticket end to
+  end. (Pure-logic expiry boundaries were already covered by
+  `src/tls/session.zig` and `src/tls/ticket_protection.zig`.)
+- **Key rotation**: `src/tls/ticket_protection.zig` —
+  `"rotation: a resolve against a decrypt-only grace-window key still
+  succeeds, and new seals use the new key (#369)"` and `"rotation: a fully
+  removed key rejects old tickets with unknown_key instead of crashing
+  (#369)"` cover both the grace-window and full-removal rotation shapes
+  against `ReloadableKeyRing`/`Protector` directly.
+- **Process restart / lost cache**: `src/tls/resumption_runtime.zig` —
+  `"restart: a stateless identity issued by one process is unresolvable by a
+  fresh process with no shared state (#369)"` and the stateful-cache
+  equivalent construct two independent `Runtime`s to model a real restart
+  (no shared ephemeral key or cache), not just a resolver returning miss.
+- **ALPN / cipher-suite mismatch on resumption**: `src/tls/
+  tls13_backend_tests.zig` — `"an ALPN mismatch falls back..."` and `"a
+  cipher-suite mismatch falls back..."`, alongside the pre-existing SNI-
+  mismatch and certificate-change tests, so every `session.ResumeMismatch`
+  variant now has an end-to-end (not just unit-level) regression.
+- **0-RTT replay rejection, anti-replay capacity, fallback to 1-RTT/full
+  handshake, and no-double-execution on 425 retry**: already extensively
+  covered by #368/#367's own test suites
+  (`src/tls/early_data_replay.zig`, `src/tls/tls13_backend_tests.zig`,
+  `src/gateway_proxy_runtime.zig`, `src/gateway_handlers.zig`,
+  `src/edge_gateway.zig`). No new tests added here in this slice — see the
+  file-level docs on those tests for the existing matrix.
+
+## Explicitly deferred to a later slice
+
+- **External interop** with OpenSSL (`s_client`/`s_server` ticket
+  round-trips) and an independent QUIC peer for H3 resumption/0-RTT. Needs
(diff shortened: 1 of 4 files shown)


docs/RESUMPTION_TEST_PLAN.md
```diff
@@ -0,0 +1,67 @@
...
+  real ClientHello/PSK-offer against a resolved-but-expired ticket end to
+  end. (Pure-logic expiry boundaries were already covered by
+  `src/tls/session.zig` and `src/tls/ticket_protection.zig`.)
+- **Key rotation**: `src/tls/ticket_protection.zig` —
+  `"rotation: a resolve against a decrypt-only grace-window key still
+  succeeds, and new seals use the new key (#369)"` and `"rotation: a fully
+  removed key rejects old tickets with unknown_key instead of crashing
+  (#369)"` cover both the grace-window and full-removal rotation shapes
+  against `ReloadableKeyRing`/`Protector` directly.
+- **Process restart / lost cache**: `src/tls/resumption_runtime.zig` —
```
Jaruso: **[P1] Keep the real restart/rotation matrix open for a later slice.** These two independent `Runtime`s are useful deterministic *models* of lost ephemeral state, but this currently reads as though #369's process-restart obligation is covered. The existing #369 restart/rotation matrix asks for operational behavior that this does not exercise: old ticket -> actual restart/composition re-init -> resolver miss -> usable full handshake -> fresh ticket issuance + bounded metrics; persistent-overlap restart with generation N retained decrypt-only and N+1 current (or a new non-overlapping nonce lease); exact lifecycle boundaries; failed reload retaining the old snapshot; and the concurrent/cross-worker cases. Please label this bullet as in-process/unit coverage and add those operational cases to the deferred list (or explicitly mark any blocked by a missing production seam), otherwise the tracker can allow #369 to close without the restart assurance it owns.

docs/RESUMPTION_TEST_PLAN.md
```diff
@@ -0,0 +1,67 @@
...
+  `"restart: a stateless identity issued by one process is unresolvable by a
+  fresh process with no shared state (#369)"` and the stateful-cache
+  equivalent construct two independent `Runtime`s to model a real restart
+  (no shared ephemeral key or cache), not just a resolver returning miss.
+- **ALPN / cipher-suite mismatch on resumption**: `src/tls/
+  tls13_backend_tests.zig` — `"an ALPN mismatch falls back..."` and `"a
+  cipher-suite mismatch falls back..."`, alongside the pre-existing SNI-
+  mismatch and certificate-change tests, so every `session.ResumeMismatch`
+  variant now has an end-to-end (not just unit-level) regression.
+- **0-RTT replay rejection, anti-replay capacity, fallback to 1-RTT/full
```
Jaruso, line 40: **[P1] Do not treat #367/#368's in-process tests as the end of #369's replay/425 obligation.** Reusing those tests as the baseline is correct, but #326 assigns #369 the *external/soak assurance* for replay/restart behavior, and #369 explicitly calls for accepted/rejected 0-RTT, replay attempts, anti-replay capacity, 425 retry, and fallback. Add a deferred process-level matrix here, e.g. run the real gateway/native TLS or H3 path with `process_local`, prove duplicate 0-RTT is rejected while the PSK connection continues as 1-RTT, prove capacity/startup-quarantine failures reject only early execution, and drive a real 425 retry through an upstream side-effect counter that ends at exactly one. That avoids duplicating the unit tests while still validating the cross-layer behavior this story exists to own.

docs/RESUMPTION_TEST_PLAN.md
```diff
@@ -0,0 +1,67 @@
...
+  `src/gateway_proxy_runtime.zig`, `src/gateway_handlers.zig`,
+  `src/edge_gateway.zig`). No new tests added here in this slice — see the
+  file-level docs on those tests for the existing matrix.
+
+## Explicitly deferred to a later slice
+
+- **External interop** with OpenSSL (`s_client`/`s_server` ticket
+  round-trips) and an independent QUIC peer for H3 resumption/0-RTT. Needs
+  subprocess-driven test harnesses, not covered here.
+- **A real distributed anti-replay backend.** `src/tls/
```
Jaruso: **[P1] This is outside #369's required implementation scope.** #368 explicitly says `Do not implement Redis, etcd, or another remote service in this story`, and the merged PR #507 likewise closed #368 with only the pluggable `Store` contract plus the documented process-local/cluster limitation. #369 should validate that current guarantee (including multi-worker/process-local behavior and the fact that it is *not* cluster-wide), not make a real distributed backend a prerequisite for closing the test epic. Please remove this as a deferred completion item; replace it with process-local/cluster-policy validation if needed.

Jaruso (commented):
Slice 1's added Zig tests look good as deterministic in-process coverage, and CI is green. I found three scope-tracking issues in the new test-plan document that should be corrected before merge because this file is now the roadmap for the remaining #369 slices.

The main concern is that the plan currently mixes three categories: coverage already proven in predecessor unit/integration suites, process-level/external assurance that still belongs to #369, and one distributed-backend implementation that is explicitly outside the current epic contract. The inline comments spell out the changes I recommend.

docs/RESUMPTION_TEST_PLAN.md
```diff
@@ -0,0 +1,105 @@
...
+    a duplicate 0-RTT claim is rejected while the underlying PSK connection
+    still completes as ordinary 1-RTT.
+  - Prove anti-replay capacity exhaustion and startup-quarantine both
+    reject only the early-data attempt, never the resumed/full-handshake
+    connection.
+  - Drive a real 425 early-retry through an upstream request-execution
+    counter and prove it lands at exactly one execution, through the actual
+    proxy path rather than a scripted attempt executor.
+  - Validate the store's documented scope directly: it is process-local
+    (each worker/process anti-replay state is independent), and it is
```
Jaruso: **[P1] Workers inside one process must share the replay scope.** This parenthetical currently says `each worker/process anti-replay state is independent`, which contradicts #368's core guarantee: there is one process-shared `LocalStore` across native TCP workers and H3, specifically so a replay accepted by worker A is rejected by worker B in the same process. Only *separate Tardigrade processes* have independent replay state when using `process_local`. Please rewrite this along the lines of: `all workers/transports in one process share one authoritative replay store; separate processes are independent, so the guarantee is not cluster-wide.` The later phrase `process-local/multi-worker guarantee` points in the right direction, but this sentence currently states the opposite security property.

Jaruso (commented):
Follow-up review: the three prior scope findings are substantially addressed. I found one new P1 wording error in the corrected process-local replay description, plus the PR body is still stale and continues to list a real distributed anti-replay backend as deferred #369 work. Please fix both so the tracker and PR summary describe the same ownership contract. CI for the new head is currently still running.

Jaruso (commented):
Re-review complete on head `64a074e8`. The remaining findings are addressed:

- `docs/RESUMPTION_TEST_PLAN.md` now correctly states that all workers/transports inside one Tardigrade process share one authoritative process-local replay store, while separate processes have independent replay state.
- The PR description now matches that contract and explicitly states that a real distributed anti-replay backend is out of scope for #369.
- The earlier restart/rotation and process-level replay/425 obligations remain explicitly open for later #369 slices rather than being implied complete by this Slice 1.

The latest commit is documentation-only, so the previously reviewed Zig test implementation is unchanged. I found no new correctness or scope issues in this pass.

CI for the latest head is still running at review time.

PR #509, opened 2026-07-25
tls: add process-level 0-RTT replay and 425 assurance (#369 Slice 2)

Jaruso:
## Summary

Slice 2 of #369 (326-J). Slice 1 (#508) added deterministic in-process resumption scenario tests plus `docs/RESUMPTION_TEST_PLAN.md`. This slice moves assurance outward across the production TLS/replay-store/H1 early-data/retry surfaces that are currently wired, while explicitly leaving the remaining native TCP/H1 production 0-RTT enablement and provenance path tracked in #510.

Central invariant under test: a replay or anti-replay-store failure may reject early execution, but must never destroy an otherwise-valid PSK-resumed connection, and retry behavior must never execute an application request twice.

**#369 stays open** — this slice does not close it. Several scenarios remain explicitly deferred in `docs/RESUMPTION_TEST_PLAN.md`, including true end-to-end native TCP/H1 0-RTT dispatch from a production-issued early-capable ticket through production server policy, record provenance, and the H1 safety gate (#510).

### TLS / replay-store layer

`src/tls/tls13_backend_tests.zig` reuses the existing production TLS harness and a real `tls_core.early_data_replay.LocalStore`/`GateAdapter`. The tests now assert real observer/event metric deltas for accepted, duplicate, capacity-rejected, and startup-quarantine outcomes instead of deriving any application-execution counter from `earlyDataAccepted()` itself.

Covered scenarios:

- `rt0.accept.first_use`
- `rt0.reject.duplicate`
- `rt0.reject.capacity`
- `rt0.reject.startup_quarantine`
- `rt0.reject.cross_worker_duplicate`

### H1 dispatch / gateway composition

`src/edge_gateway.zig` adds tests next to the private production H1 orchestration and native replay composition helpers:

- `rt0.reject.unsafe_request` drives the real `executeH1PostPreflightOrchestration` with counting hooks and proves an unsafe early POST returns 425 before route/upstream dispatch.
- `rt0.reject.store_unavailable` drives the same `nativeEarlyDataReplayComposition` helper used by `run()` with deterministic inputs, proves disabled mode withholds store/gate options even when a native path exists, and builds a real `NativeTlsConnection` with the returned null replay gate so `Tls13Backend` retains its fail-closed default.

### Process-level retry

`src/process_early_data_integration_tests.zig` drives the real `gateway_proxy_runtime.runBufferedProxyAttempts` retry/425 state machine and the real production TCP HTTP client against a loopback upstream server. The responder uses bounded accept/read loops, reads headers through `\r\n\r\n`, joins before shared-state inspection, and verifies the first request carries `Early-Data: 1`, the retry drops it, the upstream executes exactly once, and real metrics distinguish the upstream 425 from the retry.

### Docs and follow-up

`docs/RESUMPTION_TEST_PLAN.md` records the slice coverage, documents the worker-thread-pinning test seam gap, and explicitly defers external interop, operational restart/rotation, soak/CI, QUIC-specific rejection fallback, and the true native TCP/H1 production 0-RTT path. #510 now owns the remaining native TCP production prerequisites: early-capable ticket advertisement, server early-data policy composition, record-read provenance propagation, and H1 safety-gate dispatch.

## Test plan

- [x] `zig fmt --check .`
- [x] `zig build test`

Refs #369
Refs #510

### DIFF
--- a/docs/RESUMPTION_TEST_PLAN.md
+++ b/docs/RESUMPTION_TEST_PLAN.md
@@ -48,6 +48,94 @@ real OpenSSL/QUIC peer, no CI/soak changes.
   cross-layer, process-level version of this assurance to #369, which those
   predecessor suites do not exercise — see the deferred matrix below.
 
+## Slice 2 (this PR): process-level 0-RTT replay / anti-replay / 425 assurance
+
+Moves assurance one layer outward from Slice 1: exercises the real
+production TLS/replay-store/HTTP-early-data/gateway-retry code composed
+together, rather than each layer's own unit tests in isolation. Two new
+test areas, both reusing existing production code and harness patterns
+rather than scripted stand-ins:
+
+- **`src/tls/tls13_backend_tests.zig`** (real `DirectHarness`
+  ClientHello/PSK-offer/binder handshakes, a real
+  `tls_core.early_data_replay.LocalStore`/`GateAdapter`, and an atomic
+  "application executed" counter driven strictly by the real
+  `earlyDataAccepted()` outcome — not a re-derivation of the TLS decision
+  itself):
+  - `rt0.accept.first_use` — accepted 0-RTT records the replay claim,
+    executes the application exactly once, and a diagnostic-formatting
+    check confirms the raw ticket identity never appears in what a log
+    line would show (the candidate type has no PSK field at all, so that
+    leak is impossible by construction, not merely untested).
+  - `rt0.reject.duplicate` — an exact-duplicate claim on a second,
+    independent connection never executes the application; the resumed
+    connection remains usable and a later, distinct 1-RTT request executes
+    normally (total: one execution for the accept, one more for the
+    intentionally distinct later request).
+  - `rt0.reject.capacity` — a store configured with one live slot rejects a
+    second, otherwise-valid claim with the typed capacity outcome, no
+    application side effect, occupancy stays bounded, and a later request
+    on the same connection still succeeds.
+  - `rt0.reject.startup_quarantine` — a fresh store (modeling lost replay
+    history after a restart) rejects early execution during quarantine
+    without any application side effect, and the connection remains usable
+    for an ordinary request; the exact quarantine boundary is exercised
+    against #368's existing (exclusive-end) semantics
+    (`now < quarantine_end` rejects, `now == quarantine_end` is ordinary
+    eligibility) with a deterministic injected clock, not a sleep.
+  - `rt0.reject.cross_worker_duplicate` — two independent per-connection
+    `DirectHarness` instances ("worker A"/"worker B") share one real
+    `LocalStore`; worker A's accepted claim executes once, and worker B's
+    replay of the same identity is rejected and never executes. See the
+    "true OS-thread/worker-routing seam" note below.
+- **`src/process_early_data_integration_tests.zig`** (new file, wired into
+  `zig build test` via `edge_gateway.zig`'s existing `test { _ = @import(...) }`
+  aggregator pattern):
+  - `rt0.retry.425_exactly_once` — drives the real
+    `gateway_proxy_runtime.runBufferedProxyAttempts` retry/425 state
+    machine and the real production TCP HTTP client
+    (`executeBufferedDataPlaneProxyRequest`) against an actual loopback
+    upstream test server with an atomic execution counter that itself
+    returns a real `425 Too Early` whenever it sees a real `Early-Data: 1`
+    header. Proves: the retried request no longer carries the header, the
+    upstream executes exactly once total, the response returned to the
+    caller is the successful retry, and real
+    `http.metrics.Metrics.recordHttpEarlyDataUpstream425`/
+    `recordHttpEarlyDataRetry` deltas distinguish the initial 425 from the
+    successful retry — no parallel test-only metrics model.
+  - `rt0.reject.unsafe_request` — an unsafe method (POST) with current-hop
+    early data is rejected by the real, `pub`
+    `gateway_handlers.earlyDataDecisionForRequest` (the same function
+    `edge_gateway.zig`'s H1 dispatch calls) before any upstream dispatch is
+    attempted at all — the configured `proxy_pass` target is deliberately
+    nothing that listens, so the only way the test can pass is by the real
+    gateway decision short-circuiting before any network I/O, not by
+    racing a responder thread.
+  - `rt0.reject.store_unavailable` — ties two independently-proven facts
+    together: `edge_config.loadFromEnv`'s default configuration leaves
+    native 0-RTT replay mode `disabled` and defines no location routes at
+    all, and `tls13_backend.EarlyDataReplayGate`'s default (no gate
+    configured) fails closed to `.unavailable` for 0-RTT while leaving
+    ordinary 1-RTT resumption untouched (both already proven individually
+    by `edge_config.zig`'s and `tls13_backend_tests.zig`'s own suites) — so
+    a freshly started process with no operator configuration cannot accept
+    0-RTT anywhere, without needing any replay backend present.
+
+### Known gap: no deterministic worker-thread-routing test seam
+
+The cross-worker test above proves the process-shared-store guarantee
+using two independent per-connection state instances (the same object
+`edge_gateway.zig`'s real composition actually shares by reference across
+every native TCP worker and QUIC/H3 — see `initNativeEarlyDataReplayStore`/
+`GateAdapter.init` there), not two real OS worker threads. There is
+currently no API to pin a connection to a specific worker thread
+deterministically for testing, so a stronger proof (issue the accepted
+claim through worker thread A, replay it through worker thread B) is not
+available today. A future slice could add a small, explicit worker-id test
+seam to `WorkerContext` to close this gap; until then this is a
+documented, intentional limitation of this slice's assurance, not a
+weakening of the "workers share one store" guarantee itself.
+
 ## Explicitly deferred to a later slice
 
 - **External interop** with OpenSSL (`s_client`/`s_server` ticket
@@ -70,29 +158,11 @@ real OpenSSL/QUIC peer, no CI/soak changes.
   - A failed reload retains the prior snapshot rather than leaving the
     runtime with no usable key.
   - Concurrent / cross-worker restart and rotation cases.
-- **Process-level 0-RTT/replay/425 matrix** (the cross-layer assurance #326
-  assigns to #369, distinct from the unit/integration tests #367/#368
-  already carry):
-  - Drive the real gateway/native TLS or H3 path end to end with a
-    `process_local` (in-process, not distributed) anti-replay store; prove
-    a duplicate 0-RTT claim is rejected while the underlying PSK connection
-    still completes as ordinary 1-RTT.
-  - Prove anti-replay capacity exhaustion and startup-quarantine both
-    reject only the early-data attempt, never the resumed/full-handshake
-    connection.
-  - Drive a real 425 early-retry through an upstream request-execution
-    counter and prove it lands at exactly one execution, through the actual
-    proxy path rather than a scripted attempt executor.
-  - Validate the store's documented scope directly: all workers/transports
-    inside one process share one authoritative replay store (a replay
-    accepted by worker A is rejected by worker B in the same process);
-    separate Tardigrade processes are independent, so the guarantee is
-    process-local, not cluster-wide — #368 explicitly scoped out a real
-    distributed backend (Redis/etcd/similar) as out of contract for this
-    epic, so that remains this repo's committed behavior, not a gap to
-    close. This slice should validate the process-shared/multi-worker
-    guarantee and the documented non-cluster-wide limitation, not require
-    implementing a distributed backend.
+- **Deterministic worker-thread-pinning test seam** — see "Known gap"
+  above. #369 Slice 2 proves process-shared-store sharing across
+  independent per-connection state; it does not drive that proof through
+  real OS worker threads, because no seam exists yet to pin a connection to
+  one deterministically.
 - **Soak scenarios**: repeated reconnect/rotation loops with memory/metrics
   checks over long runs.
 - **CI wiring**: a smoke subset plus an optional nightly/soak job.
@@ -104,4 +174,5 @@ real OpenSSL/QUIC peer, no CI/soak changes.
 Acceptance criteria for #369 as a whole (interop reliability, safe
 fallback/rejection under mismatch, no double execution, bounded soak
 memory/cache growth, reproducible artifacts without leaking ticket keys)
-remain open until the deferred items above land in follow-up slices.
+remain open until the deferred items above land in follow-up slices. #369
+is not complete after this slice.
--- a/docs/RESUMPTION_TEST_PLAN.md
+++ b/docs/RESUMPTION_TEST_PLAN.md
@@ -52,45 +52,81 @@ real OpenSSL/QUIC peer, no CI/soak changes.
 
 Moves assurance one layer outward from Slice 1: exercises the real
 production TLS/replay-store/HTTP-early-data/gateway-retry code composed
-together, rather than each layer's own unit tests in isolation. Two new
-test areas, both reusing existing production code and harness patterns
-rather than scripted stand-ins:
+together, rather than each layer's own unit tests in isolation, and rather
+than scripted stand-ins where production code was reachable.
 
-- **`src/tls/tls13_backend_tests.zig`** (real `DirectHarness`
-  ClientHello/PSK-offer/binder handshakes, a real
-  `tls_core.early_data_replay.LocalStore`/`GateAdapter`, and an atomic
-  "application executed" counter driven strictly by the real
-  `earlyDataAccepted()` outcome — not a re-derivation of the TLS decision
-  itself):
-  - `rt0.accept.first_use` — accepted 0-RTT records the replay claim,
-    executes the application exactly once, and a diagnostic-formatting
-    check confirms the raw ticket identity never appears in what a log
-    line would show (the candidate type has no PSK field at all, so that
-    leak is impossible by construction, not merely untested).
+A review pass on the first draft of this slice found that its TLS-layer
+tests paired each real TLS/replay decision with a same-process "application
+executed" counter incremented directly from that same decision
+(`server_backend.earlyDataAccepted()`) — tautological, since it could never
+catch the actual cross-layer bug #369 cares about (HTTP/gateway dispatch
+ignoring a correct TLS decision), and, worse, unfalsifiable for the native
+TCP/H1 transport today: see "Confirmed production gap" below. The review
+also found a real test-only deadlock risk (an unbounded `accept()` a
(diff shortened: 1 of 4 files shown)


src/tls/tls13_backend_tests.zig
```diff
@@ -8124,3 +8124,367 @@ test "record mode delivers a fatal alert to the client when required client auth
...
+        .resolveFn = IdentityResolver.resolve,
+    });
+
+    try harness.client_backend.setClientEarlyDataIntent(.{ .enabled = true, .max_bytes = 16384 });
+    try harness.server_backend.setServerEarlyDataPolicy(.{ .enabled = true, .age_skew_tolerance_ms = 60_000 });
+    try harness.server_backend.setEarlyDataReplayGate(gate);
+
+    try harness.run();
+
+    if (harness.server_backend.earlyDataAccepted()) application_executions.* += 1;
```
Jaruso: **[P1] This counter makes the application-side-effect assertion tautological.** `application_executions` increments *because* `earlyDataAccepted()` returned true, so these tests cannot detect the cross-layer bug #369 is meant to catch: TLS correctly returns `replay_rejected`/`replay_unavailable`, but the HTTP/gateway layer nevertheless executes the early request. The existing #367/#368 tests already prove this TLS decision; adding a counter derived from the same decision does not move assurance outward.

Please keep the useful TLS assertions here (duplicate/capacity/quarantine reject only early data and the PSK connection survives), but prove execution through a production dispatch seam. For example, feed the resulting `EarlyDataContext` into the real H1 preflight/orchestration path and increment the counter only inside the route/application callback:

```zig
var probe = DispatchProbe{};
const hooks = H1PostPreflightHooks{
    .ctx = &probe,
    .route = DispatchProbe.route,          // increments executions
    .rejectEarly = DispatchProbe.reject,  // increments rejects
};
try executeH1PostPreflightOrchestration(..., hooks);
try expectEqual(expected_executions, probe.executions);
```

The exact helper names can follow the existing edge-gateway test seams, but the key requirement is: **do not derive `executions` from the decision being tested.** If a real dispatch seam cannot be reached in this slice, relabel these as TLS-level composition tests and leave the application-side-effect portion explicitly deferred.

src/process_early_data_integration_tests.zig
```diff
@@ -0,0 +1,422 @@
...
+    // itself is configured replay-safe and RFC-8470-aware, an unsafe
+    // method must never be admitted as early data. This is the same real,
+    // `pub` production decision function `edge_gateway.zig`'s H1 dispatch
+    // calls (via `earlyDataPreflightDecisionForH1`) before ever reaching
+    // `hooks.route`/`gproxy_runtime.handleLocationProxyPass` — see
+    // `executeH1PostPreflightOrchestration` in `edge_gateway.zig`, whose
+    // `.too_early` branch calls `hooks.rejectEarly` and returns without
+    // calling `hooks.route` at all.
+    const early_ctx = http.request_context.EarlyDataContext{ .transport_early = true };
+    const decision = ghandlers.earlyDataDecisionForRequest(&cfg, early_ctx, .POST, "/work", false);
```
Jaruso: **[P1] This does not actually test the gateway short-circuit.** The test calls `earlyDataDecisionForRequest()` and then stops. The `proxy_pass = "http://127.0.0.1:1"` value is never consumed by any code path, so an implementation bug where `executeH1PostPreflightOrchestration` ignores `.too_early` and still calls `hooks.route` would pass this test unchanged. This is effectively another unit test of the decision function, which #367 already covers.

Please drive the real H1 preflight/post-preflight orchestration seam referenced in the comment above. A minimal probe is enough:

```zig
var probe = struct { routed: usize = 0, rejected: usize = 0 }{};
// route hook: probe.routed += 1
// reject hook: probe.rejected += 1

try executeH1PostPreflightOrchestration(...);
try expectEqual(@as(usize, 0), probe.routed);
try expectEqual(@as(usize, 1), probe.rejected);
```

Using a real loopback target is also fine, but merely configuring an unreachable target without invoking the dispatch path provides no evidence of non-contact. If that orchestration function is not currently test-accessible, document the missing seam and keep `rt0.reject.unsafe_request` deferred instead of marking it covered.

src/process_early_data_integration_tests.zig
```diff
@@ -0,0 +1,422 @@
...
+
+    var cfg = try edge_config.loadFromEnv(allocator);
+    defer cfg.deinit(allocator);
+
+    const listener = try bindEphemeralLoopbackListener();
+    defer _ = std.c.close(listener.fd);
+
+    var origin = OriginState{};
+    const responder = try std.Thread.spawn(.{}, earlyRejectingOriginResponder, .{ listener.fd, &origin, @as(usize, 2) });
+    defer responder.join();
```
Jaruso: **[P1] The new 425 integration test can hang forever on the exact failures it is supposed to expose, and it also reads shared state before thread synchronization.** Defers unwind LIFO, so `responder.join()` runs **before** the listener close. If the implementation regresses and never opens the second retry connection, the responder remains blocked in `accept()` and the test hangs rather than failing. In addition, `early_header_seen` is a plain `[4]bool` written by the responder thread and read by the main thread before `join()`, which is an unsynchronized cross-thread access.

Please make responder shutdown bounded/failure-safe and establish a real happens-before relationship before reading its observations. One robust shape is a poll/accept loop with a bounded deadline plus an atomic stop flag, followed by `join()` before inspecting non-atomic state; alternatively make the observations atomic and still ensure the thread can be terminated on every error path. Also, the responder currently performs one `read()` and assumes the complete HTTP header arrived in that call; read until `\r\n\r\n` (with a bound) so normal TCP fragmentation cannot make the first request look non-early.

The important regression behavior should be `test fails with useful assertions`, never `CI blocks in Thread.join`.

src/process_early_data_integration_tests.zig
```diff
@@ -0,0 +1,422 @@
...
+    //     never touching ordinary 1-RTT resumption (already directly
+    //     proven end-to-end in `tls13_backend_tests.zig`'s "0-RTT
+    //     anti-replay defaults to unavailable (fails closed) when no gate
+    //     is configured" test).
+    //
+    // Together: a freshly started process with no operator configuration
+    // cannot accept 0-RTT anywhere, by construction, without needing a
+    // distributed or even a local replay backend to be present.
+    const allocator = testing.allocator;
+    var cfg = try edge_config.loadFromEnv(allocator);
```
Jaruso: **[P1] This is not a composed test of the production disabled/unavailable boundary, and it is not hermetic.** It separately proves (1) the ambient environment produced a config whose mode is `disabled`, then (2) a newly constructed `EarlyDataReplayGate{}` has no callback. Nothing here proves that the edge-gateway composition actually installs/leaves that fail-closed gate when the config is disabled. A wiring regression that installs a permissive gate in disabled mode would still pass.

Please construct an explicit deterministic config (rather than depending on whatever `TARDIGRADE_*` variables happen to exist in the test process), run the same composition helper used by `edge_gateway` to decide/build/install the replay store/gate, and drive an early-data attempt far enough to observe `.unavailable` while ordinary resumption remains usable. Existing helpers/tests around `nativeEarlyDataReplayStoreEnabled`, `initNativeEarlyDataReplayStore`, and the native TCP/H3 gate installation should provide the right starting point.

If current visibility makes that impossible without a production seam, keep the two baseline assertions but do not count them as closing the process-level `store_unavailable` scenario.

src/tls/tls13_backend_tests.zig
```diff
@@ -8124,3 +8124,367 @@ test "record mode delivers a fatal alert to the client when required client auth
...
+    var buf: [512]u8 = undefined;
+    const diagnostic = try std.fmt.bufPrint(&buf, "candidate={any}", .{candidate});
+    try std.testing.expect(std.mem.indexOf(u8, diagnostic, raw_ticket) == null);
+}
+
+test "#369 Slice 2 rt0.accept.first_use: accepted 0-RTT records the replay claim, executes the application exactly once, and leaks no raw ticket identity in diagnostics" {
+    var issued = try issueEarlyCapableTicket(32);
+    defer issued.deinit();
+
+    var store = try early_data_replay.LocalStore.init(std.testing.allocator, .{}, 0, 0);
```
Jaruso, line 8300: **[P2] The Slice 2 prompt called for typed metric deltas for accepted/duplicate/capacity/quarantine outcomes, but these new `LocalStore` instances have no observer installed.** The #368 unit tests prove the enum-to-metric mapping, but this slice is supposed to catch composition/observability regressions too.

Please wire the same non-secret `LocalStore.Observer`/metrics callback used by the production edge-gateway composition, snapshot the relevant counters, and assert a `+1` delta for each scenario (accepted, duplicate, capacity rejection, startup quarantine/unavailable as applicable). Do not add ticket/replay identity labels. If the production metric object cannot be reached from this harness, leave that metric assurance explicitly deferred rather than silently omitting it.

Jaruso (commented):
The 425 retry half is strong: it drives the real retry state machine, real TCP client, real loopback origin, verifies the actual `Early-Data: 1` bytes disappear on retry, and proves one upstream execution. CI is also green across the supported matrix.

I found four blocking assurance/test-harness issues plus one metrics gap. The biggest theme is that several tests are described as process-level/cross-layer but still infer application execution directly from the TLS decision, so they would not catch a composition bug where HTTP/gateway dispatch ignored that decision. Please address the inline comments before merge. I did not find a production-code regression in this PR; the requested changes are about making Slice 2 actually prove the security invariants it claims and keeping the new integration test failure-safe.

docs/RESUMPTION_TEST_PLAN.md
```diff
@@ -48,6 +48,149 @@ real OpenSSL/QUIC peer, no CI/soak changes.
...
+
+That is: for the native record/TCP transport, **no production code path
+today reads the real TLS backend's `earlyDataAccepted()`/`earlyDataDecision()`
+and forwards it into `ctx.early_data.transport_early`** for the H1 request
+that follows. `transport_early` only ever becomes `true` for H1 in
+production via this hardcoded-`false` assignment (i.e. never); the only
+other `.transport_early = true` assignments in `edge_gateway.zig` are
+either H2 frame-provenance propagation (already set from elsewhere, not
+derived from the TLS backend) or test-only fixtures. Wiring this is a
+`#366`/`#367` follow-up, not this slice's scope — but it means no test,
```
Jaruso: **[P1] Do not orphan this production gap or remove it from #369's remaining matrix.** This is a real missing link in the feature, not merely a test seam: `serveOneRequest` still hardcodes H1 `ctx.early_data.transport_early = false`, even though `NativeTlsConnection.httpConnection()` already exposes record-read provenance. That means a direct native TLS/H1 request whose bytes were accepted as 0-RTT cannot currently become replay-exposed at the HTTP safety gate. #366's own contract requires accepted bytes to retain sticky provenance before application side effects, and #367 explicitly assumes that handoff; both issues are already closed, and I could not find an open issue that now owns this missing production wiring.

The document then removes the old process-level 0-RTT/replay matrix from `Explicitly deferred`, so #369 could eventually close after external/soak work without ever proving the actual record-TLS -> H1 provenance -> safety-gate path.

Please do one of the following before merge:

1. reopen #366 or create a dedicated linked follow-up issue for the native record/H1 provenance handoff, **and** add an explicit deferred #369 item for true end-to-end native TCP/H1 0-RTT acceptance/rejection assurance; or
2. implement the missing handoff in this slice if you intentionally want to absorb that production work here.

Either way, do not mark the process-level H1 path complete until a real accepted/rejected TLS 0-RTT record can flow through production provenance into HTTP dispatch. The current scoped TLS tests + constructed `EarlyDataContext` tests are useful, but they deliberately stop on opposite sides of this missing link.

Jaruso (commented):
Second pass on the updated head. The previous tautological execution-counter finding is addressed, the unsafe POST now drives the real H1 orchestration, the loopback 425 responder is bounded/synchronized and fragmentation-safe, and the replay-store tests now assert the real Observer/Event seam. CI is green.

I found two remaining blockers. The more important one was surfaced by the fixes themselves: the tree currently has a real production native TCP/H1 provenance gap, but the tracker does not leave it owned anywhere. The second is that `rt0.reject.store_unavailable` still claims stronger composition/hermetic coverage than it provides.

Also please update the PR description before merge: it still describes the superseded first-draft tests (the `earlyDataAccepted()`-derived execution counter, decision-only unsafe POST test, and old default-config test), so it no longer matches the actual diff or the newly discovered H1 production gap.

src/edge_gateway.zig
```diff
@@ -3505,6 +3505,124 @@ test "H1 early proxy with origin capability off never reaches upstream side effe
...
+    // The route hook — which is where any upstream dispatch, real or
+    // otherwise, would originate — never ran.
+    try std.testing.expectEqual(@as(usize, 0), effects.upstream_calls);
+    try std.testing.expectEqual(@as(usize, 0), effects.handler_calls);
+}
+
+test "#369 Slice 2 rt0.reject.store_unavailable: the disabled-mode composition decision installs no replay gate on the real native TLS connection, matching Tls13Backend's fail-closed default" {
+    // Deterministic, explicit config rather than depending on ambient
+    // TARDIGRADE_* env vars — forces the exact decision `run()` makes.
+    var cfg = try edge_config.loadFromEnv(std.testing.allocator);
```
Jaruso: **[P1] This still does not satisfy the test's own “deterministic production composition” claim.** First, `loadFromEnv()` is still ambient: overriding only `tls_native_early_data_replay_mode` afterward does not erase other `TARDIGRADE_*` inputs, so e.g. an ambient location config can make the final `location_blocks.len == 0` assertion fail and an invalid env value can fail before the override. The comment/docs saying this is “not ambient” are therefore false.

More importantly, the test manually reproduces `run()`'s expected branch: it calls `nativeEarlyDataReplayStoreEnabled(...)`, then independently constructs `NativeTlsConnection` with `.{} `. If `run()` regressed tomorrow and installed a permissive gate in `.disabled` mode, both halves of this test would still pass because neither half executes the composition branch in `run()`.

Please factor the small replay-store/gate option-selection step used by `run()` into a production helper and have this test exercise that same helper/object construction, or relabel this as helper/default-backend coverage and leave the **process-level disabled/unavailable composition** scenario explicitly deferred. At minimum, remove `loadFromEnv()` from this test and construct only the deterministic inputs it actually needs.

Jaruso (commented):
Follow-up review on `1e53884d`: the two remaining P1s from my prior pass are addressed, and I found no new blocking issues.

- The native TCP/H1 provenance gap is no longer orphaned: #510 is open with explicit AC for the real record-provenance → RequestContext → H1 safety-gate handoff, and `docs/RESUMPTION_TEST_PLAN.md` keeps true end-to-end H1 coverage deferred under #369 until that lands.
- `rt0.reject.store_unavailable` is now hermetic and shares `nativeEarlyDataReplayComposition` with `run()` rather than reconstructing the disabled-mode branch independently. The test deliberately supplies a permissive candidate gate, proves disabled composition strips it, then passes the returned null option into a real `NativeTlsConnection`, which is the right boundary for this slice.
- The earlier replay/metrics, unsafe-request dispatch, and bounded 425 responder fixes remain intact.
- PR description now matches the actual scope and explicitly references #510.
- CI run 954 is green on the current head.

Given the explicitly tracked #510 production follow-up, this Slice 2 PR is clean to merge.

docs/RESUMPTION_TEST_PLAN.md
```diff
@@ -70,38 +214,28 @@ real OpenSSL/QUIC peer, no CI/soak changes.
...
+  real OS worker threads, because no seam exists yet to pin a connection to
+  one deterministically.
 - **Soak scenarios**: repeated reconnect/rotation loops with memory/metrics
   checks over long runs.
 - **CI wiring**: a smoke subset plus an optional nightly/soak job.
 - **QUIC-transport-specific 0-RTT-rejection-falls-back-to-1-RTT variant.**
   The record-transport matrix is thoroughly covered
   (`src/tls/tls13_backend_tests.zig`); an equivalent pass explicitly over
   `Transport.quic` has not been added yet.
+- **True end-to-end native TCP/H1 record-provenance dispatch coverage**
```
Jaruso: **[P1] #510 currently tracks only the final H1 provenance handoff, but production native TCP 0-RTT cannot reach that handoff yet.** There are two earlier #366 composition gaps that are still unowned:

1. `NativeTlsConnection.issueSessionTicket()` calls `prepareNewSessionTicket(...)` without `max_early_data_size`, so the record/TCP ticket issuance path does not advertise an early-capable ticket.
2. `Tls13Backend.ServerEarlyDataPolicy.enabled` defaults to `false`, while `NativeTlsConnection.createWithOptions()` installs the resolver and replay gate but never calls `setServerEarlyDataPolicy(...)`. The repository has the replay-store mode/capacity config, but I cannot find a production native-TCP server-policy enablement call.

So even after #510 replaces the hardcoded `ctx.early_data.transport_early = false`, a real production native TCP client still cannot obtain an early-capable ticket from this server and have the server accept its `early_data` attempt. The future #510 E2E could accidentally hide this by using `issueEarlyCapableTicket()` / manually setting the server policy, exactly like the TLS harness does today.

Please either broaden #510 or create a linked follow-up that owns the remaining #366 production enablement (record-ticket `max_early_data_size` advertisement + operator/composition wiring of `ServerEarlyDataPolicy`) and keep #369's true native TCP/H1 process path deferred until the E2E starts from a **production-issued** early-capable ticket with the production server policy enabled. The docs/PR body should not describe the provenance handoff as the only remaining native TCP/H1 production gap until those prerequisites are owned too.

Jaruso (commented):
Re-review complete on head `1e53884d`.

The two P1s from my previous pass are addressed:
- the H1 provenance gap now has a real owner in #510 and remains explicitly deferred from #369;
- `rt0.reject.store_unavailable` is now hermetic and exercises the same production option-selection helper `run()` uses.

CI is green on this head.

I found one additional P1 during the final end-to-end feature-path sweep: #510 currently tracks only the final record-provenance -> H1 handoff, but native TCP 0-RTT is still not production-enabled far enough upstream to ever reach that handoff. I left the detailed finding inline. This needs to be owned explicitly before I would call Slice 2 merge-ready; otherwise #510's future end-to-end test can only succeed by manually enabling test-only ticket/policy state and still miss the production composition gap.

No other new findings in the Slice 2 code/test changes.

Jaruso (commented):
Re-review complete on `2ad277bd`. No new findings.

The last remaining P1 is addressed correctly:
- #510 is now open and explicitly owns the full native TCP production 0-RTT chain: production `max_early_data_size` ticket advertisement, operator-controlled server early-data policy composition, record-read provenance propagation, and the true native TCP/H1 end-to-end safety-gate test.
- `docs/RESUMPTION_TEST_PLAN.md` now keeps that complete production path explicitly deferred under #369 instead of treating only the H1 provenance handoff as outstanding.
- #369 itself is reopened and remains the owner of the remaining interop/restart/soak assurance.
- The PR description matches the actual slice boundary.
- The only change since the prior review is documentation/ownership clarification; no previously reviewed production/test code changed.
- CI is green on this head.

All prior review blockers are now accounted for without overstating Slice 2 coverage. **Good to merge.**

PR #511, opened 2026-07-25
tls: add real external resumption interop, restart, and soak assurance (#369 Slice 3)

Jaruso:
## Summary

Slice 3 of #369 (326-J). Slices 1 (#508) and 2 (#509) covered in-process scenario tests and process-level replay/425 assurance. This slice was originally scoped as "the final #369 slice" per the issue instructions, but **stops short of that** once two of its premises turned out to be false on inspection, before any test code was written:

1. **#510 is not "wire three call sites."** Its issue text assumed `NativeTlsConnection`'s record-read early-data provenance accessor already worked. It's a stub that always returns `false`; `record_epoch_bridge.Bridge` rejects the `.zero_rtt` epoch everywhere; `feedHandshakeToDriver` hard-fails any pre-handshake `application_data` record. Real 0-RTT decryption does not exist for the record transport at all, and native QUIC 0-RTT is unconditionally rejected in code independent of config. **0-RTT does not exist end-to-end on any transport today.** #510's text has been corrected with this finding (comment posted); it stays open as its own, larger effort comparable in size to the #366–#368 QUIC slices.
2. **#338/#358 (the external OpenSSL harness this slice was told to reuse) don't exist yet either.** Both open; no `s_client`/`s_server` TLS-over-TCP subprocess abstraction exists in the repo. This slice adds the minimal, resumption-scoped helpers it needs directly (`runOpenssl`, a `bounded_process.zig` stdin extension) rather than building a second generic framework.

Given both, **#369 stays open.** This slice proves everything honestly achievable against the actual current production surface (real 1-RTT resumption, external interop, real restart, soak) and leaves every 0-RTT-specific row explicitly open against #510's corrected scope.

**Filed #512** for a third gap found along the way: #326 requires ticket-key rotation, `ReloadableKeyRing` implements and unit-tests it, but it's never composed into production (`edge_gateway.zig`/`native_tls_connection.zig` never reference it) — the same orphaned-ownership pattern as #510, now with its own tracked issue rather than a paragraph in a test-plan doc.

### A previously undiscovered, unrelated production bug found and fixed first

Pointing a real `openssl s_client` at the native TLS listener for the first time (nothing in this repo had ever done this) failed every connection with `error.InvalidRecordType` right after the client's final handshake flight. Root cause: the record layer's ciphertext-mode parser required every post-ClientHello record's outer wire type to be `application_data`, with no exception for the unprotected, single-byte `change_cipher_spec` record TLS 1.3 middlebox-compatibility mode sends by default (RFC 9846 §5 / RFC 8446 Appendix D.4) — which OpenSSL, and most real-world TLS 1.3 clients, send. This is a universal external-interop bug, unrelated to resumption, that self-testing (Tardigrade's own client against its own server) never exercised. Fixed in `record_codec.zig`/`encrypted_stream.zig`, gated on the actual RFC-defined window (after the first ClientHello, before the peer's Finished — not just the record's wire shape) with regression coverage for all three boundaries; verified the exact repro round-trips a real HTTP response before building anything else on top of it.

### What's proven (see `docs/RESUMPTION_TEST_PLAN.md` for full detail)

- **External OpenSSL interop**: `interop.openssl.h1.resume` (full handshake → OpenSSL's own `-sess_out`/`-sess_in` → resumed reconnect → real HTTP response → cross-checked against both OpenSSL's `Reused`/`New` indicator and Tardigrade's `resumption_outcome_total{outcome="accepted"}` metric), `interop.openssl.h2.tls_resume` (TLS-layer scoped only — documented that it does not prove resumed H2 dispatch), and a negative matrix: `ticket.expired`, `sni_mismatch`, `alpn_mismatch`, `ticket.tampered` (byte-flip inside the real AEAD ciphertext, located via a plain search for its own public magic, offset past the envelope's fixed header — not a DER walk, and not the nonce). `cipher_mismatch` is deliberately **not** shipped — ad hoc testing found an unexplained discrepancy that needs its own investigation before a test can assert on it honestly.
- **Real process restart**: `restart.ephemeral.ticket_miss`/`fresh_ticket` (asserting the real typed `outcome="miss"` metric the A→B reconnect causes, not just that B's own later ticket resumes) and `restart.restart_and_credential_change.ticket_rejected` (both processes are `stateless` with independent fresh ephemeral keys, so this proves restart+credential-swap safety via the same key-loss mechanism, explicitly **not** certificate/auth-binding-change rejection — that needs a live-reload path the appliance profile currently forbids, or #512's rotation), using real separate OS processes (not `Runtime.deinit()`/`init()` reused in one test object).
- **Soak**: `soak.reconnect_resumption`, proxying through a real loopback origin and counting exactly-once dispatch via that origin's own atomic request counter (not a client-side "the response looked successful" counter, which can't distinguish real exactly-once dispatch from a duplicate-dispatch bug that only sent one response back). `TARDIGRADE_SOAK_HEAVY=1` scales 40→2000 iterations (verified locally at both sizes: heavy run was 2000 iterations, ~29s, `accepted=2000 executions=4000`, exact match at both sizes).
- **CI**: new `test-integration-resumption-interop` build step. Linux already gets this coverage via the existing unfiltered full-integration-suite step, so this new step is scoped to non-Linux (macOS) specifically to fill that gap without re-running the same tests twice; new `resumption-soak.yml` scheduled workflow for the heavy tier, mirroring `pki-differential.yml`; a CI invariant that no OpenSSL interop secret ever leaks into the cached `.zig-cache` tree (session files and the generated alternate credential's key live under `$TMPDIR` in a `0700` directory instead).

### #369 requirement → proof mapping

| #369 requirement | Proof |
|---|---|
| H1 external resumption | `interop.openssl.h1.resume` |
| H2 external resumption | `interop.openssl.h2.tls_resume` (TLS-layer only — does not prove resumed H2 dispatch, see deferred matrix) |
| H3 independent-peer resumption | **Not in this slice** — deferred, existing QUIC/H3 interop tooling (`scripts/interop/run-interop.sh`) needs resumption-specific extension |
| Accepted 0-RTT | **Blocked on #510's corrected scope** — 0-RTT decryption doesn't exist for record transport; QUIC 0-RTT unconditionally rejected |
| Rejected 0-RTT → safe 1-RTT | **Blocked on #510** (same reason) |
| Restart | `restart.ephemeral.ticket_miss`/`fresh_ticket`, `restart.restart_and_credential_change.ticket_rejected` |
| Expiry/mismatch/tampering | `interop.openssl.ticket.expired`/`sni_mismatch`/`alpn_mismatch`/`ticket.tampered`; `cipher_mismatch` deferred (documented); true certificate/auth-binding-change rejection deferred (documented, see restart row) |
| Rotation | **Blocked on #512** (filed this PR) — `ReloadableKeyRing` implemented/unit-tested but not composed into production; not faked with a test-only key system |
| No duplicate execution | `soak.reconnect_resumption`'s exact execution-count invariant, counted at the real upstream dispatch hook |
| Bounded soak state | `soak.reconnect_resumption` (execution/accepted counts); server-side cache occupancy has no exposed gauge — documented gap, not asserted against a metric that doesn't exist |
| Secret-safe reproducibility | `ticket.tampered`'s fingerprint-based log/diagnostic check; restart tests' log checks; CI invariant against `.zig-cache` secret leakage |

If any row above has no honest production-level proof, #369 remains open for it — this PR leaves #369 open.

## Test plan
- [x] `zig fmt --check .`
- [x] `zig build test` — 2691/2692 (1 pre-existing skip)
- [x] `zig build test-tls` — 759/759
- [x] `zig build test -Dtls-profile=appliance` — 2662/2663 (1 pre-existing skip)
- [x] `zig build test-integration-native-tls -Dtls-profile=appliance` — 9/9
- [x] `zig build test-integration-resumption-interop -Dtls-profile=appliance` — 9/9, both default (40-iteration) and `TARDIGRADE_SOAK_HEAVY=1` (2000-iteration) runs, exact counts
- [x] `zig build test-integration -Dtls-profile=appliance` (full suite sanity check) — 92/100 (8 pre-existing skips)

Refs #369
Refs #326
Refs #510 (corrected scope, comment posted)
Refs #512 (new — production rotation composition, filed from this PR's review)

### DIFF
--- a/docs/RESUMPTION_TEST_PLAN.md
+++ b/docs/RESUMPTION_TEST_PLAN.md
@@ -199,54 +199,213 @@ seam to `WorkerContext` to close this gap; until then this is a
 documented, intentional limitation of this slice's assurance, not a
 weakening of the "workers share one store" guarantee itself.
 
-## Explicitly deferred to a later slice
+## Slice 3 (this PR): external OpenSSL interop, real process restart, and soak
 
-- **External interop** with OpenSSL (`s_client`/`s_server` ticket
-  round-trips) and an independent QUIC peer for H3 resumption/0-RTT. Needs
-  subprocess-driven test harnesses, not covered here.
-- **Operational restart/rotation matrix** (the part of #369's restart
-  obligation the in-process `Runtime` model above does not reach):
-  - Old ticket issued → actual process restart / composition re-init →
-    resolver miss on the old ticket → connection still completes as a
-    usable full handshake → fresh ticket issuance succeeds with bounded
-    metrics recorded for the miss and the reissue.
-  - A persistent-overlap restart: generation N retained decrypt-only
-    alongside a newly current generation N+1 (or, for the stateless
-    keyring, a fresh non-overlapping nonce lease on reload), proving live
-    traffic straddling the reload never double-uses a nonce or drops a
-    still-valid ticket.
-  - Exact lifecycle boundaries (not-yet-active, active, decrypt-only grace,
-    fully retired) driven through a real reload/composition path rather
-    than direct `ReloadableKeyRing` calls.
-  - A failed reload retains the prior snapshot rather than leaving the
-    runtime with no usable key.
-  - Concurrent / cross-worker restart and rotation cases.
-- **Deterministic worker-thread-pinning test seam** — see "Known gap"
-  above. #369 Slice 2 proves process-shared-store sharing across
-  independent per-connection state; it does not drive that proof through
-  real OS worker threads, because no seam exists yet to pin a connection to
-  one deterministically.
-- **Soak scenarios**: repeated reconnect/rotation loops with memory/metrics
-  checks over long runs.
-- **CI wiring**: a smoke subset plus an optional nightly/soak job.
-- **QUIC-transport-specific 0-RTT-rejection-falls-back-to-1-RTT variant.**
-  The record-transport matrix is thoroughly covered
-  (`src/tls/tls13_backend_tests.zig`); an equivalent pass explicitly over
-  `Transport.quic` has not been added yet.
-- **True end-to-end native TCP/H1 production 0-RTT dispatch coverage**
-  (#510): the E2E must start from a production-issued early-capable native
-  TCP ticket (`max_early_data_size` advertised by the production ticket
-  path), use production server composition to enable
-  `Tls13Backend.ServerEarlyDataPolicy` only when replay protection
-  prerequisites are satisfied, then carry accepted/rejected TLS 0-RTT
-  record provenance through native TCP/H1 request parsing into
-  `RequestContext.early_data` and the production H1 safety gate. The
-  current slice proves the TLS/replay-store decision and the constructed H1
-  dispatch context on either side of that wider gap; it does not close the
-  missing production enablement or provenance handoff itself.
+This slice is explicitly **not** the final #369 slice, and #369 stays
+open after it. Two premises the original plan for a final slice assumed
+turned out to be false once actually checked against current `main`,
+before any test code was written:
 
-Acceptance criteria for #369 as a whole (interop reliability, safe
-fallback/rejection under mismatch, no double execution, bounded soak
-memory/cache growth, reproducible artifacts without leaking ticket keys)
-remain open until the deferred items above land in follow-up slices. #369
-is not complete after this slice.
+1. **#510's remaining scope is not "wire three call sites."** Its issue
+   text was written on the assumption that `NativeTlsConnection`'s
+   record-read early-data provenance accessor already worked and just
+   needed plumbing into H1. Inspection found
+   `PureZigRecordStream.currentReadTransportEarly()` is a stub that always
+   returns `false`, `record_epoch_bridge.Bridge` rejects the `.zero_rtt`
+   epoch everywhere (`error.UnsupportedRecordEpoch`), and
+   `feedHandshakeToDriver` hard-fails any pre-handshake `application_data`
+   record — meaning real 0-RTT decryption does not exist at all for the
+   record (TCP) transport, not merely "isn't wired to HTTP yet". Native
+   QUIC 0-RTT is also unconditionally rejected in code, independent of
+   configuration. So 0-RTT does not exist end-to-end on **any** transport
+   today, and no amount of wiring closes that — it needs new record-layer
+   capability comparable in size to the #366–#368 QUIC slices. #510's
+   issue text has been corrected with this finding; it remains open as its
+   own, larger effort, and is not attempted in this PR.
+2. **#338/#358 (the external OpenSSL interop harness this slice was told
+   to reuse) do not exist yet either.** Both issues are open; the repo has
+   `tests/crypto_openssl_diff.zig`/`pki_openssl_diff.zig` (differential
+   crypto fixtures, not a live TLS-over-TCP subprocess harness) and a real
+   external QUIC/H3 harness (`scripts/interop/run-interop.sh` +
+   `tests/h3_interop_tool.zig`), but no `s_client`/`s_server` abstraction.
+   This slice adds the minimal, resumption-scoped subprocess helpers it
+   needs directly in `tests/integration.zig` (`runOpenssl`, a
+   `bounded_process.zig` stdin extension) rather than building a second
+   generic OpenSSL interoperability framework — #338/#358 still own that
+   broader effort.
+
+Given both, this slice proves everything that is honestly achievable
+against the *actual* current production surface — real 1-RTT resumption
+end to end — and leaves every 0-RTT-specific external/production
+acceptance-criteria row explicitly open, tracked against #510's corrected
+scope rather than faked with test-only configuration.
+
+### A previously undiscovered, unrelated production bug found and fixed first
+
+Pointing a real `openssl s_client` at the native TLS listener for the
+first time (for the H1 interop case below) failed every single time with
+`error.InvalidRecordType` immediately after the client's final handshake
+flight — before any resumption logic was even reachable. Root cause:
+`record_codec.parseHeader`'s ciphertext-mode parser required every
+record's outer wire type to be `application_data`, with no exception for
+the unprotected, single-byte `change_cipher_spec` record that TLS 1.3
+middlebox-compatibility mode sends during the handshake (RFC 8446 Appendix
+D.4) — which OpenSSL, and most other real-world TLS 1.3 stacks, send by
+default. This is a universal external-interop bug, unrelated to
+resumption, that self-testing (Tardigrade's own client against its own
+server) never exercised. Fixed in `record_codec.zig`/
+`encrypted_stream.zig` (see that commit) with new unit coverage; every
+interop case in this slice runs against the fixed code and would not have
+passed before it.
+
+### External OpenSSL interop (`tests/integration.zig`, real `openssl` subprocess)
+
+- `interop.openssl.h1.resume` — real native TLS listener, real
+  `openssl s_client`, full handshake, OpenSSL's own `-sess_out`
+  session-file capture, resumed reconnect via `-sess_in`, a real HTTP
+  request over the resumed connection, and cross-checked authoritative
+  indicators from **both** sides: OpenSSL's own `Reused`/`New` line
+  (backed by `SSL_session_reused()`) and Tardigrade's
+  `tardigrade_tls_resumption_outcome_total{outcome="accepted"}` metric —
+  not merely a second successful handshake.
+- `interop.openssl.h2.resume` — the same proof scoped to the TLS layer
+  under the h2 ALPN path (handshake, ticket, resumed reconnect, ALPN
+  renegotiated to h2 again). Hand-rolling HPACK/frame encoding through raw
+  `s_client` stdin to drive one real h2 request was judged out of
+  proportion to what this case needs to prove, given H1 above already
+  proves resumption carries a real served request end to end and
+  production H2 dispatch has its own independent coverage; documented in
+  the test itself rather than silently narrowed.
+- `interop.openssl.ticket.expired` — 1-second ticket lifetime, a real
+  2-second sleep past it (there is no injectable clock at this external
+  boundary), reconnect falls back to a full handshake, connection remains
+  usable, a fresh ticket is issued afterward, and the expired-ticket
+  attempt never counts as `accepted`.
+- `interop.openssl.sni_mismatch` — the appliance TLS profile (the only
+  profile the native listener builds under) supports exactly one identity
+  and rejects `TARDIGRADE_TLS_SNI_CERTS` outright, so the only honest
+  externally-observable case is: any SNI other than the configured one,
+  ticket or not, is a deterministic `handshake_failure` alert — no request
+  is ever served, no secret material appears in the failure diagnostics,
+  and the listener remains fully usable via the correct SNI immediately
+  afterward.
+- `interop.openssl.alpn_mismatch` — ticket obtained under `h2`, reconnect
+  offering only `http/1.1`: falls back to a full handshake under the new
+  ALPN, old ticket never silently reused.
+- `interop.openssl.ticket.tampered` — flips one byte inside the real
+  `ticket_protection` AEAD envelope (stateless mode; located by its public,
+  non-secret `"TDTK"` magic via a minimal DER walk of the OpenSSL session
+  file, landing the flip inside the OCTET STRING's declared length rather
+  than a tag/length byte a blind offset would otherwise corrupt — which
+  would only break OpenSSL's own local session-file parser and prove
+  nothing about the server). No crash, safe fallback to a full handshake,
+  and the still-authentic ciphertext fingerprint next to the flipped byte
+  never appears in the server's log or the client's own stdout/stderr.
+- **`interop.openssl.cipher_mismatch` deliberately not shipped.** Ad hoc
+  verification found the reconnect still negotiated the *original* cipher
+  suite despite the client restricting itself via `-ciphersuites` to a
+  disjoint one — behavior that needs its own investigation before a test
+  can assert on it honestly. Shipping a test that silently asserts nothing
+  useful would be worse than not shipping one; left as a known gap.
+
+### Real process restart (`tests/integration.zig`, real separate OS processes)
+
+- `restart.ephemeral.ticket_miss` / `restart.ephemeral.fresh_ticket` — a
+  real process A (not `Runtime.deinit()`/`Runtime.init()` reused inside
+  one test object) issues a stateless ticket, is terminated, and a real
+  fresh process B with the same operator configuration proves: the old
+  ticket resolves as a miss rather than authenticating, the connection
+  still completes via a full handshake, B issues its own fresh ticket that
+  genuinely resumes within B's own lifetime, and neither process's log
+  ever contains the raw ticket ciphertext.
+- `restart.cert_change.ticket_rejected` — process B boots with a
+  different, test-generated throwaway credential under the same
+  `server_name`; the old ticket must not (and does not) bypass the new
+  authentication binding.
+- Both use the ephemeral **stateless** ticket-key policy specifically —
+  the currently supported restart-safety policy per #369 section 6.
+  Stateful mode's cache-miss shape after a restart is a different,
+  already-covered concern (an empty cache, not a key-loss question).
+
+### Reconnect soak (`tests/integration.zig`, in-process client against one long-lived server)
+
+- `soak.reconnect_resumption` — a bounded loop of full handshake → ticket
+  → resumed reconnect against one process, using the fast in-process
+  `PureZigTlsClient` rather than spawning `openssl` per iteration (that
+  external proof is already covered by the interop cases above). Tracks
+  iteration count and asserts the required invariant exactly —
+  application executions equal legitimate requests expected to execute,
+  never derived from a TLS-layer decision function itself — plus that
+  every resumed reconnect actually resumed. `TARDIGRADE_SOAK_HEAVY=1`
+  scales the default 40-iteration run up to 2000 for the scheduled heavy
+  workflow; both sizes verified locally (2000 iterations: ~27s,
+  `accepted=2000 executions=4000`, exact match, no failures).
+- **Known gap**: there is no exposed gauge for server-side
+  resumption-cache occupancy today (only issuance/outcome counters), so
+  "memory/cache growth converges to configured bounds" is proven only via
+  the cache's own existing bounded-capacity unit tests
+  (`test-session-cache`, #364) plus this loop never regressing the counted
+  invariants over many iterations — not via a live occupancy sample.
+  Documented rather than asserted against a metric that doesn't exist.
+
+### CI
+
+- `zig build test-integration-resumption-interop` (new build step,
+  filtered to the `interop.`/`restart.`/`soak.` case-ID prefixes) is wired
+  into the existing appliance-profile CI job's Linux-only tier — the same
+  scoping that job already applies to the full integration suite, since
+  this new suite is similarly more prone to runner-scheduling variance
+  (real subprocesses, real process spawn/kill, bounded real-time sleeps)
+  than the pure in-process native-TLS suite.
+- `.github/workflows/resumption-soak.yml` runs the same suite with
+  `TARDIGRADE_SOAK_HEAVY=1` on a weekly schedule plus manual dispatch,
+  mirroring `pki-differential.yml`'s existing pattern.
+
+## Explicitly deferred beyond this slice
+
+- **0-RTT external/production interop, entirely** (the largest remaining
+  gap): accepted 0-RTT, rejected-0-RTT-falls-back-to-1-RTT, and any
+  related metrics/replay assertions against real external peers, for
+  every transport. Blocked on #510's corrected (large) scope — see above
+  — not merely "not yet wired." Do not close #369 by treating any
+  in-process/constructed-context 0-RTT test (Slices 1–2) as equivalent to
+  this.
+- **Independent QUIC/H3 external interop for resumption/0-RTT.** The repo
+  has real external QUIC/H3 tooling (`scripts/interop/run-interop.sh` +
+  `tests/h3_interop_tool.zig`, driving ngtcp2/quiche/aioquic peers) that a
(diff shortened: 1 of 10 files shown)


src/tls/encrypted_stream.zig
```diff
@@ -1525,6 +1532,19 @@ pub const PureZigRecordStream = struct {
...
+            // RFC 8446 Appendix D.4: a middlebox-compatibility-mode peer sends
+            // an unprotected single-byte {0x01} change_cipher_spec record at
+            // any point up to its own Finished; it carries no protocol
+            // meaning and MUST be dropped without further processing rather
+            // than opened through the bridge (it was never encrypted, so
+            // `openProtected` would try to decrypt an unprotected record).
+            // `parseHeader` already bounds it to exactly one byte; any other
+            // payload here is the "any other change_cipher_spec value" case
+            // RFC 8446 requires aborting the handshake for.
+            if (record.content_type == .change_cipher_spec) {
```
Jaruso, line 1576: **[P1] Gate dummy CCS on the RFC 9846 handshake window, not only on `{0x01}` shape.** This branch runs for the server's `.initial` epoch too, so a client can currently send a syntactically valid dummy CCS *before its first ClientHello* and we silently discard it. RFC 9846 §5 only permits this after the first ClientHello has been sent/received and before the peer Finished; outside that window it is an unexpected record and must abort. The project also names RFC 9846 as the normative TLS reference, so the new RFC 8446-only comments should be updated.

Please make the acceptance stateful, e.g. conceptually:
```zig
if (record.content_type == .change_cipher_spec) {
    if (record.payload.len != 1 or record.payload[0] != 0x01)
        return self.fail(error.UnexpectedRecordContent);
    if (!self.firstClientHelloSeen() or self.peerFinishedSeen())
        return self.fail(error.UnexpectedRecordContent);
    continue;
}
```
Use the existing handshake lifecycle/state if it exposes these boundaries rather than inventing duplicate protocol state. Add regression tests for CCS-before-ClientHello rejected, CCS after CH/before peer Finished accepted, and CCS after peer Finished rejected.

tests/integration.zig
```diff
@@ -2431,6 +2432,1067 @@ fn requireNativeTlsProfile() !void {
...
+}
+
+fn opensslConnectArg(allocator: std.mem.Allocator, port: u16) ![]u8 {
+    return std.fmt.allocPrint(allocator, "127.0.0.1:{d}", .{port});
+}
+
+fn opensslSessionPath(allocator: std.mem.Allocator, port: u16, tag: []const u8) ![]u8 {
+    const cwd = try compat.cwd().realpathAlloc(allocator, ".");
+    defer allocator.free(cwd);
+    return std.fmt.allocPrint(allocator, "{s}/.zig-cache/interop-{s}-{d}.sess", .{ cwd, tag, port });
```
Jaruso: **[P1] Do not place serialized OpenSSL sessions under the cached `.zig-cache` tree.** `-sess_out` writes an encoded `SSL_SESSION`; OpenSSL's own `sess_id` documentation explicitly warns that session output contains the session master key and requires appropriate security precautions. This workflow caches `.zig-cache`. Although the normal-path `defer deleteFile` removes these files, cleanup errors are swallowed, so a successful job can still leave a resumable secret-bearing artifact inside a cache path.

Please put these files in a dedicated OS temporary directory that is explicitly outside every Actions cache/artifact path, create them with restrictive permissions where applicable, and clean that directory on teardown. I would also add a cheap CI/test invariant that no `interop-*.sess` exists anywhere under `.zig-cache` after the suite. Secret test material should fail safe even if cleanup itself fails.

tests/integration.zig
```diff
@@ -2431,6 +2432,1067 @@ fn requireNativeTlsProfile() !void {
...
+// path -- handshake, ticket capture/offer, and the resumed connection
+// negotiating h2 again -- not a full HTTP/2 request/response. Hand-rolling
+// HPACK/frame encoding through raw `openssl s_client` stdin to drive one
+// real h2 GET is out of proportion to what this case needs to prove and is
+// intentionally left out of scope (see docs/RESUMPTION_TEST_PLAN.md); H1
+// above already proves resumption carries a real served request end to
+// end, and production H2 dispatch itself already has its own coverage
+// (`tests/integration.zig`'s "native TLS listener dispatches ALPN h2..."
+// suite) independent of resumption.
+test "interop.openssl.h2.resume" {
```
Jaruso: **[P2] Keep the H2 row open at the application-integration boundary.** The comment here is honest: this proves PSK resumption while ALPN is `h2`, but it never sends an H2 frame/request on the resumed connection. #369's criterion is resumption across HTTP/2 where enabled; separate “H2 dispatch works” and “TLS resumes with h2 ALPN” tests cannot detect a bug that affects H2 dispatch *only after resumption*.

Either drive one real external H2 request (using an existing/appropriate H2-capable peer rather than hand-rolling HPACK if possible), or rename this to something like `interop.openssl.h2.tls_resume` and explicitly leave **application-level resumed H2 interop** in the deferred matrix. The PR's requirement table should not mark the full H2 row proven by this test.

Jaruso (commented):
I agree with the two major scope discoveries in this PR: #510 is materially larger than we thought, and #369 must stay open rather than manufacturing 0-RTT assurance against capability that production does not have. The new real OpenSSL/restart coverage is also valuable, and current CI is green.

That said, I found several issues in the new assurance itself that need to be corrected before I would merge this slice. The most important ones are: the CCS interoperability fix accepts a standards-forbidden state transition; the certificate-change test cannot causally test certificate binding; the restart test does not assert the claimed old-ticket miss; the soak's “execution” counter is client-side response accounting rather than an application-side-effect counter; and OpenSSL session files containing resumable session secrets are deliberately written under `.zig-cache`, which is a configured Actions cache path.

There are also a few P2 assurance/documentation problems (H2 is TLS-layer only, the tamper helper changes the nonce despite claiming ciphertext/DER-walk coverage, `bounded_process`'s stdin write is not actually deadline-bounded, and the new filtered CI suite currently runs a second time because the same tests remain part of the full integration suite).

Finally, the rotation gap now has the same ownership problem we previously found for #510: #326 explicitly requires safe ticket-key rotation, while production does not compose `ReloadableKeyRing`. Please give that missing production capability an open implementation owner rather than leaving #369-the-test-story as its only home.

**Call: not merge-ready yet.** These are mostly assurance-integrity fixes; I do not think the overall Slice 3 direction needs to be redesigned.

tests/integration.zig
```diff
@@ -2431,6 +2432,1067 @@ fn requireNativeTlsProfile() !void {
...
+    const raw = try decodeOpensslSessionFile(allocator, path);
+    defer allocator.free(raw);
+
+    const magic = "TDTK";
+    const ticket_start = std.mem.indexOf(u8, raw, magic) orelse return error.TicketMagicNotFound;
+    // The ticket envelope's fixed header (magic + key id + nonce + AEAD
+    // fields) leaves ample authenticated ciphertext after it in any
+    // realistic ticket; flipping well past the header and short of the
+    // very end stays inside the OCTET STRING's content either way.
+    const flip_at = ticket_start + 24;
```
Jaruso: **[P2] This does not flip ciphertext, and there is no DER walk here.** `ticket_protection.fixed_header_len` is 36 bytes; the envelope layout is magic/version/AEAD/reserved + key ID through byte 23, then nonce at bytes 24..35, then ciphertext. `ticket_start + 24` therefore flips the first **nonce** byte. Likewise, the 32-byte “ciphertext fingerprint” beginning at `+25` includes the remaining nonce bytes before ciphertext. The current test is still a useful authenticated-envelope tamper test, but its comments/docs/PR description materially overstate what it proves.

If ciphertext tampering is the intended case, derive the offset from the production format constant and stay inside the actual ticket OCTET STRING, e.g. conceptually:
```zig
const ciphertext_start = ticket_start + ticket_protection.fixed_header_len;
const flip_at = ciphertext_start + safe_ciphertext_offset;
```
Also either actually parse the DER TLV boundary you claim to walk, or remove the “minimal DER walk” claim and describe the public-magic search honestly.

tests/integration.zig
```diff
@@ -2431,6 +2432,1067 @@ fn requireNativeTlsProfile() !void {
...
+    defer fourth.deinit(allocator);
+    try std.testing.expectEqual(std.meta.Tag(bounded_process.Outcome).normal_exit, std.meta.activeTag(fourth.outcome));
+    try assertContains(fourth.stdout, "Reused, TLSv1.3");
+    try assertContains(fourth.stdout, "HTTP/1.1 200 OK");
+
+    // 9: bounded miss/fallback metrics recorded for the restart, and B's
+    // real resumption of its own ticket, both on B's own metrics endpoint.
+    var metrics = try sendPureZigTlsHttp1Request(allocator, process_b.port, "/status/metrics");
+    defer metrics.deinit();
+    try std.testing.expect((prometheusLabeledMetricValue(metrics.body, "tardigrade_tls_resumption_outcome_total", &.{
```
Jaruso, line 3407: **[P1] The restart test never asserts the old process-A ticket actually produced the claimed miss.** The comment says this proves a bounded miss/fallback outcome, but the only metric checked here is `accepted >= 1`, which comes from process B resuming **B's newly issued ticket** later in the test. If OpenSSL stopped offering A's old ticket entirely, the reconnect would still be `New, TLSv1.3`, the HTTP request would work, B's own ticket would resume, and this test would pass without exercising server-side old-ticket rejection at all.

Please snapshot/assert the typed negative outcome caused by the A→B reconnect before doing the fresh-B resumption. For the current stateless ephemeral-key policy that should be the production outcome exposed for an unknown-key resolver miss (for example `resumption_outcome_total{outcome="miss"}` and/or the corresponding `ticket_resolve_total{mode="stateless",result="rejected"}` delta). Then assert B's later `accepted` separately. That makes the causal path falsifiable.

tests/integration.zig
```diff
@@ -2431,6 +2432,1067 @@ fn requireNativeTlsProfile() !void {
...
+    // certificate-bound resumption contract).
+    var process_b = try TardigradeProcess.start(allocator, .{
+        .config_text = config_text,
+        .ready_https_insecure = true,
+        .ready_path = "/healthz",
+        .extra_env = &.{
+            .{ .name = "TARDIGRADE_TLS_CERT_PATH", .value = alt_cert.cert_path },
+            .{ .name = "TARDIGRADE_TLS_KEY_PATH", .value = alt_cert.key_path },
+            .{ .name = "TARDIGRADE_TLS_SERVER_NAME", .value = "tardigrade.test" },
+            .{ .name = "TARDIGRADE_TLS_NATIVE_RESUMPTION_MODE", .value = "stateless" },
```
Jaruso, line 3503: **[P1] This certificate-change test cannot reach certificate/auth-binding validation.** Process A and process B are both `stateless`, and every process constructs a fresh ephemeral ticket-protection key. B therefore cannot decrypt A's ticket at all; it resolves as unknown-key/miss *before* the recovered session's certificate binding could be inspected. The test would pass unchanged if certificate-binding compatibility were completely broken.

Please do not map this case to “certificate change rejects the old authentication binding.” Either:
1. relabel it as a combined `restart + credential change` safe-fallback case and keep true certificate-binding-change coverage deferred; or
2. exercise a real production credential-reload path where the resumption state/key remains valid across the credential replacement, if such a production path exists.

Given the current ephemeral-only restart policy, option 1 is probably the honest scope until production persistent/rotation/reload composition exists.

tests/integration.zig
```diff
@@ -2431,6 +2432,1067 @@ fn requireNativeTlsProfile() !void {
...
+            .psk_now_ctx = &clock_dummy,
+            .psk_now_fn = SoakClientClock.now,
+        });
+        defer resumed_client.destroy();
+        if (resumed_client.backend.core.psk_authenticated) accepted_count += 1;
+        try resumed_client.writeAllPlain(openssl_health_request);
+        const resumed_raw = try resumed_client.readPlainToEnd(allocator, 64 * 1024, 5_000);
+        defer allocator.free(resumed_raw);
+        try assertContains(resumed_raw, "HTTP/1.1 200 OK");
+        execution_count += 1;
```
Jaruso: **[P1] `execution_count` is not an application-execution counter.** It is incremented by the client after receiving one successful HTTP response from a static `return 200 alive` route. If the server accidentally dispatched the same request twice but emitted one response, this counter still increments once. Therefore the PR's mapping `No duplicate execution -> soak.reconnect_resumption's exact execution-count invariant` is not actually proven by this code.

Two valid fixes:
- For this 1-RTT reconnect soak, rename this to `successful_response_count`/`completed_request_count`, keep it as a liveness invariant, and map the exactly-once retry requirement to #509's real loopback-origin atomic execution counter; **or**
- Put a loopback origin behind this soak and increment an atomic counter only inside the actual origin/application callback, then assert exactly `iterations * 2` origin hits.

Do not call a client-side response counter an application-side-effect proof.

tests/bounded_process.zig
```diff
@@ -81,6 +92,18 @@ pub fn run(allocator: std.mem.Allocator, options: Options) std.mem.Allocator.Err
     var reaped = false;
     defer if (!reaped) reapProcessGroup(&child, pgid);
 
+    if (options.stdin) |input| {
+        const stdin = child.stdin.?;
+        stdin.writeStreamingAll(io, input) catch |err| {
```
Jaruso, line 119: **[P2] The new stdin path is not actually bounded end-to-end as its caller claims.** `deadline_end` is calculated before spawn, but `writeStreamingAll()` here receives no timeout, and stdout/stderr are not drained until after this write (and optional sleep) completes. A child that does not read stdin—or one that fills stdout/stderr while waiting—can block this call before the MultiReader/deadline machinery starts. Current OpenSSL inputs are tiny enough that this is unlikely today, but this is now a general test helper and its documented guarantee is stronger than its implementation.

Please either perform the write under the same deadline while concurrently draining child output, or deliberately constrain this API to a small bounded stdin payload and document/assert that bound so it cannot grow into a pipe-capacity deadlock later.

docs/RESUMPTION_TEST_PLAN.md
```diff
@@ -199,54 +199,213 @@ seam to `WorkerContext` to close this gap; until then this is a
...
+- **Rotation / persistent-overlap key policy — Outcome B.** Inspected
+  `edge_gateway.zig`/`native_tls_connection.zig`: `ReloadableKeyRing`
+  (`src/tls/ticket_protection.zig`) is not referenced by production
+  composition at all. Persistent-overlap rotation is not a production
+  capability today, full stop — not partially wired, not test-only
+  reachable. Per #369 section 7 Outcome B, this slice does **not** build a
+  test-only persistent-key system to manufacture rotation coverage; the
+  actual supported policy is ephemeral-only restart invalidation, which
+  the restart cases above cover. If persistent-overlap rotation is still
+  required for #326/#369's closure, that needs its own production issue
```
Jaruso: **[P1] Give production rotation an explicit implementation owner now; it is not conditional for #326.** This paragraph says “If persistent-overlap rotation is still required,” but #326's acceptance criteria explicitly require ticket keys to rotate without nonce reuse/indefinite validity, and #369 owns the external/operational proof of that behavior. We now know `ReloadableKeyRing` is not composed into production, so #369 cannot implement or test that AC by itself.

Please reopen the appropriate production issue or create a narrowly scoped follow-up for production keyring persistence/reload/rotation composition, link it here and from #326/#369, and keep the operational rotation matrix blocked on that owner. We just fixed exactly this orphaned-ownership failure mode with #510; don't leave the rotation equivalent parked only in a test-plan paragraph.

src/tls/encrypted_stream.zig
Jaruso: The broad window fix is in now, but there is still one standards edge here before I’d resolve this thread. On the server, `handshake_message_seen` flips as soon as *any handshake record fragment* is opened, before `driveReceive()` has reassembled/accepted a complete ClientHello. `Tls13Backend.receiveImpl()` is allowed to buffer an incomplete handshake frame and return successfully, so a peer can send `partial ClientHello record -> dummy CCS -> rest of ClientHello` and this code drops the CCS even though RFC 9846 §5 only permits it **after the first ClientHello message has been received**.

Please key this state to actual ClientHello acceptance, not first handshake-fragment arrival, and add the adversarial regression case above. The current happy-window test manually sets `handshake_message_seen = true`, so it cannot catch this distinction.

tests/integration.zig
```diff
@@ -2431,6 +2432,1277 @@ fn requireNativeTlsProfile() !void {
...
+    // ordinary range validation, not actually be free.
+    try env_map.put("TARDIGRADE_LISTEN_HOST", test_host);
+    try env_map.put("TARDIGRADE_LISTEN_PORT", "18080");
+    try env_map.put("TARDIGRADE_TLS_CERT_PATH", cert_path);
+    try env_map.put("TARDIGRADE_TLS_KEY_PATH", key_path);
+    try env_map.put("TARDIGRADE_TLS_SERVER_NAME", server_name);
+
+    const checked = try std.process.run(allocator, compat.io(), .{
+        .argv = &.{ integration_options.tardigrade_bin_path, "check", config_rel },
+        .environ_map = &env_map,
```
Jaruso, line 3241: **[P1] Keep this fixture-validation subprocess under the suite's bounded-process contract.** `std.process.run()` waits synchronously for `tardi check` with no deadline or process-group reap path. If `check` wedges because of a parser/I/O regression, this test can sit here until the outer 30-minute Actions job timeout, even though this slice explicitly requires every external child wait to be bounded.

Please run this through the same deadline/reap machinery as the other external peers (for example, extend `bounded_process.Options` with an optional environment map/copy if needed) and preserve the current “nonzero means fixture unsuitable -> SkipZigTest” behavior only *after* the child has terminated within the bound.

docs/RESUMPTION_TEST_PLAN.md
```diff
@@ -199,54 +199,249 @@ seam to `WorkerContext` to close this gap; until then this is a
...
+  invariants over many iterations — not via a live occupancy sample.
+  Documented rather than asserted against a metric that doesn't exist.
+
+### CI
+
+- `zig build test-integration-resumption-interop` (new build step,
+  filtered to the `interop.`/`restart.`/`soak.` case-ID prefixes) is wired
+  into the existing appliance-profile CI job's Linux-only tier — the same
+  scoping that job already applies to the full integration suite, since
+  this new suite is similarly more prone to runner-scheduling variance
```
Jaruso: **[P2] Update this after the CI de-duplication fix.** The workflow no longer wires the filtered resumption target into a Linux-only tier: Linux runs the full appliance integration suite (which already contains these tests) and skips the filtered step, while non-Linux runs the filtered target because the full suite is skipped there. The code is correct now, but this assurance document still describes the pre-fix topology.

Jaruso (commented):
Re-review of current head `f3d1f4e` after the first round of fixes.

The prior findings around secret-bearing OpenSSL session files, H2 overclaiming, ticket-tamper offset/description, restart miss causality, certificate-binding overclaiming, application execution counting, bounded stdin size, CI duplication, and production rotation ownership are all substantively addressed. I resolved those old threads. #512 is now the explicit owner for production persistent ticket-key rotation, and #369 correctly remains open for #510/#512/H3 and the other deferred rows. CI is green on this head.

Two merge-blocking assurance issues remain:

1. The CCS fix now rejects before-CH and after-completion cases, but the server opens the permitted window on the first **handshake record fragment**, not after a complete ClientHello has actually been accepted. I replied on the existing CCS thread with the concrete `partial ClientHello -> CCS -> remainder` counterexample. RFC 9846 §5 requires the first ClientHello *message* to have been received.
2. The latest alternate-certificate fixture check introduced a direct `std.process.run()` for `tardi check`. That child has no deadline/reap path, so a regression can hang the interop suite until the outer Actions job timeout. This violates the bounded-subprocess contract this slice is specifically establishing.

There is also a small documentation cleanup: `RESUMPTION_TEST_PLAN.md` still describes the filtered resumption step as Linux-only, but the deduped workflow now runs the filtered target only on non-Linux while Linux gets the cases through the full integration suite.

**Call: not merge-ready yet, but very close.** I do not see another architectural redesign here; fix the two bounded/state-machine issues and the stale CI sentence, then this should be ready for another final pass.

src/tls/encrypted_stream.zig
Jaruso: Fixed in 1d541885. Replaced `handshake_message_seen` with `firstClientHelloAccepted()`, which checks `read_epoch`/`write_epoch` transitions instead of handshake-fragment arrival. Epoch only advances via `advanceEpochOnSecret`, which only fires once the driver has derived real handshake secrets from a *complete* ClientHello — so a partial-ClientHello-fragment → dummy CCS → rest-of-ClientHello sequence can no longer slip a CCS through before acceptance.

Added the adversarial regression test you described (`"middlebox-compat change_cipher_spec spliced between two records of one still-incomplete ClientHello is rejected"` in `src/tls/tls13_backend_tests.zig`). One honest note: I verified empirically (by temporarily hardcoding `firstClientHelloAccepted()` to always return `true` and confirming the test still passed) that this exact interleaving is actually caught by a different, pre-existing layer — `record_codec`'s `nextClientHelloState` interleaving guard, which rejects a non-handshake record type appearing while a ClientHello is mid-reassembly (`error.InvalidRecordType`). `firstClientHelloAccepted()` is still the correct fix for the general acceptance-window bug you flagged (it now reflects true ClientHello completion rather than fragment arrival), and the new test pins both properties: the CCS is rejected, and the handshake never completes afterward, even if the client tries to continue with the rest of the ClientHello.

tests/integration.zig
Jaruso: Fixed in 1d541885. Added `environ_map: ?*const std.process.Environ.Map` to `bounded_process.Options` and routed `verifyApplianceAcceptsCredential`'s `tardi check` invocation through `bounded_process.run` with a 10s deadline, instead of a bare `std.process.run()`. The "nonzero exit -> SkipZigTest" behavior is preserved but now only evaluated once `bounded_process` confirms the child exited normally within the bound; a non-normal-exit outcome (timeout/signal) now fails the test loudly via `error.ApplianceCheckDidNotExitNormally` instead of hanging.

docs/RESUMPTION_TEST_PLAN.md
Jaruso: Fixed in 1d541885. Rewrote the CI section of `docs/RESUMPTION_TEST_PLAN.md` to describe the actual current topology: the filtered `test-integration-resumption-interop` step runs on non-Linux (where the full appliance suite is skipped), and is deliberately not run again on Linux since the full suite there already covers the same cases.

PR #513, opened 2026-07-26
tls: enable production 0-RTT and persistent ticket-key rotation

Jaruso:
## #510 production native TCP 0-RTT
- enables native TCP server early-data policy only when resumption and replay protection are both wired
- advertises `max_early_data_size` on production-issued tickets only when the production policy permits 0-RTT
- installs/decrypts/wipes `.zero_rtt` record keys in the pure Zig record stream/epoch bridge
- preserves sticky transport provenance through native TLS/H1 into `RequestContext.early_data`
- drives the real H1 safety gate so unsafe early requests receive 425 before dispatch
- covers replay-rejected early data fallback to usable 1-RTT semantics

## #512 persistent ticket-key rotation
- adds `TARDIGRADE_TLS_NATIVE_TICKET_KEYS_PATH` as an explicit opt-in persistent stateless ticket-key snapshot path; empty keeps existing startup-only ephemeral behavior
- decodes a versioned JSON snapshot into `ticket_protection.KeyConfig` and lets existing `ReloadableKeyRing`/`Protector` own validation, nonce leases, atomic publication, and teardown
- composes persistent snapshots into the shared native resumption runtime used by native TCP and QUIC/H3 stateless/hybrid ticket issue/resolve paths
- reloads configured snapshots through the existing SIGHUP config reload path; failed candidates reject the reload and leave the previous keyring active
- rejects appliance-profile filesystem-backed persistent ticket-key configuration consistently with the existing no-filesystem-watcher policy
- adds secret-safe reload outcome metrics via `tardigrade_tls_ticket_key_reload_total`

## Tests
- `zig fmt --check .`
- focused filters for `#510`, `#512`, persistent ticket-key snapshot parsing, and native persistent ticket-key config

## Known limitations
- this does not implement the external #369 operational interop matrix; it adds the production capability and metrics that matrix can exercise
- persistent ticket keys are scoped to native stateless/hybrid tickets; stateful handles continue to use the existing stateful cache path

Fixes #510
Fixes #512
Refs #369


src/tls/encrypted_stream.zig
```diff
@@ -1525,11 +1577,20 @@ pub const PureZigRecordStream = struct {
...
-            const opened = self.bridge.openProtected(epoch, record, &plaintext_buf) catch |err| return self.fail(err);
+            const opened = self.openHandshakeRecord(epoch, record, &plaintext_buf) catch |err| switch (err) {
+                error.AuthenticationFailed => {
+                    if (self.shouldDiscardUnauthenticatedEarlyRecord(epoch)) continue;
+                    return self.fail(err);
+                },
+                else => return self.fail(err),
+            };
             switch (opened.inner.content_type) {
                 .handshake => try self.driveReceive(epoch, opened.inner.content),
```
Jaruso: **[P1] Implement the record-TLS `EndOfEarlyData` transition instead of jumping directly to Finished.**

This path now successfully opens 0-RTT records, but the handshake state machine has no handling for `MessageType.end_of_early_data` anywhere outside the enum definition. For TLS 1.3 over records, once the server accepts early data, the client must send `EndOfEarlyData` before its Finished; that message marks the switch away from the 0-RTT traffic keys. A conforming external client will therefore send a handshake message that our server currently forwards straight into `driveReceive(.handshake, ...)`, where the backend/core does not expect it.

The internal test client has the same omission, which is why the new record-stream test can complete without exposing the bug.

I would expect the TLS backend/core to model this explicitly for the **record** profile only (QUIC deliberately does not use EndOfEarlyData), e.g. conceptually:

```zig
// Client: after server Finished, before client Finished.
if (self.profile == .record and self.early_data_accepted) {
    try emitHandshake(sink, .zero_rtt, .end_of_early_data, &.{});
    try sink.push(.{ .discard_epoch = .zero_rtt });
}
try emitClientFinished(sink);
```

and on the server:

```zig
.end_of_early_data => {
    if (self.profile != .record or !self.early_data_accepted or msg.body.len != 0)

    // Transition the expected inbound message to client Finished and retire
    // the early read keys only after this record has been authenticated.
    try sink.push(.{ .discard_epoch = .zero_rtt });
    self.core.expectClientFinished();
},
```

The exact helpers can differ, but the important requirements are: EoED is accepted only when record-mode 0-RTT was actually accepted, it is ordered before Finished, and the zero-RTT keys are retired at this transition rather than surviving until the entire handshake completes. Please add an interop-style test that actually sends EoED so the test cannot pass with both sides sharing the same omission.

src/tls/encrypted_stream.zig
```diff
@@ -1242,14 +1286,19 @@ pub const PureZigRecordStream = struct {
         if (self.lifecycle == .closing or self.lifecycle == .closed or self.lifecycle == .failed) return error.StreamClosed;
         if (self.pending_terminal_read_error) |err| return err;
         if (bytes.len == 0) return 0;
-        if (self.lifecycle != .open or !self.bridge.handshake_complete) return error.WouldBlock;
+        const write_epoch: events.EncryptionEpoch = if (self.lifecycle == .open and self.bridge.handshake_complete)
+            .application
+        else if (self.bridge.hasWriteKeys(.zero_rtt))
```
Jaruso, line 1331: **[P1] Enforce the ticket's `max_early_data_size` on outbound 0-RTT writes.**

`hasWriteKeys(.zero_rtt)` currently makes every pre-handshake `writePlaintext()` eligible for 0-RTT, with no cumulative byte accounting. That means a ticket advertised for (say) 4096 bytes can be used to send an arbitrary number of 0-RTT application bytes across repeated calls.

The TLS backend already exposes `earlyDataMaxBytes()` for exactly this allowance, but `RecordTransport.Backend` does not surface it to the record stream. Please carry the bound through the transport contract and decrement it here. A shape like this would make the invariant explicit:

```zig
// PureZigRecordStream
zero_rtt_sent: u64 = 0,

const max_early = self.handshake_driver.?.backend.earlyDataMaxBytes();
if (write_epoch == .zero_rtt) {
    if (self.zero_rtt_sent >= max_early) return error.WouldBlock; // or a dedicated limit error
    const remaining: usize = @intCast(max_early - self.zero_rtt_sent);
    n = @min(n, remaining);
}

const record = try self.bridge.sealProtected(write_epoch, .application_data, bytes[0..n], &record_buf);
if (write_epoch == .zero_rtt) self.zero_rtt_sent += n;
```

The server side needs symmetric accepted-byte accounting as well: before `appendInboundPlaintext(..., true)`, reject an early-data record whose cumulative plaintext would exceed the accepted allowance. RFC 8446 treats `max_early_data_size` as the maximum 0-RTT application-data payload; advertising the value without enforcing it leaves the production behavior outside that contract.

src/tls/encrypted_stream.zig
```diff
@@ -1540,6 +1601,33 @@ pub const PureZigRecordStream = struct {
...
+        }
+        return self.bridge.openProtected(epoch, record, out);
+    }
+
+    fn shouldDiscardUnauthenticatedEarlyRecord(self: *const PureZigRecordStream, epoch: events.EncryptionEpoch) bool {
+        return self.role == .server and
+            epoch == .handshake and
+            !self.bridge.handshake_complete and
+            self.handshake_driver != null and
+            self.handshake_driver.?.backend.earlyDataAttempted();
```
Jaruso, line 1724: **[P1] Bound the rejected-early-data discard path.**

This predicate stays true for the rest of the incomplete handshake whenever the ClientHello contained `early_data`. Consequently, *any number* of records that fail both the 0-RTT and handshake-key authentication attempts are silently discarded. That turns the intended fail-closed fallback into an unbounded attacker-controlled discard loop.

The backend's `earlyDataAttempted()` documentation specifically calls out that the record carrier needs this state so it can **boundedly** discard skipped first-flight ciphertext. The carrier therefore also needs a byte limit plus a counter. Note that `earlyDataMaxBytes()` is currently cleared to zero for rejected attempts, so you may need a separate accessor that preserves the attempted ticket/policy skip allowance rather than reusing the live accepted allowance.

Conceptually:

```zig
// Transport contract
pub fn earlyDataDiscardLimit(self: Backend) u32 { ... }

// Record stream
rejected_early_bytes: u64 = 0,

fn mayDiscardRejectedEarlyRecord(self: *PureZigRecordStream, record_wire_len: usize) bool {
    if (!self.handshake_driver.?.backend.earlyDataAttempted()) return false;
    const limit = self.handshake_driver.?.backend.earlyDataDiscardLimit();
    if (self.rejected_early_bytes + record_wire_len > limit) return false;
    self.rejected_early_bytes += record_wire_len;
    return true;
}
```

Once the bound is exhausted, fail the connection rather than continuing to swallow unauthenticated records. Please test both the exact-bound case and one-record/one-byte-over behavior for replay-rejected and replay-unavailable attempts.

src/tls/tls13_backend_tests.zig
```diff
@@ -2853,6 +2853,173 @@ const SocketHarness = struct {
...
+    });
+    try first.driveUntil(struct {
+        fn done(h: *SocketHarness) bool {
+            return h.bothComplete();
+        }
+    }.done);
+
+    var sink = DirectSink{};
+    defer sink.deinit();
+    var prepared = try first.server_engine.prepareNewSessionTicket(std.testing.allocator, .{
```
Jaruso, line 2888: **[P1] This does not satisfy #510's required production end-to-end test.**

The AC explicitly asks for a true native TCP/H1 test that starts from a **production-issued** early-capable ticket with the **production server policy** enabled, then carries accepted/rejected TLS 0-RTT record provenance all the way through HTTP dispatch. This test bypasses both production seams by manually calling `prepareNewSessionTicket(...)` here and later directly calling `setServerEarlyDataPolicy(...)` / `setEarlyDataReplayGate(...)`. The separate gateway test likewise starts from a synthetic provenance harness rather than this record stream.

Please add one test that crosses the whole boundary instead of proving the pieces independently. The expected shape is roughly:

```zig
// 1. First connection: real NativeTlsConnection + resumption runtime + replay store.
//    Complete the handshake, let maybeIssueSessionTicket()/issueSessionTicket()
//    produce the NST, and capture the ticket from the wire.
//
// 2. Reconnect with that ticket and client early-data intent.
//    Drive the real record stream, send an unsafe H1 POST as 0-RTT, and feed
//    the resulting connection through the same H1 request/context/preflight path.
//    Assert 425 and zero mirror/auth/middleware/upstream/handler side effects.
//
// 3. Replay the same early attempt (or force replay-unavailable).
//    Assert the early bytes never reach HTTP/application dispatch, the resumed
//    handshake still completes as ordinary 1-RTT, and a subsequent normal H1
//    request succeeds with transport_early == false.
```

This production-shaped test should also exercise the real `EndOfEarlyData` sequence; doing so would catch the interop gap called out in the other comment rather than allowing both sides of the in-process harness to share the same state-machine omission.

Jaruso (commented):
**Blocking review findings — I would request changes here, but GitHub does not allow the PR author to submit a `REQUEST_CHANGES` review on their own PR.**

I reviewed all 8 changed files against #510, including the issue's scope-correction comment.

The overall direction is good: the record epoch bridge now has real 0-RTT key slots, plaintext provenance is tracked alongside queued bytes, native ticket issuance is gated on the production server policy, and H1 now consumes transport provenance before the safety gate. CI is green on this head.

There are still four blockers before #510's acceptance criteria are complete:

1. **Record-mode accepted 0-RTT is not TLS 1.3 interoperable yet.** The handshake engine never emits or consumes `EndOfEarlyData`. RFC 8446 requires a client whose early data was accepted to send `EndOfEarlyData` before its Finished, and record TLS uses that message to transition away from the early-data traffic keys. The in-repo harness skips that state on both ends, so it can pass while a conforming external client fails.
2. **`max_early_data_size` is advertised but not enforced by the record carrier.** A client can keep writing 0-RTT records without decrementing the advertised allowance; an accepting server does not count accepted early plaintext; and the rejected-attempt path can discard authentication failures without a byte bound. The ticket limit needs to be an enforced transport invariant, not metadata only.
3. **Rejected early-record skipping is unbounded.** This is especially important because the backend contract comment explicitly says the carrier needs the attempted state to *boundedly* discard rejected first-flight ciphertext. The carrier currently has no access to the bound and never counts what it skips.
4. **The required true production end-to-end test is still missing.** The new `#510 record stream...` test manually calls `prepareNewSessionTicket(...)`, `setServerEarlyDataPolicy(...)`, and `setEarlyDataReplayGate(...)`; the H1 425 test separately injects synthetic read provenance. #510 explicitly requires one test crossing production ticket issuance + production policy composition + real record provenance + H1 dispatch for both accepted and rejected/replay cases.

Please add coverage for the record-TLS `EndOfEarlyData` transition, exact/over-limit early-data byte accounting, bounded rejected-record skipping, and the production ticket -> resumed connection -> H1 425/replay-fallback path. With those fixed, the rest of this implementation looks structurally aligned with the ticket.

src/tls/tls13_backend.zig
```diff
@@ -2721,11 +2756,24 @@ pub const Tls13Backend = struct {
...
+        var finished_transcript_hash = transcript_hash;
+        if (self.profile == .record and self.early_data_accepted) {
+            var ebuf: [handshake_header_len]u8 = undefined;
+            var ew = Writer{ .buf = &ebuf };
+            try ew.u8_(@intFromEnum(MessageType.end_of_early_data));
+            const eom_len = try ew.reserve(3);
+            ew.patch(3, eom_len);
+            const eom = ebuf[0..ew.len];
+            self.core.recordSent(eom) catch |err| return mapCoreError(err);
+            try sink.emitCrypto(.handshake, eom);
```
Jaruso: **[P1] `EndOfEarlyData` must still be protected with the 0-RTT traffic key.**

This fixes the *ordering* omission from my previous review, but the wire epoch is still wrong. RFC 8446 §4.5 (and the 0-RTT flow in Figure 4) requires the `EndOfEarlyData` message itself to be encrypted under `client_early_traffic_secret`; it marks the boundary after which the client switches to handshake traffic keys. Emitting it as `.handshake` means our in-process client/server can agree on a non-conformant transition while an external TLS implementation is entitled to reject it.

There are two supporting places that need to move with this change: `expectedLevel(.end_of_early_data)` currently returns `.handshake`, and `record_epoch_bridge.sealProtected(.zero_rtt, ...)` currently only permits `.application_data`, so simply changing this one enum argument is not enough. The record receive path also calls `driveReceive(epoch, ...)` with the *original* `.handshake` routing epoch even when `openHandshakeRecord()` actually opened the record with `.zero_rtt`.

I would expect a shape along these lines:

```zig
// Client record profile: EoED is the last message under early traffic keys.
try sink.emitCrypto(.zero_rtt, eom);
try self.emitDiscardKeys(sink, .zero_rtt);

// Record bridge: allow the narrowly-scoped EoED handshake content under
// zero_rtt (preferably through a dedicated API rather than permitting
// arbitrary early handshake messages).

// Server record path: preserve the epoch that actually authenticated it.
.handshake => try self.driveReceive(opened.epoch, opened.inner.content),

// Backend: .zero_rtt is accepted only for record-profile EoED; every other
// handshake message at that level remains UnexpectedTransportEpoch.
```

Please add a test that proves the EoED TLSCiphertext opens with the early read key and *does not* open with the handshake read key. That prevents both sides of the harness from sharing the same epoch mistake again.

src/edge_gateway.zig
```diff
@@ -3005,6 +3026,7 @@ fn handleConnection(conn: anytype, session: *ConnectionSession, cfg: *const edge
         request = parse_result.request;
         request_initialized = true;
     }
+    const request_transport_early = h2LastReadTransportEarly(conn);
```
Jaruso: **[P1] Preserve early-data provenance for bytes already buffered from a prior H1 read.**

`request_transport_early` only reflects reads performed inside the current `beginReadScope()`. That loses provenance for HTTP/1.1 pipelining:

1. one accepted 0-RTT TLS read contains `GET /safe ...` followed by `POST /unsafe ...`;
2. `readHttpRequest()` reads both requests, returns the first boundary, and this code shifts the second request into `session.pending_buf`;
3. the first request sees `transport_early = true`;
4. the next `handleConnection()` calls `beginReadScope()`, clearing the flag;
5. `readHttpRequest()` finds the second request already complete in `pending_buf`, performs **no connection read**, and `h2LastReadTransportEarly()` therefore returns false.

The second request was transported in 0-RTT but reaches `RequestContext` as ordinary 1-RTT, so an unsafe pipelined request can bypass the 425 gate.

The pending buffer needs provenance consumed/shifted in lockstep with its bytes. Because early data is a prefix of the connection, a compact early-prefix/range representation may be enough; a parallel per-byte queue is also correct. The important invariant is that moving `pending_buf[bytes_consumed..]` must move the corresponding provenance too, and the next request must derive its flag from the bytes it actually consumes rather than only from new socket reads.

Please add a regression test with two complete H1 requests in one early TLS read (safe first request, unsafe second request) and assert the second still has `transport_early == true`, returns 425, and performs zero mirror/auth/middleware/upstream/handler side effects.

src/tls/encrypted_stream.zig
```diff
@@ -1540,6 +1633,51 @@ pub const PureZigRecordStream = struct {
...
+        const wire_len: u64 = @intCast(record_codec.header_len + record.payload.len);
+        if (self.rejected_early_bytes + wire_len > limit) return false;
+        self.rejected_early_bytes += wire_len;
+        return true;
+    }
+
+    fn rejectedEarlyWireLimit(plaintext_limit: u32) u64 {
+        if (plaintext_limit == 0) return 0;
+        const overhead_per_record = record_codec.header_len + 1 + 16;
+        const records = (@as(u64, plaintext_limit) + record_codec.max_plaintext_fragment_len - 1) / record_codec.max_plaintext_fragment_len;
```
Jaruso: **[P1] The rejected-early wire budget breaks valid 1-RTT fallback when the early payload is fragmented.**

This is now bounded, which fixes the unbounded-discard problem, but `records = ceil(plaintext_limit / max_plaintext_fragment_len)` assumes the client used the *minimum* possible number of TLS records. `max_early_data_size` counts Application Data plaintext, not record framing/tag bytes.

Concrete example with the new 32-byte test allowance: two normal `write()` calls of 16 bytes each create two early records. Each record is roughly `5-byte header + 16 payload + 1 inner type + 16 tag = 38` wire bytes, so the server sees ~76 bytes. This helper allows only `32 + 22 = 54` bytes because it assumes one record. The second record therefore stops being discardable and the resumed handshake fails, even though the client sent exactly the permitted 32 Application Data bytes. That violates #510's requirement that replay-rejected/unavailable early attempts preserve ordinary 1-RTT resumption.

A robust implementation needs to remain bounded **without** assuming maximum-size fragments. The cleanest option is to derive/install a discard-only early read key for a rejected PSK attempt, decrypt the records, count actual Application Data plaintext, and never enqueue/expose it. If the design intentionally stays ciphertext-only, define a separate conservative record-count/resource cap that tolerates ordinary fragmentation (and accounts for padding) rather than deriving record count from `max_plaintext_fragment_len`.

Please test at least: permitted payload split over multiple records still reaches 1-RTT; exact permitted payload; one Application Data byte over; and a bounded pathological empty/padded-record case.

src/tls/resumption_runtime.zig
```diff
@@ -178,6 +184,23 @@ pub const Runtime = struct {
...
+        if (self.config.stateless_ticket_key_source != .persistent) return error.StaleSnapshotGeneration;
+        const keyring = if (self.keyring) |*keyring| keyring else return error.TooManyKeys;
+        const snapshot = try ticket_protection.Snapshot.build(self.allocator, configs, generation, self.provider.capabilities());
+        try keyring.install(snapshot);
+    }
+
+    pub fn loadPersistentTicketKeysFromFile(self: *Runtime, path: []const u8) (ticket_key_snapshot.LoadError || ticket_protection.SnapshotError)!void {
+        var loaded = try ticket_key_snapshot.loadFromFile(self.allocator, path);
+        defer loaded.deinit();
+        try self.installPersistentTicketKeys(loaded.generation, loaded.configs);
```
Jaruso, line 229: **[P1] Persistently reserve the next nonce lease before installing a reused ticket key.**

This production file path installs the lease exactly as serialized. That is safe only inside one continuously-running `ReloadableKeyRing`, whose ledger remembers previous ranges. On process restart the ledger and `NonceLease.next_counter` are new again, so loading the **same file** with the same key/prefix/`start` causes the same AEAD nonce sequence to be issued again. Two processes loading the same snapshot have the same problem concurrently.

That directly violates #512's “no nonce reuse” requirement and the prerequisite contract in #363: a reused key must receive a **durably reserved, non-overlapping** range before publication after restart, or old persisted keys must be decrypt-only while a fresh current key is generated.

The current `#512 persistent stateless restart...` test does not model production: it manually gives runtime B a different `[start=1,end=2)` lease after runtime A used `[0,1)`. Nothing in this loader performs that advancement.

I would expect one of these fail-safe production policies:

```zig
// Durable lease reservation path (conceptual)
lock(snapshot_state);
const reserved = reserveNextRangeDurably(key_id, old_end, lease_size);
fsync_and_atomic_replace(snapshot_state); // reservation is durable first
unlock(snapshot_state);
try runtime.installPersistentTicketKeys(generation, configsWith(reserved));
```

or load every preexisting persisted key without an encryption lease and create a fresh random current key/ID for issuance. Multi-process use must either coordinate the same durable allocator/lock or fail rather than sharing one range.

Please add a restart test that starts two sequential runtimes from the **exact same on-disk snapshot without manually editing the lease** and proves nonce reuse is impossible (plus a multi-process/competing-reservation test if shared files are supported).

src/gateway_shutdown.zig
```diff
@@ -185,6 +185,46 @@ pub fn hotReloadConfig(
             return;
         };
     }
+    if (worker_ctx.resumption_runtime) |runtime| {
```
Jaruso, line 196: **[P1] Preflight the ticket-key owner/candidate before any live TLS mutation, and do the restart-only check even when the runtime pointer is null.**

This block is reached only *after* `tls.updateProtocolPolicy(...)` and `native_credentials.reloadFromFiles(...)` have already published live changes. If the ticket-key source check or file load then rejects the reload, `prepared_version` is destroyed and the config store remains old, but the TLS policy and/or credential provider may already be new. That leaves runtime state inconsistent with the configuration reported as active.

There is a second owner-shape hole in the same conditional: a process started with native resumption disabled has `worker_ctx.resumption_runtime == null`, so the entire source check is skipped. A SIGHUP can then publish a config that enables stateless/hybrid persistent keys even though no process-scoped runtime/keyring was constructed at startup.

Please make this a two-phase reload:

```zig
// Phase 1: no live mutation.
const current = acquireCurrentConfig();
try rejectRestartOnlyRuntimeShapeChanges(current, proposed); // independent of ptr nullness
var ticket_candidate = try prepareTicketKeyCandidate(proposed); // parse/build/validate only
var credential_candidate = try prepareCredentials(proposed);   // where applicable

// Phase 2: only after every fallible preflight succeeded.
commitProtocolPolicy(...);
commitCredentials(&credential_candidate);
commitTicketKeys(&ticket_candidate); // commit path should be non-fallible or serialized
installPreparedConfig(...);
```

At minimum add the same kind of atomicity test already present for #368: combine a protocol/credential change with a bad persistent ticket snapshot and assert the rejected SIGHUP leaves **protocol policy, credentials, config store, and keyring** all unchanged. Also test disabled/no-runtime -> persistent mode is rejected as restart-required instead of publishing an unrealizable config.

src/gateway_shutdown.zig
```diff
@@ -185,6 +185,46 @@ pub fn hotReloadConfig(
...
+            state.reload_mutex.unlock();
+            state.metricsRecordReloadFailure();
+            state.metrics_mutex.lock();
+            state.metrics.recordTicketKeyReload(.reload_rejected);
+            state.metrics_mutex.unlock();
+            state.logger.warn(null, "config reload rejected: TARDIGRADE_TLS_NATIVE_TICKET_KEYS_PATH source mode changed; restart the process to switch between ephemeral and persistent ticket keys", .{});
+            return;
+        }
+        if (cfg_ptr.tls_native_ticket_keys_path.len > 0) {
+            runtime.loadPersistentTicketKeysFromFile(cfg_ptr.tls_native_ticket_keys_path) catch |err| {
```
Jaruso, line 198: **[P1] An unchanged persistent snapshot makes every unrelated SIGHUP fail as stale.**

Startup installs file generation `G`. On every later config reload with a non-empty path, this calls `loadPersistentTicketKeysFromFile()` unconditionally, builds a fresh snapshot with the same generation `G`, and passes it to `ReloadableKeyRing.install()`. That primitive correctly rejects any distinct positive replacement with `generation <= current.generation` as `StaleSnapshotGeneration`.

So once persistent keys are enabled, a SIGHUP intended only to rotate a certificate, change logging, etc. is rejected unless the operator also increments the ticket-key generation. Worse, rebuilding the same lease must *not* be used as the no-op solution because its nonce cursor has advanced in memory.

Please give the runtime loader an explicit unchanged outcome that does not rebuild/reset/install the key state. For example, retain a bounded internal fingerprint/version of the accepted persisted candidate and return something like:

```zig
const ReloadResult = enum { unchanged, installed };

if (file_generation < current_generation) return error.StaleSnapshotGeneration;
if (file_generation == current_generation) {
    if (!sameAcceptedSnapshot(file_candidate)) return error.GenerationConflict;
    return .unchanged; // keep current Snapshot and its live nonce cursor
}
try keyring.install(new_snapshot);
return .installed;
```

The exact identity mechanism can differ, but same-generation mutation must not be silently accepted and same-generation no-change must not reset nonce state or reject unrelated config reloads. Please drive this through the real `hotReloadConfig`: unchanged ticket file + unrelated setting change should succeed and the next issued ticket should continue from the existing nonce cursor.

src/tls/ticket_key_snapshot.zig
```diff
@@ -0,0 +1,173 @@
...
+
+pub fn loadFromFile(allocator: std.mem.Allocator, path: []const u8) LoadError!OwnedSnapshot {
+    const bytes = compat.cwd().readFileAlloc(allocator, path, max_snapshot_bytes) catch |err| switch (err) {
+        error.FileTooBig => return error.SnapshotTooLarge,
+        error.OutOfMemory => return error.OutOfMemory,
+        else => return error.SnapshotUnreadable,
+    };
+    defer allocator.free(bytes);
+    return parse(allocator, bytes);
+}
```
Jaruso, line 104: **[P2] Zeroize the raw snapshot buffer before freeing it.**

`OwnedSnapshot.deinit()` correctly wipes the decoded `key_storage`, but this allocation contains the same reusable ticket key as ASCII hex and is released straight back to the allocator. That leaves a second secret-bearing copy outside the typed secret cleanup path, including on parse/build failures.

At minimum this should be:

```zig
var bytes = try compat.cwd().readFileAlloc(...);
defer {
    std.crypto.secureZero(u8, bytes);
    allocator.free(bytes);
}
return parse(allocator, bytes);
```

Please also audit whether this `std.json.Value` parsing mode materializes separate copies of the `"key"` string (for example when escapes require allocation); any parser-owned secret copy needs the same wipe-before-free treatment or the loader should parse the secret field through a representation that does not duplicate it. This keeps the production loader consistent with #363's key-material wipe guarantees.

Jaruso (commented):
**Blocking re-review for `c25e2ab`, against both #510 and #512.**

The latest commit materially improves #510: the accepted-path `max_early_data_size` accounting is now enforced on both client writes and server plaintext, so that prior finding is addressed. I re-ran the rest of the review against #510's expanded record-layer scope and #512's production rotation requirements (including #363's nonce-lease invariants).

I still do **not** consider this merge-ready.

**#510 blockers**
- `EndOfEarlyData` is now present, but it is emitted/validated as a handshake-epoch message. TLS 1.3 requires the EoED record itself to be protected with `client_early_traffic_secret`; only records after it switch to handshake traffic keys. The in-process harness currently lets both sides share the same wrong epoch.
- H1 provenance is not sticky across `ConnectionSession.pending_buf`, so a second pipelined request that arrived in the same accepted 0-RTT read can lose `transport_early` and bypass the 425 safety gate.
- The rejected-early discard budget is bounded now, but its wire allowance assumes the minimum number of TLS records. A legal early payload split across multiple records can exhaust the budget before `max_early_data_size` application bytes have been skipped, turning a replay rejection into a fatal handshake instead of usable 1-RTT fallback.
- The existing review thread for #510's required **production-shaped NativeTlsConnection -> real H1 dispatch E2E test** remains unresolved; the new tests still exercise lower-level harnesses/direct configuration rather than that full AC.

**#512 blockers**
- Loading a persistent snapshot does not durably reserve/advance its nonce lease. Restarting (or running another process) from the same file resets the process-local ledger and lease counter to the serialized `start`, so the same AEAD key can reuse the same nonce sequence. #363 explicitly requires durable non-overlap before publishing a reused key after restart, or loading old keys decrypt-only and using a fresh current key.
- Any unrelated SIGHUP while the persistent file is unchanged re-installs the same positive generation, which `ReloadableKeyRing.install()` rejects as `StaleSnapshotGeneration`.
- Ticket-key validation/publication occurs after live TLS policy/credential mutations, so a rejected key candidate can leave the config store on the old config while some TLS state has already changed. The restart-only source check is also hidden behind `resumption_runtime != null`, allowing a process that started without the owner to publish a persistent-key config it cannot actually realize.
- The raw JSON allocation contains the hex-encoded ticket key and is freed without zeroization.

There is not yet a GitHub Actions run attached to this latest head, so this review is based on the code/AC pass rather than a green CI result.

src/edge_gateway.zig
```diff
@@ -2843,6 +2863,21 @@ fn streamingRequestBodyFromHead(
     };
 }
 
+fn h1PendingReadEarlyPrefix(old_pending_len: usize, total_read: usize, old_early_prefix_len: usize, read_scope_early: bool) usize {
+    const bounded_old = @min(old_early_prefix_len, old_pending_len);
+    if (!read_scope_early) return bounded_old;
+    return total_read;
```
Jaruso: **[P1] Preserve the exact early-data prefix length here instead of promoting the whole read to early.**

The pending-buffer under-marking from the previous review is fixed, but this boolean handoff creates the opposite error. `PureZigRecordStream` already keeps provenance per plaintext byte, yet `PlaintextProvenanceQueue.discard(n)` reduces the bytes returned by one `read()` to `saw_early: bool`; `WaitingEncryptedHttpConnection` ORs that across the scope; and this branch then turns any `true` into `total_read`.

That is incorrect once the handshake transition and application records are queued behind unread early plaintext. One `stream.read()` can drain a contiguous plaintext queue containing:

```text
[ accepted 0-RTT bytes ][ ordinary 1-RTT bytes ]
```

If those bytes contain an early request followed by a post-handshake pipelined request, `read_scope_early == true` makes `pending_early_prefix_len = total_read`. After consuming the first request, the second **ordinary 1-RTT** request remains tagged early and an unsafe method can be rejected with a spurious 425. That violates #510's requirement that ordinary 1-RTT use remain usable and makes `RequestContext.early_data` an inexact transport provenance signal.

Please carry a byte count (or equivalent exact range) through this seam instead of a boolean. For example:

```zig
const ReadProvenance = struct {
    early_prefix_len: usize = 0,
};

// Plaintext provenance queue: consume n bytes and return the number of
// leading bytes in that returned slice that were transport-early.
fn discard(self: *Self, byte_len: usize) ReadProvenance { ... }

// PureZigRecordStream / EncryptedStreamHttpConnection
pub fn currentReadEarlyPrefixLen(self: *const ...) usize { ... }

// WaitingEncryptedHttpConnection tracks how many bytes appended during this
// scope were early, rather than OR-ing `saw_early`.
```

Because TLS early data is a prefix of application traffic, the H1 session can then retain the exact early prefix after each request consumes bytes instead of conservatively extending it over later 1-RTT data.

Please add a regression that deliberately crosses the epoch boundary in one plaintext read: queue an accepted early H1 request, complete the handshake, queue a normal 1-RTT unsafe request, read both across the same adapter/read scope, consume the first request, and assert the pending second request has `transport_early == false` and is not rejected with 425 solely because of the preceding early request.

Jaruso (commented):
**Blocking re-review of `6862f71b` against both #510 and #512.** GitHub still will not let the PR author submit `REQUEST_CHANGES` on their own PR, so this is a COMMENT review with blocking findings.

I re-read the current 16-file PR diff, the #510 acceptance criteria + scope correction, #512's production-composition requirements, the still-open review threads, and the latest three commits since my previous pass. CI run 969 is green.

### Fixed since my last review

Several of the previous findings are now correctly addressed:

- #510 `EndOfEarlyData` is now emitted and consumed at `.zero_rtt`, and the record bridge can actually protect/open the handshake record with the 0-RTT keys before discarding them.
- #510 outbound and accepted inbound 0-RTT application bytes are cumulatively bounded by the ticket allowance.
- #510 rejected early-data trial-decrypt is now bounded without the earlier maximum-fragment assumption.
- #510 H1 provenance now survives an already-buffered all-early pipelined request instead of being cleared at the next read scope.
- #512 identical same-generation snapshots are fingerprinted and treated as no-ops, preserving the live nonce cursor across an unrelated SIGHUP.
- #512 raw snapshot JSON and parser-owned copies are now zeroed on teardown.

### Still blocking

1. **#510 still lacks the AC's true production native TCP/H1 E2E.** The record-stream test still manually invokes `prepareNewSessionTicket(...)` / `emitPreparedNewSessionTicket(...)` and directly installs the resolver, server early-data policy, and replay gate. The gateway tests still start from a synthetic provenance harness / constructed request context. There is still no single test that obtains the ticket through the real `NativeTlsConnection.issueSessionTicket()` production path, reconnects through the production composition, sends a real unsafe H1 request as 0-RTT, and proves the real H1 path returns 425 before side effects, followed by replay rejection + usable 1-RTT. That is an explicit #510 acceptance criterion and the prior review thread remains valid.

2. **#512 still has no production-safe durable nonce reservation across restart/processes.** The new restart test demonstrates the required behavior only by manually changing the second runtime's persisted lease from `[0,1)` to `[1,2)`. Nothing in `loadPersistentTicketKeysFromFile()` reserves or durably advances that range. Restarting the real process with the same configured snapshot therefore reconstructs the same AEAD key/prefix/counter start and can reuse a ticket-protection nonce; two processes pointed at the same snapshot have the same problem. This directly violates #512's no-duplicate-nonce requirement (and #363's durable lease contract). The previous nonce-durability review thread remains blocking.

3. **#512 hot reload is still not atomic/fail-closed at the composition boundary.** `hotReloadConfig()` mutates the live TLS protocol policy and may publish new native credentials before it attempts the persistent ticket-key reload. If the ticket snapshot then fails parsing/validation/install, the function reports the whole reload as rejected and leaves the previous config-store version active, but other live TLS state has already changed. The ticket source guard is also still inside `if (worker_ctx.resumption_runtime)`, so a process started without a runtime can reload into a config that requests persistent stateless/hybrid resumption without ever constructing the runtime it claims to have. The previous hot-reload review thread remains blocking; this needs prepare/validate-all-then-publish semantics (or equivalent rollback), not just a different ordering of the same fallible mutations.

4. **New inline P1: the pending-buffer fix now over-classifies a mixed 0-RTT/1-RTT read.** The TLS record queue retains byte-level provenance, but `currentReadTransportEarly()` collapses the just-read bytes to a boolean; `WaitingEncryptedHttpConnection` ORs that boolean across the read scope; then `h1PendingReadEarlyPrefix()` turns any `true` into `total_read`. A read that contains the tail of accepted early data followed by ordinary 1-RTT bytes therefore marks the entire suffix early. A legitimate 1-RTT pipelined POST can consequently receive a false 425. See the inline comment for the expected exact-prefix carrier shape and regression test.

### Closeout/documentation

`docs/RESUMPTION_TEST_PLAN.md` on the current base still explicitly says native TCP/H1 production 0-RTT does not exist and that #510 is not attempted. Once the remaining #510 work is complete, this PR should update that document so the repository does not merge a production capability while its authoritative assurance plan still describes the old architecture.

**Verdict: not merge-ready yet.** The new work materially improves the implementation and closes most of the earlier record-layer defects, but the remaining #512 nonce/reload guarantees and #510 production E2E are acceptance-criteria blockers, and the mixed-read provenance bug needs correction before the H1 safety handoff can be considered exact.

src/edge_gateway.zig
```diff
@@ -2967,18 +3030,22 @@ fn handleConnection(conn: anytype, session: *ConnectionSession, cfg: *const edge
                     return;
                 };
                 const bytes_consumed = parse_result.bytes_consumed;
+                const request_transport_early = h1ConsumeRequestEarlyProvenance(session, fallback_old_pending_len, total_read, bytes_consumed, h2LastReadEarlyPrefixLenBounded(conn, total_read));
                 if (bytes_consumed < total_read) {
                     const remaining = total_read - bytes_consumed;
```
Jaruso, line 3034: **[P1] Preserve the provenance collected by `readHttpRequestHead()` when this path falls back to the buffered parser.**

The exact-prefix change fixes the previous over-marking bug, but this branch now has the opposite under-marking case.

`h2BeginReadScope(conn)` is called once before the head pre-read. `readHttpRequestHead()` can therefore read accepted 0-RTT bytes and increase the scope's early prefix. But on `.fallback` / `.not_applicable`, `fallback_old_pending_len` is taken **after** that read, while `session.pending_early_prefix_len` has not been updated to describe those newly buffered bytes.

For a very concrete case, let the first read already contain a complete early POST with `Content-Length: 0`:

```text
old session pending:      0
readHttpRequestHead():    N bytes, all 0-RTT
session.pending_len:      N
session.pending_early_prefix_len: 0   // never committed
fallback_old_pending_len: N
readHttpRequest():        no socket read needed; request already complete
```

Then `h1PendingReadEarlyPrefix()` sees `bounded_old == 0 < old_pending_len == N` and returns 0, so this request reaches `RequestContext` with `transport_early == false`. An unsafe accepted 0-RTT request can therefore bypass the 425 gate whenever `mayNeedStreamingRequestBodyPreRead()` is active but this request resolves to fallback/not-applicable handling.

The cleanest fix is to keep the two reads in one provenance calculation: use the **original** `old_pending_len` captured before `readHttpRequestHead()` when consuming the final buffered request, because the read scope already spans both `readHttpRequestHead()` and `readHttpRequest()`. Alternatively, explicitly commit the head-read prefix into `session.pending_early_prefix_len`, reset the read scope, and then account only the second read.

For example, the one-scope shape is roughly:

```zig
const old_pending_len = session.pending_len;
const head_read = try gconn.readHttpRequestHead(...);
...
const total_read = try gconn.readHttpRequest(...);
const request_transport_early = h1ConsumeRequestEarlyProvenance(
    session,
    old_pending_len, // before *either* read
    total_read,
    bytes_consumed,
    h2LastReadEarlyPrefixLenBounded(conn, total_read),
);
```

Please add a regression through the actual pre-read branch (not just the standalone helper): a complete unsafe POST arrives as accepted 0-RTT, `mayNeedStreamingRequestBodyPreRead()` is true, eligibility is `.fallback` or `.not_applicable`, and the request must still return 425 with zero mirror/auth/rate-limit/upstream/handler side effects.

src/gateway_shutdown.zig
```diff
@@ -185,6 +205,28 @@ pub fn hotReloadConfig(
             return;
         };
     }
+    if (worker_ctx.resumption_runtime) |runtime| {
+        if (cfg_ptr.tls_native_ticket_keys_path.len > 0) {
+            runtime.loadPersistentTicketKeysFromFile(cfg_ptr.tls_native_ticket_keys_path) catch |err| {
+                worker_ctx.config_store.destroyVersion(prepared_version);
+                const msg = std.fmt.bufPrint(&state.last_reload_error, "native ticket-key reload failed: {s}", .{@errorName(err)}) catch "native ticket-key reload failed";
```
Jaruso, line 200: **[P1] The ticket-key candidate is still applied after other live TLS state has already been mutated.**

The new restart-only owner/source preflight above this is correct and fixes that half of my previous comment. The atomicity half is still open:

1. `tls.updateProtocolPolicy(...)` publishes the new live protocol policy.
2. `native_credentials.reloadFromFiles(...)` publishes new live credentials.
3. Only here do we parse/reserve/install the persistent ticket-key candidate.
4. If this call fails, the prepared config is destroyed and reload status says the reload was rejected, but the protocol policy and/or credential store may already contain the proposed values.

That leaves runtime state inconsistent with the config version reported as active, which is exactly what the fail-closed reload requirement is meant to avoid.

Simply moving this call earlier is not sufficient: `loadPersistentTicketKeysFromFile()` itself mutates durable lease state and can publish the keyring, so a later credential failure would just create the inverse partial-commit problem.

Please make reload genuinely two-phase. Conceptually:

```zig
// Phase 1: everything fallible, nothing live published.
const policy_candidate = try tls.prepareProtocolPolicy(...);
var credential_candidate = try store.prepareReloadFromFiles(...);
defer credential_candidate.deinit();
var ticket_candidate = if (path.len != 0)
    try runtime.preparePersistentTicketKeysFromFile(path)
else
    null;
defer if (ticket_candidate) |*c| c.deinit();

// Phase 2: serialized commit after every preflight succeeds.
tls.commitProtocolPolicy(policy_candidate);
store.commitReload(&credential_candidate);
try runtime.commitPersistentTicketKeys(&ticket_candidate); // durable reservation before key publication
worker_ctx.config_store.installPrepared(prepared_version);
```

The exact API can differ, but after the first live mutation there should be no ordinary validation/I/O failure left that can report the whole reload rejected.

Please add the cross-component regression from the prior review: proposed protocol/credential changes + a bad ticket-key candidate, then assert the rejected SIGHUP leaves **protocol policy, served certificate identity, keyring, and config store** all on the previous version.

src/tls/ticket_key_snapshot.zig
```diff
@@ -0,0 +1,301 @@
...
+    defer {
+        std.crypto.secureZero(u8, bytes);
+        allocator.free(bytes);
+    }
+    return parse(allocator, bytes);
+}
+
+pub fn reserveNonceLeasesInFile(allocator: std.mem.Allocator, path: []const u8, snapshot: *OwnedSnapshot) LoadError!void {
+    var changed = false;
+    for (snapshot.configs) |*cfg| {
```
Jaruso, line 117: **[P1] Make the nonce-lease reservation an inter-process atomic operation; this read/modify/rename only proves sequential restart safety.**

The new sequential restart path is directionally correct: one runtime advances the on-disk range before using it, and the next runtime advances again. But there is no ownership primitive around `loadFromFile()` → range advance → rewrite.

Two Tardigrade processes sharing the configured snapshot can therefore do this concurrently:

```text
file: key K, prefix P, lease [0, 100)

process A reads [0,100)
process B reads [0,100)
A computes [100,200)
B computes [100,200)
A renames its .tmp and installs K/P/[100,200)
B renames its .tmp and installs K/P/[100,200)
```

Both processes can now emit the same AEAD key + nonce counters. The in-process `ReloadableKeyRing` mutex/ledger cannot prevent this because each process has an independent keyring. This violates #512's explicit composed-path requirement of concurrent publication with **no duplicate nonce**.

The fixed `{path}.tmp` name also races under this scenario, but unique temp names alone would not solve the duplicate reservation.

The reservation needs an inter-process lock or an atomic compare/exchange style state transition, with the snapshot re-read **after acquiring ownership**. A reasonable shape is:

```zig
var lock = try acquireExclusiveLeaseLock(path); // e.g. sidecar lock
 defer lock.release();

var current = try loadFromFile(allocator, path); // re-read under lock
var reserved = try reserveNextNonOverlappingRange(&current);
try validateCandidateWithoutPublishing(&reserved);
try durableAtomicReplace(path, &reserved);       // reservation durable first
// release lock only after the durable reservation is committed
try runtime.installPreparedPersistentKeys(&reserved);
```

A separate durable high-water/lease sidecar is likely cleaner than rewriting the operator snapshot itself, because another process advancing the cursor currently changes the snapshot fingerprint for every already-running process; an unrelated SIGHUP can then see a changed **same-generation** file and fall into `StaleSnapshotGeneration` instead of the intended unchanged/no-op path.

Please add a competing-reservation test using the same on-disk snapshot from two independent runtimes/processes and assert their issued `(key_id, nonce_prefix, counter)` tuples cannot collide. The current sequential test does not exercise this race.

Jaruso (commented):
Blocking re-review at `b9cdaf2e`.

This push closes a lot of the previous review correctly. In particular, I now consider the prior #510 production-E2E blocker closed: the new live-process integration test obtains a production-issued early-capable NST, sends real native record 0-RTT through H1, proves the safe request executes, proves the unsafe request is stopped with 425 before upstream execution, and proves replay-rejected early bytes fall back to a usable ordinary 1-RTT request. The exact-prefix redesign also fixes the prior early→1-RTT over-marking case.

I still would not merge this head yet. I found three blocking issues in the current code:

1. **#510:** the streaming pre-read `.fallback` / `.not_applicable` path can lose provenance for bytes already read by `readHttpRequestHead()`, allowing an accepted unsafe 0-RTT request to be reclassified as ordinary 1-RTT before the 425 gate.
2. **#512:** ticket-key reload is still not atomic with the rest of SIGHUP. Protocol policy and credentials can be published before the persistent ticket-key candidate fails, while the config store remains on the old version.
3. **#512:** the new durable lease advancement is safe for sequential restarts, but the file reservation itself is not serialized across processes. Two processes can read the same range and install the same next key/prefix/counter range, violating #512's explicit no-duplicate-nonce concurrency requirement.

The new sequential on-disk restart test is a meaningful improvement, but #512 specifically requires the composed production path to retain the primitive's concurrent-publication safety, so the reservation step needs an inter-process ownership/CAS mechanism as well as the in-process keyring mutex.

CI run 970 also currently has a failing `Test appliance profile (ubuntu-latest)` job in the full integration-suite step; the native TLS listener step itself passed. That should be green before merge, independently of the code findings below.

Because this PR is authored by the authenticated account, GitHub does not allow me to submit `REQUEST_CHANGES`; this COMMENT review is intended as blocking.

src/tls/ticket_key_snapshot.zig
```diff
@@ -0,0 +1,301 @@
...
+    }
+    try serialize(&out, snapshot);
+
+    const tmp_path = try std.fmt.allocPrint(allocator, "{s}.tmp", .{path});
+    defer allocator.free(tmp_path);
+    compat.cwd().deleteFile(tmp_path) catch {};
+    {
+        var file = compat.cwd().createFile(tmp_path, .{ .truncate = true }) catch return error.SnapshotUnwritable;
+        defer file.close();
+        file.writeAll(out.items) catch return error.SnapshotUnwritable;
```
Jaruso, line 149: **[P1] Preserve restrictive permissions when rewriting the secret-bearing snapshot.**

This temp file is created without an explicit mode and is then renamed over the operator-provisioned snapshot. `DirCompat.createFile()` forwards these flags directly to `std.Io.Dir.createFile`; Zig's `Io.File.CreateFlags` aliases `std.fs.File.CreateFlags`, whose POSIX default mode is `0o666` and is only reduced by the process umask.

That means an operator can correctly provision this raw ticket-key JSON as `0600`, then the first successful nonce reservation can replace it with (for example under umask `022`) a `0644` inode. The `.tmp` itself also contains the reusable symmetric key and can remain behind on a crash/error path.

Please create the replacement with an explicitly restrictive mode and preserve/validate the original security metadata rather than relying on umask. At minimum the shape should be something like:

```zig
const original = try statSecretSnapshot(path);
try validateSecretPermissions(original);

var file = try compat.cwd().createFile(tmp_path, .{
    .truncate = true,
    .exclusive = true,
    .mode = 0o600, // or the validated original restrictive mode
});
```

If ownership/group or platform ACLs are part of the deployment contract, preserve those as well. Please add a POSIX regression that starts from a `0600` snapshot, performs a reservation/rewrite, then asserts the resulting path has no group/other read or write bits.

src/tls/ticket_key_snapshot.zig
```diff
@@ -0,0 +1,301 @@
...
+        file.flush() catch return error.SnapshotUnwritable;
+    }
+    compat.cwd().rename(tmp_path, compat.cwd(), path) catch return error.SnapshotUnwritable;
+}
+
```
Jaruso, line 199: **[P1] Make the rename itself durable before the reserved lease can be published.**

`file.flush()` above syncs the temp file contents, but this `rename()` changes the containing directory entry and there is no directory sync afterward. On POSIX, syncing the file does not necessarily make its directory entry durable.

That is a nonce-reuse failure mode, not just a generic persistence nicety:

```text
on disk: lease [0,100)
process writes+syncs temp [100,200)
rename(path.tmp, path)
process installs [100,200) and issues tickets
POWER LOSS before the directory entry is durable
restart can observe old path [0,100)
restart reserves [100,200) again
=> same key/prefix/counters can be reused
```

#363's invariant is specifically that the non-overlapping reservation is **durable before publication**, so the commit sequence needs to include the namespace update too. Conceptually:

```zig
try writeAll(temp, serialized);
try syncFile(temp);
try rename(temp, path);
try syncParentDirectory(path); // reservation is durable only after this
// only now may the runtime publish/use the reserved range
```

A dedicated durable atomic-replace helper/sidecar is also fine. Please add a fault-injection test around the persistence phases (or at least isolate the phases behind a seam) so publication cannot occur unless the durable commit reports success.

src/tls/resumption_runtime.zig
```diff
@@ -178,6 +186,33 @@ pub const Runtime = struct {
...
+
+    pub fn loadPersistentTicketKeysFromFile(self: *Runtime, path: []const u8) (ticket_key_snapshot.LoadError || ticket_protection.SnapshotError)!void {
+        var loaded = try ticket_key_snapshot.loadFromFile(self.allocator, path);
+        defer loaded.deinit();
+        const loaded_fingerprint = fingerprintPersistentTicketKeys(loaded.generation, loaded.configs);
+        if (self.persistent_snapshot_fingerprint) |accepted| {
+            if (std.mem.eql(u8, &accepted, &loaded_fingerprint)) return;
+        }
+        try ticket_key_snapshot.reserveNonceLeasesInFile(self.allocator, path, &loaded);
+        try self.installPersistentTicketKeys(loaded.generation, loaded.configs);
```
Jaruso, line 229: **[P1] Validate the candidate before mutating durable lease state.**

This order is still not fail-closed:

```zig
try ticket_key_snapshot.reserveNonceLeasesInFile(...); // durable file mutation
try self.installPersistentTicketKeys(...);             // build/validate/install afterwards
```

`installPersistentTicketKeys()` calls `Snapshot.build()` and `ReloadableKeyRing.install()`, which can still reject things the JSON parser does not validate: unsupported AEAD capability, invalid validity windows, duplicate key/key-material IDs, overlapping historical leases, stale generation, etc. So a reload that is reported **rejected** can already have advanced and rewritten the persistent nonce state.

Please split preparation from commit. The candidate's immutable key/window/generation semantics need to be validated against the current keyring *without publication* first; only a candidate known to be admissible should acquire the inter-process reservation lock, re-read current durable state, reserve/write/sync the next range, and then publish the corresponding prepared snapshot.

Conceptually:

```zig
var candidate = try self.preparePersistentTicketKeys(raw); // build + generation/key validation, no publish
 defer candidate.deinit();

var reservation = try reserveDurablyUnderLock(path, &candidate);
try self.commitPreparedPersistentKeys(&candidate, reservation);
```

Please test a structurally parseable but semantically invalid candidate (for example duplicate key IDs or invalid validity windows) and assert both the live keyring **and the on-disk lease/generation bytes** remain unchanged after the rejected reload.

Jaruso (commented):
**Additional blocking pass on the same `b9cdaf2e` head.** There has not been another code push; this review is a deeper audit of the persistence boundary rather than a repeat of the existing findings.

The three P1s from my immediately previous review remain open: the H1 streaming pre-read fallback can under-mark accepted 0-RTT provenance, SIGHUP publication is still cross-component non-atomic, and nonce lease reservation is not serialized across processes.

I found three additional #512 persistence problems in this pass:

1. **The lease rewrite can broaden filesystem permissions on the raw ticket-key snapshot.** The replacement temp file is created with the Zig default file mode rather than preserving/enforcing the secret file's restrictive mode, then renamed over the original inode. A provisioned `0600` key snapshot can therefore become group/world-readable depending on umask; the temp copy also contains the raw reusable key.
2. **The reservation is not crash-durable at the directory-entry boundary.** The temp inode is synced, but after `rename()` the containing directory is never synced before the newly reserved range can be installed and used. A crash can therefore leave the old directory entry visible after restart, allowing the same lease to be reserved again.
3. **Durable state is mutated before the candidate passes the keyring's semantic/generation validation.** `reserveNonceLeasesInFile()` rewrites the operator file before `Snapshot.build()` / `ReloadableKeyRing.install()` can reject duplicate keys, invalid windows, unsupported capabilities, overlapping leases, or stale generation. A reload reported as rejected can therefore still partially mutate persistent nonce state.

These are all on the security boundary #512/#363 is specifically trying to harden: durable non-overlap must be committed safely before publication, but invalid candidates must not partially commit persistent state.

CI run 970 is also still red in the Ubuntu appliance-profile full integration-suite step, so this head is independently not merge-ready until that is green or conclusively shown unrelated/flaky.

Because the authenticated account is the PR author, GitHub does not permit a `REQUEST_CHANGES` review; this COMMENT review is intended as blocking.

src/gateway_shutdown.zig
```diff
@@ -113,6 +113,51 @@ pub fn hotReloadConfig(
...
+        }
+    }
+    {
+        var current_lease = worker_ctx.config_store.acquire();
+        const persistent_ticket_reload_active = current_lease.cfg.tls_native_ticket_keys_path.len > 0 and cfg_ptr.tls_native_ticket_keys_path.len > 0;
+        const tls_policy_changed = !std.meta.eql(
+            gprotocol_policy.listenerPolicyFromConfig(current_lease.cfg),
+            gprotocol_policy.listenerPolicyFromConfig(cfg_ptr),
+        );
+        const native_credential_config_changed = applianceCredentialConfigChanged(current_lease.cfg, cfg_ptr);
```
Jaruso: **[P1] This guard still misses in-place credential rotation at unchanged paths.**

`applianceCredentialConfigChanged()` only compares the configured cert/key/SNI **strings**. In the general profile, however, the code below always calls `NativeCredentialStore.reloadFromFiles(...)`, which rereads the certificate and key bytes and publishes a new `ReloadableProvider` snapshot even when every path string is unchanged.

So this sequence still partially commits a rejected SIGHUP:

1. start with cert A at `/etc/tardi/cert.pem` and persistent ticket keys;
2. overwrite `/etc/tardi/cert.pem` with valid cert B, leaving the config/path unchanged;
3. make the persistent ticket snapshot malformed/stale/unreadable;
4. SIGHUP.

This guard sees `native_credential_config_changed == false`; `reloadFromFiles()` publishes cert B; `loadPersistentTicketKeysFromFile()` then fails; the prepared config is destroyed and reload reports failure, but new handshakes are already using cert B while the config/keyring remain on the previous state. That is the same atomicity invariant the guard is trying to protect, just through file-content rotation rather than path changes.

Please make this a real prepare/commit boundary. For example, load/validate a credential candidate **without publishing it**, prepare/validate/reserve the ticket candidate, and only after every fallible operation succeeds publish both snapshots and the config. Reordering the current mutating calls is not sufficient because whichever subsystem commits first can be stranded if the second one fails.

Please add a regression using the same cert/key path: serve cert A, replace the file contents with cert B, make the ticket candidate fail, SIGHUP, then assert the reload is rejected **and the server still presents cert A**, with the old keyring/config still active.

src/tls/ticket_protection.zig
```diff
@@ -512,6 +512,29 @@ pub const ReloadableKeyRing = struct {
         }
     }
 
+    pub fn validateInstallCandidate(self: *ReloadableKeyRing, replacement: *Snapshot) SnapshotError!void {
+        self.mutex.lock();
+        defer self.mutex.unlock();
+        if (self.current) |current| {
+            if (replacement.generation != 0 and replacement.generation <= current.generation) {
+                self.record(.{ .snapshot_rejected = .stale_generation });
```
Jaruso, line 520: **[P2] Do not invoke the observer while holding the keyring spin mutex.**

`validateInstallCandidate()` acquires `self.mutex` with a deferred unlock, then calls `self.record(...)` on every rejection path. That differs from `install()`, which deliberately unlocks before observer callbacks. This file already has `ReentrantDeinitObserver` coverage proving observers are allowed to re-enter/deinit the keyring; a rejection observer that calls back into `acquireCurrent()`, `deinit()`, etc. here will spin forever trying to reacquire the mutex held by this same call.

Please separate locked validation from notification, e.g. conceptually:

```zig
self.mutex.lock();
const result = self.validateInstallCandidateLocked(replacement);
self.mutex.unlock();

result catch |err| {
    self.record(.{ .snapshot_rejected = snapshotReason(err) });
    return err;
};
```

The exact helper shape can differ, but no external observer callback should execute under `ReloadableKeyRing.mutex`. Please add a rejection-path reentrancy test alongside the existing install observer test so this preflight API cannot regress the established callback contract.

Jaruso (commented):
Blocking re-review of `d3435d67` (COMMENT because GitHub does not allow me to request changes on my own PR).

This iteration closes a lot:

- **#510:** the streaming pre-read fallback now uses the pending length from before `readHttpRequestHead()`, so head-read 0-RTT provenance is no longer lost. Combined with the production live-process 0-RTT integration test already added in the previous iteration, I now consider the #510 acceptance criteria covered.
- **#512 durable reservation:** the new sidecar exclusive lock + re-read-under-lock gives cooperating processes a real serialization point before lease advancement. The rewrite now preserves/requires restrictive permissions, fsyncs the replacement, renames it, and syncs the parent directory. The candidate is also validated before durable mutation and revalidated after the lease shift. Those address my previous nonce-reservation, secret-file-permissions, crash-durability, and validate-before-mutate findings.
- The no-runtime/source-mode guard remains correctly outside the optional runtime block.

I still do **not** consider this merge-ready:

1. **P1 — hot reload is still non-transactional for in-place certificate rotation.** I left an inline comment on the newly-added guard. It compares credential *configuration strings*, but the general native credential store rereads and publishes certificate/key file contents on every SIGHUP. Changing the bytes at the same paths therefore bypasses the guard; a later ticket-key reload failure can still leave new credentials live while the config/keyring remain old.

2. **#512's composed concurrency proof is still missing.** The file reservation implementation now looks like the correct shape, but the issue explicitly requires proving concurrent publication has no duplicate nonce. The new test is sequential Runtime A -> Runtime B. The existing concurrent test races issuance against direct `installPersistentTicketKeys()` with manually adjacent leases; it does not put two runtimes through the same file/sidecar-lock reservation path. Please add a deterministic competing-reservation test (two runtimes/threads starting from one snapshot, synchronized so both contend for the same reservation file) and assert their issued key+nonce pairs are disjoint and the durable high-water advances monotonically. This is the remaining proof requested in the existing review thread, so I am not duplicating it inline.

3. **P2 — `validateInstallCandidate()` invokes the observer while holding `ReloadableKeyRing.mutex`.** I left an inline comment. This module already explicitly tests reentrant observer behavior on install; the new preflight path reintroduces a deadlock for a reentrant rejection observer.

4. **CI run 973 is red.** Both Linux unit-test jobs (`ubuntu-latest` and `ubuntu-24.04-arm`) failed while the macOS unit-test job passed. The connector exposes the failed unit-test steps but truncates the log before the specific assertion/error, so I cannot honestly attribute those failures to one of the findings above yet. They still need to be green before merge.

Once the in-place credential atomicity case, the competing file-reservation proof, the observer-under-lock issue, and the Linux CI failures are closed, I do not currently see another #510/#512 acceptance-criteria gap in the code already reviewed.

src/tls/early_data_replay.zig
```diff
@@ -387,6 +391,12 @@ pub const LocalStore = struct {
         return .rejected_capacity;
     }
 
+    fn quarantineApplies(self: *const LocalStore, c: Claim) bool {
```
Jaruso: **[P1] Do not use the ticket's wall-clock `issued_at` as proof that it was created after this process started.**

This exception reopens exactly the restart gap that the startup quarantine is meant to close for **persistent stateless tickets**.

A concrete failure sequence is:

```text
process A, wall clock ~= 2,000 ms
  -> persistent key K issues ticket T with authenticated issued_at = 2,000
  -> T is used for 0-RTT; A's in-memory replay store records it

restart + wall-clock rollback
process B starts at wall clock = 1,000 ms
  -> persistent key K survives, so T still resolves
  -> replay store is new/empty; quarantine starts at 1,000

while B is still inside its quarantine, wall clock reaches >= 2,000
  -> T is no longer `not_yet_valid`, so ordinary resumption/freshness can reach the replay gate
  -> `issued_at (2,000) >= quarantine_started (1,000)`
  -> `quarantineApplies()` returns false
  -> the empty store accepts T for 0-RTT even though A's replay history was lost
```

The existing backward-clock handling cannot detect this case because its `high_water_unix_ms` is also process-local and was reset with the store. A modest rollback can therefore age back past `issued_at` before the quarantine window itself expires.

This also conflicts with the #368 startup/restart contract this store was built around: freshly started process-local stores must fail closed while the recording window overlaps startup; the map being empty is not evidence that a ticket has not already authorized early execution in the previous process.

Please make the exemption depend on **process-incarnation provenance**, not timestamp ordering. The safest fix is to restore unconditional startup quarantine. If avoiding quarantine for newly-issued tickets is required, carry an explicit proof from the resolver, e.g. a stateful ticket resolved from the current process's non-persistent cache can be marked current-incarnation; a stateless ticket authenticated under a persistent key must remain quarantined unless its protected state contains a fresh process-epoch/incarnation value that cannot survive restart.

Please add the regression above with a persistent stateless ticket: issue/use it in runtime/process A, restart the replay store with the same persistent ticket key at a rolled-back clock, advance the new clock to `>= ticket.issued_at` but `< quarantine_until`, and assert early data is still `.unavailable` while ordinary 1-RTT resumption remains usable.

Jaruso (commented):
Re-reviewed the full PR at `14efea26001e38199e4624b6e29109dbd4c47468` against #510 and #512.

The three issues from my previous pass are now addressed:

- same-path native credential rotation is prepared without publication before the persistent ticket-key reload, so a rejected ticket candidate no longer strands a newly loaded certificate;
- `validateInstallCandidate()` releases the keyring mutex before observer callbacks;
- #512 now has a genuine two-runtime competing reservation test proving the same persistent snapshot produces disjoint nonce counters, in addition to the durable lock/re-read/fsync/permission work already present.

I also rechecked the #510 production path: production ticket issuance/policy composition, record provenance, H1 425-before-side-effects behavior, accepted early data, replay fallback, and subsequent usable 1-RTT are all represented end-to-end. CI run 977 is green.

**One P1 remains, introduced by the latest replay-store change, so I still consider this blocking.** The new startup-quarantine exception treats the authenticated ticket `issued_at` wall-clock value as proof that a ticket belongs to the current process incarnation. That proof does not survive restart/clock rollback, particularly for #512 persistent stateless tickets: a ticket issued and potentially consumed by process A can survive into process B, while process B's replay map is empty. If B starts after the wall clock has moved backward, then later (still inside its startup quarantine) the clock reaches the old ticket's `issued_at`, this helper classifies the old ticket as post-start and permits the claim. The exact replay history that startup quarantine exists to cover has been lost.

I left the concrete restart/rollback sequence and fix/test shape inline. The safe default is to keep startup quarantine unconditional unless the resolver carries a restart-stable proof that the ticket was created by this process incarnation (stateful process-local provenance could qualify; a wall-clock timestamp alone cannot).

Once that is corrected, I do not currently have another #510/#512 code blocker on this head.

src/tls/early_data_replay.zig
```diff
@@ -387,6 +392,11 @@ pub const LocalStore = struct {
...
+        _ = self;
+        return !c.current_process_stateful;
```
Jaruso, line 399: **[P1] Do not let the current-process exemption bypass the backward-clock quarantine.**

This fixes the restart/persistent-stateless bug from my previous review, but `quarantineApplies()` is used for **both** kinds of quarantine in `claimLocked()`:

1. the initial startup quarantine, where `current_process_stateful == true` is a sound exemption because an in-memory stateful handle cannot have survived a prior process; and
2. the quarantine re-armed when `now_unix_ms < high_water_unix_ms`, which protects a different invariant: a wall-clock rollback can make previously expired replay windows fresh again.

The second one must still apply to current-process stateful tickets.

A deterministic failure sequence is:

```text
store capacity = 1; startup quarantine already over

T (current-process stateful) claims at t=160
  -> accepted, retained through t=170

t=171: claim different key U
  -> T is expired, capacity reclaim removes T
  -> U is inserted; T is no longer remembered

clock rolls back to t=165
  -> claimLocked sees 165 < high_water(171)
  -> re-arms quarantine through 171 + quarantine_duration

replay T with current_process_stateful=true and a fresh deadline
  -> quarantineApplies(T) == false
  -> T is absent because it was reclaimed
  -> replay is accepted again during the very rollback quarantine that should fail closed
```

At the TLS layer an attacker can present the same ticket identity in a new ClientHello with an age consistent with the rolled-back server time, so the resolver provenance does not make this safe. The existing `backward wall-clock movement re-arms the quarantine window` test only exercises the default `current_process_stateful=false`, so it misses this regression.

Please separate the two concepts instead of using one predicate for both. For example:

```zig
// Initial startup quarantine: current-process stateful tickets may bypass.
if (!c.current_process_stateful and now_unix_ms < self.startup_quarantine_until_unix_ms)
    return .unavailable;

// Rollback quarantine: nobody bypasses this.
if (now_unix_ms < self.rollback_quarantine_until_unix_ms)
    return .unavailable;
```

When a backward jump is detected, set/extend `rollback_quarantine_until_unix_ms = high_water + quarantine_duration` (saturating as today). Equivalent state is fine; the invariant is that process provenance only exempts the startup loss-of-history case, never the rollback case.

Please add the capacity-1 regression above with `current_process_stateful=true` and assert the replay stays `.unavailable` until the re-armed rollback quarantine expires.

Jaruso (commented):
Blocking comment because this account is the PR author and GitHub will not allow REQUEST_CHANGES.

The previous P1 is fixed for the case I called out: persistent/stateless tickets now remain subject to startup quarantine, while only a resolver-proven hit from this process's in-memory stateful cache is exempt. I rechecked the #510/#512 paths touched by this delta and did not find a regression there. CI run 978 is fully green.

There is one remaining P1 in the new exemption itself: `quarantineApplies()` now controls both the initial startup quarantine and the quarantine that `claimLocked()` deliberately re-arms after a backward wall-clock jump. A current-process stateful ticket is safe to exempt from the *startup* quarantine because its stateful handle cannot survive process restart, but it is not safe to exempt from the *rollback* quarantine. After a replay entry has expired and been reclaimed, rolling the clock backward can make that ticket's expected-arrival window fresh again; bypassing the re-armed quarantine lets the same ticket identity be accepted for 0-RTT again.

I left the concrete sequence and expected fix inline. This is the only blocker I found on this pass. Once startup-only exemption is separated from rollback quarantine and the regression is covered, I expect this PR to be merge-ready.

Jaruso (commented):
Re-reviewed head `f240693f6373eb8fd7bc90fd9feb09699e9f8813` against the prior reviewed head and the #510/#512 acceptance criteria.

The remaining replay-quarantine blocker is fixed correctly. Startup and backward-clock quarantine are now distinct: `current_process_stateful` exempts only the startup loss-of-history case, while rollback quarantine is unconditional. The new capacity-1 regression reproduces the exact accepted -> expired/reclaimed -> clock rollback -> replay sequence and keeps the replay unavailable. The earlier persistent-stateless restart fix remains intact, and this commit touches only `src/tls/early_data_replay.zig`, so the previously verified #510 transport/provenance and #512 persistent rotation/nonce/reload fixes are unchanged.

I found no new code blockers in this pass. CI run 979 is still in progress at review time; all completed core gates I inspected (Ubuntu/macOS unit tests, format, lint, appliance integration, and completed builds) are green. Code-review-wise this is clear; merge should wait only for the current CI run to finish successfully.

PR #514, opened 2026-07-26
quic: add path/Retry mechanical primitives and PathManager validate/promote split (#387 Slice 1)

Jaruso:
## Summary

Slice 1 (Phases A+B) of #387's production-integration plan for wiring
authenticated path migration and Retry into the native HTTP/3 listener.
This slice lands the mechanical primitives and `PathManager` contract
changes the later `Connection`/HTTP-3-runtime integration slices depend
on. No runtime or operator-facing behavior changes yet — `migration_policy`
has no config wiring until a later slice, and `Connection` doesn't call
`PathManager` yet (confirmed via search — only the standalone
`tests/quic_h3_smoke.zig` harness does today).

- **`src/quic/config.zig`**: fix `disable_active_migration` mapping so both
  `.disabled` and `.nat_rebinding_only` advertise it; only `.full` doesn't.
- **`src/quic/path.zig`**:
  - Retry tokens / `RetryContext` now carry the Retry SCID alongside the
    original DCID (`issueRetry`/`validateRetry` signatures updated).
  - `PathManager.init()` takes an `initial_address_validated` bool so a
    non-Retry server's initial path starts amplification-limited until the
    handshake validates it; added `markActiveValidated()`.
  - Added per-path authenticated byte accounting:
    `recordReceivedOnPath`/`recordSentOnPath`/`canSendOnPath`, isolated per
    path so candidate-path bytes never inflate the active path's budget.
  - Added `nextValidationDeadlineUs()` so a connection driver can fold path
    validation deadlines into its own timer wheel.
  - Split `onPathResponse()` into `validatePathResponse()` +
    `promoteValidated()` so host migration can claim a fresh peer CID
    between validating a candidate and promoting it to active; added
    `recordMigrationBlockedNoPeerCid()` / `migrations_blocked_no_peer_cid`
    metric for the case where no fresh CID is available.
- **`src/quic/packet.zig`**: added `writeRetryV1()` (full QUIC v1 Retry
  packet writer), refactoring the RFC 9001 §5.8 integrity-tag computation
  into one shared helper used by both the writer and `verifyRetryIntegrity`.
- **`src/quic/frame.zig`**: added `encodeNewConnectionId()`.
- **`src/quic/cid.zig`**: added `CidRoutingTable.contains()` for a
  collision preflight that doesn't perturb routing hit/miss metrics.
- **`tests/quic_h3_smoke.zig`**: updated the two `PathManager` call sites
  for the new `init()`/`validatePathResponse`+`promoteValidated` APIs.

Remaining phases (Connection path-aware ingest/egress, CID lifecycle,
HTTP/3 runtime + Retry accept state machine, edge config/gateway wiring)
are tracked in #387 and will land as further slices.

## Test plan
- [x] `zig build test --summary all` — 2701/2702 passed (1 pre-existing skip)
- [x] `zig build test-quic --summary all` — 533/533 passed
- [x] New/updated tests cover: migration-policy transport-parameter mapping
      for all three policy values; Retry token round-trip with Retry SCID;
      `writeRetryV1` round-trip through `parsePacket`/`verifyRetryIntegrity`
      plus tamper/wrong-ODCID/oversized-CID/short-buffer cases;
      `encodeNewConnectionId` round-trip and short-buffer case;
      `CidRoutingTable.contains()` metric isolation; unvalidated-initial-path
      amplification gating; per-path byte-accounting isolation between
      active and candidate paths; `nextValidationDeadlineUs`;
      validate/promote split including the no-peer-CID-block-then-later-promote
      case and promotion-without-prior-validation rejection.

Refs #387 (Slice 1 of the production-integration plan; issue stays open
for the remaining phases).

### DIFF
--- a/src/quic/packet.zig
+++ b/src/quic/packet.zig
@@ -286,29 +286,83 @@ const retry_integrity_key_v1 = [16]u8{
 const retry_integrity_nonce_v1 = [12]u8{
     0x46, 0x15, 0x99, 0xd3, 0x5d, 0x63, 0x2b, 0xf2, 0x23, 0x98, 0x25, 0xbb,
 };
+/// Largest Retry body (header + token, excluding the tag) this module builds
+/// a pseudo-packet for.
+const max_retry_pseudo_body_len = 1500;
+
+/// Compute the RFC 9001 §5.8 Retry integrity tag over the pseudo-packet
+/// (`ODCID length || ODCID || Retry packet without tag`). Shared by
+/// `writeRetryV1` (compute) and `verifyRetryIntegrity` (compare), so there is
+/// exactly one implementation of the AAD construction.
+fn computeRetryIntegrityTag(original_dcid: []const u8, retry_body: []const u8) error{ InvalidConnectionId, RetryBodyTooLong }![retry_integrity_tag_len]u8 {
+    if (original_dcid.len > max_cid_len) return error.InvalidConnectionId;
+    if (retry_body.len > max_retry_pseudo_body_len) return error.RetryBodyTooLong;
+    const Aes128Gcm = std.crypto.aead.aes_gcm.Aes128Gcm;
+
+    var pseudo: [1 + max_cid_len + max_retry_pseudo_body_len]u8 = undefined;
+    pseudo[0] = @intCast(original_dcid.len);
+    @memcpy(pseudo[1..][0..original_dcid.len], original_dcid);
+    @memcpy(pseudo[1 + original_dcid.len ..][0..retry_body.len], retry_body);
+    const aad = pseudo[0 .. 1 + original_dcid.len + retry_body.len];
+
+    var tag: [retry_integrity_tag_len]u8 = undefined;
+    var empty_out: [0]u8 = undefined;
+    Aes128Gcm.encrypt(&empty_out, &tag, "", aad, retry_integrity_nonce_v1, retry_integrity_key_v1);
+    return tag;
+}
 
 /// Verify the integrity tag of a parsed Retry packet against the DCID the
 /// client sent in its first Initial (RFC 9001 §5.8). `retry_packet` is the
 /// full packet including the tag.
 pub fn verifyRetryIntegrity(retry_packet: []const u8, original_dcid: []const u8) bool {
     if (retry_packet.len < retry_integrity_tag_len) return false;
-    if (original_dcid.len > max_cid_len) return false;
-    const Aes128Gcm = std.crypto.aead.aes_gcm.Aes128Gcm;
-
-    // Retry pseudo-packet: ODCID length + ODCID + Retry packet without tag.
-    var pseudo: [1 + max_cid_len + 1500]u8 = undefined;
     const body_len = retry_packet.len - retry_integrity_tag_len;
-    if (1 + original_dcid.len + body_len > pseudo.len) return false;
-    pseudo[0] = @intCast(original_dcid.len);
-    @memcpy(pseudo[1..][0..original_dcid.len], original_dcid);
-    @memcpy(pseudo[1 + original_dcid.len ..][0..body_len], retry_packet[0..body_len]);
-    const aad = pseudo[0 .. 1 + original_dcid.len + body_len];
-
-    var expected: [retry_integrity_tag_len]u8 = undefined;
-    var empty_out: [0]u8 = undefined;
-    Aes128Gcm.encrypt(&empty_out, &expected, "", aad, retry_integrity_nonce_v1, retry_integrity_key_v1);
+    const expected = computeRetryIntegrityTag(original_dcid, retry_packet[0..body_len]) catch return false;
     const received = retry_packet[body_len..][0..retry_integrity_tag_len];
     return std.crypto.timing_safe.eql([retry_integrity_tag_len]u8, expected, received.*);
+}
+
+/// Write a complete QUIC v1 Retry packet (RFC 9000 §17.2.5): first byte, the
+/// fixed `version = 1`, the Retry packet's DCID (the client's Initial SCID)
+/// and SCID (the server-chosen Retry SCID), the opaque token, and the
+/// trailing RFC 9001 §5.8 integrity tag computed over `original_dcid` and
+/// everything written before the tag. Retry packets are never coalesced and
+/// carry no packet number.
+pub fn writeRetryV1(
+    original_dcid: []const u8,
+    client_scid: []const u8,
+    retry_scid: []const u8,
+    token: []const u8,
+    out: []u8,
+) error{ BufferTooShort, InvalidConnectionId, RetryBodyTooLong }![]const u8 {
+    if (client_scid.len > max_cid_len or retry_scid.len > max_cid_len) return error.InvalidConnectionId;
+
+    var pos: usize = 0;
+    const header_len = 1 + 4 + 1 + client_scid.len + 1 + retry_scid.len + token.len;
+    if (out.len < header_len + retry_integrity_tag_len) return error.BufferTooShort;
+
+    // Long header, fixed bit set, type = 0b11 (Retry); the low 4 bits are
+    // unused for Retry and may be any value on send (RFC 9000 §17.2.5).
+    out[pos] = 0x80 | 0x40 | (0b11 << 4);
+    pos += 1;
+    std.mem.writeInt(u32, out[pos..][0..4], quic_v1, .big);
+    pos += 4;
+    // Retry DCID = the client's Initial SCID; Retry SCID = our fresh CID.
+    out[pos] = @intCast(client_scid.len);
+    pos += 1;
+    @memcpy(out[pos..][0..client_scid.len], client_scid);
+    pos += client_scid.len;
+    out[pos] = @intCast(retry_scid.len);
+    pos += 1;
+    @memcpy(out[pos..][0..retry_scid.len], retry_scid);
+    pos += retry_scid.len;
+    @memcpy(out[pos..][0..token.len], token);
+    pos += token.len;
+
+    const tag = try computeRetryIntegrityTag(original_dcid, out[0..pos]);
+    @memcpy(out[pos..][0..retry_integrity_tag_len], &tag);
+    pos += retry_integrity_tag_len;
+    return out[0..pos];
 }
 
 const testing = std.testing;
@@ -439,6 +493,53 @@ test "Retry packet parses and RFC 9001 A.4 integrity tag verifies" {
     var tampered = retry;
     tampered[tampered.len - 1] ^= 1;
     try testing.expect(!verifyRetryIntegrity(&tampered, &odcid));
+}
+
+test "writeRetryV1 round-trips through parsePacket and verifyRetryIntegrity" {
+    const odcid = [_]u8{ 0x83, 0x94, 0xc8, 0xf0, 0x3e, 0x51, 0x57, 0x08 };
+    const client_scid = [_]u8{ 0xc1, 0xc2, 0xc3, 0xc4, 0xc5, 0xc6, 0xc7, 0xc8 };
+    const retry_scid = [_]u8{ 0xa1, 0xa2, 0xa3, 0xa4 };
+    const token = "opaque-retry-token";
+
+    var buf: [128]u8 = undefined;
+    const written = try writeRetryV1(&odcid, &client_scid, &retry_scid, token, &buf);
+
+    const parsed = try parsePacket(written, 8);
+    try testing.expectEqual(PacketKind.retry, parsed.kind);
+    try testing.expectEqual(quic_v1, parsed.version);
+    try testing.expectEqualSlices(u8, &client_scid, parsed.dcid);
+    try testing.expectEqualSlices(u8, &retry_scid, parsed.scid);
+    try testing.expectEqualStrings(token, parsed.retry_token);
+    try testing.expect(verifyRetryIntegrity(written, &odcid));
+
+    // Wrong ODCID must fail integrity verification.
+    try testing.expect(!verifyRetryIntegrity(written, &client_scid));
+
+    // Any tamper to the written packet must fail integrity verification.
+    var tampered: [128]u8 = undefined;
+    @memcpy(tampered[0..written.len], written);
+    tampered[written.len - 1] ^= 0x01;
+    try testing.expect(!verifyRetryIntegrity(tampered[0..written.len], &odcid));
+}
+
+test "writeRetryV1 rejects oversized connection IDs and short output buffers" {
+    const odcid = [_]u8{0x01} ** 8;
+    const client_scid = [_]u8{0x02} ** 8;
+    const retry_scid = [_]u8{0x03} ** 8;
+    const oversized_cid = [_]u8{0xff} ** (max_cid_len + 1);
+
+    var buf: [128]u8 = undefined;
+    try testing.expectError(error.InvalidConnectionId, writeRetryV1(&odcid, &oversized_cid, &retry_scid, "tok", &buf));
+    try testing.expectError(error.InvalidConnectionId, writeRetryV1(&odcid, &client_scid, &oversized_cid, "tok", &buf));
+
+    // A zero-length CID pair still round-trips, but a too-small output buffer
+    // must fail deterministically rather than write a truncated packet.
+    var tiny: [8]u8 = undefined;
+    try testing.expectError(error.BufferTooShort, writeRetryV1(&odcid, &client_scid, &retry_scid, "tok", &tiny));
+
+    var exact: [1 + 4 + 1 + 8 + 1 + 8 + 3 + retry_integrity_tag_len]u8 = undefined;
+    const written = try writeRetryV1(&odcid, &client_scid, &retry_scid, "tok", &exact);
+    try testing.expectEqual(exact.len, written.len);
 }
 
 test "version negotiation packet exposes the version list" {
--- a/src/quic/packet.zig
+++ b/src/quic/packet.zig
@@ -334,8 +334,14 @@ pub fn writeRetryV1(
     retry_scid: []const u8,
     token: []const u8,
     out: []u8,
-) error{ BufferTooShort, InvalidConnectionId, RetryBodyTooLong }![]const u8 {
+) error{ BufferTooShort, InvalidConnectionId, RetryBodyTooLong, EmptyToken, RetryScidNotDistinct }![]const u8 {
     if (client_scid.len > max_cid_len or retry_scid.len > max_cid_len) return error.InvalidConnectionId;
+    // A Retry with no token defeats the point of Retry (there is nothing for
+    // a conforming client to echo back), and a Retry SCID equal to the
+    // client's original DCID is indistinguishable from "no Retry happened" —
+    // #387 requires a server-chosen SCID distinct from it.
+    if (token.len == 0) return error.EmptyToken;
+    if (std.mem.eql(u8, original_dcid, retry_scid)) return error.RetryScidNotDistinct;
 
     var pos: usize = 0;
     const header_len = 1 + 4 + 1 + client_scid.len + 1 + retry_scid.len + token.len;
@@ -520,6 +526,20 @@ test "writeRetryV1 round-trips through parsePacket and verifyRetryIntegrity" {
     @memcpy(tampered[0..written.len], written);
     tampered[written.len - 1] ^= 0x01;
     try testing.expect(!verifyRetryIntegrity(tampered[0..written.len], &odcid));
+}
+
+test "writeRetryV1 rejects an empty token and a Retry SCID matching the client's original DCID" {
+    const odcid = [_]u8{0x01} ** 8;
+    const client_scid = [_]u8{0x02} ** 8;
+    const retry_scid = [_]u8{0x03} ** 8;
+    var buf: [128]u8 = undefined;
+
+    // An empty token gives a conforming client nothing to echo back.
+    try testing.expectError(error.EmptyToken, writeRetryV1(&odcid, &client_scid, &retry_scid, "", &buf));
+
+    // A Retry SCID identical to the client's original DCID is indistinguishable
+    // from no Retry having happened at all — the server must choose a fresh one.
+    try testing.expectError(error.RetryScidNotDistinct, writeRetryV1(&odcid, &client_scid, &odcid, "tok", &buf));
 }
 
 test "writeRetryV1 rejects oversized connection IDs and short output buffers" {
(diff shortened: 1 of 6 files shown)


src/quic/path.zig
```diff
@@ -598,51 +686,75 @@ pub const PathManager = struct {
...
-        switch (path.change) {
+        return .{ .path = key, .change = candidate.change };
+    }
+
+    /// Promote a path already marked `.validated` by `validatePathResponse`
+    /// to the active path, lifting its anti-amplification limit. Returns null
+    /// if `key` is not a validated candidate. For a host migration the caller
+    /// must claim a fresh peer CID before calling this (RFC 9000 §9.5); if
+    /// none is available, call `recordMigrationBlockedNoPeerCid` instead and
+    /// leave the candidate `.validated` so a later retry can promote it
```
Jaruso: [P2] This promised state is not stable yet. After `validatePathResponse()` sets the candidate to `.validated`, any subsequent datagram on that non-active path goes through `onDatagram()`, whose `.validated` branch immediately changes it back to `.validating` and installs a new challenge. That defeats the exact no-peer-CID flow described here: if the peer later sends `NEW_CONNECTION_ID` on the candidate path, receiving that packet first destroys the completed validation, so the new CID cannot be used to promote without another validation RTT.

Please distinguish "validated and awaiting promotion" from a previously-active/old validated path, or have `onDatagram()` preserve this pending-promotion state. A regression should cover:

```zig
_ = manager.onDatagram(migrated, challenge1, 0);
_ = manager.validatePathResponse(migrated, challenge1, 100).?;
manager.recordMigrationBlockedNoPeerCid();

// More authenticated traffic arrives before a CID becomes available.
const next = manager.onDatagram(migrated, challenge2, 200);
// Must not erase the completed validation / require a new RTT.

const outcome = manager.promoteValidated(migrated) orelse return error.TestUnexpectedResult;
try testing.expect(outcome.reset_congestion);
```

A dedicated state such as `.validated_pending_promotion` would make the distinction explicit and prevent later Connection code from having to infer why a non-active path is `.validated`.

src/quic/path.zig
```diff
@@ -518,23 +558,71 @@ pub const PathManager = struct {
...
+    pub fn markActiveValidated(self: *PathManager) void {
+        self.paths[self.active].?.anti_amplification.markValidated();
+    }
+
+    /// Record datagram bytes authenticated on `path` against that path's own
+    /// anti-amplification ledger. The caller must only call this after packet
+    /// protection succeeds — an unauthenticated datagram must never buy
+    /// candidate-path send budget. A no-op if `path` is not tracked.
+    pub fn recordReceivedOnPath(self: *PathManager, path: PathKey, bytes: u64) void {
+        const index = self.find(path) orelse return;
```
Jaruso: [P2] The first authenticated datagram on a new candidate cannot be credited through this API: before `onDatagram()` runs the path is untracked, so this `orelse return` drops its bytes. But the #387 appendix deliberately specifies the post-AEAD order as **record authenticated bytes on the ingress path, then classify/start validation**. Reversing those calls in Slice 2 would work for an allowed candidate, but it would violate the contract we wrote specifically to remove implementation inference, and policy-blocked/untracked paths would still have no ledger if path-control responses ever need to be sent there.

I would make classification + first-byte accounting one operation, e.g.:

```zig
pub fn onAuthenticatedDatagram(
    self: *PathManager,
    key: PathKey,
    bytes: u64,
    challenge_entropy: [path_challenge_len]u8,
    now_us: u64,
) PathDecision {
    // ensure/find the path record, credit `bytes`, then decide active/probe/blocked
}
```

An equivalent split API is fine, but there needs to be a way to credit the first candidate datagram without creating migration state before AEAD succeeds. Please add a test where a fresh candidate's first authenticated 100 bytes immediately yield a 300-byte candidate-path amplification budget.

src/quic/packet.zig
```diff
@@ -286,31 +286,85 @@ const retry_integrity_key_v1 = [16]u8{
...
+/// everything written before the tag. Retry packets are never coalesced and
+/// carry no packet number.
+pub fn writeRetryV1(
+    original_dcid: []const u8,
+    client_scid: []const u8,
+    retry_scid: []const u8,
+    token: []const u8,
+    out: []u8,
+) error{ BufferTooShort, InvalidConnectionId, RetryBodyTooLong }![]const u8 {
+    if (client_scid.len > max_cid_len or retry_scid.len > max_cid_len) return error.InvalidConnectionId;
```
Jaruso, line 338: [P2] Please reject the Retry-invalid inputs here rather than allowing the new primitive to manufacture packets that a conforming client must discard. RFC 9000 §17.2.5.2 requires clients to discard a Retry with a zero-length token, and §17.2.5 requires the server's Retry SCID to differ from the Destination Connection ID in the client's Initial (our `original_dcid`). #387 also explicitly requires a distinct server-chosen Retry SCID and tests for zero/oversized fields.

Something along these lines keeps the runtime from having to duplicate wire-validity checks:

```zig
if (original_dcid.len > max_cid_len or
    client_scid.len > max_cid_len or
    retry_scid.len > max_cid_len)
    return error.InvalidConnectionId;
if (token.len == 0) return error.EmptyRetryToken;
if (std.mem.eql(u8, retry_scid, original_dcid))
    return error.InvalidConnectionId;
```

Please add tests for an empty token and `retry_scid == original_dcid` as well as the existing oversized/short-buffer coverage.

Jaruso (commented):
Review result: **changes needed before merge**.

This is a strong Slice 1 overall and it tracks the #387 pre-development appendix closely: the migration transport-parameter mapping is corrected, Retry tokens now bind the Retry SCID, the Retry writer reuses the packet-layer integrity implementation, NEW_CONNECTION_ID encoding is present, CID collision preflight is metric-neutral, and the validation/promotion split is the right architectural direction.

I found three issues that should be fixed in this slice because these APIs are explicitly the contracts the later Connection/runtime slices will consume:

1. A host path that validates but cannot promote for lack of a fresh peer CID does not actually remain promotable: the next authenticated datagram on that path changes `.validated` back to `.validating`. In the common case where the eventual NEW_CONNECTION_ID itself arrives on that new path, we unnecessarily lose the completed validation and require another PATH_CHALLENGE round trip.
2. The new authenticated-byte accounting API cannot implement the appendix's required ingress order for the first datagram on a new path: `recordReceivedOnPath()` silently drops the bytes until `onDatagram()` has already created the path. Please make first-datagram classification/accounting atomic (or otherwise ensure the path exists and is credited as part of the same post-AEAD operation) so Slice 2 does not have to reverse the specified sequence or bolt on another ledger.
3. `writeRetryV1()` currently accepts inputs that produce a Retry packet a conforming client must discard: an empty Retry token and a Retry SCID identical to the client's original DCID. #387 explicitly calls for a distinct server Retry SCID and zero-field coverage, so please reject these at the primitive boundary and test them.

Please add regression tests for the first two state-machine cases as part of the fixes. In particular, test `validate -> no peer CID -> another authenticated datagram -> later promote` and test that the very first authenticated candidate datagram contributes to that candidate's anti-amplification budget without any pre-auth path mutation.

src/quic/path.zig
```diff
@@ -563,14 +663,16 @@ pub const PathManager = struct {
 
         if (self.find(key)) |index| {
             const path = &self.paths[index].?;
+            path.anti_amplification.recordReceived(authenticated_bytes);
             switch (path.state) {
                 .validating => return .probing,
+                // Already validated but not yet promoted (e.g. blocked
+                // earlier on a fresh peer CID): keep the completed validation
+                // intact instead of discarding it for a fresh challenge.
+                .validated => return .validated_pending_promotion,
```
Jaruso: [P2] This still conflates **"validated candidate awaiting first promotion"** with **"previously-active validated path"**. `PathState` has only one `.validated` value, so every non-active validated path hits this branch.

That is observable today. Suppose path A is the initial path and B is a same-host/new-port NAT rebinding. After B validates and is promoted, A remains `.validated`, but its stored `change` is still the initialization value `.migration`. If traffic later arrives on A, this function correctly computes `change == .nat_rebinding` relative to active B, but returns here *before updating `path.change`*. A subsequent `promoteValidated(A)` therefore reports a host migration, increments `migrations`, and requests a congestion/RTT reset (and future Connection code can unnecessarily require a fresh peer CID) even though this is only a port rebind.

More generally, this lets any previously validated path bypass revalidation forever. RFC 9000 §9.3 allows skipping validation for a previously validated address only when it has been seen **recently**; `PathManager` has no recency state, so `.validated` cannot safely mean that indefinitely.

Please represent pending promotion explicitly, e.g.:

```zig
pub const PathState = enum {
    unvalidated,
    validating,
    validated_pending_promotion,
    validated,
    failed,
};
```

`validatePathResponse()` would set `.validated_pending_promotion`; `promoteValidated()` would require that state and transition it to `.validated`. A previously-active `.validated` path can then follow the existing conservative re-probe path (unless we deliberately add a bounded "recently validated" policy later).

Please add a regression roughly like:

```zig
const a = testKey(50_000);
const b = testKey(50_001); // same host, new port
var manager = PathManager.init(.full, a, true);
_ = manager.onDatagram(b, 1200, challenge1, 0);
_ = manager.validatePathResponse(b, challenge1, 10).?;
_ = manager.promoteValidated(b).?;

// A was validated historically, but is not a pending first-promotion candidate.
const decision = manager.onDatagram(a, 1200, challenge2, 20);
try testing.expect(decision == .probe); // or equivalent explicit recent-path policy
```

Keep the new no-peer-CID regression too; both behaviors need to coexist.

Jaruso (commented):
Re-review result: **one P2 blocker remains before merge**.

The follow-up commit correctly fixes two of the three previous findings:

- authenticated receive accounting is now atomic with first-path creation, so the first candidate packet contributes to that path's amplification budget without pre-auth mutation;
- `writeRetryV1()` now rejects an empty Retry token and a Retry SCID equal to the original DCID, with regression coverage.

The no-peer-CID fix is directionally right, but it does not actually distinguish the two states called out in the original review: **(a) a newly validated candidate awaiting its first promotion** versus **(b) a path that was active/validated previously and is merely non-active now**. Both are still stored as `PathState.validated`, and `onDatagram()` returns `validated_pending_promotion` for both.

That causes two problems: the stored `path.change` can be stale relative to the *current* active path, and an arbitrarily old previously-validated path can bypass a fresh validation indefinitely. RFC 9000 permits skipping validation for a previously validated address only when it has been seen recently; this implementation tracks no recency. The original implementation's re-probe behavior was conservative and deterministic.

Please make pending promotion explicit in path state (or an equivalent flag), set it only when `validatePathResponse()` succeeds without promotion, and clear it on promotion. A normal previously-active `.validated` path should not be mistaken for this state. The inline comment includes a concrete NAT-rebinding regression that exposes the stale `change` value.

CI is green on `0b55285`; I did not find additional issues in the Retry writer, token binding, CID helper, frame encoder, transport-parameter mapping, or smoke-harness changes.

src/quic/path.zig
```diff
@@ -598,51 +717,80 @@ pub const PathManager = struct {
...
-        if (!std.crypto.timing_safe.eql([path_challenge_len]u8, path.challenge, data)) {
+        if (!std.crypto.timing_safe.eql([path_challenge_len]u8, candidate.challenge, data)) {
             self.metrics.path_response_mismatches += 1;
             return null;
         }
 
-        path.state = .validated;
-        path.anti_amplification.markValidated();
-        self.active = index;
+        candidate.state = .validated_pending_promotion;
```
Jaruso, line 771: [P2] Successful path validation needs to lift this path's anti-amplification limit **here**, not only later in `promoteValidated()`. #387's per-path contract says to “lift the candidate-path limit only when that path validates,” and RFC 9000 §8.2.3 explicitly says that after a matching PATH_RESPONSE validates the peer address, the endpoint may send more than 3x the bytes received even if a second MTU validation is still needed.

Today a host migration can reach `.validated_pending_promotion`, fail to promote because no fresh peer CID is available, and then remain amplification-limited indefinitely despite already proving address ownership. That is observable through `canSendOnPath()` and can incorrectly suppress later probing/PATH_RESPONSE traffic while waiting for a CID.

Please mark the ledger validated as part of successful validation, e.g.:

```zig
candidate.state = .validated_pending_promotion;
candidate.anti_amplification.markValidated();
self.metrics.path_validations_succeeded += 1;
```

`promoteValidated()` may leave the idempotent mark in place or remove it. Add a regression that starts with a small candidate budget, validates the PATH_RESPONSE without promoting, and verifies `canSendOnPath(candidate, maxInt(u64))` is then true while the old path is still active.

src/quic/path.zig
```diff
@@ -563,14 +674,21 @@ pub const PathManager = struct {
...
-                // Fresh traffic on a previously failed/unvalidated tuple:
-                // start a new probe.
-                .failed, .unvalidated => {},
-                // A validated non-active path re-activates only through a new
-                // challenge round trip, keeping the switch deterministic.
-                .validated => {},
+                // Already validated but not yet promoted (e.g. blocked
+                // earlier on a fresh peer CID): keep the completed validation
+                // intact instead of discarding it for a fresh challenge.
+                .validated_pending_promotion => return .validated_pending_promotion,
```
Jaruso, line 688: [P2] The pending-promotion state is now preserved correctly, but its stored `path.change` can still become stale while it waits. `change` was computed relative to whatever path was active when this candidate was first probed. Since multiple candidates are allowed, another path can be promoted before this one gets a fresh peer CID; at that point NAT-vs-host classification must be relative to the **new** active path.

Concrete case:

```zig
// A = host 1, active
// B = host 2:port1; validates as `.migration`, but no peer CID -> pending
// C = host 2:port2; later validates/promotes using a newly available CID
// B is now only a NAT rebinding relative to active C
```

B still carries `.migration`. If Connection later retries B using that stale value, it can unnecessarily require another fresh peer CID; `promoteValidated(B)` will also return `reset_congestion = true` even though the actual transition C -> B is port-only.

Please make the promotion classification queryable/recomputed against `self.activePath()` immediately before the caller decides whether a fresh peer CID is required, and have `promoteValidated()` use the same current classification. For example:

```zig
pub fn promotionChange(self: *const PathManager, key: PathKey) ?AddressChange {
    const index = self.find(key) orelse return null;
    if (self.paths[index].?.state != .validated_pending_promotion) return null;
    return if (key.remote.sameHost(self.activePath().key.remote))
        .nat_rebinding
    else
        .migration;
}
```

Then `promoteValidated()` should derive its metric/reset outcome from `promotionChange(key)`, not the historical `candidate.change`. Please cover the A -> pending B -> promoted C -> promote B sequence and assert B is treated as NAT rebinding.

Jaruso (commented):
Re-review result: **two P2 PathManager contract issues remain before merge**.

The third commit correctly fixes the previous stale-historical-path problem: pending first-promotion state is now explicit, a previously-active path re-probes, and the regression verifies NAT-rebinding classification is recomputed after returning to an older path.

On this pass I found two additional integration-contract problems that should be fixed in Slice 1 rather than pushed into the Connection slice:

1. `validatePathResponse()` records successful validation but does not lift that path's anti-amplification ledger until `promoteValidated()`. #387 explicitly requires the candidate-path limit to lift when the path validates, and RFC 9000 §8.2.3 says successful PATH_RESPONSE validation permits sending beyond the 3x limit even if migration/promotion has not happened. A host migration that is validated but blocked waiting for a fresh peer CID therefore remains incorrectly amplification-limited today.
2. `validated_pending_promotion` preserves the `AddressChange` computed against the active path that existed when validation started. Because `PathManager` supports multiple candidates, another path can be promoted while this one waits for a peer CID. The pending path's classification can then change relative to the new active path, but the stored `change` remains stale. That can make a path that is now only a NAT rebind continue to require a fresh peer CID and later reset congestion as a host migration. The Connection layer needs a way to get the promotion classification relative to the *current* active path immediately before deciding whether `claimForMigration()` is required.

Suggested direction: mark the candidate's amplification ledger validated inside `validatePathResponse()`, and expose a current `promotionChange(key)` / `pendingPromotion(key)` query (or equivalent) that recomputes against `activePath()` before the caller claims a peer CID and promotes. `promoteValidated()` should use the same current classification for its returned `MigrationOutcome`.

Please add regressions for both cases. Aside from these two issues, I re-scanned the Retry writer/token changes, frame encoder, CID route helper, transport-parameter mapping, PathManager timeout/exact-path validation behavior, and smoke-harness migration; I did not find another blocker there.

src/quic/path.zig
```diff
@@ -563,14 +674,21 @@ pub const PathManager = struct {
...
+                // A previously-active/validated-but-now-non-active path has
+                // no tracked recency (RFC 9000 §9.3 permits skipping
+                // validation only for a *recently* seen address), and its
+                // stored `change` may be stale relative to the current active
+                // path. Treat it the same as unvalidated/failed: start a
+                // fresh, conservative probe rather than trusting it — and
+                // silently reusing a stale `change` — indefinitely.
+                .failed, .unvalidated, .validated => {},
             }
             path.state = .validating;
```
Jaruso, line 698: [P2] Re-probing a previously-active `.validated` path currently leaves its old anti-amplification validation in force. This path was marked `anti_amplification.validated = true` when it originally validated/promoted. We now intentionally send it through a fresh PATH_CHALLENGE because we do not track whether the old validation is recent, but this branch only changes `PathState`; it never resets that ledger. As a result `canSendOnPath()` remains unlimited throughout the new validation attempt.

That is internally inconsistent with the conservative revalidation policy and with #387's per-candidate contract. RFC 9000 §9.3.1 says that until a peer's address is deemed valid, sending to it is limited; only when validation is *skipped* for a recently seen address does the endpoint not need that limit.

Please re-arm the ledger when transitioning historical `.validated -> .validating`, then credit this datagram into the fresh ledger. For example, restructure before the shared `recordReceived` so only this state resets:

```zig
const prior_state = path.state;
if (prior_state == .validated) {
    path.anti_amplification = .{};
}
path.anti_amplification.recordReceived(authenticated_bytes);
```

Do **not** reset `.validated_pending_promotion`; that path just completed address validation and should remain unlimited while it waits for a CID.

Regression: A validates/is active, B validates/promotes, then A returns and is re-probed with 100 authenticated bytes. Before the new PATH_RESPONSE, `canSendOnPath(A, 300)` should be true and `canSendOnPath(A, 301)` false; after validation it should become unlimited again.

src/quic/path.zig
```diff
@@ -598,51 +717,107 @@ pub const PathManager = struct {
...
         }
 
-        path.state = .validated;
-        path.anti_amplification.markValidated();
-        self.active = index;
+        candidate.state = .validated_pending_promotion;
+        candidate.anti_amplification.markValidated();
         self.metrics.path_validations_succeeded += 1;
-        switch (path.change) {
+        return .{ .path = key, .change = candidate.change };
```
Jaruso, line 775: [P2] `ValidatedCandidate.change` can still be stale at the instant this function returns. `candidate.change` was captured when the challenge started, but another candidate may promote while this validation is in flight.

Concrete sequence:

```text
A active (host 1)
B starts validation (host 2:port1) -> stored change=migration
C starts validation (host 2:port2) -> stored change=migration
C validates/promotes -> C is now active
B's PATH_RESPONSE arrives
```

At this line B is now a NAT rebinding relative to active C, but the returned `ValidatedCandidate.change` is still `.migration`. A Connection implementation following #387's validation sequence can branch on that value immediately, unnecessarily try to claim a fresh peer CID, and potentially block a rebinding that could promote directly. `promotionChange()` only fixes this if every caller knows to ignore the `change` that `validatePathResponse()` just returned.

Please make this API impossible to misuse: either recompute the current change here and update both `candidate.change` and the returned value, or remove `change` from `ValidatedCandidate` and require callers to call `promotionChange(validated.path)` before CID selection. Keep `promotionChange()` for the case where promotion is delayed and the active path changes *after* validation too.

Add the variant where C promotes **before B's PATH_RESPONSE arrives**; the returned/queried classification for B must be `nat_rebinding`.

Jaruso (commented):
Re-review result: **two P2 PathManager contract issues remain before merge**.

The latest commit correctly fixes the previous two findings: successful PATH_RESPONSE validation now lifts that candidate's anti-amplification ledger immediately, and `promotionChange()`/`promoteValidated()` recompute NAT-rebinding-vs-host-migration classification against the current active path. CI is also fully green on `7dee95f`.

The state-transition audit exposed two remaining edge cases in the API that Slice 2 is about to consume:

1. When a previously-active `.validated` path is deliberately re-probed because we do not track recency, its `AntiAmplification.validated` flag remains true from the old validation. That means `canSendOnPath()` is unlimited during the new validation attempt, even though this code explicitly chose not to skip validation. Reset/re-arm that path's anti-amplification ledger when `.validated -> .validating`, then credit the current authenticated datagram. Keep `.validated_pending_promotion` untouched because that path just completed a valid challenge.
2. `validatePathResponse()` still returns `ValidatedCandidate.change` from the original probe, so it can already be stale *when the PATH_RESPONSE arrives* if another candidate became active while this challenge was outstanding. The later `promotionChange()` helper makes delayed promotion safe, but #387's Connection sequence branches immediately after validation to decide whether a fresh peer CID is required. Please either recompute/update `change` before returning the `ValidatedCandidate`, or remove `change` from that return type and require callers to query `promotionChange()`.

I re-scanned the remaining Slice 1 changes (Retry writer/token binding, NEW_CONNECTION_ID encoder, CID routing helper, transport-parameter mapping, timer/exact-path response handling, and smoke-harness call sites) and did not find additional blockers.

Jaruso (commented):
Re-review result: **merge-ready from my side**.

I rechecked the latest `cff6d515` delta against the two remaining PathManager findings and then re-ran the Slice 1 contract pass against #387.

Both prior blockers are fixed correctly:

- historical `.validated` paths now re-arm their anti-amplification ledger before a conservative re-probe, while `.validated_pending_promotion` keeps its just-completed validation intact;
- `validatePathResponse()` now recomputes and stores `AddressChange` against the current active path before returning, while `promotionChange()` still recomputes again immediately before delayed CID selection/promotion.

The new regressions cover both timing windows, including the case where another candidate promotes while a challenge is outstanding. The earlier fixes also remain intact: pending-promotion state survives further authenticated traffic, first-datagram receive accounting is atomic with path creation, validation lifts the candidate amplification limit, delayed promotion recomputes classification, and Retry writer validity checks are enforced.

I did another pass over the rest of the declared Slice 1 surface (`config.zig`, Retry token/writer, NEW_CONNECTION_ID encoding, CID route collision preflight, PathManager timeout/accounting/validation semantics, and the smoke-harness integration) and did not find another correctness issue that should block this PR. CI is fully green on this head.

**Non-blocking cleanup:** the PR description still says the receive-accounting API includes `recordReceivedOnPath`, but that helper was intentionally removed when accounting was folded atomically into `onDatagram(..., authenticated_bytes, ...)`. Updating that summary would keep the PR record aligned with the final API, but I would not hold the merge for it.

Because this review is being submitted from the PR author's connected GitHub account, GitHub does not allow a meaningful self-approval; treat this COMMENT review as my final technical approval of Slice 1.

PR #516, opened 2026-07-26
quic: integrate PathManager into Connection and H3 runtime (#387 Slice 2)

Jaruso:
**What changed and why**

Slice 2 of #387's production-integration plan: wires the `PathManager` primitives merged in PR #514 into the native QUIC `Connection` driver and the production HTTP/3 runtime, so authenticated path changes (NAT rebinding and host migration) are tracked safely and outbound datagrams carry an explicit destination — instead of a path-blind egress wrapper that silently chose a peer address.

- `Connection.Options` gains `initial_path`/`initial_address_validated`; `Connection` now owns a `PathManager` instead of a single `AntiAmplification` ledger and untagged PATH_CHALLENGE/PATH_RESPONSE state (`startPathValidation`/`consumePathValidated` removed in favor of the PathManager-driven flow).
- `ingestOnPath()` / `pollTransmitOnPath()` replace `ingest()` / `pollTransmit()`: path state (amplification credit, candidate creation/validation) only ever mutates *after* AEAD open succeeds, and every transmit carries its exact destination `PathKey`.
- Candidate-path egress is isolated from ordinary active-path content — PATH_CHALLENGE/PATH_RESPONSE/PING/PADDING only, gated by that path's own anti-amplification budget — until promotion; ordinary output stays on the active path throughout.
- NAT rebinding promotes immediately, preserving RTT/congestion state. Host migration claims a fresh peer CID before promoting and resets recovery exactly once; with no fresh CID available, the old path stays active and the candidate remains pending (a later NEW_CONNECTION_ID retries promotion without another challenge round trip).
- The validation deadline folds into `nextTimeoutUs()`/`onTimeout()`, using `max(default_validation_timeout_us, 3x application-space PTO)` per the #387 contract.
- `path_validated` is now a structured event (`PathKey` + `AddressChange`); `Connection.pathMetrics()`/`activePathKey()` expose state for later #387 slices to publish as operator metrics.
- `http3_runtime.zig` resolves the real bound local UDP address (`getsockname`, including `quic_port = 0`), builds `PathKey`s from the actual source/destination addresses, sends every transmit to its returned destination, and decouples half-open admission accounting from mutable peer routing via an immutable `admission_source_ip` on `ConnEntry`.

Excludes Retry accept/config/TLS-CID binding and local CID issuance, which remain later #387 slices. No file owned by PR #513 (`codex/issue-510-native-tcp-0rtt`) was touched — verified against its changed-file list before starting.

**Related issues**
Closes #515
Refs #387

**Tests**
- `src/quic/path.zig`: added `PathManager.remainingOnPath()`, `stateOf()`, `pendingPromotionCandidate()` (small additions needed for the Connection-level wiring) with unit tests.
- `src/quic/connection.zig`: new `MigrationPair` test harness (configurable migration policy) plus a dedicated battery of tests covering the issue's 15 required scenarios — same-active-path no-op egress, spoofed/undecryptable traffic creating no path state, atomic per-candidate ledger crediting, NAT rebinding promote-without-reset, host migration promote-with-reset, migration blocked on no peer CID (and later retry), wrong-path PATH_RESPONSE rejection, exact-path PATH_RESPONSE correlation, candidate-path egress isolation, per-path amplification isolation, and validation-deadline/timer integration.
- `src/http/http3_runtime.zig`: added tests for port-0 local-address resolution via `getsockname` and the `sockaddr_in` <-> `quic.udp.Address` conversion round-trip.
- `tests/quic_h3_udp_smoke.zig`, `tests/quic_h3_e2e.zig`, `tests/h3_interop_tool.zig`: migrated all existing scenarios (real-UDP smoke, deterministic loss/reorder/duplication harness, and the out-of-process interop tool) to the new path-aware API so they keep exercising real handshakes/data exchange over the new plumbing.

**Environment note:** this session's container has no local Zig toolchain and network access to ziglang.org is blocked by this workspace's egress policy, so I could not run `zig fmt`/`zig build test`/`zig build test-quic` locally. Every change was written and re-reviewed carefully against the existing code's exact patterns (packet building, sealing, path/CID APIs), and brace/paren balance was verified mechanically, but this PR has **not** been compiled or executed locally — please treat CI as the first real build/test signal and let me know if it turns up issues.

**Security impact**
Yes — this changes QUIC path/migration handling and anti-amplification accounting, both explicit RFC 9000 §8/§9 security mechanisms. Reviewed: path/amplification state changes only occur after AEAD packet-protection success (never for unauthenticated traffic); PATH_RESPONSE validates only on the exact ingress path it was challenged on; candidate-path anti-amplification budgets are isolated per path so a candidate can never borrow the active path's budget or vice versa; host migration requires a fresh, never-reused peer CID before promoting (RFC 9000 §9.5 linkability); half-open connection admission accounting now keys off an immutable per-connection source IP rather than a peer address that can change, so it can't be confused by a later path change.

**Performance impact**
Minor. `ingestOnPath`/`pollTransmitOnPath` add one `PathManager` lookup/update per packet and one extra egress-priority check (candidate-path content before ordinary content) per `pollTransmitOnPath` call; both are bounded (`max_paths = 4`) and only meaningfully exercised once a path actually changes. No change to the hot path for ordinary same-path traffic beyond that fixed-size check.

**Checklist**
- [ ] `zig fmt --check build.zig src/ tests/` passes — not run locally (see Environment note); please verify in CI
- [ ] `zig build test --summary all` green — not run locally; please verify in CI
- [ ] `zig build test-integration` green — not run locally; please verify in CI
- [ ] docs/CODE_REVIEW_CHECKLIST.md completed
- [ ] README / CHANGELOG updated if operator-visible behavior changed — no operator-visible config surface changed in this slice (migration policy wiring remains a later slice)

🤖 Generated with [Claude Code](https://claude.ai/code)

---
_Generated by [Claude Code](https://claude.ai/code/session_018Ln9iWBFKwL5ohEH7pYApg)_

### DIFF
--- a/src/quic/connection.zig
+++ b/src/quic/connection.zig
@@ -33,6 +33,7 @@ const recovery = @import("recovery.zig");
 const quic_stream = @import("stream.zig");
 const quic_cid = @import("cid.zig");
 const quic_path = @import("path.zig");
+const quic_udp = @import("udp.zig");
 
 const EncryptionLevel = tls_adapter.EncryptionLevel;
 const PacketNumberSpace = recovery.PacketNumberSpace;
@@ -107,9 +108,24 @@ pub const Event = union(enum) {
     close_sent: struct { error_code: u64 },
     close_received: struct { error_code: u64, is_application: bool },
     idle_timeout,
-    /// A PATH_RESPONSE echoed the outstanding PATH_CHALLENGE: the path is
-    /// validated (RFC 9000 §8.2.3).
-    path_validated: [frame.path_data_len]u8,
+    /// A candidate path validated and was promoted to active (RFC 9000
+    /// §8.2.3/§9.5): a NAT rebinding or a host migration, per `change`.
+    path_validated: PathValidatedEvent,
+};
+
+pub const PathValidatedEvent = struct {
+    path: quic_path.PathKey,
+    change: quic_path.AddressChange,
+};
+
+/// A datagram `pollTransmitOnPath` produced, and the exact destination it
+/// must be sent to. Never the connection's "current" address implicitly:
+/// candidate-path probes and responses target a path other than the active
+/// one, so the caller must always send to `path.remote`, never a fixed peer
+/// address cached elsewhere.
+pub const Transmit = struct {
+    bytes: []const u8,
+    path: quic_path.PathKey,
 };
 
 pub const DropReason = enum {
@@ -167,6 +183,16 @@ pub const Options = struct {
     /// this backend cannot chain-validate. Mirrors
     /// `Handshake.allow_unverified_certificate`.
     allow_unverified_certificate: bool = false,
+    /// The (local, remote) UDP tuple the handshake is conducted on — the
+    /// embedder resolves its own bound local address (`getsockname`
+    /// semantics) and the datagram's source before constructing this.
+    initial_path: quic_path.PathKey,
+    /// Whether `initial_path`'s address is already validated (e.g. a
+    /// Retry-validated server path). A client's own initial path is always
+    /// validated regardless of this value (RFC 9000 §8.1); a non-Retry
+    /// server's initial path stays amplification-limited until the
+    /// handshake completes unless this is true.
+    initial_address_validated: bool = false,
 };
 
 // ---------------------------------------------------------------------------
@@ -524,7 +550,11 @@ const SentRecord = struct {
     carried_stop_sending: ?quic_stream.StopSendingFrame = null,
     /// PATH_CHALLENGE re-arms on loss; PATH_RESPONSE does not (the peer
     /// re-challenges, RFC 9000 §8.2.2). Both force datagram expansion.
-    carried_path_challenge: bool = false,
+    /// Non-null when this record carried a PATH_CHALLENGE for the given
+    /// candidate path — re-arms that candidate's `needs_send` on loss
+    /// (RFC 9000 §8.2.2). PATH_RESPONSE does not re-arm on loss (the peer
+    /// re-challenges instead), so it needs no requeue-identifying field.
+    carried_path_challenge_path: ?quic_path.PathKey = null,
     carried_path_response: bool = false,
     carried_ack_largest: ?u64 = null,
 
@@ -533,6 +563,24 @@ const SentRecord = struct {
         range: Range,
         fin: bool,
     };
+};
+
+/// A candidate path (RFC 9000 §9.3/§9.5) we are validating: the
+/// `PathManager`-issued PATH_CHALLENGE payload and whether the packet
+/// carrying it still needs to go out (initially, and again after loss —
+/// PATH_CHALLENGE re-arms on loss, RFC 9000 §8.2.2, unlike PATH_RESPONSE).
+const CandidateChallenge = struct {
+    path: quic_path.PathKey,
+    data: [quic_path.path_challenge_len]u8,
+    needs_send: bool = true,
+};
+
+/// A PATH_RESPONSE queued to echo a received PATH_CHALLENGE back on the
+/// exact path it arrived on (RFC 9000 §8.2.2) — not necessarily the active
+/// path, since the challenge may have arrived on a peer address probing us.
+const PendingPathResponse = struct {
+    path: quic_path.PathKey,
+    data: [quic_path.path_challenge_len]u8,
 };
 
 // ---------------------------------------------------------------------------
@@ -553,7 +601,10 @@ pub const Connection = struct {
     tls: tls_handshake.TlsBackend,
 
     recovery: recovery.RecoveryController = .{},
-    amp: quic_path.AntiAmplification = .{},
+    /// Authenticated path state (#251/#515): the active path's identity and
+    /// amplification ledger, plus any candidate paths being validated. Set in
+    /// `init()`; never default-constructed since it needs the handshake path.
+    paths: quic_path.PathManager = undefined,
 
     local_cid: config.CidValue,
     peer_cid: config.CidValue,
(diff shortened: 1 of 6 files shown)


src/quic/connection.zig
```diff
@@ -956,7 +1026,7 @@ pub const Connection = struct {
         }
         self.initial_packet_processed = true;
         if (level == .handshake) {
-            if (self.role == .server) self.amp.markValidated();
+            if (self.role == .server) self.paths.markActiveValidated();
             if (!self.processed_handshake_packet) {
                 self.processed_handshake_packet = true;
```
Jaruso: **Blocking: this validates the wrong anti-amplification ledger when a Handshake packet arrives on a different tuple.**

At this point AEAD has succeeded, but `ingress_path` has not been classified/credited yet. `markActiveValidated()` therefore always lifts the *currently active* path's limit, even if this authenticated Handshake packet arrived on a NAT-rebound/new path. That breaks #515's core invariant that candidate traffic cannot buy budget for the active path (and vice versa).

A rebinding during the handshake makes the failure concrete: old path A is active and still unvalidated; a valid Handshake packet arrives on B; this line makes A unlimited, then `onDatagram(B, ...)` creates/credits B separately. The packet from B has just validated A's send budget.

Please bind handshake-derived address validation to the authenticated ingress path rather than implicitly to `active`. One shape would be to add a PathManager operation that marks a *specific tracked path* validated after `onDatagram` has classified it, e.g.:

```zig
const decision = self.paths.onDatagram(ingress_path, bytes.len, challenge_entropy, now_us);
if (self.role == .server and level == .handshake) {
    self.paths.markValidatedOnPath(ingress_path);
}
```

If the intended rule is that receiving Handshake only validates the initial/active address, then a Handshake packet from a different tuple must not be allowed to lift the old path at all. Either way, `markActiveValidated()` before path classification is unsafe here.

Please add a regression test with an unvalidated server initial path, deliver an authenticated Handshake packet on a second tuple, and assert the original active path's remaining amplification allowance is unchanged/unlimited only when *that path* itself was validated.

src/quic/connection.zig
```diff
@@ -965,6 +1035,19 @@ pub const Connection = struct {
...
+        // validation. This never moves the active path or an outbound
+        // destination by itself — only `tryPromote` (driven by a validated
+        // PATH_RESPONSE) does that. A new validation attempt uses the #387
+        // contract for its timeout: max(default, 3x application-space PTO).
+        self.paths.validation_timeout_us = @max(quic_path.default_validation_timeout_us, 3 * self.recovery.rtt.ptoDuration(.application));
+        switch (self.paths.onDatagram(ingress_path, bytes.len, challenge_entropy, now_us)) {
+            .probe => |data| self.queueCandidateChallenge(ingress_path, data),
+            .on_active_path, .probing, .validated_pending_promotion, .blocked => {},
+        }
+
```
Jaruso, line 1056: **Blocking: `.blocked` alternate paths cannot send the required PATH_RESPONSE.**

`PathManager.onDatagram()` returns `.blocked` *before it creates/credits a path record* when migration policy denies the address change. We still parse the authenticated packet below, and `PATH_CHALLENGE` correctly queues `PendingPathResponse{ .path = ingress_path, ... }`. But `pollCandidateTransmit()` later calls `buildCandidatePacket()`, whose first gate is `remainingOnPath(path)`. For an untracked blocked path that is always `0`, so the queued PATH_RESPONSE can never be emitted.

That violates #515's requirement that a received PATH_CHALLENGE queue/send PATH_RESPONSE on the exact ingress PathKey, and it couples 'may migrate/promote this path' to 'may answer path validation control on this path'. Those are separate decisions.

Please preserve per-path anti-amplification accounting even for a policy-blocked path, while keeping it ineligible for challenge initiation/promotion. For example, PathManager could create/refresh a bounded `.unvalidated`/blocked record and credit only the authenticated bytes before returning `.blocked`:

```zig
if (!allowed) {
    const index = self.find(key) orelse self.claimSlot();
    if (self.paths[index] == null) {
        self.paths[index] = .{ .key = key, .state = .unvalidated, .change = change };
    }
    self.paths[index].?.anti_amplification.recordReceived(authenticated_bytes);
    self.metrics.migrations_blocked += 1;
    return .blocked;
}
```

Then PATH_RESPONSE can be sent on that exact tuple under its own 3x budget without making it promotable.

Please add a regression test under `.disabled` (and ideally `.nat_rebinding_only` with a host change): deliver an authenticated application packet containing PATH_CHALLENGE on the alternate path and assert `pollTransmitOnPath()` returns a PATH_RESPONSE targeted to that path, while `activePathKey()` remains unchanged and no local PATH_CHALLENGE/promotion occurs.

Jaruso (commented):
I reviewed this against #515's Slice 2 acceptance criteria, including the PathManager integration, exact-path control handling, per-path anti-amplification, CID-gated host migration, timeout integration, runtime routing/admission accounting, and the updated smoke/e2e call sites.

The overall direction is good and the footprint stays within the intended QUIC/H3 slice, but I found two correctness issues that violate the per-path security invariants in #515. I would treat both as blocking before merge.

Also, the current PR head is not green: Format/Lint/ReleaseSafe build pass, but unit-test jobs are failing across Linux/macOS/ARM and the appliance-profile test jobs fail at their unit-test step. The PR description explicitly notes the branch was not compiled/tested locally, so CI needs to be green before this is merge-ready.

Please add regression coverage for both inline findings. In particular, I would add (1) a server-handshake test where an authenticated Handshake packet is delivered on a different tuple and assert the old active path remains amplification-limited, and (2) a disabled/full-policy matrix test proving an authenticated PATH_CHALLENGE always gets a PATH_RESPONSE on its ingress path without granting that path migration/promotion rights.

src/quic/connection.zig
Jaruso: Fixed in c18e4ac, essentially as suggested: added `PathManager.markValidatedOnPath(path)` and moved the handshake-validation call to run after `onDatagram(ingress_path, ...)`, using `ingress_path` instead of the implicit active path.

One refinement: `afterHandshakeProgress` (the "handshake completion validates the client's address" call in `applyFrame`'s `.crypto` case) had the same bug, so I fixed that call site too. It's invoked from two places — one with a real `ingress_path` (the CRYPTO frame handler) and one without (`driveAuthentication`, resuming a previously-parked async operation with no associated datagram). I gave it an `?quic_path.PathKey` parameter: `markValidatedOnPath(path)` when a path is known, `markActiveValidated()` as a fallback only when there's genuinely no better path to attribute it to.

Added the requested regression test (`connection: a Handshake packet arriving on a different tuple validates only that tuple, not the old active path`): an unvalidated server initial path, the client's Initial delivered normally, everything after (including the Handshake-level Finished) delivered on a distinct alternate tuple. Asserts the server's handshake completes, the active path never changes, the *original* active path's amplification limit stays exactly as unvalidated as it started, and the alternate tuple is the one that ends up validated.

Also added a `path.zig`-level unit test for `markValidatedOnPath` itself (lifts only the named path, not a different tracked one).

---

src/quic/connection.zig
Jaruso: Fixed in c18e4ac, essentially as suggested: `PathManager.onDatagram()`'s `!allowed` branch now creates/refreshes a bounded `.unvalidated` record for the blocked tuple (via `find()` or `claimSlot()`) and credits `authenticated_bytes` to it before returning `.blocked`, so it never re-triggers a challenge or becomes promotable (both require `.validating`/`.validated_pending_promotion` first, which a blocked tuple never reaches) but does get its own anti-amplification budget, so `buildCandidatePacket` can answer a queued PATH_RESPONSE for it.

Added the requested regression test (`connection: a policy-blocked server still answers a PATH_CHALLENGE on its exact ingress path without migrating`): server on `.disabled`, client on `.full` so it can generate a genuine wire PATH_CHALLENGE (by being told the server spoke from a new address), that challenge delivered to the server as if from a brand-new client tuple. Asserts the server never sends its own challenge, `migrations_blocked` increments with `migrations`/`nat_rebindings` staying zero, `activePathKey()` never changes, and `pollTransmitOnPath()` still returns a PATH_RESPONSE addressed to that exact blocked tuple.

Also added a `path.zig`-level unit test (`a policy-blocked tuple still gets its own anti-amplification ledger, never a challenge`) proving the ledger accumulates correctly across repeated blocked traffic and that `validatePathResponse` still rejects it (state stays `.unvalidated`, never `.validating`).

---

PR #517, opened 2026-07-26
tls: add resumption closeout fuzz coverage

Jaruso:
## Summary
- add a fuzz entrypoint and seed corpus for stateless ticket identity parsing/resolution
- generate authenticated protected-ticket fuzz seeds for AES-128-GCM, AES-256-GCM, and ChaCha20-Poly1305 so the resolver reaches AEAD open, `session.decode`, timestamp checks, and successful `out.moveFrom`
- assert full recovered-state equality on resolver miss/error with a non-default sentinel output, including allocation-failure error paths
- add a fuzz entrypoint and seed corpus for persistent ticket-key snapshot JSON parsing and candidate validation

## Tests
- `zig fmt src/tls/ticket_protection.zig src/tls/ticket_key_snapshot.zig`
- `git diff --check`
- `zig build test-tls --summary all --error-style verbose` — 782/782 passed
- `zig build test --summary all --error-style verbose` — 2749/2750 passed, 1 skipped
- `zig build test-integration-resumption-interop --summary all --error-style verbose` — target built successfully; 0/9 passed, 9 skipped in this environment

Refs #369

### DIFF
--- a/src/tls/ticket_protection.zig
+++ b/src/tls/ticket_protection.zig
@@ -1088,6 +1088,27 @@ fn sampleServerState(allocator: std.mem.Allocator) !session.ServerRecoverableSta
     return state;
 }
 
+fn sentinelServerState(allocator: std.mem.Allocator) !session.ServerRecoverableState {
+    var common: session.ResumableSessionCommon = .{};
+    try common.init(allocator, session.Limits.default, .{
+        .cipher_suite = .tls_chacha20_poly1305_sha256,
+        .resumption_psk = &([_]u8{0xcd} ** 32),
+        .server_name = "sentinel.example.test",
+        .application_protocol = "http/1.1",
+        .auth_binding = session.AuthBinding.fromLeafCertificateDer("sentinel-leaf"),
+        .issued_at_unix_ms = 777,
+        .lifetime_seconds = 123,
+        .early_data = .{ .early_data_capable = 99 },
+        .transport_compat = .{ .format_id = 7, .format_version = 2, .bytes = "sentinel-transport" },
+        .application_compat = .{ .format_id = 8, .format_version = 3, .bytes = "sentinel-application" },
+        .early_data_transport_compat = .{ .format_id = 9, .format_version = 4, .bytes = "sentinel-early-transport" },
+        .early_data_application_compat = .{ .format_id = 10, .format_version = 5, .bytes = "sentinel-early-application" },
+    });
+    var state: session.ServerRecoverableState = .{};
+    state.init(&common, 0x55667788);
+    return state;
+}
+
 const TestObserver = struct {
     events: std.ArrayList(Event) = .empty,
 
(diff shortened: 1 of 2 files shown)


src/tls/ticket_protection.zig
```diff
@@ -1188,6 +1188,30 @@ fn expectRoundTrip(
...
+        &keyring,
+        keyId(0xf1),
+        .aes_128_gcm,
+        .{ .prefix = .{ 0xf1, 0, 0, 0 }, .start = 0, .end_exclusive = 8 },
+        0x11,
+    ) catch return;
+
+    var protector = Protector{ .provider = testProvider(), .keyring = &keyring, .limits = limits };
+    var out: session.ServerRecoverableState = .{};
+    defer out.deinit();
```
Jaruso, line 1258: The fuzz target silently returns on every `resolve` error, so it never checks the property advertised by this test name (and required by #494) that the output stays untouched on **error** as well as on a normal miss. Please seed `out` with a non-default sentinel state, snapshot/clone it, then assert full state equivalence after both `false` and error outcomes. In particular, checking only `ticket_age_add` and `resumption_psk.len` would miss mutation of cipher suite, SNI, ALPN, auth binding, timestamps, early-data policy, or compatibility blobs.

For example, structure this roughly as:
```zig
var out = try sentinelServerState(allocator);
defer out.deinit();
var expected: session.ServerRecoverableState = .{};
try out.cloneInto(allocator, &expected);
defer expected.deinit();

const result = protector.resolve(allocator, input, 2_000, &out);
if (result) |found| {
    if (!found) try expectServerStateEqual(&expected, &out);
} else |_| {
    try expectServerStateEqual(&expected, &out);
}
```
Then add allocation-failure injection separately so the error path is actually reachable.

src/tls/ticket_protection.zig
```diff
@@ -1309,6 +1333,23 @@ test "parseEnvelope validates public structure without allocation" {
...
 }
 
+test "fuzz: ticket identity parsing and miss resolution never panic or mutate output" {
+    try testing.fuzz({}, fuzzTicketIdentityInput, .{ .corpus = &.{
+        "",
+        "TDTK",
+        "TDTK\x01\x01\x00\x00",
+        &([_]u8{0} ** (fixed_header_len + tag_len)),
+        &([_]u8{ 0x54, 0x44, 0x54, 0x4b, 0x01, 0x01, 0x00, 0x00 } ++ [_]u8{0x00} ** (fixed_header_len + tag_len)),
+        &([_]u8{0xff} ** 256),
```
Jaruso, line 1424: None of these seeds is a valid authenticated ticket for the installed `0xf1` / `0x11` key. Once an input gets past `parseEnvelope`, `Protector.resolve` therefore stops at unknown-key/authentication failure; coverage-guided mutation cannot realistically synthesize a valid AEAD tag. That means the new target does **not** fuzz authenticated open + `session.decode`, despite being described as a resolution fuzz target.

Please generate/check in at least one valid protected-ticket seed per supported AEAD (the existing `seal`/`buildAuthenticatedEnvelope` helpers can produce deterministic fixtures), then mutate those seeds so the fuzzer can reach authenticated plaintext decoding, timestamp checks, and `out.moveFrom`. That is also the explicit seed requirement in #494.

Jaruso (commented):
Blocking findings from review:

### 1. This is not the final #369 slice described by the current issue

The current #369 source of truth was updated **after PR #513** specifically because #510 and #512 are now complete. It makes the remaining work mandatory: external native TCP/H1 0-RTT accepted/unsafe-425/replay-fallback cases; persistent restart/startup quarantine; persistent ticket survival and N→N+1 live rotation; failed-reload atomicity and certificate-binding change; concurrent multi-process nonce-safety soak; the documented process-local replay limitation; application-level resumed H2; QUIC/H3 resumption plus the production H3 0-RTT capability check; the real cipher-mismatch case; and reconciliation of `docs/RESUMPTION_TEST_PLAN.md`.

PR #517 changes only `ticket_protection.zig` and `ticket_key_snapshot.zig`. It adds no external/process/interop/soak coverage and does not update the test-plan document. The document in this PR still says #510 and #512 are blockers and explicitly says #369 remains incomplete, which is stale after #513. So merging this as the “final slice of #369” would leave most of #369’s current acceptance criteria unimplemented and the authoritative coverage matrix knowingly incorrect.

At minimum, the final #369 PR needs to implement or explicitly separately-own every remaining row in the issue and update the matrix accordingly. The expected shape is roughly:

```text
tests/integration.zig
  interop.openssl.h1.early.accepted
  interop.openssl.h1.early.unsafe_425
  interop.openssl.h1.early.replay_fallback
  restart.persistent.early_startup_quarantine
  restart.persistent.ticket_survives
  rotation.persistent.n_to_n_plus_1
  rotation.persistent.failed_reload_keeps_old_state
  rotation.persistent.certificate_binding_change
  process-local replay A/A/B/B scenario
  application-level resumed H2 request/response
  resolved cipher mismatch

scripts/interop/run-interop.sh + tests/h3_interop_tool.zig
  external H3 resumption request/response
  H3 0-RTT capability regression (accepted matrix only if production actually enables it)

.github/workflows/resumption-soak.yml
  bounded PR smoke + heavier shared-persistent-key rotation/nonce soak

docs/RESUMPTION_TEST_PLAN.md
  final #508/#509/#511/#513/current coverage map; remove stale #510/#512 blocker claims
```

### 2. The new ticket fuzz target does not currently reach the security-sensitive half of `Protector.resolve`

I left two inline comments with concrete fixes. The corpus contains no authenticated ticket, so mutations cannot get through AEAD authentication into `session.decode`; and error outcomes are discarded while the “unmodified output” check covers only two fields on ordinary misses. This is especially notable because fuzzing is separately owned by #494, whose requirements explicitly call for valid protected-ticket seeds and unchanged output on miss/error.

CI is green at head (`79020eeb`), so this is not a build failure; it is an acceptance/scope and assurance-depth failure. Once the actual remaining #369 matrix is present and these fuzz assertions are made meaningful (or the fuzz work is moved/scoped to #494), I’d re-review the resulting final slice.

Jaruso (commented):
Second review on head `62bad634748bff12c250dad2ea28659f3456f821`.

The two code-level findings from the previous pass are addressed:

- the protected-ticket fuzz corpus now includes valid deterministic AES-128-GCM, AES-256-GCM, and ChaCha20-Poly1305 tickets, so `Protector.resolve` can cross AEAD authentication into `session.decode`, timestamp validation, and successful state recovery;
- resolver miss/error paths now use a fully populated sentinel output and compare the complete recoverable state, with explicit failing-allocator coverage for both plaintext allocation and decode allocation failures.

I did not find another defect in those new fuzz changes on this pass, and CI is green.

However, I still cannot consider this the final #369 slice. The current #369 source of truth explicitly says that after #513 the remaining mandatory work is production/external assurance: external native H1 0-RTT accepted/425/replay fallback; persistent restart/startup-quarantine and N→N+1 SIGHUP rotation; failed-reload atomicity; certificate-binding change; concurrent nonce-safety soak; process-local two-process replay semantics; application-level resumed H2; H3 resumption plus the QUIC 0-RTT capability/fail-closed proof; the real cipher-mismatch case; and reconciliation of `docs/RESUMPTION_TEST_PLAN.md` so #510/#512 are no longer listed as blockers.

This PR still changes only `ticket_protection.zig` and `ticket_key_snapshot.zig`, i.e. fuzz/property coverage that #369 itself identifies as separately owned by #494 rather than as a blocker for its external/process tests. None of the newly mandatory #369 integration/documentation rows are implemented here.

So: the fuzz code now looks good, but the PR is still not sufficient to close #369 or to be described as its final slice. Either this PR needs to be reframed as #494-related fuzz coverage and leave #369 open, or the actual remaining #369 integration/documentation work needs to land before #369 is closed.

PR #526, opened 2026-07-27
tls: prove resumed H2 interop dispatch

Jaruso:
## Summary
- add H2 proxy dispatch for matched proxy_pass routes in the native H2 listener path
- upgrade interop.openssl.h2.tls_resume from ALPN-only resumption to a real resumed H2 request/response against an upstream fixture
- document the cipher-mismatch closeout: external resumption-specific mismatch is unreachable under the current single-cipher native TLS profile

## Validation
- zig build test-integration-resumption-interop -Dtls-profile=appliance --summary failures
- zig build test-integration-native-tls -Dtls-profile=appliance --summary failures
- zig build test -Dtls-profile=appliance --summary failures

Refs #521
Refs #369


src/edge_gateway.zig
```diff
@@ -2537,15 +2541,21 @@ fn respondHttp2Stream(
...
-
-    status_code = 404;
-    body = "{\"error\":\"not_found\"}";
+    var content_type: []const u8 = JSON_CONTENT_TYPE;
+
+    if (try executeHttp2ProxyRoute(allocator, state, cfg, method, path, ps, correlation_id, connection_ip, &lifecycle)) |proxy_response| {
+        var response = proxy_response;
+        defer response.deinit(allocator);
+        status_code = response.status_code;
+        if (response.headerValue("content-type")) |ct| content_type = ct;
```
Jaruso: **Blocker — borrowed header lifetime.** `ct` points into `BufferedUpstreamResponse.metadata_arena`, and `defer response.deinit(allocator)` runs when this block exits. `content_type` is then used later by HPACK encoding after the arena has been destroyed.

Please keep the response alive until encoding completes or copy the value, for example:
```zig
var content_type_owned: ?[]u8 = null;
defer if (content_type_owned) |ct| allocator.free(ct);
if (response.headerValue("content-type")) |ct| {
    content_type_owned = try allocator.dupe(u8, ct);
    content_type = content_type_owned.?;
}
```
Also preserve/filter the other end-to-end upstream response headers instead of silently dropping everything except content-type.

src/edge_gateway.zig
```diff
@@ -2594,6 +2604,99 @@ fn respondHttp2Stream(
...
+    state: *GatewayState,
+    cfg: *const edge_config.EdgeConfig,
+    method: []const u8,
+    path: []const u8,
+    ps: *const Http2PendingStream,
+    correlation_id: []const u8,
+    connection_ip: []const u8,
+    lifecycle: *http.request_lifecycle.RequestLifecycle,
+) !?gp.BufferedUpstreamResponse {
+    const matched = http.location_router.matchLocation(path, cfg.location_blocks) orelse return null;
```
Jaruso: `path` is the raw H2 `:path`, which can include a query (for example `/h2-resume?x=1`). Matching it directly makes an exact `/h2-resume` location miss; prefix routing can also put the query into the suffix and then append `request.uri.query` again below.

Synthesize/parse first, then route using `request.uri.path` and append `request.uri.query` exactly once:
```zig
var request = try synthesizeHttp2Request(...);
defer request.deinit();
const matched = http.location_router.matchLocation(request.uri.path, cfg.location_blocks) orelse return null;
const suffix_path = gproxy_runtime.proxySuffixPathForLocation(request.uri.path, matched, cfg.location_blocks);
```
Please cover an exact proxy route with a query.

src/edge_gateway.zig
```diff
@@ -2594,6 +2604,99 @@ fn respondHttp2Stream(
...
+    const suffix_path = gproxy_runtime.proxySuffixPathForLocation(path, matched, cfg.location_blocks);
+    const resolved = try gp.resolveProxyTarget(allocator, cfg.upstream_base_url, target, suffix_path);
+    defer allocator.free(resolved.url);
+    var upstream_url = try gp.appendProxyQueryString(allocator, resolved.url, request.uri.query);
+    defer upstream_url.deinit(allocator);
+
+    var ctx = http.request_context.RequestContext.init(allocator, correlation_id, connection_ip);
+    ctx.lifecycle = lifecycle;
+    const forwarded_proto = if (edge_config.hasTlsFiles(cfg)) "https" else "http";
+    var proxy_response = try gproxy_runtime.executeBufferedDataPlaneProxyRequest(
```
Jaruso, line 2927: **Security blocker — this creates a privileged side path around route policy.** Calling the low-level executor directly skips checks/orchestration performed before H1/H3 upstream side effects. Most concretely, `routeHttp3Location` rejects `matched.block.auth == .required` before `.proxy_pass`; this H2 path does not, so an auth-required upstream can be reached unauthenticated over H2.

Please factor/reuse protocol-neutral policy/orchestration, or fail closed for policies H2 cannot yet enforce. Add a regression with an `auth = required` H2 `proxy_pass` route and prove the response is rejected while `upstream.requestCount() == 0`.

src/edge_gateway.zig
```diff
@@ -2594,6 +2604,99 @@ fn respondHttp2Stream(
...
+        correlation_id,
+        connection_ip,
+        forwarded_proto,
+        request.headers.get("host"),
+        null,
+        null,
+        null,
+        null,
+        false,
+        cfg.upstream_connect_timeout_ms,
```
Jaruso: This first timeout is the **attempt** timeout, but it is wired to `upstream_connect_timeout_ms`; the following argument is the connect timeout again. H1/H3 derive a per-attempt timeout from `upstream_timeout_ms` / `upstream_timeout_budget_ms` and pass the connect timeout separately. With distinct operator settings this can terminate a healthy slow response on the connect budget. Please reuse the existing attempt-budget logic.

src/edge_gateway.zig
```diff
@@ -2594,6 +2604,99 @@ fn respondHttp2Stream(
...
+        request.uri.path,
+        correlation_id,
+        null,
+        connection_ip,
+        upstream_url.value,
+        "",
+        proxy_response.statusCode(),
+        proxy_response.contentTypeOr("application/octet-stream"),
+        proxy_response.transcriptBody(),
+        &.{},
```
Jaruso: **Secret-safety blocker.** The normal H1 proxy path derives transcript redactions from the bearer token in `Authorization`; this new H2 path always passes an empty redaction set. An echoing upstream can therefore persist credentials in transcript material. Mirror the existing bearer-token redaction logic and add an H2 proxy transcript regression so #521's `secret-safe` AC is protected.

src/edge_gateway.zig
```diff
@@ -2594,6 +2604,99 @@ fn respondHttp2Stream(
...
+    path: []const u8,
+    ps: *const Http2PendingStream,
+) !http.Request {
+    var raw = std.array_list.Managed(u8).init(allocator);
+    defer raw.deinit();
+    const request_line = try std.fmt.allocPrint(allocator, "{s} {s} HTTP/1.1\r\n", .{ method, path });
+    defer allocator.free(request_line);
+    try raw.appendSlice(request_line);
+    if (ps.headers.get("host") == null) try raw.appendSlice("Host: tardigrade.test\r\n");
+    for (ps.headers.items.items) |header| {
```
Jaruso, line 3150: Normal H2 authority arrives in `:authority`, not necessarily a regular `host` header. The frame decoder currently discards pseudo-headers other than `:method`/`:path`, so this fallback rewrites ordinary H2 requests to `Host: tardigrade.test`, breaking virtual-host semantics. Retain `:authority` in `Http2PendingStream` and translate it to the incoming Host, with the same precedence as H3 (`:authority` first, `host` fallback). Add an upstream assertion for Host.

src/edge_gateway.zig
```diff
@@ -2594,6 +2604,99 @@ fn respondHttp2Stream(
...
+        if (std.mem.eql(u8, header.name, "connection")) continue;
+        const line = try std.fmt.allocPrint(allocator, "{s}: {s}\r\n", .{ header.name, header.value });
+        defer allocator.free(line);
+        try raw.appendSlice(line);
+    }
+    try raw.appendSlice("\r\n");
+    return (try http.Request.parse(allocator, raw.items, MAX_REQUEST_SIZE)).request;
+}
```
Jaruso, line 3226: `ps.body` is never added to the synthetic request. With `content-length`, `Request.parse` sees zero body bytes and fails `InvalidContentLength`; without it, DATA is silently dropped and an empty request is proxied. Please construct the request with the buffered H2 body (preferably structurally rather than round-tripping through HTTP/1 text), normalize body framing, and add a POST/PUT body round-trip regression.

tests/integration.zig
```diff
@@ -3205,14 +3300,15 @@ test "interop.openssl.alpn_mismatch" {
...
 
-// #369: `-ciphersuites` on the client only restricts what it offers, so the
-// server should have no other option but to negotiate the client's sole
-// offered suite -- but ad hoc verification against the real native listener
-// still observed the *original* cipher on the reconnect despite the client
-// restricting itself to a different one, which needs its own investigation
-// before a test can assert on it honestly. Deferred rather than shipped as
-// a test that would silently stop proving anything if that behavior is
-// itself a bug; see docs/RESUMPTION_TEST_PLAN.md.
+// #369/#521: a resumption-specific external cipher mismatch is unreachable
```
Jaruso, line 5118: The cipher conclusion here is correct, but `docs/RESUMPTION_TEST_PLAN.md` is now stale: it still says application-level resumed H2 is deferred and that the cipher-mismatch row needs investigation. Please update those canonical closeout statements in this PR so #521 does not leave contradictory assurance documentation.

Jaruso (commented):
## Review against #521 — changes required before merge

The core resumed-H2 assurance is substantially correct: the test establishes OpenSSL `Reused`, ALPN `h2`, a Tardigrade `outcome="accepted"` delta, a unique upstream response body, and exactly one upstream execution after resetting capture. The cipher-mismatch conclusion is also correct, and CI is green.

I do **not** recommend merging yet because the implementation used to obtain that application proof also enables a new production H2 `proxy_pass` path, and that path is not yet safe/correct enough.

### Blocking findings

1. **H2 `proxy_pass` bypasses route/policy enforcement and normal proxy orchestration.** In particular, an `auth = required` location can reach the upstream over H2 without the fail-closed check the H3 path performs before action execution. The direct low-level executor also skips normal retry/error/proxy-accounting orchestration. Please share/factor protocol-neutral policy/orchestration or fail closed for policies H2 cannot yet enforce. Add an H2 regression where an auth-required proxy route returns 401 (or the documented fail-closed result) and upstream execution remains zero.
2. **Definite lifetime bug:** the selected upstream `content-type` is borrowed from `BufferedUpstreamResponse.metadata_arena`, but that response is deinitialized before HPACK encoding uses the slice.
3. **Response framing can corrupt nontrivial upstream bodies.** The current single DATA write can exceed the peer's default 16 KiB max frame size. If the body exceeds the current flow-control window, only a prefix is sent while `END_STREAM` and the full `content-length` are emitted. Chunk by peer max-frame/window limits and resume after `WINDOW_UPDATE`; set `END_STREAM` only on the final chunk. Add tests crossing both 16 KiB and the flow-control window.
4. **H2 request translation is incomplete:** query-bearing `:path` values are routed incorrectly/double-appended, `:authority` is discarded and replaced with a hard-coded Host, and buffered DATA in `ps.body` is never forwarded.
5. **Transcript secret redaction is lost.** H1 deliberately supplies bearer-token redactions; this H2 path passes an empty redaction set. This conflicts directly with #521's secret-safe AC.
6. **Attempt timeout uses the connect timeout.** Reuse the normal attempt/budget timeout calculation and pass connect timeout separately.
7. **The closeout docs remain contradictory.** `docs/RESUMPTION_TEST_PLAN.md` still says application-level resumed H2 is deferred and that `interop.openssl.cipher_mismatch` still needs investigation. #521 closes those two rows; update the canonical plan here.

Please also preserve end-to-end upstream response headers with proper HTTP/2 filtering/translation. The new H2 path currently keeps only `content-type`, unlike the H3 adapter.

Because this PR adds user-visible native-H2 `proxy_pass` behavior rather than test-only assurance, it should receive an `[Unreleased]` changelog entry if that production capability remains part of the final implementation.

The cipher-mismatch closeout itself should remain as written: do **not** add a test-only second cipher or mutate protected ticket state to manufacture an external mismatch.

src/edge_gateway.zig
```diff
@@ -2536,36 +2544,67 @@ fn respondHttp2Stream(
...
+                status_code = 401;
+                body_alloc = try gp.buildApiErrorJson(allocator, "unauthorized", "Unauthorized", correlation_id);
+                body = body_alloc.?;
+                state.metricsRecordErrorCode("unauthorized");
+            },
+        }
+    }
+
+    const effective_window: i32 = @min(conn_send_window.*, stream_send_window);
+    if (body.len > @as(usize, @intCast(@max(effective_window, 0)))) {
```
Jaruso: **Blocker — valid HTTP/2 flow control is being converted into an application-level 502.** The connection send window begins at 65,535 bytes, so any otherwise-valid proxied body larger than the current credit is discarded and replaced with `h2_flow_control_blocked`. A client is allowed to replenish connection/stream credit with `WINDOW_UPDATE`; temporary lack of credit is not a gateway failure.

The new 20 KiB regression only proves DATA frames are capped below 16 KiB. It never crosses the 65,535-byte connection window, so this branch is untested.

Please keep the outbound response pending and resume it from the H2 event loop when credit arrives instead of rewriting the response, e.g. conceptually:
```zig
fn flushPendingResponse(p: *PendingResponse, conn_window: *i32, stream_window: *i32) !bool {
    while (p.offset < p.body.len) {
        const credit = @min(conn_window.*, stream_window.*);
        if (credit <= 0) return false; // park; WINDOW_UPDATE will retry
        const n = @min(@as(usize, @intCast(credit)), @min(HTTP2_MAX_FRAME_SIZE, p.body.len - p.offset));
        const end = p.offset + n;
        try writeFrame(..., if (end == p.body.len) END_STREAM else 0, ..., p.body[p.offset..end]);
        conn_window.* -= @intCast(n);
        stream_window.* -= @intCast(n);
        p.offset = end;
    }
    return true;
}
```
Then add a test with a response >65,535 bytes, have the client issue WINDOW_UPDATE, and assert the exact original body is delivered with END_STREAM only on the final DATA frame.

Jaruso (commented):
Re-review of commit `15683f0` against #521 and the prior review: **changes are still required before merge**.

A lot of the first-pass feedback is now fixed correctly. The new revision preserves `:authority`, forwards H2 request bodies, routes on parsed URI path/query, owns/copies upstream response headers safely, strips hop-by-hop response headers, redacts bearer tokens from transcripts, separates attempt/connect timeout settings, adds the missing closeout docs/changelog, and adds useful translation/auth regressions. The core #521 assurance is also strong: OpenSSL provides authoritative `Reused` evidence, ALPN is `h2`, Tardigrade's accepted-resumption counter is checked independently, and the resumed request reaches the upstream exactly once with a unique response body.

Two code blockers remain:

1. **The original policy/orchestration bypass is only partially fixed.** `auth = required` now fails closed, which closes the concrete auth hole I called out, but H2 still jumps directly to `executeBufferedDataPlaneProxyRequest`. H1 runs `runMiddlewarePipeline` first (geo policy, request limits, ACL, rate limiting, etc.) and `handleLocationProxyPass` owns upstream selection, sticky affinity, configured retries, stale-connection retries, health/failure accounting, and related proxy metrics. Those policies/configurations can therefore still behave differently or be bypassed solely by negotiating H2. Since this PR now advertises native H2 `proxy_pass` as a production capability, protocol selection cannot become a policy escape hatch. Please factor/reuse the protocol-neutral decision/execution pieces or explicitly fail closed for every policy/orchestration feature H2 cannot yet honor, with regressions proving upstream side effects stay at zero for denied requests.

2. **Flow-control exhaustion is translated into a synthetic 502.** HTTP/2 send credit is transport state, not an upstream/application error. The connection starts with only 65,535 bytes of send credit, so a perfectly valid proxied response larger than that is replaced with `h2_flow_control_blocked` instead of being continued after `WINDOW_UPDATE`. The new 20 KiB test proves 16 KiB frame chunking, but it never crosses the connection-window boundary. The response needs to remain pending when credit reaches zero, resume when WINDOW_UPDATE arrives, and set END_STREAM only on the final DATA frame. Please add a regression with a body >65,535 bytes (and ideally a deliberately small peer stream window) that sends WINDOW_UPDATE and receives the original body exactly.

CI is also currently red on this commit: the appliance-profile Ubuntu and ARM jobs fail in the full integration suite, and macOS fails in the resumption/0-RTT smoke step. Unit/native-TLS portions are green, but the failing integration/smoke jobs must be resolved before merge.

Because the connected GitHub identity is also the PR author, GitHub will not let me submit a formal `REQUEST_CHANGES`; this COMMENT review is nevertheless a **do not merge yet / changes required** recommendation.

src/edge_gateway.zig
```diff
@@ -2583,17 +2622,223 @@ fn respondHttp2Stream(
...
+    const suffix_path = gproxy_runtime.proxySuffixPathForLocation(request.uri.path, matched, route_cfg.location_blocks);
+    const resolved = try gp.resolveProxyTarget(allocator, route_cfg.upstream_base_url, target, suffix_path);
+    defer allocator.free(resolved.url);
+    var upstream_url = try gp.appendProxyQueryString(allocator, resolved.url, request.uri.query);
+    defer upstream_url.deinit(allocator);
+
+    var ctx = http.request_context.RequestContext.init(allocator, correlation_id, connection_ip);
+    ctx.lifecycle = lifecycle;
+    const forwarded_proto = if (edge_config.hasTlsFiles(route_cfg)) "https" else "http";
+    var proxy_response = try gproxy_runtime.executeBufferedDataPlaneProxyRequest(
```
Jaruso, line 2927: **The original policy/orchestration blocker is only partially addressed.** Failing closed on `auth = required` fixes the concrete auth bypass, but this direct low-level call still sidesteps the normal request and proxy orchestration.

For H1, `runMiddlewarePipeline` enforces geo blocking, URI/header/body limits, ACL, and rate limiting before any upstream side effect. `handleLocationProxyPass` then owns upstream-pool selection, sticky affinity, configured retry count/idempotency rules, stale-connection retry, health/failure accounting, and proxy metrics. None of that is reached here; H2 negotiators therefore get materially different policy and reliability semantics for the same `proxy_pass` route.

Please make the shared boundary protocol-neutral rather than calling the one-attempt transport directly. A shape like this would keep rendering protocol-specific while sharing the important decisions:
```zig
const policy = try ghandlers.evaluateRequestPolicy(allocator, route_cfg, state, &ctx, request);
if (policy != .allow) return .{ .local_rejection = policy };

var plan = try gproxy_runtime.prepareLocationProxy(...);
defer plan.deinit();
var response = try gproxy_runtime.executeBufferedLocationProxy(&plan, ...); // selection/retry/health/metrics
return .{ .response = response };
```
If factoring that is intentionally out of scope for #521, the safe alternative is to fail closed whenever an H2 request would rely on a policy/orchestration capability this path cannot honor, and avoid advertising general H2 `proxy_pass` support yet. Please include at least ACL/rate-limit denial regressions with `upstream.requestCount() == 0`, plus a configured retry/failover parity test.

src/edge_gateway.zig
```diff
@@ -2576,24 +2602,384 @@ fn respondHttp2Stream(
...
+            return .{ .status_code = 429, .code = "rate_limited", .message = "Rate limit exceeded", .retry_after = "1" };
+        }
+    }
+
+    return null;
+}
+
+fn h2UnsupportedProxyOrchestration(cfg: *const edge_config.EdgeConfig, target: []const u8) bool {
+    _ = target;
+    const pool = gs.upstreamPoolForScope(cfg, .global);
```
Jaruso, line 3061: **Blocker — the fail-closed list is still not exhaustive.** This predicate can return `false` while H1 would execute semantics that H2 still bypasses. Concrete examples: `rewrite_rules`, `return_rules`, `conditional_rules`, `internal_redirect_rules`, `mirror_rules`, `auth_request_url`, global `max_in_flight_requests`, and a location-level `proxy_streaming_policy = response|full` while global `proxy_streaming_mode == .off`.

The code already has a useful inventory in `streamingUploadEligibilityBeforeBodyRead` for rewrite/return/conditional/internal-redirect/mirror/auth-request middleware. Please either factor a shared protocol-neutral pre-route/proxy policy boundary, or make this narrow H2 assurance path fail closed for every unsupported feature, including the matched block's effective streaming policy.

For example, the narrow version should look more like:
```zig
fn h2ProxyRequiresUnsupportedSemantics(cfg: *const EdgeConfig, matched: MatchResult) bool {
    if (cfg.rewrite_rules.len > 0 or cfg.return_rules.len > 0 or
        cfg.conditional_rules.len > 0 or cfg.internal_redirect_rules.len > 0 or
        cfg.mirror_rules.len > 0 or cfg.auth_request_url.len > 0 or
        cfg.max_in_flight_requests > 0) return true;

    if (matched.block.proxy_streaming_policy.responseStreamingEnabled(
        cfg.proxy_streaming_mode.responseStreamingEnabled(),
    )) return true;
    // plus the upstream-selection/retry/health features this direct executor skips.
    ...
}
```
Please add regressions for at least one pre-route rule and a location-level streaming override, asserting zero upstream side effects.

src/edge_gateway.zig
```diff
@@ -2576,24 +2602,384 @@ fn respondHttp2Stream(
...
+                    stream_send_window.* += @intCast(inc);
+                }
+            },
+            .rst_stream => {
+                if (frame.stream_id == stream_id) return error.ClientAborted;
+            },
+            .goaway => return error.ClientAborted,
+            else => {},
+        }
+    }
```
Jaruso: **Blocker — this fixes single-stream flow control by breaking H2 multiplexing.** `waitForHttp2SendCredit` becomes a second reader of the connection. Any legal frame not listed here is silently consumed by `else => {}`. For example: stream 1 exhausts its send credit, the client sends HEADERS for stream 3, then a WINDOW_UPDATE. This function drops stream 3's HEADERS; after stream 1 finishes, the main H2 loop can never dispatch stream 3.

The response writer must not call `readFrame`. Keep outbound response state (`body`, `offset`, stream id/windows) in the connection state and have a flush routine return `blocked` when credit is zero:
```zig
const Flush = enum { complete, blocked };
fn flushPendingResponse(...) !Flush {
    while (offset < body.len) {
        const credit = @min(conn_window.*, stream_window.*);
        if (credit <= 0) return .blocked;
        // emit bounded DATA and update offset/windows
    }
    return .complete;
}
```
Then let the **existing central frame switch** remain the sole reader; its WINDOW_UPDATE branch updates credit and retries pending outbound flushes. Add a multiplexing regression that sends another stream's HEADERS while the large response is flow-control blocked, then verifies both streams are served exactly once.

src/edge_gateway.zig
```diff
@@ -2576,24 +2602,384 @@ fn respondHttp2Stream(
...
+            return .{ .status_code = 431, .code = "invalid_request", .message = "Header too large" };
+        }
+        headers_total += h.name.len + h.value.len + 4;
+    }
+    if (http.request_limits.validateHeadersTotalSize(headers_total, limits) != .ok) {
+        return .{ .status_code = 431, .code = "invalid_request", .message = "Headers too large" };
+    }
+    if (request.body) |body| {
+        if (http.request_limits.validateBodySize(body.len, limits) != .ok) {
+            return .{ .status_code = 413, .code = "invalid_request", .message = "Request body too large" };
```
Jaruso, line 3007: **Blocker — this body limit is checked too late to provide a memory bound.** By the time `h2EvaluateRequestPolicy` sees `request.body`, every DATA payload has already been appended to `ps.body`, and the DATA handler immediately sends WINDOW_UPDATE for the same bytes. A client can therefore keep sending DATA indefinitely and force unbounded `ArrayList` growth, even though the eventual request would return 413.

Enforce the effective stream/connection body bound **before** `ps.body.appendSlice` and before replenishing receive credit. One safe shape is to resolve/store the effective request body limit once the request HEADERS/authority are known, then:
```zig
if (frame.payload.len > max_body -| ps.body.items.len) {
    ps.local_rejection = .payload_too_large;
    // do not append/replenish unlimited credit; terminate/reject this stream boundedly
} else {
    try ps.body.appendSlice(frame.payload);
    // replenish accepted bytes only
}
```
Also respect the connection-memory ceiling. Please add a test that sends an over-limit body across many DATA frames and proves a bounded 413 with `upstream.requestCount() == 0`.

src/edge_gateway.zig
```diff
@@ -2576,24 +2602,384 @@ fn respondHttp2Stream(
...
+) !Http2SyntheticRequest {
+    var raw = std.array_list.Managed(u8).init(allocator);
+    errdefer raw.deinit();
+    const request_line = try std.fmt.allocPrint(allocator, "{s} {s} HTTP/1.1\r\n", .{ method, path });
+    defer allocator.free(request_line);
+    try raw.appendSlice(request_line);
+    if (ps.authority) |authority| {
+        const host_line = try std.fmt.allocPrint(allocator, "Host: {s}\r\n", .{authority});
+        defer allocator.free(host_line);
+        try raw.appendSlice(host_line);
```
Jaruso, line 3145: **Security blocker — `:authority` bypasses the header-injection validation used for normal HPACK fields.** Regular fields go through `http.Headers.append`, whose `isValidHeaderValue` rejects CR/LF/control bytes. `:authority` is stored separately from the raw HPACK decoder and interpolated here directly into HTTP/1 text. The HPACK decoder itself only decodes bytes; it does not apply that validation.

A value such as an authority containing `\r\nX-Injected: value` can therefore create a second syntactically valid HTTP/1 header when `Request.parse` consumes this synthetic buffer, and that injected field can be forwarded upstream.

Please validate pseudo-header values before storing/using them (and reject the H2 stream as malformed), or preferably avoid textual H2->H1 reconstruction entirely. At minimum:
```zig
if (!http.headers.isValidHeaderValue(authority))
    return error.InvalidHttp2Request;
```
with stricter authority syntax validation as appropriate. Add an HPACK regression with CR/LF in `:authority` and assert the request never reaches the upstream.

Jaruso (commented):
Re-review of `bdcfb068` against #521 and all prior feedback: **changes are still required before merge**.

The core #521 assurance itself is now in good shape. The resumed-H2 test independently proves ALPN `h2`, authoritative OpenSSL `Reused`, Tardigrade `outcome="accepted"`, a real H2 request/response with the dedicated upstream body, and exactly one upstream execution. The cipher-mismatch closeout remains correct, docs/changelog are aligned, the earlier lifetime/query/authority/body/redaction/timeout/header-filtering defects remain fixed, and CI on this head is green.

The last two review blockers were addressed in intent, but the implementations expose additional correctness/security problems that keep the production H2 `proxy_pass` assurance path from being merge-safe:

1. **Fail-closed orchestration is still incomplete.** `h2UnsupportedProxyOrchestration` only checks retries, active connection cap, multi-upstream pools, and the global streaming mode. H1 also executes conditional/rewrite/return/internal-redirect rules, mirror/auth-request middleware, in-flight request backpressure, and route-level streaming policy before/around proxy side effects. Several of those can be configured while this predicate returns false, so ALPN `h2` can still bypass configured request semantics. Either factor the shared pre-route/proxy policy boundary or make the narrow H2 path exhaustively fail closed for every unsupported feature; do not maintain a hand-picked partial list.

2. **The flow-control fix consumes frames outside the central H2 state machine.** `waitForHttp2SendCredit` calls `readFrame` itself and drops every frame type it does not explicitly handle. While stream 1 is blocked on send credit, a legal HEADERS/DATA frame for stream 3 can therefore be consumed and lost before the main loop sees it. Keep the response body as pending outbound state; flushing should return `blocked` without reading. The main frame loop must remain the sole reader and retry the pending flush when WINDOW_UPDATE arrives. Add a multiplexing regression: large stream 1 response -> exhaust credit -> send stream 3 HEADERS -> send WINDOW_UPDATE -> verify both streams complete exactly once.

3. **Inbound H2 DATA is still buffered without an ingestion-time bound.** The code appends to `ps.body` and immediately replenishes stream/connection credit, while `validateBodySize` runs only after END_STREAM when the whole body is already resident. A client can therefore grow this buffer indefinitely despite the configured body limit. Enforce the effective body/connection-memory bound before append/replenishment and add a multi-DATA-frame oversized-body regression proving a bounded 413/no-upstream result.

4. **`:authority` bypasses the header-value injection guard.** Regular HPACK fields pass through `http.Headers.append`, which rejects CR/LF/control bytes, but `:authority` is stored separately and later interpolated directly into raw `Host: {authority}\r\n` text. The HPACK decoder itself does not validate field values. A crafted `:authority` containing CRLF can therefore synthesize additional HTTP/1 headers that `Request.parse` will accept as separate valid fields and forward upstream. Validate pseudo-header values before storage (at minimum using the same control-byte rules) or stop round-tripping H2 request metadata through raw HTTP/1 text. Add a malformed-`:authority` regression proving no upstream request occurs.

CI being green is good evidence for the covered paths, but the existing tests are all single-stream/well-formed-input cases and therefore do not exercise these boundaries.

Because the connected GitHub identity is also the PR author, GitHub will not allow a formal `REQUEST_CHANGES`; this COMMENT review is nevertheless a **do not merge yet / changes required** recommendation.

src/edge_gateway.zig
```diff
@@ -2576,22 +2736,351 @@ fn respondHttp2Stream(
...
+        cfg.mirror_rules.len > 0 or
+        cfg.auth_request_url.len > 0 or
+        cfg.max_in_flight_requests > 0) return true;
+
+    if (matched.block.proxy_streaming_policy.responseStreamingEnabled(
+        cfg.proxy_streaming_mode.responseStreamingEnabled(),
+    )) return true;
+
+    const pool = gs.upstreamPoolForScope(cfg, .global);
+    return cfg.upstream_retry_attempts > 1 or
```
Jaruso, line 3062: **Blocker — this is still not an exhaustive fail-closed boundary.** There are at least two concrete production behaviors above the low-level executor that this predicate still allows H2 to bypass:

1. **Identity-aware policy/trust.** H1 calls `primeRequestAuthContext` before middleware, `runMiddlewarePipeline` rate-limits on `ctx.identity`, and `handleLocationProxyPass` passes `ctx.identity/user_id/device_id/scopes` into the proxy. H2 currently calls `h2RateLimitDescriptor(null, ...)` and passes four `null` auth-context arguments to `executeBufferedDataPlaneProxyRequest`. A valid bearer/session identity can therefore land in a different rate-limit bucket and lose asserted/trusted identity solely by using H2.
2. **Pool/health orchestration.** H1 wraps the low-level executor with transparent stale pooled-connection retries and records passive upstream failure/success. This predicate does not reject `upstream_pool_enabled` or `upstream_max_fails`/`upstream_fail_timeout_ms`, so those configured semantics are skipped over H2 even with a single fallback upstream.

Please either share the H1 protocol-neutral orchestration, or extend the narrow fail-closed predicate to all capabilities that the direct call skips. A conservative version would additionally reject configured auth/session identity, upstream pooling, and passive-health behavior. Factoring `primeRequestAuthContext` plus the buffered proxy orchestration would be preferable. Add regressions for identity-aware rate limiting/propagation and at least one pool-health/stale-retry configuration.

src/edge_gateway.zig
```diff
@@ -2245,16 +2246,108 @@ fn h2AppendReadyStream(ready_streams: *std.array_list.Managed(u31), sid: u31) !v
...
+    conn_send_window: *i32,
+    stream_send_window: *i32,
+) !Http2FlushResult {
+    const max_data_frame: usize = HTTP2_MAX_FRAME_SIZE;
+    if (resp.body.len == 0) {
+        try http.http2_frame.writeFrame(conn.writer(), .data, http.http2_frame.Flags.END_STREAM, stream_id, &.{});
+        return .complete;
+    }
+    while (resp.offset < resp.body.len) {
+        const credit = @min(conn_send_window.*, stream_send_window.*);
```
Jaruso, line 2287: **Blocker — this new flow-control writer assumes `stream_send_window` reflects the peer's SETTINGS, but it never does.** Incoming SETTINGS are only ACKed and every new stream is still initialized with 65,535 bytes. RFC 9113 §§5.2/6.9.2 requires the server, as DATA sender, to honor the client's `SETTINGS_INITIAL_WINDOW_SIZE` and apply a setting change's delta to all active streams.

A client can advertise `{SETTINGS_INITIAL_WINDOW_SIZE = 1024}` before opening stream 1; this function can still emit up to 65,535 bytes before blocking.

Please track `peer_initial_window` (default 65,535), parse non-ACK SETTINGS, validate 0x4 <= 2^31-1, apply the delta to active `s.send_window` values, and initialize future streams with that peer value. Add a regression with a 1 KiB or zero stream window and a >16 KiB response, proving DATA stops at the grant until WINDOW_UPDATE.

src/edge_gateway.zig
```diff
@@ -2456,12 +2581,21 @@ fn handleHttp2Connection(conn: anytype, session: *ConnectionSession, cfg: *const
                 } else {
                     if (streams.getPtr(frame.stream_id)) |s| s.send_window += @intCast(inc);
                 }
+                // Either a connection-level or a stream-level WINDOW_UPDATE
+                // can unblock a parked response; this is the only place new
+                // send credit becomes available, and it runs on the central
+                // frame loop rather than a second reader.
+                try h2FlushPendingHttp2Responses(conn, allocator, state, &pending_responses, &streams, &conn_send_window);
```
Jaruso, line 2667: **Blocker / peer-triggerable overflow — before this flush runs, WINDOW_UPDATE credit is added with unchecked `i32 +=`.** `parseWindowUpdateIncrement` accepts any nonzero 31-bit increment. From the default 65,535 window, an increment of `0x7fffffff` makes the resulting window exceed 2^31-1 immediately.

RFC 9113 §6.9.1 requires FLOW_CONTROL_ERROR when the resulting flow-control window exceeds the maximum; unchecked arithmetic can instead trap in safety-enabled builds or corrupt the window.

Please use checked/wider arithmetic, e.g. compute the prospective value in `i64`, reject `> maxInt(i32)`, then commit it. Connection overflow should produce GOAWAY(FLOW_CONTROL_ERROR); stream overflow should produce RST_STREAM(FLOW_CONTROL_ERROR). Add regressions for both maximum-increment cases.

src/edge_gateway.zig
```diff
@@ -2426,15 +2536,30 @@ fn handleHttp2Connection(conn: anytype, session: *ConnectionSession, cfg: *const
...
-                if (streams.getPtr(frame.stream_id)) |s| s.send_window -= @intCast(frame.payload.len);
-                conn_send_window -= @intCast(frame.payload.len);
                 if (pending.getPtr(frame.stream_id)) |ps| {
                     ps.transport_early = ps.transport_early or frame_transport_early;
-                    try ps.body.appendSlice(frame.payload);
-                    try http.http2_frame.writeWindowUpdate(conn.writer(), frame.stream_id, @intCast(frame.payload.len));
-                    try http.http2_frame.writeWindowUpdate(conn.writer(), 0, @intCast(frame.payload.len));
-                    if (streams.getPtr(frame.stream_id)) |s| s.send_window += @intCast(frame.payload.len);
-                    conn_send_window += @intCast(frame.payload.len);
+                    const max_body = cfg.request_limits.effectiveMaxBodySize();
```
Jaruso, line 2590: **Blocker — the ingestion bound still ignores the configured per-connection memory ceiling.** This limits each pending stream only by `request_limits.effectiveMaxBodySize()`. `EdgeConfig.max_connection_memory_bytes` is explicitly the "Maximum in-memory bytes retained per active connection", and H1 enforces that setting at its connection boundary. An H2 connection can retain a body larger than a lower configured connection cap; multiple concurrent streams can each retain up to `max_body` independently.

Please track aggregate buffered H2 request bytes for this connection and reject before append when either the stream body cap or the connection cap would be crossed. Decrement the aggregate whenever a pending stream is removed. Add a test with `MAX_CONNECTION_MEMORY_BYTES < MAX_BODY_SIZE`—ideally two concurrent streams whose combined retained bodies cross the cap—and prove bounded rejection with no unintended upstream execution.

src/edge_gateway.zig
```diff
@@ -2576,22 +2736,351 @@ fn respondHttp2Stream(
...
+    if (ps.body.items.len > 0 and !saw_content_length) {
+        const line = try std.fmt.allocPrint(allocator, "Content-Length: {d}\r\n", .{ps.body.items.len});
+        defer allocator.free(line);
+        try raw.appendSlice(line);
+    }
+    try raw.appendSlice("\r\n");
+    try raw.appendSlice(ps.body.items);
+    const owned_raw = try raw.toOwnedSlice();
+    errdefer allocator.free(owned_raw);
+    const parsed = try http.Request.parse(allocator, owned_raw, MAX_REQUEST_SIZE);
```
Jaruso, line 3167: **Blocker — mismatched H2 Content-Length can silently truncate the request here.** When a client supplies Content-Length, the adapter preserves it and appends all H2 DATA bytes. `Request.parse` consumes exactly the declared HTTP/1 Content-Length and returns `bytes_consumed`, but this function ignores that count.

Concrete case: `content-length: 5` + 10 bytes of H2 DATA. The parsed request contains only the first five bytes; the remaining five stay unconsumed in `owned_raw` and disappear, so the upstream receives a request different from the H2 message. RFC 9113 §8.1.1 requires Content-Length to equal the sum of DATA payload lengths.

Validate before/after synthesis:
```zig
if (ps.headers.get("content-length")) |raw_len| {
    const declared = std.fmt.parseInt(usize, raw_len, 10) catch return error.InvalidHttp2Request;
    if (declared != ps.body.items.len) return error.InvalidHttp2Request;
}
```
Also reject HTTP/2-forbidden connection-specific fields (`connection`, `proxy-connection`, `keep-alive`, `transfer-encoding`, `upgrade`; `te` only permits `trailers`) rather than allowing HTTP/1 parsing to reinterpret H2 DATA. Add mismatched-Content-Length and `transfer-encoding: chunked` regressions with `upstream.requestCount() == 0`.

Jaruso (commented):
Re-review of `aa56e99c` against #521 and all prior feedback: **changes are still required before merge**.

This revision closes several important blockers correctly. The H2 response writer no longer becomes a second connection reader; pending responses are parked and resumed by the central frame loop, and the new multiplexing regression exercises that behavior. The malformed-`:authority` CRLF injection path is now rejected before synthesis/upstream access. Request-body growth is bounded by the configured body-size limit at DATA ingestion time, and the new multi-frame regression proves the 413/no-upstream path. The expanded pre-route/streaming fail-closed tests are also good. The core #521 assurance remains correct: authoritative OpenSSL `Reused`, ALPN `h2`, Tardigrade `outcome="accepted"`, one real H2 application request/response, exactly one upstream execution, and the cipher-mismatch closeout remains honest. CI is green on this head.

I still do **not** recommend merging yet because the production H2 `proxy_pass` path has five remaining correctness/security boundaries:

1. **The fail-closed orchestration predicate still is not exhaustive.** H1 primes auth/session identity before middleware, uses that identity for rate-limit keys, and passes identity/user/device/scopes into the proxy trust boundary. H2 always rate-limits with `identity = null` and passes all auth context fields as null. The predicate also misses enabled upstream keep-alive/stale-retry and passive-health semantics (`upstream_pool_enabled`, `upstream_max_fails`/`upstream_fail_timeout_ms`) that live above `executeBufferedDataPlaneProxyRequest`. Negotiating H2 can therefore still change policy/health behavior on a route the predicate labels supported. Either reuse the protocol-neutral orchestration or fail closed when any such capability is enabled.

2. **Peer `SETTINGS_INITIAL_WINDOW_SIZE` is ignored.** Incoming SETTINGS are only ACKed and every stream is initialized with a hard-coded 65,535-byte server-send window. RFC 9113 requires a sender to respect the receiver's advertised stream window and to apply an INITIAL_WINDOW_SIZE delta to existing streams. A client advertising a 1 KiB (or zero) stream window can currently receive far more DATA than it granted. Track the peer initial window, parse SETTINGS, adjust active streams by the delta, and initialize new streams from that value. Add a small-window large-response regression.

3. **WINDOW_UPDATE additions are unchecked.** A legal 31-bit increment can make `conn_send_window += inc` or `stream.send_window += inc` exceed 2^31-1. RFC 9113 requires FLOW_CONTROL_ERROR; unchecked `i32` arithmetic can instead overflow/trap in safety-enabled builds or corrupt flow-control state. Use checked/wider arithmetic and emit GOAWAY for connection overflow / RST_STREAM for stream overflow, with regression coverage.

4. **The inbound body fix still bypasses `max_connection_memory_bytes`.** It caps each stream at `request_limits.max_body_size`, but `max_connection_memory_bytes` is explicitly the per-active-connection memory ceiling. A single stream can exceed a lower configured connection cap, and many concurrent streams can each retain up to the body limit. Track aggregate H2 request buffering against the connection limit before append; add a regression where the connection cap is lower than the body limit and prove bounded rejection/no upstream side effect.

5. **H2 Content-Length/message semantics are not validated before HTTP/1 reconstruction.** If the client declares `content-length: 5` but sends 10 DATA bytes, `synthesizeHttp2Request` preserves CL=5, appends all 10 bytes, and `Request.parse` consumes only the first five; `parsed.bytes_consumed` is ignored, so the extra five bytes silently disappear and the upstream receives a truncated request. RFC 9113 classifies a content-length value that differs from the sum of DATA payload lengths as malformed. Validate declared Content-Length against `ps.body.items.len` before synthesis (and reject H2-forbidden connection-specific fields such as Transfer-Encoding rather than letting HTTP/1 parsing reinterpret DATA). Add a mismatched-Content-Length regression proving zero upstream side effects.

The connected GitHub identity is the PR author, so GitHub still will not allow a formal `REQUEST_CHANGES`; this COMMENT review is a **do not merge yet / changes required** recommendation.

Jaruso (commented):
Re-review of current head `3141cad5fd7c141ee3f04c17d555d6cab86e2b19` against #521 and the prior review findings.

This iteration closes most of the previous blockers cleanly: peer initial-window handling exists, WINDOW_UPDATE addition is checked, request-body aggregate accounting exists, H2 framing validation now rejects mismatched Content-Length/connection-specific fields, identity is primed for rate limiting/upstream trust, passive health/retry configuration is failed closed, and the existing resumed-H2 acceptance evidence remains strong. CI is green on this head.

I also re-traced the pooling concern from my prior review. The low-level pooled HTTP/1 executor already performs its own one-time fresh retry when a checked-out connection is stale, so I am **not** carrying that concern forward.

Two blockers remain, both in the newly added flow-control/accounting code:

1. Applying a later `SETTINGS_INITIAL_WINDOW_SIZE` delta can overflow an already-expanded active stream window. The current comment claiming this cannot exceed `i32` is incorrect: an active stream can already be at `maxInt(i32)` after WINDOW_UPDATE, then a positive SETTINGS delta can push `adjusted` past the maximum and make `@intCast` trap. RFC 9113 §6.9.2 requires a connection-level FLOW_CONTROL_ERROR instead.
2. The stream-level WINDOW_UPDATE overflow cleanup frees a pending request body without decrementing `buffered_request_bytes`. After the RST_STREAM, that freed body continues counting against `max_connection_memory_bytes`, so later valid streams on the same connection can be rejected spuriously.

Both are peer-triggerable and straightforward to cover with focused regressions. I would not merge until they are fixed.

src/edge_gateway.zig
```diff
@@ -2382,7 +2504,31 @@ fn handleHttp2Connection(conn: anytype, session: *ConnectionSession, cfg: *const
...
+                            while (it.next()) |s| {
+                                const adjusted: i64 = @as(i64, s.send_window) + delta;
+                                // RFC 9113 §6.9.2 allows this to go negative
+                                // (it must not be treated as an error); it
+                                // simply blocks further sends until the peer
+                                // grants more credit. It cannot itself exceed
+                                // the 2^31-1 ceiling since both operands are
+                                // already within range and delta is bounded
+                                // by two values <= 2^31-1.
+                                s.send_window = @intCast(adjusted);
```
Jaruso: **Blocker — a SETTINGS delta can still overflow/trap an active stream window.** The comment above this cast is not generally true. `s.send_window` may already have been raised to `maxInt(i32)` by a valid stream WINDOW_UPDATE. If the previous `peer_initial_window` was 0 and the peer then sends `SETTINGS_INITIAL_WINDOW_SIZE = 0x7fffffff`, `delta` is +0x7fffffff and `adjusted` exceeds the representable flow-control maximum. `@intCast(adjusted)` can trap instead of producing the required connection error.

RFC 9113 §6.9.2 requires a SETTINGS change that makes **any** stream window exceed 2^31-1 to terminate the connection with `FLOW_CONTROL_ERROR`.

Please check before committing the adjusted value, e.g.:
```zig
const adjusted: i64 = @as(i64, s.send_window) + delta;
if (adjusted > std.math.maxInt(i32)) {
    try http.http2_frame.writeGoaway(
        conn.writer(),
        last_client_stream_id,
        http.http2_stream.ErrorCode.flow_control_error.value(),
    );
    return error.Http2FlowControlError;
}
s.send_window = @intCast(adjusted);
```

Add a regression with an active stream: advertise initial window 0, open the stream without END_STREAM, raise that stream to the max via WINDOW_UPDATE, then change SETTINGS_INITIAL_WINDOW_SIZE from 0 to max. Assert GOAWAY(FLOW_CONTROL_ERROR) and no process crash.

src/edge_gateway.zig
```diff
@@ -2451,14 +2621,47 @@ fn handleHttp2Connection(conn: anytype, session: *ConnectionSession, cfg: *const
...
+                    }
+                    conn_send_window = @intCast(new_window);
+                } else if (streams.getPtr(frame.stream_id)) |s| {
+                    const new_window: i64 = @as(i64, s.send_window) + @as(i64, inc);
+                    if (new_window > std.math.maxInt(i32)) {
+                        try http.http2_frame.writeRstStream(conn.writer(), frame.stream_id, http.http2_stream.ErrorCode.flow_control_error.value());
+                        _ = streams.remove(frame.stream_id);
+                        if (pending.fetchRemove(frame.stream_id)) |removed| {
+                            var tmp = removed.value;
+                            tmp.deinit(allocator);
```
Jaruso, line 2653: **Blocker — this stream-error cleanup leaks aggregate request-memory accounting.** When the overflowing WINDOW_UPDATE causes RST_STREAM, `pending.fetchRemove()` frees `tmp.body`, but unlike the normal dispatch and `.rst_stream` cleanup paths it never subtracts `tmp.body.items.len` from `buffered_request_bytes`.

That leaves `max_connection_memory_bytes` permanently inflated for the rest of the H2 connection. A peer can buffer (say) 1,000 bytes on stream 1, trigger this stream-level flow-control error, then send a valid 1,000-byte body on stream 3 under a 1,500-byte connection cap; stream 1's memory is gone, but stream 3 is still rejected because the stale counter says 2,000 bytes are retained.

Mirror the other removal paths before `deinit`:
```zig
if (pending.fetchRemove(frame.stream_id)) |removed| {
    var tmp = removed.value;
    buffered_request_bytes -= @min(buffered_request_bytes, tmp.body.items.len);
    tmp.deinit(allocator);
}
```

Please add the two-stream regression above and assert stream 3 reaches its upstream successfully after stream 1 is reset.

src/edge_gateway.zig
```diff
@@ -2382,7 +2504,41 @@ fn handleHttp2Connection(conn: anytype, session: *ConnectionSession, cfg: *const
...
+                            var it = streams.valueIterator();
+                            while (it.next()) |s| {
+                                const adjusted: i64 = @as(i64, s.send_window) + delta;
+                                // RFC 9113 §6.9.2: a stream already sitting
+                                // at (or near) maxInt(i32) from a prior
+                                // WINDOW_UPDATE can still be pushed past the
+                                // 2^31-1 ceiling by a positive delta here --
+                                // that must be a connection-level
+                                // FLOW_CONTROL_ERROR, not an unchecked cast.
+                                if (adjusted > std.math.maxInt(i32)) {
```
Jaruso, line 2525: Verified fixed on `a1730d98`. The SETTINGS delta is now evaluated in `i64`, and a result above the RFC 9113 flow-control maximum is converted to connection-level GOAWAY(FLOW_CONTROL_ERROR) before the narrowing cast. The new active-stream regression exercises the exact prior failure sequence, including a stream already at `maxInt(i32)`.

src/edge_gateway.zig
```diff
@@ -2451,14 +2631,48 @@ fn handleHttp2Connection(conn: anytype, session: *ConnectionSession, cfg: *const
...
+                    }
+                    conn_send_window = @intCast(new_window);
+                } else if (streams.getPtr(frame.stream_id)) |s| {
+                    const new_window: i64 = @as(i64, s.send_window) + @as(i64, inc);
+                    if (new_window > std.math.maxInt(i32)) {
+                        try http.http2_frame.writeRstStream(conn.writer(), frame.stream_id, http.http2_stream.ErrorCode.flow_control_error.value());
+                        _ = streams.remove(frame.stream_id);
+                        if (pending.fetchRemove(frame.stream_id)) |removed| {
+                            var tmp = removed.value;
+                            buffered_request_bytes -= @min(buffered_request_bytes, tmp.body.items.len);
```
Jaruso, line 2652: Verified fixed on `a1730d98`. This cleanup now mirrors the normal dispatch/RST paths by subtracting the removed body's retained bytes before deinit. The new two-stream test is especially useful: it proves the counter is actually reusable on the same connection rather than merely checking that stream 1 was reset.

Jaruso (commented):
Re-review of `a1730d98` against issue #521 and all prior review findings: **code review is clean; I found no remaining merge blocker.**

The two blockers from `3141cad5` are now correctly closed:

- SETTINGS_INITIAL_WINDOW_SIZE deltas are applied in widened arithmetic and a resulting stream window above `2^31-1` produces connection-level GOAWAY(FLOW_CONTROL_ERROR) instead of reaching a trapping `@intCast`. The new integration regression drives an active stream to the ceiling with WINDOW_UPDATE, applies a positive SETTINGS delta, and verifies the GOAWAY.
- Stream-level WINDOW_UPDATE overflow cleanup now releases that stream's retained request bytes from `buffered_request_bytes` before deinit. The new same-connection two-stream regression proves the reset stream no longer poisons the connection-memory budget and a later legitimate body reaches its upstream.

I also re-swept the surrounding flow-control state machine rather than reviewing only those lines. The earlier fixes remain intact: peer initial-window tracking, WINDOW_UPDATE overflow handling, parked-response multiplexing, aggregate request-body memory limits, H2 framing validation, authority validation, identity-aware rate limiting/propagation, policy fail-closed behavior, request/response translation, transcript secret redaction, and timeout/header lifetime handling. I found no new regression in the latest delta.

#521's acceptance criteria remain satisfied: the OpenSSL H2 case negotiates ALPN `h2`, obtains a real production session, reconnects with authoritative `Reused, TLSv1.3` evidence, observes the server's `outcome="accepted"` counter increase, sends a real H2 request over that resumed connection, validates the dedicated upstream body, and proves exactly one upstream execution. The cipher-mismatch closeout remains production-honest and does not introduce a test-only cipher.

CI run 1013 is still marked `in_progress` at the workflow level as of this review, but every completed job I inspected is green, including the macOS appliance resumption/0-RTT interop smoke that exercises the `interop.*` regressions added here, plus unit/format/lint/security/H3/TLS-audit jobs. **My merge recommendation is: code is ready; merge once the current CI run reaches an overall successful conclusion.**

GitHub cannot record an APPROVE from the connected identity because it is the PR author, so this is submitted as COMMENTED rather than APPROVED.

PR #527, opened 2026-07-27
quic: prove external H3 resumption/0-RTT interop through production composition (#522)

Jaruso:
## Summary

Issue #522 required extending the checked-in external interop suite so that requests from an independent H3/QUIC peer route through the *production* `http3_runtime.Runtime` composition (native resumption runtime, process-local replay gate, shared HTTP early-data policy) rather than the direct `Tls13Backend -> quic.Connection -> H3.Conn` harness `tests/h3_interop_tool.zig`/`scripts/interop/run-interop.sh` already prove wire-interop with.

This PR adds four `h3interop.quic.*` cases to `tests/integration.zig` (same file/pattern as the existing `interop.openssl.*` cases, run via the same `zig build test-integration-resumption-interop` step — `h3interop.` contains the `interop.` filter substring) that spawn the real `tardi` binary with `TARDIGRADE_HTTP3_ENABLED`/`TARDIGRADE_HTTP3_ENABLE_0RTT` and drive it with the canonical ngtcp2/GnuTLS `gtlsclient` peer already documented in `scripts/interop/README.md`:

- **`h3interop.quic.resume`** (Part A) — authoritative 1-RTT resumption: `tardigrade_tls_resumption_outcome_total{transport="quic",outcome="accepted"}`, a real H3 request/response, and zero accepted 0-RTT packets (proving this is ordinary resumption, not early data).
- **`h3interop.quic.early.accepted`** (Part B.1) — a real `type=0RTT` QUIC packet accepted end to end through the production runtime and the shared HTTP early-data policy, with the upstream hit exactly once.
- **`h3interop.quic.early.unsafe_425`** (Part B.2) — a POST offered as early data is 425'd by the same shared HTTP-level safety policy H1/H2 already use, before any upstream side effect.
- **`h3interop.quic.early.replay_fallback`** (Part B.3) — the identical early ticket replayed within one process is rejected by the process-local replay gate (`tardigrade_tls_early_data_replay_total{outcome="duplicate"}`) rather than re-executed as 0-RTT, and the connection remains usable for further requests afterward.

Also fixes `tests/h3_interop_tool.zig`'s verbose event logger, which didn't compile against the `zero_rtt_packet`/`early_data_decision` variants #523/#524 added to `quic.connection.Event` — the checked-in interop tool was broken on `main` before this PR.

### What's documented as unreachable, not faked

Three of #522's required scenarios genuinely aren't reachable through today's *production configuration surface* — verified directly against the running server, not assumed:

- **Replay-store-unavailable fail-closed**: `Runtime.zeroRttCarrierEnabled` couples replay-gate presence directly to whether the 0-RTT carrier is enabled at all, so "resumption + 0-RTT on, replay gate absent" collapses to the already-deterministically-covered `.disabled`/`keys_unavailable` state rather than a distinct `replay_unavailable` outcome.
- **H3 SETTINGS incompatibility**: `edge_gateway.zig` never threads an operator-facing H3-SETTINGS config surface into `Runtime.Config.h3_settings`; it always uses compiled-in defaults.
- **QUIC transport-parameter incompatibility**: `quicConfigFrom` in `http3_runtime.zig` maps only `max_datagram_size`/`zero_rtt_enabled` from operator config; the RFC 9000 §7.4.1 limits stay compile-time constants.

Per the issue's own instructions, these are documented (in `scripts/interop/README.md`) rather than faked with test-only session/config mutation. #523's existing deterministic driver tests remain the coverage of record for all three.

A related finding, also documented: production QUIC/H3 0-RTT is currently only reachable under the `.general` (OpenSSL-adapter) TLS build profile — the appliance profile (`-Dtls-profile=appliance`) explicitly rejects `TARDIGRADE_HTTP3_ENABLE_0RTT` at config-validation time. The new cases gate on `.general` (the default `zig build`) accordingly, the mirror image of how `interop.openssl.*` gates on appliance.

### Explicitly out of scope (per the issue)

Part C's `docs/RESUMPTION_TEST_PLAN.md` reconciliation is **not** included — the issue explicitly defers it until #519, #520, and #521 reach final disposition, and all three are still open as of this PR.

## Test plan

- [x] `zig build test-integration-resumption-interop` (default `.general` profile) with `NGTCP2_GTLSCLIENT_PATH` pointed at a locally-built ngtcp2 `gtlsclient` (per `scripts/interop/README.md`): all 4 new cases pass, all pre-existing cases in the same file skip cleanly under the wrong profile (no regressions).
- [x] `zig build -Dtls-profile=appliance test-integration-resumption-interop`: the 13 pre-existing `interop.openssl.*`/`restart.*`/`soak.*` cases still pass unchanged; the 4 new QUIC cases skip cleanly (0-RTT unsupported under this profile).
- [x] `zig build test` (full unit suite): no regressions.
- [x] Added PR CI gate: `h3-resumption-interop` builds pinned nghttp3 `v1.18.0` + ngtcp2 `v1.25.0`, exports `NGTCP2_GTLSCLIENT_PATH`, and runs `zig build test-integration-resumption-interop -Dtls-profile=general`.
- [x] Latest local review-fix check: `zig build test-integration-resumption-interop -Dtls-profile=general --summary all --error-style verbose` compiled cleanly; live external-peer cases skipped locally because `NGTCP2_GTLSCLIENT_PATH` is not set.
- [x] `zig build build-h3-interop test-quic-h3-driver`: interop tool builds and driver tests pass.
- [x] Every new metric/behavior assertion was independently verified by hand against a live `tardi` process + `gtlsclient` before being encoded as a test assertion (not guessed).

Refs #522


tests/integration.zig
```diff
@@ -2564,6 +2580,590 @@ fn expectMetricAtLeast(body: []const u8, name: []const u8, labels: []const []con
...
+    var bound_len: std.c.socklen_t = @sizeOf(std.c.sockaddr.in);
+    if (std.c.getsockname(fd, @ptrCast(&bound), &bound_len) != 0) return error.GetSockNameFailed;
+    return std.mem.bigToNative(u16, bound.port);
+}
+
+/// #522: skip the external ngtcp2/GnuTLS H3/QUIC interop cases when the
+/// canonical peer (built per `scripts/interop/README.md`, `gtlsclient`) isn't
+/// available locally -- like `requireOpenssl`, an optional external peer
+/// must never fail a stripped-down dev environment; CI builds it.
+fn requireNgtcp2Client(allocator: std.mem.Allocator) ![]u8 {
```
Jaruso, line 2619: [P1] The canonical external peer is optional here, but **no automated workflow on this PR provisions it**, so all four new #522 tests can silently skip while CI stays green. The current PR CI runs unit tests under the general profile; the appliance integration/resumption jobs skip these general-only cases; and the push-only general integration job neither builds ngtcp2 nor sets `NGTCP2_GTLSCLIENT_PATH`.

Because this PR is assurance work, the independent-peer proof needs to be an actual automated gate, not only a locally-run test. Please add reproducible/pinned ngtcp2+nghttp3 provisioning in CI, set `NGTCP2_GTLSCLIENT_PATH`, and run the general-profile `test-integration-resumption-interop` suite. It is fine to retain this local missing-peer skip, but the CI job that owns #522 should fail if its required peer cannot be built/found.

For example, conceptually:
```yaml
- name: Build pinned ngtcp2 GnuTLS peer
  run: ./scripts/interop/build-ngtcp2-ci.sh   # or equivalent pinned setup

- name: Run H3 resumption/0-RTT interop
  env:
    NGTCP2_GTLSCLIENT_PATH: ${{ runner.temp }}/ngtcp2/build/examples/gtlsclient
  run: zig build test-integration-resumption-interop -Dtls-profile=general --summary all --error-style verbose
```
Please pin the peer revision(s), since these tests also consume peer diagnostic output such as `type=0RTT` / status lines.

Jaruso (commented):
**Blocking review findings.** GitHub does not permit the connected account to submit a `REQUEST_CHANGES` review on its own PR, so this is submitted as a comment review; I would not merge until the blockers below are addressed.

The core direction is correct: these tests drive the real `tardi` edge-gateway composition with an independent ngtcp2/GnuTLS peer, the accepted 0-RTT case has strong server-side typed evidence plus an upstream exactly-once observation, and documenting genuinely unreachable SETTINGS/transport/replay-store cases instead of manufacturing test-only state follows #522's intended contract.

I found four implementation/coverage gaps called out inline, plus one issue-lifecycle blocker:

1. **The new external H3 suite is not actually exercised by automated CI.** `requireNgtcp2Client()` skips when `NGTCP2_GTLSCLIENT_PATH` is absent, but no workflow on this head builds/provisions `gtlsclient` or sets that variable. PR CI can therefore be green while every new #522 test skips.
2. **Part A does not yet satisfy #522's independent-proof / exactly-once requirements.** The server metric proves Tardigrade accepted resumption, but the test does not assert a stable external-peer reuse indicator and the local `return 200 alive` route provides no application/upstream execution counter.
3. **The unsafe-425 case proves rejection-before-side-effects but stops there.** #522 also requires proving that the same connection remains usable according to the fallback/retry contract.
4. **The replay test equates one request with one accepted QUIC 0-RTT packet.** Packet count is a transport/packetization detail; snapshot the accepted-packet counter before replay and prove it does not increase instead.

There is also a scope/closure mismatch: this PR explicitly says Part C is intentionally omitted because #519/#520/#521 are still open, while #522 still owns the final `RESUMPTION_TEST_PLAN.md` reconciliation and has acceptance criteria for that work. Therefore `Closes #522` is premature. Please use `Refs #522` (or otherwise leave #522 open) until Part C can be completed, unless the issue is explicitly re-scoped first.

Once these are fixed, the actual production 0-RTT assurance here is strong and I would expect the next pass to be primarily verification rather than redesign.

tests/integration.zig
```diff
@@ -2564,6 +2580,590 @@ fn expectMetricAtLeast(body: []const u8, name: []const u8, labels: []const []con
...
+    });
+    defer second.deinit(allocator);
+    try std.testing.expectEqual(std.meta.Tag(bounded_process.Outcome).normal_exit, std.meta.activeTag(second.outcome));
+
+    // 5-8: the resumed connection actually served a real H3 request/response
+    // exactly once, and Tardigrade's own typed QUIC resumption outcome --
+    // not just a second successful handshake -- independently reports
+    // `accepted`. No accepted 0-RTT packet proves this was ordinary 1-RTT
+    // resumption, not early data.
+    try assertContains(second.stderr, "alive");
```
Jaruso: [P1] This does not yet establish two explicit Part-A acceptance criteria. `alive` proves that a request/response succeeded, and the Tardigrade metric below proves the server selected resumption, but there is **no independent gtlsclient-side reuse assertion** and this local `return` route has no application/upstream counter with which to prove exactly-once execution.

Please make the resumed case observable at both seams. A good shape is to proxy `/safe` to `UpstreamServer`, reset capture after the ticket-bootstrap connection, then assert one hit on the resumed request, while also asserting a stable reuse/resumption indicator from the pinned gtlsclient output:
```zig
try upstream.resetCapture();
var second = try runGtlsClient(... .{ .disable_early_data = true, ... });
try assertGtlsSessionReused(second.stderr); // stable indicator from the pinned peer
try std.testing.expectEqual(@as(u32, 1), upstream.requestCount());
try expectMetricAtLeast(metrics.body,
    "tardigrade_tls_resumption_outcome_total",
    &.{ "transport=\"quic\"", "outcome=\"accepted\"" }, 1);
```
That gives the independent peer evidence + Tardigrade evidence + exactly-once application proof #522 requires.

tests/integration.zig
```diff
@@ -2564,6 +2580,590 @@ fn expectMetricAtLeast(body: []const u8, name: []const u8, labels: []const []con
...
+    // Rejected before any application/upstream side effect.
+    try std.testing.expectEqual(@as(u32, 0), upstream.requestCount());
+
+    var metrics = try sendPureZigTlsHttp1Request(allocator, tardigrade.port, "/status/metrics");
+    defer metrics.deinit();
+    // The QUIC/TLS layer still authoritatively accepted the early data --
+    // the rejection is the shared HTTP-level safety policy, not a
+    // replay/compatibility failure being misattributed.
+    try expectMetricAtLeast(metrics.body, "tardigrade_quic_zero_rtt_packet_total", &.{"outcome=\"accepted\""}, 1);
+    try expectMetricAtLeast(metrics.body, "tardigrade_http_early_data_decisions_total", &.{ "protocol=\"h3\"", "decision=\"too_early\"" }, 1);
```
Jaruso, line 3124: [P1] The unsafe case currently ends after proving `425` and zero upstream side effects, but #522 also requires proving **the connection remains usable according to the shared fallback/retry contract**. With `runGtlsClient()` always adding `--exit-on-first-stream-close`, this test never demonstrates a post-425 request on the same QUIC connection.

Please keep this connection alive long enough to issue a normal post-handshake H3 request after the rejected early stream and assert that request succeeds exactly once. If `gtlsclient` needs a different invocation shape, make `--exit-on-first-stream-close` optional or add the narrow driver control needed here. A fresh separate connection would only prove the server is still alive; it would not prove the connection that received the 425 remained usable.

tests/integration.zig
```diff
@@ -2564,6 +2580,590 @@ fn expectMetricAtLeast(body: []const u8, name: []const u8, labels: []const []con
...
+    try std.testing.expectEqual(@as(u32, 2), upstream.requestCount());
+
+    var metrics = try sendPureZigTlsHttp1Request(allocator, tardigrade.port, "/status/metrics");
+    defer metrics.deinit();
+    // The replay-specific typed outcome, not a generic ticket/key miss.
+    try expectMetricAtLeast(metrics.body, "tardigrade_tls_early_data_replay_total", &.{"outcome=\"duplicate\""}, 1);
+    // Exactly one *early* execution across both attempts -- the replay was
+    // never itself accepted as 0-RTT (the second upstream hit above came
+    // from the safe 1-RTT fallback, proven independently here).
+    try std.testing.expectEqual(@as(u64, 1), prometheusLabeledMetricValue(metrics.body, "tardigrade_quic_zero_rtt_packet_total", &.{"outcome=\"accepted\""}) orelse 0);
```
Jaruso: [P1] Please don't use `accepted 0-RTT packet count == 1` as the replay invariant. One valid H3 early request can occupy more than one QUIC 0-RTT packet depending on packetization, peer version, MTU, and framing, so request-count and packet-count are not interchangeable.

Snapshot the accepted-packet counter **after the first successful early request**, require it to be nonzero, then prove replay does not increase it:
```zig
const accepted_before_replay = prometheusLabeledMetricValue(
    metrics_before.body,
    "tardigrade_quic_zero_rtt_packet_total",
    &.{"outcome=\"accepted\""},
) orelse 0;
try std.testing.expect(accepted_before_replay > 0);

// replay identical ticket/request ...

const accepted_after_replay = prometheusLabeledMetricValue(
    metrics_after.body,
    "tardigrade_quic_zero_rtt_packet_total",
    &.{"outcome=\"accepted\""},
) orelse 0;
try std.testing.expectEqual(accepted_before_replay, accepted_after_replay);
```
Together with `early_data_replay_total{outcome="duplicate"}` and the upstream count, that proves the second attempt's early bytes were rejected without depending on how many packets the first request needed.

.github/workflows/ci.yml
```diff
@@ -499,6 +499,57 @@ jobs:
...
+
+      - name: Install system dependencies
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y \
+            cmake \
+            g++-14 \
+            gcc-14 \
+            libev-dev \
+            libgnutls28-dev \
```
Jaruso: [P1] The new gate currently violates the repository's dependency-audit policy, which is why `TLS dependency audit` is failing on this head. `scripts/audit-dependencies.sh` intentionally scans `.github/workflows` for names including `ngtcp2`, `nghttp3`, and `gnutls`, while exempting only `scripts/interop/*` as the out-of-process peer boundary. This workflow has non-comment occurrences in the package name (`libgnutls28-dev`), step/id names, helper path, and env var, so the audit cannot pass as written.

Please keep the dependency-specific provisioning/naming behind the existing `scripts/interop/` exception instead of weakening the audit. For example, make the workflow generic:
```yaml
- name: Install external H3 peer dependencies
  run: ./scripts/interop/install-h3-peer-deps-ci.sh

- name: Build pinned external H3 peer
  id: build-h3-peer
  run: ./scripts/interop/build-h3-peer-ci.sh

- name: Run H3 resumption/0-RTT interop suite
  env:
    H3_INTEROP_CLIENT_PATH: ${{ steps.build-h3-peer.outputs.client_path }}
  run: zig build test-integration-resumption-interop -Dtls-profile=general ...
```
The helper(s) under `scripts/interop/` can contain the actual ngtcp2/GnuTLS package/repository names because that directory is exactly the policy-approved external-peer boundary. The Zig harness can likewise consume a generic `H3_INTEROP_CLIENT_PATH` while still naming ngtcp2 internally/tests/docs.

scripts/interop/build-ngtcp2-ci.sh
```diff
@@ -0,0 +1,48 @@
...
+  git clone --depth 1 --branch "$NGTCP2_REF" https://github.com/ngtcp2/ngtcp2.git "$ROOT/ngtcp2"
+fi
+
+export PKG_CONFIG_PATH="$PREFIX/lib/pkgconfig:${PKG_CONFIG_PATH:-}"
+export CMAKE_PREFIX_PATH="$PREFIX:${CMAKE_PREFIX_PATH:-}"
+if command -v gcc-14 >/dev/null 2>&1 && command -v g++-14 >/dev/null 2>&1; then
+  export CC="${CC:-gcc-14}"
+  export CXX="${CXX:-g++-14}"
+fi
+
```
Jaruso: [P1] The pinned peer build is not reproducible with this toolchain/clone shape, and the new H3 CI job is already failing in `Build pinned ngtcp2 GnuTLS peer` before the tests run.

There are two concrete problems with the v1.25.0 pin here:

1. ngtcp2 v1.25.0 documents its C++23 examples as requiring a modern compiler such as **GCC >= 15 or Clang >= 19**, but this helper explicitly selects `gcc-14`/`g++-14` when present and the workflow installs exactly those versions.
2. The clone is not recursive and never runs `git submodule update --init`. v1.25.0 has `third-party/urlparse` as a submodule, root CMake always adds `third-party`, and the `gtlsclient` target directly consumes `$<TARGET_OBJECTS:urlparse>`. A shallow branch clone alone does not populate `third-party/urlparse/urlparse.c`.

Please make the pinned build satisfy the pinned upstream source rather than relying on the runner image accidentally filling gaps. A reasonable shape is:
```bash
git clone --depth 1 --branch "$NGTCP2_REF" \
  --recurse-submodules --shallow-submodules \
  https://github.com/ngtcp2/ngtcp2.git "$ROOT/ngtcp2"

# CI should install one supported compiler and fail clearly if absent.
if command -v clang-19 >/dev/null && command -v clang++-19 >/dev/null; then
  export CC=clang-19
  export CXX=clang++-19
elif command -v gcc-15 >/dev/null && command -v g++-15 >/dev/null; then
  export CC=gcc-15
  export CXX=g++-15
else
  echo 'ngtcp2 v1.25.0 examples require clang >= 19 or gcc >= 15' >&2
  exit 1
fi
```
For an already-created workdir, also run `git -C "$ROOT/ngtcp2" submodule update --init --depth 1` so a cached/partial checkout cannot remain broken.

Jaruso (commented):
Second full review against the current #522 criteria.

The original code/test blockers are substantially addressed: Part A now has independent peer-side resumption evidence plus exactly-once upstream dispatch; the accepted 0-RTT case remains strong; replay now compares the accepted-packet counter before/after rather than equating one request with one packet; and the PR correctly uses `Refs #522` rather than closing the issue before Part C.

I am **not finding a new QUIC/H3 resumption or 0-RTT semantic defect in this pass**. The remaining blockers are in the new CI gate itself, and both are observable on the current head:

1. The dependency audit fails because `.github/workflows/ci.yml` now contains forbidden external QUIC/TLS dependency identifiers directly in a scanned workflow surface. Keep those implementation-specific names/packages under `scripts/interop/` (which the policy explicitly excludes) and make the workflow-facing step/env/script names generic rather than weakening the audit.
2. The H3 interop job fails while building the pinned peer. The helper pins ngtcp2 v1.25.0 but forces GCC/G++ 14 even though that release documents GCC >=15 or Clang >=19 for its C++23 examples, and the shallow ngtcp2 clone does not initialize the `third-party/urlparse` submodule that the `gtlsclient` CMake target consumes.

Once those two CI plumbing issues are fixed and the actual `H3 resumption/0-RTT interop` job runs green, I expect this slice to be mergeable. Part C remains correctly deferred on #519/#520/#521 and #522 should remain open.

Jaruso (commented):
Third-pass review on head `475275fc`: **mergeable — no blocking findings remain.**

I re-reviewed the latest fix commit rather than only checking the previous comments. The two CI blockers are corrected:

- the workflow now keeps peer-specific dependency names behind the existing `scripts/interop/` audit exception and uses the generic `H3_INTEROP_CLIENT_PATH` seam;
- the pinned external peer build now uses a supported C++23 compiler and initializes the required submodules, including the cached/partial-checkout path.

The latest PR CI run is green, including the repository dependency audit and the new H3 resumption/0-RTT gate. The peer probe is also structurally sound: the pinned `gtlsclient --help` path exits successfully, so the mandatory general-profile job does not turn the four H3 cases into missing-peer skips.

I also rechecked the #522 behavior covered by the PR: authoritative independent-peer + Tardigrade-side 1-RTT resumption proof, exactly-once application observation, accepted real QUIC 0-RTT through production composition, unsafe 425-before-side-effects, replay rejection without additional accepted 0-RTT packets, and usable 1-RTT fallback. The externally unreachable negative cases are documented rather than manufactured with test-only state.

`Refs #522` is still the correct lifecycle behavior. #519, #520, and #521 remain open, so #522's Part C / final `RESUMPTION_TEST_PLAN.md` reconciliation should remain open until those siblings reach final disposition.

All prior blocking review threads are resolved. I consider PR #527 safe to merge.

PR #528, opened 2026-07-28
tls: prove persistent ticket-key restart, rotation, reload atomicity, and certificate binding (#369)

Jaruso:
## Summary

Composes the real production seams #513 landed (persistent stateless ticket-key snapshots, `ReloadableKeyRing`, SIGHUP reload) into four process-level proofs against a live, running `tardi` process — closing the "Rotation / persistent-overlap key policy" gap `docs/RESUMPTION_TEST_PLAN.md`'s Slice 3 explicitly left open, and the certificate-binding gap `restart.restart_and_credential_change.ticket_rejected`'s own doc comment flagged as blocked pending a persistent keyring.

- **`restart.persistent.ticket_survives`** — two real process incarnations sharing one on-disk snapshot file (no manual lease editing) durably reserve disjoint nonce-lease windows before issuing anything, and never emit a colliding `(key_id, nonce)` tuple. Asserted directly against the file's own `nonce_lease.start`/`end_exclusive`, not just "both processes worked".
- **`rotation.persistent.n_to_n_plus_1`** — a real SIGHUP rotates generation N → N+1: the old N ticket resumes through the overlap grace window, fresh tickets use only N+1, and a bounded real sleep past N's `decrypt_until` proves N stops resolving at all afterward (`outcome="miss"`, not a second `accepted`) — composing the live `encrypt_until`/`decrypt_until` boundary pair, not just asserting the reload succeeded.
- **`rotation.persistent.failed_reload_keeps_old_state`** — four representative rejected-reload classes (unreadable replacement, malformed JSON, an invalid nonce lease, a stale generation number) each leave a previously issued ticket resuming and the reload metrics recording a bounded, secret-free failure outcome. A ticket issued only *after* all four still seals under the original key id.
- **`rotation.persistent.certificate_binding_change`** — a restart (not a SIGHUP reload — see the finding below) proves an old ticket is rejected on auth-binding grounds specifically (`outcome="incompatible"` alongside a successful `ticket_resolve`), not an unknown-key/decryption miss, with a safe full-handshake fallback.

**Two real findings surfaced while writing this:**
- Carrying a retiring key into a new generation with its *unchanged* `nonce_lease` range collides with `ReloadableKeyRing`'s per-key-id ledger (`OverlappingNonceLease`) — the ledger permanently remembers the full declared width of every lease a key was ever installed with, not just how far its counter advanced. A retiring key needs no lease at all going forward (matches `#512`'s own unit-test convention: `nonce_lease: null` for the rotated-out key).
- The appliance TLS profile's real served credential (`ApplianceCredentials`) is loaded once at startup and is *never* touched by `hotReloadConfig` — there is no live credential-reload path under this profile at all. `rotation.persistent.certificate_binding_change` is therefore restart-based; `rotation.persistent.failed_reload_keeps_old_state` documents why it does not additionally bundle a credential change into a failing SIGHUP.

`build.zig`: adds a `rotation.` prefix to `test-integration-resumption-interop`'s filter list so these new cases get the same non-Linux CI coverage the existing `restart.`/`interop.` cases already have (macOS's CI job only runs this filtered step, not the full `test-integration` suite).

`docs/RESUMPTION_TEST_PLAN.md`: adds a Slice 4 section and annotates the now-closed Slice 3 deferred bullets.

## Test plan

- [x] `zig fmt --check .`
- [x] `zig build test -Dtls-profile=appliance` (2769/2770 pass, 1 pre-existing skip)
- [x] `zig build test-integration -Dtls-profile=appliance` (full suite, 122/134 pass, 12 skip — includes all 4 new cases)
- [x] `zig build test-integration-resumption-interop -Dtls-profile=appliance` (filtered suite, run twice — 38/42 pass, 4 skip; one unrelated flaky H2 case reproduced-then-passed on both this branch and `main`, confirmed pre-existing)
- [x] `zig build` / `zig build test-integration-resumption-interop` under the default general profile (all new cases correctly skip via `requireNativeTlsProfile`)

Closes #519

### DIFF
--- a/tests/integration.zig
+++ b/tests/integration.zig
@@ -7759,23 +7759,74 @@ test "rotation.persistent.failed_reload_keeps_old_state" {
         }
     }.run;
 
-    // Sub-case 1: unreadable replacement -- a file that exists (so
-    // `edge_config.validate`'s own `TARDIGRADE_TLS_NATIVE_TICKET_KEYS_PATH`
-    // existence check, which merely opens the path, passes) but exceeds
-    // `ticket_key_snapshot.max_snapshot_bytes` (64 KiB), so
-    // `ticket_key_snapshot.loadFromFile`'s own bounded `readFileAlloc`
-    // fails with `error.FileTooBig` -> `SnapshotTooLarge` regardless of
-    // filesystem permissions or which user/container CI runs the test as
-    // -- deterministic on every supported environment, unlike a
+    // Sub-case 1: unreadable replacement -- the configured path exists but
+    // is a directory, not a regular file, so it genuinely cannot be
+    // *opened/read* as a snapshot at all (unlike an oversized-but-otherwise-
+    // readable file, which fails a size guard instead -- a different
+    // `ticket_key_snapshot.LoadError` class, and not what "unreadable"
+    // means). Deterministic on every supported environment regardless of
+    // permissions or which user/container CI runs the test as, unlike a
     // permission-bit trick that a root/rootless-container CI job can
-    // silently fail to enforce.
+    // silently fail to enforce. The valid snapshot is moved aside and
+    // restored afterward rather than deleted, since deleting the path
+    // outright would also work but is already a distinct, separately
+    // observed condition below (this path both "exists" and "cannot be
+    // consumed", where a missing path only demonstrates the latter).
     {
-        const oversized = try allocator.alloc(u8, tls_core.ticket_key_snapshot.max_snapshot_bytes + 4096);
-        defer allocator.free(oversized);
-        @memset(oversized, 'x');
-        try writeInteropSecretFile(ticket_keys_path, oversized);
+        const backup_path = try std.fmt.allocPrint(allocator, "{s}.bak", .{ticket_keys_path});
+        defer allocator.free(backup_path);
+        try compat.cwd().rename(ticket_keys_path, compat.cwd(), backup_path);
+        // Restores the valid snapshot on every exit path, including a
+        // `MetricThresholdNotReached` bail-out below, so a failure here
+        // doesn't leave the configured path stuck as an empty directory
+        // for whatever runs next.
+        errdefer {
+            compat.cwd().deleteTree(ticket_keys_path) catch {};
+            compat.cwd().rename(backup_path, compat.cwd(), ticket_keys_path) catch {};
+        }
+        try compat.cwd().makePath(ticket_keys_path);
+
+        // On macOS/Linux as built here, `edge_config.validate`'s own
+        // `TARDIGRADE_TLS_NATIVE_TICKET_KEYS_PATH` existence check
+        // (`validateOptionalFileChecked`, an `openFile` that itself
+        // succeeds on a directory) passes, and the rejection actually
+        // happens one layer in, at `ticket_key_snapshot.loadFromFile`'s own
+        // `readFileAlloc` -- which fails to read a directory as file
+        // content -- landing on the ticket-key-specific
+        // `reload_rejected` outcome (and, since every rejection in that
+        // branch of `hotReloadConfig` also calls the shared
+        // `state.metricsRecordReloadFailure()`, the general
+        // `tardigrade_reload_failure_total` counter too, confirmed
+        // together in practice). Checking both rather than assuming either
+        // one keeps this robust against a platform/`std.Io.Dir` version
+        // where the earlier existence check itself rejects a directory
+        // instead.
+        const general_before = blk: {
+            var metrics = try sendPureZigTlsHttp1Request(allocator, tardigrade.port, "/status/metrics");
+            defer metrics.deinit();
+            break :blk prometheusMetricValue(metrics.body, "tardigrade_reload_failure_total") orelse 0;
+        };
+        const ticket_key_before = blk: {
+            var metrics = try sendPureZigTlsHttp1Request(allocator, tardigrade.port, "/status/metrics");
+            defer metrics.deinit();
+            break :blk prometheusLabeledMetricValue(metrics.body, "tardigrade_tls_ticket_key_reload_total", &.{"outcome=\"reload_rejected\""}) orelse 0;
+        };
+        tardigrade.sendSignal(std.posix.SIG.HUP);
+        const deadline = compat.milliTimestamp() + 5_000;
+        while (true) {
+            var metrics = try sendPureZigTlsHttp1Request(allocator, tardigrade.port, "/status/metrics");
+            defer metrics.deinit();
+            const general_after = prometheusMetricValue(metrics.body, "tardigrade_reload_failure_total") orelse 0;
+            const ticket_key_after = prometheusLabeledMetricValue(metrics.body, "tardigrade_tls_ticket_key_reload_total", &.{"outcome=\"reload_rejected\""}) orelse 0;
+            if (general_after >= general_before + 1 or ticket_key_after >= ticket_key_before + 1) break;
+            if (compat.milliTimestamp() >= deadline) return error.MetricThresholdNotReached;
+            compat.sleepNs(25 * std.time.ns_per_ms);
+        }
+
+        compat.cwd().deleteTree(ticket_keys_path) catch {};
+        try compat.cwd().rename(backup_path, compat.cwd(), ticket_keys_path);
     }
-    try rejectSighupAndVerify(allocator, &tardigrade, &ticket0, &clock_dummy);
+    try assertTicket0StillResumes(allocator, tardigrade.port, &ticket0, &clock_dummy);
 
     // Sub-case 2: malformed JSON.
     try writeInteropSecretFile(ticket_keys_path, "{not valid json");
--- a/tests/integration.zig
+++ b/tests/integration.zig
@@ -7195,8 +7195,9 @@ test "rotation.persistent.n_to_n_plus_1" {
         .{ .name = "TARDIGRADE_TLS_SERVER_NAME", .value = "tardigrade.test" },
         .{ .name = "TARDIGRADE_TLS_NATIVE_RESUMPTION_MODE", .value = "stateless" },
         .{ .name = "TARDIGRADE_TLS_NATIVE_TICKET_KEYS_PATH", .value = ticket_keys_path },
-        .{ .name = "TARDIGRADE_TLS_NATIVE_RESUMPTION_TICKET_LIFETIME_SECONDS", .value = "30" },
+        .{ .name = "TARDIGRADE_TLS_NATIVE_RESUMPTION_TICKET_LIFETIME_SECONDS", .value = "10" },
     };
+    const ticket_lifetime_seconds: i64 = 10;
 
     var tardigrade = try TardigradeProcess.start(allocator, .{
         .config_text = config_text,
@@ -7272,13 +7273,11 @@ test "rotation.persistent.n_to_n_plus_1" {
         },
     });
 
-    // 4: trigger the real production reload path.
+    // 4: trigger the real production reload path. A bounded poll, not a
+    // fixed settle sleep, so a slower-than-usual reload can't race a
+    // fixed-delay assertion.
     tardigrade.sendSignal(std.posix.SIG.HUP);
-    compat.sleepNs(300 * std.time.ns_per_ms);
-
-    var reload_metrics = try sendPureZigTlsHttp1Request(allocator, tardigrade.port, "/status/metrics");
-    defer reload_metrics.deinit();
-    try expectMetricAtLeast(reload_metrics.body, "tardigrade_tls_ticket_key_reload_total", &.{"outcome=\"reload_accepted\""}, 1);
+    try waitForLabeledMetricAtLeast(allocator, tardigrade.port, "tardigrade_tls_ticket_key_reload_total", &.{"outcome=\"reload_accepted\""}, 1, 5_000);
 
     var clock_dummy: u8 = 0;
     const ClientClock = struct {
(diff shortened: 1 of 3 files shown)


tests/integration.zig
```diff
@@ -6813,6 +6937,884 @@ test "restart.restart_and_credential_change.ticket_rejected" {
...
+// process, composing what `resumption_runtime.zig`'s own in-process unit
+// suite (`#512 persistent stateless runtime loads configured keys, rotates,
+// and retires old tickets`, etc.) already proves against a bare `Runtime`
+// with an injected clock, into the real production reload path with a real
+// wall clock. Also composes the `encrypt_until`/`decrypt_until` boundary
+// pair live: N's `encrypt_until` is already in the past at the moment N+1
+// is installed (so it never re-encrypts), and N's `decrypt_until` is a
+// short, deterministic few seconds out so a real (bounded) sleep can prove
+// the grace window actually closes. `not_before` and the
+// `session.issued_at + lifetime` boundary are deliberately not re-proven
```
Jaruso, line 7139: **Blocking — this explicitly leaves two required #519 boundaries unproven at the composed/live-process layer.** The issue does not ask only for primitive correctness; under **Rotation validity boundaries** it says to prove live-process behavior around `not_before`, `encrypt_until`, `decrypt_until`, and `session.issued_at + lifetime`. Existing #512 unit tests and `interop.openssl.ticket.expired` are useful lower-level/adjacent coverage, but they do not verify these boundaries through the persistent snapshot + real process/reload composition this ticket owns.

Please add live-process coverage for the two omitted boundaries (safe-margin timing is explicitly allowed by the issue), or re-scope #519 before this PR closes it. For example, a `not_before` case can reload N+1 with a short future activation time, verify fresh issuance remains on the currently eligible key before the boundary, then verify issuance switches once N+1 becomes eligible. Likewise, a short ticket lifetime can be used to issue/resume before `issued_at + lifetime` and prove fallback/rejection after it, while keeping the key's `decrypt_until` comfortably later so the failure is attributable to session lifetime rather than key retirement.

tests/integration.zig
```diff
@@ -6813,6 +6937,884 @@ test "restart.restart_and_credential_change.ticket_rejected" {
...
+// a live process running persistent ticket keys -- not an exhaustive replay
+// of every `ticket_key_snapshot.zig` malformed-input unit case (those stay
+// there), but the composed proof that a rejected candidate at each class
+// never disturbs the live keyring, config, or credential: an
+// already-issued ticket keeps resuming across every rejected attempt, a
+// ticket issued only *after* every attempt still uses the original key
+// (nothing partially applied), and the reload metric records a bounded,
+// secret-free failure outcome each time.
+//
+// This test does not additionally attempt to bundle a TLS credential change
```
Jaruso, line 7590: **Blocking — this is an explicit acceptance criterion, not an optional extension.** #519 requires: “At least one failure scenario must also change credential material and prove the failed two-phase reload leaves the **previous certificate credential** active together with the previous ticket state.” This comment correctly explains why the appliance-only harness cannot reach `NativeCredentialStore`, but that means this harness cannot satisfy this AC; it does not make the AC satisfied.

The certificate-binding restart test below proves a different property: a *successful restart* with credential B rejects A-bound tickets after successful ticket decryption. What is still missing is *failed reload atomicity across credentials + ticket state*.

I would expect this to be exercised through a general-profile/native-TLS integration path that actually wires `NativeCredentialStore`: start with credential A + ticket state A, prepare a reload containing credential B plus a candidate ticket/config change, make a later stage fail, then verify (1) the served certificate fingerprint is still A, (2) the pre-reload ticket still resumes, and (3) no candidate ticket-key/config state became visible. If that path cannot be built in this slice, #519 should be split/re-scoped rather than closed by this PR.

tests/integration.zig
```diff
@@ -6813,6 +6937,884 @@ test "restart.restart_and_credential_change.ticket_rejected" {
...
+        break :blk false;
+    } else |_| true;
+    const failure_count_before_subcase1 = blk: {
+        var metrics = try sendPureZigTlsHttp1Request(allocator, tardigrade.port, "/status/metrics");
+        defer metrics.deinit();
+        break :blk prometheusMetricValue(metrics.body, "tardigrade_reload_failure_total") orelse 0;
+    };
+    tardigrade.sendSignal(std.posix.SIG.HUP);
+    compat.sleepNs(300 * std.time.ns_per_ms);
+    if (permission_enforced) {
```
Jaruso: This conditional means the suite can pass **without testing the required “unreadable replacement” failure class at all**. On a root/container environment `chmod(000)` can still be readable; in that case the SIGHUP is sent, no failure assertion is made, permissions are restored, and the test continues green. That weakens an explicit required scenario into best-effort coverage.

Please make the unreadable candidate deterministic across supported CI environments. One approach is to atomically move the valid snapshot aside and replace the configured path with a filesystem object that cannot be consumed as the expected file (then restore it after the rejected reload), or use another harness seam that deterministically makes `open/read` fail. Whichever mechanism is chosen, the test should always assert that *this specific attempt* increments the bounded failure outcome and leaves the old ticket/keyring live; it should not silently skip the assertion.

tests/integration.zig
```diff
@@ -6813,6 +6937,884 @@ test "restart.restart_and_credential_change.ticket_rejected" {
...
+            .decrypt_until_unix_ms = baseline_now + 3_600_000,
+            .nonce_lease = .{ .prefix_hex = "11110000", .start = 100, .end_exclusive = 100 },
+        },
+    });
+    tardigrade.sendSignal(std.posix.SIG.HUP);
+    compat.sleepNs(300 * std.time.ns_per_ms);
+    {
+        var metrics = try sendPureZigTlsHttp1Request(allocator, tardigrade.port, "/status/metrics");
+        defer metrics.deinit();
+        try expectMetricAtLeast(metrics.body, "tardigrade_tls_ticket_key_reload_total", &.{"outcome=\"reload_rejected\""}, 1);
```
Jaruso: These rejection checks need to be **per attempt**, not cumulative. After malformed JSON has already incremented `outcome="reload_rejected"` to 1, this `>= 1` assertion (and the stale-generation one below) will still pass even if this SIGHUP never records a rejection. `ticket0` continuing to resume does not close that hole: a no-op/lost signal or an incorrectly handled candidate could leave the old ticket usable while this assertion remains green.

Please snapshot the labeled counter before each candidate, trigger SIGHUP, then require `after >= before + 1` (ideally with a bounded poll rather than a fixed 300 ms sleep). Repeat that for malformed JSON, invalid lease, and stale generation. That directly satisfies #519's “for each representative class assert SIGHUP reload is rejected” requirement and prevents an earlier subcase from masking failures in later ones.

Jaruso (commented):
I reviewed this against #519's required scenarios and acceptance criteria, not just the PR description. The restart persistence/lease proof, N→N+1 overlap/retirement proof, and certificate-binding restart proof are strong, and CI is green on the current head. However, I don't think this can close #519 yet because two explicit pieces of the issue are intentionally omitted, and there are two test-assertion holes in the failed-reload coverage.

**Blocking acceptance-criteria gaps**

1. **The live-process validity-boundary requirement is incomplete.** #519 explicitly requires composed live-process behavior around all four boundaries: `not_before`, `encrypt_until`, `decrypt_until`, and `session.issued_at + lifetime`. This PR exercises `encrypt_until`/`decrypt_until`, but explicitly delegates `not_before` and ticket lifetime to existing unit/interop coverage. That existing coverage may be valuable, but it is not the composed persistent-rotation/live-process proof requested by this issue. Either add live-process cases for those two boundaries or change #519's scope/AC before closing it.

2. **The failed two-phase reload credential requirement is not implemented.** #519 is unusually explicit here: at least one representative failed reload must also change credential material and prove the *previous certificate credential and previous ticket state remain active together*. The new test explains why the appliance path cannot exercise that mechanism, but documenting an architectural mismatch does not satisfy the criterion. The implementation needs a test path that reaches `NativeCredentialStore`'s prepare/commit behavior (likely a general-profile/native-TLS harness variant), or #519 needs to be split/re-scoped so this PR does not claim to close that AC.

**Failed-reload test correctness**

3. The unreadable-file subcase can silently stop proving the required class when permission enforcement is unavailable (notably root/container environments). The test still passes and moves on. A required representative failure class should be constructed deterministically instead of conditionally dropping its assertion.

4. The `reload_rejected` metric assertions for malformed JSON / semantic invalidity / stale generation use a cumulative `>= 1` check. Once the first rejected reload increments the counter, the later two assertions can pass even if their own SIGHUP did not produce a rejection. Baseline the labeled counter immediately before each SIGHUP and require `after >= before + 1` (or use an equivalent per-attempt observable). The same per-attempt pattern should be used anywhere the criterion says *for each representative class*.

A pattern I would expect for each failure case is roughly:

```zig
const before = blk: {
    var metrics = try sendPureZigTlsHttp1Request(allocator, tardigrade.port, "/status/metrics");
    defer metrics.deinit();
    break :blk prometheusLabeledMetricValue(
        metrics.body,
        "tardigrade_tls_ticket_key_reload_total",
        &.{"outcome=\"reload_rejected\""},
    ) orelse 0;
};

tardigrade.sendSignal(std.posix.SIG.HUP);
try waitForMetricAtLeast(..., before + 1); // preferably poll/bounded wait, not fixed sleep
try assertTicket0StillResumes(...);
```

For the credential-atomicity requirement, the important shape is: candidate reload contains both a ticket-key/config change and credential B; force a later prepare/validation step to fail; then prove the live endpoint still serves credential A **and** the pre-reload ticket still resumes. That is materially different from restarting under credential B, which is already well-covered by the certificate-binding test.

Once those are addressed, the rest of this slice looks directionally solid and the certificate-binding test in particular does a good job distinguishing `incompatible` from a decrypt/key miss.

**Review disposition: changes needed before merge / before closing #519.** GitHub does not allow this account to submit a formal REQUEST_CHANGES review on its own PR, so this is posted as a COMMENT review with the blocking findings inline.

tests/integration.zig
```diff
@@ -6591,226 +6715,1444 @@ test "restart.persistent.early_startup_quarantine" {
...
+            proc.sendSignal(std.posix.SIG.HUP);
+            try waitForLabeledMetricAtLeast(alloc, proc.port, "tardigrade_tls_ticket_key_reload_total", &.{"outcome=\"reload_rejected\""}, before + 1, 5_000);
+            try assertTicket0StillResumes(alloc, proc.port, ticket, now_ctx);
+        }
+    }.run;
+
+    // Sub-case 1: unreadable replacement -- a file that exists (so
+    // `edge_config.validate`'s own `TARDIGRADE_TLS_NATIVE_TICKET_KEYS_PATH`
+    // existence check, which merely opens the path, passes) but exceeds
+    // `ticket_key_snapshot.max_snapshot_bytes` (64 KiB), so
```
Jaruso: **Still blocking: this no longer tests the `unreadable replacement` class required by #519.** This candidate is a perfectly readable regular file; `loadFromFile` successfully opens/reads it until the bounded read rejects it as `SnapshotTooLarge`. That is deterministic and is a good failure-mode test, but it is semantically different from the issue's explicit representative class (`unreadable replacement`). Renaming the comment does not close that AC.

Please preserve a deterministic *read/open failure* instead of substituting a size-limit failure. One portable shape for this harness is:

```zig
const backup = try std.fmt.allocPrint(allocator, "{s}.bak", .{ticket_keys_path});
defer allocator.free(backup);
try compat.cwd().rename(ticket_keys_path, backup);
try compat.cwd().makeDir(ticket_keys_path); // configured path now exists but cannot be consumed as a regular snapshot file

defer {
    compat.cwd().deleteDir(ticket_keys_path) catch {};
    compat.cwd().rename(backup, ticket_keys_path) catch {};
}

const before = /* tardigrade_reload_failure_total, because validation/open may reject before ticket-key parsing */;
tardigrade.sendSignal(std.posix.SIG.HUP);
try waitForMetricAtLeast(..., before + 1, ...);
try assertTicket0StillResumes(...);
```

The exact filesystem construction can differ, but the important point is that the candidate must fail because it cannot be opened/read as the configured snapshot, not because valid file I/O reaches a different parser/size guard. Keep the per-attempt metric delta and old-ticket-resumes assertions you added; those are now correct.

Jaruso (commented):
Second-pass review against #519 and against all four findings from my previous review.

Three of the four previous blockers are now substantively fixed:

1. **Validity boundaries:** the updated `rotation.persistent.n_to_n_plus_1` now composes both missing boundaries at the live-process layer. The ticket-lifetime phase keeps N+1 decrypt-capable while proving pre-expiry resumption and post-expiry fallback, then confirms fresh issuance still uses N+1; the N+2 phase exercises a real future `not_before` handoff and observes the issued key id on both sides of the boundary. This addresses my earlier scope objection.

2. **Failed credential + ticket-state atomicity:** `rotation.persistent.quic_credential_reload_atomicity` is the right architectural correction. It switches to the general profile where `NativeCredentialStore` actually exists, uses real QUIC/H3, prepares credential B before deliberately failing the ticket-key stage, and proves the old session still resumes. I traced `hotReloadConfig`: credential preparation happens before `loadPersistentTicketKeysFromFile`, while `commitPreparedReload` happens only afterward, so the test is exercising the intended two-phase ordering rather than a synthetic approximation.

3. **Per-attempt rejection assertions:** `rejectSighupAndVerify` now snapshots `reload_rejected`, waits for `before + 1`, and then proves the old ticket still resumes. That closes the cumulative-counter false-positive I called out previously, and the bounded poll is stronger than the former fixed sleep.

There is **one remaining blocking acceptance-criteria mismatch**: the required `unreadable replacement` representative failure is no longer actually tested. The new first sub-case writes a normal readable file larger than `max_snapshot_bytes`; it deterministically exercises `SnapshotTooLarge`, which is useful, but it is a different failure class than #519's explicit `unreadable replacement`. The PR/docs still describe this as the unreadable case, so the implementation and acceptance criterion are currently out of sync. I left an inline comment with a deterministic way to preserve the intended class.

I re-scanned the new phases for timing attribution and false-positive paths. Aside from that remaining failure-class mismatch, I don't see another blocker in the updated implementation. CI on head `328611b` is still running at the time of this review; formatting, lint, builds, general unit tests, and several other jobs are already green, while the appliance integration jobs are still in progress.

Jaruso (commented):
Re-review of head `94fb69861c7ba2ccb52262b149f2afcbdc11fc4f`: I no longer see a blocking issue.

The final outstanding #519 gap is now addressed correctly. `rotation.persistent.failed_reload_keeps_old_state` now constructs a deterministic unreadable/non-consumable replacement by moving the valid snapshot aside and replacing the configured path with a directory. The test observes a per-attempt failure metric, restores the original snapshot on both success and error paths, and then proves the pre-reload ticket still resumes. This now exercises the intended unreadable/open-read failure class rather than substituting `SnapshotTooLarge`.

I also re-checked the fixes from the prior pass:
- live-process `not_before`, `encrypt_until`, `decrypt_until`, and `session.issued_at + lifetime` coverage is present;
- persistent restart lease reservation/resumption remains structurally proven;
- failed reload cases use per-attempt metric deltas and preserve live ticket state;
- the general-profile QUIC test reaches the real `NativeCredentialStore` prepare/commit ordering and verifies failed ticket-key reload prevents the prepared credential swap from becoming active;
- certificate-binding rejection remains distinguished from key/decryption failure with safe full-handshake fallback;
- the current CI run is fully green.

I resolved my previous review threads. GitHub does not permit this account to formally approve its own PR, but from this review I consider PR #528 ready to merge and sufficient to close #519.

PR #529, opened 2026-07-28
tls: multi-process persistent-key nonce safety and replay-locality soak (#369)

Jaruso:
## Summary

Closes the remaining **multi-process** persistent-key assurance and **process-local replay-scope** rows from #369's resumption/0-RTT validation matrix, using the real production ticket-key-snapshot/`ReloadableKeyRing`/SIGHUP-reload composition landed by #513/#519 — not a test-only substitute.

- **Harness**: split `TardigradeProcess.start()` into `spawn()` (creates the child, doesn't wait) and `waitReady()`. `start()` stays a thin wrapper for every existing caller. This lets two processes be spawned back-to-back and genuinely race the shared snapshot's `${path}.lock` before either is waited on — calling `start()` twice only ever proves *sequential* reservation.
- **`soak.persistent.multi_process_nonce_safety`** — two real processes concurrently reserve from one persistent snapshot (asserting the file's lease advances exactly once per successful process, and every issued ticket falls inside the correct disjoint reserved window), then drive several rounds of concurrent SIGHUP rotation, proving no duplicate `(key_id, nonce)` tuple is ever emitted across the whole run and that in-flight tickets survive each reload on both processes. The concurrent-startup phase, and every rotation round, force genuine lock contention (not just probable overlap): the test itself holds `${path}.lock` before spawning/signaling, releasing it only after observing (via a secret-free log marker, counted per occurrence) that both processes have actually reached the reservation attempt. Heavy mode runs continuous churn traffic across each round (each worker's samples proven, via bounded atomic flags, to span both the outgoing and incoming key) *and* proves real worker-side traffic overlaps the reload itself: `edge_gateway`'s single-threaded event loop blocks *new* accepts while `hotReloadConfig` holds the reservation lock, but a connection already dispatched to its separate `WorkerPool` beforehand keeps making progress regardless — a `GatedUpstream` proves exactly that, holding one real request per process in flight (dispatched before the lock is even acquired) and requiring its completion before the lock is released.
- **`soak.persistent.nonce_lease_exhaustion`** — kept separate so a failure says exactly what broke. Drives one process's deliberately tiny reserved lease to exhaustion via bounded batches (before/after metric deltas, not a per-connection poll) and proves `NonceLeaseExhausted` fails closed safely: connections stay healthy, previously issued tickets still resume, ordinary handshakes still work, and the other process's own disjoint reservation is untouched.
- **`soak.replay.process_local_scope`** — proves the `process_local` early-data replay scope documented in `docs/OBSERVABILITY.md` live: the exact same ticket identity is legitimately accepted once by *each* of two independent processes sharing one persistent snapshot (after paying the real ~60s startup quarantine once), while a same-process replay is still rejected as a duplicate and doesn't invalidate the underlying session. Heavy mode asserts exact per-identity accepted/duplicate metric deltas, not cumulative thresholds.
- `TARDIGRADE_SOAK_HEAVY=1` scales rotation-round/ticket-sample/replay-identity counts for the existing scheduled heavy-soak workflow, the same convention `soak.reconnect_resumption` already used.
- Docs: `docs/RESUMPTION_TEST_PLAN.md` gets a new Slice 5 section describing these cases and updates Slice 4's deferred note; `docs/OBSERVABILITY.md` gets a pointer to the new live proof (no change to the documented guarantee itself). CI's cache-leak check (`ci.yml` and `resumption-soak.yml`) also now covers `interop-*-ticket-keys.json` as defense in depth.

**Production code change**: a small, secret-free observability seam was added to make the multi-process startup-contention test deterministic rather than timing-based. `resumption_runtime.Observer` gained a new `onTicketKeyReservationAttemptFn` callback, fired immediately before `loadPersistentTicketKeysFromFile` calls `reserveNonceLeasesInFile` (whose first action is acquiring `${path}.lock`); `edge_gateway.zig` wires it to a single fixed, secret-free log line (no key material, no path). The test polls each spawned-but-not-yet-waited-on process's own log file for that marker before releasing a barrier lock it holds, forcing both processes' real reservation attempts to overlap instead of merely being likely to. This is the only production code touched — no locking/keyring/reload/replay bug was found.

## Test plan
- [x] `zig build test -Dtls-profile=appliance` (full unit suite)
- [x] `zig build test-integration-resumption-interop -Dtls-profile=appliance` (41/46 pass, 5 skipped — external QUIC/H3 interop needing a built ngtcp2 peer binary not present in this environment)
- [x] Each new case run individually in both default (smoke) and `TARDIGRADE_SOAK_HEAVY=1` (heavy) mode

Closes #520

### DIFF
--- a/src/edge_gateway.zig
+++ b/src/edge_gateway.zig
@@ -283,6 +283,18 @@ pub fn run(cfg: *const edge_config.EdgeConfig) !void {
         };
         if (native_resumption_runtime) |*rt| rt.setObserver(nativeResumptionMetricsObserver(&state));
         if (cfg.tls_native_ticket_keys_path.len > 0) {
+            // #520: a secret-free, startup-only observability line (no key
+            // material, no path, just a fixed marker) logged immediately
+            // before entering the persistent ticket-key load -- which
+            // itself calls into `ticket_key_snapshot.reserveNonceLeasesInFile`,
+            // whose very first action is acquiring `${path}.lock`. This is
+            // the only observable signal (short of instrumenting the lock
+            // acquisition itself) that a process has reached this point,
+            // which `tests/integration.zig`'s multi-process startup-
+            // reservation soak polls for on both processes before
+            // releasing a barrier lock it holds, so their real reservation
+            // attempts are forced to overlap rather than merely likely to.
+            state.logger.info(null, "native TLS/QUIC persistent ticket-key reservation attempt starting", .{});
             native_resumption_runtime.?.loadPersistentTicketKeysFromFile(cfg.tls_native_ticket_keys_path) catch |err| {
                 state.metrics_mutex.lock();
                 state.metrics.recordTicketKeyReload(.initial_load_failure);
--- a/src/edge_gateway.zig
+++ b/src/edge_gateway.zig
@@ -283,18 +283,6 @@ pub fn run(cfg: *const edge_config.EdgeConfig) !void {
         };
         if (native_resumption_runtime) |*rt| rt.setObserver(nativeResumptionMetricsObserver(&state));
         if (cfg.tls_native_ticket_keys_path.len > 0) {
-            // #520: a secret-free, startup-only observability line (no key
-            // material, no path, just a fixed marker) logged immediately
-            // before entering the persistent ticket-key load -- which
-            // itself calls into `ticket_key_snapshot.reserveNonceLeasesInFile`,
-            // whose very first action is acquiring `${path}.lock`. This is
-            // the only observable signal (short of instrumenting the lock
-            // acquisition itself) that a process has reached this point,
-            // which `tests/integration.zig`'s multi-process startup-
-            // reservation soak polls for on both processes before
-            // releasing a barrier lock it holds, so their real reservation
-            // attempts are forced to overlap rather than merely likely to.
-            state.logger.info(null, "native TLS/QUIC persistent ticket-key reservation attempt starting", .{});
             native_resumption_runtime.?.loadPersistentTicketKeysFromFile(cfg.tls_native_ticket_keys_path) catch |err| {
                 state.metrics_mutex.lock();
                 state.metrics.recordTicketKeyReload(.initial_load_failure);
@@ -873,7 +861,21 @@ fn nativeResumptionMetricsObserver(state: *GatewayState) tls_core.resumption_run
         .onTicketResolveFn = nativeResumptionTicketResolve,
         .onResumptionAttemptFn = nativeResumptionAttempt,
         .onResumptionOutcomeFn = nativeResumptionOutcome,
+        .onTicketKeyReservationAttemptFn = nativeResumptionTicketKeyReservationAttempt,
     };
+}
+
+// #520: a secret-free, fixed marker (no key material, no path) logged at
+// the exact instant `resumption_runtime.Runtime.loadPersistentTicketKeysFromFile`
+// is about to call `ticket_key_snapshot.reserveNonceLeasesInFile` -- whose
+// own first action is acquiring `${path}.lock`. `tests/integration.zig`'s
+// multi-process startup-reservation soak polls each process's own log
+// file for this marker before releasing a barrier lock it holds, so both
+// processes' real reservation attempts are forced to overlap rather than
+// merely likely to.
+fn nativeResumptionTicketKeyReservationAttempt(ctx: *anyopaque) void {
+    const state: *GatewayState = @ptrCast(@alignCast(ctx));
+    state.logger.info(null, "native TLS/QUIC persistent ticket-key reservation attempt starting", .{});
 }
 
 fn nativeResumptionTransport(transport: tls_core.resumption_runtime.Transport) http.metrics.ResumptionTransport {
--- a/tests/integration.zig
+++ b/tests/integration.zig
@@ -642,6 +642,154 @@ const RawTcpServer = struct {
     }
 };
 
+/// #520: a real upstream that deterministically proves a connection has
+/// been accepted, dispatched to a worker, and proxied all the way to it
+/// -- and holds it there under the test's own control -- rather than a
+/// fixed `delay_ms` sleep (the shape `UpstreamResponseSpec` already
+/// supports, used by e.g. `in-flight request completes safely across
+/// reload and new requests use new config` above, but not condition-based
+/// enough for a proof that must not depend on how long a reload happens
+/// to take). `waitEntered` confirms the request truly reached the
+/// upstream; `release` lets its response go out only once the caller
+/// chooses to, so a request can be held in flight across an arbitrarily
+/// long window (here, a real SIGHUP reload race) with no risk of the
+/// gate itself timing the proof.
+const GatedUpstream = struct {
+    allocator: std.mem.Allocator,
+    server: compat.NetServer,
+    thread: ?std.Thread,
+    stop_flag: std.atomic.Value(bool),
+    entered: std.atomic.Value(bool),
+    release_flag: std.atomic.Value(bool),
+
+    fn start(allocator: std.mem.Allocator) !GatedUpstream {
+        const server = try compat.listenTcp(test_host, 0);
+        return .{
+            .allocator = allocator,
+            .server = server,
+            .thread = null,
+            .stop_flag = std.atomic.Value(bool).init(false),
+            .entered = std.atomic.Value(bool).init(false),
+            .release_flag = std.atomic.Value(bool).init(false),
+        };
+    }
+
+    fn port(self: *const GatedUpstream) u16 {
+        return self.server.port();
+    }
+
+    fn run(self: *GatedUpstream) !void {
+        self.thread = try std.Thread.spawn(.{}, gatedUpstreamThreadMain, .{self});
+    }
+
+    fn stop(self: *GatedUpstream) void {
+        self.stop_flag.store(true, .seq_cst);
+        self.release_flag.store(true, .seq_cst);
+        wakeListener(self.port());
+        if (self.thread) |thread| thread.join();
+        self.server.deinit();
+        self.* = undefined;
+    }
+
+    /// Rearms the gate for another round: must be called before the next
+    /// request is sent, never concurrently with one already in flight.
+    fn resetGate(self: *GatedUpstream) void {
+        self.entered.store(false, .release);
+        self.release_flag.store(false, .release);
+    }
+
+    fn waitEntered(self: *const GatedUpstream, timeout_ms: u64) !void {
+        const deadline = compat.milliTimestamp() + @as(i64, @intCast(timeout_ms));
+        while (compat.milliTimestamp() < deadline) {
+            if (self.entered.load(.acquire)) return;
+            compat.sleepNs(5 * std.time.ns_per_ms);
+        }
+        return error.GatedUpstreamNotEntered;
+    }
+
+    fn release(self: *GatedUpstream) void {
+        self.release_flag.store(true, .release);
+    }
+};
+
+fn gatedUpstreamThreadMain(server: *GatedUpstream) void {
+    while (!server.stop_flag.load(.seq_cst)) {
+        var conn = server.server.accept() catch |err| {
+            if (server.stop_flag.load(.seq_cst)) return;
+            std.debug.print("gated upstream accept failed: {}\n", .{err});
+            continue;
+        };
+        defer conn.stream.close();
+
+        const req = readHttpMessage(server.allocator, conn.stream, 1024 * 1024) catch |err| {
+            if (server.stop_flag.load(.seq_cst)) return;
+            std.debug.print("gated upstream read failed: {}\n", .{err});
+            continue;
+        };
+        defer server.allocator.free(req.raw);
+        if (req.request_line.len == 0) continue;
+
+        server.entered.store(true, .release);
+        while (!server.release_flag.load(.acquire)) {
+            if (server.stop_flag.load(.seq_cst)) return;
+            compat.sleepNs(2 * std.time.ns_per_ms);
+        }
+        if (server.stop_flag.load(.seq_cst)) return;
+
+        const body = "gated-ok";
+        var out = std.array_list.Managed(u8).init(server.allocator);
+        defer out.deinit();
+        out.print("HTTP/1.1 200 OK\r\nContent-Length: {d}\r\nConnection: close\r\n\r\n{s}", .{ body.len, body }) catch continue;
+        conn.stream.writeAll(out.items) catch |err| {
+            if (server.stop_flag.load(.seq_cst)) return;
+            std.debug.print("gated upstream write failed: {}\n", .{err});
+            continue;
+        };
+    }
+}
+
+/// #520: drives one real native TLS H1 request against `port`'s `/race`
+/// route (proxied to a `GatedUpstream`) on its own thread, so the main
+/// test thread can hold a reservation lock and drive a SIGHUP race while
+/// this request sits blocked mid-flight at the gate. `done` is a plain
+/// atomic the main thread can poll (see `waitForRaceRequestComplete`)
+/// without joining the thread first -- joining would block until the
+/// gate is released, which is exactly the ordering this proof needs to
+/// control explicitly.
+const RaceRequestResult = struct {
+    done: std.atomic.Value(bool) = std.atomic.Value(bool).init(false),
+    status_ok: bool = false,
+    err: ?anyerror = null,
+};
+
+fn runTlsRaceRequest(allocator: std.mem.Allocator, port: u16, result: *RaceRequestResult) void {
+    defer result.done.store(true, .release);
+    const client = PureZigTlsClient.create(allocator, port, "http/1.1") catch |err| {
+        result.err = err;
+        return;
+    };
+    defer client.destroy();
+    client.writeAllPlain("GET /race HTTP/1.1\r\nHost: tardigrade.test\r\nConnection: close\r\n\r\n") catch |err| {
+        result.err = err;
+        return;
+    };
+    const raw = client.readPlainToEnd(allocator, 64 * 1024, 10_000) catch |err| {
+        result.err = err;
+        return;
+    };
+    defer allocator.free(raw);
+    result.status_ok = containsSubstring(raw, "HTTP/1.1 200 OK");
+}
+
+fn waitForRaceRequestComplete(result: *const RaceRequestResult, timeout_ms: u64) !void {
+    const deadline = compat.milliTimestamp() + @as(i64, @intCast(timeout_ms));
+    while (compat.milliTimestamp() < deadline) {
+        if (result.done.load(.acquire)) return;
+        compat.sleepNs(5 * std.time.ns_per_ms);
+    }
+    return error.RaceRequestNotObserved;
+}
+
 const StartTlsSmtpProcess = struct {
     allocator: std.mem.Allocator,
     child: std.process.Child,
@@ -6571,6 +6719,20 @@ const ChurnWorker = struct {
     fn run(self: *ChurnWorker) void {
         while (!self.stop.load(.acquire)) {
             const ids = issueTicketIds(std.heap.page_allocator, self.port) catch |err| {
+                // `WouldBlock` is transient local contention -- this
+                // worker's own tight reconnect loop briefly outpacing the
+                // single dispatched worker thread `TARDIGRADE_WORKER_
+                // THREADS=1` provides (e.g. while the gated race request
+                // above is deliberately occupying it), not a server-side
+                // correctness failure. Retrying past it (with a short
+                // pause so the retry doesn't just recreate the same
+                // contention) keeps the proof honest without papering
+                // over a real issuance/connection failure, which remains
+                // fatal exactly as before.
+                if (err == error.WouldBlock) {
+                    compat.sleepNs(2 * std.time.ns_per_ms);
+                    continue;
+                }
                 self.failure = err;
                 return;
             };
@@ -8852,22 +9014,58 @@ test "soak.persistent.multi_process_nonce_safety" {
         },
     });
 
-    const config_text =
-        \\location = /healthz {
+    // One gated upstream per process, proxied to from a `/race` route
+    // declared in that process's own config -- the heavy tier's real
+    // worker-overlap proof below drives one request through each of
+    // these, per round, and holds it there under the test's own control
+    // while the reservation lock is held. Created (and their `/race`
+    // route wired) regardless of tier, since the process's own config
+    // is fixed at startup and can't be changed per-round; only the
+    // smoke tier simply never drives traffic through it.
+    var gate_a = try GatedUpstream.start(allocator);
+    defer gate_a.stop();
+    try gate_a.run();
+    var gate_b = try GatedUpstream.start(allocator);
+    defer gate_b.stop();
+    try gate_b.run();
+
+    const config_text_a = try std.fmt.allocPrint(allocator,
+        \\location = /healthz {{
         \\    return 200 alive;
-        \\}
-    ;
-    const process_options = TardigradeOptions{
-        .config_text = config_text,
+        \\}}
+        \\location = /race {{
+        \\    proxy_pass http://{s}:{d};
+        \\}}
+    , .{ test_host, gate_a.port() });
+    defer allocator.free(config_text_a);
+    const config_text_b = try std.fmt.allocPrint(allocator,
+        \\location = /healthz {{
+        \\    return 200 alive;
+        \\}}
+        \\location = /race {{
+        \\    proxy_pass http://{s}:{d};
+        \\}}
+    , .{ test_host, gate_b.port() });
+    defer allocator.free(config_text_b);
+
+    const shared_extra_env = &[_]EnvPair{
+        .{ .name = "TARDIGRADE_TLS_CERT_PATH", .value = tls_paths.cert_path },
+        .{ .name = "TARDIGRADE_TLS_KEY_PATH", .value = tls_paths.key_path },
+        .{ .name = "TARDIGRADE_TLS_SERVER_NAME", .value = "tardigrade.test" },
+        .{ .name = "TARDIGRADE_TLS_NATIVE_RESUMPTION_MODE", .value = "stateless" },
+        .{ .name = "TARDIGRADE_TLS_NATIVE_TICKET_KEYS_PATH", .value = ticket_keys_path },
+    };
+    const process_options_a = TardigradeOptions{
+        .config_text = config_text_a,
         .ready_https_insecure = true,
         .ready_path = "/healthz",
-        .extra_env = &.{
-            .{ .name = "TARDIGRADE_TLS_CERT_PATH", .value = tls_paths.cert_path },
-            .{ .name = "TARDIGRADE_TLS_KEY_PATH", .value = tls_paths.key_path },
-            .{ .name = "TARDIGRADE_TLS_SERVER_NAME", .value = "tardigrade.test" },
-            .{ .name = "TARDIGRADE_TLS_NATIVE_RESUMPTION_MODE", .value = "stateless" },
-            .{ .name = "TARDIGRADE_TLS_NATIVE_TICKET_KEYS_PATH", .value = ticket_keys_path },
-        },
+        .extra_env = shared_extra_env,
+    };
+    const process_options_b = TardigradeOptions{
+        .config_text = config_text_b,
+        .ready_https_insecure = true,
+        .ready_path = "/healthz",
+        .extra_env = shared_extra_env,
     };
 
     // Phase 1: *forced*, not merely likely, concurrent startup reservation.
@@ -8897,17 +9095,17 @@ test "soak.persistent.multi_process_nonce_safety" {
         .lock = .exclusive,
     });
 
-    var process_a = try TardigradeProcess.spawn(allocator, process_options);
+    var process_a = try TardigradeProcess.spawn(allocator, process_options_a);
     defer process_a.stop();
-    var process_b = try TardigradeProcess.spawn(allocator, process_options);
+    var process_b = try TardigradeProcess.spawn(allocator, process_options_b);
     defer process_b.stop();
 
     try waitForReservationAttemptCount(allocator, process_a.log_path, 1, 5_000);
     try waitForReservationAttemptCount(allocator, process_b.log_path, 1, 5_000);
     startup_barrier.close();
 
-    try process_a.waitReady(process_options);
-    try process_b.waitReady(process_options);
+    try process_a.waitReady(process_options_a);
+    try process_b.waitReady(process_options_b);
 
     const after_both_start = try readTicketKeyLeaseWindow(allocator, ticket_keys_path, key_ids[0]);
     try std.testing.expectEqual(@as(u64, 1), after_both_start.generation);
@@ -9055,6 +9253,47 @@ test "soak.persistent.multi_process_nonce_safety" {
             try waitForChurnGenerationSeen(&churn_b.?.saw_outgoing, 5_000);
         }
 
+        // The real worker-overlap proof #520 asks for, heavy tier only:
+        // `edge_gateway.run` has a separate `WorkerPool` alongside its
+        // single-threaded event loop, so once a native H1 connection has
+        // actually been accepted and dispatched (`advanceNativeHttp1`
+        // running on a worker, `WaitingEncryptedHttpConnection.waitFor`
+        // polling its own fd directly), it can keep making progress even
+        // while the main loop is blocked inside synchronous
+        // `hotReloadConfig` -- blocking that one thread only prevents
+        // *new* accepts, not already-dispatched work. `gate_{a,b}` proxy
+        // one real request per process to a point held entirely under
+        // this test's control, so `waitEntered` below deterministically
+        // proves each request has already been accepted, dispatched, and
+        // proxied all the way to its upstream *before* either reload
+        // begins -- a real condition, not a fixed delay.
+        //
+        // The cleanup defer releases both gates and joins any spawned
+        // race thread unconditionally, registered before the fallible
+        // spawns/waits below -- if a later `try` in this round fails
+        // (reload never reaching `reload_accepted`, say), a race thread
+        // left blocked waiting on a never-released gate must not leak
+        // past this scope, the same lesson the churn-worker lifetime fix
+        // already applied.
+        var race_result_a = RaceRequestResult{};
+        var race_result_b = RaceRequestResult{};
+        var race_thread_a: ?std.Thread = null;
+        var race_thread_b: ?std.Thread = null;
+        defer {
+            gate_a.release();
+            gate_b.release();
+            if (race_thread_a) |t| t.join();
+            if (race_thread_b) |t| t.join();
+        }
+        if (soakHeavyEnabled()) {
+            gate_a.resetGate();
+            gate_b.resetGate();
+            race_thread_a = try std.Thread.spawn(.{}, runTlsRaceRequest, .{ allocator, process_a.port, &race_result_a });
+            race_thread_b = try std.Thread.spawn(.{}, runTlsRaceRequest, .{ allocator, process_b.port, &race_result_b });
+            try gate_a.waitEntered(5_000);
+            try gate_b.waitEntered(5_000);
+        }
+
         // Forces the two real SIGHUP handlers to genuinely block on
         // `${ticket_keys_path}.lock` at the same time, exactly the same
         // technique the concurrent-startup phase above uses: this test
@@ -9085,28 +9324,36 @@ test "soak.persistent.multi_process_nonce_safety" {
         try waitForReservationAttemptCount(allocator, process_a.log_path, reservation_marker_count_before_a + 1, 5_000);
         try waitForReservationAttemptCount(allocator, process_b.log_path, reservation_marker_count_before_b + 1, 5_000);
 
-        // Deliberately does *not* require a churn worker to complete a
-        // connection while the lock above is held. `edge_gateway.run`'s
-        // main loop (`http.shutdown.consumeReloadRequested()` ->
-        // `gshutdown.hotReloadConfig`) calls `hotReloadConfig`
-        // synchronously from the same single-threaded event loop that
-        // also accepts new connections (`gaccept.acceptReadyConnections`)
-        // and dispatches ready sockets to the worker pool -- with
-        // `TARDIGRADE_WORKER_THREADS=1`, blocking that one loop thread on
-        // `${path}.lock` therefore blocks *all* socket I/O for that
-        // process, including finishing already-in-flight work, not only
-        // new accepts. An earlier revision of this test required exactly
-        // that ("wait for a churn worker to complete a connection while
-        // both reload handlers are confirmed blocked") and it reliably
-        // hung for the full bounded test timeout -- a real, reproducible
-        // deadlock proving the invariant is architecturally unobservable
-        // this way, not merely hard to arrange. The honest, achievable
-        // property instead: churn traffic keeps flowing continuously
-        // across the *whole* round (started before either SIGHUP, stopped
-        // only after both post-acceptance probes below), asserted by
-        // requiring each worker's collected samples to include both the
-        // outgoing and the incoming generation's key id.
+        // Both main event loops are now confirmed blocked in reload --
+        // but each gated request (heavy tier only) was already accepted
+        // and dispatched to a worker *before* that happened, so releasing
+        // it here and requiring its completion *before* the lock itself
+        // is released is the actual proof that real request/ticket-
+        // adjacent traffic was genuinely in flight during the reload
+        // race, not merely before or after it. `ChurnWorker`'s own
+        // connections deliberately are not required to complete here
+        // (see its doc comment): only an *already-dispatched* connection
+        // can make progress while the main loop is blocked, and a churn
+        // worker's next connection attempt has no such guarantee.
+        if (soakHeavyEnabled()) {
+            gate_a.release();
+            gate_b.release();
+            try waitForRaceRequestComplete(&race_result_a, 5_000);
+            try waitForRaceRequestComplete(&race_result_b, 5_000);
+        }
+
         reload_barrier.close();
+
+        if (soakHeavyEnabled()) {
+            race_thread_a.?.join();
+            race_thread_b.?.join();
+            race_thread_a = null;
+            race_thread_b = null;
+            if (race_result_a.err) |err| return err;
+            if (race_result_b.err) |err| return err;
+            try std.testing.expect(race_result_a.status_ok);
+            try std.testing.expect(race_result_b.status_ok);
+        }
 
         // Now that the lock is released, both processes' real
         // reservations proceed and each (eventually) reports its own
--- a/tests/integration.zig
+++ b/tests/integration.zig
@@ -6448,6 +6448,32 @@ fn currentLabeledMetric(allocator: std.mem.Allocator, port: u16, name: []const u
     return prometheusLabeledMetricValue(metrics.body, name, labels) orelse 0;
 }
 
+/// #520: the observable arrival signal for the concurrent-startup
+/// reservation barrier -- polls a spawned-but-not-yet-waited-on process's
+/// own log file (already captured from its earliest stderr output, per
+/// `TardigradeProcess.spawn`'s `early_stderr` redirect, independent of
+/// whether its HTTP listener has started serving yet) for the fixed,
+/// secret-free marker `edge_gateway.zig` logs immediately before entering
+/// `loadPersistentTicketKeysFromFile` -- whose own first action is
+/// acquiring `${path}.lock`. Waiting for this marker on *both* processes
+/// before releasing a barrier lock the test holds is a real observed-
+/// arrival condition, not a fixed delay: a fixed sleep can never rule out
+/// a slower process reaching the lock attempt after the barrier was
+/// already released, silently turning the intended race back into
+/// sequential reservation.
+fn waitForReservationAttempt(allocator: std.mem.Allocator, log_path: []const u8, timeout_ms: u64) !void {
+    const marker = "persistent ticket-key reservation attempt starting";
+    const deadline = compat.milliTimestamp() + @as(i64, @intCast(timeout_ms));
+    while (compat.milliTimestamp() < deadline) {
+        if (compat.cwd().readFileAlloc(allocator, log_path, 4 * 1024 * 1024)) |log_data| {
+            defer allocator.free(log_data);
+            if (containsSubstring(log_data, marker)) return;
+        } else |_| {}
+        compat.sleepNs(10 * std.time.ns_per_ms);
+    }
+    return error.ReservationAttemptNotObserved;
+}
+
 /// #520 heavy tier: keeps real full-handshake/ticket-issuance traffic
 /// flowing against one process for as long as `stop` is clear -- run on
 /// its own `std.Thread` so the main thread can send both SIGHUPs and poll
@@ -6457,16 +6483,38 @@ fn currentLabeledMetric(allocator: std.mem.Allocator, port: u16, name: []const u
 /// smoke tier's shape) never overlaps the race itself. Uses
 /// `std.heap.page_allocator` rather than `std.testing.allocator`: the
 /// latter is not safe to call concurrently from multiple threads, and
-/// this worker's results are only ever read by the main thread after
-/// `std.Thread.join()`, so no cross-thread synchronization on `results`
-/// itself is needed.
+/// this worker's results/failure are only ever read by the main thread
+/// after `std.Thread.join()`, so no cross-thread synchronization on them
+/// is needed.
+///
+/// Every issued ticket is validated against the round's two known key
+/// ids: it must belong to either the complete outgoing or complete
+/// incoming generation, never anything else -- and once the caller marks
+/// `require_new` (only after *this specific process* has independently
+/// reported `reload_accepted`), it must belong to the incoming generation
+/// only, proving fresh issuance never regresses to the retiring key after
+/// that process's own acceptance. A worker records the first error it
+/// hits (an issuance failure, an unexpected key id, or a stale-key-after-
+/// acceptance violation) and stops immediately, rather than silently
+/// swallowing it and leaving the caller to mistake an empty result set
+/// for a healthy run.
 const ChurnWorker = struct {
     port: u16,
     stop: *std.atomic.Value(bool),
+    outgoing_key_id_hex: []const u8,
+    incoming_key_id_hex: []const u8,
+    require_new: std.atomic.Value(bool) = std.atomic.Value(bool).init(false),
     results: std.array_list.Managed(TicketEnvelopeIds),
+    failure: ?anyerror = null,
 
-    fn init(port: u16, stop: *std.atomic.Value(bool)) ChurnWorker {
-        return .{ .port = port, .stop = stop, .results = std.array_list.Managed(TicketEnvelopeIds).init(std.heap.page_allocator) };
+    fn init(port: u16, stop: *std.atomic.Value(bool), outgoing_key_id_hex: []const u8, incoming_key_id_hex: []const u8) ChurnWorker {
+        return .{
+            .port = port,
+            .stop = stop,
+            .outgoing_key_id_hex = outgoing_key_id_hex,
+            .incoming_key_id_hex = incoming_key_id_hex,
+            .results = std.array_list.Managed(TicketEnvelopeIds).init(std.heap.page_allocator),
+        };
     }
 
     fn deinit(self: *ChurnWorker) void {
@@ -6475,8 +6523,25 @@ const ChurnWorker = struct {
 
     fn run(self: *ChurnWorker) void {
         while (!self.stop.load(.acquire)) {
-            const ids = issueTicketIds(std.heap.page_allocator, self.port) catch continue;
-            self.results.append(ids) catch {};
+            const ids = issueTicketIds(std.heap.page_allocator, self.port) catch |err| {
+                self.failure = err;
+                return;
+            };
+            const hex = std.fmt.bytesToHex(ids.key_id, .lower);
+            const is_incoming = std.mem.eql(u8, self.incoming_key_id_hex, &hex);
+            const is_outgoing = std.mem.eql(u8, self.outgoing_key_id_hex, &hex);
+            if (!is_incoming and !is_outgoing) {
+                self.failure = error.ChurnObservedUnexpectedKeyId;
+                return;
+            }
+            if (self.require_new.load(.acquire) and !is_incoming) {
+                self.failure = error.ChurnObservedStaleKeyAfterAcceptance;
+                return;
+            }
+            self.results.append(ids) catch |err| {
+                self.failure = err;
+                return;
+            };
         }
     }
 };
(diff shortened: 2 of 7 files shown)


tests/integration.zig
```diff
@@ -8391,6 +8582,739 @@ test "soak.reconnect_resumption" {
...
+    };
+
+    // Phase 1: genuinely concurrent startup reservation. `spawn()` returns
+    // as soon as each child is created, before either is waited on -- both
+    // are live and already racing `${ticket_keys_path}.lock` before this
+    // test ever blocks on readiness, unlike calling `start()` twice (which
+    // only ever proves *sequential* reservation).
+    var process_a = try TardigradeProcess.spawn(allocator, process_options);
+    defer process_a.stop();
+    var process_b = try TardigradeProcess.spawn(allocator, process_options);
```
Jaruso: **Blocking — this permits contention but does not prove it.** `spawn(A)` can return and A can complete `reserveNonceLeasesInFile()` before B reaches that code at all. In that schedule every assertion below still passes even if the lock/under-lock re-read were broken, because the two reservations happened sequentially.

The test needs an enforced overlap point so it deterministically exercises the race it claims to cover. One workable shape is a bounded test seam/barrier immediately before persistent lease reservation: start A and B, wait until both have reached the pre-reservation barrier, then release both together. Another is to have the parent hold `${path}.lock`, start both children, wait for a bounded observable "reservation pending" point in each, then release the parent lock. The key requirement is: **the test must fail if the serialization primitive is removed**, not merely make simultaneous execution likely.

For example, conceptually:
```zig
var barrier = try ReservationBarrier.init(...);
var a = try TardigradeProcess.spawn(... barrier env ...);
var b = try TardigradeProcess.spawn(... barrier env ...);
try barrier.waitUntilArrived(2, 5_000);
barrier.release();
try a.waitReady(...);
try b.waitReady(...);
```
Exact plumbing can differ, but back-to-back `spawn()` alone is not a deterministic contention proof.

tests/integration.zig
```diff
@@ -8391,6 +8582,739 @@ test "soak.reconnect_resumption" {
...
+            },
+        });
+
+        const reload_accepted_before_a = try currentLabeledMetric(allocator, process_a.port, "tardigrade_tls_ticket_key_reload_total", &.{"outcome=\"reload_accepted\""});
+        const reload_accepted_before_b = try currentLabeledMetric(allocator, process_b.port, "tardigrade_tls_ticket_key_reload_total", &.{"outcome=\"reload_accepted\""});
+
+        // Sent back-to-back, without waiting between them -- both real
+        // SIGHUP handlers race the shared snapshot lock concurrently.
+        process_a.sendSignal(std.posix.SIG.HUP);
+        process_b.sendSignal(std.posix.SIG.HUP);
```
Jaruso, line 9342: **Blocking — the heavy tier never has ticket/request traffic in flight during the reload race.** Immediately after the two signals, this test waits for both `reload_accepted` counters; all `issueTicketIds()` sampling happens afterward. That misses #520's explicit heavy-tier requirement to keep issuance/request traffic flowing while the two SIGHUP handlers race, which is the part intended to expose publication/UAF/deadlock bugs and observe only complete old-or-new generations during the transition.

Please run bounded traffic concurrently with the reload in heavy mode. A practical shape is one worker per process looping full-handshake/ticket probes under a stop flag, collecting `TicketEnvelopeIds` under a mutex, while the main thread signals A/B and waits for both reload deltas; then stop/join the workers and fold those tuples into the global uniqueness/key-id assertions.

Conceptually:
```zig
var stop = std.atomic.Value(bool).init(false);
var samples = ConcurrentTicketSamples.init(allocator);
const ta = try std.Thread.spawn(.{}, ticketChurn, .{ process_a.port, &stop, &samples });
const tb = try std.Thread.spawn(.{}, ticketChurn, .{ process_b.port, &stop, &samples });

process_a.sendSignal(std.posix.SIG.HUP);
process_b.sendSignal(std.posix.SIG.HUP);
try waitFor...A;
try waitFor...B;

stop.store(true, .release);
ta.join();
tb.join();
try assertNoDuplicateTicketIds(samples.items());
```
Also validate every tuple observed during the transition belongs to either the complete outgoing generation or the complete incoming generation; after that process reports `reload_accepted`, only the incoming key id should appear.

tests/integration.zig
```diff
@@ -8391,6 +8582,739 @@ test "soak.reconnect_resumption" {
...
+            const raw = try client.readPlainToEnd(allocator, 64 * 1024, 5_000);
+            defer allocator.free(raw);
+            try assertContains(raw, "HTTP/1.1 200 OK");
+            try assertContains(raw, "a-early");
+            try std.testing.expectEqual(@as(u32, 1), upstream_a.requestCount());
+        }
+        {
+            var metrics = try sendPureZigTlsHttp1Request(allocator, process_a.port, "/status/metrics");
+            defer metrics.deinit();
+            try expectMetricAtLeast(metrics.body, "tardigrade_tls_early_data_replay_total", &.{"outcome=\"accepted\""}, 1);
```
Jaruso: **Blocking acceptance gap in heavy mode:** this is a cumulative `>= 1` assertion. With `identity_count = 4`, once identity 0 increments `accepted`, identities 1–3 can stop emitting `accepted` entirely and this still passes; the duplicate assertions below have the same problem. #520 specifically calls for independent per-process **metric deltas** for each replay identity.

Capture baselines at the start of every identity iteration and assert exact increments after each action, for both A and B. For example:
```zig
const a_accept_before = try currentLabeledMetric(...);
const a_dup_before = try currentLabeledMetric(...);
const b_accept_before = try currentLabeledMetric(...);
const b_dup_before = try currentLabeledMetric(...);

// A first use
try std.testing.expectEqual(a_accept_before + 1, try currentLabeledMetric(...));
// A replay
try std.testing.expectEqual(a_dup_before + 1, try currentLabeledMetric(...));
// same for B
```
`/status/metrics` handshakes do not themselves create replay-store accepted/duplicate events, so these replay-counter deltas are the correct invariant even though metrics reads may consume ticket nonces.

Jaruso (commented):
I found three **merge-blocking gaps** against #520's acceptance criteria. The overall structure is good: the harness split is contained, the nonce-exhaustion case exercises the intended best-effort failure semantics, docs/CI wiring are scoped correctly, and the current CI run is green. However, the two concurrency guarantees are not yet proven strongly enough, and the heavy replay metric assertions can pass after the first identity even if later identities stop emitting events.

Please address the inline comments before merge. In particular, the multi-process soak needs to fail deterministically (not probabilistically) if shared-snapshot serialization is broken, and the heavy rotation tier needs actual request/ticket activity overlapping the SIGHUP publication window rather than sampling only after both reloads finish.

Note: GitHub would not allow this account to submit a formal `REQUEST_CHANGES` review because the authenticated account is the PR author, so this is posted as a comment review; I consider the findings blocking.

tests/integration.zig
Jaruso: Fixed in 60ee6eb1: the test now acquires `${ticket_keys_path}.lock` itself (the same `createFile(..., .lock = .exclusive)` call `acquireReservationLock` uses) before spawning either process, holds it for a bounded 500ms window, then releases -- forcing both children to be genuinely blocked inside the same lock acquisition the instant it's released, rather than merely hoping the scheduler interleaves two back-to-back `spawn()` calls.

tests/integration.zig
Jaruso: Fixed in 60ee6eb1: added a `ChurnWorker` (tests/integration.zig) that runs on its own `std.Thread` per process in heavy mode, issuing real full-handshake/ticket connections continuously while the two SIGHUP signals are sent and `reload_accepted` is polled for. Every tuple collected during that window is asserted to belong to either the complete outgoing or complete incoming generation's key id, and folded into the global (key_id, nonce) uniqueness check. Verified heavy mode now collects hundreds of extra tuples during the race (448 -> 1085 total tuples across 6 rounds) with all assertions still passing.

tests/integration.zig
Jaruso: Fixed in 60ee6eb1: each identity iteration now captures accepted/duplicate baselines for both A and B at the top of the loop and asserts exact +1 deltas after each of the four actions, replacing the cumulative >= 1 checks. Re-verified with TARDIGRADE_SOAK_HEAVY=1 (4 identities) passing.

Jaruso (commented):
Second-pass review against #520 and the first review's three blockers.

The replay-metric fix is correct: every identity now takes fresh per-process accepted/duplicate baselines and asserts exact +1 deltas. CI for the new head is also green.

I still do **not** think this is ready to merge. The startup-contention fix retains a timing assumption rather than an observed barrier, and the new threaded churn harness introduces failure-path/lifetime problems that can mask or turn a real reload failure into a hanging/UAF test. There is also still a gap in the process-specific post-`reload_accepted` assertion from the prior review: churn samples are accepted as old-or-new even after that individual process has reported acceptance.

I left four inline blockers with concrete fixes. Because this connected account owns the PR, GitHub will not permit a formal `REQUEST_CHANGES`; please treat this COMMENT review as changes-required.

tests/integration.zig
```diff
@@ -8391,6 +8615,809 @@ test "soak.reconnect_resumption" {
...
+        .permissions = .fromMode(0o600),
+        .lock = .exclusive,
+    });
+
+    var process_a = try TardigradeProcess.spawn(allocator, process_options);
+    defer process_a.stop();
+    var process_b = try TardigradeProcess.spawn(allocator, process_options);
+    defer process_b.stop();
+
+    compat.sleepNs(500 * std.time.ns_per_ms);
```
Jaruso: **Still blocking from the first review:** holding the lock is the right mechanism, but the `500 ms` sleep is still not an arrival barrier. On a loaded runner, process B can take >500 ms to reach `reserveNonceLeasesInFile()`. The parent then releases the lock, A reserves, and B can arrive later and reserve sequentially; the exact same false-positive schedule from the first review is still possible. The comment above saying this “guarantees, by construction” that both children are already blocked is therefore stronger than what the code proves.

Please release the parent lock only after **observable per-child evidence** that both children have reached the reservation/lock acquisition point, with a bounded timeout. A fixed delay cannot establish that property. If the current production path has no observable pre-lock point, add a secret-free debug/observer seam around the reservation attempt and have the integration harness wait for both process-specific markers before releasing the parent lock. Conceptually:
```zig
var barrier = try holdReservationLock(ticket_keys_path);
defer barrier.close();

var a = try TardigradeProcess.spawn(...);
var b = try TardigradeProcess.spawn(...);
try waitForReservationAttempt(a, 5_000);
try waitForReservationAttempt(b, 5_000);
barrier.close();
```
The important distinction is **condition-based arrival of both children**, not making the sleep longer.

tests/integration.zig
```diff
@@ -6257,6 +6272,215 @@ fn extractTicketEnvelopeIds(ticket: *const tls_core.session.ClientTicketState) T
...
+        return .{ .port = port, .stop = stop, .results = std.array_list.Managed(TicketEnvelopeIds).init(std.heap.page_allocator) };
+    }
+
+    fn deinit(self: *ChurnWorker) void {
+        self.results.deinit();
+    }
+
+    fn run(self: *ChurnWorker) void {
+        while (!self.stop.load(.acquire)) {
+            const ids = issueTicketIds(std.heap.page_allocator, self.port) catch continue;
```
Jaruso: **Blocking — the churn worker currently hides the exact failures this soak is supposed to detect.** Both `issueTicketIds(...) catch continue` and `results.append(...) catch {}` silently discard errors. A reload/publication regression that makes every in-race handshake fail could therefore leave this worker with zero samples and the test can still continue to the post-reload probes and pass.

Record the first worker error (or at least a failure count), stop that worker, and assert after `join()` that it had no failures and produced at least one sample. For example:
```zig
const ChurnWorker = struct {
    ...
    failure: ?anyerror = null,

    fn run(self: *ChurnWorker) void {
        while (!self.stop.load(.acquire)) {
            const ids = issueTicketIds(std.heap.page_allocator, self.port) catch |err| {
                self.failure = err;
                return;
            };
            self.results.append(ids) catch |err| {
                self.failure = err;
                return;
            };
        }
    }
};

// after join
if (churn_a.failure) |err| return err;
try std.testing.expect(churn_a.results.items.len > 0);
```
The heavy tier needs to prove traffic stayed healthy *during* publication, not merely ignore failed attempts until publication finishes.

tests/integration.zig
```diff
@@ -8391,6 +8615,809 @@ test "soak.reconnect_resumption" {
...
+            churn_thread_a = try std.Thread.spawn(.{}, ChurnWorker.run, .{&churn_a.?});
+            churn_thread_b = try std.Thread.spawn(.{}, ChurnWorker.run, .{&churn_b.?});
+        }
+
+        // Sent back-to-back, without waiting between them -- both real
+        // SIGHUP handlers race the shared snapshot lock concurrently.
+        process_a.sendSignal(std.posix.SIG.HUP);
+        process_b.sendSignal(std.posix.SIG.HUP);
+
+        try waitForLabeledMetricAtLeast(allocator, process_a.port, "tardigrade_tls_ticket_key_reload_total", &.{"outcome=\"reload_accepted\""}, reload_accepted_before_a + 1, 5_000);
```
Jaruso, line 9389: **Blocking test-harness lifetime bug:** these `try` calls can unwind while both churn threads are still running. The cleanup (`stop.store`, `join`, `deinit`) only occurs below on the success path. If reload never reaches `reload_accepted` — exactly the failure this soak wants to diagnose — the test returns, the `process_*` defers kill the children, and each churn thread continues looping because its stop flag is still false (`issueTicketIds` errors are currently swallowed). The threads also retain pointers to stack-owned `churn_stop`/`ChurnWorker` values after this scope unwinds, so this can become a hang/UAF instead of a bounded test failure. The same leak occurs if spawning thread B fails after thread A was started.

Register cleanup **before any fallible operation after the first worker/thread is created**. A small owner object is cleanest, but a scope defer with nullable threads also works:
```zig
var churn_stop = std.atomic.Value(bool).init(false);
var churn_a: ?ChurnWorker = null;
var churn_b: ?ChurnWorker = null;
var ta: ?std.Thread = null;
var tb: ?std.Thread = null;
defer {
    churn_stop.store(true, .release);
    if (ta) |t| t.join();
    if (tb) |t| t.join();
    if (churn_a) |*w| w.deinit();
    if (churn_b) |*w| w.deinit();
}
// now spawn and perform fallible waits
```
Then remove the success-only joins/deinits (or make the owner idempotent). This is important to #520's explicit bounded/no-hang acceptance criterion.

tests/integration.zig
```diff
@@ -8391,6 +8615,809 @@ test "soak.reconnect_resumption" {
...
+        try waitForLabeledMetricAtLeast(allocator, process_b.port, "tardigrade_tls_ticket_key_reload_total", &.{"outcome=\"reload_accepted\""}, reload_accepted_before_b + 1, 5_000);
```
Jaruso, line 9394: **The previous review's post-acceptance invariant is still not actually checked per process.** The main thread observes A's `reload_accepted` first, then waits for B. During that second wait, A's churn worker keeps issuing tickets. Later, lines below accept *every* A churn tuple as `old OR new`, so A could regress to / continue issuing the retiring key **after A has reported `reload_accepted`** and this test would pass. The final post-both-reloads samples only prove that A eventually settles on the new key.

Please mark each worker as `must_use_new` immediately after that process's own acceptance is observed, and have subsequent worker samples reject the outgoing key. For example:
```zig
try waitFor...A;
churn_a.require_new.store(true, .release);

try waitFor...B;
churn_b.require_new.store(true, .release);
```
with the worker validating `ids.key_id == incoming_key_id` whenever `require_new` is set (recording a failure rather than silently dropping it). That preserves `old OR new` during the actual transition while proving the required monotonic boundary after each individual process reports acceptance.

tests/integration.zig
Jaruso: Fixed in a59655bd: added a secret-free, startup-only log line in edge_gateway.zig (right before entering loadPersistentTicketKeysFromFile, whose first action is acquiring ${path}.lock) and a new waitForReservationAttempt() test helper that polls each child's own log file for that marker with a bounded timeout. The barrier lock is now released only after observing both children's real arrival, not after a fixed sleep.

tests/integration.zig
Jaruso: Fixed in a59655bd: ChurnWorker now records the first error (issuance failure, unexpected key id, or stale-key-after-acceptance) in a `failure: ?anyerror` field and stops immediately instead of swallowing it. The caller asserts `failure == null` and `results.items.len > 0` for both workers after join().

tests/integration.zig
Jaruso: Fixed in a59655bd: cleanup (stop-store/join/deinit) is now registered via `defer` immediately after declaring the nullable churn_stop/churn_a/churn_b/churn_thread_a/churn_thread_b state, before any fallible operation (including the std.Thread.spawn calls and the waitForLabeledMetricAtLeast polls). The success path explicitly joins and nulls the thread handles so the deferred joins become harmless no-ops rather than an illegal double-join.

tests/integration.zig
Jaruso: Fixed in a59655bd: ChurnWorker now takes both outgoing_key_id_hex and incoming_key_id_hex at construction and exposes a require_new atomic flag. The main thread sets it on churn_a immediately after observing A's own reload_accepted, and on churn_b immediately after observing B's own reload_accepted (independently). The worker validates every ticket against outgoing-or-incoming unconditionally, and against incoming-only once require_new is set, recording a failure instead of silently accepting a post-acceptance regression.

src/edge_gateway.zig
```diff
@@ -283,6 +283,18 @@ pub fn run(cfg: *const edge_config.EdgeConfig) !void {
...
+            // before entering the persistent ticket-key load -- which
+            // itself calls into `ticket_key_snapshot.reserveNonceLeasesInFile`,
+            // whose very first action is acquiring `${path}.lock`. This is
+            // the only observable signal (short of instrumenting the lock
+            // acquisition itself) that a process has reached this point,
+            // which `tests/integration.zig`'s multi-process startup-
+            // reservation soak polls for on both processes before
+            // releasing a barrier lock it holds, so their real reservation
+            // attempts are forced to overlap rather than merely likely to.
+            state.logger.info(null, "native TLS/QUIC persistent ticket-key reservation attempt starting", .{});
```
Jaruso: **Blocking — this marker is still before the actual lock-attempt boundary.** The comment says entering `loadPersistentTicketKeysFromFile()` means the reservation attempt has started / its first action is acquiring `${path}.lock`, but that is not what the production function does. `Runtime.loadPersistentTicketKeysFromFile()` first calls `loadFromFile`, computes/compares the fingerprint, builds/validates the snapshot, and only then calls `ticket_key_snapshot.reserveNonceLeasesInFile()`. The lock is acquired inside that final call.

So this schedule is still possible while the test passes:

```text
A logs marker -> parses/validates -> blocks on parent-held lock
B logs marker -> is descheduled while still before reserveNonceLeasesInFile
parent sees both markers -> releases lock
A reserves and releases
B resumes -> reaches reserveNonceLeasesInFile later -> reserves sequentially
```

That does not establish the contention property from the original review. Please put the observable event at the **exact pre-reservation boundary** (immediately before `reserveNonceLeasesInFile`, or inside the reservation path immediately before the blocking lock acquisition), then wait for that process-specific event from both children before releasing the parent lock. If production instrumentation is needed, make it a real secret-safe observer/log event with operational meaning rather than a marker that claims a later stage than it actually represents.

tests/integration.zig
```diff
@@ -8391,6 +8680,834 @@ test "soak.reconnect_resumption" {
...
+        if (soakHeavyEnabled()) {
+            churn_a = ChurnWorker.init(process_a.port, &churn_stop, key_ids[retiring_index], key_ids[new_index]);
+            churn_thread_a = try std.Thread.spawn(.{}, ChurnWorker.run, .{&churn_a.?});
+            churn_b = ChurnWorker.init(process_b.port, &churn_stop, key_ids[retiring_index], key_ids[new_index]);
+            churn_thread_b = try std.Thread.spawn(.{}, ChurnWorker.run, .{&churn_b.?});
+        }
+
+        // Sent back-to-back, without waiting between them -- both real
+        // SIGHUP handlers race the shared snapshot lock concurrently.
+        process_a.sendSignal(std.posix.SIG.HUP);
```
Jaruso, line 9341: **Blocking acceptance gap — spawning the churn threads does not prove any churn actually overlaps the reload.** `std.Thread.spawn()` only makes the workers runnable; the main thread immediately sends both SIGHUPs. A valid scheduler can let A/B reload completely before either worker executes `issueTicketIds()`. The workers can then run after acceptance, collect one or more incoming-key samples, satisfy `results.items.len > 0`, and the heavy test passes without having exercised traffic during publication at all.

Please add an observed worker-arrival/in-flight condition before triggering reload, analogous to the startup contention barrier. At minimum each worker should expose a bounded `attempt_started`/`in_attempt` signal that the parent waits on before sending SIGHUP; ideally factor the probe so the worker can be held at a deterministic point inside a real connection/ticket attempt and release the traffic together with the reload signals. The important invariant is that the test cannot pass if all ticket traffic starts only after reload publication has finished.

tests/integration.zig
```diff
@@ -6257,6 +6272,280 @@ fn extractTicketEnvelopeIds(ticket: *const tls_core.session.ClientTicketState) T
...
+                return;
+            };
+            const hex = std.fmt.bytesToHex(ids.key_id, .lower);
+            const is_incoming = std.mem.eql(u8, self.incoming_key_id_hex, &hex);
+            const is_outgoing = std.mem.eql(u8, self.outgoing_key_id_hex, &hex);
+            if (!is_incoming and !is_outgoing) {
+                self.failure = error.ChurnObservedUnexpectedKeyId;
+                return;
+            }
+            if (self.require_new.load(.acquire) and !is_incoming) {
```
Jaruso: **Blocking flake / incorrect boundary classification:** `require_new` is loaded *after* `issueTicketIds()` completes. A ticket request can legitimately begin and be sealed with the outgoing key before this process reports `reload_accepted`, then finish reading/return to the worker after the main thread has observed acceptance and stored `require_new = true`. This code will retroactively reject that allowed in-flight old-generation ticket as `ChurnObservedStaleKeyAfterAcceptance`.

The phase must be captured at the start of the ticket attempt, not after it returns. For example:
```zig
while (!self.stop.load(.acquire)) {
    const require_new_for_this_attempt = self.require_new.load(.acquire);
    const ids = issueTicketIds(...) catch |err| { ... };
    ...
    if (require_new_for_this_attempt and !is_incoming) {
        self.failure = error.ChurnObservedStaleKeyAfterAcceptance;
        return;
    }
}
```
Then make sure each worker completes at least one attempt that *started* with `require_new_for_this_attempt == true` (or rely on an explicit post-acceptance probe for that invariant). This preserves the ticket's rule that old-or-new is allowed for work already in the transition while fresh work begun after the observed acceptance boundary must use the incoming key.

Jaruso (commented):
Third-pass review against #520. The replay per-identity metric fix remains correct, and this revision properly fixes the churn-worker error propagation and failure-path stop/join lifetime problem from the prior pass. CI 1021 is green. I still see three merge-blocking test-design issues: (1) the startup marker is emitted before the code reaches the reservation lock, so it still does not prove both children are contending the lock; (2) the heavy churn threads are spawned but not observed to have entered an issuance attempt before SIGHUP, so a schedule with all reload work completing before either churn thread runs can still pass; and (3) `require_new` is sampled after a whole request, which can retroactively classify an allowed pre-acceptance old-key issuance as a post-acceptance regression. Also please update the PR summary once these are fixed: it still says “No production TLS code changed,” but this head now changes `src/edge_gateway.zig` for the reservation marker. GitHub will not let this account submit REQUEST_CHANGES on its own PR, so this is a COMMENT review; the three inline findings should be treated as blocking.

src/edge_gateway.zig
Jaruso: Fixed in 0740978: moved the observable signal to the exact pre-reservation boundary using the existing resumption_runtime.Observer pattern -- a new onTicketKeyReservationAttemptFn callback fires immediately before loadPersistentTicketKeysFromFile calls reserveNonceLeasesInFile (after loadFromFile/fingerprint-compare/validation, right where the lock acquisition is the very next statement), wired in edge_gateway.zig to the same secret-free log line. tests/integration.zig's waitForReservationAttempt is unchanged (same marker string), it now just observes the correctly-placed event.

tests/integration.zig
Jaruso: Fixed in 0740978: ChurnWorker now sets an attempt_started flag immediately before every connection attempt (including its first), and the main thread waits (bounded, 5s) for both workers' attempt_started before sending either SIGHUP -- so real traffic is guaranteed to already be in flight when the reload race begins, not merely spawned-and-maybe-runnable.

tests/integration.zig
Jaruso: Fixed in 0740978: require_new is now sampled once at the start of each attempt (require_new_for_this_attempt) and used for that attempt's own classification, never re-read after issueTicketIds() returns -- so a request that legitimately began before this process's own reload_accepted can no longer be retroactively misclassified as a post-acceptance regression just because it happens to finish reading after the flag flips.

tests/integration.zig
```diff
@@ -6257,6 +6272,307 @@ fn extractTicketEnvelopeIds(ticket: *const tls_core.session.ClientTicketState) T
...
+
+    fn deinit(self: *ChurnWorker) void {
+        self.results.deinit();
+    }
+
+    fn run(self: *ChurnWorker) void {
+        while (!self.stop.load(.acquire)) {
+            self.attempt_started.store(true, .release);
+            const require_new_for_this_attempt = self.require_new.load(.acquire);
+            const ids = issueTicketIds(std.heap.page_allocator, self.port) catch |err| {
```
Jaruso, line 6731: **Blocking — this still does not establish the boundary the two new atomics claim to establish.** `attempt_started` is stored immediately *before* `issueTicketIds()`, and it is never cleared. A valid schedule is still:

```text
worker A: attempt_started=true; sampled require_new=false; descheduled
worker B: attempt_started=true; sampled require_new=false; descheduled
main: sees both flags -> sends both SIGHUPs -> reloads complete -> sets require_new=true
workers resume -> only now enter issueTicketIds()
```

In that schedule the heavy tier passes its “attempt started” gate even though **no request/ticket traffic was in flight during reload at all**. It also means a request that actually begins after this process reported `reload_accepted` is still classified with the stale `require_new_for_this_attempt=false`, so a transient old-key regression after acceptance can escape the worker check. The permanent boolean has the same problem if a first attempt completes before the main thread observes the flag.

Please synchronize on an event from the *real request/reload lifecycle*, not an instruction immediately before the call. A deterministic shape can reuse the shared lock you already control in this test:

```zig
// heavy round
var barrier = try holdReservationLock(ticket_keys_path);
defer barrier.close();

startChurnWorkers();
const reload_marker_a = reservationMarkerCount(process_a.log_path);
const reload_marker_b = reservationMarkerCount(process_b.log_path);

process_a.sendSignal(std.posix.SIG.HUP);
process_b.sendSignal(std.posix.SIG.HUP);

// Prove both SIGHUP handlers have entered the real reservation path and
// are blocked on the parent-held lock.
try waitForReservationMarkerCount(process_a.log_path, reload_marker_a + 1, 5_000);
try waitForReservationMarkerCount(process_b.log_path, reload_marker_b + 1, 5_000);

// While both reload handlers are definitely in flight, require actual
// successful churn progress (completed ticket/request samples), not a
// pre-call flag.
try waitForCompletedChurnDelta(&churn_a, 1, 5_000);
try waitForCompletedChurnDelta(&churn_b, 1, 5_000);

barrier.close();
// keep churn running while publication completes
try waitFor...A;
const a_after_accept = try issueTicketIds(allocator, process_a.port);
try expectKey(a_after_accept, key_ids[new_index]);
try waitFor...B;
const b_after_accept = try issueTicketIds(allocator, process_b.port);
try expectKey(b_after_accept, key_ids[new_index]);
```

Exact plumbing can differ, but the invariants should be: **(1)** actual successful traffic is observed while the SIGHUP handlers are demonstrably active, and **(2)** a fresh probe initiated only after each process's own `reload_accepted` must use the incoming key. A pre-call atomic flag cannot prove either property.

Jaruso (commented):
Fourth-pass re-review against #520. Most prior blockers are now genuinely addressed: the reservation observer moved to the correct pre-reservation location, replay-locality uses exact per-identity metric deltas, churn errors are surfaced, and worker lifetime cleanup is bounded on failure paths.

I still have one merge-blocking gap in the heavy reload proof: the new `attempt_started` / `require_new_for_this_attempt` boundary is still outside the actual network attempt, so it can both falsely claim traffic overlapped reload and classify a request that truly starts after acceptance as pre-acceptance. See the inline comment for a deterministic shape that uses the existing shared lock/reload markers rather than another pre-call flag.

Separately, the latest CI run (#1022) is red in `Packaging smoke` → `Smoke test DEB package`. The appliance resumption/0-RTT smoke job itself passed, so I am not attributing that packaging failure to #520 without evidence, but the PR should still return to green before merge.

tests/integration.zig
Jaruso: Fixed in 848ce53: replaced both flags with deterministic mechanisms. (1) The test itself now acquires ${ticket_keys_path}.lock before sending either SIGHUP -- the same createFile(..., .lock = .exclusive) call production uses -- forcing both real reload handlers to genuinely block on it, confirmed via a per-round count of the existing reservation-attempt log marker (it fires once at startup and once per reload, so counting occurrences detects this specific round's attempt). While both are confirmed still blocked, heavy mode requires each churn worker to have *completed* (not merely started) at least one real connection, via a plain atomic counter safe to poll while the worker thread runs. (2) The post-acceptance 'incoming key only' boundary is now an explicit synchronous single-shot probe issued immediately after observing each process's own reload_accepted, not inferred from a concurrently-running worker's ambiguous phase sample. ChurnWorker itself is simplified back to just the structural outgoing-or-incoming validation, which is safe for a concurrent worker to check.

tests/integration.zig
```diff
@@ -8391,6 +8719,873 @@ test "soak.reconnect_resumption" {
...
+        try waitForReservationAttemptCount(allocator, process_b.log_path, reservation_marker_count_before_b + 1, 5_000);
+
+        // Heavy tier only, while both reload handlers are confirmed still
+        // blocked on the held lock: require each worker to have
+        // *completed* at least one real connection during this stalled
+        // window -- proof that actual traffic overlapped the reload race
+        // itself, not merely that a worker thread was runnable at some
+        // point before or after it.
+        if (soakHeavyEnabled()) {
+            try waitForCompletedChurnAtLeast(&churn_a.?, 1, 5_000);
```
Jaruso: **Blocking — this is still cumulative, so it does not prove traffic completed during the blocked reload window.** The churn workers are spawned before the reload barrier is acquired and before either SIGHUP is sent. If A and B each complete one connection during that setup time, `completed` is already `>= 1`; after lines above confirm both reload handlers reached the held reservation lock, these two waits return immediately. The test then releases the lock having observed **zero new traffic while the handlers were confirmed blocked**, so the exact false-positive this change is meant to eliminate is still possible.

Take the baseline only after both current-round reservation markers have been observed, while `reload_barrier` is still held, and require a delta from that point:
```zig
try waitForReservationAttemptCount(...A...);
try waitForReservationAttemptCount(...B...);

if (soakHeavyEnabled()) {
    const completed_before_a = churn_a.?.completed.load(.acquire);
    const completed_before_b = churn_b.?.completed.load(.acquire);

    try waitForCompletedChurnAtLeast(&churn_a.?, completed_before_a + 1, 5_000);
    try waitForCompletedChurnAtLeast(&churn_b.?, completed_before_b + 1, 5_000);
}

reload_barrier.close();
```
That establishes the actual required invariant: at least one successful real connection per process **completed after both SIGHUP handlers were demonstrably in the reservation path and before the parent released the lock**. If this corrected form times out because the reload path stalls all new connection progress, that is meaningful #520 behavior to investigate rather than something the soak should mask.

Jaruso (commented):
Fifth-pass review against #520 and the prior blocker.

The prior synchronization problem is mostly fixed: the rotation test now holds the real shared reservation lock, observes each process's current-round reservation marker before releasing it, and uses explicit synchronous fresh-ticket probes immediately after each process's own `reload_accepted`. The startup contention observer, churn error propagation/lifetime cleanup, nonce-exhaustion case, and exact per-identity replay deltas all remain sound.

I still have one merge-blocking test-proof issue: the new `completed` counter is cumulative, but the heavy-tier wait uses an absolute threshold of `1`. Because the churn threads start before the reload barrier is acquired/signaled, either worker can already have `completed >= 1` before this round's SIGHUPs. Once both reload markers are observed, these waits can therefore return immediately without proving any connection completed during the confirmed-blocked window. Capture each worker's `completed` baseline only *after both current-round reservation markers have been observed while the parent still holds the lock*, then require `baseline + 1` before releasing the barrier.

The current CI run for this head is still in progress. The previously failing DEB packaging step has passed on this run, and the macOS appliance resumption/0-RTT soak smoke step is green so far.

Also update the PR summary wording after this fix: it currently says heavy mode is gated on an observed in-flight attempt before either signal, but the new design intentionally sends the signals first, pins both reload handlers behind the held lock, and then requires completed traffic during that window.

tests/integration.zig
Jaruso: Implemented exactly this in b574ba9, and it did exactly what your last sentence anticipated: the corrected form (baseline captured only after both reservation markers observed, requiring baseline+1 while the lock is held) reliably hung for the full 5-minute test-level timeout. Investigated per your note rather than masking it: edge_gateway.zig's event loop calls hotReloadConfig synchronously from the same single-threaded loop that accepts/dispatches connections. With TARDIGRADE_WORKER_THREADS=1, blocking that thread on the shared reservation lock blocks all socket I/O for that process -- including finishing already-in-flight work, not only new accepts. So no client-observable connection can ever complete while a process's own reload is confirmed stalled on the lock; the invariant is architecturally unobservable, not merely hard to arrange. Removed the completed-counter/baseline mechanism and the held-lock-plus-completion requirement; replaced with continuous churn traffic across the whole round (started before either SIGHUP, stopped only after both reload_accepted + post-acceptance probes) asserted to span both the outgoing and incoming key -- the achievable proof of continuity given this architecture. The reservation-lock-forcing via the marker count is unaffected and still works (that part never hung, since it doesn't require any connection to complete while blocked).

tests/integration.zig
```diff
@@ -8391,6 +8726,890 @@ test "soak.reconnect_resumption" {
...
+        // connection while the lock above is held. `edge_gateway.run`'s
+        // main loop (`http.shutdown.consumeReloadRequested()` ->
+        // `gshutdown.hotReloadConfig`) calls `hotReloadConfig`
+        // synchronously from the same single-threaded event loop that
+        // also accepts new connections (`gaccept.acceptReadyConnections`)
+        // and dispatches ready sockets to the worker pool -- with
+        // `TARDIGRADE_WORKER_THREADS=1`, blocking that one loop thread on
+        // `${path}.lock` therefore blocks *all* socket I/O for that
+        // process, including finishing already-in-flight work, not only
+        // new accepts. An earlier revision of this test required exactly
```
Jaruso: **Blocking — the architectural justification for dropping the in-race traffic proof is incorrect.** `hotReloadConfig()` is synchronous on the main event-loop thread, but that does **not** mean blocking that thread blocks all socket I/O. `edge_gateway.run()` owns a separate `WorkerPool` of real `std.Thread`s; the event loop submits active native sockets to those workers, and once a native H1 connection is inside `advanceNativeHttp1()`, `WaitingEncryptedHttpConnection.waitFor()` calls `std.posix.poll()` directly on the raw fd from the worker thread. An already-dispatched worker can therefore continue request/TLS I/O while the main event loop is stalled on `${path}.lock`.

The corrected `baseline + 1` experiment timing out only proves the current churn loop had no connection already dispatched into that worker phase when the main loop was stalled; it does **not** prove the #520 overlap invariant is architecturally unobservable. The replacement `old + new somewhere in the round` check also weakens the issue's explicit heavy-tier requirement to keep request/ticket traffic in flight while the SIGHUP handlers race.

Please make the overlap deterministic at the real worker boundary. One workable shape is to pre-establish a native H1 connection and drive it into an observably active worker request (for example, through a gated test upstream or another exact application-side barrier), then acquire the reservation lock, signal both processes, wait for both current-round reservation markers, and release the request/upstream gate while the reload lock is still held. That gives you a request that is **actually executing on the worker while the main loop is blocked in reload**. Keep it alive across `reload_barrier.close()` as well if needed to exercise publication/keyring replacement, then require bounded completion and the normal post-acceptance new-key probe. The important invariant remains the one in #520: real request/ticket work must overlap the reload race, not merely exist before and after it.

tests/integration.zig
```diff
@@ -8391,6 +8726,890 @@ test "soak.reconnect_resumption" {
...
+            try std.testing.expect(churn_a.?.results.items.len > 0);
+            try std.testing.expect(churn_b.?.results.items.len > 0);
+
+            // Traffic actually spanned the transition: each worker's
+            // collected samples include both the outgoing and the
+            // incoming generation's key id, not merely "some traffic
+            // happened at some point in the round".
+            try expectSpansBothGenerations(churn_a.?.results.items, key_ids[retiring_index], key_ids[new_index]);
+            try expectSpansBothGenerations(churn_b.?.results.items, key_ids[retiring_index], key_ids[new_index]);
+
```
Jaruso, line 9432: **Blocking heavy-soak flake / insufficient synchronization:** `expectSpansBothGenerations()` is asserted after the fact, but nothing guarantees either churn worker actually samples the outgoing generation before the SIGHUPs, or the incoming generation before `churn_stop` is set. The main thread starts the workers and immediately proceeds to the reload barrier/signals; after the two synchronous post-acceptance probes it immediately stops the workers. A valid scheduler can therefore give a worker only incoming samples (started late) or only outgoing samples (not rescheduled before stop), making the scheduled heavy soak fail nondeterministically.

At minimum, add bounded, thread-safe observed-generation state/counters and wait for an outgoing sample from **each** worker before sending SIGHUP, then wait for an incoming sample from each worker before stopping them. That makes this before/after continuity assertion deterministic. It still should be supplementary to — not a replacement for — the actual in-race worker-overlap proof in the other comment.

Jaruso (commented):
Sixth-pass review of #529 / #520. The previous cumulative-counter bug is gone, and the exact reservation marker/barrier plus the explicit post-`reload_accepted` new-key probes remain good. CI run #1024 is green. However, this revision replaces the missing overlap proof with an architectural claim that is not true for the native H1 path, and the replacement old+new sample assertion is itself scheduler-dependent. I still consider the two inline findings below merge-blocking for #520's heavy-tier acceptance criterion. Because this PR is authored by the authenticated account, GitHub will not allow a formal REQUEST_CHANGES review, so this is submitted as COMMENT.

tests/integration.zig
```diff
@@ -8391,6 +8726,890 @@ test "soak.reconnect_resumption" {
...
+            try std.testing.expect(churn_a.?.failure == null);
+            try std.testing.expect(churn_b.?.failure == null);
+            try std.testing.expect(churn_a.?.results.items.len > 0);
+            try std.testing.expect(churn_b.?.results.items.len > 0);
+
+            // Traffic actually spanned the transition: each worker's
+            // collected samples include both the outgoing and the
+            // incoming generation's key id, not merely "some traffic
+            // happened at some point in the round".
+            try expectSpansBothGenerations(churn_a.?.results.items, key_ids[retiring_index], key_ids[new_index]);
```
Jaruso, line 9430: **Blocking heavy-soak flake:** checking for both generations only *after* the workers are stopped makes scheduler timing part of correctness. The workers are spawned and SIGHUP is sent immediately; a loaded runner may not schedule a worker until after reload, so it sees only `incoming`. Conversely, after the post-acceptance probes the main thread immediately sets `churn_stop`, so a worker that did get an outgoing sample may not get scheduled again to complete an incoming sample before shutdown. Either schedule fails this assertion even when production is correct.

Please gate the two sides **outside** the intentionally blocked reload interval: wait for each worker to report one successful outgoing sample before sending SIGHUP, then after both reloads are accepted wait for each worker to report one successful incoming sample before setting `churn_stop`. `saw_outgoing` / `saw_incoming` atomics set after a validated+appended sample would make this deterministic without reintroducing the impossible "complete while reload is blocked" requirement.

Jaruso (commented):
Sixth-pass review on `b574ba94`.

The architectural correction in this revision is valid. I re-read the production event loop: SIGHUP reload is consumed on the main event-loop thread and `gshutdown.hotReloadConfig(...)` runs synchronously there, while that same loop owns native socket-readiness dispatch. For this native appliance composition, deliberately blocking reload on `${path}.lock` also prevents the process from making native connection progress. So my previous suggestion to require a native request to complete while that process's reload was confirmed blocked was not an achievable invariant; removing that requirement is correct.

The new `outgoing + incoming` churn-span assertion is a reasonable replacement for the operational continuity part of #520, and the deterministic reservation-contention proof, explicit per-process post-`reload_accepted` new-key probes, global `(key_id, nonce)` uniqueness checks, exhaustion case, and process-local replay case all remain intact.

I do still see one heavy-tier determinism blocker: the workers are spawned and then the test immediately acquires the reload barrier / sends SIGHUP. There is no condition proving either worker completed an outgoing-generation sample before reload starts, and after both post-acceptance probes the workers are stopped immediately without proving either worker completed an incoming-generation sample. `expectSpansBothGenerations(...)` therefore turns scheduler timing into a pass/fail condition: on a loaded runner a worker may simply not get scheduled on one side of the transition and the soak fails even though production is correct.

Please make the two sides observable without trying to make progress inside the deliberately blocked interval. For example, expose atomics on each worker that are set only after a successful sample is appended:

```zig
saw_outgoing: std.atomic.Value(bool) = std.atomic.Value(bool).init(false),
saw_incoming: std.atomic.Value(bool) = std.atomic.Value(bool).init(false),

// after validating + appending ids
if (is_outgoing) self.saw_outgoing.store(true, .release);
if (is_incoming) self.saw_incoming.store(true, .release);
```

Then:

```zig
// workers started, event loops still free
try waitForFlag(&churn_a.?.saw_outgoing, 5_000);
try waitForFlag(&churn_b.?.saw_outgoing, 5_000);

// acquire barrier, signal, observe both reservation attempts, release barrier,
// wait for each process's reload_accepted and run the explicit new-key probes

// event loops are free again; require worker-side post-transition progress
try waitForFlag(&churn_a.?.saw_incoming, 5_000);
try waitForFlag(&churn_b.?.saw_incoming, 5_000);

churn_stop.store(true, .release);
join...
```

That preserves the architecture-aware proof: successful worker traffic before the transition, the real reload race itself, and successful worker traffic after publication, with no impossible requirement to complete I/O while the event loop is intentionally stalled.

CI #1024 is green on this head. Once this scheduler-dependence is removed, I do not currently see another #520 acceptance gap.

tests/integration.zig
```diff
@@ -6257,6 +6272,359 @@ fn extractTicketEnvelopeIds(ticket: *const tls_core.session.ClientTicketState) T
...
+/// before acceptance and only finish reading after it, or vice versa),
+/// so it cannot establish that boundary correctly no matter which side of
+/// the request it is sampled on.
+///
+/// Deliberately does *not* attempt to prove a worker completed a
+/// connection *while* the reload SIGHUP handler is confirmed blocked on
+/// the shared reservation lock -- `edge_gateway.run`'s single-threaded
+/// event loop handles both accepting/dispatching connections and
+/// (synchronously) `hotReloadConfig`, so blocking that one thread on the
+/// lock blocks all socket I/O for that process, including finishing
```
Jaruso: **Blocking — this architectural claim is incorrect, and it is why the heavy-tier proof is still weaker than #520 requires.** The main event loop does own accept/readiness dispatch and calls `hotReloadConfig()` synchronously, but native connection work runs on the separate `WorkerPool`. Once a connection is already dispatched, `advanceNativeHttp1()` runs on that worker and `WaitingEncryptedHttpConnection.waitFor()` calls `std.posix.poll()` directly on the raw fd. So blocking the main event loop on `${path}.lock` does **not** block all socket I/O or prevent an already-dispatched worker from completing request work.

The previous `baseline + 1` churn experiment timed out because those churn iterations create new connections, which need the blocked main loop to accept/dispatch them. It does not establish that worker-side overlap is impossible.

Please preserve the new `saw_outgoing`/`saw_incoming` proof, but add a deterministic already-dispatched request for each process during the reload race. There is already a reusable pattern later in this file (`in-flight request completes safely across reload and new requests use new config`): run a request on its own thread and wait until the upstream has observed it before signaling reload. For this soak, use a **gate** rather than a fixed delay: wait until both process-specific upstream gates report `entered`, acquire the reservation lock, send both HUPs, wait for both reservation markers, then release the two upstream gates while the reload lock is still held and require both requests to complete boundedly. Only after that release the reservation lock and continue with the existing `reload_accepted` / incoming-key probes.

Conceptually:
```zig
startRaceRequest(A); startRaceRequest(B);
try gate_a.waitEntered(5_000);
try gate_b.waitEntered(5_000); // proves both are already on workers

var barrier = try holdReservationLock(...);
sendHup(A); sendHup(B);
try waitForReservationAttemptCount(...A...);
try waitForReservationAttemptCount(...B...);

gate_a.release(); gate_b.release();
try waitRaceRequestComplete(A, 5_000);
try waitRaceRequestComplete(B, 5_000); // worker progress while main loops blocked

barrier.close();
```
That directly exercises the operational in-race request/UAF/deadlock condition #520 calls for without depending on new accepts while the event loop is stalled.

Jaruso (commented):
Seventh review pass at `8f8258a8`.

The latest `saw_outgoing` / `saw_incoming` atomics and bounded waits are implemented correctly and close the scheduler-dependent before/after-generation flake from the prior pass. The flags are published only after a validated sample is retained, the outgoing waits happen before the reservation barrier, and the incoming waits happen after both reloads have completed. I found no new problem in that patch.

**One #520 blocker remains:** the reason given for removing the actual in-race worker-overlap proof is not consistent with the production connection model. `edge_gateway.run()` has a main-thread event loop *plus a separate `WorkerPool`*. Once a native H1 connection has been dispatched, `advanceNativeHttp1()` executes on the worker and `WaitingEncryptedHttpConnection.waitFor()` directly calls `std.posix.poll()` on the fd. Blocking the main loop inside synchronous `hotReloadConfig()` prevents new accept/readiness dispatch, but it does not stop an already-dispatched worker from continuing socket/request I/O.

That matters because #520 explicitly requires the heavy tier to keep real request/ticket traffic active while the two SIGHUP reloads race. The current churn workers now deterministically prove healthy traffic before and after the transition, but they still do not prove any request has actually crossed into a server worker before the reload begins. A valid execution can therefore have all server-side request/ticket work happen before and after `hotReloadConfig`, with no worker concurrently active during the reload/publication path.

There is already a useful harness pattern in this file: `in-flight request completes safely across reload and new requests use new config` starts a request thread and uses `waitForUpstreamCount()` to prove that request has reached the upstream before sending SIGHUP. For #520 I would make that deterministic rather than delay-based by using one gated upstream per Tardigrade process:

```zig
// Conceptual shape; keep the existing churn workers as the before/after
// ticket sampling proof.
var gate_a = try GatedUpstream.start(allocator);
defer gate_a.stop();
var gate_b = try GatedUpstream.start(allocator);
defer gate_b.stop();

// Each /race request is native TLS H1 and proxies to that process's gate.
var req_a = try std.Thread.spawn(.{}, runTlsRaceRequest, .{ process_a.port, &result_a });
var req_b = try std.Thread.spawn(.{}, runTlsRaceRequest, .{ process_b.port, &result_b });

// This is the crucial worker-side arrival condition: both requests have
// already been accepted, dispatched, processed by the worker, and reached
// their upstream before reload starts.
try gate_a.waitEntered(5_000);
try gate_b.waitEntered(5_000);

var reload_barrier = try holdReservationLock(ticket_keys_path);
process_a.sendSignal(std.posix.SIG.HUP);
process_b.sendSignal(std.posix.SIG.HUP);
try waitForReservationAttemptCount(...A...);
try waitForReservationAttemptCount(...B...);

// Main event loops are now blocked in reload, but the already-dispatched
// workers must still be able to finish their requests.
gate_a.release();
gate_b.release();
try waitForRaceRequestComplete(&result_a, 5_000);
try waitForRaceRequestComplete(&result_b, 5_000);

reload_barrier.close();
req_a.join();
req_b.join();

// Keep the existing independent reload_accepted + incoming-key probes and
// saw_incoming waits after this.
```

The existing `UpstreamServer`/request-thread test demonstrates that this worker-overlap shape is supported by the architecture; the gate should be condition-based rather than a fixed delay so the proof remains deterministic.

Everything else from #520 remains satisfied from the prior passes: deterministic multi-process lease contention, exact lease advancement/disjoint windows, global `(key_id, nonce)` uniqueness, retiring-ticket resumption, per-process post-acceptance incoming-key probes, exhaustion isolation, exact process-local replay deltas/application effects, bounded cleanup, and secret-safe diagnostics.

CI note: run #1025 is red, but the #520 smoke itself passed (`soak.persistent.multi_process_nonce_safety: heavy=false ...`). The sole failure I found is the unrelated existing macOS `interop.h2.window_update_connection_overflow_sends_goaway` case (`ConnectionResetByPeer` / `SocketReadFailed`). That should be rerun/cleared before merge, but it is not evidence against this patch.

Because this is the author's own PR I can only submit a COMMENT review. **Not merge-ready yet: one heavy-tier overlap acceptance gap remains.**

tests/integration.zig
Jaruso: Fixed in cb1b371: added GatedUpstream, a real upstream that deterministically proves a request reached it (and holds it there under test control) rather than a fixed delay. Each process gets its own gated upstream via a /race route. Per heavy round: one real native TLS H1 request per process is dispatched and confirmed to have reached its gate before the reservation barrier is even acquired; once both real SIGHUP handlers are confirmed blocked on the shared lock, the gates are released and both requests' completion is confirmed before the lock itself is released -- proving real request traffic was genuinely in flight throughout the reload race, exactly the shape you described (adapted from the existing 'in-flight request completes safely across reload' pattern, but condition-based via the gate rather than delay-based). Race-thread cleanup is deferred before any fallible spawn/wait, mirroring the churn-worker lifetime fix. Along the way this surfaced a transient WouldBlock in the churn loop (its tight reconnect pace occasionally outpacing the single TARDIGRADE_WORKER_THREADS=1 worker while it's held busy by the gated request) -- not a correctness bug, now retried specifically while every other error remains fatal. Re-verified: full unit suite, full test-integration-resumption-interop suite, and the affected test individually in both default and TARDIGRADE_SOAK_HEAVY=1 mode, heavy mode run three times back to back with no flake.

Jaruso (commented):
Re-reviewed the new head `cb1b371` against the full #520 acceptance criteria and the actual gateway worker/event-loop execution path.

The **gated-worker overlap fix is correct** and closes my previous blocker. In heavy mode each process now has a real native H1 request demonstrably accepted/dispatched all the way to a process-specific gated upstream before SIGHUP; both reload handlers are then proven blocked in the real persistent reservation path; the gates are released and both worker-side requests must complete boundedly while the handlers are still active. Combined with the existing deterministic outgoing/incoming churn samples, per-process post-`reload_accepted` incoming-key probes, retained-ticket resumption, lease-window assertions, and global `(key_id, nonce)` uniqueness checks, that satisfies the intended heavy-tier request/reload overlap proof. I traced `hotReloadConfig` -> `Runtime.loadPersistentTicketKeysFromFile` -> `reserveNonceLeasesInFile` -> `installPersistentTicketKeys` again and I am **not** asking for another new publication barrier after this.

I found **one blocker introduced by this new commit**: the new `GatedUpstream` threads and both `runTlsRaceRequest` threads are all given the test's `std.testing.allocator` concurrently. This same file already documents why `ChurnWorker` deliberately uses `std.heap.page_allocator` instead: the testing allocator is not safe for concurrent use. The heavy path can therefore race the allocator from four helper threads (and the test thread), even though smoke CI won't exercise those allocations. The inline comment has the exact minimal fix: use `std.heap.page_allocator` for the gate-owned/thread-owned allocations and race clients, while leaving ordinary main-thread test allocations on `std.testing.allocator`.

I also left a nonblocking cleanup note on the stale `ChurnWorker` architecture comment. It still says already-dispatched worker I/O cannot progress while the main loop is blocked, which this commit correctly disproves. The PR description repeats that old explanation and should be updated at the same time so future readers don't relearn the wrong model.

Everything else from #520 still checks out on this head: deterministic startup contention at the true pre-reservation boundary; exact `[2W,3W)` durable lease advancement and disjoint process windows; complete old/new generation handling; independent reload acceptance and fresh-new-key probes; retiring-ticket resumption on both processes; exhaustion isolation; exact process-local replay acceptance/duplicate deltas and side-effect behavior; bounded cleanup; secret-safe observer/logging; and existing workflow/cache-leak coverage.

CI #1026 completed successfully on this exact head.

**Conclusion: not merge-ready only because of the new concurrent `std.testing.allocator` usage. I found no additional #520 design/acceptance gaps in this pass.**

tests/integration.zig
```diff
@@ -8391,6 +8921,997 @@ test "soak.reconnect_resumption" {
...
+    // declared in that process's own config -- the heavy tier's real
+    // worker-overlap proof below drives one request through each of
+    // these, per round, and holds it there under the test's own control
+    // while the reservation lock is held. Created (and their `/race`
+    // route wired) regardless of tier, since the process's own config
+    // is fixed at startup and can't be changed per-round; only the
+    // smoke tier simply never drives traffic through it.
+    var gate_a = try GatedUpstream.start(allocator);
+    defer gate_a.stop();
+    try gate_a.run();
```
Jaruso, line 9045: **Blocking — the new heavy-only helpers share `std.testing.allocator` across multiple real threads.** This test defines `allocator = std.testing.allocator`, then both `GatedUpstream` instances retain it and allocate/free through it from their spawned server threads (`readHttpMessage`, response `ArrayList`), while both `runTlsRaceRequest` threads are also passed that same allocator and use it for `PureZigTlsClient` plus response buffers. That creates concurrent allocator access from up to four helper threads plus the main test thread.

This file already documents the required rule in `ChurnWorker`: it intentionally uses `std.heap.page_allocator` because `std.testing.allocator` is not safe to call concurrently from multiple threads. The new helpers need to follow the same rule. This is especially easy to miss in CI because the normal smoke tier starts the gate threads but never sends requests through them; the concurrent allocations happen in heavy mode.

Minimal fix:
```zig
var gate_a = try GatedUpstream.start(std.heap.page_allocator);
defer gate_a.stop();
try gate_a.run();
var gate_b = try GatedUpstream.start(std.heap.page_allocator);
defer gate_b.stop();
try gate_b.run();
```

and either pass `std.heap.page_allocator` to the race threads:
```zig
race_thread_a = try std.Thread.spawn(
    .{}, runTlsRaceRequest,
    .{ std.heap.page_allocator, process_a.port, &race_result_a },
);
race_thread_b = try std.Thread.spawn(
    .{}, runTlsRaceRequest,
    .{ std.heap.page_allocator, process_b.port, &race_result_b },
);
```

or better, make `runTlsRaceRequest` choose `std.heap.page_allocator` internally so a caller cannot accidentally pass the test allocator again. Main-thread fixture/config allocations can remain on `std.testing.allocator`.

tests/integration.zig
```diff
@@ -6257,6 +6420,373 @@ fn extractTicketEnvelopeIds(ticket: *const tls_core.session.ClientTicketState) T
...
+/// after acceptance" is inherently racy (a request can legitimately begin
+/// before acceptance and only finish reading after it, or vice versa),
+/// so it cannot establish that boundary correctly no matter which side of
+/// the request it is sampled on.
+///
+/// Deliberately does *not* attempt to prove a worker completed a
+/// connection *while* the reload SIGHUP handler is confirmed blocked on
+/// the shared reservation lock -- `edge_gateway.run`'s single-threaded
+/// event loop handles both accepting/dispatching connections and
+/// (synchronously) `hotReloadConfig`, so blocking that one thread on the
```
Jaruso: **Nonblocking cleanup, but this comment is now factually stale and contradicts the new proof below.** The latest commit correctly establishes that an already-dispatched native H1 worker *can* keep doing socket/request I/O while the main event loop is blocked in synchronous `hotReloadConfig`; that is exactly why `GatedUpstream` works. What is impossible for `ChurnWorker` specifically is guaranteeing its **next newly-created connection** gets accepted/dispatched after the main loop has blocked.

Please rewrite this paragraph to make that distinction explicit rather than preserving the old "blocks all socket I/O / architecturally unobservable" explanation. The PR description contains the same stale statement and should be updated too. This is documentation-only; the new gated runtime proof itself is correct.

Jaruso (commented):
Re-reviewed head `61ec3fd4123758f6f311ceb5e15af33c024e7ee8` against #520 and the full prior review history.

The allocator blocker is fixed correctly: both `GatedUpstream` instances and both heavy-mode `runTlsRaceRequest` threads now use `std.heap.page_allocator`, so the new concurrent helpers no longer share `std.testing.allocator` across threads. The stale `ChurnWorker` explanation was also corrected, and the PR description now accurately distinguishes blocked new accepts from already-dispatched worker progress.

I rechecked the full #520 acceptance surface, including deterministic startup contention, per-round SIGHUP contention, disjoint durable nonce reservations, global `(key_id, nonce)` uniqueness, complete-generation behavior, explicit post-own-`reload_accepted` new-key probes, real already-dispatched worker traffic completing while both reload handlers are blocked, retained-ticket resumption, nonce exhaustion isolation, and exact per-process replay-locality semantics. I found no remaining code or acceptance-criteria blockers.

**Code review: merge-ready.** Exact-head CI #1027 is still in progress at the time of this review, so final merge should wait for that run to finish successfully.

PR #524, opened 2026-07-26
quic: implement native QUIC/H3 0-RTT packet acceptance and early-data delivery

Jaruso:
## Summary

Implements the missing native QUIC/H3 0-RTT carrier: the QUIC connection driver previously dropped every `.zero_rtt` packet unconditionally (`src/quic/connection.zig`'s `ingestPacket`), and `http3_runtime.zig` logged "0-RTT is not supported" whenever `enable_0rtt` was set. Most of the surrounding TLS-side machinery (PSK/resumption decision, replay gate, H3 SETTINGS compatibility, ticket issuance) already existed from #366/#367/#368/#488/#513 — this closes the QUIC-carrier gap beneath it, reusing that machinery rather than reimplementing any of it.

- **`src/quic/connection.zig`**
  - `ingestPacket`: map `.zero_rtt` through the same decrypt / packet-number-space / recovery / ACK path as every other level, gated entirely on `adapter.protectionKeys(.zero_rtt, .read)` — which already refuses to hand back keys for an unauthorized/disabled 0-RTT attempt, so "no keys installed" naturally becomes the fail-closed drop case with no new decision logic.
  - `applyFrame`: relax the STREAM-frame level check to accept `.zero_rtt`, and mark accepted 0-RTT streams via the existing (previously unused) `markStreamZeroRtt` provenance seam.
  - `Connection.init`: wire `adapter.setZeroRttEnabled(cfg.zero_rtt_enabled)` — the adapter's own gate existed but nothing ever called it.
  - **`ensureEarlyStreamManager`** (the non-obvious part): the stream/flow-control layer was only ever created once the *full* handshake completed and peer transport parameters were authenticated — strictly after 0-RTT data arrives, which would have closed the connection on the first accepted 0-RTT STREAM frame (`stream before handshake`). Brings the stream layer up early on the first authenticated 0-RTT packet, using local transport parameters (receive-side admission depends only on those, confirmed via `stream.zig`) and a zero-send-credit placeholder for the peer side, since nothing can be sent before the handshake completes anyway. `afterHandshakeProgress` now refreshes (`StreamManager.refreshPeerParams`, monotonic raise-only, mirroring `applyMaxData`/`applyMaxStreamData`) rather than reinitializing — reinitializing would silently discard any stream state buffered from accepted 0-RTT packets.
[...]

PR #532, opened 2026-07-28
docs: add authoritative final #369 matrix to resumption test plan

Jaruso:
## Summary

All implementation work for #369's resumption/0-RTT assurance epic is already merged into `main` (#508, #509, #511, and the #518–#523 series, including #522's own H3 interop landed via PR #527). The only remaining #522 scope, per the issue, is the final documentation reconciliation of `docs/RESUMPTION_TEST_PLAN.md`.

- Adds an authoritative **Final #369 status** matrix mapping every #369 acceptance criterion to a named deterministic/process/external proof, or to an explicitly unreachable capability with its reason.
- Adds an **Intentional limitations** section covering the four negative cases that are unreachable under the current production configuration surface (cipher-suite mismatch, H3 SETTINGS mismatch, QUIC transport-parameter mismatch, H3 replay-store-unavailable) plus the worker-thread-pinning test-seam gap and the distributed-replay non-goal.
- States the 1-RTT vs 0-RTT compatibility contract, process-local (not cluster-wide) replay semantics, and transport/profile support boundaries explicitly, per section requirements.
- Reorganizes the doc so the final matrix is authoritative and the prior Slice 1–5 write-ups move under a clearly-labeled "Historical slice record" section.
- Fixes the specific stale claims called out by #522: Slice 3's "0-RTT external/production interop, entirely" and "Independent QUIC/H3 external interop" deferred bullets are now marked resolved with pointers to the PRs that closed them, and a genuinely stale Slice 4 bullet (general-profile credential-reload atomicity, which was actually already closed by a co-landed test in the same PR) is corrected.
- No new tests, production code, or configuration knobs — this is a docs-only reconciliation, as the issue explicitly requires.

## Test plan

- [x] `zig build test` passes (docs-only change; confirms nothing else regressed).
- [x] Verified every test name and metric name referenced in the new matrix/limitations text against current `main` source (`grep` for `h3interop.quic.*`, `zeroRttCarrierEnabled`, `quicConfigFrom`, `quicEarlyDataTransportCompatible`, `ResumeMismatch.cipher_suite_mismatch`, etc.).
- [x] Confirmed via `git log` that the four cross-referenced PRs (#524–#529, #518/#525's equivalent) are already on `main`.

Closes #522
Closes #369

### DIFF
--- a/docs/RESUMPTION_TEST_PLAN.md
+++ b/docs/RESUMPTION_TEST_PLAN.md
@@ -19,14 +19,15 @@ matrix wins.
 |---|---|---|---|---|---|
 | Expiry, key rotation/removal, restart/lost-state models, ALPN/cipher compatibility scenarios | native TLS record layer, all profiles | `Tls13Backend` / `ReloadableKeyRing` | `tls13_backend_tests.zig`, `ticket_protection.zig`, `resumption_runtime.zig` scenario tests (#508) | deterministic | complete |
 | Process-level replay decisions, capacity/quarantine, 425/retry exactly-once composition | native TCP/H1, all profiles | process-local replay gate + gateway retry state machine | `rt0.*` cases (`tls13_backend_tests.zig`/`edge_gateway.zig`), `rt0.retry.425_exactly_once` (#509) | process (real composition, in-process) | complete |
-| Production native TCP/H1 record 0-RTT carrier + persistent ticket-key composition | native TCP/H1; persistent keys are general-profile only | `NativeTlsConnection` 0-RTT record path; `TARDIGRADE_TLS_NATIVE_TICKET_KEYS_PATH` | `#510 native tcp production 0-rtt reaches h1 safety gate and replay fallback` (#513) | process (real production composition) | complete |
+| Production native TCP/H1 record 0-RTT carrier | native TCP/H1, appliance profile | `NativeTlsConnection` 0-RTT record path | `#510 native tcp production 0-rtt reaches h1 safety gate and replay fallback` (#513) | process (real production composition) | complete |
+| Persistent stateless/hybrid ticket-key composition | shared native resumption runtime; exercised by appliance-profile H1 (#519/#520) and general-profile QUIC/H3 (#519's credential-atomicity case) | `TARDIGRADE_TLS_NATIVE_TICKET_KEYS_PATH` / `ReloadableKeyRing` | #513, composed by #519/#520's proofs | process (real production composition) | complete |
 | Production QUIC/H3 0-RTT packet carrier, provenance, policy/configuration, deterministic compatibility | native QUIC/H3, general profile | `http3_runtime.zeroRttCarrierEnabled` / `quicConfigFrom` / `quicEarlyDataTransportCompatible` | `quicConfigFrom`/H3 runtime deterministic tests (#523/PR #524) | deterministic + process | complete |
-| External ordinary resumption, ephemeral restart, reconnect soak baseline | native TCP/H1, both profiles | native TLS listener | `interop.openssl.h1.resume`, `restart.ephemeral.*`, `soak.reconnect_resumption` (#511) | external peer (OpenSSL) | complete |
-| External accepted 0-RTT, unsafe 425, replay fallback, persistent-restart startup quarantine | native TCP/H1, general profile (persistent keys) | production replay gate + H1 safety gate | `interop.openssl.h1.early.accepted`, `interop.openssl.h1.early.unsafe_425`, `interop.openssl.h1.early.replay_fallback`, `restart.persistent.early_startup_quarantine` (#518/PR #525) | external peer (OpenSSL) | complete |
-| Real application-level H2 request/response on an authoritatively resumed external connection | native TCP/H2, both profiles | H2 dispatch over resumed TLS | `interop.openssl.h2.tls_resume` (#521/PR #526) | external peer (OpenSSL `h2`) + app dispatch | complete |
+| External ordinary resumption, ephemeral restart, reconnect soak baseline | native TCP/H1, appliance profile | native TLS listener | `interop.openssl.h1.resume`, `restart.ephemeral.*`, `soak.reconnect_resumption` (#511) | external peer (OpenSSL) | complete |
+| External accepted 0-RTT, unsafe 425, replay fallback, persistent-restart startup quarantine | native TCP/H1, appliance profile (persistent keys) | production replay gate + H1 safety gate | `interop.openssl.h1.early.accepted`, `interop.openssl.h1.early.unsafe_425`, `interop.openssl.h1.early.replay_fallback`, `restart.persistent.early_startup_quarantine` (#518/PR #525) | external peer (OpenSSL) | complete |
+| Real application-level H2 request/response on an authoritatively resumed external connection | native TCP/H2, appliance profile | H2 dispatch over resumed TLS | `interop.openssl.h2.tls_resume` (#521/PR #526) | external peer (OpenSSL `h2`) + app dispatch | complete |
 | External H3 1-RTT resumption + accepted/unsafe/replay QUIC 0-RTT production composition | QUIC/H3, general profile | `http3_runtime.Runtime` composition | `h3interop.quic.resume`, `h3interop.quic.early.accepted`, `h3interop.quic.early.unsafe_425`, `h3interop.quic.early.replay_fallback` (#522/PR #527) | external peer (ngtcp2/GnuTLS `gtlsclient`) | complete |
 | Persistent restart, N→N+1 SIGHUP rotation, live validity boundaries, failed-reload atomicity, QUIC credential/ticket two-phase reload atomicity, certificate-binding incompatibility | native TCP/H1 (appliance profile) + QUIC/H3 (general profile, credential-atomicity case) | `ReloadableKeyRing`, `gateway_shutdown.hotReloadConfig`, `native_credentials.prepareReloadFromFiles` | `restart.persistent.ticket_survives`, `rotation.persistent.n_to_n_plus_1`, `rotation.persistent.failed_reload_keeps_old_state`, `rotation.persistent.quic_credential_reload_atomicity`, `rotation.persistent.certificate_binding_change` (#519/PR #528) | process (real restarts/SIGHUPs) | complete |
-| Concurrent multi-process nonce safety, rotation soak, lease exhaustion, process-local replay scope | native TCP/H1 persistent keys, general profile | `ticket_key_snapshot.reserveNonceLeasesInFile` sidecar lock | `soak.persistent.multi_process_nonce_safety`, `soak.persistent.nonce_lease_exhaustion`, `soak.replay.process_local_scope` (#520/PR #529) | process (real OS processes) | complete |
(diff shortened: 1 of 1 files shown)


docs/RESUMPTION_TEST_PLAN.md
```diff
@@ -1,12 +1,176 @@
...
+| Concurrent multi-process nonce safety, rotation soak, lease exhaustion, process-local replay scope | native TCP/H1 persistent keys, general profile | `ticket_key_snapshot.reserveNonceLeasesInFile` sidecar lock | `soak.persistent.multi_process_nonce_safety`, `soak.persistent.nonce_lease_exhaustion`, `soak.replay.process_local_scope` (#520/PR #529) | process (real OS processes) | complete |
+| Ticket expiry (negative, external) | native TCP/H1 | — | `interop.openssl.ticket.expired` (#511) | external peer | complete |
+| SNI mismatch (negative, external) | native TCP/H1, appliance profile | — | `interop.openssl.sni_mismatch` (#511) | external peer | complete |
```
Jaruso: [P1] **These profile assignments do not match current `main`, so the authoritative matrix is overclaiming coverage.** PR #525 explicitly removed the appliance-profile rejection of `TARDIGRADE_TLS_NATIVE_TICKET_KEYS_PATH` so the persistent/stateless H1 restart-quarantine path could run there, and the relevant native TCP tests are gated by `requireNativeTlsProfile()` (which skips the OpenSSL-adapter/general build). #526's resumed-H2 proof and #529's multi-process persistent-key/replay soak were likewise validated under `-Dtls-profile=appliance`.

Concretely, these rows need correction:

- #511 ordinary native H1 resumption/restart/soak: **appliance profile**, not “both profiles”.
- #518 external H1 0-RTT + persistent restart quarantine: **appliance profile**, not “general profile”.
- #521 resumed H2 application proof: **appliance profile**, not “both profiles”.
- #520 multi-process persistent-key/replay soak: **appliance profile**, not “general profile”.
- The #513 row should not say persistent keys are general-only; #525 deliberately made the persistent key path usable by the appliance native path too.

I would also split the conflated #513 row so the transport boundary is unambiguous, e.g.:

```md
| Production native TCP/H1 record 0-RTT carrier | native TCP/H1, appliance profile | `NativeTlsConnection` 0-RTT record path | `#510 native tcp production 0-rtt reaches h1 safety gate and replay fallback` (#513) | process | complete |
| Persistent stateless/hybrid ticket-key composition | shared native resumption runtime; exercised by appliance H1 and general-profile H3 where applicable | `TARDIGRADE_TLS_NATIVE_TICKET_KEYS_PATH` / `ReloadableKeyRing` | #513 + composed #519/#520 proofs | production/process | complete |
```

Then scope the #511/#518/#521/#520 rows to appliance explicitly. The general-profile H3 proofs remain correctly separate in the #523/#527 rows.

docs/RESUMPTION_TEST_PLAN.md
```diff
@@ -1,12 +1,176 @@
...
+| H3 SETTINGS early-data incompatibility (negative) | QUIC/H3, general profile | — | remembered-SETTINGS compatibility coverage (`http3_runtime.zig`) (#523) | deterministic | intentionally unreachable externally — see limitations below |
+| QUIC transport-parameter early-data incompatibility (negative) | QUIC/H3, general profile | — | `quicEarlyDataTransportCompatible` deterministic coverage (#523) | deterministic | intentionally unreachable externally — see limitations below |
+| H3 replay-store-unavailable fail-closed state (negative) | QUIC/H3, general profile | `http3_runtime.zeroRttCarrierEnabled` | deterministic fail-closed carrier-disabled coverage (#523) | deterministic | intentionally unreachable as a distinct external state — see limitations below |
+
+Every #369 acceptance criterion maps to one of the rows above, either as a
+named deterministic/process/external proof or as an explicit
+unsupported/unreachable capability with the reason recorded in the section
```
Jaruso, line 44: [P1] **The sentence “Every #369 acceptance criterion maps to one of the rows above” is not true yet.** #369 has two explicit closeout ACs that are not represented by any row in this table:

- “Soak jobs show bounded behavior and no crash/UAF/deadlock/secret leak; heavy jobs remain scheduled/manual while a stable smoke subset runs routinely.”
- “Repro artifacts contain enough public context to diagnose failures without exposing ticket keys, PSKs, raw identities, or decrypted session state.”

Those requirements are discussed later under `CI and reproducibility`, but #522 specifically requires the **final matrix** to map every #369 AC. Please add explicit rows, for example:

```md
| Bounded smoke + scheduled/manual heavy soak; no crash/UAF/deadlock/secret leak | resumption suite / appliance + H3 general where applicable | bounded process/socket harness + existing soak workflow | `test-integration-resumption-interop`, `.github/workflows/resumption-soak.yml`, #511/#520 | CI + process soak | complete |
| Secret-safe reproducibility artifacts | all resumption/interop transports | bounded artifacts + cache leak checks + typed/public metadata only | interop secret-file helpers, CI cache-leak checks, landed secret-absence assertions | CI / assurance | complete |
```

The exact wording can differ, but both explicit parent ACs need first-class matrix mappings before this document can truthfully be the authoritative #369 closeout.

docs/RESUMPTION_TEST_PLAN.md
```diff
@@ -1,12 +1,176 @@
...
+  SETTINGS and QUIC transport-parameter snapshots to be compatibility-safe
+  (`quicEarlyDataTransportCompatible`, RFC 9001 §4.6.1 / RFC 9000 §7.4.1), on
+  top of everything 1-RTT already checks.
+
+A session can therefore legitimately resume at 1-RTT while its remembered
+application/transport state is no longer eligible for 0-RTT — that is a
+real, working fallback shape, not a defect. `h3interop.quic.resume` proves
+exactly this fallback path is usable independent of 0-RTT eligibility.
+
+## Process-local replay semantics
```
Jaruso, line 121: [P1] **`h3interop.quic.resume` does not prove the compatibility-mismatch fallback described by the preceding sentence.** In current `main`, that test reconnects with the same saved session/TP state and sets `disable_early_data = true`; it proves authoritative ordinary H3 1-RTT resumption and application dispatch, but it never creates incompatible remembered H3 SETTINGS or QUIC transport parameters.

That distinction matters because this same document correctly says those mismatch states are **externally unreachable** through today's production configuration surface. Please avoid crediting the external test with proof that only the deterministic compatibility tests can provide.

Suggested replacement:

```md
A session can therefore legitimately resume at 1-RTT while its remembered
application/transport state is no longer eligible for 0-RTT. The actual
SETTINGS/transport-mismatch cases are externally unreachable through today's
operator configuration and remain covered by #523's deterministic compatibility
tests. `h3interop.quic.resume` independently proves that ordinary H3 1-RTT
resumption and real application traffic work when early data is not attempted.
```

That preserves the important contract without overstating what #527 exercised.

docs/RESUMPTION_TEST_PLAN.md
```diff
@@ -1,12 +1,176 @@
...
+  `interop.`/`restart.`/`soak.`/`rotation.` prefixed cases; required on macOS
+  (where the full integration suite is skipped) and not duplicated on the
+  Linux job that already runs the unfiltered suite.
+- `.github/workflows/resumption-soak.yml` runs the same filtered suite with
+  `TARDIGRADE_SOAK_HEAVY=1` on a weekly schedule plus manual dispatch.
+- The H3 interop cases require a built ngtcp2/GnuTLS `gtlsclient` peer
+  (`scripts/interop/build-h3-peer-ci.sh`); CI builds and pins it, and those
+  cases skip locally when it isn't built.
+- Every subprocess/socket wait across this suite is bounded (poll-based, not
+  raw blocking `accept()`/`read()`), and session/transport-state files use
```
Jaruso: [P2] The CI description is now incomplete/misleading after #527. The appliance-profile filtered resumption suite is indeed used on macOS because the full appliance integration suite is skipped there, but **H3 external interop has its own Linux PR gate**: `.github/workflows/ci.yml::h3-resumption-interop` builds the pinned peer and runs `zig build test-integration-resumption-interop -Dtls-profile=general` on every PR.

Please distinguish the two gates instead of saying the filtered suite is “not duplicated on the Linux job” without qualification. For example:

```md
- Appliance-profile `test-integration-resumption-interop` runs on macOS where the full appliance integration suite is skipped; Linux appliance CI gets the same cases through the unfiltered integration suite.
- General-profile H3 resumption/0-RTT interop runs in the dedicated Linux `h3-resumption-interop` PR job, which builds the pinned external peer and sets `H3_INTEROP_CLIENT_PATH`.
```

That makes the final reproducibility section reflect the actual current gate topology.

Jaruso (commented):
I reviewed this as the final #522/#369 closeout against current `main`, the reconciled #522 scope, #369's explicit acceptance criteria, and the landed #525/#526/#527/#528/#529 behavior.

The overall reorganization is the right shape: making one final matrix authoritative and demoting the slice-by-slice history is exactly what #522 called for. The historical stale-claim annotations are also generally handled well, and CI is green at this head.

I do not think this is ready to close #522/#369 yet, because the authoritative section currently contains three material overclaims/factual mismatches:

1. Several transport/profile rows are wrong relative to current `main` (especially after #525 deliberately enabled persistent ticket keys in the appliance profile). The final matrix cannot call the #511/#518/#521/#520 native TCP proofs “both profiles” or “general profile” when those cases are `requireNativeTlsProfile()`/appliance-profile proofs.
2. The matrix says every #369 AC maps to a row, but #369's explicit soak-tier/boundedness AC and reproducibility-artifact/secret-safety AC are only described later and have no matrix rows. Since this is the authoritative AC mapping, add those rows explicitly.
3. `h3interop.quic.resume` is being credited with proving a compatibility-mismatch fallback it does not exercise. It disables early data and proves ordinary 1-RTT resumption; the SETTINGS/transport mismatch cases are correctly documented elsewhere in this PR as externally unreachable.

I also left a non-blocking-but-important CI wording fix: current CI has a dedicated Linux/general-profile H3 interop job in addition to the appliance-profile macOS filtered run, so the final reproducibility section should say that explicitly.

Once those are corrected, I don't see another substantive #522 implementation gap in this diff.

Jaruso (commented):
Re-review at `89733f8d15994d45a4c7f53c2532c030113f23ae`: the follow-up commit addresses all four findings from my first pass. The authoritative matrix now scopes the H1/H2/persistent-key proofs to the profiles they actually exercise, adds first-class mappings for #369's bounded smoke/heavy-soak and secret-safe reproducibility ACs, no longer over-credits `h3interop.quic.resume` with an externally unreachable compatibility-mismatch proof, and documents the dedicated general-profile H3 PR gate accurately. I walked the revised matrix back across #522 and the parent #369 acceptance criteria and did not find another implementation or documentation gap.

From a content/acceptance-criteria standpoint, this is ready to close #522 and #369.

One GitHub-mechanics note before merge: the PR description still contains only `Closes #522`. If the intent is for this merge to auto-close the parent too, add `Closes #369` to the PR body (otherwise close #369 manually after merge).

The new-head CI run is still in progress as of this review; the dedicated `H3 resumption/0-RTT interop` job has already passed, along with format/unit/security and multiple build jobs.

PR #533, opened 2026-07-29
quic: wire local CID lifecycle into H3 runtime

Jaruso:
## Summary

Refs #387. Implements the Slice 3 local-CID lifecycle integration for the native QUIC/H3 runtime.

- add transactional caller-generated local CID issuance so runtime routing is reserved before NEW_CONNECTION_ID can be emitted
- let Connection own active local CID state after handshake, accept all active local DCIDs, process RETIRE_CONNECTION_ID, and retransmit the same NEW_CONNECTION_ID frame identity on loss
- track every runtime-owned CID route per connection, reconcile retired CIDs, and remove all owned routes on teardown

## Validation

## Notes

This intentionally does not close #387. Remaining slices still need production Retry admission, CID-role split for Retry, operator retry/migration policy wiring, metrics, docs, and the broader UDP acceptance tests.

### DIFF
--- a/src/http/http3_runtime.zig
+++ b/src/http/http3_runtime.zig
@@ -99,6 +99,7 @@ pub const Config = struct {
 const max_connections: usize = 1024;
 const max_connections_per_source: u32 = 32;
 const handshake_timeout_us: u64 = 10 * std.time.us_per_s;
+const cid_generation_retries: usize = 16;
 
 const ParkedH3Retry = struct {
     stream_id: u64,
@@ -154,6 +155,8 @@ const ConnEntry = struct {
     /// Monotonic microseconds when the connection was accepted; used to reap
     /// connections that never complete the handshake.
     accepted_at_us: u64,
+    owned_cids: [quic.cid.max_local_active_cids]quic.cid.ConnectionId = undefined,
+    owned_cid_count: usize = 0,
     /// Set exactly once this connection has attempted post-handshake ticket
     /// issuance (successfully or not) — issuance is best-effort and must
     /// never be retried on the same connection (#488).
@@ -190,6 +193,7 @@ pub const Runtime = struct {
     /// `zero_rtt_enabled` gate and whether `accept()` installs a server
     /// early-data policy on new connections' TLS backends.
     zero_rtt_enabled: bool,
+    stateless_reset_key: [32]u8,
     quic_config: quic.config.Config,
     h3_settings: http3.frame.Settings,
     h3_application_compat: [http3.early_data.encoded_snapshot_len]u8 = undefined,
@@ -245,6 +249,7 @@ pub const Runtime = struct {
             .resumption_runtime = cfg.resumption_runtime,
             .early_data_replay_gate = cfg.early_data_replay_gate,
             .zero_rtt_enabled = zeroRttCarrierEnabled(cfg),
+            .stateless_reset_key = undefined,
             .quic_config = quicConfigFrom(cfg),
             .h3_settings = cfg.h3_settings,
             .early_data_compat_metrics_ctx = cfg.early_data_compat_metrics_ctx,
@@ -256,6 +261,7 @@ pub const Runtime = struct {
             .snapshot_state = .{ .quic_port = cfg.quic_port },
             .stopping = std.atomic.Value(bool).init(false),
         };
+        compat.randomBytes(&runtime.stateless_reset_key);
         http3.frame.validateLocallySupportedSettings(runtime.h3_settings) catch return error.InvalidH3Settings;
         runtime.h3_application_compat_len = (http3.early_data.encodeSettingsSnapshot(
             runtime.h3_settings,
@@ -339,6 +345,8 @@ pub const Runtime = struct {
                 while (it.next()) |kv| {
                     const entry = kv.value_ptr.*;
                     entry.conn.onTimeout(now);
+                    self.syncCidRoutes(entry, &routes);
+                    self.maintainLocalCidRoutes(entry, kv.key_ptr.*, &routes);
                     while (entry.conn.pollTransmitOnPath(&out, now)) |t| {
                         self.sendDatagram(sockaddrInFromAddress(t.path.remote), t.bytes);
                     }
@@ -451,6 +459,8 @@ pub const Runtime = struct {
             if (freshly_accepted) self.removeConnection(connections, routes, per_ip, found);
             return;
         };
+        self.syncCidRoutes(entry, routes);
+        self.maintainLocalCidRoutes(entry, found, routes);
         const authenticated = entry.conn.metrics.packets_received > packets_before;
         switch (classifyIngest(freshly_accepted, authenticated, !active_before.remote.eql(ingress_remote))) {
             // A just-accepted connection whose first datagram authenticates
@@ -568,6 +578,7 @@ pub const Runtime = struct {
             .tls = backend.backend(),
             .now_us = now,
             .initial_path = .{ .local = self.local_address, .remote = addressFromSockaddrIn(peer) },
+            .stateless_reset_key = self.stateless_reset_key,
             .events = .{ .context = self, .emitFn = quicConnectionEvent },
         }) catch {
             allocator.destroy(backend);
@@ -595,6 +606,8 @@ pub const Runtime = struct {
             allocator.destroy(entry);
             return null;
         };
+        entry.owned_cids[0] = cid;
+        entry.owned_cid_count = 1;
         routes.insert(cid, handle) catch {
             entry.deinit(allocator);
             allocator.destroy(entry);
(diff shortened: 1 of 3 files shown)


src/quic/connection.zig
```diff
@@ -885,6 +895,47 @@ pub const Connection = struct {
...
+    pub fn needsLocalCid(self: *const Connection) bool {
+        if (self.state_ != .established) return false;
+        const registry = if (self.local_cids) |*registry| registry else return false;
+        const target = @min(@as(usize, 2), @as(usize, @intCast(registry.active_limit)));
+        return registry.activeCount() < target;
+    }
+
+    pub fn advertiseLocalCid(self: *Connection, cid_value: quic_cid.ConnectionId) error{ CidLimitExceeded, DuplicateCid, OutOfMemory }!void {
+        const registry = if (self.local_cids) |*registry| registry else return error.CidLimitExceeded;
+        const ncid = registry.issueCid(cid_value) catch |err| switch (err) {
```
Jaruso, line 932: **Blocking — queued is not the same as previously sent.** `issueCid()` advances `LocalCidRegistry.next_sequence` here, before this frame is even appended, let alone committed to an outbound packet. `retire()` currently accepts any sequence `< next_sequence`, so a peer can send `RETIRE_CONNECTION_ID(1)` while seq 1 is only queued/backpressured and we accept it even though RFC 9000 §19.16 requires `PROTOCOL_VIOLATION` for a sequence greater than any **previously sent**. There is a second atomicity problem: if the following `append()` OOMs, the `errdefer retire()` removes the entry but does not restore `next_sequence`, so the next wire-visible CID can jump from 0 to 2.

Please separate the reservation/next-sequence state from an advertised/sent watermark, and make the queue mutation non-fallible after sequence allocation. For example:

```zig
try self.pending_new_connection_ids.ensureUnusedCapacity(self.allocator, 1);
const ncid = try registry.issueCid(cid_value);
self.pending_new_connection_ids.appendAssumeCapacity(ncid);
```

Then mark `ncid.sequence` sent only when the first packet carrying it is committed by `pollTransmitOnPath`, and validate peer retirement against that sent watermark (sequence 0 starts advertised). Add a test that queues seq 1 without transmitting it and verifies `RETIRE_CONNECTION_ID(1)` closes; after actually transmitting seq 1, retiring it from a packet addressed to a different CID should succeed.

src/quic/connection.zig
```diff
@@ -2826,6 +2912,14 @@ pub const Connection = struct {
             record.ack_eliciting = true;
             _ = self.pending_retires.orderedRemove(0);
         }
+        while (self.pending_new_connection_ids.items.len > 0) {
+            const ncid = self.pending_new_connection_ids.items[0];
+            const n = frame.encodeNewConnectionId(ncid, plain[plain_len..budget]) catch break;
+            plain_len += n;
+            record.ack_eliciting = true;
+            record.carried_new_connection_id = ncid;
```
Jaruso: **Blocking — this loss record can only remember the last NCID in the packet.** The surrounding `while` can encode/remove multiple `pending_new_connection_ids`, but each iteration overwrites `record.carried_new_connection_id`. If the packet is declared lost, `requeueRecord()` restores only the final frame; every earlier `NEW_CONNECTION_ID` in that packet is forgotten.

Either make this an `if` so one NCID is emitted per packet (matching the single `SentRecord` slot), or change `SentRecord` to a bounded array/count and requeue all carried frames. Please add a test that queues at least two NCIDs, loses the packet, and verifies the same sequence/CID/reset-token identity for **both** frames is still pending afterward.

src/http/http3_runtime.zig
```diff
@@ -627,14 +640,53 @@ pub const Runtime = struct {
...
+            for (0..cid_generation_retries) |_| {
+                compat.randomBytes(&entropy);
+                const candidate = quic.cid.generateCid(&entropy, @intCast(entry.cid_len)) catch return;
+                if (routes.contains(candidate)) continue;
+                cid_value = candidate;
+                break;
+            }
+            const cid = cid_value orelse return;
+            routes.insert(cid, handle) catch return;
+            entry.conn.advertiseLocalCid(cid) catch {
```
Jaruso: This is the critical Slice-3 production seam, but the PR adds no runtime tests for it. The direct `Connection` tests do not exercise route ownership/order at all. Before merge, please add focused runtime tests around `maintainLocalCidRoutes`/`syncCidRoutes` proving at least:

```text
initial route exists
handshake establishes -> spare route is inserted
only then NEW_CONNECTION_ID becomes pollable
collision -> regenerate/rollback without stealing the existing route
RETIRE(spare or initial, from another active DCID) -> old route removed + replacement route registered
lookup(retired CID) == null
removeConnection() -> every owned route is gone
```

Also cover a future/unadvertised retirement as `PROTOCOL_VIOLATION`. These are explicit Slice-3 acceptance cases from #387, not optional hardening.

Jaruso (commented):
**Verdict: please address these before merge.** GitHub does not allow this connected account to formally request changes on its own PR, so I am leaving the review as `COMMENT`; the findings below are still blocking from my review perspective.

I reviewed all three changed files against the current #387 Slice 3 handoff, including the local-CID lifecycle requirements deliberately left after #514/#516.

The overall direction is correct: the runtime reserves a route before calling `advertiseLocalCid`, `Connection` initializes `LocalCidRegistry` from the authenticated peer `active_connection_id_limit`, ingress accepts every active local CID, `RETIRE_CONNECTION_ID` is checked against the packet's DCID, the runtime tracks all owned CID routes, teardown removes every owned route, and the process-lifetime reset key is threaded into the server binding. CI is green as well.

I found three blockers:

1. **The registry currently conflates "allocated/queued" with "previously sent" CID sequences.** RFC 9000 §19.16 requires `PROTOCOL_VIOLATION` when a peer retires a sequence greater than any sequence previously sent. `issueCid()` advances `next_sequence` before the frame is wire-visible, while `retire()` uses that same watermark. A queued/backpressured sequence can therefore be retired before we ever advertised it. The same ordering also means an OOM appending the frame consumes a sequence number that never reaches the wire; the next successful `NEW_CONNECTION_ID` can skip that sequence even though newly issued CID sequence numbers must increase monotonically one-by-one. Please separate reservation/pending state from a `largest_sent_sequence` (or equivalent advertised state), and reserve queue capacity before mutating the registry so the local operation is actually transactional.

2. **A packet can carry multiple `NEW_CONNECTION_ID` frames but `SentRecord` remembers only one.** The loop overwrites `carried_new_connection_id`; loss requeues only the last frame. Either intentionally emit at most one NCID per packet or give `SentRecord` bounded storage/count and requeue every encoded frame. Add a loss test with at least two queued CIDs so this cannot regress.

3. **The required Slice-3 runtime proof is missing.** The two new `Connection` tests are useful, but they never exercise `owned_cids`, `syncCidRoutes`, or `maintainLocalCidRoutes`. The authoritative #387 handoff explicitly requires tests proving: route registration before advertisement, collision regeneration/rollback, valid retirement removes its route and replenishes the spare, invalid future retirement closes with `PROTOCOL_VIOLATION`, a retired CID no longer routes, and teardown removes every local CID route. Those behaviors are the production-integration purpose of this slice and need coverage before merge.

Suggested shape for the sequence-allocation fix:

```zig
try self.pending_new_connection_ids.ensureUnusedCapacity(self.allocator, 1);
const ncid = try registry.issueCid(cid_value);
self.pending_new_connection_ids.appendAssumeCapacity(ncid);
```

Then maintain a distinct sent/advertised watermark that advances only when the first `NEW_CONNECTION_ID` carrying that sequence is committed to an outbound transmit. `RETIRE_CONNECTION_ID` validation should compare against that watermark, not merely `next_sequence`.

For the loss-record issue, the smallest safe fix with the current target of one spare is simply to emit one NCID per packet:

```zig
if (self.pending_new_connection_ids.items.len > 0) {
    const ncid = self.pending_new_connection_ids.items[0];
    // encode...
    record.carried_new_connection_id = ncid;
    _ = self.pending_new_connection_ids.orderedRemove(0);
}
```

A bounded array in `SentRecord` is also fine if coalescing several NCIDs is intentional.

Once those are fixed and the runtime lifecycle tests are added, the remaining implementation looks consistent with the Slice-3 boundary; Retry admission/policy/CID-role separation and broader UDP acceptance correctly remain in the later #387 slice.

src/quic/connection.zig
Jaruso: Re-reviewed on `e9b5480a11b1eb830c314f97006a6ebdec808d42` — addressed. `advertiseLocalCid()` now reserves queue capacity before mutating the registry, `LocalCidRegistry` has a separate `largest_advertised_sequence`, and the watermark advances only when the NCID is committed into an outbound packet. The new tests cover both sides of the boundary: an unadvertised sequence is rejected with `PROTOCOL_VIOLATION`, while the same sequence is retireable after transmission from a packet using another active CID. This resolves the blocker.

src/quic/connection.zig
Jaruso: Re-reviewed on `e9b5480a11b1eb830c314f97006a6ebdec808d42` — addressed. The builder now emits at most one `NEW_CONNECTION_ID` per packet, matching the single `SentRecord.carried_new_connection_id` slot, and the added two-CID loss test verifies that the sent frame is requeued with identical sequence/CID/token while the second queued CID is preserved. This resolves the blocker.

src/http/http3_runtime.zig
Jaruso: Re-reviewed on `e9b5480a11b1eb830c314f97006a6ebdec808d42` — addressed. The runtime harness now covers route-before-advertise, collision/no-route-theft behavior, retired-route removal, spare replenishment, and teardown of all owned routes. Combined with the new Connection tests for future/unadvertised retirement and valid retirement, this satisfies the Slice-3 production-integration proof I was asking for.

Jaruso (commented):
Second-pass review complete on `e9b5480a11b1eb830c314f97006a6ebdec808d42`.

**Verdict: merge-ready for the Slice 3 scope.** I re-checked the full three-file PR, the new review-fix commit, the authoritative #387 Slice-3 acceptance list, and the previously landed migration behavior. I found no new blocking correctness issues.

All three prior blockers are addressed:

- local CID allocation is now transactional with queue capacity reserved before registry mutation, and retirement validity is checked against a distinct advertised-sequence watermark rather than merely allocated sequence state;
- `NEW_CONNECTION_ID` emission is limited to one frame per packet, so the single `SentRecord` identity is sufficient and loss preserves every queued CID;
- runtime-level coverage now exercises route-before-advertise, collision/no-route-theft behavior, retirement route removal, spare replenishment, and teardown of all owned routes, while direct Connection tests cover unadvertised/future retirement and valid retirement semantics.

The remaining Slice-3 requirements are also present: sequence 0 + spare maintenance from the peer's authenticated active-CID limit, acceptance of packets addressed to an active spare CID, process-lifetime reset-key wiring into the runtime/server binding, same-length CSPRNG CID generation, route ownership tracking, and existing fresh-peer-CID host migration behavior left intact.

Latest CI run `1035` is green.

GitHub will not allow this connected account to submit an `APPROVE` review on its own PR, so this is recorded as `COMMENT`, but from my review perspective there are no remaining blockers to merging this slice. #387 should remain open for Slice 4 (Retry admission/CID-role split/operator policy/metrics/broader UDP acceptance).

PR #534, opened 2026-07-29
tls: integrate non-PSK HelloRetryRequest through the shared backend

Jaruso:
## Summary
- Wires the HRR codec, `Core` transitions, and negotiation primitives added by PR #483 into `Tls13Backend`, so a full non-PSK TLS 1.3 handshake can complete through exactly one HelloRetryRequest, in both client and server roles, over both record and QUIC transports.
- Adds `InitialKeyShareMode` (`normal`/`empty`) so a client can send a legal empty `key_share` vector, exercising a native server's HRR path before #335 adds a second key-exchange group.
- `drainInput` peeks the HRR sentinel / `retry_state` before dispatching, routing HRR ServerHellos and ClientHello2 through `Core`'s dedicated `acceptHelloRetryRequest`/`recordSecondClientHello`/`recordHelloRetryRequest`/`acceptSecondClientHello` transitions instead of the ordinary accept/send path.
- Client (`onHelloRetryRequest`): decodes the HRR against retained ClientHello1, builds ClientHello2 (fresh X25519 share or byte-identical `key_share`, cookie echoed exactly when present), self-validates via `hello_retry.validateSecondClientHello` before emission, and records/emits it — never deriving a secret at the HRR.
- Server (`emitHelloRetryRequest`): reacts to a `.retry` negotiation decision, optionally consults a new `HelloRetryCookieProvider` seam (no provider ⇒ group-only HRR stays valid), and emits the HRR at the Initial epoch; ClientHello2 is validated against ClientHello1 *before* any peer metadata (negotiated tuple, ALPN, transport extension, SNI) is overwritten, and a second retry decision is `IllegalParameter` rather than a second HRR.
- Retry retention reuses the server's existing `client_hello_psk` capture buffer for the retained ClientHello1 bytes (the two uses are never live at the same time, on either role) instead of embedding a second `max_message_len`-sized buffer, keeping `Tls13Backend` under its existing size budget test.
- Ephemeral key hygiene: the client destroys ClientHello1's superseded key pair (`wipeEphemeral`) before generating ClientHello2's fresh share from a new `Entropy.retry_key_share_seed`; the server generates its one key pair only after ClientHello2 is validated.

## Test plan
- [x] `zig build test` — 2808/2809 passed (1 pre-existing skip, unrelated to this change)
- [x] `zig build test-tls` — 792/792 passed, including new HRR coverage
- [x] `zig build test-quic` — 588/588 passed
- [x] `zig build test-integration` — passed (some cases skip in this sandboxed environment, consistent with existing networking-sandbox limitations, not caused by this change)
- [x] New tests: record-mode HRR round trip, QUIC/extension-profile HRR round trip (including transport-extension carry-through), an external-style hand-crafted ClientHello1 driving the server directly into HRR, missing-`key_share` still `MissingExtension` (regression), a second retry decision rejected as `IllegalParameter`, retry state cleared on both success and failure paths, a cookie provider's create/release/validate all exercised through a full handshake, client-side cookie-only HRR preserving the original share, client-side requested-group HRR emitting a fresh share derived from the new retry seed, and pre-`start()` rejection of both an ordinary and an HRR-shaped ServerHello.

Closes #484

### DIFF
--- a/src/tls/tls13_backend.zig
+++ b/src/tls/tls13_backend.zig
@@ -1651,6 +1651,13 @@ pub const Tls13Backend = struct {
         self.retry.wipe();
         self.wipeRetryKeyShareSeed();
         self.client_hrr_selection = null;
+        // #484: a terminal failure anywhere between generating an ephemeral
+        // key pair (the initial-flight one, or the fresh one
+        // `onHelloRetryRequest` generates for ClientHello2) and
+        // `onServerHello`'s own success-path `wipeEphemeral()` call must
+        // still destroy that private key here — it must not linger until
+        // some later `deinit()`.
+        self.wipeEphemeral();
         if (self.schedule) |*schedule| schedule.wipe();
         self.schedule = null;
         self.resumption_master_secret.deinit();
@@ -3529,9 +3536,17 @@ pub const Tls13Backend = struct {
     /// provider's own buffer right after, on every path (no provider
     /// configured, the provider declines, or a later step fails).
     fn takeHelloRetryCookie(self: *Tls13Backend, selected_group: ?tls_algorithms.NamedGroup) HandshakeError!?[]const u8 {
-        defer self.hello_retry_cookie_provider.release();
         const provided = try self.hello_retry_cookie_provider.create(selected_group);
+        // #484: the provider only transfers ownership of a buffer when it
+        // actually returns one — `createFn` returning `null` or erroring
+        // means there was never an acquisition, so `releaseFn` must not run
+        // in either case (a provider that only marks ownership on a
+        // successful non-null return could double-release/assert/free
+        // stale state otherwise). Once `bytes` is non-null, ownership has
+        // transferred, so release must run regardless of what happens next
+        // in this function (including the bounds check below).
         const bytes = provided orelse return null;
+        defer self.hello_retry_cookie_provider.release();
         if (bytes.len == 0 or bytes.len > self.retry.cookie.len) return error.CredentialProviderFailed;
         @memcpy(self.retry.cookie[0..bytes.len], bytes);
         self.retry.cookie_len = bytes.len;
@@ -5086,7 +5101,13 @@ test "transport profile validation fails before lifecycle or transcript advance"
         try std.testing.expectEqual(tls_handshake_codec.HandshakeLifecycle.idle, backend.core.handshake_lifecycle);
         try std.testing.expectEqual(@as(usize, 0), sink.len);
         try std.testing.expect(!backend.key_pair_present);
-        try std.testing.expectEqualSlices(u8, &entropy.key_share_seed, &backend.entropy.key_share_seed);
+        // #484: `clearFailedHandshakeState`'s `errdefer` now wipes ephemeral
+        // key-generation material on every terminal failure, including one
+        // this early (before a key pair is ever generated from it) — once
+        // this backend has failed, the seed is no longer needed and is
+        // zeroed proactively rather than left resident until `deinit`.
+        try std.testing.expect(std.mem.allEqual(u8, &backend.entropy.key_share_seed, 0));
+        try std.testing.expect(std.mem.allEqual(u8, &backend.entropy.retry_key_share_seed, 0));
         backend.deinit();
     }
 
--- a/src/tls/tls13_backend.zig
+++ b/src/tls/tls13_backend.zig
@@ -579,6 +579,12 @@ pub const Tls13Backend = struct {
     /// default (`createFn == null`) means no cookie is offered and
     /// group-only HRR stays valid — see `HelloRetryCookieProvider`.
     hello_retry_cookie_provider: HelloRetryCookieProvider = .{},
+    /// Client (#484): the version/cipher/group committed to when this
+    /// connection accepted a HelloRetryRequest, retained through
+    /// ClientHello2 so the final ServerHello can be checked against it.
+    /// `null` until an HRR is accepted; cleared once the final ServerHello
+    /// has been checked. See `ClientHrrSelection`.
+    client_hrr_selection: ?ClientHrrSelection = null,
     core: tls_handshake_codec.Core,
     schedule: ?KeySchedule = null,
     resumption_master_secret: crypto_pkg.secrets.FixedSecret(session.max_psk_len) = .{},
@@ -835,6 +841,18 @@ pub const Tls13Backend = struct {
             self.cookie_len = 0;
             self.request = null;
         }
+    };
+
+    /// Client (#484): the RFC 8446 §4.1.4 "final ServerHello consistency"
+    /// fields a connection committed to when it accepted a
+    /// HelloRetryRequest. Deliberately does not include `Request.cookie` —
+    /// a borrowed slice into the HRR's transient input bytes, and not
+    /// something the final ServerHello ever echoes, so there is nothing to
+    /// check it against here.
+    pub const ClientHrrSelection = struct {
+        selected_version: tls_algorithms.ProtocolVersion,
+        cipher_suite: tls_algorithms.CipherSuite,
+        selected_group: ?tls_algorithms.NamedGroup,
     };
 
     /// Server (#484): optional cookie contribution to an outgoing
@@ -1511,8 +1529,10 @@ pub const Tls13Backend = struct {
         self.clearClientHelloPsk();
         self.retry.wipe();
         self.hello_retry_cookie_provider = .{};
+        self.client_hrr_selection = null;
         crypto.secureZero(u8, &self.expected_client_verify);
         self.wipeEphemeral();
+        self.wipeRetryKeyShareSeed();
         self.wipeIdentity();
         crypto.secureZero(u8, &self.peer_chain);
         self.peer_chain_count = 0;
@@ -1550,6 +1570,17 @@ pub const Tls13Backend = struct {
         crypto.secureZero(u8, &self.entropy.key_share_seed);
         crypto.secureZero(u8, std.mem.asBytes(&self.key_pair));
         self.key_pair_present = false;
+    }
+
+    /// Client (#484): zeroes the HelloRetryRequest-only key-share seed.
+    /// Deliberately separate from `wipeEphemeral` — on the retry path,
+    /// `wipeEphemeral` runs *before* this seed is consumed (to destroy
+    /// ClientHello1's superseded key pair), so folding this into it would
+    /// zero the seed before it could be used. Called immediately after the
+    /// fresh key pair is generated from it, and unconditionally on
+    /// teardown so an unconsumed seed never outlives the connection.
+    fn wipeRetryKeyShareSeed(self: *Tls13Backend) void {
+        crypto.secureZero(u8, &self.entropy.retry_key_share_seed);
     }
 
     fn wipeIdentity(self: *Tls13Backend) void {
@@ -1618,6 +1649,8 @@ pub const Tls13Backend = struct {
         self.early_data_discard_limit = 0;
         self.clearClientHelloPsk();
         self.retry.wipe();
+        self.wipeRetryKeyShareSeed();
+        self.client_hrr_selection = null;
         if (self.schedule) |*schedule| schedule.wipe();
         self.schedule = null;
         self.resumption_master_secret.deinit();
@@ -1987,12 +2020,6 @@ pub const Tls13Backend = struct {
     // -----------------------------------------------------------------------
 
     fn sendClientHello(self: *Tls13Backend, sink: *EventSink) HandshakeError!void {
-        var key_pair = X25519.KeyPair.generateDeterministic(self.entropy.key_share_seed) catch
-            return error.SecretExportFailed;
-        defer crypto.secureZero(u8, &key_pair.secret_key);
-        self.key_pair = key_pair;
-        self.key_pair_present = true;
-
         // Sized for the worst case, not the common one: maximum ALPN (255,
         // bounded by `TransportProfile.validate`), maximum SNI
         // (`max_server_name_len` = 256), and a maximum transport extension
@@ -2046,6 +2073,12 @@ pub const Tls13Backend = struct {
         try w.u16_(ext_key_share);
         switch (self.initial_key_share_mode) {
             .normal => {
+                var key_pair = X25519.KeyPair.generateDeterministic(self.entropy.key_share_seed) catch
+                    return error.SecretExportFailed;
+                defer crypto.secureZero(u8, &key_pair.secret_key);
+                self.key_pair = key_pair;
+                self.key_pair_present = true;
+
                 try w.u16_(2 + 2 + 2 + X25519.public_length);
                 try w.u16_(2 + 2 + X25519.public_length); // client_shares
                 try w.u16_(group_x25519);
@@ -2055,7 +2088,13 @@ pub const Tls13Backend = struct {
             // #484: advertise the group in `supported_groups` above but send
             // a legal empty `client_shares` vector, so a native server
             // negotiates `.retry` and this backend can exercise
-            // HelloRetryRequest before #335 adds a second key-exchange group.
+            // HelloRetryRequest before #335 adds a second key-exchange
+            // group. Deliberately generates *no* key pair at all —
+            // `self.key_pair_present` stays `false`, so if a peer
+            // incorrectly answers with an ordinary ServerHello instead of
+            // an HRR, `onServerHello`'s existing `!key_pair_present` guard
+            // rejects it before any ECDH, rather than deriving secrets
+            // from a private key whose public half was never offered.
             .empty => {
                 try w.u16_(2); // extension_data length: just the vector length field
                 try w.u16_(0); // client_shares: zero-length vector
@@ -2410,8 +2449,22 @@ pub const Tls13Backend = struct {
         if (try r.u16_() != legacy_version) return error.IllegalParameter;
         const random = try r.slice(32);
         if (std.mem.eql(u8, random, &hello_retry_request_random)) return error.IllegalParameter;
+        // #484: this is a genuine, non-HRR ServerHello (the sentinel check
+        // above already rejected an HRR-shaped one) — no retry occurred or
+        // ever will for this connection, so the ClientHello1 capture
+        // `sendClientHello` retained for a possible HelloRetryRequest is no
+        // longer needed. For a resumed handshake this buffer holds the
+        // bearer ticket identity and its binder, so this must run here —
+        // not only on the retry path — or a successful long-lived
+        // connection would retain those bytes until `deinit`.
+        self.clearClientHelloPsk();
         const session_id_len = try r.u8_();
         _ = try r.slice(session_id_len);
+        // RFC 8446 §4.1.3: `legacy_session_id_echo` must equal the
+        // ClientHello's `legacy_session_id`, which this profile always
+        // sends empty (no TLS 1.3 compatibility-mode session echo, on
+        // either ClientHello1 or ClientHello2).
+        if (session_id_len != 0) return error.IllegalParameter;
         const selected_cipher = tls_algorithms.fromInt(tls_algorithms.CipherSuite, try r.u16_()) orelse return error.IllegalParameter;
         if (try r.u8_() != 0) return error.IllegalParameter;
 
@@ -2454,6 +2507,18 @@ pub const Tls13Backend = struct {
             .alpn = null,
         }) catch |err| return mapNegotiationError(err);
         if (selected_cipher != .tls_aes_128_gcm_sha256 or named_group != .x25519 or protocol_version != .tls13) return error.IllegalParameter;
+        // #484: the final ServerHello must remain consistent with what this
+        // connection's HelloRetryRequest actually selected — a check on top
+        // of (not a substitute for) the ordinary policy-tuple check above,
+        // which alone cannot express "matches this connection's own
+        // accepted retry" as opposed to "matches policy in general."
+        if (self.client_hrr_selection) |selection| {
+            if (protocol_version != selection.selected_version or
+                selected_cipher != selection.cipher_suite or
+                (selection.selected_group != null and named_group != selection.selected_group.?))
+                return error.IllegalParameter;
+            self.client_hrr_selection = null;
+        }
         self.negotiated_version = protocol_version;
         self.negotiated_cipher_suite = selected_cipher;
         self.negotiated_named_group = named_group;
@@ -2552,6 +2617,17 @@ pub const Tls13Backend = struct {
         // ClientHello1 always carries an empty legacy_session_id (this
         // profile does not use TLS 1.3's compatibility-mode session echo).
         const request = hello_retry.decode(body, "", &parsed.offers) catch |err| return mapCoreError(err);
+        // #484: retain the committed version/cipher/group (never the HRR's
+        // `cookie`, a borrowed slice into this call's transient `body`) so
+        // `onServerHello` can check the *final* ServerHello against exactly
+        // what this connection's HRR selected — not only against local
+        // policy, which alone cannot express "matches this connection's own
+        // accepted retry."
+        self.client_hrr_selection = .{
+            .selected_version = request.selected_version,
+            .cipher_suite = request.cipher_suite,
+            .selected_group = request.selected_group,
+        };
 
         var buf: [max_message_len]u8 = undefined;
         defer crypto.secureZero(u8, &buf);
@@ -2602,6 +2678,10 @@ pub const Tls13Backend = struct {
             defer crypto.secureZero(u8, &key_pair.secret_key);
             self.key_pair = key_pair;
             self.key_pair_present = true;
+            // #484: the retry seed is consumed exactly once per connection;
+            // zero it immediately rather than leaving it resident for the
+            // rest of the backend's lifetime.
+            self.wipeRetryKeyShareSeed();
 
             try w.u16_(2 + 2 + 2 + X25519.public_length);
             try w.u16_(2 + 2 + X25519.public_length); // client_shares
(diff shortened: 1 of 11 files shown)


Jaruso (commented):
Disposition: **changes requested before merge**. I reviewed PR #534 against #484's full implementation scope and acceptance criteria, including the HRR foundation from #483, the current PSK/0-RTT backend state, cleanup behavior, and the actual QUIC test surface. CI is green on head `75bd47d`, and the core happy-path HRR flow is substantial, but the current patch does not yet satisfy #484.

Blocking findings are called out inline. The key corrections are:

1. Validate ClientHello2 against retained ClientHello1/HRR **before** re-negotiation or mutating peer metadata. The current order allows an invalid CH2 to be classified/processed by negotiation first.
2. Retain the client's committed HRR selection until the final ServerHello and enforce version/cipher/group/session-id consistency there. The current client drops the HRR request after CH2, so that AC cannot be enforced; the existing ServerHello parser also ignores `legacy_session_id_echo`.
3. Fix secret/lifetime cleanup introduced by the new retry state: wipe `retry_key_share_seed` on use/teardown, and clear the retained CH1 capture immediately when an ordinary ServerHello proves there will be no retry. Otherwise successful resumed connections retain bearer PSK identity/binder bytes until connection teardown.
4. Implement the server-side **cookie-only** HRR path required by #484. Right now the cookie provider is consulted only after key-share negotiation already decided `.retry`, so it can only produce group+cookie HRRs.
5. Add a real QUIC `Connection`/`QuicTlsAdapter` HRR loopback. The new "QUIC-style" test only selects transport extension 57 in the direct TLS harness; the changed QUIC/H3 tests merely initialize the new entropy field and never exercise HRR through QUIC CRYPTO/key-discard behavior.
6. Make `.empty` key-share mode reject an ordinary final ServerHello before HRR/CH2 rather than using the internally generated-but-never-offered key pair.

Suggested shape for the most important ordering fix:

```zig
const parsed = try parseClientHelloObserved(...);

if (self.core.retry_state == .hrr_sent) {
    const request = self.retry.request orelse return error.InvalidHandshakeState;
    const ch1 = self.client_hello_psk orelse return error.InvalidHandshakeState;
    try hello_retry.validateSecondClientHello(ch1.message[0..ch1.message_len], raw, request);
    if (request.cookie) |cookie| try self.hello_retry_cookie_provider.validate(cookie);
}

// Only after the canonical CH2 check succeeds:
const hello_selection = try tls_negotiation.negotiateServerHello(self.policy, &parsed.offers);
// commit peer metadata, then clear retry capture/state
```

For client final-SH consistency, retain a small owned HRR selection record (version/cipher/group plus original session ID; do not retain borrowed HRR cookie slices) through CH2, compare it in `onServerHello`, then wipe it after the final SH is accepted.

For the entropy cleanup, note that `wipeEphemeral()` currently runs *before* the retry seed is consumed, so simply adding `retry_key_share_seed` to that helper would zero it too early. Zero the retry seed immediately after retry key generation and also on teardown/failure when it was never consumed.

Once these are addressed, I would re-run the full #484 negative matrix in addition to the existing green suite: invalid CH2 mutation classification, final SH mismatch after HRR, ordinary SH after empty CH1, cookie-only provider round trip, cleanup after normal/non-HRR resumption, and a genuine QUIC Initial-CRYPTO HRR loopback.

src/tls/tls13_backend.zig
```diff
@@ -498,6 +500,11 @@ const max_server_name_len = 256;
 pub const Entropy = struct {
     hello_random: [32]u8,
     key_share_seed: [X25519.seed_length]u8,
+    /// Client-only: seeds the fresh X25519 share ClientHello2 emits when a
+    /// HelloRetryRequest requests a group. Never consumed by the server,
+    /// which generates at most one key pair per connection (for the final
+    /// ServerHello, after any retry has already been resolved).
+    retry_key_share_seed: [X25519.seed_length]u8,
```
Jaruso, line 507: **Blocking — retry entropy is never zeroized.** This new seed is secret key-generation material, but `wipeEphemeral()` only clears `entropy.key_share_seed` and the current `key_pair`. On the retry path, `wipeEphemeral()` is intentionally called *before* `retry_key_share_seed` is consumed, and neither the subsequent final-ServerHello path nor `deinit()` clears this field. That leaves the retry seed resident for the backend lifetime (and servers also receive a never-used copy). Please zero it immediately after generating CH2's key pair, and explicitly zero it on terminal teardown/failure when unused. A separate `wipeRetryKeyShareSeed()` is safer than adding it to `wipeEphemeral()` because the latter currently runs before retry generation. Add a teardown regression with distinct initial/retry seed bytes.

src/tls/tls13_backend.zig
```diff
@@ -776,6 +798,83 @@ pub const Tls13Backend = struct {
...
+    /// 460-byte wire cap.
+    pub const max_cookie_len = 64;
+
+    /// #484: bounded HelloRetryRequest retry state, retained by both roles
+    /// from ClientHello1 until the retry completes (client: ClientHello2 is
+    /// recorded; server: ClientHello2 is validated) or the handshake fails.
+    /// `request`/`cookie` are server-only — the committed HRR parameters
+    /// needed to validate ClientHello2; the client re-derives the
+    /// equivalent `hello_retry.Request` synchronously from the HRR it just
+    /// decoded and never needs to retain it.
```
Jaruso, line 819: **Blocking — the client does need to retain the committed HRR selection.** #484 requires the final ServerHello to remain consistent with the HRR (version, cipher suite, requested group, and session-ID echo). After `onHelloRetryRequest()` returns, this implementation retains only `core.retry_state`; `onServerHello()` validates against the general policy and currently discards `legacy_session_id_echo` without comparing it. With no durable HRR selection, the required final-SH comparison cannot be made (the current single-suite/single-group policy merely hides most of the mismatch cases). Please retain an owned small selection record through ClientHello2, validate the final ServerHello against it, then wipe it. Do not retain `Request.cookie` as a borrowed slice into the input buffer.

src/tls/tls13_backend.zig
```diff
@@ -2024,6 +2179,18 @@ pub const Tls13Backend = struct {
...
+        // `RetryContext`'s doc comment; this field is otherwise never
+        // touched by the client role) in case the server responds with a
+        // HelloRetryRequest instead of the ordinary ServerHello. Wiped once
+        // ClientHello2 is safely recorded, or by
+        // `clearFailedHandshakeState`/`deinit` on any failure path.
+        var ch1_capture: ClientHelloPskCapture = .{};
+        if (message.len > ch1_capture.message.len) return error.MalformedHandshake;
+        @memcpy(ch1_capture.message[0..message.len], message);
+        ch1_capture.message_len = message.len;
+        self.client_hello_psk = ch1_capture;
```
Jaruso, line 2276: **Blocking lifetime regression:** every client now retains an exact copy of ClientHello1 in `client_hello_psk`, but the ordinary (non-HRR) `onServerHello()` success path never clears it. For resumed handshakes this buffer includes the bearer ticket identity and binder, so a successful long-lived connection keeps those bytes until `deinit()`. That conflicts with the existing capture-lifetime discipline and #484's cleanup AC. Once the first ServerHello is identified as an ordinary final ServerHello, clear this retry-only capture immediately; the failure boundary already handles error cleanup. Please add a normal PSK/non-HRR regression asserting the capture is null after the first ordinary ServerHello / successful handshake.

src/tls/tls13_backend.zig
```diff
@@ -3031,10 +3341,32 @@ pub const Tls13Backend = struct {
...
+            },
         };
         if (selected_key_share.len != X25519.public_length) return error.MalformedHandshake;
         const client_share = selected_key_share[0..X25519.public_length].*;
+        // #484: ClientHello2 must be validated against the retained
+        // ClientHello1 and the committed HRR parameters *before* any of the
+        // peer metadata below (negotiated tuple, ALPN, transport extension,
+        // SNI, PSK capture) is overwritten with values parsed from this
+        // second hello.
+        if (self.core.retry_state == .hrr_sent) {
```
Jaruso: **Blocking — ClientHello2 is validated too late.** #484 explicitly requires the canonical CH2-vs-CH1/HRR validation *before* re-running negotiation or overwriting peer metadata. Here `parseClientHelloObserved`, `negotiateServerHello`, early-data state, signature-scheme state, and the key-share retry decision all run before this block. That means an illegal CH2 mutation can be classified by negotiation first (for example, the new test expects `MissingExtension` for a CH2 that removed the HRR-required key_share) instead of being rejected by the canonical mutation validator, and unvalidated CH2 values transiently mutate backend state. Move this validation immediately after parsing the second ClientHello and before `negotiateServerHello`/peer-state commits; only clear the retained retry context after that validation succeeds.

src/tls/tls13_backend.zig
```diff
@@ -3088,6 +3420,62 @@ pub const Tls13Backend = struct {
...
+        // PSK-capture code (further down in `onClientHello`) ever runs, so
+        // the field is always empty here.
+        var capture: ClientHelloPskCapture = .{};
+        if (ch1_raw.len > capture.message.len) return error.MalformedHandshake;
+        @memcpy(capture.message[0..ch1_raw.len], ch1_raw);
+        capture.message_len = ch1_raw.len;
+        self.client_hello_psk = capture;
+
+        defer self.hello_retry_cookie_provider.release();
+        const provided_cookie = try self.hello_retry_cookie_provider.create(group);
```
Jaruso: **Blocking AC gap — this API can never produce a cookie-only HRR.** `create()` is only reached from the `.retry` key-share branch and is always passed a non-null `group`, so a ClientHello that already has a usable share proceeds directly to ServerHello even when the configured provider wants a cookie retry. #484 explicitly requires a cookie-only path with a deterministic cookie provider, preserving the original key_share. One workable shape is: after parsing CH1, when the share is usable, consult `create(null)`; if it returns a cookie, emit HRR with `selected_group = null`; on the existing group-retry path call `create(group)`. Then add a full server-provider cookie-only round trip. The current client-only cookie test and group+cookie provider test do not cover this requirement.

src/tls/tls13_backend_tests.zig
```diff
@@ -2532,6 +2532,343 @@ fn recordOrEmpty(bytes: ?[]const u8) []const u8 {
...
+    try std.testing.expect(harness.observed.application_write_secret[0] != null);
+    try std.testing.expect(harness.observed.application_write_secret[1] != null);
+    try std.testing.expect(harness.observed.initial_discarded[0]);
+    try std.testing.expect(harness.observed.initial_discarded[1]);
+
+    try expectHrrRetryStateCleared(&harness.client_backend);
+    try expectHrrRetryStateCleared(&harness.server_backend);
+}
+
+test "#484 HRR round trip completes over the extension (QUIC-style) profile too" {
```
Jaruso, line 2598: **Blocking acceptance-test gap:** this is a direct `Tls13Backend` harness with extension type 57; it does not send HRR/CH2 through `quic.Connection`/`QuicTlsAdapter`, QUIC CRYPTO reassembly, or real Initial-key discard/install behavior. #484's AC says the shared engine must complete HRR *through QUIC Initial CRYPTO data*. The other changed QUIC/H3 tests only add `retry_key_share_seed` initializers. Please add a genuine QUIC loopback (the existing `Connection`/H3 simulation infrastructure is fine) with an `.empty` client, assert HRR and CH2 stay in Initial CRYPTO, and assert Initial keys survive HRR and are discarded only after the final ServerHello.

src/tls/tls13_backend.zig
```diff
@@ -1901,11 +2044,23 @@ pub const Tls13Backend = struct {
...
+                try w.u16_(2 + 2 + X25519.public_length); // client_shares
+                try w.u16_(group_x25519);
+                try w.u16_(X25519.public_length);
+                try w.bytes(&key_pair.public_key);
+            },
+            // #484: advertise the group in `supported_groups` above but send
+            // a legal empty `client_shares` vector, so a native server
+            // negotiates `.retry` and this backend can exercise
+            // HelloRetryRequest before #335 adds a second key-exchange group.
+            .empty => {
```
Jaruso, line 2143: **Blocking protocol edge case:** `.empty` emits no client key share, but `sendClientHello()` generated/stored an X25519 key pair unconditionally above, leaving `key_pair_present = true`. If a peer incorrectly sends an ordinary ServerHello instead of HRR, `onServerHello()` will accept the tuple and derive/install handshake secrets from this private key even though its public key was never offered, then discard Initial keys; it only fails later when Finished cannot match. A client that sent no usable share should reject that ServerHello immediately. Either do not create/store the initial key pair in `.empty` mode (and map the absence to `illegal_parameter` on an ordinary SH), or track whether the current key pair was actually emitted and require that before deriving. Add a negative test for ordinary SH immediately after an empty CH1.

Jaruso (commented):
Re-review of head `a949b8a` against #484 and the seven findings from my previous pass.

**Good progress:** the previous blockers are substantively addressed: CH2 validation now precedes re-negotiation, the client retains and checks the HRR-selected tuple, ordinary ServerHello clears the retry capture, `.empty` no longer creates an unsent key pair, cookie-only server HRR exists, retry seed zeroing was added, and there is now a genuine `quic.Connection` HRR loopback.

I found **two remaining blocking lifecycle issues**, called out inline, and CI currently has a third merge blocker:

1. `clearFailedHandshakeState()` still does not wipe the actual ephemeral `self.key_pair`. After CH2 generates the fresh retry X25519 key pair, any terminal failure before the final ServerHello leaves that retry private key resident until a later `deinit()`. #484 explicitly requires retry key material to be wiped on every error path. Add `self.wipeEphemeral()` to terminal handshake cleanup and cover a post-CH2 failure.
2. `takeHelloRetryCookie()` invokes `releaseFn` even when `createFn` returns `null` or throws. The provider only transfers ownership when it actually returns bytes; calling release without a successful acquisition can double-release/assert/free stale provider state. Pair release only with a non-null successful return.
3. Latest CI run `30419665616` fails on every test platform because the merge ref contains a newer `main` initializer in `src/http/http3_runtime.zig` that still constructs `Entropy` without `retry_key_share_seed` (`http3_runtime.zig:1313`). The builds pass, but unit/appliance test compilation fails before execution. Please merge/rebase current `main` and update that initializer, then rerun CI.

The CI failure is concrete rather than flaky: the Ubuntu log reports `error: missing struct field: retry_key_share_seed`, and the same test-compilation stage fails on macOS and ARM. Once these three items are fixed and CI is green, I do not currently see another #484 acceptance-criteria blocker in the updated implementation.

src/tls/tls13_backend.zig
```diff
@@ -1496,6 +1648,9 @@ pub const Tls13Backend = struct {
         self.early_data_max_bytes = 0;
         self.early_data_discard_limit = 0;
         self.clearClientHelloPsk();
+        self.retry.wipe();
+        self.wipeRetryKeyShareSeed();
```
Jaruso, line 1652: **Blocking — terminal failure cleanup still leaves the retry private key resident.** This now wipes `retry_key_share_seed`, but not `self.key_pair`. After `onHelloRetryRequest()` generates/stores the fresh CH2 X25519 pair, any later terminal error before `onServerHello()` reaches its normal `wipeEphemeral()` path (for example CH2 sink emission failure, malformed/mismatched final ServerHello, or a transport error) leaves that private scalar in backend storage until some later `deinit()`. #484 requires retry key material to be cleaned on every error path. Since this helper is explicitly the terminal handshake-failure boundary, add `self.wipeEphemeral()` here as well. A useful regression is: drive a valid HRR/CH2 so `key_pair_present == true`, then feed a final ServerHello that fails consistency/structure and assert `key_pair_present == false` and the stored key/seed bytes are zeroed immediately after the failed `receive()`.

src/tls/tls13_backend.zig
```diff
@@ -3088,6 +3520,70 @@ pub const Tls13Backend = struct {
...
+    /// Server (#484): asks the configured cookie provider for a cookie —
+    /// `selected_group == null` means "the client's offered share is
+    /// already usable; would you like a cookie-only retry anyway?",
+    /// non-null means "this group needs a retry regardless; would you also
+    /// like to attach a cookie?" — copies any returned bytes into the
+    /// durable `self.retry.cookie` immediately, and releases the
+    /// provider's own buffer right after, on every path (no provider
+    /// configured, the provider declines, or a later step fails).
+    fn takeHelloRetryCookie(self: *Tls13Backend, selected_group: ?tls_algorithms.NamedGroup) HandshakeError!?[]const u8 {
+        defer self.hello_retry_cookie_provider.release();
```
Jaruso: **Blocking ownership bug — `releaseFn` runs even when no cookie was acquired.** The `defer` is installed before `create()`, so it fires when `createFn` returns `null` *and* when `createFn` returns an error. But the provider contract says the returned bytes are provider-owned until release; if there are no returned bytes, there is no acquisition to release. A provider that allocates/marks ownership only on a successful non-null return can double-free, assert, or release stale state here. Pair the release with the successful acquisition instead, e.g. `const provided = try ...create(...); const bytes = provided orelse return null; defer self.hello_retry_cookie_provider.release(); ...`. Add tests for a declining provider and an erroring provider with `release_calls == 0`, plus the existing successful case with exactly one release.

src/tls/tls13_backend.zig
```diff
@@ -1687,8 +1849,25 @@ pub const Tls13Backend = struct {
...
+            // cannot tell them apart on its own — it only tracks message
+            // kind and `retry_state`. The wire bytes must be peeked first so
+            // the matching dedicated `Core` transition runs instead of
+            // `acceptReceived`.
+            const is_hello_retry_request = self.role == .client and message.kind == .server_hello and
+                self.core.retry_state == .none and hello_retry.isHelloRetryRequest(message.body);
+            const is_second_client_hello = self.role == .server and message.kind == .client_hello and
+                self.core.retry_state == .hrr_sent;
+            if (is_hello_retry_request) {
+                _ = self.core.acceptHelloRetryRequest(message.raw) catch |err| return mapCoreError(err);
```
Jaruso, line 1871: **Blocking — validate the HRR before mutating `Core`.** #484's client flow explicitly requires `hello_retry.decode()`/validation against retained ClientHello1 *before* `Core.acceptHelloRetryRequest()`. This call is not just an ordering check: `Core.acceptHelloRetryRequest()` immediately calls `transcript.rebindClientHello()`, appends the HRR bytes, and sets retry state. `onHelloRetryRequest()` only afterward checks the HRR's session-id echo, selected cipher/version/group, extension legality, etc. A ServerHello with the HRR sentinel but illegal HRR semantics therefore gets committed into the rebound transcript before being rejected. Please pre-decode/validate here (or in a helper called here), then perform the Core transition and pass the validated `Request` into the handler. Add a regression that records the transcript hash before an invalid sentinel-shaped HRR and proves the backend rejects it without committing the HRR/rebound transcript.

Jaruso (commented):
Disposition: **changes still requested before merge**, but this is now narrowly scoped. The previous re-review blockers are fixed: terminal failures wipe ephemeral key material, cookie release now follows actual acquisition, the rebase/main initializer issue is fixed, and CI is green on `f6e9c677`.

The remaining issue is the ordering at the shared-engine boundary. #484 is explicit on both sides:

- client: **decode/validate HRR against ClientHello1, then call `Core.acceptHelloRetryRequest()`**;
- server: **validate ClientHello2 against ClientHello1/HRR, then call `Core.acceptSecondClientHello()`, then re-run negotiation/replace peer metadata**.

`drainInput()` currently performs both dedicated `Core` transitions first. Those transitions are not read-only: `acceptHelloRetryRequest()` rebinds and updates the transcript; `acceptSecondClientHello()` updates the transcript and advances handshake state. Semantic validation happens only afterward in the backend handlers. Therefore a sentinel-shaped but illegal HRR, or an illegal CH2 mutation, is committed into protocol/transcript state before it is rejected. On the CH2 side `onClientHello()` also clears retained peer metadata at entry before its validator runs.

I left two inline comments on those transition sites with the suggested fix. The cleanest shape is a backend preflight before the `Core` mutation, e.g. conceptually:

```zig
var hrr_request: ?hello_retry.Request = null;
if (is_hello_retry_request) {
    hrr_request = try self.decodeHelloRetryRequest(message.body); // validates against retained CH1
    _ = try self.core.acceptHelloRetryRequest(message.raw);
} else if (is_second_client_hello) {
    try self.validateSecondClientHelloBeforeCommit(message.raw); // canonical mutation + cookie
    _ = try self.core.acceptSecondClientHello(message.raw);
} else {
    _ = try self.core.acceptReceived(message.raw);
}

try self.onMessage(..., hrr_request, ...);
```

The handler can then consume the already-validated HRR request / already-validated CH2 without duplicating validation. This also naturally ensures the `onClientHello()` resets and metadata commits happen only after CH2 is proven legal.

I also left one test comment: #484 explicitly requires a **repeated HRR rejection** at backend level. The test named "accepts exactly one" currently sends only the first HRR; the foundation `Core` test is useful but does not exercise this backend dispatch path. Add a second HRR after CH2, assert a fatal rejection and no third ClientHello/secret transition.

With those ordering fixes and the repeated-HRR regression, I do not see another #484 acceptance-criteria gap in the current patch. The record loopback, real QUIC `Connection` loopback, cookie-only path, final-SH consistency, key lifetime/cleanup, and current regression suite all look aligned with the ticket.

src/tls/tls13_backend.zig
```diff
@@ -1687,8 +1849,25 @@ pub const Tls13Backend = struct {
...
+            } else if (is_second_client_hello) {
+                _ = self.core.acceptSecondClientHello(message.raw) catch |err| return mapCoreError(err);
```
Jaruso, line 1874: **Blocking — the server has the symmetric ordering problem for ClientHello2.** `Core.acceptSecondClientHello()` updates the transcript and advances the handshake state before `onClientHello()` runs `validateSecondClientHello()`. #484 requires the opposite order: canonical CH2-vs-CH1/HRR validation, then `Core.acceptSecondClientHello()`, then negotiation/peer-metadata replacement. This is also why the current claim that validation happens before peer state mutation is not fully true: `onClientHello()` resets `peer_sig_scheme_count`, SNI presence/length, and early-data state at entry, after this Core mutation but before its validator. Preflight the canonical mutation + cookie validation here while the retained CH1/request are intact; only on success call `acceptSecondClientHello()` and dispatch to `onClientHello()`. A useful regression is to snapshot the transcript hash and retained peer metadata after HRR, feed an invalid CH2 mutation, and assert none of them were committed before the fatal error.

src/tls/tls13_backend_tests.zig
```diff
@@ -2532,6 +2532,606 @@ fn recordOrEmpty(bytes: ?[]const u8) []const u8 {
...
+    try std.testing.expect(harness.client_driver.isComplete());
+    try std.testing.expect(harness.server_driver.isComplete());
+    try std.testing.expectEqual(@as(usize, 1), fixture.create_calls);
+    try std.testing.expectEqual(@as(usize, 1), fixture.release_calls);
+    try std.testing.expectEqual(@as(usize, 1), fixture.validate_calls);
+    try expectHrrRetryStateCleared(&harness.client_backend);
+    try expectHrrRetryStateCleared(&harness.server_backend);
+}
+
+test "#484 client accepts exactly one cookie-only HelloRetryRequest and echoes the original key share unchanged" {
```
Jaruso, line 2765: **Required negative case is still missing.** This test is named "accepts exactly one" but only sends one HRR; it never proves the backend rejects a repeated HRR after ClientHello2. #484's required tests explicitly list "repeated HRR and second retry decisions are rejected." The server-side second-retry case is covered separately, but the client dispatch path is not. After the first `receive()` has emitted CH2, feed another HRR-shaped ServerHello and assert the fatal error plus `countCryptoEvents(.initial) == 2` (no ClientHello3 / no additional secret transition). This is particularly useful because a second HRR is intentionally routed differently once `retry_state != .none`.

src/tls/tls13_backend.zig
```diff
@@ -3088,6 +3557,78 @@ pub const Tls13Backend = struct {
...
+    /// Server (#484): asks the configured cookie provider for a cookie —
+    /// `selected_group == null` means "the client's offered share is
+    /// already usable; would you like a cookie-only retry anyway?",
+    /// non-null means "this group needs a retry regardless; would you also
+    /// like to attach a cookie?" — copies any returned bytes into the
+    /// durable `self.retry.cookie` immediately, and releases the
+    /// provider's own buffer right after, on every path (no provider
+    /// configured, the provider declines, or a later step fails).
+    fn takeHelloRetryCookie(self: *Tls13Backend, selected_group: ?tls_algorithms.NamedGroup) HandshakeError!?[]const u8 {
+        const provided = try self.hello_retry_cookie_provider.create(selected_group);
```
Jaruso, line 3569: **Non-blocking doc nit:** this sentence is stale after the ownership fix. It currently says the provider buffer is released on every path, including "no provider configured" and "provider declines", but the implementation and new regression correctly establish the opposite: `releaseFn` runs only after `createFn` returns a non-null buffer; null/error means no acquisition and therefore no release. Please adjust this comment to match the actual provider contract so a future provider implementation does not infer the old semantics.

Jaruso (commented):
Disposition: **ready to merge**.

Re-reviewed latest head `f1697a7` against #484's complete acceptance criteria and every blocker raised in the prior three passes. I do not see a remaining correctness, security, lifecycle, transport, or test-coverage blocker.

The final ordering fix is the important piece: `drainInput` now preflights an HRR against retained ClientHello1 before `Core.acceptHelloRetryRequest()` can rebind/update the transcript, and preflights ClientHello2 against the retained CH1/HRR request before `Core.acceptSecondClientHello()` can update transcript/state. The new regressions explicitly prove rejected HRR/CH2 input leaves those Core mutations uncommitted. The required repeated-HRR negative case is also now present.

I rechecked the rest of #484 as part of this pass:

- native record client/server complete exactly one HRR;
- a genuine `quic.Connection` loopback carries HRR and CH2 through Initial CRYPTO and keeps Initial keys until final ServerHello;
- CH2 is derived from retained CH1 and canonical mutation validation runs before semantic/state commitment;
- final ServerHello is checked against the accepted HRR selection and session-id requirements;
- HRR itself does not derive/install/discard traffic secrets;
- empty-share mode cannot derive from an unoffered key;
- fresh retry key material, retained CH1, cookie state, and ephemeral keys are cleaned on success/failure/teardown;
- cookie-only and group+cookie retry paths are covered, including provider decline/error ownership behavior;
- repeated HRR/second-retry behavior is rejected;
- non-HRR and PSK regression paths remain covered;
- latest CI run is fully green across unit/appliance/build/security/interop jobs.

I left one **non-blocking documentation nit** inline: `takeHelloRetryCookie`'s function comment still says release occurs when the provider declines/no provider is configured, which contradicts the corrected and tested ownership contract (release only after a non-null buffer is returned). That should be cleaned up, but it does not affect runtime correctness.

Because the PR author is the authenticated user, I am leaving this as a COMMENT review rather than a formal APPROVE. From a technical review standpoint, **#484 is satisfied and PR #534 is merge-ready**.

PR #535, opened 2026-07-29
Implement QUIC Retry policy for HTTP/3

Jaruso:
## Summary
- add `TARDIGRADE_HTTP3_RETRY_POLICY=off|address_validation` and wire it through edge config/gateway into the native HTTP/3 runtime
- implement stateless QUIC Retry admission with process-lifetime token keys, validated-token CID role split, invalid-token rejection, and Retry/path snapshot counters
- map disabled HTTP/3 migration to NAT-rebinding-only, document the operator contract, and reject Retry in the appliance TLS profile

Refs #387

## Validation


src/http/http3_runtime.zig
```diff
@@ -506,6 +526,11 @@ pub const Runtime = struct {
         const credential_provider = self.credential_provider orelse return null;
         if (parsed.version != quic.packet.quic_v1) return null;
         if (parsed.dcid.len < 8 or parsed.scid.len == 0) return null;
+        const retry_context: ?quic.path.RetryContext = switch (self.classifyRetry(routes, parsed, peer, now)) {
```
Jaruso, line 532: **BLOCKER — wrong-version Retry tokens bypass `invalid_tokens`.** This return happens before `classifyRetry()`, so a non-empty, otherwise-valid Retry token replayed in an Initial with a mismatched/unsupported QUIC version is dropped here without `noteInvalidToken()`. The #387 Slice 4 state machine and UDP rejection matrix explicitly require wrong-version Retry tokens to allocate no connection **and increment `invalid_tokens`**.

Please move the version decision into the Retry classifier (or otherwise ensure non-empty tokens are classified before this early return) while still avoiding a Retry response for a tokenless unsupported-version Initial. For example, the classifier can do roughly:

```zig
if (self.retry_policy == .off) {
    return if (parsed.version == quic.packet.quic_v1) .accept_unvalidated else .drop;
}
if (parsed.token.len == 0) {
    if (parsed.version != quic.packet.quic_v1) return .drop;
    self.issueRetry(...);
    return .drop;
}
const ctx = self.retry_tokens.validateRetry(...) catch {
    self.noteInvalidToken();
    return .drop;
};
if (parsed.version != quic.packet.quic_v1 or
    ctx.quic_version != parsed.version or
    !std.mem.eql(u8, parsed.dcid, ctx.retry_scid.slice()))
{
    self.noteInvalidToken();
    return .drop;
}
```

Add a regression test that creates a valid v1 token, presents it in an Initial with a different `parsed.version`, and asserts zero allocation plus `invalid_tokens == 1`.

src/quic/connection.zig
```diff
@@ -223,14 +223,23 @@ pub const CloseInfo = struct {
...
     peer_cid: []const u8 = &.{},
+    /// The client Initial DCID before any Retry. Servers bind this into
+    /// `original_destination_connection_id`; clients use it to verify the
+    /// server binding after the handshake.
+    original_destination_cid: []const u8 = &.{},
+    /// Backwards-compatible alias for callers not yet migrated. When
+    /// `original_destination_cid` is empty, this supplies the same value.
+    original_dcid: []const u8 = &.{},
+    /// Server-only after Retry: the SCID the server used in its Retry packet.
+    retry_source_cid: ?[]const u8 = null,
```
Jaruso, line 251: **BLOCKER — this compatibility alias preserves the CID-role overload Slice 4 was specifically meant to remove.** `resolveOriginalDestinationCid()` falls back to `original_dcid`, and `resolveInitialSecretDcid()` then falls back through that same value, so any unmigrated caller still has one field meaning both the pre-Retry ODCID and the Initial-secret DCID. The authoritative #387 contract makes `original_destination_cid` and `initial_secret_dcid` explicit/required and says: “Do not overload one field for more than one of these meanings again.”

Please remove the alias/fallback and migrate the remaining `Connection.init()` call sites. Non-Retry callers should be explicit even when the values are equal:

```zig
.original_destination_cid = &odcid,
.initial_secret_dcid = &odcid,
```

The post-Retry runtime keeps the split already implemented here:

```zig
.original_destination_cid = ctx.original_dcid.slice(),
.retry_source_cid = ctx.retry_scid.slice(),
.initial_secret_dcid = parsed.dcid,
```

That makes the type/API itself enforce the invariant instead of relying on callers knowing when the compatibility fallback is unsafe.

src/http/http3_runtime.zig
```diff
@@ -1076,10 +1170,42 @@ pub const Runtime = struct {
         self.snapshot_state.requests_completed += 1;
     }
 
-    fn noteMigrationEvent(self: *Runtime) void {
+    fn foldPathMetrics(self: *Runtime, entry: *ConnEntry) void {
+        const current = entry.conn.pathMetrics();
+        const previous = entry.last_path_metrics;
```
Jaruso, line 1209: **BLOCKER — counters are wired, but the required path transition diagnostics are still missing.** #387 Slice 4.6 explicitly calls for logging validation started/succeeded/failed and migration blocked/promoted with bounded path/change/reason context. `foldPathMetrics()` gives aggregate deltas only; it cannot provide the path or blocked reason, and `quicConnectionEvent()` still intentionally ignores path lifecycle events other than the existing 0-RTT variants.

Please extend the connection event seam (or an equivalent structured seam) for the transition points, e.g.:

```zig
path_validation_started: struct { path: PathKey, change: AddressChange },
path_validation_failed: struct { path: PathKey },
path_migration_blocked: struct { path: PathKey, reason: enum { policy, no_peer_cid } },
path_validated: PathValidatedEvent,
```

Then log those state transitions at an appropriate bounded level in the runtime. Keep token/key/reset-token/CID material out of the log and do not add packet-by-packet logging. The metric folding here can remain as-is for counters.

src/http/http3_runtime.zig
```diff
@@ -3521,6 +3659,221 @@ test "accept() (#523): wires an EventSink into every accepted connection so 0-RT
...
+        .family = posix.AF.INET,
+        .port = std.mem.nativeToBig(u16, port),
+        .addr = @bitCast([4]u8{ 127, 0, 0, 1 }),
+        .zero = [_]u8{0} ** 8,
+    };
+}
+
+test "http3 runtime Retry sends tokenless Initials without allocating connection state" {
+    var fixed = tls_core.credentials.FixedCredentialProvider.init(tls_core.credentials.testdata.identity());
+    defer fixed.deinit();
```
Jaruso, line 3728: **BLOCKER — these direct `accept()` tests do not satisfy the Slice 4 real-UDP acceptance criteria.** They prove pieces of the classifier, but bypass the production `serve() -> recvfrom() -> DCID routing -> accept/ingest -> sendto()` path and never prove a native client actually consumes the Retry and completes H3.

#387 explicitly requires the production socket matrix in/alongside `tests/quic_h3_udp_smoke.zig`:

- Retry off flood: caps/zero-auth cleanup/listener responsiveness;
- Retry on flood: tokenless Initial gets Retry while tracked state remains zero, no Retry loop;
- full Retry success: native client receives Retry, echoes token, completes TLS/H3, serves one request, CID bindings/metrics correct;
- rejection: tampered, expired, wrong source/port, wrong version, wrong Retry SCID -> zero allocation + invalid-token metric;
- spoofed known-DCID rebind cannot redirect or start validation;
- validated same-IP NAT rebinding promotes to socket B and preserves recovery state;
- validation timeout leaves old path active and increments failure metric;
- host migration proof (in-memory acceptable if second loopback is not portable);
- retired/teardown CID routes stop routing and leak nothing.

Please add these socket/runtime acceptance tests rather than relying only on private-method tests. This is the final production proof for the internet-facing #387 blocker.

Jaruso (commented):
**Blocking review from my perspective.** GitHub will not allow this account to submit `REQUEST_CHANGES` on its own PR, so this is posted as `COMMENT`, but I would not merge #535 yet.

I reviewed #535 end-to-end against the authoritative Slice 4 handoff on #387 and current `main` after #533. The implementation is directionally strong: the production accept path now has Retry state, tokenless Initials are handled before TLS/QUIC/H3 allocation, valid tokens feed the correct post-Retry CID roles, migration policy is finally composed, path counters are folded after ingest and timeout, and operator/docs wiring is present. CI is green.

I still have four blockers before I would merge this as the final #387 slice:

1. Wrong-version retried Initials bypass Retry invalid-token accounting because `accept()` rejects non-v1 before `classifyRetry()`. #387 requires wrong-version Retry-token rejection to increment `invalid_tokens` while allocating no state.
2. `Connection.Options.original_dcid` remains as a compatibility fallback that still supplies both the ODCID binding and Initial-secret DCID when the new fields are omitted. The Slice 4 contract explicitly requires eliminating this semantic overload, not preserving it behind resolver helpers. Make `original_destination_cid` and `initial_secret_dcid` explicit and migrate the remaining call sites.
3. Slice 4 requires bounded path lifecycle transition logging (validation started/succeeded/failed, blocked by policy/no peer CID, promoted). This PR publishes counters but still has no equivalent transition/event logging for those states.
4. The required real UDP/runtime acceptance matrix is still missing. The new tests call `Runtime.accept()` directly and do not exercise the production `serve()/recvfrom()/route/sendto()` path or a native client consuming Retry. #387 explicitly requires the UDP scenarios in `tests/quic_h3_udp_smoke.zig`: Retry flood/state-zero behavior, full Retry -> TLS/H3 request success, all rejection variants, spoofed rebind, validated NAT rebinding, validation timeout, host migration proof, and CID retirement/teardown.

Please address these before merging. I included concrete suggestions inline. Once these are fixed, I will re-review the complete delta rather than only checking the comments.

src/http/http3_runtime.zig
Jaruso: Verified on current head `1cd6ada`: fixed. Version handling now lives in `classifyRetry()`, non-empty tokens are classified before unsupported-version rejection, and the regression test proves a valid v1 token replayed under a different version allocates no state and increments `invalid_tokens`. Resolving this thread.

src/quic/connection.zig
Jaruso: Verified on current head `1cd6ada`: fixed. `Connection.Options` now requires distinct `original_destination_cid` and `initial_secret_dcid`, the `original_dcid` compatibility alias/fallback is gone, and the touched production/test/interoperability call sites were migrated explicitly. Resolving this thread.

src/http/http3_runtime.zig
Jaruso: Partially addressed on `1cd6ada`: the event seam now has started/succeeded/failed/blocked/promoted variants and the runtime logs them with bounded change/reason/family context. I’m keeping this thread open because the implementation currently emits `.path_migration_blocked` for every new authenticated packet on an already policy-blocked path, so the new warning becomes packet-by-packet logging—the exact behavior this requirement said to avoid. I left a new inline comment with the de-dup/state-transition fix expected.

src/http/http3_runtime.zig
Jaruso: Partially addressed on current head `1cd6ada`: the new real-UDP tests now prove (1) tokenless Retry traffic allocates no tracked state and (2) a native client consumes Retry and completes a real H3 request through the production runtime. Those are good additions.

I’m keeping this thread open because the authoritative #387 socket/runtime matrix still has uncovered cases: Retry-off flood/cap cleanup, production-socket rejection for tampered/expired/wrong-source-or-port/wrong-version/wrong-Retry-SCID tokens, spoofed known-DCID rebinding, validated same-IP NAT rebinding with recovery preservation, validation timeout with old path retained/failure metric, host-migration proof, and retired/teardown CID route behavior. The underlying unit/Connection tests are useful, but #387 explicitly required these final runtime-boundary acceptance proofs.

src/quic/connection.zig
```diff
@@ -1201,8 +1227,18 @@ pub const Connection = struct {
...
+                .probe => |probe| {
+                    self.queueCandidateChallenge(ingress_path, probe.data);
+                    self.events.emit(.{ .path_validation_started = .{ .path = ingress_path, .change = probe.change } });
+                },
+                .blocked => {
+                    const change = if (ingress_path.remote.sameHost(self.paths.activePath().key.remote))
+                        quic_path.AddressChange.nat_rebinding
+                    else
+                        quic_path.AddressChange.migration;
+                    self.events.emit(.{ .path_migration_blocked = .{ .path = ingress_path, .change = change, .reason = .policy } });
```
Jaruso: **BLOCKER — this turns the requested transition diagnostic into packet-by-packet warning emission.** `PathManager.onDatagram()` returns `.blocked` every time an authenticated datagram arrives on a policy-disallowed path, including when that path was already recorded as `.unvalidated`. This branch therefore emits `.path_migration_blocked` on every such packet, and the runtime logs each event at `warn`.

That violates the #387 observability requirement to log bounded state transitions rather than packet-by-packet traffic and gives an authenticated peer a straightforward log-amplification path.

Please distinguish the *first blocked transition* from subsequent packets on the same blocked path. One clean shape would be:

```zig
pub const PathDecision = union(enum) {
    // ...
    blocked: struct {
        change: AddressChange,
        first_observation: bool,
    },
};
```

Then:

```zig
.blocked => |blocked| {
    if (blocked.first_observation) {
        self.events.emit(.{ .path_migration_blocked = .{
            .path = ingress_path,
            .change = blocked.change,
            .reason = .policy,
        } });
    }
},
```

An equivalent `.blocked_new` / `.blocked_existing` split is fine. The transport metric can continue counting attempts if that is the intended metric semantics; the **log/event must represent the transition**, not every packet.

src/http/http3_runtime.zig
```diff
@@ -627,9 +658,78 @@ pub const Runtime = struct {
...
+        }
+        const remote = addressFromSockaddrIn(peer);
+        if (parsed.token.len == 0) {
+            if (parsed.version != quic.packet.quic_v1) return .drop;
+            self.issueRetry(routes, parsed, peer, remote, now);
+            return .drop;
+        }
+        const ctx = self.retry_tokens.validateRetry(parsed.token, remote, now) catch {
+            self.noteInvalidToken();
+            self.logger.info(null, "http3: rejected invalid QUIC Retry token", .{});
```
Jaruso: **BLOCKER — Retry diagnostics are also packet-attempt logging.** Every invalid-token Initial logs here, the version/Retry-SCID mismatch path logs again below, and every tokenless Initial that receives Retry logs in `issueRetry()`. An attacker can generate these paths without creating connection state, so Retry doing its DoS-protection job can itself become unbounded info-log amplification.

#387 already gives us bounded counters (`retry_packets_sent`, `invalid_tokens`, `retry_tokens_accepted`) for these high-frequency outcomes and explicitly says not to log packet-by-packet traffic. Please remove the per-attempt logs or put them behind an explicit bounded/rate-limited aggregator. The hot paths should be more like:

```zig
const ctx = self.retry_tokens.validateRetry(parsed.token, remote, now) catch {
    self.noteInvalidToken();
    return .drop;
};

// ... mismatch
self.noteInvalidToken();
return .drop;
```

and after a Retry send:

```zig
self.noteRetryPacketSent();
```

The one-time startup log that Retry policy is enabled is useful and can stay. Path **state-transition** logs can stay once they are transition-deduplicated.

tests/h3_interop_tool.zig
```diff
@@ -63,7 +63,7 @@ fn logEvent(_: ?*anyopaque, event: connection.Event) void {
         .close_sent => |c| std.fmt.bufPrint(&buf, "close sent code={d}", .{c.error_code}),
         .close_received => |c| std.fmt.bufPrint(&buf, "close received code={d} app={}", .{ c.error_code, c.is_application }),
         .idle_timeout => std.fmt.bufPrint(&buf, "idle timeout", .{}),
-        .path_validated => |p| std.fmt.bufPrint(&buf, "path validated change={s} remote_port={d}", .{ @tagName(p.change), p.path.remote.port }),
+        .path_promoted => |p| std.fmt.bufPrint(&buf, "path promoted change={s} remote_port={d}", .{ @tagName(p.change), p.path.remote.port }),
```
Jaruso, line 70: **BLOCKER — `connection.Event` gained four new tags, but this exhaustive switch only migrated the old `path_validated` arm to `path_promoted`.** The switch is now missing `path_validation_started`, `path_validation_succeeded`, `path_validation_failed`, and `path_migration_blocked`, so `build-h3-interop` will fail when this tool is analyzed.

Please migrate the event consumer completely, preferably preserving useful verbose output:

```zig
.path_validation_started => |p| std.fmt.bufPrint(
    &buf, "path validation started change={s} remote_port={d}",
    .{ @tagName(p.change), p.path.remote.port },
),
.path_validation_succeeded => |p| /* ... */,
.path_validation_failed => |p| /* ... */,
.path_migration_blocked => |p| std.fmt.bufPrint(
    &buf, "path blocked change={s} reason={s} remote_port={d}",
    .{ @tagName(p.change), @tagName(p.reason), p.path.remote.port },
),
.path_promoted => |p| /* existing */,
```

Then explicitly run `zig build build-h3-interop`. This target is separate from the normal unit-test graph, so the current CI results do not appear to have caught the incomplete switch.

Jaruso (commented):
Second-pass review on `1cd6ada` against the full #387 Slice 4 contract.

Two prior blockers are fully fixed and I resolved those threads: wrong-version Retry tokens now reach the classifier/`invalid_tokens`, and the `Connection.Options` CID roles are now structurally distinct with all touched call sites migrated.

The path-diagnostic work is directionally correct, and the two new real-UDP tests are valuable: they now prove stateless tokenless Retry and a complete native Retry -> TLS -> H3 request. However, this is still **blocking / not merge-ready** for four reasons:

1. The new path/retry diagnostics introduce packet-attempt log amplification. Policy-blocked paths emit a warning for every authenticated packet because `PathManager.onDatagram()` returns `.blocked` on every visit, and Retry/invalid-token paths log every incoming attempt. #387 explicitly requires transition diagnostics, not packet-by-packet traffic logging. I left inline fixes for both cases.
2. The `connection.Event` expansion was not fully migrated in `tests/h3_interop_tool.zig`; its exhaustive switch is missing the four new path lifecycle variants. Build `zig build build-h3-interop` as part of the fix.
3. The real-UDP acceptance thread is only partially addressed. Still required at the production runtime boundary: Retry-off admission flood/cleanup, the full invalid-token matrix (tamper/expiry/source-port/version/Retry-SCID), spoofed rebind, validated NAT rebinding with recovery preservation, validation timeout, host migration proof, and CID retirement/teardown routing.
4. Latest CI run 1041 is red. The only observed failing job is appliance-profile macOS native-TLS integration: `native TLS listener dispatches ALPN http/1.1 through keepalive requests` timed out waiting for the second upstream request. That looks plausibly unrelated to this QUIC diff, but the head still needs a green rerun or a demonstrated unrelated/flaky disposition before merge.

Once the logging semantics, event-switch regression, remaining #387 runtime acceptance matrix, and CI are clean, I think the implementation itself is close.

src/http/http3_runtime.zig
Jaruso: Current head `082e6b7` closes most of this matrix: the real-UDP suite now covers Retry-on no-allocation, Retry success, the full five-case invalid-token matrix, spoofed known-DCID traffic, validation timeout, and validated NAT rebinding; the existing in-memory migration test covers the explicitly-allowed host-migration proof, and the runtime CID harness covers retired/teardown route ownership.

I’m keeping this thread open for one remaining acceptance problem in the new Retry-off flood test: it waits on `tracked_connections == 0`, which is already true before the listener processes any of the eight datagrams, so the test can pass without proving zero-auth cleanup at all. It also does not exercise the stated cap bound or prove that the same listener remains responsive after the flood. I left a focused inline blocker with the expected test shape.

tests/quic_h3_udp_smoke.zig
```diff
@@ -275,6 +405,315 @@ test "udp smoke: native client/server complete an H3 exchange over loopback" {
...
+        var dcid = [_]u8{0x71} ** 8;
+        dcid[7] = @intCast(i);
+        var scid = [_]u8{0x72} ** 8;
+        scid[7] = @intCast(i);
+        var packet: [128]u8 = undefined;
+        const initial = try writeSyntheticInitial(quic.packet.quic_v1, &dcid, &scid, "", &packet);
+        try client_socket.sendTo(runtime_addr, initial);
+    }
+
+    const snapshot = try waitRuntimeSnapshot(&runtime, hasNoTrackedConnections);
```
Jaruso: **BLOCKER — this Retry-off flood acceptance test is currently vacuous.** `hasNoTrackedConnections()` is only `snapshot.tracked_connections == 0`, which is already true immediately after `runtime.start()`. After the eight `sendTo()` calls, `waitRuntimeSnapshot()` can therefore return on its very first read before the serve thread has received a single datagram. The assertions below then pass even if zero-auth cleanup is broken.

This also leaves two explicit #387 scenario-1 requirements unproved: **tracked state never exceeds the global/per-source admission caps** and **the same listener stays responsive after the flood**.

At minimum, wait for evidence that the flood was actually consumed before accepting `tracked_connections == 0`, e.g.:

```zig
const attempts: usize = 8;
// send attempts...

const deadline = nowUs() + 5_000_000;
var snapshot = runtime.snapshot();
while (nowUs() < deadline and
    (snapshot.datagrams_seen < attempts or snapshot.tracked_connections != 0))
{
    var ts = std.c.timespec{ .sec = 0, .nsec = 10 * std.time.ns_per_ms };
    _ = std.posix.system.nanosleep(&ts, &ts);
    snapshot = runtime.snapshot();
}
try testing.expect(snapshot.datagrams_seen >= attempts);
try testing.expectEqual(@as(usize, 0), snapshot.tracked_connections);
```

Then prove responsiveness **on this same runtime** by establishing a genuine Retry-off client/H3 request after the flood and receiving the response. For the cap portion, drive enough authenticated half-open Initials (or an equivalent runtime-boundary setup) to make state persist and assert the runtime never exceeds `max_connections_per_source` / `max_connections`; eight immediately-unauthenticated packets cannot exercise either cap because they are removed one-by-one.

The important point is that this scenario must prove all three things #387 names: the flood was processed, admission stayed bounded/cleanup occurred, and the listener still serves a valid connection afterward.

Jaruso (commented):
Third-pass review on current head `082e6b7` against the authoritative #387 acceptance matrix.

The prior production-code blockers are fixed: blocked-path diagnostics now emit only on first observation, Retry/invalid-token hot paths no longer log per packet, and the interop event switch handles the expanded event union. I resolved those threads. The expanded UDP suite is also a substantial improvement: Retry-on stateless admission, successful Retry -> TLS/H3, all five invalid-token cases, spoofed known-DCID traffic, validation timeout, and same-IP NAT rebinding are now exercised through the real runtime socket path. Existing Connection/runtime tests cover host migration/recovery semantics and CID route retirement/teardown. CI #1045 is green on this head.

I found one remaining **blocking acceptance-test defect**, not a production implementation bug: the new Retry-off flood test can succeed before the runtime processes any flood datagram because its wait predicate is `tracked_connections == 0`, which is true at startup. Consequently it does not actually prove zero-auth cleanup; it also does not demonstrate the #387 requirements that state remains bounded by admission caps and that the same listener remains responsive after the flood. I left the exact fix on the test.

Once that Retry-off flood proof is made non-vacuous and demonstrates post-flood responsiveness/cap behavior, I do not currently see another #387 blocker in this PR.

tests/quic_h3_udp_smoke.zig
Jaruso: Verified on current head `8172fc3`: the specific false-positive called out here is fixed. The test now waits until all eight Retry-off flood datagrams have actually been observed before asserting zero tracked state, and it completes a genuine Retry-off H3 request on the same runtime afterward. Resolving this thread.

There is one separate issue in the newly-added *cap-saturation* subsection: its wait uses an absolute cumulative `datagrams_seen >= 48` threshold and the `> 0` state assertion is already satisfied by the established responsiveness connection, so that subsection can still pass without proving all 40 cap attempts were processed or that the per-source cap was actually reached. I’m leaving that as a new focused inline blocker.

tests/quic_h3_udp_smoke.zig
```diff
@@ -275,6 +413,380 @@ test "udp smoke: native client/server complete an H3 exchange over loopback" {
...
+        });
+        errdefer cap_client.deinit();
+        var out: [2048]u8 = undefined;
+        while (cap_client.pollTransmitOnPath(&out, nowUs())) |t| {
+            try client_socket.sendTo(sockaddrInFromAddress(t.path.remote), t.bytes);
+        }
+        cap_client.deinit();
+    }
+
+    const cap_snapshot = try waitRuntimeSnapshot(&runtime, processedAdmissionCapFlood);
```
Jaruso: **BLOCKER — the new admission-cap proof can still pass before the cap flood is actually processed.** `processedAdmissionCapFlood()` uses an absolute cumulative `datagrams_seen >= 48`. By the time this line runs, this same runtime has already processed the eight synthetic flood datagrams **and a complete Retry-off QUIC/TLS/H3 exchange**, so `datagrams_seen` may already be at/above 48 (or need only a subset of these 40 Initials to get there). `waitRuntimeSnapshot()` can therefore return before all cap attempts are consumed.

The state assertions are also too weak to prove saturation: the successful H3 connection above is still tracked, so `tracked_connections > 0` passes even if every cap client is dropped. `<= 32` only proves the implementation did not exceed the cap; it does not prove this test actually exercised it.

Please make the wait relative to a baseline captured immediately before this flood and assert the expected saturated state. For example:

```zig
const before_cap = runtime.snapshot();
const cap_attempts: usize = 40;
// send cap_attempts authenticated Initials...

const target_datagrams = before_cap.datagrams_seen + cap_attempts;
const deadline = nowUs() + 5_000_000;
var cap_snapshot = runtime.snapshot();
while (nowUs() < deadline and cap_snapshot.datagrams_seen < target_datagrams) {
    var ts = std.c.timespec{ .sec = 0, .nsec = 10 * std.time.ns_per_ms };
    _ = std.posix.system.nanosleep(&ts, &ts);
    cap_snapshot = runtime.snapshot();
}
try testing.expect(cap_snapshot.datagrams_seen >= target_datagrams);
// One established connection already occupies one slot; the same-source
// half-open flood should fill the remaining 31 and reject the rest.
try testing.expectEqual(@as(usize, 32), cap_snapshot.tracked_connections);
try testing.expectEqual(@as(usize, 32), cap_snapshot.native_connections);
```

I would also use a separate `cap_socket` (same 127.0.0.1, different port) for these 40 clients so their handshake responses do not pollute the established client's receive queue; the per-source admission key is IP-only, so this still exercises the same `max_connections_per_source` bucket.

Jaruso (commented):
Fourth-pass review on current head `8172fc3` against the authoritative #387 Slice 4 acceptance criteria.

The previous Retry-off cleanup/responsiveness defect is fixed: the test now waits for all eight unauthenticated flood datagrams to be observed before asserting zero tracked state, and the same runtime successfully completes a genuine Retry-off H3 request afterward. The prior production-code blockers remain fixed, and CI #1047 is green.

I found one remaining **blocking acceptance-test defect**, again in the new cap proof rather than production behavior. The cap subsection waits on an absolute cumulative `datagrams_seen >= 48` threshold even though this runtime has already processed the eight synthetic flood packets plus an entire QUIC/TLS/H3 exchange. It can therefore satisfy the threshold before all 40 cap Initials have been processed. Separately, `tracked_connections > 0` is already true because the successful H3 connection remains tracked, so the assertions do not prove that any half-open cap clients were admitted, much less that the per-source cap was actually exercised.

The fix is small: snapshot `datagrams_seen` immediately before the cap flood, wait for `before_cap.datagrams_seen + cap_attempts`, and assert the expected saturated state. Because the runtime already has one established connection from this source and `max_connections_per_source == 32`, 40 authenticated half-open Initials should leave exactly 32 tracked/native connections during the test window (1 established + 31 half-open; the rest rejected). Using a separate UDP socket for the cap clients would also keep their server responses out of the established client's receive queue while preserving the same 127.0.0.1 per-source key.

Once that cap subsection is made non-vacuous, I do not currently see another #387 blocker in PR #535.

tests/quic_h3_udp_smoke.zig
Jaruso: Verified on current head `46bbe8e`: fixed exactly as requested. The cap proof now captures a baseline immediately before the cap flood, sends the 40 authenticated Initials from a dedicated UDP socket on the same source IP, waits for `before_cap.datagrams_seen + cap_attempts`, and then requires `tracked_connections == 32` and `native_connections == 32`. Because the already-established responsiveness connection occupies one IP-scoped admission slot, this proves 31 additional half-open connections were admitted, the per-source limit was actually saturated, and the excess attempts were rejected. Resolving this blocker.

src/http/http3_runtime.zig
Jaruso: Final verification on head `46bbe8e`: the #387 runtime/socket acceptance matrix is now covered to the level specified by the issue handoff. Retry-off flood processing/cleanup is non-vacuous, post-flood responsiveness is proven on the same listener, and the IP-scoped admission cap is actually saturated and asserted at 32. Retry-on stateless admission, successful Retry -> TLS/H3, the five invalid-token rejection cases, spoofed known-DCID traffic, validation timeout, and same-IP NAT rebinding are exercised through the runtime socket path; the explicitly-allowed deterministic harness covers host migration/recovery behavior, and the runtime CID tests cover route retirement/teardown. CI #1048 is green. Resolving this umbrella acceptance thread.

Jaruso (commented):
Final re-review of PR #535 on head `46bbe8e` against #387: **merge-ready**.

The last open blocker is fixed. The Retry-off acceptance test now (1) proves the initial unauthenticated flood was actually processed and cleaned up, (2) proves the same listener still completes a genuine Retry-off QUIC/TLS/H3 request afterward, and (3) proves the IP-scoped admission cap is truly exercised by taking a baseline immediately before 40 authenticated half-open Initials, waiting for all 40 additional datagrams, and asserting the saturated state is exactly 32 tracked/native connections. Using a separate UDP socket avoids contaminating the established client's receive queue while still hitting the same IP-only admission bucket.

All previously raised review threads are now resolved. The production Retry/migration/CID-role/observability fixes remain intact from the earlier passes, the required runtime/socket acceptance matrix is covered to the level specified by the #387 handoff, and CI #1048 is green on this exact head.

I do not see another blocking correctness or acceptance-criteria gap in the code reviewed across these passes. Because the authenticated GitHub account is also the PR author, I cannot submit an `APPROVE` review; this COMMENT is my explicit merge-ready verdict.

PR #536, opened 2026-07-29
tls: make PSK binders HelloRetryRequest-aware

Jaruso:
## Summary
- Complete the PSK/resumption portion of TLS 1.3 HelloRetryRequest: ClientHello2 binders are now generated and verified over the rebound transcript (`message_hash(Hash(CH1)) || HRR || Truncate(CH2)`), not `Hash(Truncate(CH2))` in isolation.
- `src/tls/pre_shared_key.zig`: refactor around new `deriveBinderFromTranscriptHash` / `verifyBinderFromTranscriptHash` primitives that take a precomputed transcript digest; the existing `deriveBinder`/`verifyBinder` ClientHello wrappers become thin callers that hash the truncated message first.
- Client (`onHelloRetryRequest`): retains the PSK offer set across the retry and re-emits it in ClientHello2 — same identity order, freshly recomputed obfuscated ticket ages, binders derived from the rebound transcript via `Transcript.peekWith` (never mutating the live transcript), `pre_shared_key` still last. Identities are never silently dropped to fit; an oversized retry ClientHello fails locally (`HandshakeBufferOverflow`) and all retained PSK/retry state is wiped on that path.
- Server (`onClientHello` / `selectPsk`): captures the exact transcript digest each ClientHello's binder must verify against at parse time. For ClientHello2 this requires the transcript state as it existed *before* `Core.acceptSecondClientHello` appends the full message — threaded through `drainInput`/`onMessage` as a plain call argument (never stored on the backend) so `Tls13Backend` stays under its 64 KiB size budget. `selectPsk` now verifies against the captured digest instead of re-hashing the truncated message.
- `hello_retry.zig`'s ClientHello2 mutation validator already enforced PSK identity ordering and `pre_shared_key`-last from the #484 foundation — untouched here.

## Test plan
- [x] `zig build test` — full suite green (2843 tests), including the new binder-primitive, client-retry, server-retry, async-resolution, and full record-loopback PSK+HRR resumption tests.
- [x] `zig fmt --check` on all changed files.
- [x] Existing non-HRR resumption and #484 non-PSK HRR tests remain green (no regressions).

Closes #485

### DIFF
--- a/src/quic/connection.zig
+++ b/src/quic/connection.zig
@@ -3605,6 +3605,212 @@ test "driver: HelloRetryRequest completes through real QUIC Initial CRYPTO data"
     }
 }
 
+// #485: a genuine two-`Connection` loopback proving PSK resumption itself
+// (not just the non-PSK HRR retry mechanics the #484 test above covers)
+// survives a real HelloRetryRequest over the QUIC adapter — the client's
+// retained offer is re-emitted in ClientHello2 with a binder derived from
+// the rebound transcript, and the server verifies it against that same
+// digest, exactly as `src/tls/tls13_backend_tests.zig`'s direct-harness
+// tests prove at the TLS-backend level.
+test "driver: PSK resumption completes through a real HelloRetryRequest over QUIC" {
+    const allocator = testing.allocator;
+
+    const LoggedEvent = union(enum) { packet_sent: PacketNumberSpace, keys_discarded: PacketNumberSpace };
+    const EventCapture = struct {
+        log: [64]LoggedEvent = undefined,
+        len: usize = 0,
+
+        fn onEvent(ctx: ?*anyopaque, event: Event) void {
+            const self: *@This() = @ptrCast(@alignCast(ctx.?));
+            const logged: LoggedEvent = switch (event) {
+                .packet_sent => |p| .{ .packet_sent = p.space },
+                .keys_discarded => |space| .{ .keys_discarded = space },
+                else => return,
+            };
+            if (self.len == self.log.len) return;
+            self.log[self.len] = logged;
+            self.len += 1;
+        }
+
+        fn initialPacketsSentBeforeInitialDiscard(self: *const @This()) usize {
+            var count: usize = 0;
+            for (self.log[0..self.len]) |entry| {
+                switch (entry) {
+                    .packet_sent => |space| if (space == .initial) {
+                        count += 1;
+                    },
+                    .keys_discarded => |space| if (space == .initial) return count,
+                }
+            }
+            return count;
+        }
+    };
+    var client_capture = EventCapture{};
+    var server_capture = EventCapture{};
+
+    const psk = [_]u8{0x64} ** tls_core.tls13_backend.hash_len;
+    var common: tls_core.session.ResumableSessionCommon = .{};
+    try common.init(allocator, tls_core.session.Limits.default, .{
+        .cipher_suite = .tls_aes_128_gcm_sha256,
+        .resumption_psk = &psk,
+        // QUIC negotiates "h3", not the record profile's "h2" — must match
+        // what `onEncryptedExtensions` actually negotiates or the resumed
+        // connection is fatally rejected as an ALPN mismatch, not merely
+        // falling back to a full handshake.
+        .application_protocol = "h3",
+        .auth_binding = tls_core.session.AuthBinding.fromLeafCertificateDer(tls_backend_mod.testdata.certificate_der),
+        .issued_at_unix_ms = 0,
+        .lifetime_seconds = 3600,
+    });
+    var ticket: tls_core.session.ClientTicketState = .{};
+    try ticket.init(allocator, tls_core.session.Limits.default, &common, .{
+        .ticket = "quic-hrr-resumption-ticket",
+        .ticket_age_add = 0,
+        .ticket_nonce = "n",
+        .received_at_unix_ms = 0,
+    });
+    var offers: tls_core.pre_shared_key.ClientPskOfferSet = .{};
+    try offers.push(&ticket);
+    var clock_dummy: u8 = 0;
+    const Clock = struct {
+        fn now(_: *anyopaque) i64 {
+            return 5_000;
+        }
+    };
+
+    var stored_common: tls_core.session.ResumableSessionCommon = .{};
+    try stored_common.init(allocator, tls_core.session.Limits.default, .{
+        .cipher_suite = .tls_aes_128_gcm_sha256,
+        .resumption_psk = &psk,
+        .application_protocol = "h3",
+        .auth_binding = tls_core.session.AuthBinding.fromLeafCertificateDer(tls_backend_mod.testdata.certificate_der),
+        .issued_at_unix_ms = 0,
+        .lifetime_seconds = 3600,
+    });
+    var stored_state: tls_core.session.ServerRecoverableState = .{};
+    stored_state.init(&stored_common, 0);
+    defer stored_state.deinit();
+
+    const Resolver = struct {
+        state: *tls_core.session.ServerRecoverableState,
+        calls: usize = 0,
+
+        fn now(_: *anyopaque) i64 {
+            return 0;
+        }
+        fn resolve(ctx: *anyopaque, identity: []const u8) tls_core.pre_shared_key.ResolveError!tls_core.pre_shared_key.ServerPskResolveResult {
+            const self: *@This() = @ptrCast(@alignCast(ctx));
+            self.calls += 1;
+            if (!std.mem.eql(u8, identity, "quic-hrr-resumption-ticket")) return .miss;
+            var out: tls_core.session.ServerRecoverableState = .{};
+            self.state.cloneInto(testing.allocator, &out) catch return error.ResolverFailed;
+            return .{ .hit = .{ .state = out, .lease = tls_core.pre_shared_key.ServerPskLease.initNoop() } };
+        }
+    };
+    var resolver_state = Resolver{ .state = &stored_state };
+    const DecisionProbe = struct {
+        count: usize = 0,
+        last: ?tls_core.tls13_backend.Tls13Backend.ResumptionDecision = null,
+        fn onDecision(ctx: *anyopaque, decision: tls_core.tls13_backend.Tls13Backend.ResumptionDecision) void {
+            const self: *@This() = @ptrCast(@alignCast(ctx));
+            self.count += 1;
+            self.last = decision;
+        }
+    };
+    var decision_probe = DecisionProbe{};
+
+    const pair = try allocator.create(TestPair);
+    errdefer allocator.destroy(pair);
+    pair.* = .{
+        .client_backend = try tls_backend_mod.Tls13Backend.initClientWithAllocatorAndOptions(
+            allocator,
+            .{ .hello_random = [_]u8{0xc1} ** 32, .key_share_seed = [_]u8{0x11} ** 32, .retry_key_share_seed = [_]u8{0x13} ** 32 },
+            .{ .pinned_certificate = tls_backend_mod.testdata.certificate_der },
+            .{ .initial_key_share_mode = .empty },
+        ),
+        .server_backend = tls_backend_mod.Tls13Backend.initServerWithAllocator(
+            allocator,
+            .{ .hello_random = [_]u8{0x51} ** 32, .key_share_seed = [_]u8{0x22} ** 32, .retry_key_share_seed = [_]u8{0x23} ** 32 },
+            try tls_backend_mod.Identity.initPkcs8(
+                tls_backend_mod.testdata.certificate_der,
+                tls_backend_mod.testdata.private_key_pkcs8_der,
+            ),
+        ),
+    };
+    // Native QUIC 1-RTT resumption deliberately ignores connection-specific
+    // transport/application snapshots (matching the runtime-composition PSK
+    // tests elsewhere in this file) — a hand-built stored ticket has no
+    // transport_compat of its own to match against this fresh connection's.
+    const resume_policy: tls_core.tls13_backend.Tls13Backend.ResumeCompatibilityPolicy = .{
+        .transport = .ignore,
+        .application = .ignore,
+    };
+    try pair.client_backend.engine.setClientPskOffers(&offers, &clock_dummy, Clock.now);
+    try pair.client_backend.setResumeCompatibilityPolicy(resume_policy);
+    try pair.server_backend.setResumeCompatibilityPolicy(resume_policy);
+    try pair.server_backend.setServerPskResolver(.{
+        .ctx = &resolver_state,
+        .nowUnixMsFn = Resolver.now,
+        .resolveFn = Resolver.resolve,
+    });
+    try pair.server_backend.setResumptionDecisionObserver(.{ .ctx = &decision_probe, .onDecisionFn = DecisionProbe.onDecision });
+
+    pair.client = try Connection.init(allocator, .{
+        .role = .client,
+        .local_cid = &TestPair.client_cid,
+        .original_dcid = &TestPair.odcid,
+        .peer_cid = &TestPair.odcid,
+        .tls = pair.client_backend.backend(),
+        .now_us = pair.now_us,
+        .initial_path = TestPair.client_path,
+        .events = .{ .context = &client_capture, .emitFn = EventCapture.onEvent },
+    });
+    errdefer pair.client.deinit();
+    pair.server = try Connection.init(allocator, .{
+        .role = .server,
+        .local_cid = &TestPair.odcid,
+        .original_dcid = &TestPair.odcid,
+        .peer_cid = &TestPair.client_cid,
+        .tls = pair.server_backend.backend(),
+        .now_us = pair.now_us,
+        .initial_path = TestPair.server_path,
+        .events = .{ .context = &server_capture, .emitFn = EventCapture.onEvent },
+    });
+    defer pair.deinit(allocator);
+
+    try pair.pump();
+
+    try testing.expectEqual(State.established, pair.client.state());
+    try testing.expectEqual(State.established, pair.server.state());
+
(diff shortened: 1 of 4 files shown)


src/tls/pre_shared_key.zig
```diff
@@ -325,10 +335,8 @@ pub fn deriveBinder(
             var finished_key = tls.hkdfExpandLabel(HkdfSha256, binder_key, "finished", "", Sha256.digest_length);
             defer crypto.secureZero(u8, &finished_key);
 
-            var transcript_hash: [Sha256.digest_length]u8 = undefined;
-            Sha256.hash(truncated_client_hello, &transcript_hash, .{});
             var mac: [HmacSha256.mac_length]u8 = undefined;
-            HmacSha256.create(&mac, &transcript_hash, &finished_key);
+            HmacSha256.create(&mac, transcript_hash[0..Sha256.digest_length], &finished_key);
```
Jaruso, line 340: **Blocking secret-lifetime issue:** the new primitive correctly wipes `psk_fixed`, `early_secret`, `binder_key`, and `finished_key`, but the HMAC output itself remains as a second secret-derived copy on the stack after `@memcpy(out, &mac)`. #485 explicitly requires intermediate binder output to be wiped. Please add `defer crypto.secureZero(u8, &mac);` here and in the SHA-384 branch before creating/copying the MAC.

src/tls/tls13_backend_tests.zig
```diff
@@ -2621,6 +2621,212 @@ test "#484 HRR round trip completes over the extension (QUIC-style) profile too"
...
+    var ticket_b = try makeCacheTicket(&psk_b, "ticket-b");
+    defer ticket_b.deinit();
+    var offers: pre_shared_key.ClientPskOfferSet = .{};
+    try offers.push(&ticket_a);
+    try offers.push(&ticket_b);
+
+    var clock_dummy: u8 = 0;
+    const Clock = struct {
+        fn now(_: *anyopaque) i64 {
+            return 5_000;
```
Jaruso: **This does not prove the required age recomputation.** The same clock function is called for ClientHello1 and ClientHello2, but it always returns `5000`, so both flights receive the same `obfuscated_ticket_age`. The test name/comment says the age is updated across HRR, yet it never observes a changed clock. Make the clock state mutable, capture CH1's age, advance it before `receive(hrr)`, and assert CH2's age changes by the expected delta while identity order remains unchanged.

src/tls/tls13_backend_tests.zig
```diff
@@ -2635,6 +2841,194 @@ fn nthInitialCryptoBytes(sink: *const DirectSink, index: usize) []const u8 {
     unreachable;
 }
 
+test "#485 PSK resumption completes through one HelloRetryRequest, with matching transcripts and traffic secrets on both sides" {
+    var harness = directHarnessWithClientKeyShareMode(.record, .record, .empty);
```
Jaruso, line 2874: **Blocking #485 AC gap:** this proves the PSK+HRR path only through the direct **record** harness. The July 27 #485 reconciliation explicitly kept “resumed HRR loopbacks for record + QUIC” in scope, and none of the three files changed by this PR exercises `quic.Connection`/`QuicTlsAdapter`. Please add a genuine two-Connection QUIC loopback by extending the existing #484 HRR test in `src/quic/connection.zig` with a client PSK offer + server resolver. Assert both sides become `psk_authenticated`, HRR/CH2 remain in Initial packets, Initial keys survive until final SH, and resumed application/stream traffic succeeds.

src/tls/tls13_backend_tests.zig
```diff
@@ -2635,6 +2841,194 @@ fn nthInitialCryptoBytes(sink: *const DirectSink, index: usize) []const u8 {
...
+    // internally without it showing up here.
+    const client_hash = harness.client_backend.core.transcriptHash();
+    const server_hash = harness.server_backend.core.transcriptHash();
+    try std.testing.expectEqualSlices(u8, &client_hash, &server_hash);
+
+    // No traffic secret was derived at the HRR itself, only after the final
+    // ServerHello — same assertions as the non-PSK #484 HRR round trip.
+    try std.testing.expect(harness.observed.handshake_write_secret[0] != null);
+    try std.testing.expect(harness.observed.handshake_write_secret[1] != null);
+    try std.testing.expect(harness.observed.application_write_secret[0] != null);
```
Jaruso, line 2920: **Blocking fixture gap:** non-null secrets and client/server agreement are useful integration assertions, but they do not satisfy #485's deterministic traffic-secret fixture requirement; both sides can share the same transcript/key-schedule bug and still agree. With this test's fixed entropy and PSK, check in independently generated expected handshake (and preferably application) traffic-secret literals and compare the captured bytes exactly. The new primitive fixture already does this well for the binder; the end-to-end schedule needs the same independent known-answer check.

src/tls/tls13_backend_tests.zig
```diff
@@ -2635,6 +2841,194 @@ fn nthInitialCryptoBytes(sink: *const DirectSink, index: usize) []const u8 {
...
+    // Genuinely usable: application data flows both ways under the
+    // PSK-resumed, post-HRR keys.
+    var protected: [record_codec.max_ciphertext_record_len]u8 = undefined;
+    var plaintext: [record_codec.max_ciphertext_fragment_len]u8 = undefined;
+    const request = try harness.client_bridge.sealApplicationData("resumed after hrr", &protected);
+    const opened_request = try harness.server_bridge.openApplicationData(try parseSingleRecord(.ciphertext, request), &plaintext);
+    try std.testing.expectEqualStrings("resumed after hrr", opened_request.inner.content);
+}
+
+test "#485 server rejects a ClientHello2 PSK binder computed only over Hash(Truncate(CH2)), requiring the rebound transcript instead" {
```
Jaruso, line 3066: This covers the **fatal authenticated-binder** branch, but #485 also requires the existing full-handshake fallback behavior to survive HRR. Please add post-HRR cases for an unknown/miss identity and an expired/incompatible identity (plus client-auth forcing full handshake), asserting the handshake completes via certificate auth, `psk_authenticated` stays false, the resumption-decision classification is unchanged, and the new capture/digest is wiped. Those paths are important here because #536 changed the retained CH2 capture used before `selectPsk()` decides to fall back.

Jaruso (commented):
Disposition: **changes requested before merge**. I reviewed PR #536 against the current #485 issue plus the dated remaining-work reconciliation on #485. The core implementation is strong: the transcript-digest binder API is the right abstraction, CH2 generation uses `Transcript.peekWith()` over the rebound transcript without mutating the live transcript, the server snapshots the pre-CH2 transcript before `Core.acceptSecondClientHello()`, the isolated-CH2 binder negative is correct, and the async *credential-selection* suspension proves the owned capture survives a real backend suspension. CI is green, and the PR is based on current `main`.

I am **not** treating the stale “async PSK resolver” wording in the issue body as required; the July 27 reconciliation explicitly removed that as non-production scope. The remaining blockers are:

1. **Wipe the intermediate HMAC result in `deriveBinderFromTranscriptHash`.** Both SHA-256 and SHA-384 branches leave the secret-derived `mac` stack copy resident after copying it to `out`. #485 explicitly calls for binder/intermediate material to be wiped. Add `defer crypto.secureZero(u8, &mac);` in both branches.

2. **Add the required real QUIC PSK+HRR loopback.** The new end-to-end test is record-only (`directHarnessWithClientKeyShareMode(.record, .record, .empty)`). #485’s reconciled scope still explicitly requires resumed HRR loopbacks for **record + QUIC**. Reuse the existing #484 `src/quic/connection.zig` real-HRR test and combine it with the existing resumed-QUIC PSK setup: configure the client with `.initial_key_share_mode = .empty`, install a real client ticket and server resolver, pump two real `Connection`s, then assert both backends are `psk_authenticated`, exactly one HRR occurred, HRR/CH2 stayed in Initial packets, Initial keys were not discarded before the final ServerHello, and application/stream data works under the resumed keys.

   Expected shape:

```zig
pair.client_backend = try Tls13Backend.initClientWithAllocatorAndOptions(
    allocator,
    client_entropy,
    trust,
    .{ .initial_key_share_mode = .empty },
);
try pair.client_backend.setClientPskOffers(&offers, &clock, Clock.now);
try pair.server_backend.setServerPskResolver(resolver.provider());

try pair.pump();
try testing.expectEqual(State.established, pair.client.state());
try testing.expectEqual(State.established, pair.server.state());
try testing.expect(pair.client_backend.engine.core.psk_authenticated);
try testing.expect(pair.server_backend.engine.core.psk_authenticated);
try testing.expectEqual(RetryState.hrr_received, pair.client_backend.engine.core.retry_state);
try testing.expectEqual(RetryState.hrr_sent, pair.server_backend.engine.core.retry_state);
try testing.expect(client_capture.initialPacketsSentBeforeInitialDiscard() >= 2);
try testing.expect(server_capture.initialPacketsSentBeforeInitialDiscard() >= 2);
```

3. **Add deterministic final-handshake/traffic-secret known-answer fixtures.** The current record test proves only that both sides agree and that secret slots are non-null. Two implementations can agree on the same wrong transcript/key schedule. The reconciled #485 scope explicitly asks for deterministic HRR binder/transcript/**traffic-secret** fixtures. With the fixed entropy/PSK already used here, check in independently generated literal values for at least the rebound transcript hash / final CH2 binder and the post-final-ServerHello handshake traffic secrets (ideally application secrets too), then compare the observed bytes to those literals rather than only `!= null` / client==server.

```zig
const expected_client_hs = hexBytes("<independently generated fixture>");
const expected_server_hs = hexBytes("<independently generated fixture>");
try testing.expectEqualSlices(u8, &expected_client_hs, &harness.observed.handshake_write_secret[0].?);
try testing.expectEqualSlices(u8, &expected_server_hs, &harness.observed.handshake_write_secret[1].?);
```

4. **Make the age-recomputation test actually advance time.** Its clock always returns `5000`, so CH1 and CH2 necessarily carry the same age; the test currently proves “age equals 5000,” not “CH2 recomputed age after HRR.” Use mutable clock state, parse CH1’s PSK age, advance the clock before feeding HRR, then assert CH2’s age changed by the expected amount while identity order remains identical.

```zig
const Clock = struct {
    now_ms: i64,
    fn now(ctx: *anyopaque) i64 {
        return (@as(*@This(), @ptrCast(@alignCast(ctx)))).now_ms;
    }
};
var clock = Clock{ .now_ms = 5_000 };
try client.setClientPskOffers(&offers, &clock, Clock.now);
try client.backend().start(.client, {}, &sink);
const ch1_age = ...;
clock.now_ms = 6_250;
try client.backend().receive(.initial, hrr_raw, &sink);
const ch2_age = ...;
try testing.expectEqual(@as(u32, ch1_age + 1_250), ch2_age);
```

5. **Finish the HRR fallback/cleanup matrix.** The new tests cover successful resumption and fatal bad-binder behavior, but not the reconciled requirement that ordinary fallback semantics remain intact after HRR. Add at least an unknown/miss and expired/incompatible identity through HRR that completes as a full certificate handshake, plus a client-auth-enabled case that forces full-handshake behavior. For each, assert `!psk_authenticated`, expected resumption decision/error classification, and that `client_hello_psk`/its binder digest are cleared. These are the paths most likely to regress specifically because #536 changed what is retained in the CH2 capture.

Once those are addressed, I would expect #485 to be complete; I did not find another protocol-state or binder-transcript bug in the changed production path.

src/quic/connection.zig
```diff
@@ -3557,6 +3557,212 @@ test "driver: HelloRetryRequest completes through real QUIC Initial CRYPTO data"
...
+        .resolveFn = Resolver.resolve,
+    });
+    try pair.server_backend.setResumptionDecisionObserver(.{ .ctx = &decision_probe, .onDecisionFn = DecisionProbe.onDecision });
+
+    pair.client = try Connection.init(allocator, .{
+        .role = .client,
+        .local_cid = &TestPair.client_cid,
+        .original_dcid = &TestPair.odcid,
+        .peer_cid = &TestPair.odcid,
+        .tls = pair.client_backend.backend(),
```
Jaruso, line 3764: **Blocking — this no longer compiles against current `main`.** PR #535 merged during this review and replaced `Options.original_dcid` with `original_destination_cid` plus `initial_secret_dcid`. The current merge-ref CI fails here on Linux/ARM/macOS/appliance builds before tests run. Rebase and mirror the adjacent #484 HRR test's current initialization on both client and server:

```zig
.original_destination_cid = &TestPair.odcid,
.initial_secret_dcid = &TestPair.odcid,
```

There is no QUIC Retry packet in this #485 loopback, so both are the original DCID. After the rebase, please run the actual merge-ref test suite again rather than relying on the branch's pre-#535 local result.

Jaruso (commented):
Re-review disposition: **changes requested before merge**.

The five findings from my previous pass are mostly addressed correctly in the branch code:

- intermediate binder HMAC output is now zeroized in both SHA-256 and SHA-384 paths;
- the age-recomputation test now advances a mutable clock and proves CH2's age actually changed;
- the record-mode KAT now checks the rebound digest, CH2 binder, handshake traffic secrets, and application traffic secrets against independent literals;
- fallback coverage now includes unknown identity, incompatible identity, and client-auth-forced full handshake;
- a genuine two-`Connection` QUIC PSK+HRR loopback was added.

Two blockers remain:

1. **Current merge-ref CI does not compile.** PR #535 merged while this branch was in review and replaced QUIC `Options.original_dcid` with the split `original_destination_cid` / `initial_secret_dcid` model. The new #485 QUIC test still initializes the removed field, so Linux, ARM, macOS, and appliance test jobs all fail before running the suite. Rebase onto current `main` and update both sides of this test using the same no-Retry initialization now used by the adjacent #484 test:

```zig
.original_destination_cid = &TestPair.odcid,
.initial_secret_dcid = &TestPair.odcid,
```

For this particular HRR test there is no QUIC Retry packet, so both values are the ODCID.

2. **The required expired-ticket-through-HRR fallback case is still absent.** The issue's end-to-end matrix explicitly calls out unknown/**expired** identity through HRR, and my previous review requested unknown/expired/incompatible coverage. The new tests cover unknown and an auth-binding incompatibility, but not expiry. Please add an expired server-side recoverable ticket that is still offered by the client, drive it through HRR, and prove it falls back to the certificate handshake with `psk_authenticated == false`, an `.incompatible` resumption decision, completed certificate authentication, and cleared CH2 capture/retry state. For example, give the server-side recovered session `issued_at_unix_ms = 0`, `lifetime_seconds = 1`, and make the resolver clock return `5_000`; keep the client-side offered ticket itself unexpired so the offer reaches the server and exercises server compatibility/fallback rather than being filtered locally.

Everything else I rechecked against the reconciled #485 scope now looks aligned: precomputed-digest binder API, rebound CH2 binder generation/verification, PSK-last/identity ordering, record-mode resumed HRR, deterministic KATs, cleanup on fatal binder failure/fallback, and the actual QUIC carrier test design. Once the branch is rebased/fixed and the expired fallback scenario is added with green merge-ref CI, I expect #485 to be closable.

src/tls/tls13_backend_tests.zig
```diff
@@ -2635,6 +2870,484 @@ fn nthInitialCryptoBytes(sink: *const DirectSink, index: usize) []const u8 {
...
+// #485: ordinary fallback/cleanup semantics must survive a HelloRetryRequest
+// unchanged. #536 changed what `onClientHello` retains for ClientHello2's
+// binder capture — these prove that change never leaks into the paths where
+// `selectPsk` declines to resume and the connection falls back to full
+// certificate authentication.
+// --------------------------------------------------------------------------
+
+test "#485 an unknown PSK identity through HelloRetryRequest falls back to a full certificate handshake" {
+    var harness = directHarnessWithClientKeyShareMode(.record, .record, .empty);
+    defer harness.deinit();
```
Jaruso, line 3200: **Blocking acceptance-evidence gap:** the fallback matrix now covers unknown, auth-binding-incompatible, and client-auth-forced full handshake, but #485 explicitly requires an **expired** identity through HRR to preserve ordinary fallback behavior too (and my previous review called this out as unknown/expired/incompatible). Please add a server-side recovered session that expires before `resolver.nowUnixMs()` while the client-side ticket remains eligible to offer, then drive the normal CH1 → HRR → CH2 path and assert: resolver consulted once, no PSK authentication, full certificate handshake completes, resumption decision is `.incompatible`, and `client_hello_psk`/retry state are cleared. That distinguishes expiry handling from the generic auth-binding mismatch already covered below.

Jaruso (commented):
Re-review at `f1469860`: the two prior blockers are addressed, and I do not have any remaining code changes to request.

- The branch is rebased onto current `main` after #535, and the real QUIC PSK+HRR loopback now uses the current `original_destination_cid` / `initial_secret_dcid` initialization shape.
- The missing expired-ticket path is now covered distinctly from generic incompatibility: the client still emits the live offer, the server resolves an authoritatively expired session, resumption is classified as incompatible, both sides remain non-PSK-authenticated, and the certificate handshake completes.
- Prior blockers remain fixed: binder MAC stack copies are zeroized; CH2 ticket ages are proven to change under an advancing clock; the independent KAT checks rebound binder/transcript material plus handshake/application traffic secrets; unknown/incompatible/client-auth fallback behavior is covered; and the real two-Connection QUIC path establishes a usable resumed connection through exactly one HRR.
- The production flow still preserves the CH1-selected offer order through CH2, computes the binder from the pre-CH2 rebound transcript snapshot, verifies server-side against the owned captured digest, and wipes the capture/digest on terminal cleanup.
- I am continuing to apply the July 27 scope reconciliation: the issue body's async-PSK-resolver wording is stale because the production resolver is synchronous; the existing async credential-selection suspension is sufficient evidence that the owned capture survives a real backend suspension.

CI run #1051 is still globally `in_progress` at review time, but the code-relevant matrix already passing includes Linux/ARM/macOS unit tests, H3 resumption/0-RTT interop, lint, TLS dependency audit, Debug/ReleaseFast builds, and perf smoke. No failures are currently present on this head.

**Disposition: code/AC ready to merge once the currently-running CI workflow finishes green. #485 can close with this PR.**

PR #538, opened 2026-07-29
Implement HTTP/3 RFC 9218 priority hints

Jaruso:
## Summary
- parse HTTP/3 `priority` request headers into RFC 9218 urgency/incremental hints while preserving the header
- process request `PRIORITY_UPDATE` frames on the peer control stream and count parse/update metrics
- poll completed buffered HTTP/3 requests by urgency with deterministic equal-priority ordering
- document the experimental scheduling behavior and update the changelog

## Validation

Closes #254


src/http3/session.zig
```diff
@@ -133,6 +138,15 @@ pub const RequestStream = struct {
     }
 
     fn appendHeader(self: *RequestStream, name: []const u8, value: []const u8) !void {
+        if (std.ascii.eqlIgnoreCase(name, "priority")) {
+            if (priority.parse(value)) |parsed| {
+                self.priority = parsed.effective();
```
Jaruso: **Blocking — combine all `Priority` field lines before parsing.** RFC 8941 Structured Fields requires matching field lines in one field section to be combined with `, ` and parsed as a single Dictionary. Parsing each occurrence here independently makes omitted members reset values from an earlier line. For example:

```http
priority: u=1
priority: i
```

is semantically `u=1, i`, but the current code first sets `{u=1,i=false}` and then overwrites it with `{u=3,i=true}`. A malformed second line similarly leaves the first line’s priority active instead of making the combined field invalid/defaulted.

Please preserve each raw header as you do now, but collect the `priority` values while decoding the HEADERS field section and call `priority.parse` once after the loop. Conceptually:

```zig
var priority_value: std.ArrayList(u8) = .empty;
defer priority_value.deinit(self.allocator);

// While walking decoded fields:
if (std.ascii.eqlIgnoreCase(field.name, "priority")) {
    if (priority_value.items.len != 0)
        try priority_value.appendSlice(self.allocator, ", ");
    try priority_value.appendSlice(self.allocator, field.value);
}
try self.appendHeader(field.name, field.value); // still preserve wire fields

// After the complete field section:
if (priority_value.items.len != 0) {
    const parsed = priority.parse(priority_value.items) catch {
        self.priority = priority.Priority.default;
        self.priority_parse_error = true;
        return;
    };
    self.priority = parsed.effective();
    self.priority_invalid_parameter = parsed.invalid_parameter;
    self.priority_duplicate_parameter = parsed.duplicate_parameter;
}
```

Please add a regression test for the two-line example above and for `priority: u=1` followed by a malformed second value.

src/http3/conn.zig
```diff
@@ -372,6 +379,69 @@ pub fn Conn(comptime Transport: type) type {
...
+                else => try self.peer_control_view.ingestFrame(raw),
+            }
+        }
+
+        fn applyPriorityUpdate(self: *Self, raw: frame.RawFrame) frame.DecodeError!void {
+            const kind = priority.kindFromFrameType(raw.typ) orelse return;
+            const update = priority.decodePayload(raw.payload) catch return error.FrameTooLarge;
+            switch (kind) {
+                .push => {
+                    self.metrics.priority_updates_ignored_push += 1;
```
Jaruso: **Blocking — ignoring push updates (and accepting updates on a client) is not spec-compatible.** RFC 9218’s HTTP/3 rules make `PRIORITY_UPDATE` client-to-server only: a client that receives either PRIORITY_UPDATE variant from a server must close with `H3_FRAME_UNEXPECTED`. The request variant must identify a client-initiated bidirectional request stream; an invalid target is `H3_ID_ERROR`. The push variant must identify a promised Push ID; this implementation never promises server push, so every push update is invalid and should produce `H3_ID_ERROR`, not an ignored-update metric.

I’d validate direction and target before parsing the field value, roughly:

```zig
if (self.role == .client)
    return self.fail(.frame_unexpected);

switch (kind) {
    .request => {
        if ((update.element_id & 0x3) != 0)
            return self.fail(.id_error);
        // Also reject IDs outside the peer's permitted request-stream space.
    },
    .push => return self.fail(.id_error), // no Push Promise / MAX_PUSH_ID support
}
```

Because `applyPriorityUpdate` currently returns `frame.DecodeError`, this probably needs the control-dispatch path to return `H3Error` (or a typed error that maps explicitly to `.frame_unexpected` / `.id_error`) rather than squeezing these protocol errors through `failControl`. Please cover server->client updates plus request/push invalid-ID cases.

src/http3/conn.zig
```diff
@@ -372,6 +379,69 @@ pub fn Conn(comptime Transport: type) type {
...
+            }
+
+            self.metrics.priority_updates_received += 1;
+            const parsed = priority.parse(update.field_value) catch {
+                self.metrics.priority_parse_errors += 1;
+                return;
+            };
+            if (parsed.invalid_parameter) self.metrics.priority_invalid_parameters += 1;
+            if (parsed.duplicate_parameter) self.metrics.priority_duplicate_parameters += 1;
+            if (self.requests.get(update.element_id)) |request| {
```
Jaruso, line 449: **Blocking — this drops a valid cross-stream `PRIORITY_UPDATE` race.** A client can send the update on the control stream before the referenced request stream is observed; HTTP/3 gives no ordering guarantee across those streams. RFC 9218 explicitly recommends retaining the most recent update and applying it when the request stream opens (subject to a bounded resource policy). Here the update is silently lost whenever `self.requests` does not contain the ID yet.

Please keep a bounded pending map and use last-update-wins semantics, then consume it when the request stream is registered. For example:

```zig
pending_priority_updates: std.AutoHashMap(u64, priority.Priority),

// after validating/parsing the update
if (self.requests.get(update.element_id)) |request| {
    request.stream.priority = parsed.effective();
} else {
    // enforce a documented cap before inserting
    try self.pending_priority_updates.put(update.element_id, parsed.effective());
}

// when accepting/registering a request stream
if (self.pending_priority_updates.fetchRemove(id)) |pending| {
    request.stream.priority = pending.value;
}
```

The exact storage can differ, but it should be bounded and a later update for the same ID must replace the earlier one. Add a test that pumps the control-stream update first, then makes the request stream visible, and verifies the retained priority is applied.

src/http3/conn.zig
```diff
@@ -531,21 +625,35 @@ pub fn Conn(comptime Transport: type) type {
...
             exchange: stream_transport.Exchange,
             transport_early: bool,
+            priority: priority.Priority,
         };
 
         /// Pop the next fully received request. The exchange borrows the
         /// request session's memory; call `finishRequest` after responding.
         pub fn pollRequest(self: *Self) H3Error!?IncomingRequest {
+            var selected_id: ?u64 = null;
+            var selected_request: ?*ServerRequest = null;
```
Jaruso, line 678: **Blocking — this is request dispatch ordering, not the stream scheduling required by #254.** The priority selected here is returned in `IncomingRequest`, but `src/http/http3_runtime.zig` does not consume `incoming.priority`: it immediately runs `serveRequest`, queues the response, and moves to the next request. QUIC egress in `Connection.buildAppFrames` still iterates `send_queues` without any priority metadata. So once multiple response streams have bytes queued, RFC 9218 hints no longer affect which stream gets wire bandwidth, and the acceptance criterion that equal-urgency stream scheduling be fair enough to avoid starvation is not exercised.

Please carry a transport-neutral scheduling hint down to the active QUIC stream scheduler (or implement an equivalent bounded H3 scheduling layer). One reasonable shape would be:

```zig
pub const StreamSchedulingHint = struct {
    urgency: u3 = 3,
    incremental: bool = false,
};

pub fn setStreamSchedulingHint(
    self: *Connection,
    id: StreamId,
    hint: StreamSchedulingHint,
) !void { ... }
```

Then have `buildAppFrames` select ready streams by urgency and use a bounded quantum / round-robin cursor within equal urgency, following the policy already documented in `priority.zig`, instead of raw `AutoHashMap` iteration. `http3_runtime` can set the hint from `incoming.priority` before response bytes are queued.

Please add a scheduler test with multiple simultaneously active/backpressured streams proving (a) lower urgency wins while both are ready and (b) equal-urgency streams both make forward progress. The current two-request `pollRequest` test does not establish that.

Jaruso (commented):
**Blocking review findings for #254.** GitHub does not permit the PR author to submit a formal `REQUEST_CHANGES` review on their own PR, so this is posted as a comment review; I do not consider the current head ready to merge.

The existing priority model/parser/serializer is strong, and this PR wires useful header/update metrics plus request ordering, but four integration gaps leave the issue’s HTTP/3 correctness and scheduling acceptance criteria incomplete.

1. Multiple `Priority` field lines are parsed independently instead of as one Structured Fields Dictionary. That can change the effective priority and mishandle a malformed continuation.
2. `PRIORITY_UPDATE` is accepted in directions/for targets that RFC 9218 requires to be connection errors. In particular, servers MUST NOT send these frames; request updates must target request streams; and because this stack never promises push, a push update cannot be silently ignored.
3. A legal `PRIORITY_UPDATE` that wins the cross-stream race and arrives before its request stream is registered is dropped. RFC 9218 explicitly accounts for that race; retain the most recent bounded update and apply it when the stream appears.
4. The scheduling AC is not actually wired to active response streams. `pollRequest` only chooses which fully buffered request is handed to the synchronous handler first. `http3_runtime` then queues the response without consuming `IncomingRequest.priority`, while QUIC egress still iterates `send_queues` with no priority metadata. Please wire a small per-stream scheduling hint into the active-stream egress path (or an equivalent bounded H3 scheduler) and test multiple simultaneously backpressured streams, including equal-urgency fairness.

CI is green on the current head, so these require targeted behavioral tests rather than build fixes. I’d expect regression coverage for: split Priority field lines, server->client PRIORITY_UPDATE, invalid request/push IDs, PRIORITY_UPDATE-before-request-stream ordering, and multiple active equal-urgency response streams.

Jaruso (commented):
Re-review of the amended head (`88a09a7`) against #254 and the previous four blockers.

The previous fixes are substantial: combined `Priority` field parsing is now correct, control-stream direction/target validation is improved, pre-request updates are bounded and retained, and the H3 hint is now wired into QUIC egress. CI is green.

I still don't think #254 is ready to close. There are four remaining behavioral gaps:

1. A PRIORITY_UPDATE received after the request stream is registered but before HEADERS are decoded is overwritten by the later `Priority` header. RFC 9218 treats the update as the newer signal and says it overrides other priority signals.
2. Once `sendResponse()` removes the H3 request, later PRIORITY_UPDATE frames for a response that still has QUIC bytes queued are buffered as if the request had not opened yet and never reach the active QUIC send queue. That defeats the core reprioritization use case for an in-flight response.
3. PRIORITY_UPDATE placement is only validated on the control-stream path. The server request-stream path maps it to `H3_MESSAGE_ERROR`, and the client response-stream decoder ignores it, but RFC 9218 requires `H3_FRAME_UNEXPECTED` whenever PRIORITY_UPDATE is received anywhere other than the client control stream.
4. The QUIC comparator always ranks `incremental=true` ahead of `incremental=false` at equal urgency. A continuously-ready incremental stream can therefore starve a non-incremental peer forever, violating #254's explicit equal-urgency starvation AC and the round-robin policy documented in `priority.zig`/SUPPORT_MATRIX.

I've left concrete fix/test sketches inline. After these are addressed, I would expect tests covering fragmented/late updates, wrong-stream placement, and persistent mixed-incremental equal-urgency traffic rather than only short responses that fit in one packet.

src/http3/conn.zig
```diff
@@ -372,6 +390,76 @@ pub fn Conn(comptime Transport: type) type {
...
+            }
+
+            self.metrics.priority_updates_received += 1;
+            const parsed = priority.parse(update.field_value) catch {
+                self.metrics.priority_parse_errors += 1;
+                return;
+            };
+            if (parsed.invalid_parameter) self.metrics.priority_invalid_parameters += 1;
+            if (parsed.duplicate_parameter) self.metrics.priority_duplicate_parameters += 1;
+            if (self.requests.get(update.element_id)) |request| {
```
Jaruso, line 449: **Blocking — an update can still be overwritten by the request header.** This branch applies the PRIORITY_UPDATE directly to `request.stream.priority` and clears `pending_priority_update`. If the request stream has been registered but its HEADERS frame has not been fully decoded yet (including the ordinary fragmented-HEADERS case), `RequestStream.ingestHeaders()` later assigns the parsed `Priority` header and silently replaces the newer update.

RFC 9218 §7 explicitly says the most recently received PRIORITY_UPDATE can be treated as the most up-to-date information that **overrides any other signal**, specifically because control/request stream ordering can race.

Please keep the latest update authoritative until the request header has been processed. For example:

```zig
if (self.requests.get(update.element_id)) |request| {
    request.pending_priority_update = parsed.effective();
    if (request.stream.saw_headers) {
        request.stream.priority = request.pending_priority_update.?;
        request.pending_priority_update = null;
    }
}
```

Or keep a separate `latest_priority_update` for the lifetime of the response and derive the effective priority from that. The important invariant is that decoding HEADERS can never roll priority back after a PRIORITY_UPDATE has been received.

Please add a regression test where the request stream is accepted, only part of its HEADERS bytes are made readable, then a control-stream update arrives, then the rest of HEADERS arrives with `priority: u=7`; the effective result must remain the update value.

src/http3/conn.zig
```diff
@@ -372,6 +390,76 @@ pub fn Conn(comptime Transport: type) type {
...
+                request.stream.priority = parsed.effective();
+                request.pending_priority_update = null;
+            } else {
+                if (!self.pending_priority_updates.contains(update.element_id) and
+                    self.pending_priority_updates.count() >= max_pending_priority_updates)
+                {
+                    return self.fail(.excessive_load);
+                }
+                self.pending_priority_updates.put(update.element_id, parsed.effective()) catch return error.OutOfMemory;
```
Jaruso, line 460: **Blocking — late reprioritization of an active response is lost.** `sendResponse()` queues response bytes into QUIC and immediately calls `finishRequest()`, which removes this stream from `self.requests`. If the client then sends PRIORITY_UPDATE while those response bytes are still queued/in flight, this branch treats the ID as a pre-request race and stores it in `pending_priority_updates`. That stream will never be opened again, so the update never reaches `Connection.send_queues[id].scheduling_hint`.

That misses the primary RFC 9218 §6 use case: changing the priority of a response *while it is in progress*.

I would make the transport scheduling hook reachable from the update path, not only once in `http3_runtime.serveRequest`. One possible shape:

```zig
fn applyTransportPriority(
    transport: *Transport,
    id: u64,
    p: priority.Priority,
) bool {
    if (comptime @hasDecl(Transport, "setStreamSchedulingHint")) {
        transport.setStreamSchedulingHint(id, .{
            .urgency = p.urgency,
            .incremental = p.incremental,
        }) catch return false;
        return true;
    }
    return false;
}
```

Pass `transport` through control-frame ingestion and apply the update to an already-known/active QUIC stream immediately; only use the bounded pending map when the referenced request stream genuinely has not appeared yet. `setStreamSchedulingHint` may also need to allow updating an existing send queue after local FIN has been queued, because retransmission/queued data can still need scheduling.

Please test: queue a response larger than one packet, transmit only part of it, send PRIORITY_UPDATE, then verify subsequent STREAM selection uses the new urgency.

src/http3/conn.zig
```diff
@@ -372,6 +390,76 @@ pub fn Conn(comptime Transport: type) type {
...
+                    else => return self.failControl(err),
+                };
+                try self.ingestControlFrame(raw);
+                discardPrefix(&self.peer_control_view.pending, raw.len);
+            }
+        }
+
+        fn ingestControlFrame(self: *Self, raw: frame.RawFrame) H3Error!void {
+            switch (raw.typ) {
+                .priority_update_request, .priority_update_push => {
```
Jaruso, line 425: **Blocking — the `H3_FRAME_UNEXPECTED` rule still only works on the control-stream path.** RFC 9218 §7.2 says PRIORITY_UPDATE **MUST** be on the client control stream and receiving it on any other stream is `H3_FRAME_UNEXPECTED`.

Current non-control paths still violate that:

- server `RequestStream.ingestFrame()` returns `InvalidRequestFrame` for PRIORITY_UPDATE, but `pumpRequests()` maps every session error to `H3_MESSAGE_ERROR`;
- client `pollResponse()` falls through `else => {}` for PRIORITY_UPDATE on the request/response stream, so a server-sent update there is silently ignored.

Please make placement validation explicit on both directions. For example, map an `UnexpectedFrame`/`InvalidRequestFrame` session error to `.frame_unexpected`, and in `pollResponse()` explicitly reject `.priority_update_request, .priority_update_push` via `return self.fail(.frame_unexpected)` rather than treating them like unknown extension frames.

Add tests for a client-sent PRIORITY_UPDATE on a request stream and a server-sent PRIORITY_UPDATE on a response stream; both should close with `0x0105`.

src/quic/connection.zig
```diff
@@ -3045,10 +3069,60 @@ pub const Connection = struct {
...
+        const frame_room: u64 = @intCast(budget -| plain_len -| frame.max_stream_overhead);
+        const n_bytes = @min(@min(unsent, @min(stream_window, conn_window)), frame_room);
+        return n_bytes != 0 or (want_fin and unsent == 0);
+    }
+
+    fn sendQueueBefore(lhs: SelectedSendQueue, rhs: SelectedSendQueue, cursor: StreamId) bool {
+        const lhs_hint = lhs.queue.scheduling_hint;
+        const rhs_hint = rhs.queue.scheduling_hint;
+        if (lhs_hint.urgency != rhs_hint.urgency) return lhs_hint.urgency < rhs_hint.urgency;
+        if (lhs_hint.incremental != rhs_hint.incremental) return lhs_hint.incremental and !rhs_hint.incremental;
```
Jaruso: **Blocking — mixed incremental/non-incremental streams at the same urgency can still starve.** This comparator makes every `incremental=true` stream rank ahead of every `incremental=false` stream before the round-robin cursor is considered. If an incremental stream remains continuously ready (a long response or CONNECT-style stream), the equal-urgency non-incremental peer is never selected.

That directly violates #254's acceptance criterion that equal-urgency scheduling be fair enough to avoid starvation, and it contradicts the policy already documented in `src/http3/priority.zig`: all ready responses within one urgency band get a bounded turn so every response makes forward progress. RFC 9218 §10 also explicitly recommends avoiding starvation when incremental and non-incremental responses share an urgency.

For the simple policy this issue chose, the safest fix is to make urgency the strict tier and use the cursor across **all** streams in that tier:

```zig
fn sendQueueBefore(lhs: SelectedSendQueue, rhs: SelectedSendQueue, cursor: StreamId) bool {
    const lu = lhs.queue.scheduling_hint.urgency;
    const ru = rhs.queue.scheduling_hint.urgency;
    if (lu != ru) return lu < ru;
    return distanceFromCursor(lhs.id, cursor) < distanceFromCursor(rhs.id, cursor);
}
```

You can still retain `incremental` in the hint for future/adaptive policy; it just shouldn't form an absolute sub-priority that defeats the fairness guarantee.

The current fairness test uses two tiny `incremental=true` responses that both fit in one packet. Please add a persistent mixed case (`i=true` + `i=false`, same urgency, bodies > one datagram / multiple transmit polls) and assert both stream IDs make forward progress.

src/http3/conn.zig
```diff
@@ -397,6 +490,47 @@ pub fn Conn(comptime Transport: type) type {
             }
         }
 
+        fn applyPendingPriorityUpdate(_: *Self, request: *ServerRequest) void {
```
Jaruso, line 493: **Blocking — the PRIORITY_UPDATE override is still cleared too early for multi-chunk HEADERS.**

This helper is called after *every* non-empty `readStream()` result in `pumpRequests()`, even when `RequestStream.ingestBytes()` only buffered another incomplete fragment and `request.stream.saw_headers` is still false. That leaves this sequence broken:

1. request stream is registered and receives the first byte of HEADERS;
2. control stream delivers `PRIORITY_UPDATE u=0` → `pending_priority_update = u=0`;
3. request stream receives a **second partial** HEADERS chunk that still does not complete the frame;
4. `applyPendingPriorityUpdate()` runs here, writes `u=0`, then clears the pending value;
5. final HEADERS bytes arrive with `priority: u=7`; `ingestHeaders()` assigns `u=7`, rolling back the newer update.

The new regression test only has one partial request chunk before the update and then sends the entire remainder, so it does not exercise step 3.

RFC 9218 §7 says the most recently received PRIORITY_UPDATE can be treated as the most up-to-date information that overrides other signals specifically to survive cross-stream ordering races. Keep the pending update until HEADERS has actually been processed:

```zig
fn applyPendingPriorityUpdate(_: *Self, request: *ServerRequest) void {
    if (!request.stream.saw_headers) return;
    if (request.pending_priority_update) |update| {
        request.stream.priority = update;
        request.pending_priority_update = null;
    }
}
```

Please extend the fragmented regression test to send at least one additional incomplete request-stream chunk *after* the update and pump it before sending the final HEADERS bytes.

src/http3/session.zig
```diff
@@ -112,15 +119,34 @@ pub const RequestStream = struct {
...
+        var priority_value: std.ArrayList(u8) = .empty;
+        defer priority_value.deinit(self.allocator);
         for (fields[0..count]) |field| {
             if (field.name.len > 0 and field.name[0] == ':') {
                 if (regular_seen) return error.PseudoHeaderAfterRegularHeader;
                 try self.applyPseudoHeader(field);
             } else {
                 regular_seen = true;
+                if (std.ascii.eqlIgnoreCase(field.name, "priority")) {
+                    if (priority_value.items.len != 0) try priority_value.appendSlice(self.allocator, ", ");
```
Jaruso: **Blocking — use field presence, not accumulated byte length, when combining `Priority` lines.**

RFC 8941 §4.2 requires parsers to combine **all** matching field lines into one comma-separated value before parsing. `priority_value.items.len` is not equivalent to “a Priority field was already seen”: an empty field value leaves the buffer length at zero.

For example:

```http
priority:
priority: u=1
```

must combine to `, u=1`. That is malformed Structured Fields syntax, so the entire Priority field is ignored/defaulted. The current code drops the empty first line and parses only `u=1`, changing the semantics.

Track presence separately:

```zig
var priority_seen = false;
var priority_value: std.ArrayList(u8) = .empty;

...
if (std.ascii.eqlIgnoreCase(field.name, "priority")) {
    if (priority_seen)
        try priority_value.appendSlice(self.allocator, ", ");
    try priority_value.appendSlice(self.allocator, field.value);
    priority_seen = true;
}
...
if (priority_seen) {
    // parse even when the resulting single-line value is empty
}
```

Please add regressions for an empty line followed by `u=1`, and ideally `u=1` followed by an empty line. This is part of #254's requirement that Priority header values be parsed/validated according to the Structured Fields rules.

Jaruso (commented):
Re-review of the latest head against #254 and the prior review threads. The previous four blockers are mostly addressed: late reprioritization now reaches the transport scheduler, PRIORITY_UPDATE placement errors are mapped to H3_FRAME_UNEXPECTED, and equal-urgency QUIC scheduling no longer gives `incremental=true` an absolute tier over non-incremental peers. The new mixed fairness test is also materially better.

I still found two blocking correctness gaps:

1. The latest PRIORITY_UPDATE can still be cleared before a fragmented HEADERS frame has actually completed. `applyPendingPriorityUpdate()` runs after every non-empty request read and does not check `request.stream.saw_headers`. A second partial HEADERS chunk after the update clears the pending override; the final chunk then parses the request `Priority` header and rolls the effective priority back. RFC 9218 §7 requires the most recently received update to remain authoritative over other signals across these cross-stream races.

2. Multiple `Priority` field lines are combined based on `priority_value.items.len` rather than whether a matching field line has been seen. An empty first line is therefore erased from the combined value. RFC 8941 §4.2 requires every matching field line to participate in the comma-separated field value. `priority: ` followed by `priority: u=1` must be parsed as `, u=1` (and therefore fail/ignore the whole structured field), not silently become `u=1`.

Both are small fixes, but they are directly inside #254's parser/update acceptance criteria. CI for this head is still running at the time of this review; completed build/format/unit/security jobs seen so far are green.

src/http3/conn.zig
```diff
@@ -99,8 +104,16 @@ pub fn Conn(comptime Transport: type) type {
             responses_decoded: u64 = 0,
             settings_received: bool = false,
             unknown_uni_streams: u64 = 0,
+            priority_updates_received: u64 = 0,
+            priority_updates_ignored_push: u64 = 0,
```
Jaruso, line 108: **Non-blocking cleanup:** this metric became unreachable once push PRIORITY_UPDATE was changed to fail immediately with `H3_ID_ERROR`; there is no increment path anymore. I’d remove it rather than leave a counter whose name implies push updates can be ignored. This does not affect #254 correctness.

Jaruso (commented):
Re-review of `d7d8647`: **no blocking code findings remain.** I rechecked the full six-file change against #254 and the RFC 9218 behaviors we previously flagged.

The two blockers from the last pass are fixed correctly: `Priority` field combination now tracks field presence (including empty repeated lines), and the pending PRIORITY_UPDATE is retained until HEADERS has actually decoded, with a multi-chunk regression test covering the race.

Acceptance-criteria check:
- `priority` values are parsed/validated as a combined Structured Fields Dictionary, including malformed/invalid/duplicate cases.
- defaults are modeled/documented as `u=3`, `i=false`.
- HTTP/3 PRIORITY_UPDATE direction, placement, request-vs-push target validation, pre-open buffering, last-update-wins behavior, and late active-response reprioritization are implemented.
- priority reaches the native QUIC send scheduler; lower urgency wins and all equal-urgency streams share a bounded round-robin tier, including mixed incremental/non-incremental streams.
- tests cover header parsing, urgency/incremental values, malformed headers, control/request stream races, invalid updates, late updates, and active-stream scheduling fairness.

I also rescanned the scheduler/retransmission path and runtime handoff and did not find another correctness regression. RFC 9218 only says IDs beyond the QUIC client-bidi stream limit SHOULD produce H3_ID_ERROR and explicitly allows implementations not to do so where that limit is not practical to surface at H3, so I do not consider the current type-level request-stream validation a blocker.

CI run 1056 is still in progress, but the unit tests on Linux/macOS/ARM, H3 resumption/0-RTT interop, builds, formatting, lint, security, appliance tests, and perf smoke that have completed are green. Wait for the remaining required checks before merging.

Two cleanup nits are noted, neither merge-blocking: the dead `priority_updates_ignored_push` metric below, and the unchanged comment in `src/http3/priority.zig` that still says scheduler wiring is a future follow-up/out of scope for #254 even though this PR now wires it.

**Code verdict: ready to merge once required CI completes.**

PR #539, opened 2026-07-29
http3: finish rollout advertisement and drain lifecycle

Jaruso:
Claims #257

## Summary
- make HTTP/3 Alt-Svc an explicit effective rollout state with `off|auto`, bounded `ma`, and `clear` support
- advertise only after the runtime is actually bootstrapped; reject listener-owned H3 reload changes as restart-required while allowing advertisement-only reloads
- add runtime H3 drain API with GOAWAY, no-new-work admission, deadline close, CID cleanup, and effective-state metrics
- apply the gateway-owned Alt-Svc policy consistently across buffered, streamed, and error response paths, stripping upstream Alt-Svc
- document the v1 UDP/LB/reload/drain rollout contract

## Validation
- `zig build test`
- `zig build`

### DIFF
--- a/src/gateway_shutdown.zig
+++ b/src/gateway_shutdown.zig
@@ -380,6 +380,27 @@ test "http3ListenerConfigChanged permits advertisement-only reloads" {
     try std.testing.expect(http3ListenerConfigChanged(&base, &proposed));
 }
 
+test "computeReloadedHttp3Advertisement withdraws active auto advertisement when reloaded off" {
+    const allocator = std.testing.allocator;
+    var cfg = try edge_config.loadFromEnv(allocator);
+    defer cfg.deinit(allocator);
+
+    cfg.http3_enabled = true;
+    cfg.http3_alt_svc = .off;
+    cfg.http3_alt_svc_max_age_seconds = 60;
+
+    const withdrawal = computeReloadedHttp3Advertisement(&cfg, true, 8443, .advertising);
+    try std.testing.expectEqual(http.http3_handler.Advertisement.clear, withdrawal);
+    try std.testing.expectEqual(http.http3_runtime.EffectiveAdvertisementState.clearing, stateForHttp3Advertisement(true, true, withdrawal));
+
+    const already_off = computeReloadedHttp3Advertisement(&cfg, true, 8443, .ready_advertisement_disabled);
+    try std.testing.expectEqual(http.http3_handler.Advertisement.disabled, already_off);
+
+    cfg.http3_alt_svc = .auto;
+    const active = computeReloadedHttp3Advertisement(&cfg, true, 8443, .clearing);
+    try std.testing.expectEqualDeep(http.http3_handler.Advertisement{ .active = .{ .port = 8443, .max_age_seconds = 60 } }, active);
+}
+
 extern "c" fn setenv(name: [*:0]const u8, value: [*:0]const u8, overwrite: c_int) c_int;
 extern "c" fn unsetenv(name: [*:0]const u8) c_int;
 
@@ -582,21 +603,17 @@ pub fn applyReloadedRuntimeConfig(cfg: *const edge_config.EdgeConfig, state: *Ga
 
     state.runtime_mutex.lock();
     state.add_headers = cfg.add_headers;
+    const previous_h3_advertisement_state = state.http3_advertisement_state;
     if (state.http3_alt_svc) |value| state.allocator.free(value);
     const runtime_ready = if (state.http3_runtime) |runtime| runtime.snapshot().server_bootstrapped else false;
-    const advertisement: http.http3_handler.Advertisement = if (!cfg.http3_enabled or !runtime_ready or cfg.http3_alt_svc == .off)
-        .disabled
-    else
-        .{ .active = .{
-            .port = if (state.http3_runtime) |runtime| runtime.snapshot().quic_port else cfg.quic_port,
-            .max_age_seconds = cfg.http3_alt_svc_max_age_seconds,
-        } };
+    const advertisement = computeReloadedHttp3Advertisement(
+        cfg,
+        runtime_ready,
+        if (state.http3_runtime) |runtime| runtime.snapshot().quic_port else cfg.quic_port,
+        previous_h3_advertisement_state,
+    );
     state.http3_alt_svc = http.http3_handler.formatAdvertisement(state.allocator, advertisement) catch null;
-    state.http3_advertisement_state = switch (advertisement) {
-        .disabled => if (!cfg.http3_enabled) .disabled else if (!runtime_ready) .configured_unavailable else .ready_advertisement_disabled,
-        .clear => .clearing,
-        .active => .advertising,
-    };
+    state.http3_advertisement_state = stateForHttp3Advertisement(cfg.http3_enabled, runtime_ready, advertisement);
     if (state.hsts_value.len > 0) state.allocator.free(state.hsts_value);
     state.hsts_value = computeHstsValue(state.allocator, cfg) catch &.{};
     state.security_headers = blk: {
@@ -622,6 +639,37 @@ pub fn applyReloadedRuntimeConfig(cfg: *const edge_config.EdgeConfig, state: *Ga
     };
     state.logger.min_level = cfg.log_level;
     state.runtime_mutex.unlock();
+}
+
+fn computeReloadedHttp3Advertisement(
+    cfg: *const edge_config.EdgeConfig,
+    runtime_ready: bool,
+    runtime_port: u16,
+    previous_state: http.http3_runtime.EffectiveAdvertisementState,
+) http.http3_handler.Advertisement {
+    if (!cfg.http3_enabled or !runtime_ready) return .disabled;
+    if (cfg.http3_alt_svc == .off) {
+        return switch (previous_state) {
+            .advertising, .clearing, .draining => .clear,
+            else => .disabled,
+        };
+    }
+    return .{ .active = .{
+        .port = runtime_port,
+        .max_age_seconds = cfg.http3_alt_svc_max_age_seconds,
+    } };
+}
+
+fn stateForHttp3Advertisement(
+    http3_enabled: bool,
+    runtime_ready: bool,
+    advertisement: http.http3_handler.Advertisement,
+) http.http3_runtime.EffectiveAdvertisementState {
+    return switch (advertisement) {
+        .disabled => if (!http3_enabled) .disabled else if (!runtime_ready) .configured_unavailable else .ready_advertisement_disabled,
+        .clear => .clearing,
+        .active => .advertising,
+    };
 }
 
 pub fn reopenErrorLog(cfg: *const edge_config.EdgeConfig) !void {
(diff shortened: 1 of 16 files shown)


src/http/http3_runtime.zig
```diff
@@ -418,6 +460,7 @@ pub const Runtime = struct {
                 if (n == 0) continue;
                 const datagram = buf[0..@intCast(n)];
                 if (from.family != posix.AF.INET) continue;
+                if (draining) continue;
```
Jaruso: **Blocking — this prevents existing H3 streams from draining gracefully.** Once `draining` is true, this drops *all* inbound UDP datagrams, not just packets that would create new connections. Existing QUIC connections still need inbound ACKs, request-body STREAM frames, flow-control updates, path traffic, etc. An admitted request that is still receiving a body (or whose response needs normal QUIC progress) can no longer complete, which directly violates #257's requirement that existing streams may finish until the deadline.

Please keep routing/ingesting datagrams for already-tracked connections while refusing only new connection admission. Conceptually the receive path should distinguish a datagram whose DCID resolves to an existing `ConnEntry` from an unknown/new Initial; existing routes should continue through normal QUIC processing during drain, while unknown/new connection attempts are discarded/rejected. Add a real UDP-runtime test where an established request continues exchanging packets after `beginDrain()` and completes before the deadline.

src/http/http3_runtime.zig
```diff
@@ -837,10 +880,24 @@ pub const Runtime = struct {
...
 
+    fn sendDrainGoaway(self: *Runtime, entry: *ConnEntry) void {
+        if (!entry.conn.isEstablished() or !entry.h3_started) return;
+        entry.h3.sendGoaway(entry.conn, 0) catch |err| {
+            self.logger.warn(null, "http3: failed to queue GOAWAY during drain: {s}", .{@errorName(err)});
+            return;
+        };
+        entry.drain_goaway_sent = true;
+    }
+
```
Jaruso, line 915: **Blocking — `GOAWAY(0)` is not a valid graceful boundary once any request has been accepted.** In HTTP/3, a server GOAWAY carries a client-initiated bidirectional stream ID identifying the first request the server will not process. Sending `0` says request stream 0 and every later request is not accepted. If stream 0 (or any later request) is already admitted and being allowed to finish, that contradicts the GOAWAY boundary and can cause the client to treat an actually-processing request as retryable/unprocessed.

Please track/select a boundary above all requests already admitted at drain start (or use the RFC's two-stage graceful GOAWAY pattern: an initial maximum valid client-bidi stream ID, then a lower final boundary once the accepted set is known). The admission check must use the same boundary rather than unconditionally rejecting every `pollRequest()` result merely because `drain_requested` is set. A test should admit at least one request, begin drain, verify its GOAWAY semantics permit that request to finish, and verify a request at/above the boundary is rejected with `H3_REQUEST_REJECTED`.

src/gateway_shutdown.zig
```diff
@@ -533,7 +583,20 @@ pub fn applyReloadedRuntimeConfig(cfg: *const edge_config.EdgeConfig, state: *Ga
     state.runtime_mutex.lock();
     state.add_headers = cfg.add_headers;
     if (state.http3_alt_svc) |value| state.allocator.free(value);
-    state.http3_alt_svc = if (cfg.http3_enabled) http.http3_handler.formatAltSvc(state.allocator, cfg.quic_port) catch null else null;
+    const runtime_ready = if (state.http3_runtime) |runtime| runtime.snapshot().server_bootstrapped else false;
+    const advertisement: http.http3_handler.Advertisement = if (!cfg.http3_enabled or !runtime_ready or cfg.http3_alt_svc == .off)
+        .disabled
+    else
+        .{ .active = .{
```
Jaruso: **Blocking for rollback semantics — `auto -> off` silently removes the header instead of withdrawing the cached alternative.** #257 explicitly requires disabling/rollback to be able to emit `Alt-Svc: clear`. Here an advertisement-only hot reload from `.auto` to `.off` produces `.disabled`, so subsequent TCP responses carry no Alt-Svc at all. Clients that cached the prior active advertisement keep using it until `ma` expires, meaning the documented staged rollback does not actually withdraw H3.

Please model the transition from an effective active advertisement to operator-disabled state as a withdrawal/`clear` phase (with a defined bounded lifetime/transition back to disabled), or otherwise provide an explicit rollback-withdrawal state/config that hot reload can select. Add a reload test that starts from an actively advertised runtime, reloads the rollback setting without reconstructing the runtime, and proves eligible TCP responses emit exactly `Alt-Svc: clear`.

Jaruso (commented):
**Changes needed before merge.** I found three lifecycle issues that prevent this from satisfying #257's drain/rollback acceptance criteria. The largest two are in the H3 drain path: once drain begins the socket loop drops every inbound UDP datagram, including packets for established connections, and GOAWAY is always sent with ID 0 even when request streams are already admitted. Together those make the advertised graceful-drain semantics incorrect. There is also no `Alt-Svc: clear` transition for an operator rollback from `auto` to `off`, so clients can retain the previously cached H3 alternative until `ma` expires.

Please address the inline comments and add production-level tests covering: (1) an established H3 request continuing bidirectional QUIC traffic and completing after drain begins, while a new connection/request is refused; (2) GOAWAY boundary selection with at least one already-admitted request stream; and (3) hot reload `auto -> off` actively emitting `Alt-Svc: clear` without reconstructing the listener.

I would treat these as blocking even though GitHub will not let me submit `REQUEST_CHANGES` on my own PR.

Jaruso (commented):
Second-pass review against #257 and the fixes for the prior three blockers.

The original defects are substantially addressed: existing routed QUIC traffic is no longer blanket-dropped during drain, GOAWAY now uses a request-stream boundary derived from admitted work, and `auto -> off` reload produces `Alt-Svc: clear`.

I found two remaining blockers before merge:

1. The runtime can enter its GOAWAY/draining behavior from the global shutdown flag before `beginDrain()` sets `drain_requested`. During that interval it sends a GOAWAY boundary, but `accept()` and `pumpH3()` still use only `drain_requested` for admission. A request/connection can therefore be accepted after the advertised GOAWAY boundary. The drain trigger needs one coherent source of truth.
2. #257 explicitly requires a production UDP/native-H3 lifecycle test proving an admitted request completes after drain begins, new work is refused according to GOAWAY semantics, and connections/CID routes reach zero. The added helper-level boundary test does not exercise any of those runtime guarantees, so the acceptance/test criteria remain incomplete.

CI for current head `cff40b56` is still running as of this review. Changes still needed before merge.

src/http/http3_runtime.zig
```diff
@@ -356,8 +391,11 @@ pub const Runtime = struct {
             per_ip.deinit();
         }
 
-        while (!self.stopping.load(.acquire) and !shutdown.isShutdownRequested()) {
+        while (!self.stopping.load(.acquire)) {
             const now = nowUs();
+            const draining = self.drain_requested.load(.acquire) or shutdown.isShutdownRequested();
+            const drain_deadline = self.drain_deadline_us.load(.acquire);
+            const drain_expired = draining and drain_deadline != 0 and now >= drain_deadline;
```
Jaruso, line 400: **Blocking — the drain trigger is split between two different predicates, which creates a GOAWAY/admission race.** Here `draining` becomes true as soon as the global `shutdown.isShutdownRequested()` flag flips, so this loop can call `sendDrainGoaway()` and advertise a fixed boundary *before* `edge_gateway` reaches `runtime.beginDrain()`. But both `accept()` and the request rejection path in `pumpH3()` only check `self.drain_requested`. During that window the runtime can therefore send (for example) `GOAWAY(4)` and then still accept a new connection or process request stream 4+, contradicting the boundary it just sent.

Please make the drain state single-source. The safest shape here is to let the gateway own the transition and have the runtime act only on `drain_requested`:

```zig
const draining = self.drain_requested.load(.acquire);
```

`edge_gateway` already observes shutdown and calls `beginDrain(deadline)`, so the runtime does not need to independently infer graceful drain from the process-global flag. Alternatively, if the global flag must remain a trigger, factor a helper such as `isDraining()` and use that exact predicate in the loop, `accept()`, and `pumpH3()` so admission closes atomically with GOAWAY behavior. Add a regression test where the shutdown signal becomes visible before `beginDrain()` and prove no request can be admitted beyond a sent boundary.

src/http/http3_runtime.zig
```diff
@@ -1906,6 +1986,16 @@ test "admissionAllowed enforces global and per-source caps at the boundary" {
     try testing.expect(!admissionAllowed(max_connections, max_connections_per_source));
 }
 
+test "drainBoundaryAfter permits admitted client-bidi streams" {
+    try testing.expectEqual(@as(u64, 0), drainBoundaryAfter(null));
+    try testing.expectEqual(@as(u64, 4), drainBoundaryAfter(0));
+    try testing.expectEqual(@as(u64, 12), drainBoundaryAfter(8));
```
Jaruso, line 2017: **Blocking acceptance/test gap — this helper test is not the graceful-drain coverage required by #257.** It verifies only arithmetic on the chosen boundary. The ticket explicitly requires the production UDP runtime/native H3 path to: establish H3, have an admitted request in flight, call gateway/runtime drain, observe GOAWAY, let that admitted request finish, reject a subsequent request, then prove `tracked_connections == 0` and CID routes are gone after completion/deadline (with Linux + macOS release coverage).

Please add an integration-style runtime test using the existing native QUIC/H3 client/test harness rather than only testing `drainBoundaryAfter()`. Conceptually:

```zig
var runtime = try Runtime.init(... .quic_port = 0 ...);
runtime.start();
// connect native H3 client and complete SETTINGS
// start request A and ensure the server has admitted it
runtime.beginDrain(Runtime.nowUsPublic() + test_timeout_us);
// observe GOAWAY boundary > request A's stream id
// finish request A and assert its response succeeds
// attempt request B at/above boundary and assert H3_REQUEST_REJECTED / no app dispatch
// drive until drained
try expect(runtime.isDrained());
const snap = runtime.snapshot();
try expectEqual(@as(usize, 0), snap.tracked_connections);
// assert CID-route count/metric is zero through the runtime's existing observable seam
```

This is particularly important because the first review already exposed a bug that helper/state tests would not catch: graceful drain depends on continued packet exchange after `beginDrain()`.

Jaruso (commented):
Re-review of current head `57adee15f1f23624cd0b90fdc10e9fbaab247695`: the two remaining blockers from my prior pass are addressed.

- The H3 runtime now uses `drain_requested` as the single drain predicate; it no longer infers graceful drain independently from the process-global shutdown flag, so GOAWAY/admission behavior is coherent.
- The new real UDP/native-H3 lifecycle test establishes H3, admits request A, starts drain, receives A successfully after drain starts, verifies request B is rejected without app dispatch, verifies a fresh Initial cannot create another tracked connection, and finally checks `tracked_connections == 0`, `active_cid_routes == 0`, and `isDrained()`.
- The UDP smoke harness is wired into the default `zig build test`, and CI runs that target on Linux and macOS, satisfying the cross-platform release guard once the current run is green.

I did another pass over the Alt-Svc rollout/reload path, GOAWAY boundary logic, connection/CID teardown, and effective-state metrics and did not find another blocking defect in this head. I consider the implementation ready to merge once CI completes successfully.

PR #540, opened 2026-07-29
Expand deterministic QUIC parser fuzz coverage

Jaruso:
## Summary
- add a checked-in QUIC/H3 fuzz and property matrix for #247/#537, including existing coverage and explicit follow-up gaps
- add packet/header/coalescing fuzz properties for bounded parsed slices, nonzero packet length, and coalesced progress/error behavior
- add QUIC frame/ACK fuzz properties for bounded consumption, payload slice ownership, ACK arithmetic boundaries, fixed-size truncation rejection, and canonical encoder round-trips
- add `-Dquic-test-filter` so focused QUIC/H3 fuzz targets can be selected through `zig build test-quic`

Refs #537.

## Validation
- `zig build test-quic -Doptimize=ReleaseFast -Dquic-test-filter="fuzz: packet parser" --fuzz=1 --summary all --error-style verbose`
- `zig build test-quic -Doptimize=ReleaseFast -Dquic-test-filter="fuzz: frame decoder" --fuzz=1 --summary all --error-style verbose`


src/quic/frame.zig
```diff
@@ -721,3 +772,90 @@ test "every decoded frame length covers exactly the consumed bytes" {
...
+
+    const decoded = decodeFrame(input) catch return;
+    try testing.expect(decoded.len > 0);
+    try testing.expect(decoded.len <= input.len);
+    try expectFrameSlicesWithin(input, decoded.frame);
+
+    var parser = Parser.init(input);
+    var last_pos: usize = 0;
+    var count: usize = 0;
+    while (try parser.next()) |frame| {
```
Jaruso: **Blocking: typed malformed-tail errors become false fuzz failures.**

At this point the callback has already established that the first frame is structurally valid. A later frame in the same arbitrary input is allowed to be malformed/truncated/unknown; #537's parser invariant is that iteration makes progress **or returns a typed error**. `try parser.next()` propagates that ordinary `DecodeError` out of the fuzz callback, so an input such as a valid PING followed by an unknown/truncated frame is reported as a fuzz failure instead of a successful typed rejection.

Please consume later parser errors as an allowed terminal outcome while still checking progress for every successful frame, e.g.:

```zig
while (true) {
    const maybe_frame = parser.next() catch return;
    const frame = maybe_frame orelse break;

    try testing.expect(parser.pos > last_pos);
    try testing.expect(parser.pos <= input.len);
    try expectFrameSlicesWithin(input, frame);
    last_pos = parser.pos;
    count += 1;
}
```

A small permanent regression with `{ 0x01, 0x21 }` (PING + unknown frame) would also make the intended property explicit.

src/quic/frame.zig
```diff
@@ -721,3 +772,90 @@ test "every decoded frame length covers exactly the consumed bytes" {
...
+        1 => _ = try roundtripOne(buf[0..try encodeHandshakeDone(&buf)]),
+        2 => {
+            const frame = try roundtripOne(buf[0..try encodeCrypto(@as(u64, smith.value(u16)), data[0..data_len], &buf)]);
+            try testing.expectEqualSlices(u8, data[0..data_len], frame.crypto.data);
+        },
+        3 => {
+            const frame = try roundtripOne(buf[0..try encodeStream(id, @as(u64, smith.value(u16)), data[0..data_len], smith.value(u1) == 1, &buf)]);
+            try testing.expectEqualSlices(u8, data[0..data_len], frame.stream.data);
+        },
+        4 => _ = try roundtripOne(buf[0..try encodeMaxData(limit, &buf)]),
```
Jaruso: **Blocking: this is not yet a semantic round-trip property.**

`roundtripOne()` only asserts that the decoder consumed the encoder's byte count. In this branch, for example, if `encodeMaxData(limit)` accidentally encoded `0` for every nonzero `limit`, the decoder would consume the entire generated frame and this fuzz case would still pass. The same issue applies to most branches below/above this one; CRYPTO and STREAM compare only payload bytes and omit offset/id/FIN.

#537 explicitly requires successful canonical encode/decode round-trips to preserve semantic fields. Please compare every generated field to the decoded frame. For example:

```zig
4 => {
    const frame = try roundtripOne(buf[0..try encodeMaxData(limit, &buf)]);
    try testing.expectEqual(limit, frame.max_data);
},
```

For STREAM, retain the generated values and verify `id`, `offset`, `fin`, and `data`; do the equivalent for MAX_STREAM_DATA/MAX_STREAMS, RESET_STREAM, STOP_SENDING, PATH_* and CONNECTION_CLOSE. It would also be appropriate to bring the existing ACK / NEW_CONNECTION_ID / RETIRE_CONNECTION_ID encoders into this property target so the matrix's “supported canonical encodings” claim matches the production encoder surface.

docs/QUIC_H3_FUZZ_MATRIX.md
```diff
@@ -0,0 +1,43 @@
...
+| QUIC varints | `src/quic/varint.zig` `fuzz: varint decode and minimal re-encode never panic`; `fuzz: varint encode round-trips arbitrary in-range values` | Minimal re-encode, in-range encode/decode round-trip, truncation rejection. | None for the current varint codec surface. |
+| QUIC packet numbers | `src/quic/packet.zig` `fuzz: packet number truncation reconstructs recent sends` | Packet-number length selection, truncation, and reconstruction near recent-send windows. | Add recovered packet-number comparisons from loss/reordering driver state when that state exposes a narrow property API. |
+| QUIC packet/header/coalescing | `src/quic/packet.zig` `fuzz: packet parser preserves bounded slice and progress invariants`; deterministic coalesced invalid-tail regression tests | Empty/truncated inputs, fixed-bit failures, long/short header CID boundaries, version-negotiation length boundaries, Initial token/Length cursor handling, Retry token/tag split, short-header caller CID length, coalesced progress, parsed slices staying inside the input, nonzero bounded `packet_len`. | Add canonical public-header re-encode properties if a read-side public-header writer is introduced. |
```
Jaruso: **A-scope gap: the prerequisite in this follow-up is already present.**

`src/quic/packet.zig` already exposes `writeLongHeader()` and `writeShortHeader()` and has a Retry writer. #537 requires deterministic public-header fields to round-trip where canonical encoding exists. The fixed-value unit tests prove a couple of examples, but this PR's new Smith packet property only mutates the parser input; it does not generate writer inputs and compare the parsed public fields back to those inputs.

Please either add the property in this slice or mark this row as genuinely still-open instead of saying it is waiting for a writer. A useful property would vary long-header kind/version/DCID/SCID/token/pn length/Length, then assert parsed kind/version/CIDs/token/pn offset/packet length; similarly vary short-header CID/key-phase/pn length for the fields the parser exposes.

docs/QUIC_H3_FUZZ_MATRIX.md
```diff
@@ -0,0 +1,43 @@
...
+external peers out of these loops; ngtcp2/nghttp3, quiche, and aioquic remain under
+`scripts/interop/run-interop.sh`.
+
+Use `-Doptimize=ReleaseFast` for coverage-guided runs with Zig 0.16.0; the
+ordinary deterministic smoke tests continue to run in the default Debug mode.
+
+| Area | Existing or New Target | Properties Covered | Open Follow-Up |
+| --- | --- | --- | --- |
+| QUIC varints | `src/quic/varint.zig` `fuzz: varint decode and minimal re-encode never panic`; `fuzz: varint encode round-trips arbitrary in-range values` | Minimal re-encode, in-range encode/decode round-trip, truncation rejection. | None for the current varint codec surface. |
+| QUIC packet numbers | `src/quic/packet.zig` `fuzz: packet number truncation reconstructs recent sends` | Packet-number length selection, truncation, and reconstruction near recent-send windows. | Add recovered packet-number comparisons from loss/reordering driver state when that state exposes a narrow property API. |
```
Jaruso: **A-scope gap: the packet-number property does not reach the codec's upper boundaries.**

The referenced `fuzzPacketNumberRoundTrip` currently builds `largest` from `smith.value(u32)` and `delta` from `smith.value(u16)`, so the generated full packet number stays around the 32-bit range. The public codec allows values through `max_packet_number == 2^62 - 1`, and #537 calls for packet-number encode/truncate/reconstruction boundary properties.

Please extend the property (or add deterministic property cases) to exercise the upper legal range, especially values/windows near `max_packet_number`. One safe shape is:

```zig
const largest = smith.value(u64) & max_packet_number;
if (largest == max_packet_number) return;
const remaining = max_packet_number - largest;
const delta = 1 + (@as(u64, smith.value(u32)) % @min(remaining, @as(u64, std.math.maxInt(u32))));
const full = largest + delta;
```

and seed exact cases near `max_packet_number` / reconstruction half-window boundaries so the default deterministic pass retains them.

Jaruso (commented):

I reviewed this as the packet/frame first slice of #537 rather than as closure of the whole issue. The inventory is useful and the overall direction is right, and CI is green at the current head. I am not blocking this PR on the explicitly documented transport-parameter, Retry-token, CRYPTO-reassembly, stream/flow-control, stateful-QPACK, or H3 state-machine follow-ups.

I found four items that are part of the A/B scope this PR claims to cover and should be fixed before merge:

1. The frame parser fuzz callback treats a typed error after a valid frame prefix as a fuzz failure. A malformed trailing frame is expected parser behavior under #537; the invariant is progress *or a typed error*. This will create false-positive failures during longer fuzz runs.
2. The canonical frame round-trip fuzz target mostly checks only that decode consumed the bytes emitted by encode. It does not compare the generated semantic fields to the decoded fields, so an encoder that writes the wrong value can still pass. #537 explicitly requires semantic-field preservation.
3. The matrix defers canonical packet-header round-trip fuzzing until a writer exists, but `packet.zig` already has `writeLongHeader` and `writeShortHeader` (plus Retry writing). The stated follow-up condition is already satisfied, so the A-slice is missing the requested property coverage.
4. The packet-number fuzz generator only samples a `u32` largest packet number plus a `u16` delta, while the codec accepts packet numbers through `2^62 - 1`. That leaves the upper-range reconstruction/truncation boundaries outside the property test even though the matrix presents packet-number boundary coverage as complete.

I left concrete fixes inline. Once these are addressed, the rest of this PR looks appropriate as a first #537 slice.

_Note: GitHub does not allow the PR author to submit a formal `REQUEST_CHANGES` review on their own PR, so this is submitted as a comment review; the findings above are blocking recommendations._

Jaruso (commented):
## Follow-up review: looks good to merge

Re-reviewed the new commit (`a33679a`) against all four findings from my first pass and then swept the updated packet/frame fuzz changes again for regressions.

All four blockers are addressed:

1. Frame iteration now treats a typed malformed/truncated tail as an allowed terminal result while preserving progress checks for successfully parsed frames, with a permanent PING+unknown-tail regression.
2. Canonical frame fuzzing now verifies semantic fields rather than only byte consumption, including CRYPTO/STREAM metadata, flow-control frames, reset/stop-sending, PATH frames, ACK ranges/delay, NEW_CONNECTION_ID, RETIRE_CONNECTION_ID, and CONNECTION_CLOSE.
3. Packet writer/parser property coverage now exercises long headers, short headers, and Retry and compares the public fields exposed by `parsePacket`.
4. Packet-number generation now spans the legal 62-bit packet-number space and includes deterministic upper-bound regressions near `max_packet_number`.

I also checked the new generators for bounded buffers, Retry writer preconditions, ACK range construction, packet Length construction, and the matrix claims against the A/B scope of #537. I did not find an additional blocking defect in the updated code.

The remaining transport-parameter, Retry-token envelope, CRYPTO reassembly, stream/flow-control state, stateful QPACK, and H3 state-machine items are still correctly recorded as follow-up work for #537 rather than being silently claimed complete here.

CI run 1064 is still in progress as of this review, but the completed format, unit-test, dependency-review, and multiple build jobs are green. No further code changes requested from this review.

PR #541, opened 2026-07-29
test(quic): expand deterministic fuzz coverage

Jaruso:
## Summary

Extends #537 parser/state fuzz coverage in PR #541:

- Keeps the A/B packet/header/coalescing + frame/ACK closeout from the earlier PR revision.
- Adds C transport-parameter raw decode + semantic-validation fuzz/property coverage in `src/quic/tls_backend.zig`.
- Adds D Retry-token offline issue/validate + mutation/property coverage in `src/quic/path.zig`.
- Updates `docs/QUIC_H3_FUZZ_MATRIX.md` with exact target names, focused fuzz commands, covered properties, and remaining #537 follow-up.

Transport-parameter coverage includes structural malformed input, duplicate known/tracked-unknown IDs, unknown skippability past the bounded duplicate cache, semantic numeric boundaries, symmetric local config/encode/decode enforcement of `initial_max_streams_* <= 2^60`, migration flag handling, CID binding lengths, stateless reset-token length checks, bounded arbitrary decode invariants, and canonical encode/decode round trips for supported `TransportParameters` and `CidBinding` fields.

Retry-token coverage includes deterministic IPv4/IPv6 issue/validate round trips, scoped IPv6, key rotation/retirement, nonce/ciphertext/tag mutations, public short/oversized/truncated input, authenticated malformed plaintext, wrong-kind tokens, address/port binding, exact expiry/future-skew boundaries, and `u64` saturation behavior.

Refs #537

## Production fixes

- Enforce RFC 9000's `initial_max_streams_bidi` / `initial_max_streams_uni` upper bound of `2^60` in local config validation, outbound transport-parameter encoding, and peer transport-parameter decoding.
- Keep distinct well-formed unknown transport parameters skippable after the bounded duplicate-tracking cache fills, while preserving deterministic duplicate rejection for known parameters and tracked unknown IDs.

## Validation

- `zig build test-quic --summary all --error-style verbose` (663/663 passed)
- `zig build test --summary all --error-style verbose` (2905/2906 passed, 1 skipped)
- `zig build test-quic -Doptimize=ReleaseFast -Dquic-test-filter="fuzz: transport parameter decoder" --fuzz=1 --summary all --error-style verbose` (228/228 passed)
- `zig build test-quic -Doptimize=ReleaseFast -Dquic-test-filter="fuzz: transport parameters canonical" --fuzz=1 --summary all --error-style verbose` (228/228 passed)
- `zig build test-quic -Doptimize=ReleaseFast -Dquic-test-filter="fuzz: Retry token issue validate" --fuzz=1 --summary all --error-style verbose` (228/228 passed)

## Remaining #537 work

E-H remain open after this PR:

- CRYPTO reassembly
- stream/flow-control command sequences
- stateful QPACK
- H3 control/request/critical-stream state


src/quic/packet.zig
```diff
@@ -510,6 +510,34 @@ test "coalesced packets split on the long-header Length field" {
...
+    buf[handshake_end + 1] = 0x00;
+
+    const one = try parsePacket(buf[0 .. handshake_end + 2], dcid.len);
+    try testing.expectEqual(initial_end, one.packet_len);
+    const two = try parsePacket(buf[one.packet_len .. handshake_end + 2], dcid.len);
+    try testing.expectEqual(PacketKind.handshake, two.kind);
+    try testing.expectEqual(handshake_end - initial_end, two.packet_len);
+    try testing.expectError(error.TruncatedPacket, parsePacket(buf[one.packet_len + two.packet_len .. handshake_end + 2], dcid.len));
+
+    try testing.expectError(error.TruncatedPacket, parsePacket(buf[handshake_end .. handshake_end + 2], dcid.len));
```
Jaruso: **[P1] Add the required `invalid + valid` coalesced ordering instead of repeating the same suffix check.**

The assertion immediately above already evaluates `buf[handshake_end .. handshake_end + 2]` through the equivalent `one.packet_len + two.packet_len` expression, so this line does not add a new case. #537 explicitly requires coalesced `valid+invalid`, `invalid+valid`, and multiple-valid-prefix combinations. This test currently covers multiple valid prefixes followed by an invalid tail, but never an invalid first packet with a valid packet behind it.

Please replace this duplicate with a deterministic invalid-prefix + valid-packet case. A robust shape is to make the first long header fail on an oversized DCID, then append a fully valid Handshake packet and assert that parsing the whole datagram still returns `error.InvalidConnectionId` rather than skipping/reinterpreting the malformed prefix, e.g. conceptually:

```zig
var datagram: [256]u8 = undefined;
datagram[0] = 0xc0;
std.mem.writeInt(u32, datagram[1..5], quic_v1, .big);
datagram[5] = max_cid_len + 1;
@memset(datagram[6..27], 0);

const valid_start = 27;
const valid = try writeLongHeader(.handshake, quic_v1, &dcid, &scid, "", 1, datagram[valid_start..]);
patchLongHeaderLength(datagram[valid_start..], valid.length_offset, 1);
datagram[valid_start + valid.pn_offset] = 0xaa;
const end = valid_start + valid.pn_offset + 1;

try testing.expectError(error.InvalidConnectionId, parsePacket(datagram[0..end], dcid.len));
```

The exact malformed prefix can differ; the important property is that a valid later packet cannot make an invalid leading packet disappear.

src/quic/packet.zig
```diff
@@ -522,6 +550,51 @@ test "short header parses with caller-provided DCID length" {
...
+    buf[5] = 0;
+    buf[6] = 0;
+    try testing.expectError(error.MalformedPacket, parsePacket(buf[0..7], 0));
+    std.mem.writeInt(u32, buf[7..11], quic_v1, .big);
+    const vn = try parsePacket(buf[0..11], 0);
+    try testing.expectEqual(PacketKind.version_negotiation, vn.kind);
+    try testing.expectEqual(@as(usize, 4), vn.supported_versions.len);
+    try testing.expectError(error.MalformedPacket, parsePacket(buf[0..10], 0));
+
+    const one_over = [_]u8{ 0xc0, 0, 0, 0, 1, max_cid_len + 1 } ++ [_]u8{0} ** (max_cid_len + 1);
```
Jaruso, line 588: **[P1] Cover max+1 SCID as well as max+1 DCID.**

This hits the `dcid_len > max_cid_len` branch, but `parsePacket()` has an independent `scid_len > max_cid_len` rejection path. #537 explicitly calls out **DCID/SCID lengths `0`, exact max, max+1** where wire-reachable, and the matrix now says there is no remaining packet/header follow-up.

Please add the sibling case so the separate branch is permanently covered:

```zig
const scid_one_over = [_]u8{
    0xc0, 0, 0, 0, 1,
    0,                 // DCID length
    max_cid_len + 1,   // SCID length
} ++ [_]u8{0} ** (max_cid_len + 1);
try testing.expectError(error.InvalidConnectionId, parsePacket(&scid_one_over, 0));
```

That gives deterministic coverage for both length guards rather than relying on arbitrary fuzz mutation to eventually reach the SCID branch.

Jaruso (commented):

I reviewed PR #541 end-to-end against #537 and the current parser implementations.

The frame additions look correct: decoder-only families are exercised with valid semantics, implicit STREAM consumption is checked, and ACK_ECN/truncation/huge-gap cases hit the intended typed-error paths. The packet boundary tests are also directionally good, and the current CI run is green for the completed jobs.

I found two remaining A-scope acceptance gaps in the new boundary matrix:

1. The new coalescing test covers valid + valid + invalid, but not the explicit `invalid + valid` ordering required by #537. The final assertion currently repeats the same invalid suffix that was already checked on the previous line.
2. The max+1 CID regression only exercises the DCID rejection branch. `parsePacket()` has a separate SCID length check, and #537 explicitly requires DCID/SCID lengths at 0, max, and max+1 where wire-reachable. Add a max+1 SCID case before the matrix continues to claim no packet/header follow-up.

There is also a scope/tracking point: this PR is an A/B boundary-hardening follow-up to #540, not the transport-parameter + Retry-token C/D slice we discussed as the next #537 step. The matrix accurately leaves C and D open, so I would merge this as an A/B closeout after the two gaps above are fixed, then do C+D in the next PR rather than counting #541 as that slice.

Because the connected account owns this PR, GitHub will not allow a formal `REQUEST_CHANGES` review; this COMMENT review should be treated as blocking until the two inline findings are addressed.

src/quic/tls_backend.zig
```diff
@@ -599,6 +599,243 @@ test "QUIC adapter owns CID binding round trip" {
...
+    try expectTpIntRejected(tp_ack_delay_exponent, 21);
+    try std.testing.expectEqual(@as(u64, (1 << 14) - 1), (try decodeOneTpInt(tp_max_ack_delay, (1 << 14) - 1)).max_ack_delay_ms);
+    try expectTpIntRejected(tp_max_ack_delay, 1 << 14);
+
+    try std.testing.expectEqual(@as(u64, 0), (try decodeOneTpInt(tp_initial_max_data, 0)).initial_max_data);
+    try std.testing.expectEqual(@as(u64, 7), (try decodeOneTpInt(tp_initial_max_stream_data_bidi_local, 7)).initial_max_stream_data_bidi_local);
+    try std.testing.expectEqual(max_quic_varint, (try decodeOneTpInt(tp_initial_max_stream_data_bidi_remote, max_quic_varint)).initial_max_stream_data_bidi_remote);
+    try std.testing.expectEqual(@as(u64, 0), (try decodeOneTpInt(tp_initial_max_stream_data_uni, 0)).initial_max_stream_data_uni);
+    try std.testing.expectEqual(@as(u64, 3), (try decodeOneTpInt(tp_initial_max_streams_bidi, 3)).initial_max_streams_bidi);
+    try std.testing.expectEqual(max_quic_varint, (try decodeOneTpInt(tp_initial_max_streams_uni, max_quic_varint)).initial_max_streams_uni);
```
Jaruso: **[P1] `initial_max_streams_*` is capped at `2^60`, not the QUIC varint maximum.**

This regression currently asserts that `initial_max_streams_uni = 2^62 - 1` is valid, which locks in a protocol error. RFC 9000 §4.6 requires a `max_streams` transport parameter greater than `2^60` to close the connection with `TRANSPORT_PARAMETER_ERROR`; the same bound applies to both `initial_max_streams_bidi` and `initial_max_streams_uni`.

The decoder currently assigns both values directly from `integerParameter()` with no semantic bound. Please add the production validation and test both exact boundaries, e.g.:

```zig
const max_initial_streams: u64 = 1 << 60;

// in decodeTransportParametersBound
 tp_initial_max_streams_bidi => {
    const value = try integerParameter(value_bytes);
    if (value > max_initial_streams) return error.InvalidTransportParameters;
    params.initial_max_streams_bidi = value;
},
 tp_initial_max_streams_uni => {
    const value = try integerParameter(value_bytes);
    if (value > max_initial_streams) return error.InvalidTransportParameters;
    params.initial_max_streams_uni = value;
},
```

Then permanently assert `2^60` is accepted and `2^60 + 1` is rejected for **both** parameters. The canonical fuzz generator also needs to generate these fields only in `0...2^60`, otherwise fixing the decoder will turn valid-generator cases into fuzz failures.

src/quic/tls_backend.zig
```diff
@@ -599,6 +599,243 @@ test "QUIC adapter owns CID binding round trip" {
...
+
+    var many: [512]u8 = undefined;
+    var pos: usize = 0;
+    var id: u64 = 0x21;
+    while (id < 0x21 + max_distinct_transport_parameters) : (id += 1) {
+        try appendTpRaw(&many, &pos, id, &.{});
+    }
+    _ = try decodeTransportParameters(many[0..pos]);
+    try appendTpRaw(&many, &pos, 0x21 + max_distinct_transport_parameters, &.{});
+    try std.testing.expectError(error.InvalidTransportParameters, decodeTransportParameters(many[0..pos]));
```
Jaruso: **[P1] Do not make the 65th distinct unknown transport parameter a protocol error.**

This test codifies `max_distinct_transport_parameters` as a wire-visible limit: 64 distinct well-formed unknown IDs succeed, then the 65th produces `InvalidTransportParameters`. RFC 9000 §7.4.2 says an endpoint **MUST ignore transport parameters that it does not support**, and §18.1 reserves IDs specifically to exercise that property. #537 likewise requires unknown well-formed parameters to remain skippable.

The fixed `seen_ids` array is useful for bounded duplicate detection, but saturation cannot reject a new unknown ID. One bounded shape is to track known singleton IDs separately and make unknown duplicate tracking best-effort:

```zig
var seen_known: u32 = 0;
var seen_unknown: [max_tracked_unknown_parameters]u64 = undefined;
var unknown_count: usize = 0;

if (knownTransportParameterBit(id.value)) |bit| {
    if (seen_known & bit != 0) return error.InvalidTransportParameters;
    seen_known |= bit;
} else {
    for (seen_unknown[0..unknown_count]) |seen| {
        if (seen == id.value) return error.InvalidTransportParameters;
    }
    // Duplicate rejection is a SHOULD; unknown-parameter ignore is a MUST.
    // Once the bounded cache is full, keep parsing/ignoring new unknown IDs.
    if (unknown_count < seen_unknown.len) {
        seen_unknown[unknown_count] = id.value;
        unknown_count += 1;
    }
}
```

Equivalent bounded designs are fine. The regression here should prove that **more than 64 distinct unknown parameters still decode successfully**, rather than expecting an error solely because the tracking cache filled.

Jaruso (commented):
## Follow-up review: two C-scope changes required before merge

I re-reviewed the force-pushed head end-to-end against #537 A-D, including the two prior A/B findings and the newly added transport-parameter / Retry-token work.

The two previous A/B blockers are fixed: the coalescing matrix now contains an actual invalid+valid ordering, and the independent max+1 SCID branch is covered. The Retry-token D slice also looks good: deterministic issue/validate, IPv4/IPv6 + scope binding, key rotation/retirement, mutation/truncation, authenticated malformed plaintext, expiry/skew/u64 boundaries, and the existing runtime tests already cover the delegated wrong-Retry-SCID and wrong-QUIC-version comparisons.

I found two blocking issues in the C transport-parameter slice:

1. The new semantic regression treats `initial_max_streams_uni = 2^62-1` as valid. RFC 9000 limits both `initial_max_streams_bidi` and `_uni` to `2^60`; a larger transport-parameter value requires `TRANSPORT_PARAMETER_ERROR`. The production decoder currently lacks this validation.
2. The new distinct-ID regression intentionally rejects the 65th distinct well-formed unknown transport parameter because `seen_ids` is full. RFC 9000 requires unsupported transport parameters to be ignored; a bounded duplicate-tracking implementation cannot turn additional distinct unknown IDs into a connection error. This also conflicts with #537's unknown-skippability property.

I left concrete bounded fixes inline. After those are corrected, I do not currently see another A-D blocker. CI is green at this head.

Because the connected account owns this PR, GitHub will not accept a formal REQUEST_CHANGES self-review, so this COMMENT review should be treated as changes-required.

src/quic/tls_backend.zig
```diff
@@ -54,6 +54,7 @@ const tp_ack_delay_exponent: u64 = 0x0a;
 const tp_max_ack_delay: u64 = 0x0b;
 const tp_initial_source_connection_id: u64 = 0x0f;
 const tp_retry_source_connection_id: u64 = 0x10;
+const max_initial_streams_transport_parameter: u64 = 1 << 60;
 
 pub const max_transport_parameters_len = 11 * (2 + 1 + 8) + 3 + 3 * (2 + 1 + config.max_cid_len) + (2 + 1 + 16);
```
Jaruso, line 58: **[P1] Apply the new `2^60` stream-count ceiling to the outbound path too.**

The receive-side fix is correct, but this constant currently only protects `decodeTransportParametersBound()`. The normal local path can still emit an invalid transport parameter: `Connection.init()` accepts `options.config.transportParameters()`, `Config.validate()` does not cap either `initial_max_streams_bidi` or `initial_max_streams_uni`, and `Tls13Backend.start()` passes those values straight into `encodeTransportParametersBound()`. A configured value in `(2^60, 2^62-1]` therefore encodes successfully on the wire even though RFC 9000 §4.6 requires the peer to close with `TRANSPORT_PARAMETER_ERROR` when it receives it.

Please make the bound symmetric so this codec never emits a value its decoder correctly rejects. At minimum, reject it in the encoder:

```zig
pub fn encodeTransportParametersBound(
    params: config.TransportParameters,
    binding: config.CidBinding,
    buf: []u8,
) HandshakeError![]const u8 {
    if (params.initial_max_streams_bidi > max_initial_streams_transport_parameter or
        params.initial_max_streams_uni > max_initial_streams_transport_parameter)
    {
        return error.InvalidTransportParameters;
    }
    // existing encoding...
}
```

Even better, share the limit with `config.zig` and reject it in `Config.validate()` as well so operator configuration fails before the handshake starts. Please add permanent tests for both bidi and uni at `2^60` (accepted) and `2^60 + 1` (rejected on the local/encode path), not only on decode.

Jaruso (commented):
Follow-up review of PR #541 at `5deed065e8ba858000d012318eaf8ed5d50a5708`.

The two blockers from my previous pass are fixed correctly:
- `initial_max_streams_bidi` / `_uni` now reject values above `2^60`, with exact-boundary regressions and a valid-range canonical fuzz generator.
- the bounded unknown-parameter duplicate cache no longer turns the 65th+ distinct well-formed unknown parameter into a protocol error; known duplicates remain deterministic and tracked unknown duplicates remain rejected.

I re-swept the A-D integration. The A/B regressions remain good, and the D Retry-token slice still looks correctly scoped: offline issue/validate/mutation coverage is bounded, while wrong Retry-SCID/version comparison remains owned and permanently tested at the runtime boundary.

One remaining change is required before merge: the new `2^60` semantic limit is only enforced when decoding peer parameters. The local config/encoder path can still successfully emit `initial_max_streams_* > 2^60`, producing a transport parameter a conforming peer must reject. I left the concrete fix and regression shape inline.

CI run 1070 is green. After outbound validation is made symmetric, I do not currently see another A-D blocker.

Jaruso (commented):
Re-review complete at `5c7aeb4ac3845e2665276d0ef726f888742cfa23`.

The remaining outbound transport-parameter blocker is fixed correctly:

- `config.max_initial_streams_transport_parameter` centralizes the RFC 9000 `2^60` ceiling.
- `Config.validate()` rejects local bidi/uni stream-count configuration above that ceiling.
- `encodeTransportParametersBound()` independently refuses to emit an invalid value, so direct codec callers cannot bypass config validation.
- `decodeTransportParametersBound()` enforces the same ceiling on peer input.
- Deterministic tests cover the exact accepted boundary and `+1` rejection, and the canonical fuzz generator remains inside the legal range.

I also rechecked the previous A/B fixes, bounded unknown-parameter handling, C transport-parameter coverage, D Retry-token coverage, and the documented remaining #537 scope. I found no new blocking issues.

CI run 1071 is green.

Because this PR is authored by the connected GitHub user, I cannot submit an APPROVE review on the user's own PR, but my review verdict is: **merge-ready**. This completes the intended A-D coverage in this PR; #537 E-H remain for subsequent slices.

PR #544, opened 2026-07-30
crypto: add provider-owned QUIC header protection

Jaruso:
## Summary
- add a narrow CryptoProvider operation for QUIC AES-128 header-protection masks and advertise it in the checked-in crypto profile
- add provider-backed QUIC initial/key-update/AEAD/header-protection APIs while documenting existing compatibility exceptions
- add a deterministic crypto-boundary audit guard plus native crypto ownership audit docs

## Tests
- zig build test-crypto -freference-trace
- zig build test-quic -freference-trace
- zig build test -freference-trace

Refs #490


src/quic/tls_adapter.zig
```diff
@@ -206,6 +208,28 @@ pub const PacketProtectionKeys = struct {
         return out[0..required_len];
     }
 
+    pub fn sealPayloadWithProvider(
+        self: *const PacketProtectionKeys,
+        provider: crypto_provider.CryptoProvider,
```
Jaruso, line 226: [P1] This creates a second opt-in path but does not migrate the live QUIC ownership boundary. `QuicTlsAdapter.installInitialSecrets`, `protectionKeys`, `sealPacketPayload`, `openPacketPayload`, `updateApplicationWriteKeys`, and the read-key update methods still call the concrete helpers, while `connection.zig` still calls `applyHeaderProtection` / `removeHeaderProtection`. Consequently these provider methods are not used by production QUIC and several are not exercised at all.

Please make the provider a required adapter dependency and route the canonical APIs through it, for example:

```zig
pub const QuicTlsAdapter = struct {
    crypto_provider: crypto_provider.CryptoProvider,
    // ...

    pub fn protectionKeys(self: *const QuicTlsAdapter, level: EncryptionLevel, direction: Direction) !?PacketProtectionKeys {
        // existing secret checks...
        return try deriveAes128GcmKeysWithProvider(self.crypto_provider, secret_bytes[0..traffic_secret_len].*);
    }

    pub fn sealPacketPayload(...) ![]u8 {
        const keys = try self.protectionKeys(level, direction) orelse return error.KeysUnavailable;
        return keys.sealPayloadWithProvider(self.crypto_provider, packet_number, header, plaintext, out);
    }
};
```

Then route header apply/remove and key-update derivation through the same stored provider. If compatibility wrappers must remain temporarily, keep them clearly test-only or behind a narrowly documented legacy constructor so the normal runtime cannot silently choose the concrete path.

scripts/audit-crypto-boundary.sh
```diff
@@ -0,0 +1,46 @@
...
+ROOT="${1:-.}"
+
+cd "$ROOT"
+
+status=0
+
+check_file() {
+    file="$1"
+    pattern="$2"
+    rationale="$3"
```
Jaruso: [P2] This guard fails open when `rg` is unavailable. Because every invocation is the condition of an `if`, `set -e` does not terminate on command-not-found; stderr is discarded, all checks evaluate false, and the script exits 0. Fail closed before running any checks:

```sh
if ! command -v rg >/dev/null 2>&1; then
    printf '%s\n' 'audit-crypto-boundary: ripgrep is required' >&2
    exit 2
fi
```

Using a repository-provided Zig/source-audit executable instead of an ambient tool would be even more deterministic.

scripts/audit-crypto-boundary.sh
```diff
@@ -0,0 +1,46 @@
...
+        cat /tmp/tardigrade-crypto-boundary.$$
+        printf '%s\n' "rationale: $rationale"
+        status=1
+    fi
+    rm -f /tmp/tardigrade-crypto-boundary.$$
+}
+
+# Protocol modules must not add new concrete keyed crypto dependencies. Keep
+# public transcript hashing, parsing, nonce arithmetic, and shared secret
+# containers out of this grep; those are reviewed separately by code owners.
```
Jaruso: [P1] The forbidden pattern does not include `std.crypto.core.aes`, `crypto.core.aes`, `Aes128.initEnc`, or an aliased AES block `encrypt` call. That means the exact direct keyed header-protection implementation being moved out of protocol code can be copied into `connection.zig` and this guard will still pass.

Please match the block-cipher ownership forms explicitly, e.g. include patterns such as `std\\.crypto\\.core\\.aes`, `crypto\\.core\\.aes`, and `Aes(128|256)\\.initEnc`, and add a small audit fixture/test that proves AEAD, HKDF, ECDH/signature, and AES header-protection examples each make the guard fail. Keep the existing Retry-integrity exception path-specific rather than omitting AES block primitives globally.

src/crypto/profile.zig
```diff
@@ -87,9 +89,10 @@ pub const rows = [_]Row{
...
+    row(.{ .quic_header_protection = .aes_128 }, "QUIC AES-128 header protection", .zig_std_crypto, .supported, .openssl_provider, .provider_deferred, "RFC 9001 header-protection sample and provider mask parity", .{.quic_packet_protection}, "Narrow provider-owned keyed QUIC header-protection operation for TLS_AES_128_GCM_SHA256."),
     row(.{ .group = .x25519 }, "X25519", .zig_std_crypto, .supported, .openssl_provider, .provider_deferred, "std.crypto scalar multiplication parity, low-order rejection", .{ .tls_handshake, .quic_tls_bridge }, "Group additions require key-share and shared-secret vectors."),
     row(.{ .group = .secp256r1 }, "secp256r1 (P-256)", .unavailable, .provider_deferred, .openssl_provider, .provider_deferred, "capability rejection tests", .{ .tls_handshake, .quic_tls_bridge, .pki }, "P-256 support requires explicit provider implementation and ECDH vectors."),
```
Jaruso: [P1] This row now claims a live `quic_packet_protection` integration, but the provider-owned mask operation is only exercised by the vector test; the runtime still uses the concrete `headerProtectionMask`. More broadly, `Row` still has provider status plus a consumer set and free-form review text, so it does not explicitly distinguish provider capability, protocol integration status, and named product-profile enablement as #490 requires.

Please either mark this integration deferred until the live adapter is migrated, or land the runtime migration in this PR. I would also model the missing dimensions directly, for example with an `IntegrationStatus` (or per-consumer integration map) and `enabled_profiles: EnumSet(ProductProfile)` containing named profiles such as `.native_appliance` and `.general_purpose`, rather than relying on prose to carry product enablement.

Jaruso (commented):
**Blocking findings — changes needed before merge.** GitHub does not permit the PR author to submit a formal `REQUEST_CHANGES` review on their own PR, so this is recorded as a comment review.

The narrow `CryptoProvider` header-protection primitive and pure-Zig implementation are well shaped, and the AES-256/ChaCha consumer corrections improve the profile. The QUIC ownership slice is not actually wired into the live runtime, however, and the new enforcement guard can be bypassed.

The main runtime path still derives keys, seals/opens payloads, performs key updates, and applies/removes header protection through the original concrete `std.crypto` methods. The `*WithProvider` methods are parallel opt-in APIs used only by limited vector coverage, so `docs/CRYPTO_PROVIDER.md` currently overstates that native QUIC/H3 packet protection runs through `CryptoProvider`.

The source guard also misses the exact `crypto.core.aes.Aes128` / `Aes128.initEnc` form used for AES header protection, and it silently exits successfully when `rg` is unavailable. Both undermine #490's deterministic enforcement requirement.

Please also make the profile's three dimensions explicit rather than encoding protocol integration and product enablement through a consumer set plus free-form review text. A typed integration status and named enabled-product-profile set would make the acceptance criterion enforceable and prevent rows from claiming a live provider-owned consumer before the runtime is migrated.

I am treating this as an incremental PR because it says `Refs #490` and the audit explicitly leaves TLS key-schedule work open; #490 should remain open after this merges. For this PR, however, the QUIC slice needs to make the provider-backed path canonical and exercise it through the existing end-to-end QUIC tests. The current CI run is also red in one x86 appliance smoke test (`HTTP/3 runtime drain lets admitted work finish and rejects new work`); it looks unrelated because the other unit/appliance matrices passed, but the updated head should be rerun to green.

src/quic/tls_adapter.zig
```diff
@@ -51,6 +61,25 @@ pub const header_protection_sample_len = 16;
...
+fn unusedDefaultProviderEntropyFill(_: *anyopaque, buffer: []u8) crypto_provider.EntropyError!void {
+    _ = buffer;
+    return error.EntropyFailure;
+}
+
+const default_provider_entropy_context: u8 = 0;
+
+const default_provider_backing: crypto_pkg.pure_zig.Provider = crypto_pkg.pure_zig.Provider.init(.{
+    .context = @constCast(&default_provider_entropy_context),
+    .fillFn = unusedDefaultProviderEntropyFill,
```
Jaruso: [P1] This moves primitive calls behind the vtable, but it makes protocol code choose and construct the concrete `pure_zig.Provider`. #490 explicitly requires protocol code not to know whether crypto is implemented by `std.crypto`, project code, or a future backend; concrete provider selection belongs in a native HTTP/QUIC composition root.

Please inject the provider through an adapter/connection constructor or options object, and keep any pure-Zig convenience default outside `src/quic/` (or test-only). For example:

```zig
pub fn init(provider: crypto_provider.CryptoProvider) !QuicTlsAdapter {
    try validatePacketProtectionProvider(provider);
    return .{ .provider = provider };
}
```

Then the appliance/general composition root can construct the selected backend and pass only the erased `CryptoProvider` into QUIC.

src/quic/tls_adapter.zig
```diff
@@ -599,6 +756,15 @@ pub const QuicTlsAdapter = struct {
...
     metrics: Metrics = .{},
+    /// Provider-owned crypto for HKDF, AEAD, and QUIC header-protection
+    /// operations (#490). Defaults to the pure-Zig provider so the live send/
+    /// receive path is provider-backed without every call site threading one
+    /// through; override with `setProvider` to select a different backend.
+    provider: crypto_provider.CryptoProvider = default_provider_backing.cryptoProvider(),
+
+    pub fn setProvider(self: *QuicTlsAdapter, provider: crypto_provider.CryptoProvider) void {
+        self.provider = provider;
+    }
```
Jaruso, line 772: [P1] `setProvider` accepts an arbitrary provider without validating the capabilities required by the fixed QUIC profile. A provider missing SHA-256 HKDF, AES-128-GCM, or AES-128 header protection is stored successfully; later `protectionKeys` converts `ProviderUnsupported` to `null`, receive paths report it as peer authentication failure, and transmit header protection uses `catch unreachable`.

Validate before storing and preserve configuration errors as configuration errors:

```zig
pub fn setProvider(self: *QuicTlsAdapter, cp: CryptoProvider) error{ProviderUnsupported}!void {
    const caps = cp.capabilities();
    if (!caps.supportsHash(.sha256) or
        !caps.supportsAead(.aes_128_gcm) or
        !caps.supportsQuicHeaderProtection(.aes_128))
        return error.ProviderUnsupported;
    self.provider = cp;
}
```

`protectionKeys` should also return `error{ProviderUnsupported}!?PacketProtectionKeys` rather than swallowing that error, so callers never conflate a broken provider configuration with absent traffic secrets or forged packets.

src/crypto/profile.zig
```diff
@@ -122,14 +179,25 @@ fn row(
...
+    inline for (items, 0..) |item, i| list[i] = .{ .consumer = item[0], .status = item[1] };
+    return list;
+}
+
+/// Mechanical derivation, not a separate judgment call: the product profile
+/// that selects a backend is exactly the profile "enabled" once that
+/// backend reports `.supported` for this row.
+fn productProfilesFor(pure_zig_status: Status, openssl_status: Status) ProductProfileSet {
+    var set = ProductProfileSet{};
+    if (pure_zig_status == .supported) set.insert(.native_appliance);
```
Jaruso: [P1] This still collapses provider capability into product enablement. Every pure-Zig primitive marked `.supported` is automatically labeled enabled in `.native_appliance`, even when the same row says there is no live integration. For example ChaCha20-Poly1305 has no consumers and AES-256-GCM is not negotiated, yet both receive `.native_appliance` here. Conversely, `general_purpose_openssl` is modeled from the deferred in-process OpenSSL-provider column rather than what the existing general-purpose TLS product can actually select.

That is exactly the distinction #490 asks the profile to preserve. Make product enablement explicit per row/profile (or derive it from named, live selectable integrations), e.g. pass an `enabled_profiles` set to `row(...)`. Primitive support should remain independent and must not automatically imply product-profile availability.

scripts/audit-crypto-boundary.sh
```diff
@@ -0,0 +1,56 @@
...
+    "QUIC packet parsing/encoding is public protocol logic. The existing AES-GCM Retry integrity test vector is allowed because RFC 9001 fixes the public key/nonce and it is not packet-protection key material."
+
+check_file "src/quic/path.zig" \
+    "std\\.crypto\\.(dh|sign|kdf)|crypto\\.(dh|sign|kdf)|X25519\\.|Ed25519\\.|Ecdsa|Rsa|hkdfExpandLabel\\(|$aes_block_pattern" \
+    "QUIC path validation may use public constants and existing token/retry exceptions, but must not add key exchange, signatures, or KDF shortcuts."
+
+if rg -n "std\\.crypto\\.(aead|dh|sign|kdf)|crypto\\.(aead|dh|sign|kdf)|Aes[0-9]+Gcm\\.(encrypt|decrypt)|ChaCha20Poly1305\\.(encrypt|decrypt)|X25519\\.|Ed25519\\.|Ecdsa|Rsa|hkdfExpandLabel\\(|$aes_block_pattern" src/quic \
+    --glob '*.zig' --glob '!tls_adapter.zig' --glob '!tls_handshake.zig' --glob '!path.zig' --glob '!connection.zig' --glob '!packet.zig' >/tmp/tardigrade-crypto-boundary.$$ 2>/dev/null; then
+    printf '%s\n' "disallowed direct keyed crypto in QUIC protocol modules outside the allowlist"
+    cat /tmp/tardigrade-crypto-boundary.$$
```
Jaruso: [P1] The guard still cannot catch the exact ownership regression this PR fixes. Reverting a live call to `keys.applyHeaderProtection(...)`, `keys.removeHeaderProtection(...)`, `keys.openPayload(...)`, or `deriveAes128GcmKeys(...)` contains none of the primitive regexes, and `tls_adapter.zig`—where all concrete wrappers live—is excluded wholesale. The profile test only checks the hard-coded metadata, so it would also continue passing.

Please make the allowlist narrow enough to enforce call ownership. The cleanest option is to move the concrete differential/reference implementations into test code and scan all of `src/quic/tls_adapter.zig`. Otherwise explicitly reject legacy wrapper call names in runtime modules and add a fixture that reintroduces each bypass and asserts the audit fails.

scripts/audit-crypto-boundary.sh
```diff
@@ -0,0 +1,56 @@
+#!/usr/bin/env sh
+set -eu
+
+ROOT="${1:-.}"
+
+cd "$ROOT"
+
+if ! command -v rg >/dev/null 2>&1; then
```
Jaruso: [P1] This correctly fails closed now, but it is not CI-safe: the current `9991770` workflow is failing in the normal Ubuntu, ARM, and appliance test jobs because those runners do not have `rg` installed (`audit-crypto-boundary: ripgrep (rg) is required but not found in PATH`).

Please either make the guard self-contained—preferably a small checked-in Zig audit invoked by the build—or install/pin ripgrep in every job and supported platform that runs `zig build test` / `test-crypto`. Until then the PR's required CI matrix cannot pass.

Jaruso (commented):
Second-pass review of `9991770`.

The previous runtime-wiring issue is substantially addressed: Initial/key-update derivation, payload AEAD, and header protection now cross `CryptoProvider` on the live QUIC path. The fail-open behavior and missing AES primitive pattern were also addressed.

There are still five blocking issues before merge:

1. `tls_adapter.zig` now selects and constructs the concrete pure-Zig backend inside protocol code, which violates #490's provider-neutral composition rule.
2. Injected providers are not capability-validated; unsupported providers can be downgraded to missing keys, misreported as peer authentication failures, or hit `unreachable` on transmit.
3. `enabled_product_profiles` is mechanically copied from provider primitive status, so it still does not represent product-level selectability.
4. The architecture guard can still miss the exact regression this PR fixes because legacy concrete wrapper calls are not forbidden and `tls_adapter.zig` is excluded wholesale.
5. CI is currently red because the guard now requires `rg`, but the existing unit-test/appliance jobs do not install it.

I left concrete fixes inline. This remains a useful QUIC-first slice of #490, but the parent issue should stay open for the documented TLS key-schedule/key-share/CertificateVerify follow-up.

src/quic/tls_adapter.zig
```diff
@@ -51,6 +65,30 @@ pub const header_protection_sample_len = 16;
...
+/// composition root, have a ready capability-valid provider instead of each
+/// duplicating the construction boilerplate.
+fn testOnlyEntropyFill(_: *anyopaque, buffer: []u8) crypto_provider.EntropyError!void {
+    _ = buffer;
+    return error.EntropyFailure;
+}
+
+const test_only_entropy_context: u8 = 0;
+
+const test_only_provider_backing: crypto_pkg.pure_zig.Provider = crypto_pkg.pure_zig.Provider.init(.{
```
Jaruso: [P1] `src/quic/` still selects and constructs the concrete pure-Zig backend here, despite the module comment and commit message saying it never names one. `testOnlyDefaultProvider` is public and is not compile-time gated; it is also used by `tests/h3_interop_tool.zig`'s executable client/server paths, not solely inside Zig `test` blocks. That leaves the protocol package aware of `crypto.pure_zig.Provider` and provides a production-reachable concrete-provider escape hatch.

Move this construction into a dedicated test-support/composition-root module and inject only the erased `CryptoProvider` into `src/quic/`. For example, a `tests/support/quic_crypto.zig` module can own deterministic entropy plus `pure_zig.Provider`, while the interop tool can construct its own OS-backed provider as its composition root. Then add an audit rule that rejects `pure_zig` references anywhere under `src/quic/`.

scripts/audit_crypto_boundary.zig
```diff
@@ -0,0 +1,389 @@
...
+const dir_checks = [_]DirCheck{
+    .{
+        .dir = "src/quic",
+        // tls_adapter.zig is the approved provider-owned adapter (and the
+        // one file allowed to define the legacy wrapper names as
+        // differential fixtures); tls_handshake.zig's in-memory handshake
+        // test backend uses deterministic transcript/HKDF fixtures and is
+        // excluded from the protocol guard; path.zig, connection.zig, and
+        // packet.zig have their own narrower checks above.
+        .excluded = &.{ "tls_adapter.zig", "tls_handshake.zig", "path.zig", "connection.zig", "packet.zig", "cid.zig" },
```
Jaruso: [P1] The guard still excludes the exact canonical implementation whose ownership it needs to enforce. A regression such as changing `QuicTlsAdapter.sealPacketPayload` back to `keys.sealPayload(...)`, or `protectionKeys` back to `deriveAes128GcmKeys(...)`, happens inside `tls_adapter.zig` and passes because the whole file is excluded. Likewise, excluding all of `tls_handshake.zig` exempts the production handshake driver merely because the test backend shares that file.

The robust fix is to move the concrete differential implementations and `TestTlsBackend` into test-support files, leaving the production adapter/driver fully scannable. Alternatively, add dedicated checks for the canonical method bodies—not just callers in `connection.zig`—and an end-to-end fixture that rewrites `sealPacketPayload`/`protectionKeys` to the legacy forms and proves the audit fails.

scripts/audit_crypto_boundary.zig
```diff
@@ -0,0 +1,389 @@
...
+
+const Violation = struct {
+    path: []const u8,
+    needle: []const u8,
+    rationale: []const u8,
+};
+
+fn checkFile(allocator: std.mem.Allocator, root: compat.DirCompat, path: []const u8, forbidden: []const []const u8, rationale: []const u8, violations: *std.ArrayList(Violation)) !void {
+    const contents = root.readFileAlloc(allocator, path, 16 * 1024 * 1024) catch |err| switch (err) {
+        error.FileNotFound => return,
```
Jaruso: [P2] Explicit checks fail open when a protected file disappears: renaming or moving `src/http/http3_runtime.zig` (or any other listed path) silently removes that enforcement. The directory scan is also intentionally non-recursive, so a new `src/quic/subdir/packet_crypto.zig` bypasses it entirely.

For an architecture guard, absence of an expected path should be an audit error, and the QUIC tree should be walked recursively. Please add fixtures for both cases: delete an explicitly protected file and place a direct AEAD call in a nested QUIC directory; each should produce a violation.

src/crypto/profile.zig
```diff
@@ -78,29 +120,44 @@ pub const Row = struct {
...
+    row(.{ .hkdf = .sha256 }, "HKDF-SHA256", .zig_std_crypto, .supported, .openssl_provider, .provider_deferred, "RFC 5869/std.crypto and TLS expand-label parity", .{ .{ .tls_handshake, .not_integrated }, .{ .quic_tls_bridge, .live }, .{ .quic_packet_protection, .live } }, .{ .native_appliance, .general_purpose_openssl }, "QUIC Initial/Handshake/1-RTT key derivation in src/quic/tls_adapter.zig runs through CryptoProvider; the TLS 1.3 key schedule (src/tls/key_schedule.zig) still calls std.crypto directly and is tracked as open follow-up for #490."),
+    row(.{ .hkdf = .sha384 }, "HKDF-SHA384", .zig_std_crypto, .supported, .openssl_provider, .provider_deferred, "provider capability and expand-label tests", .{.{ .tls_handshake, .not_integrated }}, .{ .native_appliance, .general_purpose_openssl }, "src/tls/key_schedule.zig calls std.crypto directly; HKDF additions require TLS 1.3 label coverage."),
```
Jaruso: [P1] Product enablement is explicit now, but several values still describe primitive availability rather than the live native product. SHA-384/HKDF-SHA384 are marked `.native_appliance`, and RSA-PSS is also marked native below, while `tls13_backend.native_capabilities` exposes only `TLS_AES_128_GCM_SHA256` and only Ed25519/ECDSA signatures. The native engine cannot select those rows end to end.

There is also no product-aware consumer of this data: `tls.crypto_profile.fromProvider(profile.capabilities(.pure_zig))` still returns all three cipher suites, and the documentation recommends that function for configuration. Introduce a product-aware derivation such as `fromProfile(.native_appliance, caps)` (or filter by `enabled_product_profiles` plus selectable integrations), update the native test to expect AES-128/SHA-256 only, and remove `.native_appliance` from rows the native engine cannot actually select. Otherwise the profile still claims support the live consumer cannot execute, contrary to #490.

Jaruso (commented):
Third-pass review of `19acc84`. The five findings from the prior round are substantially addressed: provider injection is required and capability-checked, product profiles are explicit, the runtime path stays provider-backed, the guard is now self-contained, and CI run 1096 is green.

There are still six blocking issues in the new implementation. The most important are that concrete provider selection remains inside `src/quic/` through a public test helper, and the audit still excludes the canonical adapter methods whose regression it is meant to prevent. The product-profile data also remains inconsistent with the actual native TLS engine and is not used by the recommended TLS capability derivation path.

These comments review this PR as an incremental QUIC/provider slice. They do not imply that the full #490 story is complete; the documented TLS key-schedule/key-exchange/CertificateVerify follow-ups should remain open.

src/http/http3_runtime.zig
```diff
@@ -300,7 +315,9 @@ pub const Runtime = struct {
             .stopping = std.atomic.Value(bool).init(false),
             .drain_requested = std.atomic.Value(bool).init(false),
             .drain_deadline_us = std.atomic.Value(u64).init(0),
+            .crypto_provider_state = undefined,
         };
+        runtime.crypto_provider_state = tls_core.production_crypto.Provider.init(runtime.crypto_provider_entropy.entropy());
```
Jaruso, line 321: [P2] This stores a self-referential pointer into a value that is immediately returned by value. `OsEntropy.entropy()` sets `context = &runtime.crypto_provider_entropy`; after `Runtime.init` returns or the `Runtime` is moved, the copied provider retains the address of the local pre-return field. `OsEntropy.fill` currently ignores `context`, so this is dormant today, but `CryptoProvider.entropy` exposes the dangling pointer and any future stateful entropy implementation or `randomBytes` call would dereference invalid storage.

Make the entropy context move-stable—for example, use a static context for this stateless OS entropy callback, store the provider state in separately allocated stable storage, or initialize the self-reference only after the final `Runtime` has a stable address.

build.zig
```diff
@@ -324,6 +326,33 @@ pub fn build(b: *std.Build) void {
     test_step.dependOn(&run_crypto_tests.step);
     test_step.dependOn(&run_crypto_secret_tests.step);
 
+    // Deterministic, dependency-free crypto-boundary audit (#490): a checked-in
+    // Zig program rather than a shell script shelling out to an ambient `rg`,
+    // so every CI runner and platform enforces it identically with nothing
+    // extra to install.
+    const crypto_boundary_audit_mod = b.createModule(.{
+        .root_source_file = b.path("scripts/audit_crypto_boundary.zig"),
+        .target = target,
```
Jaruso: [P2] This audit is a build-host tool, but it is compiled for the user-selected target and then executed with `addRunArtifact`. A cross invocation such as `zig build audit-crypto-boundary -Dtarget=aarch64-linux` on an x86 host will build a foreign executable and attempt to run it. The imported `compat_mod` is target-specific as well.

Build the audit executable and its tests for the host target (`b.graph.host`, or the Zig 0.16 equivalent) and give it a host-target `zig_compat` module. The application/test artifacts can continue using `target`; source-audit tooling that executes during the build must use the host.

src/quic/connection.zig
```diff
@@ -3619,6 +3637,8 @@ const TestPair = struct {
             .initial_secret_dcid = &odcid,
             .peer_cid = &odcid,
             .tls = pair.client_backend.backend(),
+            var test_crypto_storage_17 = tls_core.production_crypto.StackProviderStorage{};
```
Jaruso: [P0] This is not valid Zig: declarations cannot appear inside the `Connection.Options` struct literal, which is one of the parser failures currently breaking CI. Do not fix this by merely hoisting `test_crypto_storage_17` to a local variable above the call: `pair.client` escapes `TestPair.init`, while `CryptoProvider.context` borrows the storage's `provider` field, so that would leave the connection holding a pointer into a dead stack frame.

Make the storage part of the stable heap-allocated owner instead:

```zig
const TestPair = struct {
    client_crypto: tls_core.production_crypto.StackProviderStorage = .{},
    server_crypto: tls_core.production_crypto.StackProviderStorage = .{},
    // existing fields...
};

pair.* = .{
    .client_backend = ...,
    .server_backend = ...,
};
pair.client = try Connection.init(allocator, .{
    // ...
    .crypto_provider = tls_core.production_crypto.stackProvider(&pair.client_crypto),
});
pair.server = try Connection.init(allocator, .{
    // ...
    .crypto_provider = tls_core.production_crypto.stackProvider(&pair.server_crypto),
});
```

Apply the same lifetime rule to every constructor that returns or stores a connection beyond the current function.

src/http/http3_runtime.zig
```diff
@@ -1597,6 +1615,8 @@ const RuntimeCidHarness = struct {
             .initial_secret_dcid = &odcid,
             .peer_cid = &odcid,
             .tls = self.client_backend.backend(),
+            var test_crypto_storage_51 = tls_core.production_crypto.StackProviderStorage{};
```
Jaruso: [P0] This generated declaration inside the options literal is the first format/parse failure reported by CI. There is also a second error hidden behind it: the field initializer uses `quic.tls_core.production_crypto`, but this module already imports `tls_core`; `quic.tls_core` is not the composition API.

For `RuntimeCidHarness`, the client/server connections survive this initialization method, so provider storage must be fields on the harness (or use a stable test-support provider), not locals. For a test where the connection remains entirely inside one function, a local declared *before* the literal is sufficient:

```zig
var crypto_storage: tls_core.production_crypto.StackProviderStorage = .{};
const conn = try Connection.init(allocator, .{
    // ...
    .crypto_provider = tls_core.production_crypto.stackProvider(&crypto_storage),
});
```

Please repair all generated `_51` through `_63` replacements in this file rather than only the first parser error.

src/quic/tls_adapter.zig
```diff
@@ -999,23 +1178,26 @@ test "adapter installs Initial secrets by endpoint perspective" {
     var dcid: [8]u8 = undefined;
     _ = try std.fmt.hexToBytes(&dcid, "8394c8f03e515708");
 
-    var client = QuicTlsAdapter{};
+    var test_crypto_storage_1 = tls_core.production_crypto.StackProviderStorage{};
+    var client = QuicTlsAdapter{ .provider = testOnlyDefaultProvider() };
```
Jaruso: [P0] `testOnlyDefaultProvider` was removed from this module, but the replacement stopped halfway: the new `test_crypto_storage_1` value is unused and this initializer still calls the deleted symbol. The same pattern repeats throughout this file, and unchanged callers remain in `tests/h3_interop_tool.zig`, `tests/quic_h3_e2e.zig`, `tests/quic_h3_smoke.zig`, and `tests/quic_h3_udp_smoke.zig`. Those will become the next compile failures after the parser errors are fixed.

Either complete each local replacement:

```zig
var crypto_storage: tls_core.production_crypto.StackProviderStorage = .{};
var client = QuicTlsAdapter{
    .provider = tls_core.production_crypto.stackProvider(&crypto_storage),
};
```

or, preferably for test/tool call sites, actually wire the new `tests/support/quic_crypto.zig` module into `build.zig` and use its stable test provider. At present `testDefaultProvider` has no callers and the new file has no build import, so it does not replace the removed API.

src/quic/tls_backend.zig
```diff
@@ -949,6 +952,10 @@ const RealHandshakeHarness = struct {
                 try Identity.initPkcs8(testdata.certificate_der, testdata.private_key_pkcs8_der),
             ),
         };
+        harness.client_adapter = .{ .provider = tls_core.production_crypto.stackProvider(&harness.client_crypto) };
+        harness.server_adapter = .{ .provider = tls_core.production_crypto.stackProvider(&harness.server_crypto) };
+        return harness;
+    }
     }
```
Jaruso, line 953: [P0] The extra closing brace here is the `expected ';' after declaration` parser failure reported by CI. There are two correctness issues immediately around it that also need repair:

- `init()` builds `CryptoProvider` views borrowing `harness.client_crypto` / `server_crypto` and then returns `harness` by value. The returned adapters still point at the pre-return local copy.
- `initWithSniProvider()` returns with both adapter fields still `undefined`, but its tests call `wire()` and use them.

Initialize borrowed views only after the harness has reached its final address, for example:

```zig
fn initBackends(...) !RealHandshakeHarness {
    return .{ .client_backend = ..., .server_backend = ... };
}

fn initCrypto(self: *RealHandshakeHarness) void {
    self.client_adapter = .{
        .provider = production_crypto.stackProvider(&self.client_crypto),
    };
    self.server_adapter = .{
        .provider = production_crypto.stackProvider(&self.server_crypto),
    };
}

var harness = try RealHandshakeHarness.initBackends(...);
harness.initCrypto();
```

Both constructors must follow the same initialization path.

src/quic/tls_handshake.zig
```diff
@@ -738,13 +738,22 @@ fn defaultParams() config.TransportParameters {
...
+    server_crypto: tls_core.production_crypto.StackProviderStorage = .{},
+    client_adapter: QuicTlsAdapter = undefined,
+    server_adapter: QuicTlsAdapter = undefined,
     client_backend: TestTlsBackend = .{},
     server_backend: TestTlsBackend = .{},
     client: Handshake = undefined,
     server: Handshake = undefined,
 
+    fn init() Harness {
+        var harness: Harness = .{};
```
Jaruso: [P1] This constructor has the same borrowed-self move problem: `stackProvider` stores a context pointer to fields of the local `harness`, then `return harness` copies/moves the struct. In addition, the tests below still instantiate `Harness{}` directly instead of calling this function, leaving `client_adapter` and `server_adapter` undefined.

Use in-place initialization after the caller owns the final variable:

```zig
fn init(self: *Harness) void {
    self.* = .{};
    self.client_adapter = .{
        .provider = production_crypto.stackProvider(&self.client_crypto),
    };
    self.server_adapter = .{
        .provider = production_crypto.stackProvider(&self.server_crypto),
    };
}

var h: Harness = undefined;
h.init();
```

A stable provider from the new test-support module would also avoid this per-harness self-reference entirely.

src/tls/crypto_profile.zig
```diff
@@ -46,15 +46,38 @@ pub const TlsCapabilities = struct {
 };
 
 pub fn fromProvider(caps: provider.Capabilities) TlsCapabilities {
+    return fromProfile(.general_purpose_openssl, caps);
+}
```
Jaruso: [P1] Adding `fromProfile` is the right direction, but keeping `fromProvider` as a silent alias for `.general_purpose_openssl` preserves the original product-confusion bug. `docs/CRYPTO_PROVIDER.md` still instructs every protocol configuration to call `fromProvider(provider.capabilities())`; with pure-Zig capabilities this still returns AES-256-GCM, ChaCha20-Poly1305, and RSA-PSS even when the caller is the native appliance.

Please remove the ambiguous wrapper or require the product explicitly at every call site:

```zig
pub fn fromProvider(
    product: profile.ProductProfile,
    caps: provider.Capabilities,
) TlsCapabilities {
    return fromProfile(product, caps);
}
```

Also gate each cipher suite on all product dimensions it requires, not only its AEAD row. For AES-256, for example, require the profile to enable `.aead = .aes_256_gcm`, `.hash = .sha384`, and `.hkdf = .sha384` before advertising `TLS_AES_256_GCM_SHA384`. Update the documentation to recommend the product-aware API.

Jaruso (commented):
Blocking follow-up review on `cc6d241` (`cursor WIP`). This is a partial fix and the branch is not currently mergeable.

What improved:
- The production `Runtime` entropy callback now uses a static context, so the previous return-by-value self-reference is removed.
- `src/quic/tls_adapter.zig` no longer constructs or names `crypto.pure_zig.Provider` in its production setup.
- The profile removes `.native_appliance` from SHA-384, HKDF-SHA384, and RSA-PSS, and adds a product-aware `fromProfile` path with a native-profile test.

Current blockers:
1. The mechanical provider replacement inserted declarations inside struct literals in `http3_runtime.zig` and `connection.zig`, plus an extra brace in `tls_backend.zig`. CI run 1097 fails format and every build/test job at parsing.
2. Removing `testOnlyDefaultProvider` was incomplete: `tls_adapter.zig` tests and unchanged interop/e2e/smoke call sites still reference it, while `tests/support/quic_crypto.zig` is not wired into the build or referenced anywhere.
3. Several proposed `StackProviderStorage` uses have invalid lifetimes. Moving the generated declarations above the literals is not sufficient where a `Connection` or harness escapes the function; the provider borrows the storage and that storage must live in the owning heap object or other stable location.
4. The two returned-by-value handshake harness constructors create provider views into their own local fields before returning, so those views point at the pre-return copy. `initWithSniProvider` additionally leaves both adapters undefined.
5. Product-aware filtering is only partially adopted: `fromProvider` silently assumes `.general_purpose_openssl`, while the checked-in documentation still tells all protocol configuration—including pure-Zig/native configuration—to call it. This can still advertise AES-256, ChaCha20, and RSA-PSS for the native provider.
6. The previous guard findings remain untouched: `tls_adapter.zig` / `tls_handshake.zig` are still excluded from the ownership scan, missing protected files still fail open and nested QUIC directories are not scanned, and the audit executable is still built for the application target rather than the build host.

The live QUIC provider migration itself remains directionally correct, but this commit needs to be repaired to compile and the remaining enforcement issues need to be completed before merge. As before, GitHub does not allow the PR author's connected account to formally request changes on its own PR, so this is submitted as a blocking comment review.

scripts/audit_crypto_boundary.zig
```diff
@@ -0,0 +1,627 @@
...
+/// itself stays excluded from the whole-file/whole-directory scans above
+/// because it legitimately defines the legacy concrete primitives as
+/// differential test-vector fixtures (and its own tests call them to compare
+/// against the provider path) — a blanket scan of the file cannot tell that
+/// apart from a live call site regressing. This targets exactly the
+/// regression the third-pass review named: reverting `sealPacketPayload`
+/// back to `keys.sealPayload(...)`, or `protectionKeys` back to
+/// `deriveAes128GcmKeys(...)`.
+const function_body_checks = [_]FunctionBodyCheck{
+    .{ .path = "src/quic/tls_adapter.zig", .function_name = "protectionKeys", .forbidden = &legacy_wrapper_calls, .rationale = "QuicTlsAdapter.protectionKeys must derive packet-protection keys through deriveAes128GcmKeysWithProvider, not the legacy deriveAes128GcmKeys." },
```
Jaruso: [P1] These checks enforce the spelling of the six canonical callers, but not the provider-backed implementation they call. Their forbidden set contains only `legacy_wrapper_calls`, and there are no body checks for `sealPayloadWithProvider`, `openPayloadWithProvider`, `headerProtectionMaskWithProvider`, `deriveInitialSecretsV1WithProvider`, `deriveAes128GcmKeysWithProvider`, or `deriveNextGenerationSecretWithProvider`.

For example, this regression still passes the audit:

```zig
pub fn sealPayloadWithProvider(...) ![]u8 {
    return self.sealPayload(packet_number, header, plaintext, out);
}
```

`QuicTlsAdapter.sealPacketPayload` continues to spell `sealPayloadWithProvider`, so every current function check stays clean while the live path has returned to concrete AES-GCM. Direct `Aes128Gcm.encrypt` inside the canonical method would also pass because these checks do not include `keyed_crypto_core` or `aes_block_cipher`.

Please either move all legacy/reference implementations into test-only source and scan `tls_adapter.zig` normally, or add body checks for every provider entry point and canonical method using the full concrete-primitive plus legacy-call set. Add an end-to-end fixture where a `*WithProvider` function delegates to its concrete sibling and assert the audit rejects it.

scripts/audit_crypto_boundary.zig
```diff
@@ -0,0 +1,627 @@
...
+
+/// Same as `keyed_crypto_core` except for KDF: `tls_handshake.zig`'s
+/// in-memory `TestTlsBackend` test fixture builds a deterministic transcript
+/// hash with `std.crypto.kdf.hkdf.HkdfSha256` directly, since it stands in
+/// for a TLS engine rather than driving one. The production `Handshake`/
+/// `CoreDriver` code sharing this file has no keyed-crypto dependency at
+/// all, so this file is fully scannable for every other category, including
+/// the legacy wrapper names (#490 third-pass review: this file must not stay
+/// wholesale-excluded from the guard just because a test fixture shares it).
+const tls_handshake_zig_forbidden = [_][]const u8{
```
Jaruso: [P1] The intended exception is one deterministic HKDF use inside `TestTlsBackend`, but this pattern set exempts **all KDF use anywhere in `tls_handshake.zig`**. A future production change such as this inside `Handshake`, `CoreDriver`, or another runtime helper passes:

```zig
const next = std.crypto.kdf.hkdf.HkdfSha256.extract(salt, secret);
```

That is a secret-bearing KDF in protocol code and violates #490's provider ownership rule. Please scope the exception to the exact test-backend function/body, or move `TestTlsBackend` into test support and apply the full `keyed_crypto_core` set to the production driver. The fixture suite should prove that KDF is allowed only in the named fixture and rejected elsewhere in the same file.

scripts/audit_crypto_boundary.zig
```diff
@@ -0,0 +1,627 @@
...
+        // QUIC path validation keeps its existing address-validation
+        // token/Retry-integrity exception (public constants, process keys);
+        // see docs/CRYPTO_PROVIDER_AUDIT.md. No key exchange, signatures, KDF,
+        // or new AES block-cipher shortcuts may be added here.
+        .forbidden = &path_zig_forbidden,
+        .rationale = "QUIC path validation may use public constants and the existing token/retry exception, but must not add key exchange, signatures, KDF, or AES block-cipher shortcuts.",
+    },
+    .{
+        .path = "src/http/http3_runtime.zig",
+        .forbidden = &legacy_wrapper_calls,
```
Jaruso: [P1] The native composition root is checked only for retired wrapper names, so it can introduce direct concrete keyed crypto without tripping the audit. For example, `std.crypto.aead.*`, `std.crypto.kdf.*`, `X25519.*`, or `Aes128.initEnc(...)` in `http3_runtime.zig` all pass today.

The composition root may choose/construct a provider, but it should not perform packet crypto itself. Use the full concrete-primitive set here, with any legitimate provider-construction or OS-entropy exceptions expressed narrowly. For example:

```zig
const http3_runtime_forbidden =
    keyed_crypto_core ++ aes_block_cipher ++ legacy_wrapper_calls;
```

Then add a fixture that places a direct AEAD or AES-header-protection call in `src/http/http3_runtime.zig` and expects a violation.

scripts/audit_crypto_boundary.zig
```diff
@@ -0,0 +1,627 @@
...
+        // tls_adapter.zig is the approved provider-owned adapter and the
+        // one file allowed to define the legacy wrapper names as
+        // differential fixtures, so it cannot be scanned wholesale for
+        // either keyed-crypto primitives or the legacy names themselves —
+        // `function_body_checks` above targets its canonical methods
+        // specifically instead. path.zig, connection.zig, packet.zig,
+        // cid.zig, and tls_handshake.zig have their own narrower checks in
+        // `file_checks` above. Recursive: a new nested directory under
+        // src/quic gets no free pass.
+        .excluded = &.{ "tls_adapter.zig", "path.zig", "connection.zig", "packet.zig", "cid.zig", "tls_handshake.zig" },
```
Jaruso: [P1] The recursive scan still has a basename escape hatch. `checkDir` compares every `entry.name` against this list at every depth, so files such as `src/quic/nested/connection.zig`, `src/quic/experimental/tls_adapter.zig`, or `src/quic/foo/path.zig` are skipped entirely. That contradicts the adjacent claim that nested directories get no free pass.

Make exclusions exact root-relative paths rather than basenames. One simple shape is:

```zig
.excluded_paths = &.{
    "src/quic/tls_adapter.zig",
    "src/quic/path.zig",
    // ...
},
```

Build `rel` before checking exclusions and compare the complete normalized path. Add the exact fixture `src/quic/nested/connection.zig` containing `std.crypto.aead...`; the current `nested/packet_crypto.zig` fixture does not exercise this bypass.

scripts/audit_crypto_boundary.zig
```diff
@@ -0,0 +1,627 @@
...
+    ".openPayload(",
+    ".applyHeaderProtection(",
+    ".removeHeaderProtection(",
+    ".headerProtectionMask(",
+    "deriveAes128GcmKeys(",
+    "deriveInitialSecretsV1(",
+    "deriveNextGenerationSecret(",
+};
+
+const packet_zig_forbidden = [_][]const u8{
```
Jaruso: [P1] The approved exceptions are still file-wide rather than narrowly allowlisted. Because `packet_zig_forbidden` omits AEAD, any unrelated new `Aes128Gcm.encrypt` in `packet.zig` passes alongside the fixed-public Retry vector. `path_zig_forbidden` similarly permits arbitrary new AEAD/auth operations anywhere in `path.zig`, and the CID check permits arbitrary HMAC/auth use anywhere in `cid.zig`.

#490 explicitly requires exceptions to be narrow and rationale-backed. Please either move the Retry/vector code into test or crypto-owned support, or apply full-file concrete checks and carve out only the exact approved function bodies. Add fixtures proving that the existing named exception passes while a second unrelated AEAD/HMAC function in the same file fails.

build.zig
```diff
@@ -313,7 +313,24 @@ pub fn build(b: *std.Build) void {
...
+    // composition root never do. Every module whose test blocks or tools
+    // exercise the QUIC seam directly (without the real Runtime composition
+    // root) imports this; production code must never call it.
+    const test_quic_crypto_mod = b.createModule(.{
+        .root_source_file = b.path("tests/support/quic_crypto.zig"),
+        .target = target,
+        .optimize = optimize,
+    });
+    test_quic_crypto_mod.addImport("crypto", crypto_mod);
+    quic_mod.addImport("test_quic_crypto", test_quic_crypto_mod);
```
Jaruso: [P2] `test_quic_crypto` is described as unavailable to production, but it is injected unconditionally into both `quic_mod` and `exe_mod`, and production source files import it at top level. Nothing prevents a live protocol method from calling `testDefaultProvider()` and silently selecting the concrete pure-Zig singleton inside `src/quic/`; the current audit also does not reject such a call.

Please make the boundary enforceable rather than comment-only. Prefer separate test module roots/import graphs so production `quic_mod` cannot resolve `test_quic_crypto` at all. If Zig's module-sharing constraints require the import to exist, add an explicit compile-time test-only facade plus an audit rule that rejects `test_quic_crypto` references outside the known test sections/files. The interop executable is itself a composition root and can construct/inject its own appropriately backed provider without making the helper available throughout production QUIC.

docs/CRYPTO_PROVIDER.md
```diff
@@ -60,43 +65,96 @@ changes when it lands.
...
-| RSA-PSS-RSAE-SHA256 | supported | project verifier | provider deferred | CertificateVerify, PKI |
-| DER parser, chain builder, WebPKI validation | provider deferred | project code / unavailable | provider deferred | PKI |
-| injected random bytes, secure zero, constant-time compare | supported | project code | provider deferred / project code | all secret-bearing paths |
+The table below summarizes the checked-in profile for review; "Integration"
+lists only the consumers whose live runtime calls `CryptoProvider` today
+(`.live`), not every consumer the row names:
+
+| Capability | Pure-Zig status | Pure-Zig implementation | OpenSSL status | Consumers | Live integration | Product profiles |
+| --- | --- | --- | --- | --- | --- | --- |
+| SHA-256, SHA-384 | supported | `std.crypto` | provider deferred | TLS transcript, HKDF, QUIC TLS bridge | none — unkeyed hashing has no `CryptoProvider` entry point by design | native appliance, general-purpose OpenSSL |
```
Jaruso: [P2] This summary table no longer matches the typed source of truth. The combined row says both SHA-256 and SHA-384 are enabled for the native appliance, and the next row says both HKDF-SHA256 and HKDF-SHA384 are native-enabled. `src/crypto/profile.zig` now correctly enables SHA-384/HKDF-SHA384 only for `.general_purpose_openssl`. The RSA-PSS row below likewise still says native appliance even though its typed row is general-purpose only.

Please split the combined rows or qualify product enablement per algorithm, and update RSA-PSS. For example:

```text
SHA-256       ... native appliance, general-purpose OpenSSL
SHA-384       ... general-purpose OpenSSL only
HKDF-SHA256   ... native appliance, general-purpose OpenSSL
HKDF-SHA384   ... general-purpose OpenSSL only
RSA-PSS       ... general-purpose OpenSSL only
```

The acceptance criteria require `docs/CRYPTO_PROVIDER.md` to match the implemented profile, so the prose table cannot contradict the checked-in rows even if it calls them the source of truth.

Jaruso (commented):
Re-review of `00ff6d8` after the Cursor-WIP repair.

This is a major improvement over the prior head: the generated syntax breakage is gone, provider lifetime/validation is sound on the live `Connection.init` path, product-aware capability derivation is explicit, the audit tool runs for the build host, and CI run 1098 is fully green.

I still found seven blocking enforcement/documentation gaps, detailed inline. The packet runtime itself now looks correctly provider-routed; the remaining concern is that the architecture guard can still miss several concrete regressions it claims to prevent, and the checked-in summary table no longer matches the typed profile. The cleanest resolution is to move the legacy concrete QUIC implementations and test backend into test-support modules, then scan production source without broad exceptions. If they remain in production files, the audit needs function/path-scoped exceptions for every approved use and fixtures for each bypass below.

This PR remains an incremental slice of #490 (`Refs #490`), which is appropriate because the issue's full TLS key-schedule/key-share/CertificateVerify migration is still documented as follow-up. I am submitting a COMMENT review because GitHub does not permit requesting changes on my own PR.

scripts/audit_crypto_boundary.zig
```diff
@@ -0,0 +1,856 @@
...
+
+/// Concrete legacy wrapper names retired from the live QUIC send/receive path
+/// in favor of the `*WithProvider` entry points on the same types. Kept only
+/// as differential test-vector fixtures in `src/quic/tls_adapter.zig` and
+/// `tests/crypto_vectors.zig` / `tests/crypto_openssl_diff.zig`; a runtime
+/// module calling one directly is exactly the regression this audit exists
+/// to catch. Each literal ends at the call parenthesis so it does not also
+/// match its own `...WithProvider(` sibling.
+const legacy_wrapper_calls = [_][]const u8{
+    ".sealPayload(",
```
Jaruso: [P1] The provider-entry-point regression is still catchable only when the legacy method is invoked with dot syntax. Inside the same Zig container, a sibling function can be called unqualified, so this live-path regression contains neither `.sealPayload(` nor a concrete primitive and passes after the legacy function body is blanked:

```zig
pub fn sealPayloadWithProvider(
    self: *const PacketProtectionKeys,
    provider: CryptoProvider,
    packet_number: u64,
    header: []const u8,
    plaintext: []const u8,
    out: []u8,
) ![]u8 {
    _ = provider;
    return sealPayload(self, packet_number, header, plaintext, out);
}
```

The same bypass exists for `openPayload`, `headerProtectionMask`, `applyHeaderProtection`, and `removeHeaderProtection`. Add the unqualified call forms to the forbidden set (the exempt legacy declarations/bodies are already redacted, so `sealPayload(` will not reject their definitions), or perform a more structural call check. Please add the exact unqualified-delegation fixture; the current fixture uses `self.sealPayload(...)` and therefore does not cover this form.

scripts/audit_crypto_boundary.zig
```diff
@@ -0,0 +1,856 @@
...
+        // integrity tag use AES-GCM with process keys or RFC-fixed public
+        // constants; sealTokenPlaintextForTest is the one test-only helper
+        // building a token payload the same way for fuzz coverage.
+        .exempt_functions = &.{ "issueRetry", "validateRetry", "retryIntegrityTag", "sealTokenPlaintextForTest" },
+        .rationale = "QUIC path validation may use public constants and the existing token/Retry-integrity exception, but must not add key exchange, signatures, KDF, or AES block-cipher shortcuts anywhere else in this file.",
+    },
+    .{
+        .path = "src/quic/cid.zig",
+        .forbidden = &full_forbidden,
+        .exempt_exact = &.{"const HmacSha256 = std.crypto.auth.hmac.sha2.HmacSha256;"},
```
Jaruso, line 221: [P1] Blanking the import alias does not scope the exception to the approved operation. Once this exact declaration is removed from the scan, `full_forbidden` has no pattern for `HmacSha256.` itself, so any additional HMAC use anywhere in `cid.zig` passes:

```zig
fn unrelatedToken(key: [32]u8, input: []const u8) [32]u8 {
    var mac: [HmacSha256.mac_length]u8 = undefined;
    HmacSha256.create(&mac, input, &key);
    return mac;
}
```

The `HkdfSha256` exemption immediately below has the same problem: a production `Handshake` helper can call `HkdfSha256.extract(...)` without containing `std.crypto.kdf.`. This means the earlier file-wide HMAC/KDF exceptions have effectively been replaced by alias-wide exceptions rather than function-scoped ones.

Add alias-qualified patterns such as `HmacSha256.` and `HkdfSha256.` to the relevant forbidden sets, then allow only `statelessResetToken` and `TestTlsBackend.deriveSecret` (preferably by moving the test backend out of the production driver file). Add fixtures using the already-declared aliases—not fresh `std.crypto...` expressions—because the current fixtures only prove that a second import spelling is rejected.

Jaruso (commented):
Sixth-pass review of `3e7b3ed`. This commit substantially addresses the prior round: `tls_adapter.zig`'s production/provider implementations are now scanned, `http3_runtime.zig` uses the full forbidden set, recursive exclusions are root-relative, the product-profile documentation is corrected, and the narrow exception model is much better than the previous category-wide omissions.

Five enforcement gaps still remain in the source guard, detailed inline below. The most direct live-path bypasses are an unqualified call to a legacy sibling inside a `*WithProvider` method and legacy-wrapper calls moved into any QUIC file covered only by the recursive directory scan. The HMAC/HKDF alias exceptions also still apply file-wide in practice, and the test-provider protection is based on a fixed file list plus a textual prefix rather than all production declarations reachable from `quic_mod`/`exe_mod`.

CI run 1099 is also red. The audit itself passed, but `Test appliance profile (ubuntu-latest)` failed in `quic_h3_udp_smoke.test.udp smoke: HTTP/3 runtime drain lets admitted work finish and rejects new work` at `tests/quic_h3_udp_smoke.zig:971` (`new_initial_seen` was false). This commit does not modify that runtime/test path, and the same matrix passed on the previous head, so it looks likely to be a timing-sensitive flake; nevertheless the PR currently needs a clean rerun before it can be considered green.

#490 remains an incremental/open story beyond this PR, as documented. I am leaving this as a COMMENT review because GitHub does not permit requesting changes on my own pull request.

scripts/audit_crypto_boundary.zig
```diff
@@ -0,0 +1,856 @@
...
+        },
+        // Everything from this marker to end of file is the differential
+        // test-vector suite, which legitimately calls every exempted
+        // primitive/legacy function directly to compare against the
+        // provider path.
+        .production_only_marker = "\nconst testing = std.testing;",
+        .rationale = "QuicTlsAdapter owns provider-backed QUIC packet protection; every canonical method and every *WithProvider entry point must reach CryptoProvider only, never a concrete primitive or its own legacy differential-fixture sibling.",
+    },
+};
+
```
Jaruso, line 272: [P1] The recursive QUIC scan still omits `legacy_wrapper_calls`. Only the explicitly listed files receive `full_forbidden`; every other existing or future QUIC module is checked for concrete primitives and AES block construction only. Moving the regression into a helper therefore passes:

```zig
// src/quic/nested/protection.zig
fn seal(keys: PacketProtectionKeys, ...) ![]u8 {
    return keys.sealPayload(packet_number, header, plaintext, out);
}
```

`deriveInitialSecretsV1(...)` and the other legacy wrappers are equally invisible here. That undermines the guard whenever packet-protection code is refactored out of `connection.zig` rather than changed in place.

Use `full_forbidden` for the recursive directory scan, with explicit exceptions only where required, and add a nested-file fixture containing `keys.sealPayload(...)` or `deriveAes128GcmKeys(...)`. The existing nested fixtures use direct `std.crypto.aead` calls, so they do not exercise this ownership bypass.

scripts/audit_crypto_boundary.zig
```diff
@@ -0,0 +1,856 @@
...
+/// them (#490 fourth-pass review). Enforced the same way as
+/// `production_only_marker` above: only usages *before* the file's test
+/// boundary are violations.
+const TestOnlyImportCheck = struct {
+    path: []const u8,
+    boundary_marker: []const u8,
+    rationale: []const u8,
+};
+
+const test_quic_crypto_import_checks = [_]TestOnlyImportCheck{
```
Jaruso: [P1] `test_quic_crypto` is available module-wide, but this enforcement is a closed list of five current source files. `build.zig` injects the helper into both the entire production `quic_mod` and the entire production `exe_mod`; a new `src/quic/provider_helper.zig`, or an existing executable-module file such as `src/main.zig`, can import/use `test_quic_crypto.testDefaultProvider()` without appearing in this array and without matching the crypto directory patterns.

The architecture boundary needs to cover the module graph, not only today's importers. Prefer separate production and test module roots so the production graphs cannot resolve this helper. If the shared-module constraint is retained, recursively reject `test_quic_crypto.` throughout every production source reachable from `quic_mod` and `exe_mod`, with explicit test-only exceptions, and fail closed when a protected path moves. A fixture in a newly created QUIC file—and one in an executable-module source file—should both fail.

scripts/audit_crypto_boundary.zig
```diff
@@ -0,0 +1,856 @@
...
+}
+
+fn checkTestOnlyImport(allocator: std.mem.Allocator, root: compat.DirCompat, check: TestOnlyImportCheck, violations: *std.ArrayList(Violation)) !void {
+    const contents = root.readFileAlloc(allocator, check.path, 16 * 1024 * 1024) catch |err| switch (err) {
+        error.FileNotFound => return,
+        else => return err,
+    };
+    defer allocator.free(contents);
+
+    const boundary = std.mem.indexOf(u8, contents, check.boundary_marker) orelse contents.len;
```
Jaruso: [P1] A textual marker does not create a test-only region in Zig. Ordinary functions and public declarations may legally appear after `const testing = std.testing;` or after the first `test` block, and they remain production-compilable if referenced. This bypasses both this check and `FileCheckWithExceptions.production_only_marker`:

```zig
const testing = std.testing;

test "fixture" {}

pub fn liveProvider() CryptoProvider {
    return test_quic_crypto.testDefaultProvider();
}
```

Likewise, a direct `Aes128Gcm.encrypt(...)` added after `tls_adapter.zig`'s marker is excluded from the full crypto scan. The current fixture explicitly treats a helper call appended after the marker as clean, so it codifies the bypass rather than proving the code is genuinely test-gated.

Move the tests/differential fixtures into separate source modules, or redact actual `test` declaration bodies while continuing to scan every ordinary declaration regardless of file order. Add a fixture with a normal `pub fn` after the marker and require it to fail.

scripts/audit_crypto_boundary.zig
```diff
@@ -0,0 +1,960 @@
...
+/// establishes a *presumption* that everything after it is test-only and
+/// unreachable from outside the file; only a `pub` declaration could
+/// actually break that presumption by being reachable from elsewhere, so
+/// this is the structural check that makes the presumption enforceable
+/// rather than merely assumed (#490 fifth-pass review).
+fn checkTestOnlyBoundary(allocator: std.mem.Allocator, path: []const u8, contents: []const u8, violations: *std.ArrayList(Violation)) !void {
+    const boundary = findTestBoundary(contents);
+
+    const production_region = contents[0..boundary];
+    if (std.mem.indexOf(u8, production_region, "test_quic_crypto.") != null) {
```
Jaruso: [P1] This checks the **identifier spelling** `test_quic_crypto`, not use of the imported module. A new production importer can simply bind the same module under another name and pass:

```zig
const test_crypto = @import("test_quic_crypto");

pub fn liveProvider() CryptoProvider {
    return test_crypto.testDefaultProvider();
}
```

`scanTestQuicCryptoImporters` finds the file because the import string is present, but `production_region` contains no `test_quic_crypto.` substring. The direct expression `@import("test_quic_crypto").testDefaultProvider()` evades the same check.

Do not key enforcement to a conventional local variable name. Either parse/capture the import binding and reject all uses of that binding before the boundary, require and validate one exact canonical import declaration while rejecting every other form, or—preferably—remove the module from the production import graph entirely. Add fixtures for a renamed binding and the inline-import form.

scripts/audit_crypto_boundary.zig
```diff
@@ -0,0 +1,960 @@
...
+        try violations.append(allocator, .{
+            .path = try allocator.dupe(u8, path),
+            .needle = "test_quic_crypto.",
+            .rationale = "test_quic_crypto is a test-only provider; production code must be injected a provider by its caller, never construct one, and this usage falls before the file's test boundary.",
+        });
+    }
+
+    const post_boundary_region = contents[boundary..];
```
Jaruso: [P1] The absence of a `pub` declaration after the marker does not make that region unreachable. Zig declaration order is irrelevant, so a public production API before the marker can call a private helper declared afterward:

```zig
pub fn liveProvider() CryptoProvider {
    return providerForLivePath();
}

const testing = std.testing;

fn providerForLivePath() CryptoProvider {
    return test_quic_crypto.testDefaultProvider();
}
```

This passes: the production region contains no direct `test_quic_crypto.` use, and the post-boundary declaration is private, but `liveProvider` still exposes the test singleton to production. The public-pattern list is also incomplete (`pub inline fn`, `pub noinline fn`, `pub usingnamespace`, etc.), though even an exhaustive `pub` matcher would not fix the private-helper path.

A textual boundary cannot establish call-graph reachability. Move these consumers into separate test-only modules or declarations local to `test` blocks, so production code cannot name them. At minimum, enforce a file layout where no top-level declarations—public or private—exist after the boundary except test blocks, and add the private-helper fixture above.

Jaruso (commented):
Follow-up review on `a83537e`. This commit fixes the five findings from the previous pass: unqualified legacy calls are covered, the HMAC/HKDF exceptions are narrowed, the recursive QUIC scan includes legacy wrappers, importer discovery is recursive, and CI run 1100 is fully green (including the appliance profile).

Two blockers remain in the test-only provider enforcement:

- The check is tied to the local identifier spelling `test_quic_crypto`, not to the imported module. Renaming the binding or calling through an inline `@import` makes a production call invisible.
- A textual test boundary plus “no `pub` declarations afterward” does not prove the remainder is unreachable. Zig declarations are order-independent, so a public production function before the marker can call a private helper after it; the current scan accepts that path. The public-declaration pattern list is also not syntactically complete (`pub inline fn`, `pub usingnamespace`, etc.).

The robust resolution is to make `test_quic_crypto` genuinely absent from production module graphs, or move every consumer into separate test-support roots / local `test` scopes. A source guard can supplement that structure, but the current textual partition cannot establish production unreachability.

The live QUIC implementation, typed profile, documentation matrix, host-target audit build, and all previously reviewed runtime/lifetime fixes otherwise look good. #490 remains explicitly incremental beyond this PR.

docs/CRYPTO_PROVIDER_AUDIT.md
```diff
@@ -0,0 +1,82 @@
...
...ure-Zig provider singleton used by QUIC/H3 tests and tools that exercise the seam without the native `Runtime` composition root. Isolation from production code is structural, not scanned: `build.zig` wires it only into `quic_test_mod`/`exe_test_mod` (used solely by `zig build test-quic`/`zig build test`'s `quic_tests`/`exe_unit_tests`), never into `quic_mod`/`exe_mod` (used by `exe`/`run_cmd` and every other production consumer). Any reachable reference to it from code compiled as part of the production module graph — under any local binding name, any indirection such as an inline `@import`, at any position in the file — is therefore a Zig compiler error, not something `scripts/audit_crypto_boundary.zig` has to detect after the fact (#490 sixth-pass review: an earlier revision tried a source-text scan for this — a fixed importer list, then a "before/after a textual test-boundary marker" heuristic — which a renamed binding or an inline `@import` defeated, and which couldn't see that Zig declarations are order-independent, so a public declaration before the marker could call a private helper physically written after it). See the comment beside `test_quic_crypto_mod` in `build.zig`. |
```
Jaruso: [P2] The structural conclusion here is correct, but the statement that `build.zig` wires this module **only** into `quic_test_mod` / `exe_test_mod` is not accurate. The same helper is also deliberately injected into `quic_h3_smoke_mod`, `quic_h3_e2e_mod`, `quic_h3_udp_mod`, the test-only `udp_http3_runtime_mod`, and the opt-in `h3_interop_mod` executable root.

That does not reopen the production leak—the important property is that it is absent from `quic_mod`, `exe_mod`, and the shipped `tardi` graph—but the ownership matrix is supposed to be the precise checked-in disposition for #490. Please describe the actual rule instead, for example:

```text
Wired only into explicit test/tool roots (`quic_test_mod`, `exe_test_mod`,
QUIC/H3 smoke/e2e/UDP test modules, the test Runtime clone, and the opt-in
H3 interop tool), never into `quic_mod`, `exe_mod`, or the shipped `tardi`
module graph.
```

The adjacent comment beside `test_quic_crypto_mod` in `build.zig` should use the same wording rather than “wired only into quic_test_mod and exe_test_mod.”

Jaruso (commented):
Re-review of `1223ed1` after replacing the source-text test-boundary heuristic with structural module isolation.

The actual enforcement now looks sound: `test_quic_crypto` is absent from the production `quic_mod`/`exe_mod` graphs, reachable production use therefore fails at Zig module resolution regardless of aliasing or declaration order, the weaker scanner was removed, and CI run 1101 passed the full build/test/appliance/dependency/packaging matrix.

I found no remaining runtime, provider-routing, capability-profile, lifetime, or audit-pattern blocker. One P2 documentation mismatch remains inline: the ownership matrix says the helper is wired only into the two test twins, while `build.zig` also intentionally wires it into several standalone test and interop-tool roots. The architectural conclusion is still correct, but #490 requires the checked-in disposition to describe the actual boundary precisely.

PR #545, opened 2026-08-01
crypto: complete native TLS CryptoProvider ownership

Jaruso:
## Summary

Closes the last two "Open follow-up" rows in `docs/CRYPTO_PROVIDER_AUDIT.md` and completes issue #490. PR #544 already migrated QUIC packet/header protection onto `crypto.provider.CryptoProvider`; this PR does the same for the native TLS 1.3 key schedule and handshake engine, so every keyed cryptographic operation in native TLS/QUIC/PKI/resumption code now has a final provider-owned disposition.

### Migrated operations

- **`src/tls/key_schedule.zig`**: HKDF-Extract, HKDF-Expand-Label, early/handshake/master-secret derivation, handshake/application traffic secrets, resumption master secret, resumption PSK, client early traffic secret, Finished-key derivation, and Finished `verify_data` all cross `CryptoProvider`. `verify_data` is expressed as `HKDF-Extract(salt = finished_key, ikm = transcript_hash)` per RFC 5869's `HKDF-Extract(salt, IKM) = HMAC-Hash(salt, IKM)` identity, rather than adding a new generic HMAC entry point to the provider boundary — cross-checked in tests against a direct HMAC-SHA256 computation. `KeySchedule.init`/`.initWithPsk`/`.applicationSecrets` and the free HKDF-driven functions are now fallible, propagating the provider's typed `InvalidInput`/`UnsupportedCapability` errors. Every output buffer's `errdefer`/`defer` cleanup is armed *before* the fallible provider call that fills it (the provider interface never promises `out` is untouched on error), using the canonical `provider.secureZero` rather than `std.crypto.secureZero` directly — proven by a fault-provider fixture that writes a recognizable prefix into every output buffer and then errors. Unkeyed transcript hashing (`Sha256.hash` for the comptime empty-transcript constant) stays provider-independent by design — it has no secret input. The one documented exception is `derived_early_secret`, the zero-PSK schedule's fixed "derived" early secret: a public constant (all-zero PSK) computed once at compile time because no provider exists at comptime and none is needed. This module has no test block of its own — its direct-crypto fixtures (the HMAC cross-check, raw `CryptoProvider` vtable construction) live in a new `key_schedule_tests.zig`, compiled only through a dedicated `key_schedule_test_root_mod` in `build.zig` that is never reachable from the production `tls_core` module (`src/tls/root.zig`) or any of its consumers (native TLS/QUIC/HTTP, the `tardi` executable) — a Zig compiler-enforced guarantee, the same class as PR #544's `test_quic_crypto` module-graph isolation.
- **`src/tls/tls13_backend.zig`**: X25519 ephemeral key-share generation and shared-secret derivation route through `CryptoProvider.generateKeyShare`/`.deriveSharedSecret`, with `errdefer`/`defer` wipes armed before each fallible call. CertificateVerify peer-signature authentication routes through `CryptoProvider.verify`, preflighted against `capabilities().supportsSignature`. `error.InvalidInput` from `.verify` means a structurally malformed *public key* and maps to `.invalid_certificate` (`bad_certificate`); a malformed *or* incorrect *signature* is `error.AuthenticationFailed` and maps to `.invalid_signature` (`decrypt_error`, RFC 8446 §4.4.3) — the same split the pre-migration direct-`std.crypto.sign` code made. A provider missing a negotiated signature capability (or a post-capability-check `error.UnsupportedCapability`) is a local fault (`error.SecretExportFailed`), not attributed to the peer, since the scheme was already screened against this side's own policy and the peer's actual offer. This file has zero remaining direct keyed-crypto calls of its own.
- **`src/tls/credentials.zig`**: `Identity.sign` (the fixed/native credential's local CertificateVerify signing) now signs through the opaque `provider.SigningKey` handle instead of calling `Ed25519.KeyPair.sign`/`EcdsaP256Sha256.KeyPair.sign` directly. A new `pure_zig.SoftwareEcdsaP256SigningKey` joins the existing Ed25519 `SoftwareSigningKey` so both fixed-identity key types go through the same seam. The existing `CredentialProvider`/`SelectedCredential` async pending/cancel/release contract is unchanged — only what the concrete credential calls internally changed.

### Provider injection and lifetime

`Tls13Backend` now takes an explicit `crypto.provider.CryptoProvider` constructor parameter, threaded from the same provider each composition root already owns (`http3_runtime.Runtime`, `native_tls_connection.Connection`, `edge_gateway.zig`) rather than a hidden default. Ephemeral key-share entropy now comes from the provider's own injected `Entropy` (production: OS CSPRNG via `production_crypto.OsEntropy`; tests: `pure_zig.DeterministicEntropy`), not a caller-supplied seed — `Tls13Backend.Entropy` drops `key_share_seed`/`retry_key_share_seed` entirely, keeping only `hello_random` (ClientHello/ServerHello random, unrelated to key exchange). Every ephemeral private scalar and shared secret is wiped via `secureZero`/`wipeEphemeral()` on every success, failure, and teardown path.

Native QUIC and record-mode TLS use the same explicit provider selected by their composition root. Test roots each own an independent, per-instance deterministic provider — never a shared process-wide mutable stream. There is no `testHandshakeProvider()` or equivalent global anywhere in the tree anymore: `tests/support/quic_crypto.zig` exposes only `HandshakeProviderStorage`, an owned-storage type every caller (harness struct field, or a same-function local for a self-contained test) seeds with its own call to `.init(seed)`. This is threaded through `DirectHarness`, `SocketHarness`, `RealHandshakeHarness`, `TestPair`, `MigrationPair`, `RuntimeCidHarness`, the QUIC/H3 `Smoke`/`Sim` harnesses, and ~150 standalone test/helper functions across `tls13_backend.zig`'s own inline tests, `tls13_backend_tests.zig`, `connection.zig`, `tls_backend.zig`, `http3_runtime.zig`, `quic_h3_smoke.zig`, and `quic_h3_udp_smoke.zig`. `tests/h3_interop_tool.zig` — a real network-facing process — owns an OS-backed `production_crypto.Provider` for its own lifetime instead of a deterministic test provider, since predictable ephemeral X25519 keys would be a real forward-secrecy concern there. Verified empirically, not just by inspection: the two golden-value tests in `tls13_backend_tests.zig` pass under `--test-filter` in complete process isolation, byte-identical to a full-suite run — their literals were recomputed against this now-genuinely-hermetic entropy. PR #544's `test_quic_crypto` production/test module-graph isolation is untouched.

### Why unkeyed transcript hashing stays local

SHA-256 transcript hashing has no secret input — hashing a public handshake-message transcript is not a keyed operation, and the provider boundary deliberately has no `hash` entry point for it (documented in `docs/CRYPTO_PROVIDER.md`'s capability matrix). This is unchanged by this PR.

### Remaining narrow exceptions

- `key_schedule.zig`'s `derived_early_secret` comptime block (two exact-line exemptions in `scripts/audit_crypto_boundary.zig`) — documented above.
- `RSA-PSS-RSAE-SHA256` and `HKDF-SHA384` deliberately stay `not_integrated` for the `tls_handshake` consumer in `src/crypto/profile.zig`: the native engine's CertificateVerify only negotiates Ed25519/ECDSA-P256 (never RSA), and its live protocol path negotiates `TLS_AES_128_GCM_SHA256` only (never selecting SHA-384), even though the provider primitives exist, `key_schedule.zig`'s generic resumption helpers correctly route `.sha384` when called with it, and RSA's PKI integration is separately `.live`. The profile does not claim consumer support the live protocol path cannot execute.

## Architecture enforcement

`scripts/audit_crypto_boundary.zig` now scans both migrated files: `tls13_backend.zig` as a plain whole-file check with **no exceptions** (zero direct keyed-crypto patterns remain, including in its own inline test block), and `key_schedule.zig` with the two narrow exact-line exemptions above — no `production_only_marker`, no test-block carve-out. A textual "scan only before this position" marker is not a sound production/test reachability boundary (Zig declarations are order-independent: a production declaration positioned before a marker could still call a private helper positioned after it without the scan ever seeing the call). `key_schedule.zig`'s own tests were moved out into the separate, unscanned, structurally-unreachable-from-production `key_schedule_tests.zig` instead, closing that gap the same way PR #544 closed it for `test_quic_crypto` by using the Zig module graph rather than a text scan. New fixture tests prove each of the three bypass classes the issue calls out is rejected for both files: a direct call, an aliased call (a differently-named local alias for the same concrete type), and a helper-function indirection (a forbidden call wrapped in an arbitrarily-named function), plus a fixture proving the specific caller-before/helper-after "decoy marker" bypass class is caught even without a marker configured.

## Documentation and capability profile

- `docs/CRYPTO_PROVIDER_AUDIT.md`: both "Open follow-up" rows now read "Provider-owned", with the guard-extension mechanism (including the dedicated test-module boundary) documented accurately against the final implementation.
- `docs/CRYPTO_PROVIDER.md`: the supported-profile-matrix table's HKDF-SHA256, X25519, Ed25519, and ECDSA-P256-SHA256 rows are updated to "live" integration; HKDF-SHA384 stays explicitly not-live with the reasoning above; the entropy-injection and SigningKey sections describe the TLS engine as a real user of those abstractions rather than only their documented intent.
- `src/crypto/profile.zig`: `tls_handshake` consumer integration flips to `.live` for HKDF-SHA256, X25519, Ed25519, ECDSA-P256-SHA256, and injected-random-bytes; HKDF-SHA384 and RSA-PSS-RSAE-SHA256 stay `.not_integrated` deliberately (see above). A pinning test, `"native TLS 1.3 handshake engine is live-integrated through CryptoProvider"`, anchors all of this the same way the existing QUIC pinning test does.

## Acceptance criteria (issue #490)

- [x] Every keyed cryptographic operation in native TLS/QUIC/PKI/resumption code has an explicit owner and audit disposition — `docs/CRYPTO_PROVIDER_AUDIT.md`'s table has no remaining "Open follow-up" rows.
- [x] Disallowed direct keyed `std.crypto` operations in protocol code are migrated to the shared provider/security seam — `key_schedule.zig`/`tls13_backend.zig` migrated; `credentials.zig`'s local signing migrated.
- [x] QUIC header protection has a provider-neutral ownership path for supported algorithms — completed by PR #544, unchanged here, still passing.
- [x] No duplicate per-protocol crypto abstraction is introduced — `credentials.zig`'s existing `CredentialProvider`/`SelectedCredential`/`PeerVerifier` contracts are preserved as-is; no new TLS-specific crypto vtable added.
- [x] The crypto profile distinguishes provider capability from protocol integration/product enablement — `src/crypto/profile.zig` rows updated with per-consumer `IntegrationStatus`.
- [x] The profile no longer claims consumer support the live protocol path cannot execute — RSA-PSS and HKDF-SHA384 stay `not_integrated` for `tls_handshake` explicitly, with a pinning test.
- [x] An automated architecture guard prevents new unapproved direct keyed-crypto dependencies — `scripts/audit_crypto_boundary.zig` extended with fixture tests for both files, and the guard's own test-only exemption is now a structural module-graph boundary rather than a text-scan marker.
- [x] OpenSSL remains out of the native/appliance link path — unchanged; `zig build -Dtls-profile=appliance` still builds and passes with no libcrypto linkage.
- [x] `docs/CRYPTO_PROVIDER.md` matches the implemented architecture — updated.
- [x] Existing deterministic vectors and protocol tests remain green — RFC 8448 vectors byte-identical; the harness-driven golden-value tests were recomputed against the now-verified-hermetic key-exchange entropy source and pass under `--test-filter` in isolation.

## Test plan

At minimum covers: RFC 8448 key-schedule outputs through the provider path; zero-PSK and PSK-resumed schedule initialization; SHA-256/SHA-384 resumption derivations; Finished key and `verify_data` (cross-checked against direct HMAC); client early traffic secret; deterministic X25519 key-share generation via injected test entropy; shared-secret derivation success and invalid-peer-key handling; missing X25519/hash capability; entropy failure during key-share generation; a fault provider that partially writes an HKDF output before failing, proving every entry point wipes it; CertificateVerify success, authentication failure (malformed key vs. malformed/wrong signature, mapped to the correct distinct alert), missing-capability local-fault mapping, and unsupported-capability handling; local signing through the opaque credential/provider-backed handle; teardown/failure-path wiping; QUIC and record-mode handshakes completing full end-to-end through the same migrated engine; audit fixtures for each newly forbidden bypass class, including the marker-based decoy-position bypass; hermeticity verified directly via `--test-filter` on the previously order-dependent golden-value tests.

### Validation commands run and results

```
zig fmt --check .                                        # pass
zig build                                                 # pass
zig build audit-crypto-boundary                           # pass
zig build test-crypto -freference-trace                   # pass
zig build test-tls -freference-trace                      # pass
zig build test-quic -freference-trace                     # pass
zig build test -freference-trace                          # pass
zig build -Dtls-profile=appliance                         # pass
zig build -Dtls-profile=appliance test -freference-trace  # pass
```

Also ran, all passing:
```
zig build test-crypto-vectors -freference-trace
zig build test-crypto-openssl -freference-trace   # against the real openssl CLI as an out-of-process oracle
zig build test-quic-h3-driver -freference-trace
zig build test-crypto-corpus -freference-trace
zig build build-h3-interop
```

Closes #490


Jaruso (commented):
Reviewed the full 24-file diff on `c74ae8d` against every acceptance criterion in #490. CI run 1104 is fully green across format, multi-platform builds/tests, appliance profiles, dependency/linkage audit, H3 interop, PKI differential, packaging, perf, and security.

The core migration is real: runtime HKDF/Finished operations, X25519 key exchange, CertificateVerify verification, and fixed-credential signing now cross the intended provider/security seams, and production composition roots inject providers explicitly. However, six blockers mean this cannot yet close #490:

1. `key_schedule.zig`'s architecture check trusts a textual test marker as a reachability boundary; a production declaration above it can call a direct-crypto helper below it, bypassing the guard.
2. The capability registry marks HKDF-SHA384 `.live` for `tls_handshake`, although no live native TLS path can select SHA-384 and general-purpose OpenSSL runs outside this provider seam.
3. `CryptoProvider.verify`'s `InvalidInput` conflates malformed public-key and malformed signature encodings, so malformed CertificateVerify signatures now map to `bad_certificate` instead of the pre-migration `decrypt_error` behavior.
4. A provider missing the selected signature capability is a local configuration/provider failure, but the engine maps it to `unsupported_certificate`, incorrectly blaming the peer.
5. Handshake tests and the H3 interop tool use one process-global mutable deterministic entropy stream. Focused/filter runs and test ordering can change captured key shares/goldens, and the network interop tool uses predictable X25519 private keys.
6. Provider output cleanup is armed only after successful calls. The provider interface does not promise output buffers remain untouched on error, so a future provider that writes a prefix and then fails can leave key-schedule or key-share material resident; the new code also continues bypassing the canonical `provider.secureZero` helper.

The docs and PR acceptance checklist currently assert the opposite of several of these points, so they need to be corrected with the code. Once these are addressed, re-run the same full matrix and walk #490's criteria again before retaining `Closes #490`.

scripts/audit_crypto_boundary.zig
```diff
@@ -268,6 +314,31 @@ const file_checks_with_exceptions = [_]FileCheckWithExceptions{
...
+        .exempt_exact = &.{
+            "const HkdfSha256 = crypto.kdf.hkdf.HkdfSha256;",
+            "break :blk crypto.tls.hkdfExpandLabel(HkdfSha256, early_secret, \"derived\", &empty_transcript_hash, hash_len);",
+        },
+        // This module's own test block (after the marker) cross-checks the
+        // HKDF-Extract-as-HMAC `verifyData` trick against a direct
+        // `std.crypto.auth.hmac.sha2.HmacSha256` computation and builds raw
+        // `CryptoProvider` vtables for typed-error-propagation fixtures —
+        // legitimate test-only direct crypto use, not production code.
+        .production_only_marker = "\nconst testing = std.testing;",
```
Jaruso: [P1] This textual marker is not a production/test reachability boundary in Zig. Declarations are order-independent, so the following passes the audit while putting direct keyed crypto on a live path:

```zig
pub fn verifyData(...) [32]u8 {
    return legacyVerifyData(...);
}

const testing = std.testing;

fn legacyVerifyData(...) [32]u8 {
    const Hmac = std.crypto.auth.hmac.sha2.HmacSha256;
    var out: [32]u8 = undefined;
    Hmac.create(&out, message, key);
    return out;
}
```

The scanned prefix contains only the unrecognized helper name, and the forbidden call is after the truncation point. This is the same structural class of bug #544 removed for `test_quic_crypto`: source position cannot establish reachability.

Move the direct-HMAC cross-check and raw-vtable fixtures into a separate test-only module/file, then scan all of `key_schedule.zig` with no marker. Add the exact caller-before-marker/private-helper-after-marker fixture above; the current helper-indirection fixture places the helper inside the scanned region and therefore does not prove this bypass is closed. Update the audit document's claim accordingly.

src/crypto/profile.zig
```diff
@@ -141,21 +141,21 @@ pub const Row = struct {
...
+    row(.{ .hkdf = .sha256 }, "HKDF-SHA256", .zig_std_crypto, .supported, .openssl_provider, .provider_deferred, "RFC 5869/std.crypto and TLS expand-label parity", .{ .{ .tls_handshake, .live }, .{ .quic_tls_bridge, .live }, .{ .quic_packet_protection, .live } }, .{ .native_appliance, .general_purpose_openssl }, "QUIC Initial/Handshake/1-RTT key derivation in src/quic/tls_adapter.zig and the TLS 1.3 key schedule (src/tls/key_schedule.zig: HKDF-Extract, HKDF-Expand-Label, and Finished verify_data) both run through CryptoProvider (#490)."),
+    row(.{ .hkdf = .sha384 }, "HKDF-SHA384", .zig_std_crypto, .supported, .openssl_provider, .provider_deferred, "provider capability and expand-label tests", .{.{ .tls_handshake, .live }}, .{.general_purpose_openssl}, "The TLS 1.3 key schedule's SHA-384 resumption-master-secret/PSK derivations (src/tls/key_schedule.zig) run through CryptoProvider (#490); the native appliance negotiates TLS_AES_128_GCM_SHA256 only, so HKDF-SHA384 is not selected there. HKDF additions require TLS 1.3 label coverage."),
```
Jaruso: [P1] `.tls_handshake = .live` overclaims SHA-384 protocol integration. The native TLS engine still negotiates only `TLS_AES_128_GCM_SHA256`; its live resumption/ticket path therefore derives SHA-256 secrets. The generic SHA-384 helper is callable and tested, but that proves provider capability/helper coverage, not a live protocol path. The general-purpose OpenSSL backend can negotiate SHA-384 only outside this `CryptoProvider` seam.

Keep the primitive/provider capability as supported, but leave the TLS consumer `.not_integrated` (or otherwise non-live) until a native suite/path can actually select SHA-384. The new pinning test and both documentation tables must reflect that distinction. Avoid using “a helper can execute” as the definition of consumer integration—the acceptance criterion explicitly forbids that inference.

src/tls/tls13_backend.zig
```diff
@@ -3239,53 +3315,70 @@ pub const Tls13Backend = struct {
...
                 }
-                const public_key = EcdsaP256.PublicKey.fromSec1(parsed.pubKey()) catch return .invalid_certificate;
-                const sig = EcdsaP256.Signature.fromDer(signature) catch return .invalid_signature;
-                sig.verify(content, public_key) catch return .invalid_signature;
+                break :blk .ecdsa_secp256r1_sha256;
             },
             else => unreachable,
-        }
+        };
+        if (!self.crypto_provider.capabilities().supportsSignature(scheme)) return .unsupported_certificate;
```
Jaruso: [P1] A missing signature capability is a local provider/configuration failure, not an unsupported peer certificate. With an injected provider lacking Ed25519, a perfectly valid Ed25519 certificate and CertificateVerify reach this branch and the engine sends the peer-facing `unsupported_certificate` alert. The later `error.UnsupportedCapability` branch has the same problem.

Treat this like missing X25519/HKDF support: validate the required/intersecting provider capabilities before advertising/accepting the profile, or make `checkProofOfPossession` fallible and map both cases to the engine's local provider failure (`SecretExportFailed`/`internal_error`) rather than a `ProofResult` attributed to the peer. Add a provider-without-selected-signature test that asserts the local failure/alert mapping.

src/tls/tls13_backend.zig
```diff
@@ -3239,53 +3315,70 @@ pub const Tls13Backend = struct {
...
+            // possession failure (decrypt_error, RFC 8446 §4.4.3).
+            error.AuthenticationFailed => .invalid_signature,
+            // `provider.verify` reports `InvalidInput` for a structurally
+            // malformed public-key or signature encoding (see
+            // `pure_zig.verifyImpl`) — the same defect class the
+            // pre-migration code caught with its own `PublicKey.fromBytes`
+            // check and mapped to `.invalid_certificate` (bad_certificate),
+            // distinct from a validly-encoded signature that simply fails to
+            // verify.
+            error.InvalidInput => .invalid_certificate,
```
Jaruso, line 3396: [P1] `InvalidInput` cannot be mapped wholesale to `.invalid_certificate`. The provider currently returns that same error for two different inputs:

- malformed/invalid `parsed.pubKey()`;
- malformed CertificateVerify signature encoding (wrong Ed25519 length or invalid ECDSA DER).

Before this migration, malformed key material mapped to `bad_certificate`, but malformed CertificateVerify signatures mapped to `.invalid_signature`/`decrypt_error`. This branch changes the latter into `bad_certificate`, and the new comment/audit documentation incorrectly says the old implementation treated both alike.

Preserve the distinction in the provider taxonomy—for example, separate invalid public-key encoding from invalid signature encoding, or have signature parse failures report `AuthenticationFailed` while key parse failures remain `InvalidInput`. Then test malformed Ed25519 length and malformed ECDSA DER as `decrypt_error`, malformed certificate public keys as `bad_certificate`, and well-formed wrong signatures as `decrypt_error`.

tests/support/quic_crypto.zig
```diff
@@ -30,3 +30,52 @@ const test_provider_backing: crypto.pure_zig.Provider = crypto.pure_zig.Provider
...
+// `testDefaultProvider` above deliberately fails every entropy draw — correct
+// for its existing callers, QUIC packet-protection-only paths that derive
+// AEAD/header-protection keys from already-negotiated TLS secrets and must
+// never need ambient randomness. Ephemeral X25519 key-share generation is
+// different: it is itself a real entropy draw, now routed through the same
+// injected `CryptoProvider` (#490's native TLS engine migration), so a QUIC/H3
+// test or tool that constructs a full `Tls13Backend` end-to-end needs a
+// provider whose entropy source actually produces bytes.
+
+var handshake_entropy_state = crypto.pure_zig.DeterministicEntropy.init(0x71a5_c0de);
```
Jaruso: [P1] This process-global mutable entropy stream makes the new handshake tests non-hermetic. Every `testHandshakeProvider()` call advances the same state, so a test's X25519 share depends on which tests ran before it and whether a filter selected it. That is especially visible in `tls13_backend_tests.zig`, where exact traffic-secret goldens are now captured while `clientProvider()`/`serverProvider()` also retain mutable static streams: running the KAT alone can start from different entropy state than running the full file. It is also unsafe if tests are ever parallelized.

The same helper is passed to `h3_interop_tool`, giving a network-facing interop process predictable, repeatable ephemeral private keys after every restart.

Use the `HandshakeProviderStorage` added below as an owned field/local of each harness/backend pair, with explicit domain-separated seeds and stable addresses. Reset or embed the analogous client/server provider storage in `tls13_backend_tests.zig`. The interop executable should instead own an OS-backed `production_crypto.Provider` for its lifetime. Remove the shared helper entirely so the PR body's “no shared mutable global” claim becomes true.

src/tls/key_schedule.zig
```diff
@@ -54,38 +89,57 @@ pub const KeySchedule = struct {
...
+        var master_secret: [hash_len]u8 = undefined;
+        try crypto_provider.hkdfExtract(.sha256, &derived_handshake, &zeros, &master_secret);
+        errdefer crypto.secureZero(u8, &master_secret);
+
+        var client_handshake_traffic: [hash_len]u8 = undefined;
+        try crypto_provider.hkdfExpandLabel(.sha256, &handshake_secret, "c hs traffic", &hello_transcript_hash, &client_handshake_traffic);
+        errdefer crypto.secureZero(u8, &client_handshake_traffic);
+
+        var server_handshake_traffic: [hash_len]u8 = undefined;
+        try crypto_provider.hkdfExpandLabel(.sha256, &handshake_secret, "s hs traffic", &hello_transcript_hash, &server_handshake_traffic);
```
Jaruso, line 156: [P1] Failure cleanup is registered only after provider calls succeed. `CryptoProvider.hkdfExtract`/`hkdfExpandLabel` do not promise that `out` remains untouched or is zeroed when they return an error, so a conforming future provider may write part of `server_handshake_traffic` and then fail here, leaving secret-derived bytes on the stack. The same ordering appears for `early_secret`, `derived_early`, every retained schedule field, `applicationSecrets`, `finishedKey`/`verifyData`, and the backend's `generateKeyShare`/`deriveSharedSecret` output buffers.

Arm cleanup before the fallible call, using the canonical project helper:

```zig
var server_handshake_traffic: [hash_len]u8 = undefined;
errdefer provider.secureZero(&server_handshake_traffic);
try crypto_provider.hkdfExpandLabel(
    .sha256,
    &handshake_secret,
    "s hs traffic",
    &hello_transcript_hash,
    &server_handshake_traffic,
);
```

For transient values use `defer provider.secureZero(...)` before `try`; for key-share storage likewise register the private-scalar wipe before `generateKeyShare`. Apply this systematically and add a fault-provider fixture that writes a recognizable prefix into each output and then errors, proving the caller clears it. This also completes #490's requirement that protocol code use the shared security zeroization seam rather than direct `std.crypto.secureZero` calls.

src/tls/root.zig
```diff
@@ -12,6 +12,7 @@ pub const events = @import("events.zig");
 pub const algorithms = @import("algorithms.zig");
 pub const crypto_profile = @import("crypto_profile.zig");
 pub const key_schedule = @import("key_schedule.zig");
+pub const key_schedule_tests = @import("key_schedule_tests.zig");
```
Jaruso: [P1] Keep the direct-crypto test module out of the production `tls_core` graph

`build.zig` uses this file as the root of the production `tls_core_mod`, which is imported by the native TLS, QUIC, HTTP, and executable modules. Exporting `key_schedule_tests` here therefore lets production code resolve `tls_core.key_schedule_tests`, while `scripts/audit_crypto_boundary.zig` intentionally does not scan that file. A future production declaration in `key_schedule_tests.zig` could call its direct HMAC/HKDF fixtures and then be reached from production without triggering the guard—the same declaration-order/reachability class we were trying to eliminate.

Please remove this import from `src/tls/root.zig` and wire the test file through a genuinely test-only module/root in `build.zig`. For example:

```zig
// src/tls/test_root.zig, used only by the TLS test step
const tls_core = @import("tls_core");
test {
    _ = tls_core;
    _ = @import("key_schedule_tests.zig");
}
```

Then have `key_schedule_tests.zig` import `@import("tls_core").key_schedule` rather than becoming part of the production root. The invariant should be structural: no production module can resolve the file containing approved direct-crypto test fixtures.

src/tls/tls13_backend.zig
```diff
@@ -5214,14 +5357,30 @@ fn buildMaxNewSessionTicketMessage(allocator: std.mem.Allocator) ![]u8 {
...
+/// A fresh deterministic pure-Zig `CryptoProvider` for this module's own
+/// tests (#490), matching the fixture pattern `key_schedule.zig`'s test
+/// block uses: `DeterministicEntropy` (explicitly not a CSPRNG) plus the
+/// pure-Zig `Provider`, held as a persistent local so every call in this
+/// file's test block gets a live, valid provider without a hidden global
+/// default.
+fn testCryptoProvider() crypto_provider_pkg.CryptoProvider {
+    const pure_zig = crypto_pkg.pure_zig;
+    const State = struct {
+        var entropy = pure_zig.DeterministicEntropy.init(0x71357_490);
```
Jaruso: [P2] Remove the remaining process-global mutable test provider

This is still a hidden process-global deterministic entropy stream. The first inline test that generates an X25519 share advances it, so later tests receive values determined by execution order; parallel test execution would also race on `State.entropy`/`State.backing`. That leaves the prior non-hermetic-test finding only partially fixed despite removing `testHandshakeProvider()` elsewhere.

Use caller-owned storage per test instead of a function-static `State`, e.g.:

```zig
const TestProviderStorage = struct {
    entropy: pure_zig.DeterministicEntropy,
    backing: pure_zig.Provider,

    fn init(self: *TestProviderStorage, seed: u64) CryptoProvider {
        self.entropy = pure_zig.DeterministicEntropy.init(seed);
        self.backing = pure_zig.Provider.init(self.entropy.entropy());
        return self.backing.cryptoProvider();
    }
};

var storage: TestProviderStorage = undefined;
const cp = storage.init(0x71357_490);
```

Moving these inline tests into the new test-only module would also make it straightforward to give every test its own stable provider storage.

docs/CRYPTO_PROVIDER_AUDIT.md
```diff
@@ -15,31 +15,50 @@ fixed public constants, and documented temporary exceptions.
...
+same way: `tls13_backend.zig` carries no exceptions at all (a plain
+whole-file check, like `connection.zig`/`http3_runtime.zig`), and
+`key_schedule.zig` carries exactly two narrow exact-line exemptions for its
+one documented comptime "derived early secret" public constant, scanned
+against a forbidden-pattern variant that permits the provider-routed
+`crypto_provider.hkdfExpandLabel(`/`self.provider.hkdfExpandLabel(` calls
+while still blocking the legacy `crypto.tls.hkdfExpandLabel(` spelling — the
+same qualified/bare-call distinction `tls_adapter_zig_forbidden` already
+makes for the QUIC seam. `key_schedule.zig`'s own test block, after a
+`production_only_marker`, is exempt from scanning because it legitimately
```
Jaruso: [P2] Update this section to describe the new test-module boundary

This paragraph still says `key_schedule.zig` has an exempt test block after a `production_only_marker`, but the latest commit removed both the marker and the tests from that file. It now contradicts the implementation and the audit row above.

After fixing the module-graph issue in `src/tls/root.zig`, replace this with the actual rule: `key_schedule.zig` is scanned in full except for the two exact comptime lines, while direct-crypto differential fixtures live in a separate test-only module that production graphs cannot resolve and that is wired only into the TLS test target. That distinction is important evidence for #490’s documentation and enforcement acceptance criteria.

Jaruso (commented):
Re-reviewed the complete PR at `3463d3c` against issue #490 and the six findings from the previous pass.

The latest commit correctly fixes the provider error taxonomy, the HKDF-SHA384 capability overclaim, H3 interop entropy, provider-output failure cleanup, and the source-marker bypass inside `key_schedule.zig`. CI run 1105 is green across the build/test matrix, appliance profiles, linkage audit, H3 interop, PKI differential, packaging, and performance jobs.

Three blockers remain before this can close #490:

1. The new direct-crypto test module is exported by the production `tls_core` root, so it is not structurally test-only and remains outside the audit.
2. `tls13_backend.zig` still contains a process-global mutable deterministic provider used by inline tests, so focused/reordered/parallel runs are not hermetic.
3. The checked-in audit documentation still describes the removed `production_only_marker` carve-out.

Because of (1), the acceptance criterion requiring an automated guard that prevents new unapproved direct keyed-crypto dependencies is not yet satisfied. Because of (3), the documentation-match criterion is also still open. The PR description should also be refreshed before merge: it still says malformed signatures map through `InvalidInput`, advertises HKDF-SHA384 as `.live`, and describes the removed shared `testHandshakeProvider()`/marker design.

Once the production/test module boundary is real, the remaining global test provider is removed, and the docs/PR body are synchronized, the implementation portion of #490 looks clean.

src/tls/tls13_backend_tests.zig
```diff
@@ -6255,8 +6442,18 @@ test "provider and verifier mocks under allocation failure clean up" {
...
+/// storage, and this helper returns one by value, so a same-function local
+/// would dangle (#490 review). Function-static storage scoped to just this
+/// one helper's callers, *reset* to the same seed on every call rather than
+/// advanced — every call is independent of process history, which is
+/// sufficient here since no single test calls this helper more than once for
+/// two simultaneously-live backends.
 fn serverWithProvider(mock: *credentials.MockCredentialProvider) tls_backend.Tls13Backend {
-    return tls_backend.Tls13Backend.initServerWithProvider(serverEntropy(), mock.provider(), .record);
+    const State = struct {
+        var storage: ProviderStorage = .{};
```
Jaruso: [P2] This is still process-global mutable provider storage, despite the PR now claiming every test backend owns independent per-instance storage. Resetting the seed makes sequential calls deterministic, but each returned `Tls13Backend.crypto_provider` borrows `State.storage`. A second invocation before the first backend is finished—parallel test execution, or a future test that needs two simultaneously-live instances—overwrites the entropy/provider backing that the first backend still references. The same pattern remains in `clientWithLocalCredential` and `serverRequestingClientAuth`.

Please make the storage caller-owned, for example:

```zig
fn serverWithProvider(
    storage: *ProviderStorage,
    mock: *credentials.MockCredentialProvider,
) tls_backend.Tls13Backend {
    return tls_backend.Tls13Backend.initServerWithProvider(
        serverEntropy(),
        storage.init(server_provider_seed),
        mock.provider(),
        .record,
    );
}

var provider_storage: ProviderStorage = .{};
var server = serverWithProvider(&provider_storage, &mock);
defer server.deinit();
```

Apply the same out-parameter/caller-owned-storage shape to the client-credential and client-auth server helpers. That gives each returned backend a distinct stable owner and closes the remaining race/lifetime hole rather than relying on the current test runner executing these callers serially.

Jaruso (commented):
Re-reviewed the full PR at `b61c33f` against issue #490 and the previous findings.

The three blockers from the last pass are correctly addressed:

- `key_schedule_tests.zig` is no longer exported from the production `tls_core` root and now runs through a dedicated test-only artifact;
- the remaining function-static provider inside `tls13_backend.zig` itself was replaced with caller-owned per-test storage;
- the architecture documentation and PR description now match the marker-free implementation and corrected capability/error mappings.

CI run 1106 is fully green across format, security, Debug/ReleaseSafe/ReleaseFast, Linux/macOS/ARM, appliance linkage/tests, H3 interop, PKI differential, packaging, and performance smoke. The main `zig build test` job reports 2,948 passing tests and one skip, including the new 16-test key-schedule artifact.

One P2 test-lifetime issue remains: three factory helpers in `tls13_backend_tests.zig` still return backends borrowing function-static mutable `ProviderStorage`. Resetting those stores removes sequential order dependence, but it does not make them caller-owned or safe for overlapping invocations. Details and a concrete fix are in the inline comment.

Aside from that remaining helper pattern, I found no additional production, protocol, provider-boundary, profile, guard, or documentation blockers.

Jaruso (commented):
Final re-review at `ec6ac3d`: no code findings.

The final provider-storage fix is correct and complete: `serverWithProvider`, `clientWithLocalCredential`, and `serverRequestingClientAuth` now borrow caller-owned `ProviderStorage`, every call site supplies distinct stable storage, and no function-static provider state remains in `tls13_backend_tests.zig`. This closes the remaining parallelism/lifetime hole without changing handshake behavior.

I rechecked the full #490 acceptance criteria and the earlier findings: provider ownership, capability/profile truthfulness, CertificateVerify alert taxonomy, partial-write cleanup, structural test-module isolation, production entropy, documentation, and architecture enforcement all remain correct. All prior review threads are resolved. From the code-review perspective, this is ready to merge and close #490.

CI run 1107 currently has one macOS appliance-profile failure in the unchanged timing-sensitive UDP drain test (`quic_h3_udp_smoke...new_initial_seen`); the modified TLS test module and the general Linux/macOS unit jobs pass. That failure does not appear caused by this commit, but the required check should be rerun/green before merge.

PR #547, opened 2026-08-01
crypto: add reduced ECDSA-P256-SHA256 corpus and finalize corpus waivers

Jaruso:
## Summary
- Adds a bounded, provenance-preserving `wycheproof-ecdsa-p256-sha256-verify-reduced` suite (upstream `testvectors_v1/ecdsa_secp256r1_sha256_test.json`, pinned commit `fc24cd5b787d8e496bff31b0468af693a652b0f2`) covering `tcId`s 7, 6, 8, 33, 168, 192, and 448, copied verbatim from upstream (group index, classification, comment, flags, message, signature, public key).
- Generalizes `tests/crypto_corpus.zig`'s Ed25519-only verify execution into a single generic `executeVerify` path selected by an `Algorithm.signatureScheme()` helper, so both `.ed25519` and `.ecdsa_secp256r1_sha256` share one runner instead of duplicating it.
- Removes `ECDSA-P256-SHA256` from `skippedSuites`/manifest and registers it as an executable suite (7 cases) in `tests/crypto_corpus_manifest.zig`.
- Converts the RSA-PSS skip from temporary "outside this slice" wording into an explicit **permanent waiver**, pointing at the independent strict coverage already delivered in `src/crypto/rsa.zig` (#413 / PR #428).
- Updates `tests/vectors/wycheproof/README.md` with the final executable-coverage/waiver policy, including why hashes/HKDF/header-protection/entropy/helper operations are intentionally not corpus candidates.
- Registers the new suite id in `tests/crypto_vectors.zig`'s case registry so the top-level vector harness (#373) builds cleanly against the updated manifest.

## Test plan
- [x] `zig build test-crypto-corpus --summary all --error-style verbose` — 7/7 tests pass
- [x] `zig build test-crypto --summary all --error-style verbose` — 77/77 tests pass
- [x] `zig build test --summary all --error-style verbose` — 2948/2949 pass (1 pre-existing skip: UDP smoke test under socket-blocking sandbox, unrelated to this change)
- [x] `git diff --check` — clean
- [x] Verified malformed/non-canonical DER (missing zero-pad, BER long-form length, truncated length, zero/out-of-range `r`,`s`) all resolve to `authentication-failed` without panic; canonical valid signature and edge-case public key both succeed

Closes #374


tests/crypto_vectors.zig
```diff
@@ -91,6 +91,7 @@ const case_registry = [_]CaseMeta{
     .{ .id = "wycheproof-chacha20-poly1305-reduced", .algorithm = .{ .aead = .chacha20_poly1305 }, .providers = pure_zig_only, .class = .negative, .source = "tests/vectors/wycheproof/corpus.json", .license = "Apache-2.0", .reproduction = "zig build test-crypto-corpus" },
     .{ .id = "wycheproof-x25519-reduced", .algorithm = .{ .group = .x25519 }, .providers = pure_zig_only, .class = .negative, .source = "tests/vectors/wycheproof/corpus.json", .license = "Apache-2.0", .reproduction = "zig build test-crypto-corpus" },
     .{ .id = "wycheproof-ed25519-verify-reduced", .algorithm = .{ .signature = .ed25519 }, .providers = pure_zig_only, .class = .negative, .source = "tests/vectors/wycheproof/corpus.json", .license = "Apache-2.0", .reproduction = "zig build test-crypto-corpus" },
+    .{ .id = "wycheproof-ecdsa-p256-sha256-verify-reduced", .algorithm = .{ .signature = .ecdsa_secp256r1_sha256 }, .providers = pure_zig_only, .class = .negative, .source = "tests/vectors/wycheproof/corpus.json", .license = "Apache-2.0", .reproduction = "zig build test-crypto-corpus" },
```
Jaruso, line 94: **Blocking:** this registers the whole ECDSA corpus suite only as `.negative`, so `validateCapabilityMatrix()` still needs the existing ECDSA waiver to satisfy the supported algorithm's **positive** coverage requirement. That waiver says the harness does not carry an independent DER fixture, which is now factually false because this PR adds valid Wycheproof DER cases (`tcId` 7 and 448).

Please remove the ECDSA waiver and make executable valid ECDSA coverage visible to this harness. A minimal approach consistent with the current model is to add a positive `CaseMeta` and execute the pinned valid vector in `runKeyExchangeAndSignatureVectors`, for example:

```zig
.{
    .id = "wycheproof-ecdsa-p256-sha256-valid",
    .algorithm = .{ .signature = .ecdsa_secp256r1_sha256 },
    .providers = pure_zig_only,
    .class = .positive,
    .source = "C2SP Wycheproof fc24cd5... tcId 7",
    .license = "Apache-2.0",
    .reproduction = "zig build test-crypto-vectors",
},
```

Then log that ID and call `cp.verify(.ecdsa_secp256r1_sha256, public_key, message, signature)` with the pinned `tcId` 7 values, and delete the old ECDSA waiver. Alternatively, extend the corpus registry metadata so one suite can explicitly contribute both positive and negative classes. Either fix should ensure `validateCapabilityMatrix()` fails if positive ECDSA coverage is later removed, rather than silently falling back to a stale waiver.

Jaruso (commented):
The bounded ECDSA suite, generic verify dispatch, RSA-PSS permanent waiver, manifest updates, and README policy are otherwise aligned with #374. The selected cases and expected provider outcomes match the requested closeout shape.

One blocking acceptance issue remains in the top-level vector coverage accounting: this PR registers the new ECDSA corpus suite only as `.negative`, while retaining the existing pure-Zig ECDSA waiver that says the harness has no independent DER fixture. That statement is now false, and `validateCapabilityMatrix()` still uses the waiver—not executable positive ECDSA coverage—to satisfy the supported capability's positive requirement. Please remove the stale waiver and make the valid corpus coverage (`tcId` 7 and/or 448) count as positive coverage in `tests/crypto_vectors.zig`.

CI is currently red because `Test appliance profile (ubuntu-latest)` hit the unrelated UDP drain smoke assertion at `tests/quic_h3_udp_smoke.zig:993`; the same appliance profile passed on ARM and macOS, and all crypto-related/unit jobs passed. That looks unrelated/flaky, but the workflow should be rerun after the coverage-accounting fix.

tests/crypto_vectors.zig
Jaruso: Fixed at `6be5d13`: the stale ECDSA waiver is removed, pinned Wycheproof `tcId` 7 is now executed and registered as explicit `.positive` coverage, and the reduced corpus suite remains `.negative` coverage. `validateCapabilityMatrix()` now requires the executable positive vector because there is no ECDSA fallback waiver. Re-reviewed the full PR against #374; this thread is resolved.

Jaruso (commented):
Re-review complete at `6be5d13`. No additional code findings.

The previous blocker is fully addressed:

- the stale pure-Zig ECDSA waiver is removed;
- pinned Wycheproof `tcId` 7 is registered and actually executed as `.positive` ECDSA coverage;
- the reduced ECDSA corpus suite remains registered as `.negative` coverage;
- `validateCapabilityMatrix()` can no longer satisfy positive ECDSA coverage through a fallback waiver.

I rechecked all five changed files against #374. The generic signature dispatch is correctly shared by Ed25519 and ECDSA; the seven selected ECDSA cases preserve the pinned upstream public keys, group indices, `tcId`s, classifications, comments, flags, messages, and signatures; the expected provider outcomes match the documented contract; ECDSA is removed from skipped metadata; the RSA-PSS disposition is explicit and permanent; manifest/registry counts agree; parser bounds and provenance-safe failure reporting remain intact; and the README documents executable coverage, waivers, and intentionally non-applicable operations.

From the #374 implementation and acceptance-criteria perspective, this is ready. I am intentionally not treating the separate CI smoke-test failure as part of this re-review because it is being handled independently.

PR #548, opened 2026-08-01
http3: count post-GOAWAY stream rejections and deflake UDP drain smoke

Jaruso:
## Summary
- A request stream rejected directly by `http3.Conn.pump()` at `local_goaway_id` never became an `IncomingRequest`, so `Runtime` had no way to observe it — the runtime-level boundary rejection only counted requests that survived to `pollRequest()`. When the server installed GOAWAY before the client observed it, `Snapshot.h3_drain_request_rejections` stayed at zero, and `tests/quic_h3_udp_smoke.zig`'s drain smoke could time out waiting for `new_initial_seen` depending on which side won that race.
- Add `Metrics.goaway_request_rejections` to `http3.Conn` and increment it exactly at the admission-rejection site in `pump()` (`src/http3/conn.zig`).
- Fold that counter's delta into `Runtime`'s `Snapshot.h3_drain_request_rejections` around every `entry.h3.pump()` call, using a result-first shape so an unrelated `pump()` error from the same call can't suppress an already-real rejection (`src/http/http3_runtime.zig`). The existing runtime-level boundary-rejection path (for requests admitted before drain but rejected later) is unchanged and remains mutually exclusive with this one.
- Extend the existing `H3 conn: sent GOAWAY rejects boundary...` unit test with an assertion on the new counter, and add a new runtime-level test proving the H3 metric delta folds into the snapshot.
- Split the UDP smoke test's single `new_initial_seen` assertion into two (`sent_new_initial` first), so a future failure distinguishes "the drain rejection was never observed" from "the synthetic Initial was sent but never ingested."

## Test plan
- [x] `zig build test` — full suite green
- [x] `zig build test -Dtls-profile=appliance --seed 0x811b55b0` — reproduces the original CI shape, green (2708/2709, 1 pre-existing skip)
- [x] `zig build test-quic -Dquic-test-filter="udp smoke: HTTP/3 runtime drain lets admitted work finish and rejects new work" --seed 0x811b55b0` — the exact CI-failing seed, passes
- [x] Same focused test repeated with 6+ additional randomized seeds — all pass

Closes #546

### DIFF
--- a/src/http/http3_runtime.zig
+++ b/src/http/http3_runtime.zig
@@ -890,7 +890,20 @@ pub const Runtime = struct {
         // any response `serveRequest` queues still can't reach the wire
         // until `.application` packet building unlocks at establishment.
         if (!established and !self.zero_rtt_enabled) return;
-        entry.h3.pump(entry.conn) catch |err| {
+        // #546: a peer request stream that arrives at/above `local_goaway_id`
+        // is rejected by `entry.h3.pump()` itself, during admission, before
+        // it can ever become a `pollRequest()`-visible `IncomingRequest` —
+        // the boundary-rejection branch below never sees it. That's a
+        // distinct, mutually-exclusive rejection owner from this one, so
+        // fold its delta into the same snapshot counter here, unconditionally
+        // (a result-first shape, not inside the `catch`) so an unrelated
+        // protocol error from this same `pump()` call can't suppress an
+        // already-real rejection that happened earlier in the same call.
+        const rejected_before = entry.h3.metrics.goaway_request_rejections;
+        const pump_result = entry.h3.pump(entry.conn);
+        const rejected_after = entry.h3.metrics.goaway_request_rejections;
+        self.noteDrainRequestRejections(@intCast(rejected_after - rejected_before));
+        pump_result catch |err| {
             // An H3-level protocol error closes the connection with the
             // specific RFC 9114 §8.1 code the session layer recorded.
             const code = entry.h3.closeCode();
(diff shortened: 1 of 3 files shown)


src/http/http3_runtime.zig
```diff
@@ -890,7 +890,20 @@ pub const Runtime = struct {
...
+        // the boundary-rejection branch below never sees it. That's a
+        // distinct, mutually-exclusive rejection owner from this one, so
+        // fold its delta into the same snapshot counter here, unconditionally
+        // (a result-first shape, not inside the `catch`) so an unrelated
+        // protocol error from this same `pump()` call can't suppress an
+        // already-real rejection that happened earlier in the same call.
+        const rejected_before = entry.h3.metrics.goaway_request_rejections;
+        const pump_result = entry.h3.pump(entry.conn);
+        const rejected_after = entry.h3.metrics.goaway_request_rejections;
+        self.noteDrainRequestRejections(@intCast(rejected_after - rejected_before));
```
Jaruso, line 905: This ordering is the key correctness detail: folding the delta before handling `pump_result` preserves an admission rejection even if the same pump later returns an unrelated H3 error. I also verified this is the only production `entry.h3.pump()` call, so the runtime has no uninstrumented pump path.

Jaruso (commented):
Re-reviewed the full PR against issue #546. No blocking findings.

Verified:

- `http3.Conn.pump()` increments the new admission-rejection counter exactly where a post-GOAWAY request stream is reset and discarded.
- The runtime folds only the per-call counter delta, before inspecting `pump_result`, so an H3 error later in the same pump cannot hide a real rejection.
- This is the only production `entry.h3.pump()` call, so there is no uninstrumented runtime path.
- The admission-time path and the existing `pollRequest()` boundary path are mutually exclusive: a stream rejected during admission never enters `requests`, while a stream already admitted before drain can only be counted by the runtime boundary.
- The H3 unit test covers multiple direct admission rejections exactly once; the runtime test proves the delta reaches `Snapshot.h3_drain_request_rejections`; and the UDP smoke now distinguishes “rejection not observed” from “Initial sent but not ingested.”
- No timeout was increased or used as a workaround.
- CI run 1111 is green across Linux, ARM, macOS general, and macOS appliance-profile jobs, including the H3 interop and full relevant unit suites.

The implementation satisfies #546’s acceptance criteria and is ready to merge. Because this review is being submitted through the PR author’s account, I am leaving the approval-equivalent result as a comment rather than `APPROVE`.

PR #549, opened 2026-08-01
fix: harden secret lifecycle cleanup paths

Jaruso:
## Summary
- harden QUIC Retry key-ring lifecycle by wiping keys on overwrite/retire/deinit and avoiding by-value key retrieval from the ring
- harden local CID stateless-reset handling with owned secret-key storage, explicit registry deinit, token wipe-on-retire, and borrowed-key token derivation
- route TLS identity-loader key-material cleanup through canonical secret helpers and wipe PEM/base64/decode scratch + retained DER on cleanup paths
- tighten ticket-protection lifecycle cleanup by wiping ledger fingerprints on deinit, wiping temporary fingerprints after validation, and using canonical secure-zero in ticket plaintext free path while preserving deterministic zero-before-free behavior
- ensure QUIC connection teardown deinitializes local CID registry and wipes stateless reset key material

## Validation
- `ZIG_GLOBAL_CACHE_DIR="$TMPDIR/zig-global-cache" ZIG_LOCAL_CACHE_DIR="$TMPDIR/zig-local-cache" zig build test` (pass)
- VS Code diagnostics: no errors in touched files

## Issue
- Refs #375

### DIFF
--- a/scripts/audit_crypto_boundary.zig
+++ b/scripts/audit_crypto_boundary.zig
@@ -265,7 +265,7 @@ const file_checks_with_exceptions = [_]FileCheckWithExceptions{
         // anywhere else in the file would pass (#490 fifth-pass review).
         .forbidden = &(full_forbidden ++ [_][]const u8{"HmacSha256."}),
         .exempt_exact = &.{"const HmacSha256 = std.crypto.auth.hmac.sha2.HmacSha256;"},
-        .exempt_functions = &.{"statelessResetToken"},
+        .exempt_functions = &.{ "statelessResetToken", "statelessResetTokenInto" },
         .rationale = "Connection-ID/stateless-reset-token derivation (RFC 9000 §10.3.1) is a documented HMAC-SHA256 exception under a static process-lifetime key (docs/CRYPTO_PROVIDER_AUDIT.md), not TLS/QUIC-negotiated packet protection; no AEAD, ECDH, signature, KDF, or AES header-protection shortcuts, and no other HmacSha256 use, may be added here.",
     },
     .{
--- a/scripts/audit_crypto_boundary.zig
+++ b/scripts/audit_crypto_boundary.zig
@@ -265,7 +265,7 @@ const file_checks_with_exceptions = [_]FileCheckWithExceptions{
         // anywhere else in the file would pass (#490 fifth-pass review).
         .forbidden = &(full_forbidden ++ [_][]const u8{"HmacSha256."}),
         .exempt_exact = &.{"const HmacSha256 = std.crypto.auth.hmac.sha2.HmacSha256;"},
-        .exempt_functions = &.{ "statelessResetToken", "statelessResetTokenInto" },
+        .exempt_functions = &.{"statelessResetTokenInto"},
         .rationale = "Connection-ID/stateless-reset-token derivation (RFC 9000 §10.3.1) is a documented HMAC-SHA256 exception under a static process-lifetime key (docs/CRYPTO_PROVIDER_AUDIT.md), not TLS/QUIC-negotiated packet protection; no AEAD, ECDH, signature, KDF, or AES header-protection shortcuts, and no other HmacSha256 use, may be added here.",
     },
     .{
--- a/src/http/http3_runtime.zig
+++ b/src/http/http3_runtime.zig
@@ -338,7 +338,7 @@ pub const Runtime = struct {
         runtime.crypto_provider_state = tls_core.production_crypto.Provider.init(runtime.crypto_provider_entropy.entropy());
         var retry_key: [quic.path.token_key_len]u8 = undefined;
         compat.randomBytes(&retry_key);
-        runtime.secrets.retry_tokens.keys.install(0, retry_key);
+        runtime.secrets.retry_tokens.keys.install(0, &retry_key);
         crypto_pkg.secrets.secureZero(&retry_key);
         compat.randomBytes(&runtime.secrets.stateless_reset_key);
         errdefer runtime.secrets.deinit();
@@ -694,7 +694,7 @@ pub const Runtime = struct {
             .now_us = now,
             .initial_path = .{ .local = self.local_address, .remote = addressFromSockaddrIn(peer) },
             .initial_address_validated = retry_context != null,
-            .stateless_reset_key = self.secrets.stateless_reset_key,
+            .stateless_reset_key = &self.secrets.stateless_reset_key,
             .events = .{ .context = self, .emitFn = quicConnectionEvent },
         }) catch {
             allocator.destroy(backend);
@@ -1678,7 +1678,7 @@ const RuntimeCidHarness = struct {
             .crypto_provider = test_quic_crypto.testDefaultProvider(),
             .now_us = self.now_us,
             .initial_path = server_path,
-            .stateless_reset_key = [_]u8{0x44} ** 32,
+            .stateless_reset_key = &([_]u8{0x44} ** 32),
         });
         entry.* = .{
             .backend = server_backend,
--- a/src/http/http3_runtime.zig
+++ b/src/http/http3_runtime.zig
@@ -322,7 +322,9 @@ pub const Runtime = struct {
         var retry_key: [quic.path.token_key_len]u8 = undefined;
         compat.randomBytes(&retry_key);
         runtime.retry_tokens.keys.install(0, retry_key);
+        crypto_pkg.secrets.secureZero(&retry_key);
         compat.randomBytes(&runtime.stateless_reset_key);
+        errdefer runtime.wipeSecrets();
         http3.frame.validateLocallySupportedSettings(runtime.h3_settings) catch return error.InvalidH3Settings;
         runtime.h3_application_compat_len = (http3.early_data.encodeSettingsSnapshot(
             runtime.h3_settings,
@@ -354,7 +356,20 @@ pub const Runtime = struct {
         self.stopping.store(true, .release);
         if (self.thread) |thread| thread.join();
         _ = std.c.close(self.socket_fd);
+        self.wipeSecrets();
         self.* = undefined;
+    }
+
+    /// Wipe the Retry token key ring and the stateless-reset key. Factored
+    /// out of `deinit` (rather than inlined before its trailing
+    /// `self.* = undefined`) so it stays independently callable — including
+    /// from tests, which would otherwise be unable to observe the wipe at
+    /// all: safety-checked builds poison-fill a whole struct on `= undefined`,
+    /// which would overwrite the very zero bytes a post-`deinit` byte-level
+    /// assertion is trying to verify.
+    fn wipeSecrets(self: *Runtime) void {
+        self.retry_tokens.keys.deinit();
+        crypto_pkg.secrets.secureZero(&self.stateless_reset_key);
     }
 
     pub fn snapshot(self: *Runtime) Snapshot {
@@ -2043,6 +2058,45 @@ test "runtime init rejects unsupported local H3 setting h3_datagram" {
 
 test "runtime init rejects unsupported local H3 setting max_field_section_size" {
     try expectInvalidH3SettingsAtRuntimeInit(.{ .max_field_section_size = 1024 });
+}
+
+test "Runtime.deinit wipes the retry token key ring and the stateless reset key" {
+    var logger = logger_mod.Logger.init(.err, "http3-runtime-deinit-wipe-test");
+    var runtime = try Runtime.init(testing.allocator, &logger, .{
+        .listen_host = "127.0.0.1",
+        .quic_port = 0,
+    });
+    // `deinit()` ends with `self.* = undefined`, and safety-checked builds
+    // poison-fill the *whole* struct on that assignment — which would
+    // overwrite our zeroed bytes with poison before a post-`deinit` byte
+    // check ever ran, making the wipe unobservable either way. Call
+    // `wipeSecrets()` (the exact cleanup `deinit()` delegates to) directly
+    // instead, on this real `Runtime.init`-produced key material, and close
+    // the socket by hand since we're bypassing `deinit()`.
+    defer _ = std.c.close(runtime.socket_fd);
+
+    // Capture raw pointers into the still-installed key storage before the
+    // wipe: asserting only that lookups fail afterward would also pass an
+    // implementation that just cleared option tags without ever touching
+    // the key bytes.
+    const key0_ptr: *const [quic.path.token_key_len]u8 = &(runtime.retry_tokens.keys.keys[0].?);
+    const reset_key_ptr = &runtime.stateless_reset_key;
+
+    var saw_key_nonzero = false;
+    for (key0_ptr) |b| {
+        if (b != 0) saw_key_nonzero = true;
+    }
+    try testing.expect(saw_key_nonzero);
+    var saw_reset_nonzero = false;
+    for (reset_key_ptr) |b| {
+        if (b != 0) saw_reset_nonzero = true;
+    }
+    try testing.expect(saw_reset_nonzero);
+
+    runtime.wipeSecrets();
+
+    for (key0_ptr) |byte| try testing.expectEqual(@as(u8, 0), byte);
+    for (reset_key_ptr) |byte| try testing.expectEqual(@as(u8, 0), byte);
 }
 
 test "admissionAllowed enforces global and per-source caps at the boundary" {
--- a/src/http/http3_runtime.zig
+++ b/src/http/http3_runtime.zig
@@ -206,6 +206,25 @@ const ConnEntry = struct {
     }
 };
 
+/// The Retry token key ring and the process stateless-reset key are the
+/// only long-lived secrets `Runtime` owns. Grouping them here makes their
+/// lifecycle (installed at `init`, wiped on teardown or a later init
+/// failure) independently testable via `deinit` — including calling the
+/// exact cleanup the production owner uses — without fighting
+/// `Runtime.deinit`'s own trailing `self.* = undefined`, which
+/// safety-checked builds use to poison-fill the *entire* enclosing struct
+/// and would otherwise make any post-deinit byte-level assertion on these
+/// fields meaningless.
+const RuntimeSecrets = struct {
+    retry_tokens: quic.path.RetryTokens = .{},
+    stateless_reset_key: [32]u8 = [_]u8{0} ** 32,
+
+    fn deinit(self: *RuntimeSecrets) void {
+        self.retry_tokens.keys.deinit();
+        crypto_pkg.secrets.secureZero(&self.stateless_reset_key);
+    }
+};
+
 pub const Runtime = struct {
     allocator: std.mem.Allocator,
     socket_fd: std.c.fd_t,
@@ -228,8 +247,7 @@ pub const Runtime = struct {
     /// early-data policy on new connections' TLS backends.
     zero_rtt_enabled: bool,
     retry_policy: quic.config.RetryPolicy,
-    retry_tokens: quic.path.RetryTokens,
-    stateless_reset_key: [32]u8,
+    secrets: RuntimeSecrets,
     quic_config: quic.config.Config,
     h3_settings: http3.frame.Settings,
     h3_application_compat: [http3.early_data.encoded_snapshot_len]u8 = undefined,
@@ -302,8 +320,7 @@ pub const Runtime = struct {
             .early_data_replay_gate = cfg.early_data_replay_gate,
             .zero_rtt_enabled = zeroRttCarrierEnabled(cfg),
             .retry_policy = cfg.retry_policy,
-            .retry_tokens = .{},
-            .stateless_reset_key = undefined,
+            .secrets = .{},
             .quic_config = quicConfigFrom(cfg),
             .h3_settings = cfg.h3_settings,
             .early_data_compat_metrics_ctx = cfg.early_data_compat_metrics_ctx,
@@ -321,10 +338,10 @@ pub const Runtime = struct {
         runtime.crypto_provider_state = tls_core.production_crypto.Provider.init(runtime.crypto_provider_entropy.entropy());
         var retry_key: [quic.path.token_key_len]u8 = undefined;
         compat.randomBytes(&retry_key);
-        runtime.retry_tokens.keys.install(0, retry_key);
+        runtime.secrets.retry_tokens.keys.install(0, retry_key);
         crypto_pkg.secrets.secureZero(&retry_key);
-        compat.randomBytes(&runtime.stateless_reset_key);
-        errdefer runtime.wipeSecrets();
+        compat.randomBytes(&runtime.secrets.stateless_reset_key);
+        errdefer runtime.secrets.deinit();
         http3.frame.validateLocallySupportedSettings(runtime.h3_settings) catch return error.InvalidH3Settings;
         runtime.h3_application_compat_len = (http3.early_data.encodeSettingsSnapshot(
             runtime.h3_settings,
@@ -356,20 +373,8 @@ pub const Runtime = struct {
         self.stopping.store(true, .release);
         if (self.thread) |thread| thread.join();
         _ = std.c.close(self.socket_fd);
-        self.wipeSecrets();
+        self.secrets.deinit();
         self.* = undefined;
-    }
-
-    /// Wipe the Retry token key ring and the stateless-reset key. Factored
-    /// out of `deinit` (rather than inlined before its trailing
-    /// `self.* = undefined`) so it stays independently callable — including
-    /// from tests, which would otherwise be unable to observe the wipe at
-    /// all: safety-checked builds poison-fill a whole struct on `= undefined`,
-    /// which would overwrite the very zero bytes a post-`deinit` byte-level
-    /// assertion is trying to verify.
-    fn wipeSecrets(self: *Runtime) void {
-        self.retry_tokens.keys.deinit();
-        crypto_pkg.secrets.secureZero(&self.stateless_reset_key);
     }
 
     pub fn snapshot(self: *Runtime) Snapshot {
@@ -689,7 +694,7 @@ pub const Runtime = struct {
             .now_us = now,
             .initial_path = .{ .local = self.local_address, .remote = addressFromSockaddrIn(peer) },
             .initial_address_validated = retry_context != null,
-            .stateless_reset_key = self.stateless_reset_key,
+            .stateless_reset_key = self.secrets.stateless_reset_key,
             .events = .{ .context = self, .emitFn = quicConnectionEvent },
         }) catch {
             allocator.destroy(backend);
@@ -767,7 +772,7 @@ pub const Runtime = struct {
             self.issueRetry(routes, parsed, peer, remote, now);
             return .drop;
         }
-        const ctx = self.retry_tokens.validateRetry(parsed.token, remote, now) catch {
+        const ctx = self.secrets.retry_tokens.validateRetry(parsed.token, remote, now) catch {
             self.noteInvalidToken();
             return .drop;
         };
@@ -790,7 +795,7 @@ pub const Runtime = struct {
         var nonce: [quic.path.token_nonce_len]u8 = undefined;
         compat.randomBytes(&nonce);
         var token_buf: [quic.path.max_token_len]u8 = undefined;
-        const token = self.retry_tokens.issueRetry(parsed.dcid, retry_scid.slice(), parsed.version, remote, now, nonce, &token_buf) catch return;
+        const token = self.secrets.retry_tokens.issueRetry(parsed.dcid, retry_scid.slice(), parsed.version, remote, now, nonce, &token_buf) catch return;
         var retry_buf: [512]u8 = undefined;
         const retry = quic.packet.writeRetryV1(parsed.dcid, parsed.scid, retry_scid.slice(), token, &retry_buf) catch return;
         self.sendDatagram(peer, retry);
@@ -2060,27 +2065,29 @@ test "runtime init rejects unsupported local H3 setting max_field_section_size" 
     try expectInvalidH3SettingsAtRuntimeInit(.{ .max_field_section_size = 1024 });
 }
 
-test "Runtime.deinit wipes the retry token key ring and the stateless reset key" {
+test "RuntimeSecrets.deinit wipes the retry token key ring and the stateless reset key" {
     var logger = logger_mod.Logger.init(.err, "http3-runtime-deinit-wipe-test");
     var runtime = try Runtime.init(testing.allocator, &logger, .{
         .listen_host = "127.0.0.1",
         .quic_port = 0,
     });
-    // `deinit()` ends with `self.* = undefined`, and safety-checked builds
-    // poison-fill the *whole* struct on that assignment — which would
-    // overwrite our zeroed bytes with poison before a post-`deinit` byte
-    // check ever ran, making the wipe unobservable either way. Call
-    // `wipeSecrets()` (the exact cleanup `deinit()` delegates to) directly
-    // instead, on this real `Runtime.init`-produced key material, and close
-    // the socket by hand since we're bypassing `deinit()`.
+    // Exercise `RuntimeSecrets.deinit()` directly — the exact cleanup
+    // `Runtime.deinit()` delegates to via `self.secrets.deinit()` — rather
+    // than the full `Runtime.deinit()`, whose trailing `self.* = undefined`
+    // poison-fills the *whole* struct in safety-checked builds. That would
+    // overwrite these zeroed bytes with poison before any post-deinit byte
+    // check ran, making the wipe unobservable regardless of whether it
+    // happened. `RuntimeSecrets` has no such trailing assignment, so its
+    // fields stay genuinely inspectable after `deinit()`. Close the socket
+    // by hand since we're bypassing `Runtime.deinit()`.
     defer _ = std.c.close(runtime.socket_fd);
 
     // Capture raw pointers into the still-installed key storage before the
     // wipe: asserting only that lookups fail afterward would also pass an
-    // implementation that just cleared option tags without ever touching
-    // the key bytes.
-    const key0_ptr: *const [quic.path.token_key_len]u8 = &(runtime.retry_tokens.keys.keys[0].?);
-    const reset_key_ptr = &runtime.stateless_reset_key;
+    // implementation that never touched the key bytes, only an occupancy
+    // flag/length.
+    const key0_ptr: *const [quic.path.token_key_len]u8 = &runtime.secrets.retry_tokens.keys.keys[0].bytes;
+    const reset_key_ptr = &runtime.secrets.stateless_reset_key;
 
     var saw_key_nonzero = false;
     for (key0_ptr) |b| {
@@ -2093,7 +2100,7 @@ test "Runtime.deinit wipes the retry token key ring and the stateless reset key"
     }
     try testing.expect(saw_reset_nonzero);
 
-    runtime.wipeSecrets();
+    runtime.secrets.deinit();
 
     for (key0_ptr) |byte| try testing.expectEqual(@as(u8, 0), byte);
     for (reset_key_ptr) |byte| try testing.expectEqual(@as(u8, 0), byte);
@@ -4119,7 +4126,7 @@ test "http3 runtime Retry accepts validated tokens with split CID roles and vali
     const client_scid = [_]u8{0x22} ** 8;
     const peer = testPeerSockaddr(44_332);
     var token_buf: [quic.path.max_token_len]u8 = undefined;
-    const token = try runtime.retry_tokens.issueRetry(
+    const token = try runtime.secrets.retry_tokens.issueRetry(
         &odcid,
         &retry_scid,
         quic.packet.quic_v1,
@@ -4181,7 +4188,7 @@ test "http3 runtime Retry rejects valid tokens replayed with the wrong Retry SCI
     const client_scid = [_]u8{0x22} ** 8;
     const peer = testPeerSockaddr(44_333);
     var token_buf: [quic.path.max_token_len]u8 = undefined;
-    const token = try runtime.retry_tokens.issueRetry(
+    const token = try runtime.secrets.retry_tokens.issueRetry(
         &odcid,
         &retry_scid,
         quic.packet.quic_v1,
@@ -4235,7 +4242,7 @@ test "http3 runtime Retry counts valid tokens replayed with the wrong QUIC versi
     const client_scid = [_]u8{0x22} ** 8;
     const peer = testPeerSockaddr(44_334);
     var token_buf: [quic.path.max_token_len]u8 = undefined;
-    const token = try runtime.retry_tokens.issueRetry(
+    const token = try runtime.secrets.retry_tokens.issueRetry(
         &odcid,
         &retry_scid,
         quic.packet.quic_v1,
--- a/src/quic/cid.zig
+++ b/src/quic/cid.zig
@@ -654,9 +654,50 @@ test "local registry deinit clears issued entries" {
     _ = try registry.issue(&entropy, 8);
     try testing.expect(registry.activeCount() > 0);
 
+    // Capture raw pointers into the still-populated storage before deinit:
+    // `activeCount() == 0` and `reset_token_key.len == 0` alone would also
+    // pass an implementation that only flips the `?Entry` option tags
+    // without ever touching the token/key payload bytes.
+    var token_ptrs: [max_local_active_cids]?*const [stateless_reset_token_len]u8 = .{null} ** max_local_active_cids;
+    for (&registry.entries, 0..) |*slot, i| {
+        if (slot.*) |*entry| token_ptrs[i] = &entry.stateless_reset_token;
+    }
+    try testing.expect(token_ptrs[0] != null);
+    const key_ptr = &registry.reset_token_key.bytes;
+
     registry.deinit();
     try testing.expectEqual(@as(usize, 0), registry.activeCount());
     try testing.expectEqual(@as(usize, 0), registry.reset_token_key.len);
+    for (token_ptrs) |maybe_ptr| {
+        const ptr = maybe_ptr orelse continue;
+        for (ptr) |byte| try testing.expectEqual(@as(u8, 0), byte);
+    }
+    for (key_ptr) |byte| try testing.expectEqual(@as(u8, 0), byte);
+}
+
+test "retire wipes the stateless reset token bytes, not just the entry slot" {
+    var registry = LocalCidRegistry.init(4, [_]u8{0x66} ** 32);
+    defer registry.deinit();
+    _ = try registry.registerInitial(try ConnectionId.init(&.{ 9, 9, 9, 9 }));
+    var entropy = [_]u8{0x77} ** 8;
+    const issued = try registry.issue(&entropy, 8);
+    try registry.markAdvertised(issued.sequence);
+
+    var token_ptr: ?*const [stateless_reset_token_len]u8 = null;
+    for (&registry.entries) |*slot| {
+        if (slot.*) |*entry| {
+            if (entry.sequence == issued.sequence) token_ptr = &entry.stateless_reset_token;
+        }
+    }
+    const token = token_ptr orelse return error.TestUnexpectedResult;
+    var saw_nonzero = false;
+    for (token) |byte| {
+        if (byte != 0) saw_nonzero = true;
+    }
+    try testing.expect(saw_nonzero);
+
+    _ = try registry.retire(.{ .sequence = issued.sequence });
+    for (token) |byte| try testing.expectEqual(@as(u8, 0), byte);
 }
 
 test "peer pool validates NEW_CONNECTION_ID and sweeps retire_prior_to" {
--- a/src/quic/cid.zig
+++ b/src/quic/cid.zig
@@ -190,12 +190,15 @@ pub const LocalCidRegistry = struct {
     }
 
     /// Register the CID chosen during the handshake as sequence 0
-    /// (RFC 9000 §5.1.1). Call once, before any `issue`.
-    pub fn registerInitial(self: *LocalCidRegistry, cid: ConnectionId) error{ CidLimitExceeded, DuplicateCid }!Entry {
+    /// (RFC 9000 §5.1.1). Call once, before any `issue`. Returns only the
+    /// (public, wire-visible) CID rather than the full `Entry`, so the
+    /// registry's copy of the derived reset token is never duplicated onto
+    /// the caller's stack.
+    pub fn registerInitial(self: *LocalCidRegistry, cid: ConnectionId) error{ CidLimitExceeded, DuplicateCid }!ConnectionId {
         std.debug.assert(self.next_sequence == 0);
         const entry = try self.store(cid);
         self.largest_advertised_sequence = entry.sequence;
-        return entry;
+        return entry.cid;
     }
 
     /// Issue a fresh CID from caller-supplied entropy and return the
@@ -207,13 +210,16 @@ pub const LocalCidRegistry = struct {
         return try self.issueCid(cid);
     }
 
-    /// Issue a caller-generated CID. Production endpoints use this
-    /// transactional form so a route table entry can be reserved before the
-    /// exact CID is queued in a NEW_CONNECTION_ID frame.
-    pub fn issueCid(self: *LocalCidRegistry, cid: ConnectionId) error{ CidLimitExceeded, DuplicateCid }!NewConnectionIdFrame {
+    /// Issue a caller-generated CID, writing the NEW_CONNECTION_ID frame
+    /// model directly into `out`. Production endpoints use this
+    /// destination-oriented form so the reset token is copied exactly once,
+    /// from the registry's owned entry into the caller's owned destination
+    /// (e.g. a queue slot), instead of round-tripping through an
+    /// intermediate by-value return.
+    pub fn issueCidInto(self: *LocalCidRegistry, cid: ConnectionId, out: *NewConnectionIdFrame) error{ CidLimitExceeded, DuplicateCid }!void {
         const entry = try self.store(cid);
         self.metrics.cids_issued += 1;
-        return .{
+        out.* = .{
             .sequence = entry.sequence,
             .retire_prior_to = 0,
             .cid = entry.cid,
@@ -221,7 +227,18 @@ pub const LocalCidRegistry = struct {
         };
     }
 
-    fn store(self: *LocalCidRegistry, cid: ConnectionId) error{ CidLimitExceeded, DuplicateCid }!Entry {
+    /// Convenience wrapper over `issueCidInto` for callers (mostly tests)
+    /// that want an owned return value rather than a destination pointer.
+    pub fn issueCid(self: *LocalCidRegistry, cid: ConnectionId) error{ CidLimitExceeded, DuplicateCid }!NewConnectionIdFrame {
+        var out: NewConnectionIdFrame = undefined;
+        try self.issueCidInto(cid, &out);
+        return out;
+    }
+
+    /// Reserve and populate a fresh entry for `cid`, returning a pointer into
+    /// the registry's own storage. Callers must not retain the pointer past
+    /// their next mutating call into the registry.
+    fn store(self: *LocalCidRegistry, cid: ConnectionId) error{ CidLimitExceeded, DuplicateCid }!*Entry {
         if (self.activeCount() >= self.active_limit) return error.CidLimitExceeded;
         // Repeated caller entropy must not mint the same CID twice under
         // different sequence numbers.
@@ -231,12 +248,13 @@ pub const LocalCidRegistry = struct {
         const slot_idx = for (self.occupied, 0..) |occupied, i| {
             if (!occupied) break i;
         } else return error.CidLimitExceeded;
-        const entry = Entry{
+        const entry = &self.entries[slot_idx];
+        entry.* = .{
             .sequence = self.next_sequence,
             .cid = cid,
-            .stateless_reset_token = statelessResetToken(self.reset_token_key.slice(), cid.slice()),
+            .stateless_reset_token = [_]u8{0} ** stateless_reset_token_len,
         };
-        self.entries[slot_idx] = entry;
+        statelessResetTokenInto(&entry.stateless_reset_token, self.reset_token_key.slice(), cid.slice());
         self.occupied[slot_idx] = true;
         self.next_sequence += 1;
         return entry;
@@ -462,12 +480,20 @@ pub const stateless_reset_token_len = 16;
 /// HMAC-SHA256(static_key, connection_id); it MUST be hard to guess, so the
 /// static key must be secret and stable across the connection's lifetime.
 pub fn statelessResetToken(static_key: []const u8, connection_id: []const u8) [stateless_reset_token_len]u8 {
+    var token: [stateless_reset_token_len]u8 = undefined;
+    statelessResetTokenInto(&token, static_key, connection_id);
+    return token;
+}
+
+/// Same derivation as `statelessResetToken`, writing directly into `out`
+/// instead of returning by value. Production issuance paths use this so the
+/// token is written once into its final owner, rather than materializing an
+/// extra by-value stack copy that a caller then copies again.
+fn statelessResetTokenInto(out: *[stateless_reset_token_len]u8, static_key: []const u8, connection_id: []const u8) void {
     var mac: [HmacSha256.mac_length]u8 = undefined;
     defer secrets.secureZero(&mac);
     HmacSha256.create(&mac, connection_id, static_key);
-    var token: [stateless_reset_token_len]u8 = undefined;
-    @memcpy(&token, mac[0..stateless_reset_token_len]);
-    return token;
+    @memcpy(out, mac[0..stateless_reset_token_len]);
 }
 
 /// Emission rules and packet construction for stateless resets (RFC 9000 §10.3).
@@ -615,8 +641,8 @@ test "retired local CIDs stop routing and unknown retire sequences are violation
     var registry = LocalCidRegistry.init(4, [_]u8{0xab} ** 32);
 
     const initial = try ConnectionId.init(&.{ 0xaa, 0xaa, 0xaa, 0xaa, 0xaa, 0xaa, 0xaa, 0xaa });
-    const initial_entry = try registry.registerInitial(initial);
-    try table.insert(initial_entry.cid, 1);
+    const initial_cid = try registry.registerInitial(initial);
+    try table.insert(initial_cid, 1);
 
     var entropy = [_]u8{0x10} ** 8;
     const issued = try registry.issue(&entropy, 8);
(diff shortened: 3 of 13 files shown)


src/quic/path.zig
```diff
@@ -133,17 +134,30 @@ pub const RetryTokenKeyRing = struct {
...
 
     pub fn retire(self: *RetryTokenKeyRing, key_id: u8) void {
-        if (key_id < max_token_keys) self.keys[key_id] = null;
+        if (key_id >= max_token_keys) return;
+        if (self.keys[key_id]) |*slot| secrets.secureZero(slot);
+        self.keys[key_id] = null;
     }
 
-    fn get(self: *const RetryTokenKeyRing, key_id: u8) ?[token_key_len]u8 {
+    pub fn deinit(self: *RetryTokenKeyRing) void {
```
Jaruso, line 154: **[P1] This teardown is never invoked by the production owner.** `src/http/http3_runtime.zig::Runtime` owns `retry_tokens`, but `Runtime.deinit()` currently stops the thread, closes the socket, and assigns `self.* = undefined` without calling this method or wiping `Runtime.stateless_reset_key`. `Runtime.init()` also generates a stack-local `retry_key`, installs it, and can subsequently fail settings validation without cleaning either key.

Please wire the full owner lifecycle, for example:

```zig
var retry_key: [quic.path.token_key_len]u8 = undefined;
compat.randomBytes(&retry_key);
defer crypto_pkg.secrets.secureZero(&retry_key);
runtime.retry_tokens.keys.install(0, retry_key);
errdefer runtime.retry_tokens.keys.deinit();
errdefer crypto_pkg.secrets.secureZero(&runtime.stateless_reset_key);
```

and after the listener thread has joined:

```zig
self.retry_tokens.keys.deinit();
crypto_pkg.secrets.secureZero(&self.stateless_reset_key);
self.* = undefined;
```

As written, `deinit()` is exercised only by the new unit test, while the real process-lifetime secrets still have no cleanup path.

src/tls/identity_loader.zig
```diff
@@ -50,6 +63,7 @@ pub fn readSmallFile(allocator: std.mem.Allocator, path: []const u8) ![]u8 {
     var out: std.ArrayList(u8) = .empty;
     errdefer out.deinit(allocator);
     var buf: [4096]u8 = undefined;
+    defer secrets.secureZero(&buf);
```
Jaruso: **[P1] The stack scratch is wiped, but the dynamic accumulator directly above it is still freed unwiped on every error.** If `read()` fails after one or more successful chunks, the file exceeds the limit, or `appendSlice()` runs out of memory, `out.items` contains private-key file bytes and `errdefer out.deinit(allocator)` releases them unchanged.

Please wipe the populated allocation before deinitializing it:

```zig
var out: std.ArrayList(u8) = .empty;
errdefer {
    secrets.secureZero(out.items);
    out.deinit(allocator);
}
```

Add a fault-injected read/allocator test that fails after at least one appended key chunk and inspects the released storage. The current happy-path changes do not cover the allocator/error cleanup criterion from #375.

Jaruso (commented):
I reviewed all five changed files against the concrete secret-lifetime requirements in #375 and traced the new owners through the QUIC/HTTP3 composition root. CI is green, and the direction is correct, but this should not merge yet.

The blocking gaps are:

1. `RetryTokenKeyRing.deinit()` is not called by its production owner, so the HTTP/3 runtime still discards the installed Retry key and process stateless-reset key with `self.* = undefined`; init failures after key generation also retain those bytes.
2. `readSmallFile()` still frees its partially populated heap accumulator without wiping it on read, size-limit, or allocation errors.
3. Locally generated stateless-reset tokens are copied into `pending_new_connection_ids` and `sent_records`; those copies are removed/freed without wiping. The connection also retains a raw reset-key copy after constructing the registry, and construction currently copies the `FixedSecret` through a stack-local registry value.
4. Wiping only the current ticket ledger allocation does not cover old `ArrayList` backing allocations freed during growth, so prior secret-derived fingerprints can remain in allocator memory.
5. The two ledger fingerprint comparisons remain ordinary `std.mem.eql` despite #375 explicitly classifying them as secret-derived and requiring the canonical constant-time helper.
6. `zeroAndFree()` still contains a second local volatile-zero primitive and bypasses normal allocator ownership with `rawFree`; this does not satisfy the issue's removal of ad hoc zeroization.
7. The new QUIC cleanup tests mostly prove logical null/length state and would pass implementations that never erased the underlying bytes. They need byte-level replacement, retirement, failure, and final-owner teardown coverage.

I left concrete fixes and test shapes inline. The full #375 audit matrix/checklist can remain for later PRs because this PR only references the issue, but each lifecycle remediation claimed by this slice needs to be complete and regression-tested before merge.

src/quic/connection.zig
```diff
@@ -959,6 +959,9 @@ pub const Connection = struct {
         self.pending_path_responses.deinit(self.allocator);
         self.candidate_challenges.deinit(self.allocator);
         self.retry_token.deinit(self.allocator);
+        if (self.local_cids) |*registry| registry.deinit();
```
Jaruso, line 1064: **[P1] This only wipes the registry's primary copies; the connection owns several additional copies of the same reset material.** Every issued `NewConnectionIdFrame` copies the token into `pending_new_connection_ids`, then into `SentRecord.carried_new_connection_id`, and loss requeue copies it back again. ACK/loss removal and `deinitPartial()` currently `swapRemove`/`orderedRemove`/free those arrays without scrubbing the token bytes. A peer can retire a CID while an unacknowledged frame copy still exists, so `registry.retire()` does not close the lifetime.

There is a second ownership problem when the registry is created: `self.stateless_reset_key` remains live even though it is no longer used, and `var registry = LocalCidRegistry.init(...); self.local_cids = registry;` copies the `FixedSecret` through a stack-local value that is never deinitialized.

Please initialize the optional owner in place and transfer/wipe the raw key immediately, e.g.:

```zig
self.local_cids = LocalCidRegistry.init(limit, self.stateless_reset_key);
const registry = &self.local_cids.?;
errdefer {
    registry.deinit();
    self.local_cids = null;
}
crypto_secrets.secureZero(&self.stateless_reset_key);
```

Then either redesign queued/sent CID frames to borrow/derive the token from the registry, or explicitly scrub every frame copy on ACK, loss, retirement, queue removal, and teardown (including stale capacity left by array removals). Add a connection-level test covering a pending and an in-flight `NEW_CONNECTION_ID` during teardown/retirement.

src/tls/ticket_protection.zig
```diff
@@ -439,6 +444,7 @@ pub const ReloadableKeyRing = struct {
         self.mutex.unlock();
         if (retired) |snapshot| snapshot.release();
         var mutable_ledger = ledger;
+        for (mutable_ledger.items) |*entry| entry.deinit();
```
Jaruso: **[P1] Wiping the final ledger allocation does not wipe allocations retired during `ArrayList` growth.** `validateReplacementLocked()` calls `ledger.ensureUnusedCapacity()`. When that grows the list, the allocator copies the `LeaseHighWater` values and frees the old backing allocation without calling `LeaseHighWater.deinit()`, leaving every old `key_fingerprint` in freed memory. `appendAssumeCapacity(LeaseHighWater.init(key))` can also create a transient by-value fingerprint copy.

Use storage whose secret-bearing elements have stable ownership, or implement a secure grow operation. A straightforward shape is an array list of pointers:

```zig
ledger: std.ArrayList(*LeaseHighWater) = .empty,

const entry = try self.allocator.create(LeaseHighWater);
errdefer self.allocator.destroy(entry);
entry.* = LeaseHighWater.init(key);
try self.ledger.append(self.allocator, entry);

// deinit
for (self.ledger.items) |entry| {
    entry.deinit();
    self.allocator.destroy(entry);
}
self.ledger.deinit(self.allocator);
```

A custom contiguous container is also fine, but it must wipe the old backing bytes before freeing on every reallocation. Please add a test that forces multiple ledger growth events and verifies the allocator sees zeroed retired allocations, not only the final one.

src/tls/ticket_protection.zig
```diff
@@ -554,7 +560,8 @@ pub const ReloadableKeyRing = struct {
                 if (prior.aead == key.aead and prior.key.eql(&key.key)) return error.DuplicateKeyId;
             }
 
-            const key_fingerprint = fingerprintKey(key.aead, key.key.slice());
+            var key_fingerprint = fingerprintKey(key.aead, key.key.slice());
+            defer secrets.secureZero(&key_fingerprint);
```
Jaruso, line 606: **[P1] The temporary fingerprint is now wiped, but both comparisons immediately below still use ordinary equality on secret-derived data.** #375 explicitly calls out `key_fingerprint` and requires the canonical constant-time helper; only public `key_id` routing should remain ordinary.

Please change both ledger comparisons to:

```zig
if (entry.aead != key.aead or
    !secrets.constantTimeEqual(&entry.key_fingerprint, &key_fingerprint))
    return error.DuplicateKeyId;
```

and:

```zig
if (entry.aead == key.aead and
    secrets.constantTimeEqual(&entry.key_fingerprint, &key_fingerprint))
    return error.DuplicateKeyId;
```

Add equal and first/middle/last-byte mismatch coverage through the shared helper or the replacement-validation path.

src/tls/ticket_protection.zig
```diff
@@ -893,6 +900,10 @@ fn checkedProtectedLen(plaintext_len: usize, limits: session.Limits) SealError!u
 }
 
 fn zeroAndFree(allocator: std.mem.Allocator, buffer: []u8) void {
+    if (buffer.len == 0) return;
+    secrets.secureZero(buffer);
```
Jaruso: **[P2] This still retains the ad hoc zeroization primitive that #375 requires us to remove.** The canonical `secrets.secureZero()` is immediately followed by a second hand-written volatile loop, a debug scan, and `allocator.rawFree()`. That duplicates the reviewed primitive and bypasses ordinary allocator ownership.

The file's `ZeroCheckingAllocator.free()` already inspects the bytes before delegating to its child, so the local path should be:

```zig
fn zeroAndFree(allocator: std.mem.Allocator, buffer: []u8) void {
    if (buffer.len == 0) return;
    secrets.secureZero(buffer);
    allocator.free(buffer);
}
```

If Zig's typed `Allocator.free` semantics make that impossible for a supported allocator, put the exceptional raw-free behavior in one reviewed canonical `crypto.secrets` helper with focused allocator tests and use that helper here. It should not remain a ticket-specific primitive.

src/quic/path.zig
```diff
@@ -1096,6 +1110,18 @@ test "retry token survives key rotation while a key is retained" {
     try testing.expectError(error.UnknownTokenKey, tokens.validateRetry(token, loopbackV4(4433), 1_000_000));
 }
 
+test "retry token key ring deinit clears installed keys" {
```
Jaruso: **[P1 test gap] This test proves only that the optional tags were set to `null`; the pre-change implementation (`keys[key_id] = null`) would pass the same assertions while leaving the payload bytes untouched.** Please inspect the backing bytes and cover all required transitions:

- replacement wipes the old key before the new key becomes current;
- `retire()` wipes the payload before nulling it;
- `deinit()` wipes every installed slot;
- the real `http3_runtime.Runtime.deinit()` invokes ring teardown;
- init failure after installation cleans the key.

A test-only raw-slot accessor or a `FixedSecret`-backed slot makes these assertions much less fragile than scanning only the logical API state.

src/quic/cid.zig
```diff
@@ -628,6 +647,18 @@ test "local registry enforces the peer's active CID limit and derives reset toke
     try testing.expectEqual(@as(usize, 2), registry.activeCount());
 }
 
+test "local registry deinit clears issued entries" {
```
Jaruso, line 703: **[P1 test gap] `activeCount() == 0` and `reset_token_key.len == 0` do not prove that either the key bytes or token bytes were erased.** A broken implementation that only clears option tags and sets `len = 0` would pass.

Please assert the actual `reset_token_key.bytes` are zero, capture/inspect the entry-token storage across `retire()`, and add a connection-level teardown case with issued tokens in the pending and sent-frame copies. The regression test should fail if any old key/token pattern remains in the owner or its queued/retransmission state.

src/tls/ticket_protection.zig
```diff
@@ -594,18 +609,25 @@ pub const ReloadableKeyRing = struct {
...
+                    entry.prefix = lease.prefix;
+                    entry.end_exclusive = @max(entry.end_exclusive, lease.currentEnd());
                 }
             } else {
-                self.ledger.appendAssumeCapacity(LeaseHighWater.init(key));
+                // `validateReplacementLocked` already reserved pointer-array
+                // capacity for every entry counted here, so only the
+                // individual entry allocation can fail; skip on OOM rather
+                // than leaving the ledger inconsistent post-commit.
+                const entry = self.allocator.create(LeaseHighWater) catch continue;
```
Jaruso: **[P1] This turns allocation failure into a successful install with incomplete security state.** `install()` has already assigned `self.current = replacement` before calling `updateLedgerLocked()`. If `create(LeaseHighWater)` fails here, `continue` leaves the new key active but omits its durable fingerprint and nonce-lease high-water mark. A later reload can then reuse that key ID with different key material or reuse/roll back a nonce range because the ledger has no record of the committed key.

Allocate every missing entry transactionally **before** mutating `self.current` or existing ledger entries. One safe shape is:

```zig
var additions: std.ArrayList(*LeaseHighWater) = .empty;
errdefer {
    for (additions.items) |entry| {
        entry.deinit();
        self.allocator.destroy(entry);
    }
    additions.deinit(self.allocator);
}

// Allocate and initialize every missing entry first; any OOM rejects install.
// Ensure self.ledger capacity next. Only after all fallible work succeeds,
// commit self.current, update existing high-water values, and append additions.
```

After the commit point, ledger publication should be infallible. Also, `entry.* = LeaseHighWater.init(key)` still recreates the transient by-value fingerprint copy called out in the first review; initialize through a local that is wiped with `defer`, or provide an `initInto(entry, key)` path whose temporary fingerprint is explicitly destroyed.

src/quic/connection.zig
```diff
@@ -950,15 +987,24 @@ pub const Connection = struct {
         self.stream_transport_early.deinit();
         self.accept_queue.deinit(self.allocator);
         for (&self.crypto_tx) |*tx| tx.deinit(self.allocator);
+        // Any still-outstanding (unacked/unlost) or still-queued
+        // NEW_CONNECTION_ID carries a stateless-reset token copy; teardown
+        // must scrub those before the backing allocations are freed, not
+        // just the registry's own copy below.
+        wipeSentRecordTokens(self.sent_records.items);
```
Jaruso, line 1053: **[P1] Final teardown wiping does not cover `ArrayList` growth.** Both `sent_records.append()` and `pending_new_connection_ids.append()` can reallocate: Zig copies every live element into a new backing allocation and frees the old allocation without invoking these element-level wipe helpers. Any reset token present during that growth remains in the retired allocation even though the final live array is scrubbed here.

These collections are already bounded by protocol state: `sent_records` cannot exceed `recovery.max_tracked_packets` (128), and locally active/queued CIDs are bounded by `quic_cid.max_local_active_cids` (8). Reserve the full capacity before the first secret-bearing element can be inserted, or replace them with bounded/stable storage or a secure-grow container that wipes the old backing allocation before raw free. For example, perform fallible capacity reservation during connection/registry setup, before any token is queued:

```zig
try self.sent_records.ensureTotalCapacity(self.allocator, recovery.max_tracked_packets);
try self.pending_new_connection_ids.ensureTotalCapacity(
    self.allocator,
    quic_cid.max_local_active_cids,
);
```

Add an allocator-backed regression test that forces the old growth pattern and verifies no retired allocation retains a token pattern.

src/quic/connection.zig
```diff
@@ -720,6 +720,43 @@ const SentRecord = struct {
...
+/// `std.ArrayList.swapRemove` moves the last live element into the removed
+/// slot but only assigns `undefined` (a no-op at runtime, not a real write)
+/// to the vacated tail slot, so a byte-for-byte duplicate of whatever
+/// occupied that tail position survives at `items.ptr[items.len]` — within
+/// the backing allocation, just past the shrunk `.items` slice — until the
+/// allocator reuses it. Call this immediately after every
+/// `sent_records.swapRemove(...)` to scrub that residue.
+fn wipeSentRecordsSwapRemoveResidue(records: *std.ArrayList(SentRecord)) void {
+    wipeSentRecordToken(&records.allocatedSlice()[records.items.len]);
+}
```
Jaruso, line 784: **[P1] This reads an uninitialized `SentRecord` outside the list's live range.** After `swapRemove`, `allocatedSlice()[items.len]` is the vacated capacity slot, not an initialized element in `.items`. Interpreting it as a `SentRecord` and pattern-matching `carried_new_connection_id` reads an undefined optional tag. That is the same class of invalid-storage assumption causing the current Linux CI failures.

Wipe the live element's token **before** removal, then clear the vacated tail as raw bytes without interpreting any field:

```zig
wipeSentRecordToken(&records.items[index]);
_ = records.swapRemove(index);
crypto_secrets.secureZero(
    std.mem.asBytes(&records.allocatedSlice()[records.items.len]),
);
```

For the loss path, requeue from the live element first, then wipe it, remove it, and zero the raw tail. Apply the same rule to `wipePendingNewConnectionIdsOrderedRemoveResidue`: zero the whole vacated frame slot as bytes rather than selecting a field from an undefined `NewConnectionIdFrame`.

src/quic/connection.zig
```diff
@@ -1601,6 +1647,7 @@ pub const Connection = struct {
             }
             self.onRecordAcked(record);
             _ = self.sent_records.swapRemove(index);
+            wipeSentRecordsSwapRemoveResidue(&self.sent_records);
```
Jaruso, line 1737: **[P1] The tail wipe still leaves the by-value source copies untouched.** Immediately above this line, `const record = self.sent_records.items[index]` copies the entire `SentRecord`, including any reset token, onto the stack; `onRecordAcked(record)` takes another by-value copy. `detectAndRequeueLost()`/`requeueRecord()` repeat the pattern, and packet construction keeps a local `record` plus a local `ncid` copy after the owned queue/record copies are created.

Process stored records by pointer and wipe the live source before removal:

```zig
const record = &self.sent_records.items[index];
self.onRecordAcked(record); // change helper to borrow
wipeSentRecordToken(record);
_ = self.sent_records.swapRemove(index);
// zero raw vacated tail
```

For packet construction, arm `defer wipeSentRecordToken(&record)` as soon as the local record is created; the appended recovery copy remains owned, while the temporary is destroyed on every return path. Avoid the local `const ncid = ...` by encoding from the queued element by pointer, or explicitly wipe that local after copying it into the record.

Jaruso (commented):
Re-reviewed the complete PR at `da46802c7c077448ced27518e7c972b8dc80913f` against #375 and every finding from the first review.

The revision correctly addresses several earlier blockers: the production HTTP/3 runtime now invokes Retry/stateless-reset cleanup on teardown and init failure, `readSmallFile` wipes its dynamic accumulator on failure, ticket fingerprints use the canonical constant-time helper, and ticket plaintext cleanup is routed through a canonical helper.

**This remains blocking despite being submitted as a COMMENT review:** GitHub does not allow the PR author to formally request changes on their own pull request.

Changes are still required before merge:

1. Ticket-key installation can now commit a replacement while silently dropping a required ledger entry on OOM. That loses the key fingerprint and nonce high-water security state.
2. `sent_records` and `pending_new_connection_ids` remain ordinary growable `ArrayList`s. Growth copies reset tokens into a new allocation and frees the prior allocation unwiped; final teardown and removal-residue cleanup do not cover this.
3. The new `swapRemove` residue helper interprets an element outside the initialized `.items` range, including reading an undefined optional tag.
4. Several by-value `SentRecord`/`NewConnectionIdFrame` and `LocalCidRegistry` construction paths still leave temporary reset-token/key copies on the stack.
5. The new byte-level tests dereference pointers into optional payloads after those optionals are set to `null`. Those pointers no longer designate active values.

The CI failure is deterministic and caused by item 5, not by cache or an unrelated base-branch regression. Both `Test (ubuntu-latest)` and `Test appliance profile (ubuntu-latest)` fail the same five tests. Linux x64 Debug poison-fills the inactive payloads with `0xAA`, producing `expected 0, found 170`; macOS and ARM passing is platform-dependent behavior from reading invalid/inactive storage, not evidence the tests are sound. The failing tests are:

- `cid.test.local registry deinit clears issued entries`
- `cid.test.retire wipes the stateless reset token bytes, not just the entry slot`
- `path.test.retry token key ring deinit clears installed keys down to the byte level`
- `path.test.retry token key ring retire wipes the retired key, not just its slot tag`
- `http.http3_runtime.test.Runtime.deinit wipes the retry token key ring and the stateless reset key`

Please fix the production ownership issues and replace the stale-pointer assertions with an observable, always-live storage seam or instrumented cleanup path. Then rerun both the normal and appliance test commands on Linux x64.

src/quic/cid.zig
```diff
@@ -145,19 +146,33 @@ pub const LocalCidRegistry = struct {
...
     /// Static secret for stateless-reset token derivation (RFC 9000 §10.3.1).
-    reset_token_key: [32]u8,
+    reset_token_key: secrets.FixedSecret(32),
     entries: [max_local_active_cids]?Entry = [_]?Entry{null} ** max_local_active_cids,
     next_sequence: u64 = 0,
     largest_advertised_sequence: ?u64 = null,
     metrics: Metrics = .{},
 
     pub fn init(peer_active_connection_id_limit: u64, reset_token_key: [32]u8) LocalCidRegistry {
+        var key = secrets.FixedSecret(32){};
```
Jaruso: **[P1] The registry is still constructed through by-value secret copies.** Moving the call expression directly into `self.local_cids = ...` does not make this API an in-place initializer: the raw `[32]u8` parameter is passed by value, this local `FixedSecret` owns another copy, and the function returns a `LocalCidRegistry` containing the secret by value. The audit cannot rely on result-location optimization to erase those semantic copies.

Use a borrowed input and explicit destination, for example:

```zig
pub fn initInto(
    out: *LocalCidRegistry,
    peer_limit: u64,
    reset_key: []const u8,
) void {
    out.* = .{
        .active_limit = @min(peer_limit, max_local_active_cids),
        .reset_token_key = .{},
    };
    out.reset_token_key.replace(reset_key) catch unreachable;
}
```

Initialize the optional payload through that destination, and wipe `Connection.stateless_reset_key` only after `initInto` succeeds. This removes the local secret owner and the raw key's by-value parameter copy.

src/quic/path.zig
```diff
@@ -1096,6 +1110,55 @@ test "retry token survives key rotation while a key is retained" {
...
+    // deinit: asserting only `get(id) == null` afterward would also pass an
+    // implementation that just flips the `?[N]u8` option tag without ever
+    // touching the key bytes.
+    const key0_ptr: *const [token_key_len]u8 = &(ring.keys[0].?);
+    const key1_ptr: *const [token_key_len]u8 = &(ring.keys[1].?);
+
+    ring.deinit();
+    try testing.expect(ring.get(0) == null);
+    try testing.expect(ring.get(1) == null);
+    for (key0_ptr) |byte| try testing.expectEqual(@as(u8, 0), byte);
```
Jaruso, line 1133: **[P1 / CI root cause] `key0_ptr` and `key1_ptr` no longer point to active values after `ring.deinit()` sets the optionals to `null`.** Dereferencing an inactive optional payload is invalid; Linux x64 Debug exposes this by poison-filling the payload with `0xAA`, which is why CI reports `expected 0, found 170`. The retire test has the same problem, as do the new CID and runtime tests. macOS/ARM passing is incidental platform behavior.

Do not inspect stale pointers after the optional tag transition. Either use an always-live storage representation, such as a `FixedSecret` plus an explicit `occupied` flag, or expose/instrument a cleanup helper that verifies the active payload immediately after `secureZero` and before setting the logical state to empty. The replacement test is different and valid because the slot remains active after replacement.

src/http/http3_runtime.zig
```diff
@@ -2045,6 +2060,45 @@ test "runtime init rejects unsupported local H3 setting max_field_section_size"
...
+    for (key0_ptr) |b| {
+        if (b != 0) saw_key_nonzero = true;
+    }
+    try testing.expect(saw_key_nonzero);
+    var saw_reset_nonzero = false;
+    for (reset_key_ptr) |b| {
+        if (b != 0) saw_reset_nonzero = true;
+    }
+    try testing.expect(saw_reset_nonzero);
+
```
Jaruso, line 2102: **[P2 test gap] This test never calls `Runtime.deinit()`.** It verifies `wipeSecrets()` in isolation, so a future regression that removes `self.wipeSecrets()` from `Runtime.deinit()` would still pass despite the test name and the original owner-lifecycle requirement. It also inherits the inactive-optional pointer problem noted in `path.zig`.

Make the secret state an observable sub-owner (for example, a `RuntimeSecrets` value whose `deinit` can be tested with always-live secret storage or a test hook), then call the real `Runtime.deinit()` and prove that owner teardown occurred. The test needs to cover both the cleanup primitive and the production delegation/wiring.

src/tls/identity_loader.zig
```diff
@@ -48,8 +61,12 @@ pub fn readSmallFile(allocator: std.mem.Allocator, path: []const u8) ![]u8 {
     defer _ = std.c.close(fd);
 
     var out: std.ArrayList(u8) = .empty;
-    errdefer out.deinit(allocator);
+    errdefer {
+        secrets.secureZero(out.items);
```
Jaruso: **[P1] This wipes only the ArrayList's current allocation, not prior allocations retired during growth or a moved allocation created by `toOwnedSlice()`.** A private-key file larger than the initial capacity can be copied through several backing allocations; `appendSlice()` may free each old allocation before this `errdefer` runs, and `toOwnedSlice()` may allocate/copy/free again on the success path. In ReleaseFast, the allocator's `undefined` fill is not guaranteed to overwrite those bytes, which is exactly why this PR added `secureZeroAndFree` elsewhere.

Please use a secret-aware growth/ownership-transfer path. For example, implement a small buffer owner whose grow operation allocates a new block, copies the live bytes, calls `secrets.secureZeroAndFree` on the old block, and whose final exact-size transfer does the same when an in-place remap is unavailable. A simpler regular-file path can `fstat`, reject sizes over 256 KiB, allocate the exact size, and read directly, with a bounded fallback for non-regular files.

The regression test must use an allocator that **forces moved growth** and separately force the success-path final move; the current `FixedBufferAllocator` can expand its last allocation in place, so it does not prove retired allocations were scrubbed.

src/quic/connection.zig
```diff
@@ -885,6 +926,20 @@ pub const Connection = struct {
...
+        // (every append is gated by a successful recovery-tracker insert,
+        // itself bounded there), and `pending_new_connection_ids` cannot
+        // exceed `quic_cid.max_local_active_cids` (the local CID registry's
+        // own issuance limit). With capacity reserved up front, later
+        // `append`/`ensureUnusedCapacity` calls never trigger `std.ArrayList`'s
+        // grow-and-free-the-old-backing-unwiped path, which would otherwise
+        // copy a live reset token into a new allocation and free the old one
+        // without scrubbing it.
+        try conn.sent_records.ensureTotalCapacityPrecise(allocator, recovery.max_tracked_packets);
+        try conn.pending_new_connection_ids.ensureTotalCapacityPrecise(allocator, quic_cid.max_local_active_cids);
```
Jaruso, line 975: **[P1] `pending_new_connection_ids` is not bounded by `max_local_active_cids`.** The registry bounds distinct active CIDs, but retransmission creates duplicate frame copies. `firePto()` calls `requeueRecord()` without removing the original sent record; calling PTO repeatedly before transmission/ACK appends the same `NEW_CONNECTION_ID` over and over. Multiple outstanding retransmissions can also be declared lost together and all requeue the same sequence. The ninth pending copy grows this ArrayList and frees the old token-bearing backing allocation unwiped.

Do not rely on capacity 8. Deduplicate queued frames by `sequence` before appending, or replace this with bounded/stable storage whose invariant is enforced. For example:

```zig
fn queueNewConnectionId(self: *Connection, source: *const quic_cid.NewConnectionIdFrame) !void {
    for (self.pending_new_connection_ids.items) |*queued| {
        if (queued.sequence == source.sequence) return;
    }
    const dst = try self.pending_new_connection_ids.addOne(self.allocator);
    dst.* = source.*;
}
```

Then route issuance, loss, and PTO requeue through that helper. Add a test that sends one CID frame, fires PTO more than `max_local_active_cids` times without polling, and proves the queue remains deduplicated and its backing pointer never changes.

Jaruso (commented):
Third-pass review on `85853ed` against the secret-lifetime portion of #375.

The latest commit correctly fixes the Linux x64 inactive-optional CI failure, makes the ticket ledger allocation transactional, uses constant-time fingerprint comparison, replaces typed reads of vacated ArrayList slots with raw-byte wipes, and wires the runtime secret sub-owner into production teardown. CI run 30707691320 is green.

This is still not merge-ready. Five secret-copy/lifetime gaps remain:

1. `readSmallFile` wipes only the current ArrayList allocation; growth and `toOwnedSlice` can retire earlier allocations containing private-key bytes without a secure wipe.
2. `pending_new_connection_ids` is not actually bounded by the number of active CIDs. PTO requeues an existing sent record without removing it, so repeated PTOs can enqueue duplicate copies and exceed the reserved capacity, re-enabling the unsafe growth path.
3. `RetryTokenKeyRing.install` still takes the secret key by value, leaving an unowned callee-stack copy after the caller wipes its own buffer.
4. The process stateless-reset key is copied by value through the `Connection.Options` temporary and the by-value `Connection.init` parameter into another raw-array owner. Those intermediate copies are never wiped.
5. The CID issuance/encoding chain still passes locally derived reset tokens through by-value `Entry`/`NewConnectionIdFrame` results and locals. The new cleanup handles selected connection locals, but the lower-level API itself continues creating unowned stack copies.

The full checked-in audit matrix/checklist from #375 is not treated as a blocker for this PR because it says `Refs #375`, not `Closes #375`; the findings above are blockers because they are directly within this PR's claimed cleanup scope.

src/quic/path.zig
```diff
@@ -128,22 +129,37 @@ pub const TokenError = error{
...
+    /// free to poison-fill an inactive optional payload — observably so on
+    /// Linux x64 — which would make it impossible to assert the bytes were
+    /// actually zeroed rather than merely tagged absent.
+    keys: [max_token_keys]secrets.FixedSecret(token_key_len) = [_]secrets.FixedSecret(token_key_len){secrets.FixedSecret(token_key_len){}} ** max_token_keys,
     current: u8 = 0,
 
     pub fn install(self: *RetryTokenKeyRing, key_id: u8, key: [token_key_len]u8) void {
         std.debug.assert(key_id < max_token_keys);
-        self.keys[key_id] = key;
+        self.keys[key_id].replace(&key) catch unreachable;
```
Jaruso: **[P1] The ring storage is now typed, but `install` still creates an unowned by-value copy of the Retry key.** `Runtime.init` wipes its local `retry_key`, but the `[token_key_len]u8` parameter here is a separate semantic copy and is never wiped when this function returns. The compiler/ABI may optimize it away, but #375 explicitly requires owned, auditable cleanup rather than relying on copy elision.

Borrow the key instead:

```zig
pub fn install(
    self: *RetryTokenKeyRing,
    key_id: u8,
    key: *const [token_key_len]u8,
) void {
    std.debug.assert(key_id < max_token_keys);
    self.keys[key_id].replace(key) catch unreachable;
    self.current = key_id;
}
```

and call it as `install(0, &retry_key)`. Update test constants through addressable locals where necessary.

src/http/http3_runtime.zig
```diff
@@ -674,7 +694,7 @@ pub const Runtime = struct {
             .now_us = now,
             .initial_path = .{ .local = self.local_address, .remote = addressFromSockaddrIn(peer) },
             .initial_address_validated = retry_context != null,
-            .stateless_reset_key = self.stateless_reset_key,
+            .stateless_reset_key = self.secrets.stateless_reset_key,
```
Jaruso: **[P1] This still copies the process stateless-reset key through unowned raw stack values.** `Connection.Options.stateless_reset_key` is `[32]u8`, the struct literal copies the key into an `Options` temporary, `Connection.init(..., options: Options)` takes that whole value by value, and initialization copies it again into `Connection.stateless_reset_key`. Only the final connection field is later wiped; the temporary/parameter copies are not.

Make the input borrowed and the connection owner typed:

```zig
// Options
stateless_reset_key: *const [32]u8,

// Connection
stateless_reset_key: crypto_secrets.FixedSecret(32) = .{},

// init
conn.stateless_reset_key.replace(options.stateless_reset_key) catch unreachable;
```

Then use `conn.stateless_reset_key.slice()` for token derivation and `LocalCidRegistry.initInto`, and pass `&self.secrets.stateless_reset_key` here. This removes the intermediate raw-key copies and gives the pre-registry connection lifetime an explicit owner.

src/quic/cid.zig
```diff
@@ -194,20 +225,19 @@ pub const LocalCidRegistry = struct {
...
+            if (!occupied) break i;
         } else return error.CidLimitExceeded;
         const entry = Entry{
             .sequence = self.next_sequence,
             .cid = cid,
-            .stateless_reset_token = statelessResetToken(self.reset_token_key, cid.slice()),
+            .stateless_reset_token = statelessResetToken(self.reset_token_key.slice(), cid.slice()),
         };
-        slot.* = entry;
+        self.entries[slot_idx] = entry;
```
Jaruso: **[P1] The reset-token copy chain is still by-value below the connection cleanup helpers.** `store()` builds a local `Entry` containing the token, copies it into `entries`, and returns it by value. `registerInitial()` returns another `Entry`; `issueCid()` turns the returned copy into a by-value `NewConnectionIdFrame`; `Connection.advertiseLocalCid()` stores that in a local `ncid`; and `frame.encodeNewConnectionId(value: NewConnectionIdFrame, ...)` takes another by-value parameter. Most of these stack copies have no owner or wipe.

This should be destination-oriented instead of relying on return-value optimization. One workable shape is:

```zig
fn store(self: *LocalCidRegistry, cid: ConnectionId) !*Entry {
    const idx = try self.findFreeSlot(cid);
    const entry = &self.entries[idx];
    entry.* = .{
        .sequence = self.next_sequence,
        .cid = cid,
        .stateless_reset_token = [_]u8{0} ** stateless_reset_token_len,
    };
    statelessResetTokenInto(
        &entry.stateless_reset_token,
        self.reset_token_key.slice(),
        cid.slice(),
    );
    self.occupied[idx] = true;
    self.next_sequence += 1;
    return entry;
}

pub fn issueCidInto(self: *LocalCidRegistry, cid: ConnectionId, out: *NewConnectionIdFrame) !void {
    const entry = try self.store(cid);
    out.* = .{ /* copy once into the caller-owned destination */ };
}
```

Have `advertiseLocalCid` obtain an ArrayList destination with `addOne` and initialize it in place, make `registerInitial` return `void`/sequence rather than a token-bearing `Entry`, and change the encoder to borrow `*const NewConnectionIdFrame`. Add a test allocator/stack-scratch hook that fails if the token pattern remains outside the registry and explicitly owned queue/record slots.

src/quic/connection.zig
```diff
@@ -885,6 +929,20 @@ pub const Connection = struct {
...
```
Jaruso, line 975: **[P1] The claimed eight-element bound is still false once CID retirement is included.** Deduplicating PTO requeues prevents multiple pending copies of the *same* sequence, but the `RETIRE_CONNECTION_ID` handler only calls `registry.retire()`. It does not remove a matching frame already in `pending_new_connection_ids`, and it does not clear `carried_new_connection_id` from matching in-flight `SentRecord`s.

That has two consequences:

1. A later PTO/loss can retransmit a `NEW_CONNECTION_ID` after the peer has explicitly retired that sequence.
2. Retirement frees a registry slot, so `needsLocalCid()` permits a replacement sequence while the old sequence remains queued/in flight. Repeating that cycle can leave more than `max_local_active_cids` distinct frame copies in these collections; the next `ensureUnusedCapacity()` then grows the queue and frees token-bearing backing storage unwiped.

Cancel every owned frame copy when retirement is processed, not only the registry entry. For example:

```zig
fn cancelLocalCidFrameCopies(self: *Connection, sequence: u64) void {
    var i: usize = 0;
    while (i < self.pending_new_connection_ids.items.len) {
        if (self.pending_new_connection_ids.items[i].sequence != sequence) {
            i += 1;
            continue;
        }
        crypto_secrets.secureZero(
            &self.pending_new_connection_ids.items[i].stateless_reset_token,
        );
        _ = self.pending_new_connection_ids.orderedRemove(i);
        wipePendingNewConnectionIdsOrderedRemoveResidue(
            &self.pending_new_connection_ids,
        );
    }

    for (self.sent_records.items) |*record| {
        if (record.carried_new_connection_id) |*ncid| {
            if (ncid.sequence != sequence) continue;
            crypto_secrets.secureZero(&ncid.stateless_reset_token);
            record.carried_new_connection_id = null;
        }
    }
}
```

Call this after a valid retire, including an idempotent repeated retire. Add a connection test with one pending and one in-flight copy, retire the sequence, fire PTO/loss, and prove it is never requeued. Then perform more than eight issue/send/retire/replace cycles and assert the queue backing pointer never changes and no retired token remains.

Jaruso (commented):
Re-reviewed the complete current head (`40e73c5`) against the prior findings and the secret-lifetime requirements represented by #375.

The third-round revision correctly fixes the previously reported moved-allocation leak in `readSmallFile`, borrows the Retry key and process reset key at the API boundaries, makes the primary local-CID issuance path destination-oriented, changes the frame encoder to borrow, and deduplicates repeated PTO requeues. CI is fully green across the complete matrix.

This is still not merge-ready because five ownership gaps remain:

1. `RETIRE_CONNECTION_ID` wipes only the registry entry. Pending and in-flight copies of that sequence remain live and can be retransmitted. Retire/replacement cycles also invalidate the claimed eight-element queue bound and can trigger the unwiped growth path again.
2. The requeue path still creates token-bearing stack copies and passes token-bearing structs by value into `ArrayList.append`; the same pattern remains when publishing a local `SentRecord` into `sent_records`.
3. The server's initial reset token is copied through the TLS CID-binding API and retained in `Tls13Backend.cid_binding`, but neither the transient copies nor the retained backend owner are wiped.
4. Public convenience APIs in `LocalCidRegistry` still accept or return the key/token-bearing objects by value, leaving the unsafe ownership pattern available even though the production path avoids it.
5. The crypto-boundary audit now exempts both reset-token functions even though only `statelessResetTokenInto` contains the direct HMAC call, unnecessarily weakening the guard.

I am continuing to treat this PR as a focused remediation slice because it says `Refs #375`, not `Closes #375`; the missing repository-wide audit matrix/checklist is therefore not independently blocking this PR. The eventual #375 audit still needs to disposition the separate peer-CID token owner and the remaining broader surfaces.

src/quic/connection.zig
```diff
@@ -1701,6 +1792,21 @@ pub const Connection = struct {
...
+    /// record it requeues from), so without this guard a repeated PTO would
+    /// append duplicate copies of the same NEW_CONNECTION_ID frame — pushing
+    /// `pending_new_connection_ids` past the capacity `init()` reserves once
+    /// up front and reopening the unwiped-growth path that reservation
+    /// exists to close.
+    fn queueNewConnectionId(self: *Connection, source: *const quic_cid.NewConnectionIdFrame) void {
+        for (self.pending_new_connection_ids.items) |*queued| {
+            if (queued.sequence == source.sequence) return;
+        }
+        self.pending_new_connection_ids.append(self.allocator, source.*) catch {};
```
Jaruso: **[P1] This still passes reset-token material through unowned by-value copies.** In `requeueRecord`, `|ncid_value|` captures the optional payload by value, `var ncid = ncid_value` makes another stack copy, and only the second copy is wiped. This `append(..., source.*)` then passes the complete token-bearing frame by value again. The packet publication path has the same final issue at `sent_records.append(self.allocator, record)`.

Borrow the optional payload and initialize the already-reserved destination directly:

```zig
if (record.carried_new_connection_id) |*ncid| {
    self.queueNewConnectionId(ncid);
}

fn queueNewConnectionId(
    self: *Connection,
    source: *const quic_cid.NewConnectionIdFrame,
) void {
    for (self.pending_new_connection_ids.items) |*queued| {
        if (queued.sequence == source.sequence) return;
    }
    std.debug.assert(
        self.pending_new_connection_ids.items.len <
            self.pending_new_connection_ids.capacity,
    );
    const dst = self.pending_new_connection_ids.addOneAssumeCapacity();
    dst.* = source.*;
}
```

Once the retirement invariant is fixed, the reserved-capacity assertion is valid. Likewise, publish the local recovery record with `addOneAssumeCapacity()` and a direct destination assignment rather than `ArrayList.append(record)`, then let the existing `defer wipeSentRecordToken(&record)` destroy only the explicitly owned local source.

src/quic/connection.zig
```diff
@@ -893,7 +951,7 @@ pub const Connection = struct {
         if (options.role == .server) {
             binding.original_destination_connection_id = conn.original_dcid;
             if (conn.retry_scid) |retry_source| binding.retry_source_connection_id = retry_source;
-            binding.stateless_reset_token = quic_cid.statelessResetToken(options.stateless_reset_key, conn.local_cid.slice());
+            binding.stateless_reset_token = quic_cid.statelessResetToken(options.stateless_reset_key[0..], conn.local_cid.slice());
```
Jaruso: **[P1] Borrowing the derivation key fixed that key's copy chain, but the derived initial reset token immediately starts another unowned copy chain here.** `statelessResetToken()` returns the token by value into local `binding`; `TlsBackend.setCidBinding(binding)` takes the whole binding by value; its vtable callback takes another value; and `tls_backend.zig` retains it in `Tls13Backend.cid_binding`. The local/parameter copies are never wiped, and `Tls13Backend.deinit()` currently wipes `local_transport_parameters` but not `cid_binding.stateless_reset_token` (or the peer binding owner).

Make this boundary borrowed and give the retained backend object an explicit cleanup path. A safe shape is:

```zig
// TlsBackend/vtable
setCidBindingFn: ?*const fn (*anyopaque, *const config.CidBinding) void,

pub fn setCidBinding(self: TlsBackend, binding: *const config.CidBinding) void {
    if (self.setCidBindingFn) |set| set(self.transport.ptr, binding);
}

// Connection.init
var binding = config.CidBinding{ ... };
defer binding.deinit(); // wipes only security-sensitive token fields
quic_cid.statelessResetTokenInto(
    &binding.stateless_reset_token.?,
    options.stateless_reset_key,
    conn.local_cid.slice(),
);
options.tls.setCidBinding(&binding);
```

The concrete backend should copy directly into its owned field, `encodeTransportParametersBound` should borrow that owner rather than taking `CidBinding` by value, and backend teardown must wipe both retained token fields before the object is released. Add a backend-level teardown test that begins with a non-zero local token and observes the retained storage after the cleanup helper runs.

src/quic/cid.zig
```diff
@@ -176,38 +210,52 @@ pub const LocalCidRegistry = struct {
...
             .retire_prior_to = 0,
             .cid = entry.cid,
             .stateless_reset_token = entry.stateless_reset_token,
         };
     }
 
-    fn store(self: *LocalCidRegistry, cid: ConnectionId) error{ CidLimitExceeded, DuplicateCid }!Entry {
+    /// Convenience wrapper over `issueCidInto` for callers (mostly tests)
+    /// that want an owned return value rather than a destination pointer.
+    pub fn issueCid(self: *LocalCidRegistry, cid: ConnectionId) error{ CidLimitExceeded, DuplicateCid }!NewConnectionIdFrame {
```
Jaruso: **[P2] The primary production path is destination-oriented now, but the public API still exposes the exact by-value ownership pattern this change is intended to eliminate.** `issue()` and this `issueCid()` return `NewConnectionIdFrame` by value, `get()` returns the token-bearing `Entry` by value, and `init()` still takes the 32-byte derivation key by value and returns the secret-owning registry by value. These are not compile-time test-only functions, so another production caller can silently recreate the unwiped copies the new APIs avoid.

Please make the safe API the only public shape:

```zig
pub fn get(self: *const LocalCidRegistry, sequence: u64) ?*const Entry { ... }

pub fn issueInto(
    self: *LocalCidRegistry,
    entropy: []const u8,
    cid_len: u8,
    out: *NewConnectionIdFrame,
) !void { ... }
```

Use `initInto` everywhere, or at minimum change `init` to borrow `*const [32]u8`. The return-by-value conveniences can be private test helpers whose callers explicitly wipe their local frame, but they should not remain part of the production-facing secret API.

scripts/audit_crypto_boundary.zig
```diff
@@ -265,7 +265,7 @@ const file_checks_with_exceptions = [_]FileCheckWithExceptions{
         // anywhere else in the file would pass (#490 fifth-pass review).
         .forbidden = &(full_forbidden ++ [_][]const u8{"HmacSha256."}),
         .exempt_exact = &.{"const HmacSha256 = std.crypto.auth.hmac.sha2.HmacSha256;"},
-        .exempt_functions = &.{"statelessResetToken"},
+        .exempt_functions = &.{ "statelessResetToken", "statelessResetTokenInto" },
```
Jaruso: **[P2] This exemption is broader than the implementation and weakens the boundary audit.** After this revision, `statelessResetToken` is only a wrapper; the sole direct `HmacSha256` call lives in `statelessResetTokenInto`. Exempting both functions means a future direct HMAC shortcut added back to the wrapper would pass the audit despite the rationale immediately below saying no other HMAC use may be added.

Keep the exception exact:

```zig
.exempt_functions = &.{"statelessResetTokenInto"},
```

The wrapper will continue to pass because it contains no forbidden primitive call.

src/quic/connection.zig
```diff
@@ -1032,11 +1129,18 @@ pub const Connection = struct {
     pub fn advertiseLocalCid(self: *Connection, cid_value: quic_cid.ConnectionId) error{ CidLimitExceeded, DuplicateCid, OutOfMemory }!void {
         try self.pending_new_connection_ids.ensureUnusedCapacity(self.allocator, 1);
```
Jaruso: **[P1] This can still reallocate and free a token-bearing queue before the registry rejects the operation.** When `pending_new_connection_ids.len == capacity == 8`, `ensureUnusedCapacity(1)` grows the `ArrayList` immediately, copying all live reset tokens and freeing the old backing allocation unwiped. Only afterward does `issueCidInto()` discover that the registry is at its active-CID limit and return `CidLimitExceeded`.

The queue was deliberately preallocated to its hard bound, so this path must never call a grow-capable API. Check the invariant before obtaining the destination:

```zig
const registry = if (self.local_cids) |*value| value else
    return error.CidLimitExceeded;

if (self.pending_new_connection_ids.items.len ==
    self.pending_new_connection_ids.capacity)
{
    return error.CidLimitExceeded;
}

const dst = self.pending_new_connection_ids.addOneAssumeCapacity();
registry.issueCidInto(cid_value, dst) catch |err| {
    self.pending_new_connection_ids.shrinkRetainingCapacity(
        self.pending_new_connection_ids.items.len - 1,
    );
    return err;
};
```

Preferably expose a registry preflight/reservation operation so queue publication is transactional and the destination is added only after CID validation succeeds. Add a test that reaches the real maximum, attempts one more advertisement, and proves the queue backing pointer/capacity did not change.

src/quic/tls_backend.zig
```diff
@@ -61,12 +61,12 @@ const tp_retry_source_connection_id: u64 = 0x10;
...
 
 pub fn encodeTransportParameters(params: config.TransportParameters, buf: []u8) HandshakeError![]const u8 {
-    return encodeTransportParametersBound(params, .{}, buf);
+    return encodeTransportParametersBound(params, &.{}, buf);
 }
 
 pub fn encodeTransportParametersBound(
     params: config.TransportParameters,
-    binding: config.CidBinding,
+    binding: *const config.CidBinding,
```
Jaruso, line 69: **[P1] Borrowing the binding at the function boundary does not yet eliminate the internal token copies.** Later in this function, `if (binding.stateless_reset_token) |token|` captures the 16-byte optional payload by value and leaves that stack copy unwiped. Borrow the payload too:

```zig
if (binding.stateless_reset_token) |*token| {
    // encode lengths...
    @memcpy(buf[len..][0..token.len], token);
    len += token.len;
}
```

There is a second discarded-owner path in this file: `decodeTransportParameters()` creates a local `CidBinding`, decodes a peer token into it, returns only `TransportParameters`, and never wipes the binding. That wrapper is used by the HTTP/3 early-data compatibility path. Now that `CidBinding.deinit()` exists, it should be:

```zig
var binding = config.CidBinding{};
defer binding.deinit();
return decodeTransportParametersBound(bytes, &binding);
```

Please cover both the encode scratch-copy avoidance and the decode-wrapper cleanup with focused tests.

src/quic/tls_handshake.zig
```diff
@@ -70,7 +70,7 @@ pub const TlsBackend = struct {
     /// Optional RFC 9000 §7.3 authentication-binding hooks. A backend that
     /// carries connection IDs in its transport parameters implements both; the
     /// in-memory test backend leaves them null.
-    setCidBindingFn: ?*const fn (ptr: *anyopaque, binding: config.CidBinding) void = null,
+    setCidBindingFn: ?*const fn (ptr: *anyopaque, binding: *const config.CidBinding) void = null,
```
Jaruso, line 73: **[P1] The peer CID-binding half of this interface still returns the token-bearing owner by value.** The concrete callback returns `self.peer_cid_binding`, this wrapper returns another value, `Connection.afterHandshakeProgress()` stores it in a local, and `validateCidBinding(peer_binding: CidBinding)` takes another by-value parameter. A server's stateless-reset token therefore crosses multiple unowned stack copies, and the connection-local copy is never deinitialized.

Make the getter borrowed end-to-end, mirroring the setter:

```zig
peerCidBindingFn: ?*const fn (*anyopaque) *const config.CidBinding = null,

pub fn peerCidBinding(self: TlsBackend) ?*const config.CidBinding {
    const get = self.peerCidBindingFn orelse return null;
    return get(self.transport.ptr);
}
```

The concrete backend should return `&self.peer_cid_binding`, and `Connection.validateCidBinding` should accept `*const CidBinding`. For backends without the hook, treat `null` as the current empty-binding behavior. This leaves exactly one retained owner, which the new backend `deinit()` already wipes.

src/quic/connection.zig
```diff
@@ -1701,6 +1825,61 @@ pub const Connection = struct {
...
+    /// to wipe its own entry; it says nothing about copies of that
+    /// sequence's NEW_CONNECTION_ID frame already queued or in flight.
+    /// Without also canceling those: a later PTO/loss could retransmit a
+    /// frame for a sequence the peer already retired, and retiring frees a
+    /// registry slot that `needsLocalCid()` may immediately reuse while the
+    /// old sequence's copies are still live, silently reopening the
+    /// capacity bound `queueNewConnectionId` depends on staying fixed.
+    /// Called for every valid retire, including an idempotent repeat of an
+    /// already-retired sequence.
+    fn cancelLocalCidFrameCopies(self: *Connection, sequence: u64) void {
```
Jaruso, line 1879: **[P1 test gap] The new retirement invariant is not exercised by a regression test.** The existing retirement test verifies only the registry/`needsLocalCid()` result, and the teardown test calls wipe helpers directly. Neither would fail if this function stopped removing queued copies, stopped clearing in-flight copies, or allowed PTO/loss to resurrect the retired sequence.

Add a connection-level test that:

1. creates an in-flight `NEW_CONNECTION_ID` and requeues the same sequence so one copy is also pending;
2. processes `RETIRE_CONNECTION_ID` from a different active CID;
3. proves the pending copy is removed and every matching sent-record optional is cleared with zeroed token bytes;
4. fires PTO and loss processing and proves the retired sequence is never requeued; and
5. performs more than eight issue/send/retire/replace cycles while asserting the pending-array backing pointer never changes.

That last loop is necessary to validate the exact capacity proof this function is now responsible for maintaining.

Jaruso (commented):
Blocking review on `b29a030a62be6e1070285fd9805337466a3ee925`.

The fourth-round commit substantially fixes the previous findings: retirement now cancels queued/in-flight frame copies, requeue and sent-record publication are destination-oriented, the local CID-binding setter is borrowed, the unsafe registry convenience APIs are private, and the HMAC audit exception is exact.

This is still not merge-ready because the fixed-capacity guarantee can still be defeated before registry validation, and two TLS CID-binding paths still copy/discard stateless-reset tokens without an owner cleanup path. The new retirement invariant also has no focused regression test, despite that test being necessary to make the queue-capacity proof durable.

CI run 30719441162 currently has no failed jobs; the ordinary Linux/macOS/ARM tests, Ubuntu appliance tests, builds, security, interop, lint, format, and dependency review have passed. A few appliance/packaging/dependency-audit jobs are still completing as of this review.

Additional replacement-lifecycle note: `Tls13Backend.setCidBinding()` directly overwrites `self.cid_binding`. If this hook can be called more than once, wipe the prior owner before replacement; assigning a binding with `stateless_reset_token = null` can otherwise leave the old optional payload bytes resident until final teardown.

Jaruso (commented):
Re-reviewed the latest head (`7c1dce2161dc4ab4c02a5d6b69c1b2f4a39c6768`) against the prior three production findings and the requested retirement coverage.

The production changes are now correct:

- `advertiseLocalCid()` no longer invokes a grow-capable queue API before CID validation.
- transport-parameter encoding borrows the optional reset token, and the decode-only wrapper now destroys its temporary binding.
- peer CID binding is borrowed end-to-end, and replacement of the backend's retained local binding wipes the prior token first.

I did not find another production ownership defect in the new commit. However, the new retirement tests still do not satisfy the cleanup/capacity regression criterion from #375, for the two reasons called out inline: the byte scan can pass without a secure wipe because nulling an optional may overwrite/poison its payload, and the multi-cycle test never invokes PTO/loss after each retirement, so stale in-flight copies would never accumulate in the pending queue it claims to bound.

Please make the cleanup state observable after the logical transition (an always-live payload plus an explicit presence flag is the most robust shape, consistent with the key-ring/registry fixes), inspect the vacated pending slot as raw bytes, and drive retransmission after every retire in the capacity loop.

CI run `30720324243` has no failures at review time: Linux/macOS/ARM unit tests, lint, format, security, PKI differential, dependency review, and completed builds are green. Remaining appliance, packaging, TLS-dependency, interop, performance, and ReleaseFast jobs were still running.

Because this PR uses `Refs #375` rather than closing it, the broader repository-wide audit matrix/checklist acceptance criteria remain outside this focused lifecycle patch.

src/quic/connection.zig
```diff
@@ -6832,6 +7059,96 @@ test "connection: advertised local CID can be retired from another active CID" {
...
+    try testing.expectEqual(State.established, pair.server.state());
+
+    // 3) Both the pending and the in-flight copy are gone. The sent
+    // record's storage no longer contains the original token bytes
+    // anywhere, proven via a raw byte scan rather than reading the
+    // optional's payload after it has been nulled.
+    try testing.expectEqual(@as(usize, 0), pair.server.pending_new_connection_ids.items.len);
+    try testing.expectEqual(@as(?quic_cid.NewConnectionIdFrame, null), pair.server.sent_records.items[sent_index].carried_new_connection_id);
+    const record_bytes = std.mem.asBytes(&pair.server.sent_records.items[sent_index]);
+    try testing.expect(std.mem.indexOf(u8, record_bytes, &original_token) == null);
```
Jaruso: **[P1 test gap] This assertion does not prove that `secureZero()` ran.** Assigning `record.carried_new_connection_id = null` is allowed to overwrite or poison the optional payload in safety builds, so an implementation that simply removes the wipe and nulls the optional can still make `original_token` disappear from `record_bytes`. The pending assertion has the same problem in the other direction: `items.len == 0` says nothing about the vacated slot in `allocatedSlice()`.

Please make the cleanup observable independently of the logical empty transition. The robust shape is the same one already adopted for the Retry ring and CID registry: keep an always-live `NewConnectionIdFrame` payload plus an explicit `present` flag, wipe the token, then clear the flag. That lets the regression test inspect the token bytes after cancellation without reading an inactive optional payload. At minimum, factor the active-payload wipe into a helper that can be asserted before nulling, and capture the pending array's vacated slot as raw bytes so the test verifies that whole slot is zero after `orderedRemove`.

Also avoid `const sent_copy = ...` here or explicitly wipe it: this security-lifetime test currently creates another token-bearing by-value stack copy of the exact record it is trying to prove has a single audited lifetime.

src/quic/connection.zig
```diff
@@ -6832,6 +7059,96 @@ test "connection: advertised local CID can be retired from another active CID" {
...
+        try pair.server.advertiseLocalCid(cid);
+        _ = pair.server.pollTransmitOnPath(&out, pair.now_us) orelse return error.TestExpectedEqual;
+        try pair.server.applyFrame(
+            .application,
+            .{ .retire_connection_id = .{ .sequence = @as(u64, i) + 1 } },
+            TestPair.server_path,
+            0,
+            pair.now_us,
+        );
+        try testing.expectEqual(State.established, pair.server.state());
```
Jaruso, line 7165: **[P1 test gap] This loop does not exercise the capacity failure mode it claims to prove.** Each iteration sends the frame and retires it, but never fires PTO or loss processing afterward. If `cancelLocalCidFrameCopies()` regressed so it retired the registry entry while leaving each in-flight `SentRecord` copy intact, this loop would still keep `pending_new_connection_ids` empty and both backing-pointer assertions would pass.

Drive the retransmission path after every retirement, before issuing the next CID, for example:

```zig
pair.server.firePto(.application);
pair.server.firePto(.application);
try testing.expectEqual(@as(usize, 0), pair.server.pending_new_connection_ids.items.len);
for (pair.server.sent_records.items) |record| {
    if (record.carried_new_connection_id) |ncid| {
        try testing.expect(ncid.sequence != @as(u64, i) + 1);
    }
}
```

A forced loss/detection path is even better because it covers both requeue callers. Then continue for more than `max_local_active_cids` cycles and assert the queue pointer/capacity never changes. Without retransmission pressure, the test does not validate the invariant that prevents stale retired sequences from eventually filling and growing the queue.

Jaruso (commented):
Re-reviewed at `8e425c62b53503438b2257f7a27d3d81517f698f`.

No blocking findings remain in this PR's scope.

The latest commit closes both issues from the previous pass:

- `SentRecord` now uses an always-live `NewConnectionIdFrame` payload plus an explicit presence flag, so retirement tests can directly verify the token bytes were actually zeroized rather than merely hidden or poison-filled by an optional-state transition.
- The retirement regression now inspects the pending queue's vacated backing slot and the retained sent-record token storage, then proves PTO/direct requeue cannot resurrect the retired sequence.
- The multi-cycle test applies retransmission pressure after every retire, so stale in-flight frame copies would immediately re-enter the queue and fail the invariant instead of remaining unobserved.

I also rechecked all production transitions introduced by the representation change: packet construction sets the flag only after populating the payload, recovery requeue checks the flag, ACK/loss teardown wipes before removal, cancellation wipes before clearing the flag, and connection teardown wipes every still-live owner. I did not find a stale-frame reuse or unwiped-copy path introduced by this revision.

CI run `30723280898` completed successfully across unit tests, appliance profiles, Debug/ReleaseSafe/ReleaseFast builds on Linux/ARM/macOS, H3 resumption/0-RTT interop, TLS dependency audit, security, packaging, performance, lint, format, dependency review, and PKI differential checks.

Because this PR says `Refs #375` rather than closing it, I am not treating the repository-wide audit matrix/checklist and the remaining broader #375 targets as merge blockers for this focused remediation slice.

PR #550, opened 2026-08-01
crypto: complete pure-Zig ECDSA P-256 signing

Jaruso:
## Summary
- complete `pure_zig.SoftwareEcdsaP256SigningKey` construction, scalar validation, entropy-backed signing noise, DER output ordering, and wipe behavior
- thread injected provider entropy through fixed and identity-backed concrete credential providers without changing the provider-neutral handshake/transport seams
- add focused provider and credential tests for ECDSA signing, entropy failure/order, DER canonicality, invalid constructors, selection compatibility, and retained Ed25519 behavior
- reconcile `docs/CRYPTO_PROVIDER.md` and `src/crypto/profile.zig` with live ECDSA signing support

Fixes #432

## Validation
- `zig fmt --check .`
- `zig build`
- `zig build audit-crypto-boundary`
- `zig build test-crypto -freference-trace`
- `zig build test-tls -freference-trace`
- `zig build test-quic -freference-trace`
- `zig build test -freference-trace`
- `zig build -Dtls-profile=appliance`
- `zig build -Dtls-profile=appliance test -freference-trace`


Jaruso (commented):
**Review disposition: changes required before merge.** GitHub does not permit the authenticated PR author to formally request changes on their own pull request, so this is submitted as a comment review.

The core implementation is strong: ECDSA signing now consumes injected entropy, rejects undersized output before entropy consumption, stages complete DER before publishing output, threads production entropy through the fixed/SNI credential paths, and keeps HTTP/QUIC code provider-neutral. The full GitHub Actions run is green, including the appliance no-OpenSSL audits and appliance test suites.

Five requirements from #432 are not yet fully satisfied:

1. **Secret lifecycle is still incomplete in the scalar constructors.** Valid construction leaves typed `SecretKey` copies in implementation frames without explicit wiping, and the documented consuming constructors do not consume malformed-length `FixedSecret` inputs.
2. **The new secret-owning type remains ordinarily formattable/loggable.** #432 explicitly prohibits ordinary formatting from exposing private material; this should use the same compile-time formatting prohibition as `crypto.secrets.FixedSecret`.
3. **The ECDSA S-value normalization policy is undocumented.** #432 explicitly requires an intentional policy. The current `std.crypto` path does not appear to normalize low-S, so document that high-S signatures may be emitted and are accepted, or normalize deliberately and test that behavior.
4. **Provider-capability withdrawal is not covered.** The new credential test proves peer-offer incompatibility only. Add a test proving that removing ECDSA from provider/profile capabilities removes it before credential selection/signing, and enforce the gate if the current hand-written-policy path can still select it.
5. **Process-global mutable deterministic entropy was reintroduced across test modules.** This regresses the per-instance entropy isolation established in #490 and makes tests order-dependent and unsafe for parallel execution. Use per-test/per-harness state, or a stateless non-mutating entropy stub where Ed25519 ignores the input.

Once these are addressed, the remaining #432 acceptance criteria appear covered: construction and scalar rejection, opaque `SigningKey`, scheme reporting, canonical DER verification, entropy/output failure behavior, credential compatibility, retained Ed25519 behavior, provider-neutral TLS integration, documentation, and the no-OpenSSL appliance path.

src/crypto/pure_zig.zig
```diff
@@ -562,14 +563,69 @@ pub const SoftwareSigningKey = struct {
...
+
+    /// Load a raw scalar from typed secret storage, consuming and clearing the
+    /// source container before key derivation.
+    pub fn fromScalarSecret(scalar: *secrets.FixedSecret(EcdsaP256Sha256.SecretKey.encoded_length)) provider.SignError!SoftwareEcdsaP256SigningKey {
+        if (scalar.len != EcdsaP256Sha256.SecretKey.encoded_length) return error.InvalidInput;
+        var bridge: [EcdsaP256Sha256.SecretKey.encoded_length]u8 = undefined;
+        defer crypto.secureZero(u8, &bridge);
+        @memcpy(&bridge, scalar.slice());
+        scalar.deinit();
+        const secret_key = try validatedSecretKey(bridge);
```
Jaruso: This valid-input path still leaves private scalar copies unwiped. `validatedSecretKey()` returns a `SecretKey` by value, `secret_key` remains in this frame, and `fromSecretKey(secret_key)` creates another by-value copy; `fromSecretKey` also validates through another returned `SecretKey` that is immediately discarded. That misses #432's requirement to wipe temporary private-key material under the shared #372 lifecycle rules.

The consuming constructors also return on a wrong `FixedSecret.len` before calling `deinit`, despite documenting that they consume the source.

Please make consumption unconditional and wipe each implementation-owned typed secret, for example:

```zig
pub fn fromScalarSecret(
    scalar: *secrets.FixedSecret(EcdsaP256Sha256.SecretKey.encoded_length),
) provider.SignError!SoftwareEcdsaP256SigningKey {
    defer scalar.deinit();
    if (scalar.len != EcdsaP256Sha256.SecretKey.encoded_length)
        return error.InvalidInput;

    var bridge: [EcdsaP256Sha256.SecretKey.encoded_length]u8 = undefined;
    defer crypto.secureZero(u8, &bridge);
    @memcpy(&bridge, scalar.slice());

    var secret_key = try validatedSecretKey(bridge);
    defer crypto.secureZero(u8, &secret_key.bytes);
    return deriveValidatedSecretKey(&secret_key);
}
```

Apply the same ownership rule to `fromSeedSecret`, `fromScalarBytes`, and the public `fromSecretKey` boundary so every local copy is wiped on success and error.

src/crypto/pure_zig.zig
```diff
@@ -562,14 +563,69 @@ pub const SoftwareSigningKey = struct {
 pub const SoftwareEcdsaP256SigningKey = struct {
     key_pair: EcdsaP256Sha256.KeyPair,
 
+    /// Deterministically derive a P-256 key pair from fixture seed material.
+    /// This mirrors `std.crypto`'s deterministic constructor while wiping this
+    /// frame's seed copy after derivation.
+    pub fn fromSeed(seed: [EcdsaP256Sha256.KeyPair.seed_length]u8) provider.SignError!SoftwareEcdsaP256SigningKey {
```
Jaruso, line 578: Now that this is a public constructible secret-owning type, it needs the observability protection required by #432. The struct currently has no `format` override, so ordinary generic formatting can recursively expose `key_pair.secret_key`.

Please mirror the established secret-container rule:

```zig
pub fn format(
    _: SoftwareEcdsaP256SigningKey,
    comptime _: []const u8,
    _: std.fmt.FormatOptions,
    _: anytype,
) !void {
    @compileError("signing keys must not be formatted or logged");
}
```

Also verify that formatting the owning `Identity`/key union cannot bypass this guard. The issue explicitly requires ordinary formatting/logging never to expose private scalar or signing noise.

docs/CRYPTO_PROVIDER.md
```diff
@@ -120,7 +120,7 @@ lists only the consumers whose live runtime calls `CryptoProvider` today
...
 | Ed25519 | supported | `std.crypto` | provider deferred | CertificateVerify, PKI | live for both — PKI chain-signature verification and the handshake's own CertificateVerify proof-of-possession (`src/tls/tls13_backend.zig`) call `CryptoProvider.verify`; local CertificateVerify signing (`src/tls/credentials.zig`'s `Identity.sign`) signs through the opaque `provider.SigningKey` handle rather than a concrete `std.crypto.sign.Ed25519.KeyPair` | native appliance, general-purpose OpenSSL |
-| ECDSA-P256-SHA256 | supported | `std.crypto` verification | provider deferred | CertificateVerify, PKI | live for both, same split as Ed25519 — chain verification and CertificateVerify both call `CryptoProvider.verify`, and local signing goes through `provider.SigningKey` | native appliance, general-purpose OpenSSL |
+| ECDSA-P256-SHA256 | supported | `std.crypto` signing/verification | provider deferred | CertificateVerify, PKI | live for both, same split as Ed25519 — chain verification and CertificateVerify both call `CryptoProvider.verify`, and local signing goes through `provider.SigningKey` with per-signature noise from injected provider entropy | native appliance, general-purpose OpenSSL |
```
Jaruso: #432 explicitly requires the provider contract to state whether ECDSA signatures are normalized (for example, low-S). This row documents entropy and DER encoding, but not the S-value policy. The current `std.crypto` signing path appears to emit the primitive's `s` directly rather than replacing it with `min(s, n - s)`.

Please choose and document one policy, then pin it with a test:

- **Non-normalizing policy:** state that canonical DER does not imply low-S, high-S signatures may be emitted, and verification accepts either valid in-range form without silently rewriting it; or
- **Low-S policy:** normalize before DER serialization and assert `s <= n/2` for every generated signature.

Without that, the implementation leaves a security/interoperability behavior deliberately called out by the issue unspecified.

src/tls/credentials.zig
```diff
@@ -1163,6 +1161,11 @@ pub const MockVerifier = struct {
 // ===========================================================================
 
 const testing = std.testing;
+var mock_sign_entropy = pure_zig.DeterministicEntropy.init(0x4325);
```
Jaruso: This reintroduces process-global mutable deterministic entropy. The same pattern appears in `sni_provider.zig`, `edge_gateway.zig`, `http3_runtime.zig`, `native_tls_connection.zig`, `quic/tls_backend.zig`, `tls13_backend_tests.zig`, and `quic_h3_udp_smoke.zig`.

#490 deliberately removed shared mutable test-provider streams in favor of per-instance storage. These globals make a test's signature depend on what ran before it and become a data race if the runner executes tests concurrently.

Please use per-test/per-harness entropy storage and pass its borrowed `Entropy` only for that owner's lifetime. For Ed25519-only mocks, a stateless fill function is sufficient because the signer ignores entropy:

```zig
fn ignoredEntropyFill(_: *anyopaque, out: []u8) provider.EntropyError!void {
    @memset(out, 0);
}
```

ECDSA tests that exercise randomness should own an independent `DeterministicEntropy` value locally, as the new provider tests already do.

src/tls/credentials.zig
```diff
@@ -1232,8 +1237,34 @@ test "fixed provider signs a bounded output the certificate can verify" {
     try sig.verify(message, public_key);
 }
 
+test "fixed provider selects and signs ECDSA only for compatible offers" {
+    var det = pure_zig.DeterministicEntropy.init(0x4326);
+    var fixed = FixedCredentialProvider.init(testdata.p256Identity(), det.entropy());
+    defer fixed.deinit();
+    const provider_view = fixed.provider();
+
+    const incompatible = testSelection(&.{0x0807});
```
Jaruso, line 1262: This proves only that an ECDSA credential is rejected when the peer offers Ed25519. It does not cover #432's separate requirement that **provider capability absence** removes ECDSA from the selectable matrix before signing.

Please add a provider/profile-backed test that removes `.ecdsa_secp256r1_sha256` from `Capabilities.signatures`, derives or validates the TLS policy, and proves the ECDSA credential is never selected. In particular, exercise the live backend boundary, not only `FixedCredentialProvider`, because the fixed provider now receives only `Entropy` and therefore cannot itself observe capability withdrawal.

Expected shape:

```zig
var caps = pure_zig.Provider.capabilities();
caps.signatures.remove(.ecdsa_secp256r1_sha256);
const tls_caps = crypto_profile.fromProfile(.native_appliance, caps);
try expect(!contains(tls_caps.signature_schemes, .ecdsa_secp256r1_sha256));
// Then prove a handshake/selection path cannot reach the ECDSA signer.
```

If a hand-written policy can still contain ECDSA while the injected provider lacks it, add a pre-selection/selected-credential capability check so this fails before any Certificate flight is emitted rather than later as a local signing failure.

src/crypto/pure_zig.zig
```diff
@@ -562,21 +572,97 @@ pub const SoftwareSigningKey = struct {
...
+        var local_seed = seed;
+        defer crypto.secureZero(u8, &local_seed);
+        const key_pair = EcdsaP256Sha256.KeyPair.generateDeterministic(local_seed) catch return error.ProviderFailure;
+        return .{ .key_pair = key_pair };
+    }
+
+    /// Load from a seed already held in a typed secret container, consuming
+    /// and clearing it before key derivation.
+    pub fn fromSeedSecret(seed: *secrets.FixedSecret(EcdsaP256Sha256.KeyPair.seed_length)) provider.SignError!SoftwareEcdsaP256SigningKey {
+        defer seed.deinit();
```
Jaruso, line 588: The `defer` fixes consumption on the wrong-length/error path, but it no longer clears the source **before key derivation**, as this API promises. During `generateDeterministic`, both `seed.bytes` and `bridge` remain live; `fromScalarSecret` similarly keeps `scalar.bytes`, `bridge`, and `secret_key` live through public-key derivation.

Please keep the unconditional cleanup and also clear immediately after copying:

```zig
pub fn fromSeedSecret(seed: *secrets.FixedSecret(seed_len)) provider.SignError!Self {
    defer seed.deinit(); // covers wrong length and copy-time failures
    if (seed.len != seed_len) return error.InvalidInput;

    var bridge: [seed_len]u8 = undefined;
    defer crypto.secureZero(u8, &bridge);
    @memcpy(&bridge, seed.slice());
    seed.deinit(); // shorten ownership now; deferred call becomes a no-op

    const key_pair = EcdsaP256Sha256.KeyPair.generateDeterministic(bridge) catch
        return error.ProviderFailure;
    return .{ .key_pair = key_pair };
}
```

Apply the same immediate-consume pattern to `fromScalarSecret`.

Also, `fromSecretKey(secret_key: SecretKey)` still receives a private scalar **by value** and wipes only the additional `local_secret_key`; the original parameter copy is never cleared. Since the production PKCS#8 path now uses `fromScalarSecret`, please remove that public by-value entry point or replace it with a documented pointer-consuming API, e.g. `fromSecretKey(secret_key: *SecretKey)` with a `defer secureZero(&secret_key.bytes)`. This is the remaining part of the prior secret-copy finding.

Jaruso (commented):
Second-pass review against issue #432 and the five findings from the first pass.

The follow-up commit correctly added the formatting guards, documented the non-low-S policy, added capability-withdrawal coverage before any Certificate flight, and removed the process-global mutable deterministic entropy. The complete CI matrix for this head is green, including appliance/no-OpenSSL checks on Linux, macOS, and ARM.

Two acceptance-level items remain before merge:

1. The consuming typed-secret constructors still retain their source containers until function exit rather than clearing them immediately after the bridge copy, and the public by-value `fromSecretKey` boundary still leaves its parameter copy outside the wipe.
2. The new alternate-S test pins verification acceptance, but it does not pin the documented non-normalizing signing policy; it would continue passing if the signer later added low-S normalization.

Because GitHub does not allow the PR author account to formally request changes on its own PR, this is submitted as a comment review, but I consider both inline findings blocking for closing #432.

src/crypto/pure_zig.zig
```diff
@@ -953,6 +1040,182 @@ test "SoftwareEcdsaP256SigningKey signs then verifies, with tamper and wrong-key
...
+    const len_b = try signer.sign(message, det_b.entropy(), &sig_b);
+
+    try testing.expectEqualSlices(u8, sig_a1[0..len_a1], sig_a2[0..len_a2]);
+    try testing.expect(!std.mem.eql(u8, sig_a1[0..len_a1], sig_b[0..len_b]));
+    try expectCanonicalEcdsaDer(sig_a1[0..len_a1]);
+    try expectCanonicalEcdsaDer(sig_b[0..len_b]);
+    try cp.verify(.ecdsa_secp256r1_sha256, &public_key, message, sig_a1[0..len_a1]);
+    try cp.verify(.ecdsa_secp256r1_sha256, &public_key, message, sig_b[0..len_b]);
+
+    const scalars = try parseDerSignatureScalars(sig_a1[0..len_a1]);
```
Jaruso, line 1155: This proves that verification accepts both `s` and `n - s`, but it does **not** pin the documented non-normalizing signing policy. The test still passes unchanged if `signingSignImpl` later normalizes every output to low-S, so the documentation could silently become false.

Please cross-check one provider signature against the direct primitive output under known fixed noise, or use a known fixture whose emitted `s` is high and assert that it remains high. A direct parity check is straightforward:

```zig
const noise = [_]u8{0xa5} ** EcdsaP256Sha256.noise_length;
var entropy = FixedEntropy.init(noise);
const written = try signer.sign(message, entropy.entropy(), &actual);

const direct_sig = try key.key_pair.sign(message, noise);
var expected_buf: [EcdsaP256Sha256.Signature.der_encoded_length_max]u8 = undefined;
const expected = direct_sig.toDer(&expected_buf);
try testing.expectEqualSlices(u8, expected, actual[0..written]);
```

That pins “emit the primitive's S directly / do not normalize.” Keep the existing `n - s` verification check as the separate proof that verification accepts either valid in-range form.

Jaruso (commented):
Re-review complete at `2fc4db3`: **no remaining findings**.

The two outstanding security-contract issues are now correctly addressed:

- `fromSeedSecret` / `fromScalarSecret` clear the caller-owned typed secret immediately after copying while retaining unconditional deferred cleanup for every early return, and the unsafe public by-value `fromSecretKey` boundary is replaced with a pointer-consuming API that wipes the caller-owned scalar on success and failure.
- The non-normalizing ECDSA signing policy is now pinned by comparing the provider output byte-for-byte with the direct `std.crypto` primitive under identical fixed noise, while the separate alternate-S test continues to prove verification accepts both valid forms.

I also rechecked the complete #432 scope: injected entropy use and failure mapping, maximum-DER output preflight, no partial output publication, canonical DER, temporary-noise/DER/key cleanup, formatting/logging guards, malformed scalar rejection, credential and SNI integration, provider-capability withdrawal before flight emission, Ed25519 preservation, documentation/profile updates, and appliance/no-OpenSSL isolation. No regression was introduced by this follow-up.

The latest full CI run passes across Linux, macOS, ARM, appliance-profile tests, TLS dependency audit, PKI differential testing, H3 interop, security, packaging, and performance smoke.

GitHub does not permit the PR author to formally approve their own pull request, but from this review's perspective the implementation is ready to merge.

PR #551, opened 2026-08-01
crypto: add test-only EVP differential oracle

Jaruso:
Claims #431

## Summary
- add a standalone test-only OpenSSL EVP oracle child process for AEAD, X25519, Ed25519, and ECDSA-P256 verification primitives
- wire the oracle only into the general-profile `test-crypto-openssl` build path and pass its installed path to the Zig harness
- replace #431 OpenSSL differential waivers with executable positive/negative registry coverage while leaving unrelated #362 PSK waivers intact

## Validation
- `zig build test-crypto-openssl`
- `zig build test -Dtls-profile=appliance`
- `zig build test`

### DIFF
--- a/tests/crypto_openssl_diff.zig
+++ b/tests/crypto_openssl_diff.zig
@@ -7,6 +7,7 @@
 
 const std = @import("std");
 const compat = @import("zig_compat");
+const diff_options = @import("crypto_openssl_diff_options");
 const crypto_pkg = @import("crypto");
 const quic = @import("quic");
 const tls_core = @import("tls_core");
@@ -15,10 +16,13 @@ const testing = std.testing;
 const provider = crypto_pkg.provider;
 const profile = crypto_pkg.profile;
 const pure_zig = crypto_pkg.pure_zig;
+const X25519 = std.crypto.dh.X25519;
+const Ed25519 = std.crypto.sign.Ed25519;
 
 const OpenSslError = error{
     MissingOpenSslKdfOracle,
     OpenSslOracleFailed,
+    EvpOracleFailed,
     DifferentialMismatch,
 };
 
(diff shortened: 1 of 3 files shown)


tests/crypto_openssl_diff.zig
```diff
@@ -157,6 +161,409 @@ fn hexAlloc(allocator: std.mem.Allocator, bytes: []const u8) ![]u8 {
...
+fn runEvpOracleRaw(allocator: std.mem.Allocator, args: []const []const u8) ![]u8 {
+    var argv = try allocator.alloc([]const u8, args.len + 1);
+    defer allocator.free(argv);
+    argv[0] = diff_options.evp_oracle_path;
+    for (args, 0..) |arg, i| argv[i + 1] = arg;
+
+    const result = try std.process.run(allocator, compat.io(), .{
+        .argv = argv,
+        .stdout_limit = .limited(8 * 1024),
+        .stderr_limit = .limited(1024),
```
Jaruso: `8 * 1024` is smaller than the largest valid response this oracle permits. `EVP_ORACLE_MAX_INPUT` accepts a 4096-byte AEAD plaintext, but `aead-seal` returns `"ok "` + 8192 ciphertext hex chars + a separator + 32 tag hex chars + newline (8229 bytes). That valid command will therefore fail with the stdout limiter instead of returning an oracle result. Please either raise this cap above the maximum encoded response (with a small fixed margin, e.g. `2 * EVP_ORACLE_MAX_INPUT + protocol overhead`) or reduce the oracle's accepted payload bound so the two contracts agree, and add a boundary test.

tests/evp_oracle.c
```diff
@@ -0,0 +1,246 @@
...
+        return status("oracle_error");
+    }
+    ec = NULL;
+    EVP_MD_CTX *ctx = EVP_MD_CTX_new();
+    int result = 0;
+    int ok = ctx != NULL &&
+        EVP_DigestVerifyInit(ctx, NULL, EVP_sha256(), NULL, key) == 1;
+    if (ok) result = EVP_DigestVerify(ctx, sig.bytes, sig.len, message.bytes, message.len);
+    EVP_MD_CTX_free(ctx);
+    EVP_PKEY_free(key);
```
Jaruso, line 232: This collapses malformed DER and an ordinary invalid signature into the same `auth_fail` status. `EVP_DigestVerify` returns `1` for valid, `0` for a correctly processed but invalid signature, and a negative value for an error/malformed verification path. #431 explicitly requires malformed DER coverage and asks the oracle to distinguish malformed input from verification failure where EVP exposes a stable distinction. Please branch on all three outcomes, for example:
```c
if (!ok) return status("oracle_error");
if (result == 1) return status("ok");
if (result == 0) return status("auth_fail");
return status("malformed");
```
Then update the Zig malformed-DER case to expect `.malformed` while retaining `.auth_fail` for a well-formed mutated signature.

Jaruso (commented):
The architecture and registry integration are headed in the right direction, but two correctness issues currently violate the oracle contract and should be fixed before merge.

1. The harness stdout cap is smaller than the oracle's own maximum valid AEAD response. A 4096-byte plaintext is accepted by `EVP_ORACLE_MAX_INPUT`, but `aead-seal` emits 8192 hex characters plus status, separator, 32 tag characters, and newline, so the valid response exceeds the 8192-byte process limit and fails in the harness.
2. ECDSA verification collapses OpenSSL's malformed-signature result into `auth_fail`. `EVP_DigestVerify` distinguishes `0` (validly encoded but incorrect signature) from a negative return (verification/error path, including malformed DER), and #431 explicitly requires malformed DER coverage and stable malformed-vs-authentication classification where EVP supports it.

Please also add the focused harness tests required by #431 for child-process abnormal/nonzero termination and stdout/stderr limit handling. The current code configures caps but does not exercise those failure paths, so regressions there would not be detected.

I reviewed all three changed files against #431. The appliance isolation appears structurally correct because the executable/module are created only under the existing OpenSSL-enabled general-profile branch, and the normal Zig differential test executable itself imports only the generated path rather than linking libcrypto.

Jaruso (commented):
Re-reviewed the updated head against issue #431 and the prior findings. No additional issues found.

The follow-up correctly:
- raises the harness stdout bound to cover the maximum valid hex-encoded AEAD response and adds a 4096-byte boundary test;
- distinguishes malformed ECDSA verification input from an ordinary invalid signature and updates the differential expectation;
- adds explicit coverage for nonzero exit, signal termination, and stdout/stderr cap failures.

I also rechecked the complete PR for the issue's primitive coverage, waiver replacement, bounded protocol behavior, secret-output handling, and build/dependency isolation. The oracle remains confined to the OpenSSL differential path, the appliance validation remains represented, and CI is green on `bb641b7`.

From review perspective, this is ready to merge.

PR #557, opened 2026-08-02
test: complete OpenSSL crypto differential coverage

Jaruso:
## Summary

- expands `tests/crypto_openssl_diff.zig` coverage vocabulary for resumption master secrets, per-ticket PSKs, PSK binders, protected tickets, and RSA-PSS accounting
- replaces stale PSK-binder #362 waivers with executable SHA-256/SHA-384 binder fixtures covering write/parse/derive/verify and mutation/length failures
- adds deterministic protected-ticket checks for AES-128-GCM, AES-256-GCM, and ChaCha20-Poly1305 through native provider, EVP oracle, `Protector.seal`, and `Protector.resolve`
- records RSA-PSS as an exact reviewed reference to the existing OpenSSL-derived `tests/crypto_vectors.zig` corpus from #413/#428

Refs #377

## Validation

- `zig build test-crypto-openssl`
- `zig build test`
- `zig build test -Dtls-profile=appliance`

### DIFF
--- a/tests/crypto_openssl_diff.zig
+++ b/tests/crypto_openssl_diff.zig
@@ -1253,6 +1253,11 @@ fn runResumptionForHash(allocator: std.mem.Allocator, hash: provider.Hash) !void
     const psk_a = psk_a_storage[0..digest_len];
     const psk_b = psk_b_storage[0..digest_len];
     try key_schedule.KeySchedule.resumptionPsk(cp, hash, rms, "\x01", psk_a);
+    var rms_mutated_storage = rms_storage;
+    rms_mutated_storage[0] ^= 0x01;
+    try key_schedule.KeySchedule.resumptionPsk(cp, hash, rms_mutated_storage[0..digest_len], "\x01", psk_b);
+    try testing.expect(!std.mem.eql(u8, psk_a, psk_b));
+
     var nonce_mut = [_]u8{0x01};
     nonce_mut[0] ^= 0x80;
     try key_schedule.KeySchedule.resumptionPsk(cp, hash, rms, &nonce_mut, psk_b);
@@ -1280,6 +1285,34 @@ const BinderFixture = struct {
     ext_data: []const u8,
     slot: pre_shared_key.BinderSlot,
 };
+
+const expected_binder_truncated_sha256 = hexBytes("010000700303111111111111111111111111111111111111111111111111111111111111111100000213010100004500290041001c00167469636b65742d6964656e746974792d73686132353610203040");
+const expected_binder_truncated_sha384 = hexBytes("010000800303111111111111111111111111111111111111111111111111111111111111111100000213010100005500290051001c00167469636b65742d6964656e746974792d73686133383410203040");
+const expected_binder_hash_sha256 = hexBytes("9136468c5ddea96b88cdfc816087bc23b820b42646335d609c82b99320a4364e");
+const expected_binder_hash_sha384 = hexBytes("e6f3b0ea1616e04d028eff2e3ed727e088b711cf985a60c0b02ea999522183becd01cd0f8a1392e7e2603f93cf2ce949");
+const expected_binder_sha256 = hexBytes("5d93fb0a6d879504d8ada12cb6484fc3dbe9e0ae6f46e3c3131fa4941d0ac057");
+const expected_binder_sha384 = hexBytes("d1b2ab1ddf233db1f82ae5748987b65180564038960391930320455f7cbfba5177d71099a59448c51bddb44cad526e10");
+
+fn expectedBinderTruncated(hash: provider.Hash) []const u8 {
+    return switch (hash) {
+        .sha256 => &expected_binder_truncated_sha256,
+        .sha384 => &expected_binder_truncated_sha384,
+    };
+}
+
+fn expectedBinderTranscriptHash(hash: provider.Hash) []const u8 {
+    return switch (hash) {
+        .sha256 => &expected_binder_hash_sha256,
+        .sha384 => &expected_binder_hash_sha384,
+    };
+}
+
+fn expectedBinderBytes(hash: provider.Hash) []const u8 {
+    return switch (hash) {
+        .sha256 => &expected_binder_sha256,
+        .sha384 => &expected_binder_sha384,
+    };
+}
 
 fn buildBinderFixture(hash: provider.Hash, identity: []const u8, binder: []const u8, out: []u8) !BinderFixture {
     var w = tls_core.messages.Writer{ .buf = out };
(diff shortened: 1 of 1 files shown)


tests/crypto_openssl_diff.zig
```diff
@@ -1152,6 +1200,405 @@ fn runTlsKeySchedule(allocator: std.mem.Allocator) !void {
...
+
+    var psk_a_storage: [provider.max_digest_len]u8 = undefined;
+    var psk_b_storage: [provider.max_digest_len]u8 = undefined;
+    const psk_a = psk_a_storage[0..digest_len];
+    const psk_b = psk_b_storage[0..digest_len];
+    try key_schedule.KeySchedule.resumptionPsk(cp, hash, rms, "\x01", psk_a);
+    var nonce_mut = [_]u8{0x01};
+    nonce_mut[0] ^= 0x80;
+    try key_schedule.KeySchedule.resumptionPsk(cp, hash, rms, &nonce_mut, psk_b);
+    try testing.expect(!std.mem.eql(u8, psk_a, psk_b));
```
Jaruso, line 1264: The issue explicitly requires a one-bit **RMS** mutation to change the derived per-ticket PSK. This currently proves transcript sensitivity and nonce sensitivity, but not RMS sensitivity. Please add a separate mutation of `rms` and derive with the same nonce, for example:

```zig
var rms_mutated_storage = rms_storage;
rms_mutated_storage[0] ^= 0x01;
try key_schedule.KeySchedule.resumptionPsk(
    cp,
    hash,
    rms_mutated_storage[0..digest_len],
    "\x01",
    psk_b,
);
try testing.expect(!std.mem.eql(u8, psk_a, psk_b));
```

Keep this separate from the nonce mutation so a regression in either input dependency is independently identified.

Jaruso (commented):
This is a strong closeout pass and the new registry accounting is headed in the right direction, but I found three acceptance-criteria gaps that should be addressed before #377 is considered complete:

1. The resumption fixture mutates the transcript and nonce, but never mutates the RMS input itself and proves the per-ticket PSK changes.
2. The binder fixture is generated and immediately treated as the expected framing. It does not pin the exact truncated ClientHello bytes/hash required to detect a truncation-boundary or framing drift.
3. The protected-ticket negative coverage stops at `Protector.resolve`; it does not compare native-provider and EVP-oracle authentication/malformed outcomes, and it does not exercise unsupported provider capability. The envelope/ciphertext/tag are also derived at runtime rather than pinned as deterministic expected fixture bytes.

CI is still in progress at the current head, but these are behavioral coverage omissions independent of the build result.

tests/crypto_openssl_diff.zig
```diff
@@ -1152,6 +1200,405 @@ fn runTlsKeySchedule(allocator: std.mem.Allocator) !void {
...
+    try hashBytes(hash, fixture.truncated, transcript_hash);
+    var tmp = testing.tmpDir(.{});
+    defer tmp.cleanup();
+    try tmp.dir.writeFile(testing.io, .{ .sub_path = "truncated-client-hello.bin", .data = fixture.truncated });
+    const truncated_path = try tmp.dir.realPathFileAlloc(testing.io, "truncated-client-hello.bin", allocator);
+    defer allocator.free(truncated_path);
+    const openssl_transcript_hash = try opensslDigestFile(allocator, "psk binder truncated ClientHello hash", hash, truncated_path);
+    defer allocator.free(openssl_transcript_hash);
+    try expectStage("psk binder truncated ClientHello hash", transcript_hash, openssl_transcript_hash);
+
```
Jaruso, line 1400: This compares two hashes of the same dynamically generated `fixture.truncated`, but it does not assert the **exact truncated ClientHello bytes** or a pinned expected transcript hash as required by #377. If `writeOffer`, the enclosing length patching, and `truncated_len` all drift together, both sides still agree and the test passes.

Please add checked-in SHA-256/SHA-384 fixture constants and assert them before the OpenSSL comparison, e.g.:

```zig
const expected_truncated = switch (hash) {
    .sha256 => &hexBytes("..."),
    .sha384 => &hexBytes("..."),
};
const expected_hash = switch (hash) {
    .sha256 => &hexBytes("...32-byte digest..."),
    .sha384 => &hexBytes("...48-byte digest..."),
};
try expectStage("psk binder exact truncated ClientHello", fixture.truncated, expected_truncated);
try expectStage("psk binder exact transcript hash", transcript_hash, expected_hash);
```

The same fixture should also pin the expected binder bytes, rather than only deriving them during the test. That is what makes offset/framing drift reviewable and reproducible.

tests/crypto_openssl_diff.zig
```diff
@@ -1152,6 +1200,405 @@ fn runTlsKeySchedule(allocator: std.mem.Allocator) !void {
...
+    wrong_key_storage[0] ^= 0x44;
+    var wrong_config = config;
+    wrong_config.key_bytes = wrong_key_storage[0..aead.keyLength()];
+    wrong_config.nonce_lease = null;
+    const wrong_snapshot = try wrong_keyring.buildSnapshot(&.{wrong_config}, cp.capabilities());
+    try wrong_keyring.install(wrong_snapshot);
+    var wrong_protector = ticket_protection.Protector{ .provider = cp, .keyring = &wrong_keyring, .limits = session.Limits.default };
+    try expectTicketResolveMiss(&wrong_protector, allocator, sealed);
+
+    var opened = try runEvpAeadOpen(allocator, aead, key, &nonce, &aad, parsed.ciphertext, parsed.tag);
```
Jaruso, line 1751: The negative ticket cases above only assert normalized misses at `Protector.resolve`. #377 also requires raw provider/oracle negative parity and an unsupported-provider-capability case. Right now the EVP oracle is only opened on the valid envelope, so an AEAD auth/malformed divergence can slip through while `resolve` still returns `false`.

Please exercise at least the mutated tag/ciphertext/AAD inputs through both layers and compare normalized outcomes, for example:

```zig
try testing.expectError(error.AuthenticationFailed, cp.aeadOpen(
    aead, key, &nonce, &aad, parsed.ciphertext, bad_tag_bytes, plaintext_out,
));
var oracle_bad = try runEvpAeadOpen(
    allocator, aead, key, &nonce, &aad, parsed.ciphertext, bad_tag_bytes,
);
defer oracle_bad.deinit(allocator);
try expectEvpStatus(&oracle_bad, .auth_failed);
```

Also build a keyring snapshot using capabilities with this AEAD disabled (or the existing test provider used elsewhere for capability denial) and assert the documented rejection. Finally, pin expected ciphertext/tag/envelope constants for each AEAD; constructing `expected_envelope` from the native result only proves internal composition, not deterministic fixture stability.

Jaruso (commented):
Re-review complete. The follow-up commit addresses all three prior findings:

- RMS sensitivity is now tested independently from nonce sensitivity.
- SHA-256/SHA-384 binder fixtures pin the exact truncated ClientHello, transcript hash, and binder bytes before comparing the production helpers and OpenSSL oracle.
- Protected-ticket fixtures now pin ciphertext/tag/envelope bytes for every supported AEAD, compare native and EVP authentication failures for tag/ciphertext/AAD mutations, and cover provider capability denial.

I did not find any additional correctness or acceptance-criteria gaps in the updated diff. The current CI run has one failing `Test appliance profile (ubuntu-latest)` job, but the failure is in the unrelated timing-sensitive `gateway_handlers.test.h3 proxy parked early 425 fails without origin retry after budget expires` test; the same appliance suite passed on macOS and Ubuntu ARM, and all directly relevant crypto, dependency-audit, unit-test, formatting, and security jobs passed. From a code-review standpoint, this is ready.

PR #560, opened 2026-08-02
crypto: guard #375's constant-time/zeroization findings against regression

Jaruso:
## Summary
- Extends `scripts/audit_crypto_boundary.zig` (`zig build audit-crypto-boundary`) with two more regression guards, using the same named-exception approach as its pre-existing #490 provider-boundary checks:
  - A raw `std.crypto.timing_safe.*`/`crypto.timing_safe.*` call reappearing anywhere in `src/tls`, `src/quic`, or `src/pki` (recursive, no exceptions — #375 already migrated every secret/authentication-derived comparison in these trees onto `crypto.secrets.constantTimeEqual`/`crypto.provider.constantTimeEqual`).
  - The specific ad hoc zero-and-free / raw-`secureZero`-spelling findings #375 fixed, reappearing in the exact files/functions they were found: `BoundedSecret.deinit` (`src/crypto/secrets.zig`), `ticket_key_snapshot.zig`'s four fixed sites (`OwnedSnapshot.deinit`, `loadFromFile`, `reserveNonceLeasesInFile`, `parse`'s `key_storage` `errdefer`), and `sni_provider.zig`'s `SignAdapter.release`.
- Deliberately **not** a project-wide ban on either raw spelling: #375's own audit matrix classifies several comparisons in this scope as public/attacker-controlled and correctly left as ordinary `std.mem.eql`, and dozens of call sites throughout `src/tls`/`src/quic`/`src/http` legitimately zero a stack-local buffer with no accompanying free at all. A mechanical "convert everything" guard would flag all of those, which is exactly what #375 and this issue both warn against.
- Rewords two `ticket_key_snapshot.zig` comments that quoted the legacy bug shape verbatim in prose (which tripped the new literal-pattern scan on comment text, not real code) — meaning unchanged.
- Updates `docs/CRYPTO_SECURITY_AUDIT.md`'s deferred-findings section to mark #554 done and describe the new enforcement, plus a `CHANGELOG.md` entry.

## Test plan
- [x] `zig build audit-crypto-boundary` passes against the current tree (no violations)
- [x] New unit tests (`firstForbidden` against the new forbidden sets) prove each pattern fires on the bad shape and not on the canonical/fixed one
- [x] Extended the existing end-to-end fixture-tree test with `src/crypto/secrets.zig`, `src/tls/ticket_key_snapshot.zig`, `src/tls/sni_provider.zig`, and `src/pki`, plus bypass cases reproducing each of #375's fixed findings (raw `timing_safe` in tls/quic/pki, `BoundedSecret.deinit`'s zero-then-plain-free, all four `ticket_key_snapshot.zig` shapes, `sni_provider.zig`'s raw-spelling revert) — each fails the audit, and the tree passes with zero violations once reset
- [x] `zig build test` (full suite) passes

Closes #554

### DIFF
--- a/scripts/audit_crypto_boundary.zig
+++ b/scripts/audit_crypto_boundary.zig
@@ -431,24 +431,31 @@ const dir_checks = [_]DirCheck{
 /// `crypto.secrets.constantTimeEqual` itself, in `src/crypto/secrets.zig`,
 /// which the `src/crypto` directory check below excludes by exact path.
 ///
-/// No trailing `.` on either needle (#554 review, first pass): a
-/// namespace-capture alias reads and calls the member off a *different*
-/// token than the qualifier —
+/// Matches the bare `timing_safe` token itself, not a qualified spelling
+/// (#554 review, second pass): a qualified needle can always be defeated by
+/// aliasing one *more* level up the path than the needle anticipates —
 ///
-///     const timing_safe = std.crypto.timing_safe;
-///     return timing_safe.eql(...);
+///     const timing_safe = std.crypto.timing_safe;  // defeats a dotted
+///     return timing_safe.eql(...);                 // "std.crypto.timing_safe." needle
 ///
-/// — so the declaration contains `std.crypto.timing_safe;`, never
-/// `std.crypto.timing_safe.`, and the call contains only `timing_safe.eql`,
-/// neither of which the dot-suffixed needle would catch. Dropping the
-/// trailing dot catches the qualified spelling regardless of what follows it
-/// (`.eql(`, `.compare(`, `;`, `,`, `)`) at the cost of also matching inside
-/// a hypothetical unrelated identifier that happens to start with
-/// `timing_safe` (e.g. a contrived `timing_safeguard`) — no such identifier
-/// exists in `std.crypto` or this project today.
+///     const std_crypto = std.crypto;                // defeats "std.crypto.timing_safe"/
+///     return std_crypto.timing_safe.eql(...);       // "crypto.timing_safe" too if the
+///                                                    // alias name doesn't itself end in
+///                                                    // "crypto"
+///
+/// — and Zig has no bound on how many hops of local aliasing a call site can
+/// introduce, so no finite list of qualified spellings can be complete.
+/// `timing_safe` cannot be renamed by an alias — Zig aliases bind a new name
+/// to a value, they do not let you address a namespace member under a
+/// different member name — so the literal identifier `timing_safe` must
+/// appear in the source at least once for any access to reach it, direct or
+/// aliased, at any depth. The identifier does not exist anywhere in this
+/// project or in `std.crypto` for any other purpose, so a bare, qualifier-
+/// independent match is both sound (matches only genuine `timing_safe`
+/// references) and complete (no aliasing depth evades it) rather than a
+/// finite guess at qualifier spellings.
 const timing_safe_forbidden = [_][]const u8{
-    "std.crypto.timing_safe",
-    "crypto.timing_safe",
+    "timing_safe",
 };
 
 /// The raw, non-canonical spelling of a secret-buffer zero. Reserved for
@@ -776,21 +783,127 @@ fn containsPlainFreeOf(window: []const u8, key: []const u8) bool {
     return false;
 }
 
-/// The buffer-identity token of the first zero call in `contents` whose
-/// zeroed buffer is later handed to an ordinary free/deinit instead of
-/// `secureZeroAndFree`/`secureZeroAndFreeAligned`, scoped to the rest of the
-/// enclosing function/declaration — or `null` if every zero call in the
-/// file is either not followed by a plain free of the same buffer, or is
-/// itself part of `secureZeroAndFree`/`secureZeroAndFreeAligned`'s own
-/// implementation (neither contains the literal substring `secureZero(`:
-/// the character right after `secureZero` in both is `A`, not `(`, the same
-/// exact-boundary trick `extractFunctionBody` uses to tell `sealPayload` and
-/// `sealPayloadWithProvider` apart).
-fn firstZeroThenPlainFree(contents: []const u8) ?[]const u8 {
+fn isSimpleIdent(text: []const u8) bool {
+    if (text.len == 0 or !isIdentStart(text[0])) return false;
+    for (text) |c| if (!isIdentCont(c)) return false;
+    return true;
+}
+
+fn isSimplePathExpr(text: []const u8) bool {
+    for (text) |c| {
+        if (!isIdentCont(c) and c != '.' and !std.ascii.isWhitespace(c)) return false;
+    }
+    return true;
+}
+
+/// The start of the statement containing position `at`: just past the
+/// nearest preceding `;` or `{`, or 0. Used to bound alias-declaration
+/// lookups to "the current statement," not an unrelated one earlier in the
+/// file that happens to share punctuation.
+fn statementStart(contents: []const u8, at: usize) usize {
+    var i = at;
+    while (i > 0) : (i -= 1) {
+        if (contents[i - 1] == ';' or contents[i - 1] == '{') return i;
+    }
+    return 0;
+}
+
+const max_aliases = 8;
+
+/// Local callee aliases of the zero helper — `const wipe =
+/// crypto.secrets.secureZero;` followed by `wipe(buf)` — bound anywhere in
+/// `contents` (#554 review, second pass: a check that only recognizes the
+/// literal `secureZero(` spelling is defeated by aliasing the function
+/// value itself before calling it). One hop: `NAME` must be a simple
+/// identifier bound directly to a bare `secureZero` reference (not
+/// `secureZeroAndFree`/`secureZeroAndFreeAligned` — excluded by the `next
+/// char is 'A'` check, same as the direct-call scan) by a simple dotted-path
+/// expression, not an arbitrary computation. Bounded to `max_aliases`
+/// entries.
+fn findSecureZeroAliases(contents: []const u8, out: *[max_aliases][]const u8) usize {
+    var count: usize = 0;
     var search_from: usize = 0;
-    while (std.mem.indexOfPos(u8, contents, search_from, "secureZero(")) |m| {
-        const open_paren = m + "secureZero(".len - 1;
+    while (count < max_aliases) {
+        const rel = std.mem.indexOfPos(u8, contents, search_from, "secureZero") orelse break;
+        search_from = rel + 1;
+        const after = rel + "secureZero".len;
+        if (after >= contents.len) continue;
+        if (contents[after] == 'A') continue; // ...AndFree/...AndFreeAligned
+        if (contents[after] != ';' and contents[after] != ',' and contents[after] != ')') continue;
+
+        const stmt = contents[statementStart(contents, rel)..rel];
+        const eq_rel = std.mem.lastIndexOfScalar(u8, stmt, '=') orelse continue;
+        if (!isSimplePathExpr(stmt[eq_rel + 1 ..])) continue;
+
+        const before_eq = std.mem.trimEnd(u8, stmt[0..eq_rel], " \t\r\n");
+        const kw = "const ";
+        const kw_at = std.mem.lastIndexOf(u8, before_eq, kw) orelse continue;
+        const name = std.mem.trim(u8, before_eq[kw_at + kw.len ..], " \t\r\n");
+        if (!isSimpleIdent(name)) continue;
+        out[count] = name;
+        count += 1;
+    }
+    return count;
+}
+
+/// Finds a local rebinding `const NAME = <expr>;` in `text` whose
+/// right-hand side (trimmed, one leading `&` stripped) is exactly `key` —
+/// one hop of "the buffer got a new local name" resolution (#554 review,
+/// second pass: `const doomed = secret_buf; allocator.free(doomed);` defeats
+/// a check keyed only on the zero call's own argument expression).
+fn resolveBufferRename(text: []const u8, key: []const u8) ?[]const u8 {
+    if (key.len == 0) return null;
+    var search_from: usize = 0;
+    while (std.mem.indexOfPos(u8, text, search_from, "const ")) |kw_pos| {
+        search_from = kw_pos + 1;
+        const after_kw = kw_pos + "const ".len;
+        const eq_rel = std.mem.indexOfScalarPos(u8, text, after_kw, '=') orelse continue;
+        const name = std.mem.trim(u8, text[after_kw..eq_rel], " \t\r\n");
+        if (!isSimpleIdent(name)) continue;
+        const semi_rel = std.mem.indexOfScalarPos(u8, text, eq_rel, ';') orelse continue;
+        var rhs = std.mem.trim(u8, text[eq_rel + 1 .. semi_rel], " \t\r\n");
+        if (rhs.len > 0 and rhs[0] == '&') rhs = std.mem.trim(u8, rhs[1..], " \t\r\n");
+        if (std.mem.eql(u8, rhs, key)) return name;
+    }
+    return null;
+}
+
+/// The position of the next call `NAME(` in `contents` at or after `from` —
+/// `NAME` immediately followed by `(` — or `null`. Shared by the direct
+/// `secureZero(` scan and the alias-callee scan below so both go through
+/// the same call-site recognition.
+fn findNextCall(contents: []const u8, from: usize, name: []const u8) ?usize {
+    var search_from = from;
+    while (std.mem.indexOfPos(u8, contents, search_from, name)) |rel| {
+        search_from = rel + 1;
+        const after = rel + name.len;
+        if (after < contents.len and contents[after] == '(') return rel;
+    }
+    return null;
+}
+
+/// The buffer-identity token of the first zero call — direct `secureZero(`
+/// or through a local callee alias (`findSecureZeroAliases`) — in `contents`
+/// whose zeroed buffer, or a local rename of it (`resolveBufferRename`), is
+/// later handed to an ordinary free/deinit instead of
+/// `secureZeroAndFree`/`secureZeroAndFreeAligned`, scoped to the rest of the
+/// enclosing function/declaration.
+fn firstZeroThenPlainFree(contents: []const u8) ?[]const u8 {
+    var alias_buf: [max_aliases][]const u8 = undefined;
+    const alias_count = findSecureZeroAliases(contents, &alias_buf);
+
+    if (scanZeroCalls(contents, "secureZero")) |key| return key;
+    for (alias_buf[0..alias_count]) |alias| {
+        if (scanZeroCalls(contents, alias)) |key| return key;
+    }
+    return null;
+}
+
+fn scanZeroCalls(contents: []const u8, trigger: []const u8) ?[]const u8 {
+    var search_from: usize = 0;
+    while (findNextCall(contents, search_from, trigger)) |m| {
         search_from = m + 1;
+        const open_paren = m + trigger.len;
         const args = extractCallArgs(contents, open_paren) orelse continue;
         const buf_expr = lastArgument(args);
         const window_start = open_paren + args.len + 2; // past the matching ')'
@@ -801,6 +914,9 @@ fn firstZeroThenPlainFree(contents: []const u8) ?[]const u8 {
         while (tokens.next()) |key| {
             if (key.len < 2) continue;
             if (containsPlainFreeOf(window, key)) return key;
+            if (resolveBufferRename(window, key)) |renamed| {
+                if (containsPlainFreeOf(window, renamed)) return renamed;
+            }
             // The container-field fallback: `secureZero(out.items)` zeroes
             // an `ArrayList`'s current contents, but the matching free is
             // `out.deinit()` on the container itself, not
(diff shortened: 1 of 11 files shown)


scripts/audit_crypto_boundary.zig
```diff
@@ -379,6 +417,92 @@ const dir_checks = [_]DirCheck{
...
+/// `src/quic`, and `src/pki` onto `crypto.secrets.constantTimeEqual` /
+/// `crypto.provider.constantTimeEqual`; a new raw call over either spelling
+/// reappearing anywhere in those trees is exactly the regression this guards
+/// against. The one legitimate raw call left project-wide is inside
+/// `crypto.secrets.constantTimeEqual` itself, in `src/crypto/secrets.zig`,
+/// which this check's directory scope (`src/tls`/`src/quic`/`src/pki`) never
+/// reaches.
+const timing_safe_forbidden = [_][]const u8{
+    "std.crypto.timing_safe.",
+    "crypto.timing_safe.",
```
Jaruso: [P1] Match namespace captures, not only direct member access. Because both needles end in `.`, this direct raw use passes the audit in every protected directory:

```zig
const timing_safe = std.crypto.timing_safe;
return timing_safe.eql([32]u8, expected, candidate);
```

The declaration contains `std.crypto.timing_safe;`, not `std.crypto.timing_safe.`, and the call contains only `timing_safe.eql`. That defeats the canonical-helper boundary this check is supposed to enforce. Please either forbid the namespace spelling without the trailing dot (for example `"std.crypto.timing_safe"` / `"crypto.timing_safe"`) or explicitly resolve aliases, and add an end-to-end fixture for this capture-and-call shape.

scripts/audit_crypto_boundary.zig
```diff
@@ -379,6 +417,92 @@ const dir_checks = [_]DirCheck{
...
+/// route every zero exclusively through the canonical wrapper because #375
+/// found and fixed a zero-then-plain-free defect there; *not* forbidden
+/// project-wide, since many other files legitimately call this directly on a
+/// stack-local buffer with nothing to free at all.
+const zeroization_forbidden = [_][]const u8{
+    "std.crypto.secureZero(",
+    "std_crypto.secureZero(",
+};
+
+const dir_checks_375 = [_]DirCheck{
```
Jaruso, line 473: [P1] Include the audited `src/crypto` consumers in this timing-safe guard. #554 requires coverage for the files in `docs/CRYPTO_SECURITY_AUDIT.md` outside the canonical helper implementations, and that matrix explicitly includes `src/crypto/rsa.zig`'s final authentication comparison. With only `src/tls`, `src/quic`, and `src/pki` here, this regression currently passes:

```zig
// src/crypto/rsa.zig
if (!crypto.timing_safe.eql([h_len]u8, expected, h_array))
    return error.InvalidInput;
```

A `src/crypto` directory check should fail closed, with a narrow named/exact exception only for the raw implementation inside `crypto.secrets.constantTimeEqual` (and any other genuinely reviewed helper implementation), plus a fixture proving a regression in `rsa.zig` is rejected.

scripts/audit_crypto_boundary.zig
```diff
@@ -379,6 +417,92 @@ const dir_checks = [_]DirCheck{
...
+        .rationale = "src/pki has no raw std.crypto.timing_safe call today (RSA-PSS's final hash comparison routes through crypto.secrets.constantTimeEqual; its EMSA-PSS structural checks are public/attacker-controlled and correctly left as ordinary std.mem.eql, per #375). A new raw timing_safe call anywhere in this tree needs the same audit-and-classify treatment #375 gave rsa.zig, not a silent reintroduction.",
+    },
+};
+
+/// Named-exception guards for the specific ad hoc zero-and-free /
+/// raw-secureZero-spelling findings #375 fixed (see the file-level doc
+/// comment's point 5). Reuses `file_checks`'/`file_checks_with_exceptions`'
+/// mechanism and struct types rather than inventing a new one.
+const file_checks_375 = [_]FileCheck{
+    .{
```
Jaruso, line 511: [P1] This does not satisfy the zero-and-free acceptance criterion because it protects only three files and four exact historical strings. A new ad hoc implementation anywhere else in the #375 scope passes, and even the same defect in `ticket_key_snapshot.zig` passes after a harmless local rename:

```zig
crypto.secrets.secureZero(secret_buf);
allocator.free(secret_buf); // no forbidden raw secureZero spelling or listed variable name
```

```zig
const storage = self.key_storage;
crypto.secrets.secureZero(std.mem.sliceAsBytes(storage));
self.allocator.free(storage);
```

#554 calls for the guard to reject a *new ad hoc zero-and-free implementation outside the canonical helpers in the audited scope*, using named file/function exceptions to avoid false positives. Please scan the relevant secret-owning cleanup functions/files for ordinary `free`/`deinit` paths (or otherwise pair zeroization with the subsequent free) and allowlist reviewed non-secret frees, rather than encoding only the old variable names. Add fixtures in a different audited TLS/QUIC/PKI file and a renamed-local variant of one historical finding.

Jaruso (commented):
Blocking findings: the implementation is well structured and the fixture expansion is substantial, but the guard currently proves only the exact spellings chosen by this PR rather than the enforcement contract in #554.

I found three bypass/scope gaps that leave both core acceptance criteria incomplete:

1. namespace capture can hide a raw `timing_safe` use from the suffix-based needles;
2. `src/crypto` consumers covered by the #375 matrix are outside the timing-safe directory scan;
3. ad hoc zero-and-free is checked only in three files and mostly by exact historical variable names, so a new implementation elsewhere—or the same defect after a local rename—passes.

The currently completed format/unit-test jobs are green at `0c72b0b`, but those tests do not exercise these cases. Please broaden the checks using the named file/function exception model requested by #554 and add end-to-end fixtures for each gap before marking the audit follow-up complete. The documentation and changelog should describe the broader enforcement only after those fixtures pass.

I would submit this as **Request changes**, but GitHub does not permit the PR author account to request changes on its own pull request.

Jaruso (commented):
Second-pass review of `fa94159b`: the production `secureZeroAndFree` migrations look ownership-correct, the `src/crypto` directory is now included, the original direct namespace-capture fixture is covered, and CI run 1152 is green. However, the enforcement still has three straightforward source-level bypasses, including one that can reintroduce the original `BoundedSecret.deinit` defect. These remain blocking for #554 because the PR is itself the durable regression guard.

scripts/audit_crypto_boundary.zig
```diff
@@ -379,6 +417,116 @@ const dir_checks = [_]DirCheck{
...
+///
+/// — so the declaration contains `std.crypto.timing_safe;`, never
+/// `std.crypto.timing_safe.`, and the call contains only `timing_safe.eql`,
+/// neither of which the dot-suffixed needle would catch. Dropping the
+/// trailing dot catches the qualified spelling regardless of what follows it
+/// (`.eql(`, `.compare(`, `;`, `,`, `)`) at the cost of also matching inside
+/// a hypothetical unrelated identifier that happens to start with
+/// `timing_safe` (e.g. a contrived `timing_safeguard`) — no such identifier
+/// exists in `std.crypto` or this project today.
+const timing_safe_forbidden = [_][]const u8{
```
Jaruso, line 457: [P1] The timing-safe guard still misses a parent-namespace alias, including a spelling already used elsewhere in this repository:

```zig
const std_crypto = std.crypto;
return std_crypto.timing_safe.eql([32]u8, expected, candidate);
```

Neither `std.crypto.timing_safe` nor `crypto.timing_safe` occurs in that source, so every directory check passes. The current fix only catches aliasing `timing_safe` itself; aliasing `std.crypto` one level earlier defeats it. Since there are no legitimate raw timing-safe uses outside the named implementation in `secrets.zig`, prefer matching the `timing_safe` namespace token independently of its qualifier (with the existing named exception), or explicitly recognize parent namespace aliases. Please add an end-to-end fixture using `std_crypto.timing_safe`.

scripts/audit_crypto_boundary.zig
```diff
@@ -489,6 +651,259 @@ fn blank(buf: []u8, needle: []const u8) void {
...
+/// enclosing function/declaration — or `null` if every zero call in the
+/// file is either not followed by a plain free of the same buffer, or is
+/// itself part of `secureZeroAndFree`/`secureZeroAndFreeAligned`'s own
+/// implementation (neither contains the literal substring `secureZero(`:
+/// the character right after `secureZero` in both is `A`, not `(`, the same
+/// exact-boundary trick `extractFunctionBody` uses to tell `sealPayload` and
+/// `sealPayloadWithProvider` apart).
+fn firstZeroThenPlainFree(contents: []const u8) ?[]const u8 {
+    var search_from: usize = 0;
+    while (std.mem.indexOfPos(u8, contents, search_from, "secureZero(")) |m| {
```
Jaruso: [P1] This is still identifier-equality matching rather than rename tracking. Both of these ad hoc zero-and-free implementations pass:

```zig
const wipe = crypto.secrets.secureZero;
wipe(secret_buf);
allocator.free(secret_buf);
```

```zig
crypto.secrets.secureZero(secret_buf);
const doomed = secret_buf;
allocator.free(doomed);
```

The first avoids the literal `secureZero(` search; the second changes the identifier between the zero and free, so `containsPlainFreeOf` never sees the extracted key. The latter directly contradicts the documented claim that rename indirection is covered. A conservative function-level rule—any secure-zero operation plus any plain free/deinit in the same cleanup function, with named reviewed exceptions—would be harder to bypass than trying to infer buffer identity lexically. Otherwise this needs alias propagation for both the callee and the buffer, plus fixtures for both shapes.

scripts/audit_crypto_boundary.zig
```diff
@@ -489,6 +651,259 @@ fn blank(buf: []u8, needle: []const u8) void {
...
+    .{
+        .dir = "src/crypto",
+        // secureZero/secureZeroAndFree/secureZeroAndFreeAligned's own
+        // implementations in secrets.zig legitimately zero then free/rawFree
+        // the same buffer — that pairing *is* the canonical helper.
+        // rawFree/vtable.free spellings don't match `containsPlainFreeOf`'s
+        // lowercase-`free(`/`.free(`/`.deinit(` patterns in the first place
+        // (capital F in `rawFree`), but excluding the file entirely avoids
+        // relying on that incidentally rather than by design.
+        .excluded_paths = &.{"src/crypto/secrets.zig"},
```
Jaruso, line 1382: [P1] Excluding all of `secrets.zig` from the structural scan leaves the exact original `BoundedSecret.deinit` regression reintroducible under a harmless local rename:

```zig
pub fn deinit(self: *BoundedSecret) void {
    const bytes = self.bytes;
    self.clearAll();
    const allocator = self.allocator orelse return;
    allocator.free(bytes);
}
```

This has no raw `secureZero` in `deinit`, does not contain the separately forbidden `.free(self.bytes)` string, and the general zero/free scan never opens this file. The audit therefore passes while the defect #554 specifically exists to prevent is back. Please scan a copy of `secrets.zig` with only `secureZeroAndFree`/`secureZeroAndFreeAligned` blanked as named exceptions, or add a dedicated `BoundedSecret.deinit` structural check that rejects any ordinary free of its backing allocation regardless of local naming. Add this renamed fixture to the end-to-end test.

scripts/audit_crypto_boundary.zig
```diff
@@ -489,6 +658,470 @@ fn blank(buf: []u8, needle: []const u8) void {
...
+/// or through a local callee alias (`findSecureZeroAliases`) — in `contents`
+/// whose zeroed buffer, or a local rename of it (`resolveBufferRename`), is
+/// later handed to an ordinary free/deinit instead of
+/// `secureZeroAndFree`/`secureZeroAndFreeAligned`, scoped to the rest of the
+/// enclosing function/declaration.
+fn firstZeroThenPlainFree(contents: []const u8) ?[]const u8 {
+    var alias_buf: [max_aliases][]const u8 = undefined;
+    const alias_count = findSecureZeroAliases(contents, &alias_buf);
+
+    if (scanZeroCalls(contents, "secureZero")) |key| return key;
```
Jaruso, line 997: [P1] The structural guard still recognizes only calls whose callee is literally `secureZero` or a discovered alias. #554's objective also covers a new ad hoc **manual zero-and-free / volatile-clear** implementation, so this passes unchanged:

```zig
fn release(allocator: std.mem.Allocator, secret_buf: []u8) void {
    @memset(secret_buf, 0);
    allocator.free(secret_buf);
}
```

That is precisely an implementation outside `crypto.secrets.secureZeroAndFree`, but neither `findSecureZeroAliases` nor `scanZeroCalls(..., "secureZero")` sees a trigger. Please include the manual-clear spellings from #375's inventory (`@memset`/reviewed volatile-clear forms) in the structural rule, with named exceptions for legitimate non-secret clearing, and add an end-to-end fixture in a new audited file.

scripts/audit_crypto_boundary.zig
```diff
@@ -489,6 +658,470 @@ fn blank(buf: []u8, needle: []const u8) void {
...
+
+fn scanZeroCalls(contents: []const u8, trigger: []const u8) ?[]const u8 {
+    var search_from: usize = 0;
+    while (findNextCall(contents, search_from, trigger)) |m| {
+        search_from = m + 1;
+        const open_paren = m + trigger.len;
+        const args = extractCallArgs(contents, open_paren) orelse continue;
+        const buf_expr = lastArgument(args);
+        const window_start = open_paren + args.len + 2; // past the matching ')'
+        if (window_start >= contents.len) continue;
```
Jaruso: [P1] Scanning only the text *after* the zero call misses the normal Zig `defer` formulation where the free appears first in source but executes second:

```zig
fn release(allocator: std.mem.Allocator, secret_buf: []u8) void {
    defer allocator.free(secret_buf);
    defer crypto.secrets.secureZero(secret_buf);
}
```

Defers execute LIFO, so this performs the prohibited zero-then-ordinary-free sequence at runtime. `window_start` begins after the second defer's `secureZero` call, therefore the earlier `free` is invisible and the audit passes. Inspect the whole enclosing function (or conservatively flag any zero/manual-clear plus plain free/deinit in that function, then use named exceptions) rather than relying on forward textual order. Please add this exact fixture.

scripts/audit_crypto_boundary.zig
```diff
@@ -489,6 +658,470 @@ fn blank(buf: []u8, needle: []const u8) void {
...
+        count += 1;
+    }
+    return count;
+}
+
+/// Finds a local rebinding `const NAME = <expr>;` in `text` whose
+/// right-hand side (trimmed, one leading `&` stripped) is exactly `key` —
+/// one hop of "the buffer got a new local name" resolution (#554 review,
+/// second pass: `const doomed = secret_buf; allocator.free(doomed);` defeats
+/// a check keyed only on the zero call's own argument expression).
```
Jaruso, line 941: [P1] The advertised one-hop rename handling returns only the **first** alias of `key`, so a second direct one-hop alias defeats it without any multi-hop indirection:

```zig
crypto.secrets.secureZero(secret_buf);
const retained_view = secret_buf;
_ = retained_view;
const doomed = secret_buf;
allocator.free(doomed);
```

`resolveBufferRename(window, "secret_buf")` returns `retained_view`; `scanZeroCalls` checks that one name, finds no free, and never examines `doomed`. Please iterate every matching rename (bounded if necessary) rather than returning the first, and add a fixture with one harmless alias preceding the freed alias.

Jaruso (commented):
The three second-pass reproductions are addressed, and the new tests plus CI are green. I still cannot approve the implementation against #554's acceptance criteria, though: the generalized enforcement remains tied to source ordering and a narrow definition of "zero," and the dedicated BoundedSecret rule does not actually enforce its stated canonical-helper invariant. The inline cases below all pass this audit while reintroducing the prohibited cleanup class. I recommend replacing the growing expression-identity heuristic with a conservative function-level rule (zero/manual-clear plus any plain free/deinit in either textual order), backed by named reviewed exceptions, and making BoundedSecret.deinit positively require the canonical helper rather than trying to enumerate noncanonical frees.

scripts/audit_crypto_boundary.zig
```diff
@@ -489,6 +658,470 @@ fn blank(buf: []u8, needle: []const u8) void {
...
+    return true;
+}
+
+/// True if `struct_body` frees `self.bytes` — or a local rename of it, one
+/// hop — any way other than `secureZeroAndFree`/`secureZeroAndFreeAligned`.
+/// `containsPlainFreeOf`'s `free(` search already excludes both (the
+/// character right after `secureZero` in each is `A`, not the case-sensitive
+/// lowercase `f` `containsPlainFreeOf` looks for immediately after `free`),
+/// so no separate exemption is needed for the canonical call itself.
+fn boundedSecretFreesBytesPlainly(struct_body: []const u8) bool {
```
Jaruso: [P1] This dedicated check does not enforce the stated invariant that `BoundedSecret` frees its backing storage **only** through `secureZeroAndFree`. `containsPlainFreeOf` deliberately ignores `rawFree` because of the capital `F`, so replacing the correct call with this passes and releases live secret bytes without wiping them:

```zig
pub fn deinit(self: *BoundedSecret) void {
    self.len = 0;
    const allocator = self.allocator orelse return;
    allocator.rawFree(
        self.bytes,
        .fromByteUnits(@alignOf(u8)),
        @returnAddress(),
    );
}
```

The whole `src/crypto/secrets.zig` file remains excluded from the general structural scan, and the literal file check does not forbid `rawFree`, so no other rule catches this. For this asset-specific case, positively require `BoundedSecret.deinit` to invoke `secureZeroAndFree(allocator, self.bytes)` (or an explicitly reviewed equivalent) and reject any direct `free`/`rawFree`/vtable-free of the backing allocation. Add a fixture for direct `rawFree`.

scripts/audit_crypto_boundary.zig
```diff
@@ -489,6 +678,571 @@ fn blank(buf: []u8, needle: []const u8) void {
...
+    var alias_buf: [max_aliases][]const u8 = undefined;
+    const alias_count = findSecureZeroAliases(contents, &alias_buf);
+
+    if (scanZeroCalls(contents, "secureZero")) |key| return key;
+    for (alias_buf[0..alias_count]) |alias| {
+        if (scanZeroCalls(contents, alias)) |key| return key;
+    }
+    // Manual zero-and-free / volatile-clear (#554 review, third pass): a
+    // clear implementation outside the canonical `secureZero` wrapper
+    // entirely, e.g. `@memset(secret_buf, 0)` followed by a plain free.
```
Jaruso, line 1008: [P1] This still does not cover the **manual zero-and-free / volatile-clear** half of #554. The only non-wrapper trigger is `@memset`, so an ordinary hand-written clear loop passes:

```zig
fn release(allocator: std.mem.Allocator, secret_buf: []u8) void {
    for (secret_buf) |*byte| byte.* = 0;
    allocator.free(secret_buf);
}
```

The same is true for a volatile-pointer loop. Neither source contains `secureZero(` nor `@memset(`, even though it is exactly a new ad hoc manual clear followed by a plain free outside `crypto.secrets`. The audit matrix's inventory explicitly searches `volatile`, and the issue objective explicitly names volatile-clear implementations.

Please recognize the reviewed loop/volatile-clear forms, or preferably apply a conservative whole-function policy in named secret-cleanup functions and allowlist reviewed non-secret cases. Add unit and end-to-end fixtures for a loop-based manual clear, not only `@memset`.

scripts/audit_crypto_boundary.zig
```diff
@@ -489,6 +678,571 @@ fn blank(buf: []u8, needle: []const u8) void {
...
+/// `.deinit(`/`.free(` called *on* `key` as receiver (the
+/// `ArrayList`/similar-container shape). Deliberately agnostic about what
+/// precedes `free(` (the allocator expression — `allocator.free(key)`,
+/// `self.allocator.free(key)`, and `fba.allocator().free(key)` are all
+/// caught the same way) by searching for the bare `free(` substring rather
+/// than requiring a specific receiver spelling.
+fn containsPlainFreeOf(window: []const u8, key: []const u8) bool {
+    if (key.len == 0) return false;
+    var search_from: usize = 0;
+    while (std.mem.indexOfPos(u8, window, search_from, "free(")) |p| {
```
Jaruso: [P1] The release matcher still accepts ordinary alternate spellings of the same prohibited free. It requires the key to appear immediately after lowercase `free(` and immediately before `)`, and deliberately ignores `rawFree`, so both of these pass:

```zig
crypto.secrets.secureZero(secret_buf);
allocator.free(
    secret_buf,
);
```

```zig
crypto.secrets.secureZero(secret_buf);
allocator.rawFree(
    secret_buf,
    .fromByteUnits(@alignOf(u8)),
    @returnAddress(),
);
```

The second is especially important because `rawFree` is part of #375's inventory. `src/crypto/secrets.zig` is already excluded from the general scan, so the canonical helper implementation can remain the named exception while direct `rawFree`/vtable-free use elsewhere is rejected.

Please parse the free call's arguments with whitespace/trailing-comma tolerance, and include direct `rawFree` (and any reviewed allocator-vtable release form) in the deallocator set. Add fixtures for multiline `Allocator.free` and direct `rawFree` in a new audited file.

scripts/audit_crypto_boundary.zig
```diff
@@ -489,6 +678,571 @@ fn blank(buf: []u8, needle: []const u8) void {
...
+        if (std.mem.eql(u8, trigger, "@memset")) {
+            if (!isZeroLiteral(lastArgument(args))) continue;
+            // `@memset(buf, 0)` alone cannot be told apart, lexically, from
+            // deliberately zero-filling test fixture data — `test` blocks
+            // do this constantly for reasons unrelated to secret hygiene,
+            // and pairing it with the test's own ordinary
+            // `defer allocator.free(buf)` is not the #375 regression class
+            // this trigger exists to catch. `secureZero`, unambiguous
+            // either way, still scans test blocks in full.
+            if (std.mem.startsWith(u8, contents[window_start..], "test ")) continue;
```
Jaruso: [P1] This blanket test exclusion is neither fail-closed nor a named/reviewed exception, while the generic production rule is also overbroad.

A real ad hoc cleanup in a test block now passes solely because it is a test:

```zig
test "secret fixture cleanup" {
    const secret_buf = try testing.allocator.alloc(u8, 32);
    defer testing.allocator.free(secret_buf);
    @memset(secret_buf, 0);
}
```

Conversely, an ordinary production zero-initialization is rejected even though it is not a secret wipe:

```zig
fn allocateFrame(allocator: std.mem.Allocator, len: usize) ![]u8 {
    const frame = try allocator.alloc(u8, len);
    errdefer allocator.free(frame);
    @memset(frame, 0); // initialization, not cleanup
    return frame;
}
```

#554 explicitly says the hard part is encoding the audit classification precisely with named per-file/per-function exceptions rather than a mechanical denylist. Please replace the all-tests escape hatch with specific reviewed exceptions, and classify production `@memset` sites so normal initialization does not become a CI violation. Add one clean production zero-init fixture and one violating test cleanup fixture.

Jaruso (commented):
Blocking follow-up review. The four third-pass reproductions are fixed, the existing production migrations still look correct, and CI run 1156 is green. However, the zeroization guard still does not meet #554's acceptance criterion for a new ad hoc manual zero-and-free/volatile-clear implementation: it recognizes only two clear spellings, recognizes only a narrow subset of release spellings, and the broad `@memset` handling is not the named/reviewed classification the issue calls for. I also found a file-global alias-scoping false positive that will make the guard reject unrelated functions. Please address the inline findings and add both unit-level and end-to-end fixtures. At this point I recommend simplifying toward a conservative whole-function clear/release rule with explicit per-file/per-function exceptions, rather than extending lexical expression identity further.

scripts/audit_crypto_boundary.zig
```diff
@@ -489,6 +678,571 @@ fn blank(buf: []u8, needle: []const u8) void {
...
+    return null;
+}
+
+/// The buffer-identity token of the first zero call — direct `secureZero(`
+/// or through a local callee alias (`findSecureZeroAliases`) — in `contents`
+/// whose zeroed buffer, or a local rename of it (`resolveBufferRename`), is
+/// later handed to an ordinary free/deinit instead of
+/// `secureZeroAndFree`/`secureZeroAndFreeAligned`, scoped to the rest of the
+/// enclosing function/declaration.
+fn firstZeroThenPlainFree(contents: []const u8) ?[]const u8 {
```
Jaruso, line 993: [P2] Callee aliases are collected file-wide and then every call with that bare name is scanned as a zero operation, even outside the alias's lexical scope. A local alias in one function can therefore make an unrelated same-named top-level function produce a false violation elsewhere:

```zig
fn actualWipe(buf: []u8) void {
    const scrub = crypto.secrets.secureZero;
    scrub(buf);
}

fn scrub(buf: []u8) void {
    transform(buf); // unrelated top-level function
}

fn release(allocator: std.mem.Allocator, buf: []u8) void {
    scrub(buf);
    allocator.free(buf);
}
```

`findSecureZeroAliases` records `scrub`, and this loop treats the call in `release` as the local alias even though that binding exists only inside `actualWipe`. That conflicts with the issue's requirement for a precise guard without constant false positives.

Please retain the enclosing declaration range with each alias and scan only calls inside that scope, or remove callee-alias inference in favor of explicit named cleanup-function policy. Add a fixture with a local alias and an unrelated same-named function in another scope.

PR #561, opened 2026-08-02
test: cover streamed Alt-Svc rollout paths

Jaruso:
## Summary
- Add regression coverage for streamed upstream responses emitting the gateway-owned effective Alt-Svc policy
- Add regression coverage for raw streamed upstream response heads stripping upstream Alt-Svc and Server while emitting a single effective Alt-Svc value

## Issue 257 reconciliation
Current main already contains the H3 rollout/lifecycle implementation for live-readiness-derived advertisement, restart-required listener-owned reload rejection, runtime drain hooks, effective-state metrics, and HTTP/3 rollout docs. This PR tightens the remaining response-path coverage for streamed proxy heads so the issue's Alt-Svc consistency contract is pinned alongside the existing buffered-path test.

## Validation
- zig build test --summary all
- zig build test-quic --summary all
- zig build test-integration --summary all

Refs #257

### DIFF
--- a/src/gateway_proxy_response.zig
+++ b/src/gateway_proxy_response.zig
@@ -439,5 +439,61 @@ test "writeBufferedUpstreamResponse strips upstream Alt-Svc and emits effective 
 
     const output = stream.getWritten();
     try std.testing.expect(std.mem.find(u8, output, "Alt-Svc: h3=\":443\"") == null);
+    try std.testing.expectEqual(@as(usize, 1), std.mem.count(u8, output, "Alt-Svc:"));
+    try std.testing.expectEqual(@as(usize, 1), std.mem.count(u8, output, "Alt-Svc: clear\r\n"));
     try std.testing.expect(std.mem.find(u8, output, "Alt-Svc: clear\r\n") != null);
+}
+
+test "writeStreamedUpstreamResponse emits effective Alt-Svc policy" {
+    var buf: [4096]u8 = undefined;
+    var stream = compat.fixedBufferStream(&buf);
+
+    try writeStreamedUpstreamResponse(
+        stream.writer(),
+        200,
+        "OK",
+        "text/plain",
+        null,
+        "req-stream-alt-svc",
+        &http.security_headers.SecurityHeaders.api,
+        "h3=\":8443\"; ma=120",
+        null,
+    );
+
+    const output = stream.getWritten();
+    try std.testing.expect(std.mem.find(u8, output, "Transfer-Encoding: chunked\r\n") != null);
+    try std.testing.expectEqual(@as(usize, 1), std.mem.count(u8, output, "Alt-Svc:"));
+    try std.testing.expectEqual(@as(usize, 1), std.mem.count(u8, output, "Alt-Svc: h3=\":8443\"; ma=120\r\n"));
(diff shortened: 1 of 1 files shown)


src/gateway_proxy_response.zig
```diff
@@ -441,3 +441,53 @@ test "writeBufferedUpstreamResponse strips upstream Alt-Svc and emits effective
...
+        true,
+        "req-stream-headers-alt-svc",
+        &http.security_headers.SecurityHeaders.api,
+        "clear",
+        null,
+    );
+
+    const output = stream.getWritten();
+    try std.testing.expect(std.mem.find(u8, output, "Alt-Svc: h3=\":443\"") == null);
+    try std.testing.expect(std.mem.find(u8, output, "Alt-Svc: clear\r\n") != null);
```
Jaruso, line 496: **Blocking test gap — this does not verify the “once” part of the contract.** These assertions prove that the upstream alternative is absent and the gateway-owned `clear` value is present, but a regression that emits `Alt-Svc: clear` twice would still pass. That leaves the issue’s “single effective policy / never contradictory alternatives” guarantee unpinned despite the test name and PR summary.

Please count the serialized header fields explicitly, for example:

```zig
try std.testing.expectEqual(
    @as(usize, 1),
    std.mem.count(u8, output, "Alt-Svc:"),
);
try std.testing.expectEqual(
    @as(usize, 1),
    std.mem.count(u8, output, "Alt-Svc: clear\r\n"),
);
```

Please add the same exact-count protection to the active streamed-policy test above as well; otherwise duplication of the active value would also remain undetected.

Jaruso (commented):
Changes needed before merge. The underlying #257 rollout/lifecycle implementation was already covered through the prior #539 review and its final production UDP/native-H3 lifecycle test. This PR correctly targets the remaining streamed response-head coverage, but the core claim that the gateway emits a **single** effective `Alt-Svc` value is not actually asserted: a regression that writes the effective header twice would still leave every new test green.

Please strengthen the streamed assertions to count exactly one `Alt-Svc` field. I would apply the same exact-count assertion to both new tests (and ideally the existing buffered test) so active and withdrawal policies are protected consistently.

The relevant unit-test jobs are currently green on Linux, macOS, and ARM; the broader CI workflow is still in progress. Because this is my own PR, GitHub does not allow a formal `REQUEST_CHANGES`, so I am submitting this as a blocking comment review.

Jaruso (commented):
Re-review of current head `689e5724f0d7a57afc559bcea734e9a3f6fefef5`: the prior blocker is fully addressed.

The buffered, active streamed, and raw streamed-header tests now each assert exactly one serialized `Alt-Svc` field and exactly one expected effective value. That protects both halves of the contract: upstream `Alt-Svc` cannot leak through, and the gateway-owned active/withdrawal policy cannot be duplicated. The updated head contains no unrelated changes, and I did not find another issue in the one-file diff.

Unit tests are green on Linux, macOS, and ARM, along with format, lint, security, H3 interop, and the completed build jobs. Several broader CI jobs are still running, so this is ready to merge once the workflow finishes successfully.

Because this is my own PR, GitHub does not permit a formal approval; this comment records the equivalent re-review result.

PR #562, opened 2026-08-02
test: fuzz ticket snapshot config invariants

Jaruso:
## Summary
- add a bounded ticket snapshot config fuzz helper for Snapshot.build inputs
- seed key count, duplicate id/key, AEAD capability, key length, validity window, nonce lease, generation boundary, stale-generation, ambiguous-window, and replacement overlap/non-overlap cases
- assert rejected snapshot/candidate/install paths leave the live keyring publication state unchanged, including current key nonce state and non-secret ledger contents
- verify partial Snapshot.build failures wipe every initialized copied key through a test-only thread-local cleanup probe, including temporary key cleanup before KeyRecord construction
- avoid copying or diagnosing secret-derived key fingerprints in publication-state assertions

Refs #376

## Tests

### DIFF
--- a/src/tls/ticket_protection.zig
+++ b/src/tls/ticket_protection.zig
@@ -5,6 +5,7 @@
 //! rotation scheduling, metrics export, and TLS PSK binder policy stay outside
 //! this module.
 
+const builtin = @import("builtin");
 const std = @import("std");
 const crypto = @import("crypto");
 const pre_shared_key = @import("pre_shared_key.zig");
@@ -37,7 +38,9 @@ const KeyDeinitProbe = struct {
         self.observed += 1;
     }
 };
-var test_key_deinit_probe: ?*KeyDeinitProbe = null;
+const TestKeyDeinitProbeStorage = if (builtin.is_test) struct {
+    threadlocal var probe: ?*KeyDeinitProbe = null;
+} else struct {};
 
 pub const ParseError = error{
     MalformedEnvelope,
@@ -282,7 +285,9 @@ const KeyRecord = struct {
     fn deinit(self: *KeyRecord) void {
         const key_len = self.key.len;
         self.key.deinit();
-        if (test_key_deinit_probe) |probe| probe.record(&self.key, key_len);
+        if (builtin.is_test) {
+            if (TestKeyDeinitProbeStorage.probe) |probe| probe.record(&self.key, key_len);
+        }
     }
 
     fn canEncryptAt(self: *const KeyRecord, now_unix_ms: i64) bool {
@@ -1554,7 +1559,6 @@ const LeaseState = struct {
 const LedgerState = struct {
     lease: LeaseState,
     aead: provider.Aead,
-    key_fingerprint: KeyFingerprint,
 };
 
 fn captureKeyringState(keyring: *ReloadableKeyRing) KeyringState {
@@ -1592,7 +1596,6 @@ fn captureKeyringState(keyring: *ReloadableKeyRing) KeyringState {
                 .has_lease = entry.has_lease,
             },
             .aead = entry.aead,
-            .key_fingerprint = entry.key_fingerprint,
         };
     }
     return state;
@@ -1646,7 +1649,6 @@ fn expectLedgerStateEqual(expected: LedgerState, entry: *const LeaseHighWater) !
         .has_lease = entry.has_lease,
     });
     try testing.expectEqual(expected.aead, entry.aead);
-    try testing.expectEqualSlices(u8, &expected.key_fingerprint, &entry.key_fingerprint);
 }
 
 fn expectLeaseStateEqual(expected: LeaseState, actual: LeaseState) !void {
@@ -1687,8 +1689,8 @@ fn expectPartialBuildWipesCopiedKey(input: []const u8) !void {
     if (expected_deinit_count == 0) return;
 
     var probe = KeyDeinitProbe{};
-    test_key_deinit_probe = &probe;
-    defer test_key_deinit_probe = null;
+    TestKeyDeinitProbeStorage.probe = &probe;
+    defer TestKeyDeinitProbeStorage.probe = null;
 
     const snapshot = Snapshot.build(testing.allocator, configs_slice, generation, caps) catch {
         try testing.expectEqual(expected_deinit_count, probe.observed);
@@ -1713,7 +1715,7 @@ fn initializedPrefixBeforeBuildFailure(configs: []const KeyConfig, caps: provide
         }
         initialized += 1;
     }
-    return 0;
+    return initialized;
 }
 
 fn byteAt(input: []const u8, idx: usize) u8 {
@@ -2031,6 +2033,20 @@ test "fuzz: ticket snapshot configs preserve bounded build invariants" {
     writeSnapshotSeedRecord(&partial_build_wipe, 0, .{ .aead = 0, .key_seed = 0x90, .id_seed = 0x90 });
     writeSnapshotSeedRecord(&partial_build_wipe, 1, .{ .aead = 1, .key_seed = 0xa0, .id_seed = 0xa0, .key_len_mode = 1 });
 
+    var ambiguous_windows = snapshotSeed(2, 0xff, 3, 0);
+    writeSnapshotSeedRecord(&ambiguous_windows, 0, .{
+        .aead = 0,
+        .key_seed = 0xd0,
+        .id_seed = 0xd0,
+        .lease_enabled = true,
+    });
+    writeSnapshotSeedRecord(&ambiguous_windows, 1, .{
+        .aead = 1,
+        .key_seed = 0xe0,
+        .id_seed = 0xe0,
+        .lease_enabled = true,
+    });
+
     var exact_max_keys = snapshotSeed(max_keys, 0xff, 3, 0);
     for (0..max_keys) |i| {
         writeSnapshotSeedRecord(&exact_max_keys, i, .{ .aead = @truncate(i % 3), .key_seed = @intCast(0x20 + i), .id_seed = @intCast(0x40 + i), .lease_enabled = false });
@@ -2072,6 +2088,7 @@ test "fuzz: ticket snapshot configs preserve bounded build invariants" {
     try expectSnapshotSeedOutcome(&invalid_window, .invalid_validity_window);
     try expectSnapshotSeedOutcome(&invalid_lease, .invalid_nonce_lease);
     try expectSnapshotSeedOutcome(&partial_build_wipe, .invalid_key_length);
+    try expectSnapshotSeedOutcome(&ambiguous_windows, .ambiguous_encryption_window);
     try expectSnapshotSeedOutcome(&exact_max_keys, .build_success);
     try expectSnapshotSeedOutcome(&max_plus_one_keys, .too_many_keys);
     try expectSnapshotSeedOutcome(&generation_overflow, .generation_overflow);
@@ -2089,6 +2106,7 @@ test "fuzz: ticket snapshot configs preserve bounded build invariants" {
         &invalid_window,
         &invalid_lease,
         &partial_build_wipe,
+        &ambiguous_windows,
         &exact_max_keys,
         &max_plus_one_keys,
         &generation_overflow,
@@ -2198,7 +2216,10 @@ fn expectSnapshotSeedOutcome(input: []const u8, outcome: SnapshotFuzzOutcome) !v
         .unsupported_capability => try testing.expectError(error.UnsupportedCapability, result),
         .invalid_validity_window => try testing.expectError(error.InvalidValidityWindow, result),
         .invalid_nonce_lease => try testing.expectError(error.InvalidNonceLease, result),
-        .ambiguous_encryption_window => try testing.expectError(error.AmbiguousEncryptionWindow, result),
+        .ambiguous_encryption_window => {
+            try testing.expectError(error.AmbiguousEncryptionWindow, result);
+            try expectPartialBuildWipesCopiedKey(input);
+        },
         .stale_generation, .generation_overflow, .non_overlapping_nonce_lease, .overlapping_nonce_lease => unreachable,
     }
 }
--- a/src/tls/ticket_protection.zig
+++ b/src/tls/ticket_protection.zig
@@ -42,6 +42,14 @@ const TestKeyDeinitProbeStorage = if (builtin.is_test) struct {
     threadlocal var probe: ?*KeyDeinitProbe = null;
 } else struct {};
 
+fn deinitTicketKey(key: *TicketKeySecret) void {
+    const key_len = key.len;
+    key.deinit();
+    if (builtin.is_test) {
+        if (TestKeyDeinitProbeStorage.probe) |probe| probe.record(key, key_len);
+    }
+}
+
 pub const ParseError = error{
     MalformedEnvelope,
     UnsupportedVersion,
@@ -264,7 +272,7 @@ const KeyRecord = struct {
             return error.InvalidValidityWindow;
 
         var key = TicketKeySecret.init(config.key_bytes) catch return error.InvalidKeyLength;
-        errdefer key.deinit();
+        errdefer deinitTicketKey(&key);
 
         const lease = if (config.nonce_lease) |lease_config|
             try NonceLease.init(lease_config)
@@ -283,11 +291,7 @@ const KeyRecord = struct {
     }
 
     fn deinit(self: *KeyRecord) void {
-        const key_len = self.key.len;
-        self.key.deinit();
-        if (builtin.is_test) {
-            if (TestKeyDeinitProbeStorage.probe) |probe| probe.record(&self.key, key_len);
-        }
+        deinitTicketKey(&self.key);
     }
 
     fn canEncryptAt(self: *const KeyRecord, now_unix_ms: i64) bool {
@@ -1700,6 +1704,7 @@ fn expectPartialBuildWipesCopiedKey(input: []const u8) !void {
 }
 
 fn initializedPrefixBeforeBuildFailure(configs: []const KeyConfig, caps: provider.Capabilities) usize {
+    if (configs.len == 0 or configs.len > max_keys) return 0;
     var initialized: usize = 0;
     for (configs, 0..) |config, i| {
         for (configs[0..i]) |prior| {
@@ -1711,7 +1716,7 @@ fn initializedPrefixBeforeBuildFailure(configs: []const KeyConfig, caps: provide
         if (!(config.not_before_unix_ms < config.encrypt_until_unix_ms and
             config.encrypt_until_unix_ms <= config.decrypt_until_unix_ms)) return initialized;
         if (config.nonce_lease) |lease| {
-            if (lease.start >= lease.end_exclusive) return initialized;
+            if (lease.start >= lease.end_exclusive) return initialized + 1;
         }
         initialized += 1;
     }
@@ -2052,6 +2057,9 @@ test "fuzz: ticket snapshot configs preserve bounded build invariants" {
         writeSnapshotSeedRecord(&exact_max_keys, i, .{ .aead = @truncate(i % 3), .key_seed = @intCast(0x20 + i), .id_seed = @intCast(0x40 + i), .lease_enabled = false });
     }
     var max_plus_one_keys = snapshotSeed(max_keys + 1, 0xff, 3, 0);
+    for (0..max_keys + 1) |i| {
+        writeSnapshotSeedRecord(&max_plus_one_keys, i, .{ .aead = @truncate(i % 3), .key_seed = @intCast(0x50 + i), .id_seed = @intCast(0x70 + i), .lease_enabled = false });
+    }
 
     var generation_overflow = snapshotSeed(1, 0xff, 3, 3);
     writeSnapshotSeedRecord(&generation_overflow, 0, .{ .aead = 0, .key_seed = 0xb0, .id_seed = 0xb0 });
@@ -2215,7 +2223,10 @@ fn expectSnapshotSeedOutcome(input: []const u8, outcome: SnapshotFuzzOutcome) !v
         },
         .unsupported_capability => try testing.expectError(error.UnsupportedCapability, result),
         .invalid_validity_window => try testing.expectError(error.InvalidValidityWindow, result),
-        .invalid_nonce_lease => try testing.expectError(error.InvalidNonceLease, result),
+        .invalid_nonce_lease => {
+            try testing.expectError(error.InvalidNonceLease, result);
+            try expectPartialBuildWipesCopiedKey(input);
+        },
         .ambiguous_encryption_window => {
             try testing.expectError(error.AmbiguousEncryptionWindow, result);
             try expectPartialBuildWipesCopiedKey(input);
--- a/src/tls/ticket_protection.zig
+++ b/src/tls/ticket_protection.zig
@@ -1335,13 +1335,64 @@ pub fn fuzzTicketIdentity(allocator: std.mem.Allocator, input: []const u8) !void
     }
 }
 
+const snapshot_fuzz_control_len = 4;
+const snapshot_fuzz_config_stride = 28;
+const SnapshotFuzzOutcome = enum {
+    build_success,
+    too_many_keys,
+    duplicate_key_id,
+    invalid_key_length,
+    unsupported_capability,
+    invalid_validity_window,
+    invalid_nonce_lease,
+    ambiguous_encryption_window,
+    generation_overflow,
+    overlapping_nonce_lease,
+};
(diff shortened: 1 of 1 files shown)


src/tls/ticket_protection.zig
```diff
@@ -1335,6 +1335,186 @@ pub fn fuzzTicketIdentity(allocator: std.mem.Allocator, input: []const u8) !void
     }
 }
 
+pub fn fuzzSnapshotConfig(allocator: std.mem.Allocator, input: []const u8) !void {
+    const config_count = if (input.len == 0) 0 else @as(usize, input[0] % (max_keys + 2));
+    var key_storage: [max_keys + 1][provider.max_aead_key_len + 1]u8 = undefined;
+    var configs: [max_keys + 1]KeyConfig = undefined;
+
+    for (configs[0..@min(config_count, configs.len)], 0..) |*config, i| {
+        const base = i * 13 + 1;
```
Jaruso: The decoder layout does not match the corpus layout. Bytes `0..3` are being used as global controls (`count`, capabilities, duplicate mode, generation), but key 0 starts at byte 1. Each key also reads through `base + 25` (`fuzzKeyId` consumes 16 bytes from `base + 10`) while the stride and Smith buffer reserve only 13 bytes per key. That couples adjacent configs and makes the named seeds decode incorrectly—for example, every `exact_one_*` seed has `input[2] == 0xff`, so this switch chooses length 0 and the seed dies as `InvalidKeyLength` instead of exercising a successful exact-length build.

Please give controls and records disjoint fixed-width regions and validate the seed outcomes. For example:

```zig
const control_len = 4;
const config_stride = 28; // fields 0..11, then a 16-byte id
const base = control_len + i * config_stride;
...
.id = fuzzKeyId(input, base + 12, @intCast(i)),
```

and size the Smith buffer as:

```zig
var input_buf: [control_len + (max_keys + 1) * config_stride]u8 = undefined;
```

Add deterministic assertions that each checked-in seed reaches its intended success/error class; otherwise a future layout change can silently turn a targeted seed into an unrelated early rejection again.

src/tls/ticket_protection.zig
```diff
@@ -1335,6 +1335,186 @@ pub fn fuzzTicketIdentity(allocator: std.mem.Allocator, input: []const u8) !void
...
+                configs[1].encrypt_until_unix_ms = configs[0].encrypt_until_unix_ms;
+            },
+            else => {},
+        }
+    }
+
+    const caps = fuzzCapabilities(input);
+    const generation = fuzzGeneration(input);
+    const snapshot = Snapshot.build(allocator, configs[0..@min(config_count, configs.len)], generation, caps) catch {
+        try expectKeyringStillPublishesCurrent(allocator);
```
Jaruso: This does not test the required cleanup property for the fuzzed rejection. `Snapshot.build` fails, then `expectKeyringStillPublishesCurrent` creates a separate keyring and deliberately rejects an unrelated empty config. A regression that stops wiping a key copied before a later config fails would still pass this helper.

Please run the actual fuzz-generated build through a zeroization probe (for example, a test allocator that inspects freed secret allocations, or a test-only FixedSecret/deinit hook) and include failures after at least one valid key has been initialized. The shape should be closer to:

```zig
const candidate = keyring.buildSnapshot(configs_slice, caps) catch |err| {
    try expectPublicationUnchanged(&keyring, before);
    try wipe_probe.expectAllReleasedSecretRegionsZero();
    _ = err;
    return;
};
```

The important part is that the assertion observes cleanup from this exact partial build, not a second independent rejection.

src/tls/ticket_protection.zig
```diff
@@ -1335,6 +1335,186 @@ pub fn fuzzTicketIdentity(allocator: std.mem.Allocator, input: []const u8) !void
...
+    defer keyring.deinit();
+    const current = sampleKeyConfigWithByte(keyId(0xee), .aes_128_gcm, .{ .prefix = .{ 0xee, 0, 0, 0 }, .start = 0, .end_exclusive = 4 }, 0x11);
+    try keyring.install(try keyring.buildSnapshot(&.{current}, testCapabilities()));
+    const before_current = keyring.current.?;
+    const before_generation = before_current.generation;
+    const before_next_generation = keyring.next_generation;
+    const before_ledger_len = keyring.ledger.items.len;
+
+    const candidate = Snapshot.build(allocator, configs[0..@min(config_count, configs.len)], generation, caps) catch return;
+    keyring.validateInstallCandidate(candidate) catch {
```
Jaruso: Two required properties are unreachable here.

First, the installed key is fixed to ID `0xee`, AES-128, and key bytes all `0x11`, while fuzzed key bytes are generated as `seed + j`. A fuzzed key can therefore never have the same fingerprint as the installed key. If its ID is `0xee`, validation returns `DuplicateKeyId` before checking the lease; otherwise there is no same-ID ledger entry. Consequently `OverlappingNonceLease` for a replacement can never be reached.

Second, `validateInstallCandidate` is intentionally a dry run, so it cannot exercise the atomic publication/consumption behavior of a rejected `install`.

Please add an explicit replacement mode that derives the installed baseline from the candidate's same `id`/AEAD/key bytes, gives the baseline a prior lease, then selects overlapping and non-overlapping candidate ranges. Exercise the real operation:

```zig
const before = captureKeyringState(&keyring);
keyring.install(candidate) catch |err| {
    try testing.expectEqual(error.OverlappingNonceLease, err); // for the targeted mode
    try expectKeyringStateEqual(before, &keyring);
    return; // install consumed candidate
};
```

For general fuzz cases the expected error can remain unconstrained, but the permanent overlap seed should assert this exact class so the coverage cannot regress.

Jaruso (commented):
Changes are required before this is ready to merge.

The bounded structure is a good direction and CI is green, but the current harness does not actually exercise several of the ticket snapshot target's required cases:

1. The input decoder and checked-in corpus use incompatible layouts. Most named seeds are rejected at an earlier, unintended stage, so the deterministic smoke corpus does not cover valid per-AEAD builds, duplicate IDs, invalid windows/leases, or generation overflow as labeled. There is also no checked-in exact-`max_keys` seed.
2. The required overlapping replacement-lease path is unreachable with the fixed current key, and `validateInstallCandidate` does not exercise the transactional rejection paths in `install` that can actually publish state.
3. The failed-build path does not verify that partially copied key material was wiped; it runs an unrelated empty-config rejection against a fresh keyring.

I left implementation-shaped suggestions inline. Please also replace `Fixes #376` with `Refs #376` / `Part of #376` (or link a narrower child issue). #376 still includes the shared fuzzing contract, provider AEAD/KEX/signature/signing-key/shared-secret targets, typed diagnostics, CI wiring, and longer-run documentation; merging this one snapshot target must not close the entire story.

src/tls/ticket_protection.zig
```diff
@@ -1335,6 +1335,333 @@ pub fn fuzzTicketIdentity(allocator: std.mem.Allocator, input: []const u8) !void
...
+    if (candidate_lease.start >= candidate_lease.end_exclusive) return;
+
+    var keyring = ReloadableKeyRing.init(allocator);
+    defer keyring.deinit();
+
+    var baseline = candidate_config;
+    baseline.nonce_lease = .{
+        .prefix = candidate_lease.prefix,
+        .start = 0,
+        .end_exclusive = @max(candidate_lease.start + 1, candidate_lease.end_exclusive),
```
Jaruso: `expect_overlap` does not affect the lease construction, so this helper always creates an overlap. For every valid candidate lease, `max(candidate_lease.start + 1, candidate_lease.end_exclusive)` is strictly greater than `candidate_lease.start`; with the same prefix, `rangesOverlap(0, baseline_end, candidate_start, candidate_end)` is therefore always true. As a result, `exerciseReplacementInstall(..., false)` never exercises a successful non-overlapping replacement—it merely accepts whatever rejection occurs.

Please construct the prior range from the expected mode and assert both outcomes. For example:

```zig
const baseline_end = if (expect_overlap)
    candidate_lease.end_exclusive
else blk: {
    if (candidate_lease.start == 0) return;
    break :blk candidate_lease.start;
};
baseline.nonce_lease = .{
    .prefix = candidate_lease.prefix,
    .start = 0,
    .end_exclusive = baseline_end,
};

if (expect_overlap) {
    try testing.expectError(error.OverlappingNonceLease, keyring.install(candidate));
    try expectKeyringStateEqual(before, &keyring);
} else {
    try keyring.install(candidate);
    try testing.expect(keyring.current != before.current);
}
```

A permanent non-overlap seed such as prior `[0, 1)` followed by candidate `[1, 3)` would ensure valid key rotation remains expressible.

src/tls/ticket_protection.zig
```diff
@@ -1335,6 +1335,333 @@ pub fn fuzzTicketIdentity(allocator: std.mem.Allocator, input: []const u8) !void
...
+        return;
+    };
+    if (expect_overlap) return error.TestUnexpectedResult;
+}
+
+const KeyringState = struct {
+    current: ?*Snapshot,
+    current_generation: u64,
+    next_generation: u64,
+    ledger_len: usize,
```
Jaruso, line 1539: This state snapshot omits the mutable data that the atomic-rejection property is intended to protect. Pointer equality, generation, and ledger length would still pass if a rejected install changed an existing ledger entry's `prefix`, `end_exclusive`, or `has_lease`, or changed the current key's nonce cursor/end before returning the error.

Please capture and compare the bounded contents, not only container identity/length. At minimum, record for every current key and ledger entry:

```zig
const LeaseState = struct {
    key_id: KeyId,
    prefix: [4]u8,
    next_counter: u64,
    end_exclusive: u64,
    has_lease: bool,
};
```

Also retain the current key metadata relevant to publication (`id`, AEAD, validity windows, and whether a lease exists). Then compare all entries after every rejected build/validation/install. That makes the test fail on a partial nonce-high-water or window mutation even when `current` and `ledger.items.len` remain unchanged.

src/tls/ticket_protection.zig
```diff
@@ -1335,6 +1335,333 @@ pub fn fuzzTicketIdentity(allocator: std.mem.Allocator, input: []const u8) !void
...
+    var configs: [max_keys + 1]KeyConfig = undefined;
+    const configs_slice = decodeSnapshotFuzzConfigs(input, &key_storage, &configs);
+    const caps = fuzzCapabilities(input);
+    const generation = fuzzGeneration(input);
+    const should_copy_first_key = configs_slice.len > 1 and firstConfigCopiesKey(configs_slice[0], caps);
+    const copied_key = if (should_copy_first_key) configs_slice[0].key_bytes else "";
+
+    const snapshot = Snapshot.build(fba.allocator(), configs_slice, generation, caps) catch {
+        if (should_copy_first_key) {
+            try testing.expect(std.mem.indexOf(u8, &backing, copied_key) == null);
```
Jaruso: This is not a valid zeroization probe. `Snapshot.build` eventually calls ordinary `allocator.free(keys)` on the failed allocation. As documented in `crypto/secrets.zig`, `Allocator.free` writes `undefined`/poison in safety builds before dispatching to the backing allocator. Therefore, even if the `for (keys[0..initialized]) |*key| key.deinit()` cleanup were deleted, the original byte sequence would disappear and this `indexOf == null` assertion would still pass.

It also checks only the first copied key, so a failure after two or more initialized records could leave a later key uncleared without detection.

Please observe each initialized `TicketKeySecret` immediately after `deinit` and before the containing `KeyRecord` allocation is freed—for example through a test-only key-deinit/cleanup probe—and assert that every logical key byte is zero and that the observed count equals the number initialized before the forced failure. The permanent regression should fail if any individual `key.deinit()` call is removed; searching poisoned freed storage cannot provide that guarantee.

src/tls/ticket_protection.zig
```diff
@@ -1527,6 +1854,177 @@ test "fuzz helper preserves output on resolver allocation errors" {
...
+    writeSnapshotSeedRecord(&partial_build_wipe, 0, .{ .aead = 0, .key_seed = 0x90, .id_seed = 0x90 });
+    writeSnapshotSeedRecord(&partial_build_wipe, 1, .{ .aead = 1, .key_seed = 0xa0, .id_seed = 0xa0, .key_len_mode = 1 });
+
+    var exact_max_keys = snapshotSeed(max_keys, 0xff, 3, 0);
+    for (0..max_keys) |i| {
+        writeSnapshotSeedRecord(&exact_max_keys, i, .{ .aead = @truncate(i % 3), .key_seed = @intCast(0x20 + i), .id_seed = @intCast(0x40 + i), .lease_enabled = false });
+    }
+    var max_plus_one_keys = snapshotSeed(max_keys + 1, 0xff, 3, 0);
+
+    var generation_overflow = snapshotSeed(1, 0xff, 3, 3);
```
Jaruso, line 2064: The target specification explicitly requires both stale and overflowed generation cases. Overflow now has an exact deterministic seed, but stale generation has no `SnapshotFuzzOutcome` variant, seed, or exact assertion; relying on Smith eventually mutating `input[3]` is not permanent replay/smoke coverage.

Please add a `.stale_generation` outcome and a valid candidate with generation `1` against an installed generation-`1` baseline, then assert:

```zig
try testing.expectError(error.StaleSnapshotGeneration, keyring.install(candidate));
try expectKeyringStateEqual(before, &keyring);
```

That ensures later changes cannot silently make the stale-generation path unreachable or reclassify it as another earlier rejection.

Jaruso (commented):
The first revision fixed the control/record layout, added exact seed-outcome assertions, covered exact `max_keys`, changed the PR to `Refs #376`, and made the overlap rejection reachable through the real `install` path. CI is green across the full matrix.

Four blocking gaps remain before this snapshot fuzz target fully enforces the #376/#363 contract:

1. the replacement helper still cannot exercise a successful non-overlapping lease rotation;
2. the transactional-state snapshot omits mutable nonce and ledger contents;
3. the partial-build wipe probe would pass even if key deinitialization were removed and only observes one copied key;
4. stale generation has no deterministic replay seed or exact outcome assertion.

Because this is the author's own PR, GitHub records this as a comment review rather than a formal request-changes review.

src/tls/ticket_protection.zig
```diff
@@ -1335,6 +1348,465 @@ pub fn fuzzTicketIdentity(allocator: std.mem.Allocator, input: []const u8) !void
...
+                .key_id = entry.key_id,
+                .prefix = entry.prefix,
+                .next_counter = 0,
+                .end_exclusive = entry.end_exclusive,
+                .has_lease = entry.has_lease,
+            },
+            .aead = entry.aead,
+            .key_fingerprint = entry.key_fingerprint,
+        };
+    }
```
Jaruso, line 1604: This reintroduces the secret-lifetime pattern that `LeaseHighWater.initInto` was specifically written to avoid. `key_fingerprint` is secret-derived, but it is copied into a `KeyringState` that is returned by value, passed by value to `expectKeyringStateEqual`, and never wiped. The later `expectEqualSlices` can also include fingerprint bytes in failure diagnostics, violating the target's secret-safe diagnostic contract.

Please either omit the fingerprint from this snapshot or give the snapshot explicit secret ownership. If it must be compared, use an in-place initialized state and wipe it:

```zig
fn captureKeyringState(keyring: *ReloadableKeyRing, out: *KeyringState) void {
    out.* = .{ ... };
    ...
    @memcpy(&out.ledger_entries[i].key_fingerprint, &entry.key_fingerprint);
}

fn deinit(self: *KeyringState) void {
    for (self.ledger_entries[0..self.ledger_count]) |*entry|
        secrets.secureZero(&entry.key_fingerprint);
    self.* = undefined;
}

var before: KeyringState = undefined;
captureKeyringState(&keyring, &before);
defer before.deinit();
```

Pass `*const KeyringState` rather than copying it again, and compare with a non-dumping assertion:

```zig
try testing.expect(secrets.constantTimeEqual(
    &expected.key_fingerprint,
    &entry.key_fingerprint,
));
```

Do not use `expectEqualSlices` for the fingerprint.

src/tls/ticket_protection.zig
```diff
@@ -1335,6 +1348,465 @@ pub fn fuzzTicketIdentity(allocator: std.mem.Allocator, input: []const u8) !void
...
+        if (!caps.supportsAead(config.aead)) return initialized;
+        if (config.key_bytes.len != config.aead.keyLength()) return initialized;
+        if (!(config.not_before_unix_ms < config.encrypt_until_unix_ms and
+            config.encrypt_until_unix_ms <= config.decrypt_until_unix_ms)) return initialized;
+        if (config.nonce_lease) |lease| {
+            if (lease.start >= lease.end_exclusive) return initialized;
+        }
+        initialized += 1;
+    }
+    return 0;
```
Jaruso: Returning `0` here misses the most important post-construction cleanup case. When every individual record builds successfully but `validateEncryptionWindows(keys)` rejects the completed array with `AmbiguousEncryptionWindow`, all keys have already been copied. The original build enters the catch path, but this model returns zero, so `expectPartialBuildWipesCopiedKey` exits without observing any deinit.

Because this helper is called only after the same deterministic build has already failed, reaching the end means the failure occurred after the per-record loop. Return the initialized count (or explicitly model `validateEncryptionWindows`) and add a permanent exact seed:

```zig
// Two distinct valid keys, both with leases and overlapping active windows.
var ambiguous_windows = snapshotSeed(2, 0xff, 3, 0);
writeSnapshotSeedRecord(&ambiguous_windows, 0, .{
    .aead = 0, .key_seed = 0xd0, .id_seed = 0xd0,
    .lease_enabled = true,
});
writeSnapshotSeedRecord(&ambiguous_windows, 1, .{
    .aead = 1, .key_seed = 0xe0, .id_seed = 0xe0,
    .lease_enabled = true,
});
try expectSnapshotSeedOutcome(&ambiguous_windows, .ambiguous_encryption_window);
```

That outcome should also assert that the cleanup probe observed both initialized keys.

src/tls/ticket_protection.zig
```diff
@@ -27,6 +27,17 @@ const key_fingerprint_domain = "tardigrade/tls-ticket/key-fingerprint/v1\x00";
...
+    observed: usize = 0,
+
+    fn record(self: *KeyDeinitProbe, key: *const TicketKeySecret, logical_len: usize) void {
+        for (key.bytes[0..logical_len]) |byte| {
+            if (byte != 0) @panic("ticket key was not zeroized before cleanup observation");
+        }
+        self.observed += 1;
+    }
+};
+var test_key_deinit_probe: ?*KeyDeinitProbe = null;
```
Jaruso: This test probe is currently a normal mutable module global, and every production `KeyRecord.deinit` now reads it. Besides adding test-only state and a branch to the crypto lifecycle, concurrent fuzz invocations in one process can race on the pointer and mix `observed` counts.

Please make the instrumentation compile-time test-only and thread-local. For example:

```zig
const builtin = @import("builtin");
threadlocal var test_key_deinit_probe: ?*KeyDeinitProbe = null;

fn deinit(self: *KeyRecord) void {
    const key_len = self.key.len;
    self.key.deinit();
    if (builtin.is_test) {
        if (test_key_deinit_probe) |probe| probe.record(&self.key, key_len);
    }
}
```

Since `builtin.is_test` is comptime-known, the probe read and branch are eliminated from non-test builds; `threadlocal` prevents parallel fuzz workers from corrupting each other's observation state.

Jaruso (commented):
The latest commit correctly adds successful non-overlapping rotation coverage, exact stale-generation coverage, deep publication-state comparison, and an immediate post-deinit cleanup probe. CI is green across formatting, unit tests, platform/profile builds, security, packaging, and interop.

Three blocking issues remain in the new test instrumentation:

1. the publication snapshot copies the secret-derived key fingerprint through unwiped return/pass-by-value stack state and may print it on assertion failure;
2. the cleanup model returns zero after every record was initialized, so `AmbiguousEncryptionWindow` failures still skip the wipe probe;
3. the mutable probe is installed in the normal `KeyRecord.deinit` path without a test-only/thread-local guard.

Because this is the author's own PR, GitHub records this as a comment review rather than a formal request-changes review.

src/tls/ticket_protection.zig
```diff
@@ -1335,6 +1353,462 @@ pub fn fuzzTicketIdentity(allocator: std.mem.Allocator, input: []const u8) !void
...
+        if (config.key_bytes.len != config.aead.keyLength()) return initialized;
+        if (!(config.not_before_unix_ms < config.encrypt_until_unix_ms and
+            config.encrypt_until_unix_ms <= config.decrypt_until_unix_ms)) return initialized;
+        if (config.nonce_lease) |lease| {
+            if (lease.start >= lease.end_exclusive) return initialized;
+        }
+        initialized += 1;
+    }
+    return initialized;
+}
```
Jaruso, line 1724: This model still disagrees with the first guard in `Snapshot.build`. `Snapshot.build` returns `TooManyKeys` immediately when `configs.len > max_keys`, before allocating or copying any `TicketKeySecret`. Here, however, a one-over input whose records happen to be individually valid walks all 17 configs and returns 17. The probe then expects 17 deinitializations but observes 0, so a valid bounded rejection becomes a false fuzz failure.

The checked-in `max_plus_one_keys` seed does not expose this because its unwritten records have zero-length keys and the model returns at config 0. Smith can still generate a valid-looking 17-record input.

Please mirror the outer build guard before modeling per-record initialization:

```zig
fn initializedPrefixBeforeBuildFailure(
    configs: []const KeyConfig,
    caps: provider.Capabilities,
) usize {
    if (configs.len == 0 or configs.len > max_keys) return 0;
    // existing per-record model...
}
```

It would also be useful to populate all `max_keys + 1` records in the permanent one-over seed so this ordering remains regression-tested rather than depending on random mutation.

src/tls/ticket_protection.zig
```diff
@@ -269,7 +283,11 @@ const KeyRecord = struct {
     }
 
     fn deinit(self: *KeyRecord) void {
+        const key_len = self.key.len;
         self.key.deinit();
+        if (builtin.is_test) {
+            if (TestKeyDeinitProbeStorage.probe) |probe| probe.record(&self.key, key_len);
```
Jaruso: The probe only observes fully constructed `KeyRecord`s that reach `KeyRecord.deinit`, but `KeyRecord.build` has another secret cleanup path. It copies `config.key_bytes` into the local `key`, then `NonceLease.init` may return `InvalidNonceLease`; that path runs the local `errdefer key.deinit()` before a `KeyRecord` exists, so this hook never fires.

Consequently the permanent `invalid_lease` seed does not verify the copied temporary key was wiped. `initializedPrefixBeforeBuildFailure` returns the number of prior records (zero for the current seed), and `expectPartialBuildWipesCopiedKey` exits without probing. Removing the local `errdefer`, or breaking its zeroization, would not fail this target.

Please funnel both cleanup paths through one test-observable helper, for example:

```zig
fn deinitTicketKey(key: *TicketKeySecret) void {
    const logical_len = key.len;
    key.deinit();
    if (builtin.is_test) {
        if (TestKeyDeinitProbeStorage.probe) |probe|
            probe.record(key, logical_len);
    }
}

// KeyRecord.build
errdefer deinitTicketKey(&key);

// KeyRecord.deinit
deinitTicketKey(&self.key);
```

Then model an invalid lease as `initialized + 1` because the current key was copied before lease validation, and explicitly call the wipe assertion for `.invalid_nonce_lease`. The expected count should cover the temporary failed record plus every previously initialized record.

Jaruso (commented):
The previous three instrumentation issues are fixed: fingerprint state is no longer copied into diagnostics, the cleanup probe is test-only/thread-local, and the ambiguous-window seed now verifies both copied keys are wiped.

Two cleanup-model gaps remain:

1. one-over-`max_keys` inputs are rejected before any key is copied, but the model can predict deinitialization for valid-looking records and turn a correct `TooManyKeys` rejection into a fuzz-harness failure;
2. an invalid nonce lease copies a temporary key and wipes it through `KeyRecord.build`'s local `errdefer`, which the `KeyRecord.deinit`-only probe cannot observe.

Most CI jobs are green; the current workflow run is still finishing several appliance/packaging jobs. Because this is the author's own PR, GitHub records this as a comment review rather than a formal request-changes review.

Jaruso (commented):
Re-reviewed the current head against the remaining cleanup findings and the snapshot-fuzz acceptance criteria. The latest commit fixes both outstanding issues:

- the cleanup model now mirrors `Snapshot.build`'s immediate empty/over-`max_keys` rejection and the permanent `max_keys + 1` seed contains otherwise valid records;
- temporary key material created inside `KeyRecord.build` now uses the same test-only cleanup observer, so invalid nonce-lease failures verify the local `errdefer` zeroization path as well as completed-record cleanup.

I also rechecked the validation order, ambiguous-window cleanup, stale/overflow generation cases, replacement overlap/non-overlap behavior, publication-state assertions, and secret-safe diagnostics. I found no further issues. The complete CI matrix is green.

This is ready to merge. GitHub records this as a comment review because the PR belongs to the authenticated user.

PR #567, opened 2026-08-02
crypto: add pure-Zig P-256 ECDH provider support

Jaruso:
## Summary
- implement pure-Zig secp256r1 key-share generation and ECDH shared-secret derivation behind CryptoProvider
- add P-256 scalar/point validation, entropy failure handling, and failure-output preservation coverage
- update crypto profile/docs/vector registries and extend the EVP oracle differential coverage for P-256 ECDH
- keep native TLS profile negotiation on X25519 until #335 wires secp256r1 into the live handshake matrix

## Validation
- zig build test-crypto
- zig build test-tls
- zig build -Dtls-profile=appliance test-crypto
- zig build -Dtls-profile=appliance test-tls
- git diff --check

Fixes #563


src/crypto/pure_zig.zig
```diff
@@ -405,10 +422,39 @@ fn deriveSharedSecretImpl(
...
+
+            var peer = P256.fromSec1(peer_public) catch return error.InvalidInput;
+            defer crypto.secureZero(u8, std.mem.asBytes(&peer));
+            peer.rejectIdentity() catch return error.InvalidInput;
+            const canonical = peer.toUncompressedSec1();
+            if (!std.mem.eql(u8, &canonical, peer_public)) return error.InvalidInput;
+
+            var shared_point = peer.mul(scalar, .big) catch return error.InvalidInput;
+            defer crypto.secureZero(u8, std.mem.asBytes(&shared_point));
+            const affine = shared_point.affineCoordinates();
```
Jaruso: **[P1] Scrub the affine shared-point copy.** `affineCoordinates()` materializes another full point representation: `affine.x` is the ECDH shared secret and `affine.y` is equally secret-derived. The current defers wipe `shared_point` and the serialized `shared_x`, but this `const affine` value itself remains in stack storage after return. That violates #563's requirement to wipe local point/shared-secret temporaries and the new documentation's cleanup guarantee.

Please make the value mutable and register an explicit wipe before serializing it, for example:

```zig
var affine = shared_point.affineCoordinates();
defer crypto.secureZero(u8, std.mem.asBytes(&affine));
var shared_x = affine.x.toBytes(.big);
defer crypto.secureZero(u8, &shared_x);
```

The wipe needs to cover both success and any future error path added after affine conversion.

tests/evp_oracle.c
```diff
@@ -303,6 +303,63 @@ static int cmd_ecdsa_p256_verify(int argc, char **argv) {
...
+static int cmd_p256_ecdh(int argc, char **argv) {
+    if (argc != 4) return status("malformed");
+    Blob scalar, peer;
+    if (!parse_hex(argv[2], &scalar) || !parse_hex(argv[3], &peer) || scalar.len != 32 || peer.len != 65 || peer.bytes[0] != 0x04) {
+        return status("malformed");
+    }
+
+    unsigned char pub[65];
+    size_t pub_len = sizeof(pub);
+    EVP_PKEY *private_key = ecdsa_p256_private_from_scalar(scalar.bytes, scalar.len, pub, &pub_len);
```
Jaruso, line 321: **[P1] Make the oracle reject non-canonical private scalars.** The reused `ecdsa_p256_private_from_scalar()` helper checks `BN_is_zero(priv)` but never verifies `priv < group_order`. OpenSSL's `EC_POINT_mul` accepts a 32-byte value greater than or equal to the P-256 order and effectively performs the multiplication modulo the order, so an input such as 32 bytes of `0xff` can produce a key and ECDH result instead of `malformed`.

That means this new oracle cannot independently confirm #563's required out-of-range-scalar rejection; the current differential negative case only tests zero. Retrieve the order with `EC_GROUP_get_order`, reject `BN_cmp(priv, order) >= 0` (and clear/free the order on every path), then add an exact-order or all-`0xff` case to `runP256EcdhNegative` and require `.malformed`. For example, the helper's validation should include the equivalent of:

```c
BIGNUM *order = BN_new();
if (order == NULL ||
    EC_GROUP_get_order(group, order, NULL) != 1 ||
    BN_is_zero(priv) ||
    BN_cmp(priv, order) >= 0) {
    BN_clear_free(order);
    /* existing cleanup */
    return NULL;
}
BN_clear_free(order);
```

src/crypto/pure_zig.zig
```diff
@@ -373,7 +370,27 @@ fn generateKeyShareImpl(
...
-        .secp256r1 => return error.UnsupportedCapability,
+        .secp256r1 => {
+            if (public_out.len != provider.Group.secp256r1.publicKeyLength()) return error.InvalidInput;
+            if (private_out.len != provider.max_private_scalar_len) return error.InvalidInput;
+
+            var scalar_bytes: [provider.max_private_scalar_len]u8 = undefined;
+            defer crypto.secureZero(u8, &scalar_bytes);
+
+            var attempts: usize = 0;
+            while (attempts < p256_keygen_attempts) : (attempts += 1) {
```
Jaruso, line 381: **[P2] Exercise exhaustion of this rejection bound.** The implementation has the required fixed 16-attempt limit, but the new tests only cover an entropy callback that returns an error immediately. They do not cover a source that succeeds while repeatedly returning unusable scalar bytes, which is the distinct behavior this loop was added to handle.

Please add a deterministic zero-filling (or exact-order-filling) entropy source and assert all of the contract here: `generateKeyShare` returns `error.EntropyFailure`, the source was called exactly `p256_keygen_attempts` times, and both caller output buffers remain unchanged. That locks down the reviewed bound and prevents a later change from becoming unbounded or publishing partial output.

Jaruso (commented):
Changes are needed before this is ready to merge. I reviewed the implementation against every acceptance criterion in #563 and inspected all seven changed files, the provider/profile/TLS policy interactions, the fixed and differential vectors, and the successful CI matrix.

The core ECDH behavior is otherwise well structured: the provider advertises secp256r1 only after both operations exist; key generation uses injected entropy with canonical non-zero scalar validation; derivation requires an exact canonical uncompressed SEC1 point, rejects malformed/off-curve/identity inputs, uses constant-time multiplication, and delays caller-output writes until success; profile/docs correctly keep native TLS negotiation disabled pending #335; and general/appliance CI plus the no-OpenSSL dependency audit are green.

Three acceptance gaps remain:

1. The affine ECDH shared-point copy is not scrubbed, so secret-derived coordinates remain in a local stack value after return.
2. The new EVP oracle path does not reject scalars greater than or equal to the P-256 order, so it cannot independently validate the required out-of-range-scalar behavior.
3. The fixed rejection-sampling bound is implemented but not exercised with a successful entropy source that repeatedly returns unusable scalar material.

I left detailed inline fixes for each. This review is submitted as COMMENT rather than REQUEST_CHANGES because the authenticated account owns the PR.

src/crypto/pure_zig.zig
```diff
@@ -953,6 +998,159 @@ test "X25519 rejects an all-zero (low-order) peer point as InvalidInput" {
...
+    var wrong_prefix = valid_peer;
+    wrong_prefix[0] = 0x02;
+    try testing.expectError(error.InvalidInput, cp.deriveSharedSecret(.secp256r1, &valid_scalar, &wrong_prefix, &out));
+    var off_curve = valid_peer;
+    off_curve[64] ^= 0x01;
+    try testing.expectError(error.InvalidInput, cp.deriveSharedSecret(.secp256r1, &valid_scalar, &off_curve, &out));
+    var noncanonical = valid_peer;
+    @memset(noncanonical[1..33], 0xff);
+    try testing.expectError(error.InvalidInput, cp.deriveSharedSecret(.secp256r1, &valid_scalar, &noncanonical, &out));
+    try testing.expectError(error.InvalidInput, cp.deriveSharedSecret(.secp256r1, &valid_scalar, valid_peer[0..64], &out));
```
Jaruso, line 1082: **[P2] Add direct coverage for the point-at-infinity/identity branch.** Issue #563 explicitly requires focused coverage for malformed, wrong-length, off-curve, **and infinity** peer points. These cases cover the first three, but none reaches `peer.rejectIdentity()` above. In Zig's P-256 implementation, uncompressed coordinates `(0, 0)` deserialize to the identity element, so add a fixture such as:

```zig
var identity = [_]u8{0} ** 65;
identity[0] = 0x04;
try testing.expectError(
    error.InvalidInput,
    cp.deriveSharedSecret(.secp256r1, &valid_scalar, &identity, &out),
);
```

Also retain the unchanged-output assertion. This locks down the explicit identity rejection rather than relying on the wrong-length or generic off-curve cases.

Jaruso (commented):
Second-pass review against issue #563 and the prior findings:

- The affine-coordinate secret copy is now explicitly wiped.
- The EVP helper now enforces `0 < scalar < group_order`, and the differential suite covers an out-of-range scalar.
- Rejection-sampling exhaustion now verifies the exact 16-call bound, `EntropyFailure`, and unchanged outputs.
- The replacement CI run is green across general/appliance, Linux/macOS/ARM, formatting, security, and the OpenSSL-free dependency audit.

I found one remaining acceptance-criteria coverage gap: the issue explicitly requires focused coverage for point-at-infinity rejection, but the current negative fixtures do not exercise the `peer.rejectIdentity()` branch. The implementation itself looks correct; please add the direct test described inline. Since this PR is owned by the authenticated account, this is submitted as a comment rather than a formal request-changes review.

Jaruso (commented):
Re-review complete: no remaining findings.

The final issue-level assurance gap is now covered directly: the uncompressed `(0, 0)` SEC1 fixture reaches the P-256 identity rejection path, returns `error.InvalidInput`, and the existing sentinel assertion verifies that derivation failures do not publish output. The prior affine-coordinate cleanup, bounded entropy-exhaustion coverage, and OpenSSL scalar-range validation remain correct.

I rechecked the complete PR against issue #563's acceptance criteria. Provider capability, deterministic generation and symmetry, scalar and SEC1 point validation, wrong-size/output-preservation behavior, temporary-secret cleanup, independent EVP parity, profile/documentation separation, and deferred TLS negotiation are all addressed. CI run 1166 is fully green, including general/appliance tests on Linux, macOS, and ARM plus the OpenSSL-free appliance dependency audit.

Ready to merge. (Submitted as COMMENT because GitHub does not permit approving one's own PR.)

PR #568, opened 2026-08-02
tls: [323-G3] make the shared TLS 1.3 handshake cipher/hash agile

Jaruso:
## Summary

Makes the transport-neutral TLS 1.3 handshake, transcript, key schedule, PSK/Finished path, and secret-export contract follow the negotiated cipher suite's hash family, so the native engine can complete record-mode handshakes for `TLS_AES_256_GCM_SHA384` and `TLS_CHACHA20_POLY1305_SHA256` in addition to the existing `TLS_AES_128_GCM_SHA256` baseline — without duplicating protocol state machines. QUIC packet/header protection is out of scope (separate child issue per #335), and X25519 remains the only key-exchange group (per #563).

- **`transcript.zig`**: replaced the fixed-SHA-256 `Transcript` with a hash-agile one (SHA-256/SHA-384). The negotiated hash isn't known when the very first transcript bytes arrive (server learns it while parsing ClientHello; client learns it only from the server's response, after `Core` has already fed those bytes in) — `update`/`rebindClientHello` transparently buffer at most one framed handshake message until `selectFamily` runs, then replay it into the real hash. Never panics on attacker-controlled input (`error.TranscriptOverflow` instead of an assert).
- **`key_schedule.zig`**: `KeySchedule` and its Finished/early-traffic-secret helpers now take an explicit `provider.Hash` and use `max_digest_len`-bounded storage instead of a fixed 32-byte schedule. The SHA-256 zero-PSK comptime fast path is preserved byte-for-byte (and stays within `scripts/audit_crypto_boundary.zig`'s exact-text allowlist); SHA-384 derives the same value through the provider at runtime.
- **`events.zig` / `transport.zig`**: new `negotiated_parameters` event (cipher suite + transcript hash), emitted exactly once per connection before any `handshake`-epoch traffic secret, on both full and resumed handshakes. `record_epoch_bridge.Bridge` consumes it to pick the right AEAD for `record_protection.TrafficKeys.derive` — no second AES-256/ChaCha record implementation.
- **`tls13_backend.zig`**: negotiates all three suites, validates the live `CryptoProvider`'s capability for the full suite tuple (AEAD + hash + HKDF) before selection, and threads the negotiated hash through every Finished/CertificateVerify/PSK-binder/resumption call site (including per-ticket hash for offered PSKs, since a resumption ticket's own hash can differ from the suite about to be negotiated).
- **`crypto/profile.zig`, `tls/crypto_profile.zig`, `docs/CRYPTO_PROVIDER.md`**: updated to reflect that AES-256-GCM, ChaCha20-Poly1305, SHA-384, and HKDF-SHA384 are now live (not just provider-capable) for the native appliance product profile.

## Test plan

- [x] `zig build test` — full suite passes, including new SHA-384 key-schedule/transcript known-answer vectors (computed independently via Python, not the code under test) and 3 new native record-mode client/server loopback tests (AES-128-GCM/SHA-256 baseline, AES-256-GCM/SHA-384, ChaCha20-Poly1305/SHA-256) verifying negotiated suite/hash, matching transcripts across roles, and successful encrypted application-data round trips.
- [x] `zig build audit-crypto-boundary` — passes; no new direct keyed-crypto call sites outside the provider seam.
- [x] Existing HRR, resumption, 0-RTT, and client-auth coverage in `tls13_backend_tests.zig` still passes unmodified (aside from mechanical `Digest`/`SecretSnapshot` API updates for the now-variable digest length).

Closes #564

### DIFF
--- a/src/tls/key_schedule.zig
+++ b/src/tls/key_schedule.zig
@@ -115,12 +115,17 @@ fn emptyTranscriptHashFor(hash: provider.Hash) [max_digest_len]u8 {
 }
 
 /// The "derived" early secret for the zero-PSK key-schedule chain, under
-/// whichever hash the negotiated suite selected. SHA-256 returns the
-/// precomputed public constant; every other hash derives it fresh through
-/// the provider (two HKDF calls, once per full handshake — not worth a
-/// second comptime/audit exemption).
-fn zeroPskDerivedEarlySecret(crypto_provider: provider.CryptoProvider, hash: provider.Hash) provider.HkdfError![max_digest_len]u8 {
-    var out: [max_digest_len]u8 = [_]u8{0} ** max_digest_len;
+/// whichever hash the negotiated suite selected, written into `out`
+/// (zeroed first so bytes beyond the active hash's digest length are never
+/// left uninitialized). SHA-256 copies the precomputed public constant;
+/// every other hash derives it fresh through the provider (two HKDF calls,
+/// once per full handshake — not worth a second comptime/audit exemption).
+/// Out-parameter style (#564 review), like every other retained-or-caller-
+/// owned secret in this file: writing directly into caller storage means
+/// there is never a separate successful-path local left unwiped after a
+/// by-value return.
+fn zeroPskDerivedEarlySecret(crypto_provider: provider.CryptoProvider, hash: provider.Hash, out: *[max_digest_len]u8) provider.HkdfError!void {
+    out.* = [_]u8{0} ** max_digest_len;
     switch (hash) {
         .sha256 => @memcpy(out[0..hash_len], &derived_early_secret_sha256),
         .sha384 => {
@@ -133,7 +138,6 @@ fn zeroPskDerivedEarlySecret(crypto_provider: provider.CryptoProvider, hash: pro
             try crypto_provider.hkdfExpandLabel(.sha384, &early_secret, "derived", empty_hash[0..n], out[0..n]);
         },
     }
-    return out;
 }
 
 pub const KeySchedule = struct {
@@ -156,12 +160,14 @@ pub const KeySchedule = struct {
         hash: provider.Hash,
         shared: []const u8,
         hello_transcript_hash: []const u8,
-    ) Error!KeySchedule {
+        out: *KeySchedule,
+    ) Error!void {
         const n = hash.digestLength();
         if (hello_transcript_hash.len != n) return error.InvalidSecretLength;
-        var derived_early = try zeroPskDerivedEarlySecret(crypto_provider, hash);
+        var derived_early: [max_digest_len]u8 = undefined;
         defer provider.secureZero(&derived_early);
-        return initFromEarlySecret(crypto_provider, hash, derived_early[0..n], shared, hello_transcript_hash);
+        try zeroPskDerivedEarlySecret(crypto_provider, hash, &derived_early);
+        try initFromEarlySecret(crypto_provider, hash, derived_early[0..n], shared, hello_transcript_hash, out);
     }
 
     /// PSK-resumed (`psk_dhe_ke`) handshake: the same key-schedule chain as
@@ -178,7 +184,8 @@ pub const KeySchedule = struct {
         psk: []const u8,
         shared: []const u8,
         hello_transcript_hash: []const u8,
-    ) Error!KeySchedule {
+        out: *KeySchedule,
+    ) Error!void {
         const n = hash.digestLength();
         if (psk.len != n) return error.InvalidSecretLength;
 
@@ -191,7 +198,7 @@ pub const KeySchedule = struct {
         const empty_hash = emptyTranscriptHashFor(hash);
         try crypto_provider.hkdfExpandLabel(hash, early_secret[0..n], "derived", empty_hash[0..n], derived_early[0..n]);
 
-        return initFromEarlySecret(crypto_provider, hash, derived_early[0..n], shared, hello_transcript_hash);
+        try initFromEarlySecret(crypto_provider, hash, derived_early[0..n], shared, hello_transcript_hash, out);
     }
 
     /// Shared continuation from a "derived" early secret (RFC 8446 §7.1)
@@ -204,47 +211,37 @@ pub const KeySchedule = struct {
         derived_early: []const u8,
         shared: []const u8,
         hello_transcript_hash: []const u8,
-    ) Error!KeySchedule {
+        out: *KeySchedule,
+    ) Error!void {
         const n = hash.digestLength();
         if (shared.len != shared_secret_len or derived_early.len != n or hello_transcript_hash.len != n)
             return error.InvalidSecretLength;
         const zeros = [_]u8{0} ** max_digest_len;
 
-        // handshake_secret, master_secret, client_handshake_traffic, and
-        // server_handshake_traffic are all retained in the returned
-        // KeySchedule on success, so each gets an errdefer (wipe only on
-        // failure) rather than an unconditional defer.
-        var handshake_secret: [max_digest_len]u8 = [_]u8{0} ** max_digest_len;
-        errdefer provider.secureZero(&handshake_secret);
-        try crypto_provider.hkdfExtract(hash, derived_early, shared, handshake_secret[0..n]);
+        // #564 review: every retained secret is derived directly into
+        // `out.*` — never a separate local later copied in by a by-value
+        // `return .{...}` — so there is no successful-path stack copy left
+        // unwiped once this returns. `out.*` is reset to its (zeroed)
+        // default field values first so a single `errdefer out.wipe()`,
+        // armed from that point on, covers every failure below uniformly,
+        // partial writes included.
+        out.* = .{ .provider = crypto_provider, .hash = hash };
+        errdefer out.wipe();
+
+        try crypto_provider.hkdfExtract(hash, derived_early, shared, out.handshake_secret[0..n]);
 
         // derived_handshake is purely transient — never retained — so it
         // always gets wiped, success or failure.
         var derived_handshake: [max_digest_len]u8 = [_]u8{0} ** max_digest_len;
         defer provider.secureZero(&derived_handshake);
         const empty_hash = emptyTranscriptHashFor(hash);
-        try crypto_provider.hkdfExpandLabel(hash, handshake_secret[0..n], "derived", empty_hash[0..n], derived_handshake[0..n]);
+        try crypto_provider.hkdfExpandLabel(hash, out.handshake_secret[0..n], "derived", empty_hash[0..n], derived_handshake[0..n]);
 
-        var master_secret: [max_digest_len]u8 = [_]u8{0} ** max_digest_len;
-        errdefer provider.secureZero(&master_secret);
-        try crypto_provider.hkdfExtract(hash, derived_handshake[0..n], zeros[0..n], master_secret[0..n]);
+        try crypto_provider.hkdfExtract(hash, derived_handshake[0..n], zeros[0..n], out.master_secret[0..n]);
 
-        var client_handshake_traffic: [max_digest_len]u8 = [_]u8{0} ** max_digest_len;
-        errdefer provider.secureZero(&client_handshake_traffic);
-        try crypto_provider.hkdfExpandLabel(hash, handshake_secret[0..n], "c hs traffic", hello_transcript_hash, client_handshake_traffic[0..n]);
+        try crypto_provider.hkdfExpandLabel(hash, out.handshake_secret[0..n], "c hs traffic", hello_transcript_hash, out.client_handshake_traffic[0..n]);
 
-        var server_handshake_traffic: [max_digest_len]u8 = [_]u8{0} ** max_digest_len;
-        errdefer provider.secureZero(&server_handshake_traffic);
-        try crypto_provider.hkdfExpandLabel(hash, handshake_secret[0..n], "s hs traffic", hello_transcript_hash, server_handshake_traffic[0..n]);
-
-        return .{
-            .provider = crypto_provider,
-            .hash = hash,
-            .handshake_secret = handshake_secret,
-            .master_secret = master_secret,
-            .client_handshake_traffic = client_handshake_traffic,
-            .server_handshake_traffic = server_handshake_traffic,
-        };
+        try crypto_provider.hkdfExpandLabel(hash, out.handshake_secret[0..n], "s hs traffic", hello_transcript_hash, out.server_handshake_traffic[0..n]);
     }
 
     pub const ApplicationSecrets = struct {
@@ -265,19 +262,16 @@ pub const KeySchedule = struct {
         }
     };
 
-    pub fn applicationSecrets(self: *const KeySchedule, finished_transcript_hash: []const u8) Error!ApplicationSecrets {
+    /// Out-parameter style (#564 review): both secrets are derived directly
+    /// into `out.*`, never a pair of locals copied in by a by-value return.
+    pub fn applicationSecrets(self: *const KeySchedule, finished_transcript_hash: []const u8, out: *ApplicationSecrets) Error!void {
         const n = self.digestLen();
         if (finished_transcript_hash.len != n) return error.InvalidSecretLength;
 
-        var client: [max_digest_len]u8 = [_]u8{0} ** max_digest_len;
-        errdefer provider.secureZero(&client);
-        try self.provider.hkdfExpandLabel(self.hash, self.master_secret[0..n], "c ap traffic", finished_transcript_hash, client[0..n]);
-
-        var server: [max_digest_len]u8 = [_]u8{0} ** max_digest_len;
-        errdefer provider.secureZero(&server);
-        try self.provider.hkdfExpandLabel(self.hash, self.master_secret[0..n], "s ap traffic", finished_transcript_hash, server[0..n]);
-
-        return .{ .client = client, .server = server, .len = n };
+        out.* = .{ .len = n };
+        errdefer out.wipe();
+        try self.provider.hkdfExpandLabel(self.hash, self.master_secret[0..n], "c ap traffic", finished_transcript_hash, out.client[0..n]);
+        try self.provider.hkdfExpandLabel(self.hash, self.master_secret[0..n], "s ap traffic", finished_transcript_hash, out.server[0..n]);
     }
 
     pub fn resumptionMasterSecret(
--- a/src/tls/key_schedule.zig
+++ b/src/tls/key_schedule.zig
@@ -1,4 +1,8 @@
-//! Protocol-neutral TLS 1.3 SHA-256 key schedule.
+//! Protocol-neutral TLS 1.3 key schedule, hash-agile across SHA-256 and
+//! SHA-384 (#564): every secret length and HKDF call follows the hash the
+//! caller names, which must always be `algorithms.transcriptHash(suite)` for
+//! the negotiated cipher suite — this module does not select or persist a
+//! hash of its own, so it can never disagree with the negotiated suite.
 //!
 //! This module has no QUIC, HTTP, socket, or record-layer types. QUIC and
 //! future TCP-record integrations both derive handshake/application traffic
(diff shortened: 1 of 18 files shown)


src/tls/transcript.zig
```diff
@@ -1,109 +1,349 @@
...
 
-pub const Transcript = struct {
-    hash: Hash = Hash.init(.{}),
+/// One framed handshake message's worth of pre-selection buffering — see
+/// "Bounding the pre-selection buffer" above. Exactly
+/// `tls13_backend.max_message_len` (8 KiB) plus its 4-byte handshake
+/// header — the same bound the reassembler already admits a single
+/// handshake message up to, so a legitimate ClientHello the engine accepts
+/// can never itself overflow this buffer before its suite is known.
+pub const max_pending_len: usize = 8 * 1024 + 4;
```
Jaruso, line 61: [P1] This one-message bound does not match the client call order. `sendClientHello()` records the complete ClientHello1 while the family is unresolved, and `planPskOffer()` permits that message to reach `max_message_len`. Then `drainInput()` calls `Core.acceptReceived()` / `Core.acceptHelloRetryRequest()`, which appends the complete ServerHello/HRR, **before** `onServerHello()` / `onHelloRetryRequest()` calls `selectFamily()`. A legal near-8 KiB ClientHello plus any nontrivial server response therefore hits `TranscriptOverflow`; the comment above says `drainInput` selects first, but the implementation currently does the opposite. Please select the validated ServerHello/HRR suite before Core records that message (preferred), or size this storage for both pending messages, and add max-size ClientHello1 + ordinary ServerHello and HRR regressions.

src/tls/tls13_backend.zig
```diff
@@ -1542,6 +1568,33 @@ pub const Tls13Backend = struct {
...
+    /// before this side ever derives a secret under it. Ordinary policy
+    /// negotiation (`negotiation.zig`) already restricts both sides to
+    /// suites present in `self.policy.cipher_suites`; this is the
+    /// belt-and-braces check that the live provider backing that policy
+    /// actually agrees, so a provider/policy mismatch fails closed right at
+    /// selection instead of surfacing later as an opaque HKDF capability
+    /// error deep in the key schedule.
+    fn validateNegotiatedSuiteCapability(self: *const Tls13Backend) HandshakeError!void {
+        const caps = self.crypto_provider.capabilities();
+        if (!tls_crypto_profile.supportsCipherSuite(caps, self.negotiated_cipher_suite)) return error.SecretExportFailed;
```
Jaruso, line 1595: [P1] This is a post-selection guard, not capability-driven negotiation. With a hand-written policy containing AES-256 and AES-128 and a provider that only supports AES-128, the unsupported suite can still be advertised/selected and this returns `SecretExportFailed` instead of removing AES-256 and negotiating AES-128. #564 explicitly requires capability absence to remove suites **before selection**. Please build an effective cipher-suite list from `self.policy.cipher_suites` intersected with the live provider capabilities and use it for both ClientHello emission and server negotiation. Add a partial-capability fallback test and a true no-overlap test that fails before any ServerHello or traffic-secret event.

src/tls/key_schedule.zig
```diff
@@ -123,40 +200,46 @@ pub const KeySchedule = struct {
...
 
-        var server_handshake_traffic: [hash_len]u8 = undefined;
+        var server_handshake_traffic: [max_digest_len]u8 = [_]u8{0} ** max_digest_len;
         errdefer provider.secureZero(&server_handshake_traffic);
-        try crypto_provider.hkdfExpandLabel(.sha256, &handshake_secret, "s hs traffic", &hello_transcript_hash, &server_handshake_traffic);
+        try crypto_provider.hkdfExpandLabel(hash, handshake_secret[0..n], "s hs traffic", hello_transcript_hash, server_handshake_traffic[0..n]);
 
         return .{
             .provider = crypto_provider,
+            .hash = hash,
```
Jaruso: [P1] The `errdefer`s above only wipe on failure. On success, these four local secret arrays are copied into the returned `KeySchedule`, and their original stack storage is left unwiped; `applicationSecrets()` repeats the same pattern with `client` and `server`. Security cannot rely on return-value optimization eliminating those copies, and this is the exact by-value temporary-copy case #564 says to avoid. Please change schedule/application-secret derivation to out-parameter APIs and derive directly into the destination with `errdefer out.wipe()`. The same pass should audit other secret-bearing by-value returns such as PSK-selection results.

src/tls/tls13_backend_tests.zig
```diff
@@ -2728,6 +2742,103 @@ fn directHarnessWithClientKeyShareMode(
...
+            server_crypto_provider,
+            fixtureIdentity(),
+            tls_backend.recordConfig(policy),
+        ),
+        .client_bridge = Bridge.init(client_bridge_crypto_provider, .tls_aes_128_gcm_sha256),
+        .server_bridge = Bridge.init(server_bridge_crypto_provider, .tls_aes_128_gcm_sha256),
+    };
+}
+
+fn expectNativeSuiteLoopback(comptime cipher_suite: tls_core.algorithms.CipherSuite, expected_digest_len: usize) !void {
```
Jaruso, line 3227: [P1] This only exercises a full, happy-path handshake with the default credential. #564's required coverage is explicitly per SHA-256 and SHA-384 for resumed/PSK-binder, HRR, and 0-RTT paths, and also calls for tampered/wrong-length Finished, capability removal/no-overlap, and Ed25519/ECDSA coverage where compatible. The pre-existing tests for those paths remain SHA-256-only, so the newly added SHA-384 integrated branches are unproven. Please parameterize the relevant HRR, resumption, 0-RTT, credential, and failure tests for `TLS_AES_256_GCM_SHA384` (plus the appropriate SHA-256 suites), including partial-provider fallback.

Jaruso (commented):
I reviewed all 18 changed files against #564 and checked the current CI run at this head. The overall suite/hash threading, negotiated-parameters event, generic record-protection reuse, and SHA-384 known-answer work are substantial and internally coherent, but I found four blocking gaps against the issue's explicit architecture and coverage requirements:

1. The deferred transcript buffer reserves one maximum message, while the client actually buffers ClientHello1 **and** ServerHello/HRR before selecting the family. A legal large ClientHello can therefore fail with `TranscriptOverflow`.
2. Provider capability is validated only after policy negotiation has selected a suite. Unsupported suites are not removed from the offer/selection set, so a mutually supported fallback can be missed and the handshake fails as `SecretExportFailed` instead.
3. `KeySchedule` and `ApplicationSecrets` still copy secret-bearing local arrays into structs returned by value, leaving the original successful-path stack copies unwiped. This is the exact temporary-copy pattern #564 says to avoid.
4. The new integrated suite tests cover only full, happy-path record handshakes. The SHA-384 branches in resumption/binders, HRR, 0-RTT, client authentication, Finished failure handling, and capability-removal/no-overlap remain unexercised even though #564 requires those paths for both hash families.

The intended implementation shape for the two structural issues is roughly:

```zig
fn effectiveCipherSuites(
    self: *const Tls13Backend,
    out: *[native_cipher_suites.len]CipherSuite,
) []const CipherSuite {
    var len: usize = 0;
    const caps = self.crypto_provider.capabilities();
    for (self.policy.cipher_suites) |suite| {
        if (tls_crypto_profile.supportsCipherSuite(caps, suite)) {
            out[len] = suite;
            len += 1;
        }
    }
    return out[0..len];
}
```

Use that effective list for both ClientHello emission and server selection, so partial capability produces fallback and an empty intersection produces the ordinary no-mutual-suite failure before ServerHello.

```zig
pub fn init(..., out: *KeySchedule) Error!void {
    out.* = .{ .provider = crypto_provider, .hash = hash };
    errdefer out.wipe();

    try crypto_provider.hkdfExtract(
        hash,
        derived_early,
        shared,
        out.handshake_secret[0..n],
    );
    // Derive the remaining retained secrets directly into `out`.
}
```

`applicationSecrets` should follow the same out-parameter pattern. Please also audit the other secret-bearing by-value paths while making that change.

CI is green at `1573b78c1f7304b62b40b838e4998b75f423217d`, but the current tests do not reach the failing acceptance cases above. I submitted this as a comment review because the authenticated account is also the PR author; the findings are blocking from a code-review standpoint.

src/tls/tls13_backend.zig
```diff
@@ -2503,7 +2666,7 @@ pub const Tls13Backend = struct {
         const common = &ticket.common;
         if (common.isExpired(now_ms) or common.isNotYetValid(now_ms)) return false;
         if (!self.policy.containsCipherSuite(common.cipher_suite)) return false;
```
Jaruso, line 2687: [P1] Filter cached PSKs against the live provider here as well. `effectiveCipherSuites()` removes an unsupported suite from ClientHello, but this policy-only gate can retain a ticket for that same suite. With a configured policy containing AES-256 and AES-128, a provider without SHA-384/AES-256, and an early-data-capable AES-256 ticket, `planEarlyDataAttempt()` selects that ticket and `sendClientHello()` calls `KeySchedule.clientEarlyTrafficSecret(.sha384, ...)`; the provider then returns `UnsupportedCapability`/`SecretExportFailed` before the connection can fall back to a full AES-128 handshake. Without 0-RTT, the client still sends an identity for a suite it can no longer negotiate.

Please include provider capability in ticket eligibility, for example:

```zig
if (!tls_crypto_profile.supportsCipherSuite(
    self.crypto_provider.capabilities(),
    common.cipher_suite,
)) return false;
```

Add a regression with a stale SHA-384 ticket, client 0-RTT intent enabled, and a provider that only supports AES-128/SHA-256; the ticket should be removed before early-data planning and the full handshake should complete under AES-128.

Jaruso (commented):
Second-pass review of `278d0f1a60f949a69d41a61266093f6dd76a9d56` against the original findings and every #564 acceptance criterion.

The update correctly fixes two of the original structural blockers:

- the client now resolves the transcript family before recording a valid ServerHello/HRR, with near-maximum ClientHello regressions;
- retained key-schedule and application secrets are derived through out-parameters directly into their final storage and wiped on failures.

The normal full-handshake capability path is also substantially improved: both roles use an effective provider-filtered suite list, and SHA-384 integration coverage now includes HRR, resumption, 0-RTT, client authentication, and tampered Finished handling.

Five blocking gaps remain:

1. Client PSK offer planning still filters tickets against configured policy rather than the live provider, so an unsupported SHA-384 early-data ticket can abort startup before the supported full-handshake fallback is attempted.
2. An empty effective client suite set is encoded as a zero-length `cipher_suites` vector instead of failing locally before emitting ClientHello.
3. The new partial-provider test does not actually force fallback because the default preference order selects AES-128 before AES-256 even without the new filtering; a true no-overlap case is still absent.
4. The SHA-384 HRR transcript test compares two paths through the same production implementation rather than checking an independent expected digest.
5. The required credential/failure matrix still lacks an ECDSA-P256 record-mode loopback under SHA-384 and a wrong-length SHA-384 Finished rejection.

Current GitHub CI is green at this head. I could not independently rerun the checkout locally because this execution environment could not resolve GitHub, so this pass is based on the exact head sources and the successful CI run. As before, GitHub records this as a comment review because the authenticated account is also the PR author; the inline findings remain blocking.

src/tls/tls13_backend.zig
```diff
@@ -2184,8 +2320,15 @@ pub const Tls13Backend = struct {
...
         try w.u8_(0); // legacy_session_id: this profile does not use compatibility mode
-        try w.u16_(@intCast(2 * self.policy.cipher_suites.len)); // cipher_suites
-        for (self.policy.cipher_suites) |cipher_suite| {
+        // #564 review: offer only what the live provider can actually
+        // perform — see `effectiveCipherSuites`. A capability-unsupported
+        // suite must never reach the wire, or the server could select it
+        // and this side would fail closed after ServerHello instead of the
+        // ordinary no-mutual-suite path.
+        var suites_storage: [native_cipher_suites.len]tls_algorithms.CipherSuite = undefined;
+        const offered_suites = self.effectiveCipherSuites(&suites_storage);
```
Jaruso, line 2341: [P1] Handle an empty effective suite set before encoding or mutating handshake state. If the configured policy and live provider have no overlap, `offered_suites.len` is zero and this writes a zero-length `cipher_suites` vector. That is a malformed ClientHello (`parseClientHelloObserved` rejects it as `MissingCipherSuites`), not the required local capability/no-overlap failure.

Please preflight this before `core.start()` and before key-share generation or sink output, e.g. compute the effective list in `startImpl` and return `InvalidTransportProfile` (or another explicit local configuration error) when it is empty. Add a client-side no-overlap regression asserting no ClientHello, traffic-secret event, or retained ephemeral key is produced.

src/tls/tls13_backend_tests.zig
```diff
@@ -2728,6 +2997,197 @@ fn directHarnessWithClientKeyShareMode(
...
+
+    try std.testing.expect(harness.client_driver.isComplete());
+    try std.testing.expect(harness.server_driver.isComplete());
+    // The server's own preference list still lists AES-256-GCM/SHA-384 (the
+    // default `.record` policy offers all three native suites) — capability
+    // filtering must have quietly removed it from `effectiveCipherSuites`
+    // before negotiation, landing on one of the two suites the provider can
+    // actually perform, rather than selecting SHA-384 and only discovering
+    // the gap later at `SecretExportFailed`.
+    try std.testing.expect(harness.server_backend.negotiated_cipher_suite != .tls_aes_256_gcm_sha384);
```
Jaruso: [P1] This assertion does not prove the new fallback behavior. The default native policy orders suites as AES-128, then AES-256, then ChaCha; therefore the old unfiltered implementation would also select AES-128 first and satisfy `!= .tls_aes_256_gcm_sha384`. Construct an explicit server policy ordered `{ AES-256/SHA-384, AES-128/SHA-256 }`, withdraw SHA-384/AES-256 from the provider, and assert that AES-128 is selected. Also add the required empty-intersection/no-overlap case; otherwise this test cannot regress the post-selection failure that the original review identified.

src/tls/transcript.zig
```diff
@@ -1,109 +1,349 @@
...
+    // Live: the family is selected up front, so every call hashes
+    // immediately — the baseline this deferred path must reproduce.
+    var live = Transcript{};
+    live.selectFamily(.sha384);
+    try live.update(&client_hello_1);
+    live.rebindClientHello();
+    try live.update(&hello_retry_request);
+    try live.update(&client_hello_2);
+
+    try testing.expect(deferred.peek().eql(&live.peek()));
```
Jaruso, line 327: [P1] The SHA-384 HRR check is still self-referential. Both `deferred` and `live` run through the same `HashState`, `replace`, and `rebindClientHello` implementation, so a shared framing or SHA-384 bug can make both sides equal. #564 explicitly requires independent expected bytes for HRR rebinding under both hash families. Add a checked-in literal or OpenSSL/Python-derived SHA-384 digest for `message_hash(SHA384(ClientHello1)) || HRR || ClientHello2` (and the mutation case), then compare the production transcript directly to that oracle.

src/tls/tls13_backend_tests.zig
```diff
@@ -4254,6 +4846,30 @@ test "required client authentication completes with a valid client certificate"
     try std.testing.expectEqualStrings("mutually authenticated", opened.inner.content);
 }
 
+test "#564 required client authentication completes end to end under AES-256-GCM/SHA-384, not only the SHA-256 baseline" {
```
Jaruso, line 5132: [P1] Two explicit #564 coverage requirements are still missing. This SHA-384 client-auth test reuses `fixtureIdentity()`, which is the self-signed Ed25519 fixture, so ECDSA-P256 CertificateVerify is not exercised under `TLS_AES_256_GCM_SHA384`. The added bad-Finished test mutates a correctly sized 48-byte value but does not test the required invalid-length path. Please add an ECDSA credential loopback under SHA-384 and reject both short and long Finished bodies (for example 47 and 49 bytes), verifying failure cleanup and no application-secret installation.

src/tls/tls13_backend.zig
```diff
@@ -2184,8 +2320,24 @@ pub const Tls13Backend = struct {
...
+        // #568 review: an empty intersection must fail locally, before any
+        // byte reaches the wire — otherwise this would emit a ClientHello
+        // with a zero-length `cipher_suites` vector, which is not a legal
+        // offer RFC 8446 §4.1.2 permits a peer to answer at all. Reported as
+        // the same `IllegalParameter` the server-side no-mutual-suite path
+        // (`mapNegotiationError`) surfaces, since this is exactly that
+        // failure discovered one step earlier — a local capability gap, not
+        // a distinct capability error.
+        if (offered_suites.len == 0) return error.IllegalParameter;
+        try w.u16_(@intCast(2 * offered_suites.len)); // cipher_suites
```
Jaruso, line 2349: [P1] This check is still after the startup mutation boundary. `startImpl` has already run `planPskOffer()`, `planEarlyDataAttempt()`, and `core.start()` before entering `sendClientHello`; returning here triggers `clearFailedHandshakeState()`, aborts any offer lease, and leaves `core.handshake_lifecycle = .failed`. That makes a purely local provider/policy no-overlap behave differently from the existing server-name/size preflights and prevents retrying an otherwise untouched backend.

Please compute the effective list in `startImpl` before PSK planning and before `core.start()`, for example:

```zig
if (self.role == .client) {
    var suites: [native_cipher_suites.len]CipherSuite = undefined;
    if (self.effectiveCipherSuites(&suites).len == 0)
        return error.InvalidTransportProfile;

    const base_len = try self.clientHelloEncodedLen();
    // plan offers / early data only after all local configuration preflights
}
```

`sendClientHello` can retain a defensive assertion/check. Extend the regression to assert the lifecycle remains `.idle`, `key_pair_present` is false, early data was not planned, the offer lease was not consumed, and the sink is empty.

src/tls/tls13_backend_tests.zig
```diff
@@ -8086,6 +8987,125 @@ test "a bad client Finished after PSK selection clears the accepted session and
...
+    for (server_flight_sink.items[0..server_flight_sink.len]) |event| {
+        if (event != .handshake_bytes) continue;
+        var client_step_sink = DirectSink{};
+        defer client_step_sink.deinit();
+        try harness.client_backend.backend().receive(event.handshake_bytes.epoch, event.handshake_bytes.data, &client_step_sink);
+    }
+
+    var wrong_length_buf: [4 + 32]u8 = undefined;
+    const wrong_length_finished = try tls_core.messages.encode(.finished, &([_]u8{0xaa} ** 32), &wrong_length_buf);
+
```
Jaruso, line 9143: [P1] This covers only an undersized **client** Finished received by the server. It does not prove the required “no application-secret installation” ordering: the server derives/emits application secrets when it sends its own Finished, before it waits for the client's Finished, so those secrets already existed before this malformed message arrived. It also does not cover the oversized boundary requested in the prior review.

Please test the client receiving malformed **server** Finished bodies of both 47 and 49 bytes under SHA-384. Drive the client through ServerHello/EncryptedExtensions/Certificate/CertificateVerify, replace the server Finished with each malformed encoding, then assert `MalformedHandshake`, no application-secret event was emitted, the schedule/RMS were wiped, and the lifecycle is failed. A small table-driven loop over `{ 47, 49 }` will exercise both sides of the exact-length guard.

Jaruso (commented):
Re-reviewed the new `a08ac59` head against #564 and every outstanding thread. The new commit correctly fixes provider filtering for cached PSKs, makes the fallback test non-vacuous, adds a real server no-overlap case, adds independent SHA-384 HRR transcript reconstruction, and exercises ECDSA-P256 under SHA-384. CI is green at this head.

Two blocking gaps remain:

1. The empty effective-suite guard is inside `sendClientHello`, after `startImpl` has already planned PSKs/0-RTT and called `core.start()`. The error path therefore aborts and marks the backend failed instead of behaving like the other local startup preflights; it can also consume/abort offer-planning state before discovering that no handshake is possible. Move this check before all lifecycle/offer/key state mutation.
2. The new wrong-length Finished regression sends a short **client** Finished to the server. At that point the server has already derived/emitted application secrets, so it cannot prove the required failure ordering. It also leaves the oversized case uncovered. Exercise malformed server Finished lengths on the client, where successful verification is the gate before application-secret installation.

I left the corresponding original threads unresolved and resolved the findings that are fully addressed. This remains a comment review because the authenticated account is also the PR author.

Jaruso (commented):
Re-reviewed the complete PR at `ee26804885aac752c5be7711f76d9ee0748e041f` against issue #564 and every prior review thread. I found no remaining blocking defect.

The final follow-up correctly moves the client empty-effective-suite preflight ahead of PSK/0-RTT planning, `Core.start()`, key-share generation, and sink output, while retaining a defensive encoder check. The SHA-384 malformed-Finished regression now exercises the security-relevant direction: the client receives both undersized and oversized server Finished bodies before application-secret derivation, verifies `MalformedHandshake`, observes no emitted event for the failing call, and confirms terminal schedule cleanup.

The wider PR now covers the requested native AES-128/SHA-256, AES-256/SHA-384, and ChaCha/SHA-256 loopbacks; SHA-384 HRR, resumption/binders, 0-RTT, Ed25519 and ECDSA-P256 authentication, tampered and invalid-length Finished behavior, capability fallback/no-overlap, negotiated-parameter ordering, out-parameter secret construction, and generic record-protection reuse. All prior review threads have been resolved.

CI run 1170 is green at this head. I could not independently rerun the suite locally because this environment cannot resolve `github.com`. Submitted as `COMMENT` because the authenticated GitHub account is also the PR author; from a code-review standpoint, this is approved.

PR #569, opened 2026-08-03
quic: add negotiated TLS suite packet protection

Jaruso:
## Summary
- add QUIC packet-protection profiles for AES-128/SHA-256, AES-256/SHA-384, and ChaCha20/SHA-256, keeping Initial on the RFC 9001 fixed AES-128/SHA-256 profile
- route Handshake, 0-RTT, and 1-RTT key derivation, AEAD, header protection, and key updates from explicit negotiated TLS parameters
- expand provider capabilities, crypto profile docs/vectors, and loopback coverage for all negotiated QUIC suites

## Tests

Closes #566


src/quic/tls_adapter.zig
```diff
@@ -890,8 +1030,12 @@ pub const QuicTlsAdapter = struct {
         if (level == .zero_rtt and !self.zero_rtt_enabled) return null;
         const installed_secret = self.secret(level, direction) orelse return null;
         const secret_bytes = installed_secret.slice();
-        if (secret_bytes.len != traffic_secret_len) return null;
-        return try deriveAes128GcmKeysWithProvider(self.provider, secret_bytes[0..traffic_secret_len].*);
+        const profile = self.profileForLevel(level) orelse return null;
+        if (secret_bytes.len != profile.traffic_secret_len) return null;
```
Jaruso, line 1075: `derivePacketProtectionKeysWithProvider` returns an owning aggregate containing four `FixedSecret` values, but the production callers (`sealPacketPayload`, `openPacketPayload`, header-protection call sites, and trial key-update paths) currently discard these by-value results without `deinit()`. `FixedSecret` explicitly requires deinitialization, so each packet leaves copied traffic-secret/key/IV/HP bytes in stack storage.

At minimum, every scoped caller needs mutable ownership and guaranteed cleanup:

```zig
var keys = (try self.protectionKeys(level, direction)) orelse
    return error.KeysUnavailable;
defer keys.deinit();
```

Please audit every returned `PacketProtectionKeys` value, including `nextApplicationReadKeys` and `InitialSecrets`. A stronger design would derive into caller-owned scoped storage or retain keys in an owned slot, avoiding repeated by-value secret copies entirely. Also wipe `initial_secret`, `client_secret`, and `server_secret` derivation temporaries on all paths. This is required by #566's bounded-ownership and no-unwiped-temporary criteria.

src/quic/tls_adapter.zig
```diff
@@ -963,26 +1112,49 @@ pub const QuicTlsAdapter = struct {
...
-        return secret_bytes[0..traffic_secret_len].*;
+        const profile = self.profileForLevel(.application) orelse return null;
+        if (secret_bytes.len != profile.traffic_secret_len) return null;
+        return secret_bytes;
+    }
+
+    fn profileForLevel(self: *const QuicTlsAdapter, level: EncryptionLevel) ?PacketProtectionProfile {
+        return switch (level) {
+            .initial => PacketProtectionProfile.forInitial(),
+            .zero_rtt, .handshake, .application => self.negotiated_profile,
```
Jaruso: `zero_rtt` cannot use the same `negotiated_profile` as Handshake/1-RTT. The TLS backend emits the early traffic secret while constructing ClientHello, before ServerHello and before `emitHandshakeSecrets()` emits `negotiated_parameters`. Consequently, `protectionKeys(.zero_rtt, ...)` is unavailable at the moment the client must send early packets. Once ServerHello arrives, this also risks selecting the current handshake's suite rather than explicitly preserving the resumed ticket's suite.

Please store an independent early-data profile derived from the selected ticket/session metadata and install it before the zero-RTT secret, for example:

```zig
.zero_rtt => self.zero_rtt_profile,
.handshake, .application => self.negotiated_profile,
```

The event contract should guarantee `zero_rtt_profile` arrives before the `.zero_rtt` secret. Add a test for each supported suite that seals a real 0-RTT packet before any ServerHello/`negotiated_parameters` event and verifies the server opens it using the resumed session's suite.

src/crypto/pure_zig.zig
```diff
@@ -319,21 +324,50 @@ fn quicHeaderProtectionMaskImpl(
...
             @memcpy(&s, sample);
             defer crypto.secureZero(u8, &k);
+            defer crypto.secureZero(u8, &s);
 
-            const aes = Aes128.initEnc(k);
             var block: [provider.quic_header_protection_sample_len]u8 = undefined;
-            aes.encrypt(&block, &s);
+            switch (hp) {
+                .aes_128 => {
+                    const aes = Aes128.initEnc(k[0..16].*);
```
Jaruso: The raw `k` buffer is wiped, but `Aes128.initEnc` creates an expanded encryption key schedule that also contains secret-derived round keys. That temporary is currently dropped without scrubbing, contrary to #566's requirement that key and temporary state be wiped. The AES-256 branch has the same issue.

Please make each context mutable and zero it on exit, e.g.:

```zig
var aes = Aes128.initEnc(k[0..16].*);
defer crypto.secureZero(u8, std.mem.asBytes(&aes));
aes.encrypt(&block, &s);
```

Apply the equivalent cleanup to `Aes256`, preferably through a small helper so both branches cannot diverge.

Jaruso (commented):
**Technical verdict: changes required.** GitHub does not allow the author to formally request changes on their own PR, so this is submitted as a comment review.

The negotiated Handshake/1-RTT implementation is generally well-structured: the suite profile is explicit, Initial remains fixed, provider breadth is added cleanly, real connection loopbacks cover all three suites, and all completed CI jobs are green. However, three security/lifecycle requirements from #566 are not met yet:

1. **0-RTT is bound to the wrong profile lifecycle.** The native TLS backend exports the client early traffic secret while building ClientHello, before ServerHello and before `emitHandshakeSecrets` emits `negotiated_parameters`. The adapter therefore has no `negotiated_profile` when an actual early packet must be protected. After negotiation it would also use the new connection's selected suite rather than explicitly retaining the resumed ticket's suite. #566 specifically requires 0-RTT to use the resumed session's suite consistently.
2. **The new bounded key owners are discarded without zeroization.** `PacketProtectionKeys` contains four `FixedSecret` values, whose contract requires `deinit`, but `protectionKeys()` returns that owner by value and production packet paths drop the result without calling `deinit`. This leaves copied traffic secrets, packet keys, IVs, and HP keys in stack storage after every packet operation and violates the explicit “no by-value secret copies through unwiped temporaries” criterion. Initial derivation/result ownership needs the same audit.
3. **AES header-protection expanded key schedules are not wiped.** The raw key buffer is scrubbed, but the `Aes128`/`Aes256` encryption context containing expanded round keys is left in temporary storage.

Please add focused lifecycle tests proving cleanup/replacement behavior and a real all-suite 0-RTT test that seals a packet before ServerHello/negotiated-parameter delivery, then decrypts it using the resumed session's suite. The existing all-suite loopback begins after establishment, so it cannot detect the 0-RTT ordering bug.

src/quic/tls_adapter.zig
```diff
@@ -949,10 +1134,15 @@ pub const QuicTlsAdapter = struct {
...
-        const next_write = try deriveNextGenerationSecretWithProvider(self.provider, write_secret);
-        // Secrets are exactly traffic_secret_len, so Secret.init cannot overflow.
-        self.installSecret(Secret.init(.application, .write, &next_write) catch unreachable);
+        var next_write: [max_traffic_secret_len]u8 = undefined;
+        defer crypto_provider.secureZero(&next_write);
+        deriveNextGenerationSecretForProfileWithProvider(self.provider, profile, write_secret, next_write[0..profile.traffic_secret_len]) catch |err| switch (err) {
+            error.InvalidTrafficSecretLength => return error.ApplicationSecretsMissing,
+            error.ProviderUnsupported => return error.ProviderUnsupported,
+        };
+        self.installSecret(Secret.init(.application, .write, next_write[0..profile.traffic_secret_len]) catch unreachable);
```
Jaruso, line 1156: This replaces the application traffic secret, after which `protectionKeys(.application, ...)` derives **all** material again—including `quic hp`—from the new secret. That violates RFC 9001 §§5.4/6.1: the header-protection key remains fixed for the duration of the connection; only the packet key and IV rotate. With the current code, the first new-phase packet is header-protected with a new HP key while its peer must remove protection with the original HP key, so it cannot reliably recover the Key Phase bit or packet number.

Please retain application HP material separately from the rotating traffic secret, for example:

```zig
const ApplicationProtectionState = struct {
    read_hp: HeaderProtectionSecret = .{},
    write_hp: HeaderProtectionSecret = .{},
    // rotating read/write traffic secrets remain in SecretStore
};
```

Install each HP key once from the initial application secret, reuse it from `protectionKeys` and `nextApplicationReadKeys`, and wipe it on application discard/teardown. Add an all-suite short-header test that applies header protection before the update, sends/deprotects a new-phase packet after the update, and asserts `before.hp == after.hp` while `before.key != after.key` and `before.iv != after.iv`.

src/quic/tls_adapter.zig
```diff
@@ -436,37 +540,65 @@ pub fn deriveInitialSecretsV1WithProvider(provider: crypto_provider.CryptoProvid
...
+    defer crypto_provider.secureZero(&initial_secret);
     provider.hkdfExtract(.sha256, &initial_salt_v1, client_initial_dcid, &initial_secret) catch return error.ProviderUnsupported;
     var client_secret: [traffic_secret_len]u8 = undefined;
+    defer crypto_provider.secureZero(&client_secret);
     var server_secret: [traffic_secret_len]u8 = undefined;
+    defer crypto_provider.secureZero(&server_secret);
     provider.hkdfExpandLabel(.sha256, &initial_secret, "client in", "", &client_secret) catch return error.ProviderUnsupported;
     provider.hkdfExpandLabel(.sha256, &initial_secret, "server in", "", &server_secret) catch return error.ProviderUnsupported;
-    return .{
+    return InitialSecrets{
```
Jaruso: There is still a secret-cleanup hole on partial construction. If client key derivation succeeds and server key derivation returns an error, the completed client `PacketProtectionKeys` temporary is abandoned without `deinit()`. Also, `deriveAes128GcmKeysWithProvider(provider, client_secret)` takes the fixed-size traffic secret by value, creating exactly the semantic secret copy the issue says to avoid.

Please construct into owned locals/out storage with cleanup armed before the second derivation, and pass borrowed secret bytes:

```zig
var out = InitialSecrets{
    .initial_secret = initial_secret,
    .client = .{},
    .server = .{},
};
errdefer out.deinit();
out.client = try derivePacketProtectionKeysWithProvider(
    provider,
    PacketProtectionProfile.forInitial(),
    &client_secret,
);
out.server = try derivePacketProtectionKeysWithProvider(
    provider,
    PacketProtectionProfile.forInitial(),
    &server_secret,
);
return out;
```

An out-parameter for `InitialSecrets` would be even stronger if you want to eliminate return-value copies as well. Please add a failing-provider fixture that succeeds through client derivation, fails during server derivation, and observes cleanup.

src/quic/tls_adapter.zig
```diff
@@ -1210,22 +1447,138 @@ test "adapter derives Handshake and 1-RTT protection keys from installed traffic
...
+        var client_keys = (try client.protectionKeys(.zero_rtt, .write)).?;
+        defer client_keys.deinit();
+        try testing.expectEqual(profile.aead, client_keys.profile.aead);
+        try testing.expectEqual(profile.header_protection, client_keys.profile.header_protection);
+        try testing.expectEqual(@as(?PacketProtectionKeys, null), client.protectionKeys(.handshake, .write));
+
+        const header = "\xd0\x00\x00\x00\x01\x00";
+        const plaintext = "real zero-rtt payload before server hello";
+        var sealed: [128]u8 = undefined;
+        const protected = try client.sealPacketPayload(.zero_rtt, .write, 1, header, plaintext, &sealed);
```
Jaruso, line 1702: This verifies the resumed profile selects the right AEAD for payload seal/open, but it does not build a QUIC packet, apply/remove header protection, or exercise the TLS `early_data_parameters -> zero_rtt traffic_secret` event ordering. The existing real wire/resumption tests call `sealTestZeroRttPacket` only with AES-128, so AES-256 and ChaCha20 0-RTT can regress independently while this stays green.

Please drive each suite through the real resumed-session event path and ingest a fully protected 0-RTT long-header packet **before ServerHello**. At minimum the test should prove, for all three suites:

```zig
// client start emits early_data_parameters, then zero_rtt write secret
// server ClientHello processing emits matching early_data_parameters, then read secret
// packet uses AEAD + suite-specific HP and is accepted before negotiated_parameters
```

That is the coverage requested by #566 and by the prior review thread.

Jaruso (commented):
Second-pass verdict: **changes still required**.

The follow-up commit correctly addresses the three originally reported mechanisms:

- 0-RTT now has independent resumed-session metadata emitted before the early traffic secret;
- production/test `PacketProtectionKeys` owners are generally scoped and deinitialized, and `InitialSecrets` has explicit cleanup;
- AES-128/AES-256 header-protection schedules are wiped;
- the full CI run for `eff916f0` is green, including appliance tests, H3 resumption/0-RTT interop, dependency audit, and all build/test matrices.

Three issues remain against #566:

1. **Key updates rotate the header-protection key.** RFC 9001 requires one header-protection key for the connection; only the packet key and IV change. Replacing the application traffic secret and deriving the complete profile again changes `quic hp`, so a real updated short-header packet cannot be unmasked by its peer. The current test checks only that the AEAD key changes and therefore misses this.
2. **Initial derivation still has an unwiped partial-success/error path and by-value secret copy.** If client key derivation succeeds and server derivation fails, the client aggregate has no `errdefer deinit`; the fixed-size traffic secret is also passed by value into the wrapper.
3. **The new all-suite 0-RTT test does not exercise the wire/event path.** It directly installs metadata and seals only the payload. The real wire/resumption tests still use AES-128, so AES-256 and ChaCha20 early-data metadata ordering plus header protection are not covered end to end before ServerHello.

Please add a real all-suite key-update packet round trip and a real all-suite pre-ServerHello 0-RTT wire round trip; those should prevent both protocol regressions from returning.

Jaruso (commented):
**Technical verdict: approved.** GitHub does not permit the author to formally approve their own pull request, so this is submitted as a comment review.

I re-reviewed the incremental commit against #566 and the prior findings. The requested changes are now satisfied:

- Application key updates retain independently owned read/write header-protection keys while rotating the traffic secret, packet key, and IV. The retained HP material is wiped on application discard and adapter teardown, and the all-suite short-header test proves a new-phase packet can be deprotected using the original HP key.
- Initial-secret construction now arms aggregate cleanup before the second derivation, uses borrowed traffic-secret slices, and includes a failing-provider path covering partial construction.
- The 0-RTT integration test now runs a genuine ticket issuance/resumption flow for AES-128, AES-256, and ChaCha20, verifies early suite metadata and secrets are installed before ServerHello, ingests a fully AEAD- and header-protected 0-RTT packet through the real connection receive path, and completes the resumed handshake afterward.

I also rechecked the incremental production changes for key ownership, discard/teardown cleanup, suite separation, packet-key derivation, and key-phase processing. I found no remaining blocking issue. All completed CI jobs are green, including unit tests, appliance tests, H3 resumption/0-RTT interop, security, formatting, and multi-platform builds; TLS dependency audit and packaging smoke were still running at review time.

PR #570, opened 2026-08-03
tls: add RSA-PSS server signing and credential integration (323-G4)

Jaruso:
## Summary

- Adds a pure-Zig RSA private-key owner and `.rsa_pss_rsae_sha256` signing (`src/crypto/rsa.zig`: `PrivateKey` + `signPssSha256`) behind the existing `provider.SigningKey` seam (`pure_zig.SoftwareRsaSigningKey`), mirroring the existing Ed25519/ECDSA-P256 signing-key shape (constructors, `deinit` wipe, `format` ban, entropy-before-output-size ordering).
- The private-exponent modular exponentiation runs through `std.crypto.ff.Modulus.powWithEncodedExponent` — the stdlib's constant-time-oriented (non-public-exponent) path. No CRT: `p`/`q`/`dp`/`dq`/`qInv` are used only for import-time validation, never the signing operation itself, a deliberate scope narrowing documented in `docs/CRYPTO_SECURITY_AUDIT.md`.
- **Full import-time validation of the RSA private key**, not just structural/range checks: `parsePrivateKeyDer` requires `n = p·q`, `p ≠ q`, `dP = d mod (p−1)`, `dQ = d mod (q−1)`, `qInv·q ≡ 1 (mod p)`, and `e·d ≡ 1 (mod lcm(p−1, q−1))` (via `validateComponentRelationships`, using `std.math.big.int.Managed` over a bounded, wiped `FixedBufferAllocator` scratch buffer), **and** that `p`/`q` are actually prime (`isProbablePrime`: Miller-Rabin, 64 rounds, witnesses drawn uniformly at random from an injected `provider.Entropy` — not a fixed set, which is unsound against an adversarially chosen key). Regression-tested against real Carmichael numbers, the smallest Fermat pseudoprime to base 2, and a full offline-constructed 2048-bit key whose `p`/`q` are themselves composite yet satisfy every algebraic relationship exactly.
- `credentials.Identity.initPkcs8`'s signature is unchanged (44 pre-existing call sites, none needing modification) and now fails closed for an RSA key rather than silently skipping the primality check; a new `Identity.initPkcs8WithEntropy` is the only path that can construct an RSA identity, and also rejects a certificate/private-key mismatch before advertising a credential.
- Hardened PKCS#8 wrapper parsing (mandatory NULL AlgorithmIdentifier params, full DER exhaustion at every level).
- Wires the scheme end to end: `credentials.SignatureScheme`/`Identity.Key`, `policy.IdentityKey`, `sni_provider.KeyKind`/scheme mappings, and `tls13_backend.zig`'s `native_signature_schemes`/`checkProofOfPossession`/`leafSupportsSignatureScheme` — so the native TLS 1.3 engine can select an RSA credential and complete a CertificateVerify with it, including RSA-specific negative coverage (tampered CertificateVerify → `DecryptError`, no-overlap failing before the signer runs, scheme/leaf-key mismatches in both directions).
- `crypto/profile.zig`'s capability matrix flips RSA-PSS's `tls_handshake` integration to `.live`, but its `enabled_product_profiles` deliberately still excludes the native appliance pending a separate #391 policy decision — engine *capability* and product *enablement* are tracked independently, per the issue's explicit instruction not to silently broaden the appliance profile.
- Extends the OpenSSL EVP oracle (`tests/evp_oracle.c`) with `rsa-pss-verify` and adds a differential test cross-verifying native RSA-PSS signatures against it.

## Test plan

- [x] `zig build test-crypto` — RSA private-key parsing/validation (including full component-relationship and primality checks), PSS sign/verify roundtrip, entropy-failure and undersized-output non-partial-signature checks, `deinit` wipe, crypto-boundary audit.
- [x] `zig build test-tls` — credential selection (key-type/offer compatibility, cert/key match), policy identity constraints, RSA-specific negative loopbacks, and a full native record-mode client/server loopback with an RSA-PSS-only policy on both sides, directly asserting the negotiated scheme.
- [x] `zig build test-crypto-openssl` — native RSA-PSS signatures independently verified by OpenSSL via the extended EVP oracle, plus tamper/wrong-key/malformed-key rejection on both sides.
- [x] `zig build test` (full suite) — green.
- [x] `zig build test-integration` — pre-existing `ConnectionResetByPeer` failure reproduces identically on unmodified `main` in this sandbox (unrelated to this change).

Closes #565


Jaruso (commented):
Blocking review findings: the end-to-end RSA-PSS happy path is well integrated, the OpenSSL differential oracle is useful, and the current CI run is green. However, the implementation does not yet satisfy several explicit #565 acceptance requirements at the credential-import boundary:

1. The PKCS#1 loader accepts mathematically inconsistent RSA private keys; it only performs structural/range checks even though #565 requires validating the modulus, exponents, CRT components, and their relationships.
2. An RSA certificate and an unrelated RSA private key are treated as a compatible identity and advertised; the mismatch is discovered only when the peer rejects CertificateVerify.
3. The new PKCS#8 unwrap accepts malformed wrappers because required AlgorithmIdentifier fields and DER exhaustion are not checked.
4. The required RSA-specific negative TLS coverage is missing; the added loopback proves only the successful path.

Please address the inline comments and add focused mutations for every RSA component plus RSA-specific TLS failure tests. The controlled-appliance profile remains correctly unchanged, and the existing Ed25519/ECDSA paths appear preserved.

Note: this is submitted as COMMENT only because GitHub does not allow the PR author to request changes on their own pull request; the findings are blocking.

src/crypto/rsa.zig
```diff
@@ -160,6 +190,179 @@ pub fn verifyPssSha256(public_key_der: []const u8, message: []const u8, signatur
...
+    if (exponent.len == 0 or !lessThanUnsigned(exponent, modulus)) return error.InvalidInput;
+    if (exponent.len == 1 and exponent[0] < 3) return error.InvalidInput;
+    if (exponent[exponent.len - 1] & 1 == 0) return error.InvalidInput;
+
+    const private_exponent = stripLeadingZero(private_exponent_encoded);
+    if (private_exponent.len == 0) return error.InvalidInput;
+    if (private_exponent.len == 1 and private_exponent[0] == 0) return error.InvalidInput;
+    if (!lessThanUnsigned(private_exponent, modulus)) return error.InvalidInput;
+
+    inline for (.{ prime1_encoded, prime2_encoded, exponent1_encoded, exponent2_encoded, coefficient_encoded }) |encoded| {
```
Jaruso: [P1] Validate the RSA component relationships before constructing the signing key. This loop currently proves only that `p`, `q`, `dp`, `dq`, and `qInv` are nonzero and numerically smaller than `n`; values can be changed arbitrarily while remaining in range and `parsePrivateKeyDer` will still accept the key. That directly misses #565's requirement to validate the modulus, private exponent/CRT components, and component relationships, and means values such as `d = 1` or `n != p*q` become a live `provider.SigningKey` and fail only later during peer verification.

At import time, please enforce at least:

```zig
try expectEqualBig(n, try mulBig(p, q));
if (equalBig(p, q)) return error.InvalidInput;
try expectCongruentOne(e, d, try lcmBig(p_minus_one, q_minus_one));
try expectEqualBig(dp, try modBig(d, p_minus_one));
try expectEqualBig(dq, try modBig(d, q_minus_one));
try expectCongruentOne(q_inv, q, p); // qInv * q mod p == 1
```

The exact helper API can use the project's bounded fixed-field/big-int primitives, but all checks must happen before any private bytes are copied into the returned owner. Add mutation tests for each field where the DER remains structurally valid and in range. If the project intentionally wants only structural CRT validation, #565's acceptance criteria must be explicitly amended rather than narrowed in this PR's audit prose.

src/tls/credentials.zig
```diff
@@ -531,19 +533,25 @@ pub const Identity = struct {
...
+            var scalar = scalar_local;
+            defer crypto.secureZero(u8, &scalar);
+            var scalar_secret = crypto_pkg.secrets.FixedSecret(32).init(&scalar) catch return error.InvalidPrivateKey;
+            defer scalar_secret.deinit();
+            const software_key = pure_zig.SoftwareEcdsaP256SigningKey.fromScalarSecret(&scalar_secret) catch return error.InvalidPrivateKey;
+            return .{ .certificate_der = certificate_der, .key = .{ .ecdsa_p256 = software_key } };
+        } else |_| {}
+        const rsa_key_der = try rsaKeyDerFromPkcs8(pkcs8_key_der);
+        const software_key = pure_zig.SoftwareRsaSigningKey.fromDer(rsa_key_der) catch return error.InvalidPrivateKey;
+        return .{ .certificate_der = certificate_der, .key = .{ .rsa = software_key } };
```
Jaruso, line 603: [P1] Reject an RSA certificate/private-key mismatch here instead of advertising it as a usable credential. `initPkcs8` never parses `certificate_der`, and the later selection checks only that both sides are RSA-shaped. Therefore certificate A plus unrelated RSA key B is selected and signs CertificateVerify; the first actual compatibility check is the peer rejecting the signature, contrary to #565's requirement that a credential be advertised only for a compatible signer + leaf and that bad pairings fail at snapshot build/selection rather than sign time.

Please retain or otherwise expose the parsed public exponent alongside `n`, then compare the leaf's `RSAPublicKey` to the private key before returning:

```zig
var software_key = pure_zig.SoftwareRsaSigningKey.fromDer(rsa_key_der) catch
    return error.InvalidPrivateKey;
errdefer software_key.deinit();

const leaf = (crypto.Certificate{ .buffer = certificate_der, .index = 0 }).parse() catch
    return error.InvalidPrivateKey;
if (leaf.pub_key_algo != .rsaEncryption or
    !software_key.matchesPublicKeyDer(leaf.pubKey()))
{
    return error.InvalidPrivateKey;
}
```

`matchesPublicKeyDer` should strictly parse and compare both modulus and public exponent. Add a test pairing the fixture certificate with a different valid RSA private key and assert construction/snapshot build rejects it before entropy or signing is invoked.

src/tls/credentials.zig
```diff
@@ -617,6 +626,21 @@ pub const Identity = struct {
...
+    /// return the borrowed `RSAPrivateKey` DER `pure_zig.SoftwareRsaSigningKey
+    /// .fromDer` parses directly.
+    fn rsaKeyDerFromPkcs8(der: []const u8) InitError![]const u8 {
+        const oid_rsa_encryption = [_]u8{ 0x06, 0x09, 0x2a, 0x86, 0x48, 0x86, 0xf7, 0x0d, 0x01, 0x01, 0x01 };
+        var walker = DerWalker{ .bytes = der };
+        var outer = try walker.sequence();
+        try outer.expectInteger(0);
+        var alg = try outer.sequence();
+        try alg.expectBytes(&oid_rsa_encryption);
+        return outer.octetString();
```
Jaruso: [P2] Fully validate the PKCS#8 wrapper instead of returning after the first OCTET STRING. The parser currently does not consume the required `NULL` parameters after the `rsaEncryption` OID, does not require the AlgorithmIdentifier walker to be exhausted, and does not require either the outer SEQUENCE or top-level input to be exhausted. As a result, missing/garbage algorithm parameters and `valid_pkcs8 ++ trailing_bytes` are accepted as valid RSA keys, which violates the malformed-key rejection requirement.

For the exact wrapper documented above, this should look like:

```zig
try alg.expectBytes(&oid_rsa_encryption);
try alg.expectBytes(&[_]u8{ 0x05, 0x00 });
if (alg.pos != alg.bytes.len) return error.InvalidPrivateKey;

const key_der = try outer.octetString();
if (outer.pos != outer.bytes.len) return error.InvalidPrivateKey;
if (walker.pos != walker.bytes.len) return error.InvalidPrivateKey;
return key_der;
```

If RFC 5958 attributes are intentionally supported, parse those explicitly rather than silently ignoring arbitrary bytes. Please add tests for absent/wrong parameters, extra AlgorithmIdentifier fields, extra outer fields, and trailing bytes after the top-level SEQUENCE.

src/tls/tls13_backend_tests.zig
```diff
@@ -588,6 +588,55 @@ test "direct shared driver preserves derivation, sequence, discard, and teardown
     try std.testing.expect(std.mem.allEqual(u8, std.mem.asBytes(&harness.server_backend.identity), 0));
 }
 
+// #565: an RSA-2048 server credential, negotiated exclusively (the shared
+// policy on both sides names only `.rsa_pss_rsae_sha256`, so a successful
+// handshake is only reachable through the native RSA-PSS signing, provider
+// verification, and credential-selection path added by this change — there
+// is no other scheme to fall back to).
+test "native RSA-PSS server credential completes a record-mode handshake" {
```
Jaruso, line 596: [P2] Add the RSA-specific negative loopbacks required by #565. This test proves only the successful record-mode path. The issue also requires: a modified RSA CertificateVerify mapping to `DecryptError`, malformed/wrong key mapping to the existing certificate taxonomy, and no-overlap failing before signing. Those properties are especially important here because RSA introduced new algorithm/key-type branches and a variable-size signature.

Please add focused cases using the RSA-only policy, for example:

```zig
test "RSA-PSS tampered CertificateVerify maps to DecryptError" { ... }
test "RSA-PSS no-overlap fails before signer invocation" { ... } // counting signer/entropy, calls == 0
test "RSA-PSS malformed or mismatched key is rejected before flight emission" { ... }
```

Also assert the negotiated CertificateVerify scheme is `rsa_pss_rsae_sha256` directly rather than only inferring it from the policy. Provider-level tests should cover successful 3072/4096-bit keys and deterministic rejection of an unsupported size, since the implementation advertises all three sizes but currently exercises only RSA-2048.

src/crypto/rsa.zig
```diff
@@ -160,6 +190,303 @@ pub fn verifyPssSha256(public_key_der: []const u8, message: []const u8, signatur
...
+    e: []const u8,
+    d: []const u8,
+    p: []const u8,
+    q: []const u8,
+    dp: []const u8,
+    dq: []const u8,
+    qinv: []const u8,
+) Error!void {
+    const big = std.math.big.int;
+    var scratch: [131072]u8 = undefined;
```
Jaruso, line 294: [P1] Zeroize this scratch buffer on **every** exit. `bigFromBytes` copies `d`, `p`, `q`, `dP`, `dQ`, and `qInv` into allocations backed by these 128 KiB, but the function currently returns with those private values still resident in stack memory. Going out of scope—or the audit text's claim that a later stack frame will overwrite it—is not a wipe and does not satisfy #565's explicit scratch-state/failure-path requirement.

Put the wipe immediately after declaration, before any fallible operation:

```zig
var scratch: [relationship_scratch_len]u8 = undefined;
defer secrets.secureZero(&scratch);
var fba = std.heap.FixedBufferAllocator.init(&scratch);
```

Please update the audit row to state that this buffer is explicitly wiped on both success and every validation failure. A small internal helper that accepts caller-provided scratch would also let a test assert the buffer is zero after a successful validation and after a deliberately corrupted component returns `InvalidInput`.

src/crypto/rsa.zig
```diff
@@ -160,6 +190,303 @@ pub fn verifyPssSha256(public_key_der: []const u8, message: []const u8, signatur
...
+    const big_n = bigFromBytes(allocator, n) catch return error.InvalidInput;
+    const big_e = bigFromBytes(allocator, e) catch return error.InvalidInput;
+    const big_d = bigFromBytes(allocator, d) catch return error.InvalidInput;
+    const big_p = bigFromBytes(allocator, p) catch return error.InvalidInput;
+    const big_q = bigFromBytes(allocator, q) catch return error.InvalidInput;
+    const big_dp = bigFromBytes(allocator, dp) catch return error.InvalidInput;
+    const big_dq = bigFromBytes(allocator, dq) catch return error.InvalidInput;
+    const big_qinv = bigFromBytes(allocator, qinv) catch return error.InvalidInput;
+
+    if (big_p.eql(big_q)) return error.InvalidInput;
```
Jaruso, line 318: [P1] These relationship checks still do not validate that `p` and `q` are actually primes—the defining property of the PKCS#1 version-0 **two-prime** key this parser claims to accept. A caller can choose distinct composite (including even) factors, set `n = p*q`, compute `d` as an inverse modulo this code's `lcm(p-1, q-1)`, and derive matching `dP`/`dQ`/`qInv`; every check below can pass even though the structure is not a valid two-prime RSA key. Depending on the factors, the later field modulus construction can reject it or the resulting private operation can fail for some representatives, moving malformed-key detection back to sign time.

Validate both factors before trusting the equations, for example:

```zig
if (!isOdd(&big_p) or !isOdd(&big_q)) return error.InvalidInput;
if (!isProbablePrime(&big_p) or !isProbablePrime(&big_q))
    return error.InvalidInput;
```

The implementation can use bounded small-prime trial division plus a reviewed fixed-round Miller–Rabin/BPSW-style probable-prime check; it does not need ambient entropy if bases are selected deterministically. Add regression fixtures whose composite `p` or `q` have **all other fields recomputed consistently**, so rejection proves primality validation rather than merely another equation catching a stale field. Also include an even-factor case to ensure an invalid modulus cannot become a live signer.

Jaruso (commented):
Follow-up review on `1c29261c`.

The four findings from the first review are addressed, and I resolved those threads: the key now checks the requested RSA component equations, certificate/private-key mismatch is rejected during identity construction, the RSA PKCS#8 wrapper is exhausted strictly, and the RSA-specific negative/3072/4096/unsupported-size coverage is present. CI is green.

This remains **blocking** because the newly added relationship validator still does not satisfy the private-key import contract in two ways:

1. Its 128 KiB big-integer scratch buffer contains copies of every private component and is never zeroized on success or failure.
2. It checks equations involving `p` and `q` but never verifies that they are primes (or even odd), so a composite-factor structure can satisfy every current equation and still be accepted as a PKCS#1 two-prime RSA key. That can produce a weak/non-RSA modulus or defer failure until signing, which is exactly what import-time validation is intended to prevent.

I left detailed inline fixes and regression-test expectations. As before, GitHub does not allow the PR author to submit `REQUEST_CHANGES` on their own PR, so this COMMENT review should be treated as changes requested.

src/crypto/rsa.zig
```diff
@@ -160,6 +190,416 @@ pub fn verifyPssSha256(public_key_der: []const u8, message: []const u8, signatur
...
+    for (buf) |*byte| {
+        const next_carry: u8 = byte.* & 1;
+        byte.* = (byte.* >> 1) | (carry << 7);
+        carry = next_carry;
+    }
+}
+
+/// Small-prime witness bases for `isProbablePrime`'s Miller-Rabin rounds (the
+/// first 40 odd primes).
+const miller_rabin_witnesses = [_]u8{
```
Jaruso: [P1] Do not use a public fixed list of 40 small bases as the security boundary for attacker-controlled RSA key import. This is concretely bypassable, not merely a theoretical caveat: François Arnault's 1995 construction (`Journal of Symbolic Computation` 20(2), DOI `10.1006/jsco.1995.1042`) includes a 397-digit composite that is a strong Miller–Rabin pseudoprime to **every base below 307**. It therefore passes every witness in this array (`2..173`) and is only about 1,319 bits, well within the component widths this parser accepts. A caller can use such a value as `p`, choose a coprime `q`, and recompute `n`, `d`, `dP`, `dQ`, and `qInv` so every later algebraic check also succeeds. The checked-in composite fixture only proves that these particular 40 bases catch that fixture; it does not establish adversarial primality validation.

For an import boundary, use independently sampled witnesses from injected entropy (or a reviewed primality proof with an equivalent adversarial guarantee), not fixed or candidate-derived public bases. For example, thread the project's entropy seam into RSA import and perform 64 uniform independent rounds, which gives the standard `<= 4^-64 = 2^-128` false-prime bound for any composite:

```zig
fn isProbablePrime(candidate: []const u8, entropy: provider.Entropy) Error!bool {
    if (!passesSmallPrimeDivision(candidate)) return false;

    var state = try MillerRabinState.init(candidate);
    defer state.deinitAndWipe();

    for (0..64) |_| {
        // Rejection-sample uniformly from [2, candidate - 2]; modulo bias
        // would invalidate the stated bound.
        const witness = try state.sampleWitness(entropy);
        if (!try state.passesRound(witness)) return false;
    }
    return true;
}
```

That likely requires an entropy-aware `parsePrivateKeyDer` / `SoftwareRsaSigningKey.fromDer` / identity-import path, with entropy failure mapped explicitly. Add Arnault's published composite (or another documented composite passing all 40 current bases) as a regression: the old fixed-base implementation must accept it, while the replacement must reject it using deterministic test entropy that supplies a known compositeness witness.

src/crypto/rsa.zig
```diff
@@ -160,6 +190,416 @@ pub fn verifyPssSha256(public_key_der: []const u8, message: []const u8, signatur
...
+/// avoids only to keep RSA private-key parsing free of an entropy-source
+/// dependency. Recorded here per the project's practice of documenting
+/// toolchain/scope limitations honestly rather than leaving them implicit;
+/// see `docs/CRYPTO_SECURITY_AUDIT.md`.
+fn isProbablePrime(candidate: []const u8) bool {
+    if (candidate.len == 0) return false;
+    if (candidate.len == 1) return candidate[0] == 2 or candidate[0] == 3;
+    if (candidate[candidate.len - 1] & 1 == 0) return false; // even, and > 1 byte so > 3: composite
+
+    var n_minus_one_buf: [max_modulus_bytes]u8 = undefined;
```
Jaruso, line 469: [P1] The primality fix reintroduces unwiped private scratch outside the arena fixed in the previous thread. `n_minus_one_buf` receives a direct copy of `p` or `q`; `d_buf` receives another candidate-derived copy; and the subsequent `Modulus`/`Fe` values and per-round `x` contain representations and residues derived from that private factor. None are covered by `validateComponentRelationships`'s `defer secrets.secureZero(&scratch)`, because they live in this callee's separate stack frame. Every success and early-return path currently leaves them behind, contrary to #565's requirement that private scratch be wiped.

At minimum, wipe both byte arrays immediately from their declarations, and preferably group all candidate-derived field state into one explicitly wiped owner:

```zig
var n_minus_one_buf: [max_modulus_bytes]u8 = undefined;
defer secrets.secureZero(&n_minus_one_buf);
var d_buf: [max_modulus_bytes]u8 = undefined;
defer secrets.secureZero(&d_buf);

var prime_state = try PrimeTestState.init(candidate);
defer prime_state.deinitAndWipe(); // modulus, n-1 Fe, and round temporaries
```

If retaining the current inline shape, declare the `Modulus`/`Fe` values mutable and wipe `std.mem.asBytes(&value)` on scope exit, including each round's `x`; alternatively, move the whole primality operation onto caller-provided scratch owned by the already-unconditional outer wipe. Add an internal test seam that supplies observable scratch/state and asserts zeroization after both a passing prime and an early composite/fallible path. Update the audit row to cover primality-test scratch as well as the big-int relationship arena.

Jaruso (commented):
Blocking follow-up review at `d7de2426`.

The two findings from the previous pass were implemented, so I resolved those threads: the 128 KiB relationship-validation arena is now wiped unconditionally, and the parser now performs a primality gate with a full algebraically-consistent composite-factor regression fixture. CI run 1178 is green across unit tests, Linux/macOS/ARM builds, appliance tests, PKI differential coverage, security, packaging, and perf smoke.

Two P1 issues remain in the newly added primality implementation:

1. The public fixed witness set is not a sound primality check for adversarially supplied private keys. Published strong pseudoprimes already pass every witness used here, so a composite `p`/`q` can still be imported after recomputing the other PKCS#1 fields consistently.
2. `isProbablePrime` creates new stack-resident copies and field representations of `p`/`q` without wiping them, reintroducing the private-scratch lifetime problem outside the now-wiped relationship arena.

Treat this review as blocking even though GitHub does not allow the authenticated PR author to submit `REQUEST_CHANGES`. Once these two import-path issues are fixed, the previously reviewed TLS/signing/selection behavior and issue #565 coverage otherwise look complete. Please also refresh the PR summary afterward: it still describes the CRT fields as only range-validated, which no longer matches the implementation.

src/tls/credentials.zig
```diff
@@ -522,28 +523,68 @@ pub const Identity = struct {
...
+            var scalar = scalar_local;
+            defer crypto.secureZero(u8, &scalar);
+            var scalar_secret = crypto_pkg.secrets.FixedSecret(32).init(&scalar) catch return error.InvalidPrivateKey;
+            defer scalar_secret.deinit();
+            const software_key = pure_zig.SoftwareEcdsaP256SigningKey.fromScalarSecret(&scalar_secret) catch return error.InvalidPrivateKey;
+            return .{ .certificate_der = certificate_der, .key = .{ .ecdsa_p256 = software_key } };
+        } else |_| {}
+        const rsa_key_der = try rsaKeyDerFromPkcs8(pkcs8_key_der);
+        const rsa_entropy = entropy orelse return error.InvalidPrivateKey;
+        var software_key = pure_zig.SoftwareRsaSigningKey.fromDer(rsa_key_der, rsa_entropy) catch return error.InvalidPrivateKey;
```
Jaruso: [P2] Preserve the entropy failure through this entropy-aware import API instead of classifying it as an invalid private key. `SoftwareRsaSigningKey.fromDer` already distinguishes `error.EntropyFailure`, but this catch collapses it into `InvalidPrivateKey`; callers therefore cannot tell a valid RSA credential plus a failed/randomness provider from malformed key material, cannot retry appropriately, and the new import seam does not provide the explicit failure mapping requested in the previous review.

Give the entropy-taking API a wider typed error while keeping the legacy API narrow, for example:

```zig
pub const InitError = error{InvalidPrivateKey};
pub const InitWithEntropyError = InitError || error{EntropyFailure};

pub fn initPkcs8WithEntropy(...) InitWithEntropyError!Identity {
    return initPkcs8Impl(..., entropy);
}

// In the RSA branch:
var software_key = pure_zig.SoftwareRsaSigningKey.fromDer(rsa_key_der, rsa_entropy) catch |err| switch (err) {
    error.InvalidInput => return error.InvalidPrivateKey,
    error.EntropyFailure => return error.EntropyFailure,
    else => return error.InvalidPrivateKey,
};
```

`initPkcs8` can catch the wider internal error and mark `EntropyFailure` unreachable because it never supplies RSA entropy. Add a failing/partially-writing entropy test asserting `initPkcs8WithEntropy` returns `EntropyFailure`, creates no identity, and leaves no private validation scratch behind.

Jaruso (commented):
Re-reviewed the latest head against #565 and the two open v3 findings.

The fixed/public witness issue is resolved: `isProbablePrime` now performs 64 independently entropy-drawn rounds and `drawWitness` rejection-samples over `[2, n-2]`, so I resolved that thread. CI run 1179 is green.

This is still blocking on the remaining P1 scratch-state thread. The byte arrays are now wiped, but `modulus_fe`, `n_minus_one_fe`, `one_fe`, each `base_fe`, and the evolving `x` residues are separate stack values containing the private `p`/`q` modulus or values derived from it. They remain live on normal return, composite rejection, and entropy-failure exits. The audit and #565 require all private validation scratch—not only the source byte arrays—to be explicitly wiped. Please move the whole primality state into caller-owned memory covered by an unconditional wipe, or explicitly zero every field value/round temporary on scope exit and add observable success/early-failure wipe tests.

I also left one P2 inline comment: the new entropy-aware identity import currently collapses `EntropyFailure` into `InvalidPrivateKey`, so a transient entropy/provider failure is misreported as malformed credential data. Preserve that typed failure through `initPkcs8WithEntropy` and test it.

Because GitHub does not permit requesting changes on one's own PR, treat this COMMENT review as blocking. Not ready to merge yet.

src/tls/credentials.zig
```diff
@@ -522,28 +523,77 @@ pub const Identity = struct {
...
+    /// Load an Ed25519 or ECDSA-P256 PKCS#8 private key. Returns
+    /// `error.InvalidPrivateKey` for an RSA key — RSA's `p`/`q` primality
+    /// check needs an injected entropy source for its random Miller-Rabin
+    /// witnesses (a fixed/public witness set is not sound against an
+    /// adversarially chosen key; see `rsa.isProbablePrime`'s doc comment),
+    /// which this entry point has none of. Use `initPkcs8WithEntropy` for
+    /// RSA support.
     pub fn initPkcs8(certificate_der: []const u8, pkcs8_key_der: []const u8) InitError!Identity {
+        return initPkcs8Impl(certificate_der, pkcs8_key_der, null);
+    }
```
Jaruso, line 561: [P1] Thread this entropy-aware RSA import through the production file-loading composition before marking RSA TLS integration live. This legacy entry point intentionally rejects RSA when `entropy == null`, but `src/tls/identity_loader.zig` still constructs every file-backed identity with `Identity.initPkcs8(cert_chain[0], key_der)`. `NativeCredentialStore.reloadFromFiles` reaches that loader for both the default and every SNI bundle, so a valid RSA PKCS#8 file deterministically fails before snapshot construction. Even if an RSA identity were injected around the loader, `src/http/native_tls_connection.zig`'s `keyKindForScheme` currently handles only Ed25519/ECDSA and sends RSA to `unreachable`.

The lower-level tests bypass both gaps by calling `testdata.rsaIdentity()` / `initPkcs8WithEntropy` directly, so the advertised live capability is not provisionable through the native server's canonical file path.

Please thread import entropy through the loader and production store, for example:

```zig
pub fn loadIdentityWithEntropy(
    allocator: std.mem.Allocator,
    cert_path: []const u8,
    key_path: []const u8,
    entropy: provider.Entropy,
) !LoadedIdentity {
    // existing file/PEM handling...
    const identity = try credentials.Identity.initPkcs8WithEntropy(
        cert_chain[0],
        key_der,
        entropy,
    );
    return ...;
}
```

Then pass `self.entropy_source.entropy()` through `LoadedBundle.initDefault` / `initNamed` / `init`, and complete the projection:

```zig
fn keyKindForScheme(scheme: credentials.SignatureScheme) sni_provider.KeyKind {
    return switch (scheme) {
        .ed25519 => .ed25519,
        .ecdsa_secp256r1_sha256 => .ecdsa_p256,
        .rsa_pss_rsae_sha256 => .rsa,
        else => unreachable,
    };
}
```

Add a composition-level test that writes the RSA fixture certificate/key to temporary files, calls `NativeCredentialStore.reloadFromFiles`, selects `rsa_pss_rsae_sha256`, and successfully signs/verifies. That single test should catch both the missing entropy thread and the missing key-kind arm.

src/tls/credentials.zig
```diff
@@ -522,28 +523,77 @@ pub const Identity = struct {
...
-    pub const InitError = error{InvalidPrivateKey};
-
+    /// `EntropyFailure` is distinct from `InvalidPrivateKey`: it means the
+    /// key itself may well be fine but `initPkcs8WithEntropy`'s injected
+    /// entropy source failed mid-RSA-import (drawing Miller-Rabin
+    /// witnesses), a transient/local provider fault, not malformed
+    /// credential data — callers that log or alert on these should not
+    /// conflate "bad key on disk" with "entropy source is unhealthy."
+    pub const InitError = error{ InvalidPrivateKey, EntropyFailure };
+
```
Jaruso, line 544: [P2] Keep the legacy constructor's public error set narrow. Adding `EntropyFailure` to `InitError` changes the function type of `Identity.initPkcs8`, even though that no-entropy entry point can never return it; exhaustive callers and stored function pointers now have to account for a new impossible error. That also conflicts with the PR's compatibility claim that the existing constructor signature/call sites remain unchanged.

Use separate error sets as originally suggested:

```zig
pub const InitError = error{InvalidPrivateKey};
pub const InitWithEntropyError = InitError || error{EntropyFailure};

pub fn initPkcs8(...) InitError!Identity {
    return initPkcs8Impl(..., null) catch |err| switch (err) {
        error.InvalidPrivateKey => error.InvalidPrivateKey,
        error.EntropyFailure => unreachable,
    };
}

pub fn initPkcs8WithEntropy(...) InitWithEntropyError!Identity {
    return initPkcs8Impl(..., entropy);
}
```

This preserves the typed entropy failure exactly where it is possible without widening the established Ed25519/ECDSA API.

Jaruso (commented):
Re-reviewed the complete delta from `af051282` to `df321433` and rechecked #565's acceptance criteria. The two previous findings are fixed: all explicit Miller–Rabin field state is now wiped on every return path, and the entropy-aware import preserves `EntropyFailure`; I resolved both threads.

One blocking integration gap remains. The production file-loading composition cannot construct an RSA identity: `identity_loader.loadIdentity` still calls the no-entropy `Identity.initPkcs8`, which this PR deliberately makes reject RSA, and `NativeCredentialStore`'s scheme-to-key-kind projection still has no RSA arm. The lower-level fixed/SNI and loopback tests construct RSA identities directly, so they do not exercise the actual reload-from-files path. This leaves RSA-PSS marked live while the native server's canonical credential-loading path cannot provision it.

I also left a P2 API comment: broadening the legacy constructor's `InitError` changes its public function type and exposes an impossible `EntropyFailure`; the wider error set should be limited to `initPkcs8WithEntropy`.

CI run 1180 is currently red because the macOS job hit the unrelated QUIC Retry smoke assertion (`expected 4, found 3`); the crypto/TLS tests and Linux/ARM jobs passed. The production RSA loading gap is independently blocking.

src/tls/credentials.zig
Jaruso: The legacy function's return type is narrow again, but the public compatibility issue is not fully resolved: the pre-existing `Identity.InitError` alias itself has still changed from `error{InvalidPrivateKey}` to `error{InvalidPrivateKey, EntropyFailure}`. Callers that name that public type directly now inherit the new impossible branch even though `initPkcs8` was moved to `LegacyInitError`.

Please preserve the established name for the established API and give the entropy-taking path the new name:

```zig
pub const InitError = error{InvalidPrivateKey};
pub const InitWithEntropyError = InitError || error{EntropyFailure};

pub fn initPkcs8(...) InitError!Identity {
    return initPkcs8Impl(..., null) catch |err| switch (err) {
        error.InvalidPrivateKey => error.InvalidPrivateKey,
        error.EntropyFailure => unreachable,
    };
}

pub fn initPkcs8WithEntropy(...) InitWithEntropyError!Identity {
    return initPkcs8Impl(..., entropy);
}

fn initPkcs8Impl(...) InitWithEntropyError!Identity { ... }
```

That keeps both the function type and the existing public error-type alias source-compatible, while exposing the typed entropy failure only on the new API.

Jaruso (commented):
Follow-up review on `9923eafd`.

The production RSA provisioning gap is fixed: entropy is now threaded through `identity_loader.loadIdentity` and `NativeCredentialStore`, the RSA key-kind projection is present, and the new composition-level test exercises the real reload-from-files path through selection and signing. I resolved that P1 thread. CI run 1181 is green.

I found no new functional or security regressions in the latest changes. One P2 compatibility detail remains in the existing thread: although `initPkcs8` now returns a narrow `LegacyInitError`, the pre-existing public `Identity.InitError` alias is still widened to include `EntropyFailure`. Callers that reference that public alias directly remain source-incompatible. Restore `InitError = error{InvalidPrivateKey}` and introduce a separate `InitWithEntropyError` for `initPkcs8WithEntropy` / the shared implementation.

This is submitted as COMMENT because GitHub does not allow the PR author to request changes on their own PR; the remaining thread should be addressed before merge.

Jaruso (commented):
Follow-up review on `fb458e711`.

The remaining compatibility issue is fixed correctly: `Identity.InitError` is restored to the pre-existing `error{InvalidPrivateKey}` API, while the new entropy-aware constructor and shared implementation use the distinct wider `InitWithEntropyError`. The final thread is resolved.

I rechecked the accumulated #565 acceptance criteria across the previous review rounds: RSA component and primality validation, random Miller–Rabin witnesses, private scratch zeroization, certificate/key binding, strict PKCS#8 parsing, typed entropy/output failures, RSA credential selection, production reload-from-files integration, native TLS positive and negative paths, 2048/3072/4096 coverage, OpenSSL differential verification, profile/documentation boundaries, and preservation of the Ed25519/ECDSA paths. I found no remaining issues.

CI run 1182 is green. This is ready to merge.

Submitted as COMMENT because GitHub does not allow the PR author to approve their own pull request.

PR #571, opened 2026-08-03
tls: close native TLS algorithm matrix

Jaruso:
Addresses #335.

## Summary
- Wire dynamic TLS 1.3 key-share negotiation for X25519 and P-256, including HRR regeneration, provider capability filtering, and private-buffer wiping when replacing shares.
- Expand CertificateVerify capacity for RSA-PSS up to RSA-4096 and add RSA-3072/RSA-4096 handshake fixtures and coverage.
- Add final matrix coverage: record-mode 3 suites x 2 groups x 3 signature schemes, QUIC 3 suites x 2 groups, malformed P-256 shares, provider fallback, and no-overlap group failure.
- Update crypto provider/profile and QUIC TLS docs for the final native TLS matrix state.

## Validation
- `zig build test --summary all`
  - 52/52 steps succeeded
  - 3102/3103 tests passed (1 skipped)


src/tls/tls13_backend.zig
```diff
@@ -1603,6 +1612,58 @@ pub const Tls13Backend = struct {
...
+    fn generateEphemeral(self: *Tls13Backend, group: tls_algorithms.NamedGroup) HandshakeError!EphemeralKeyShare {
+        const provider_group = try providerGroupOrIllegal(group);
+        const public_len = provider_group.publicKeyLength();
+        const private_len = try groupPrivateLen(group);
+        var share = EphemeralKeyShare{
+            .group = group,
+            .public_key_len = public_len,
+            .private_key_len = private_len,
+        };
+        self.crypto_provider.generateKeyShare(provider_group, share.public_key[0..public_len], share.private_key[0..private_len]) catch
```
Jaruso: **[P1] Wipe the helper-owned private scalar on provider failure.** The caller’s `defer` is attached to its still-unassigned local `share`; when this helper returns an error, that defer wipes the caller’s unrelated/undefined stack slot, not this function’s `share` that a conforming provider may already have partially filled. That violates #335’s explicit wipe-on-failure requirement.

Please arm cleanup inside this helper, for example:

```zig
var share = EphemeralKeyShare{
    .group = group,
    .public_key_len = public_len,
    .private_key_len = private_len,
};
errdefer crypto_provider_pkg.secureZero(&share.private_key);
self.crypto_provider.generateKeyShare(
    provider_group,
    share.public_key[0..public_len],
    share.private_key[0..private_len],
) catch return error.SecretExportFailed;
return share;
```

Please also add a provider regression that writes a recognizable partial scalar and then fails, proving the helper clears that exact buffer.

src/tls/tls13_backend.zig
```diff
@@ -2751,13 +2846,21 @@ pub const Tls13Backend = struct {
...
+        len = try checkedAdd(len, 2 + 2 * offered_suites.len); // cipher_suites vector
         len = try checkedAdd(len, 1 + 1); // compression_methods vector + null
         len = try checkedAdd(len, 2); // extensions vector length
         len = try checkedAdd(len, 2 + 2 + 1 + 2 * self.policy.protocol_versions.len); // supported_versions
-        len = try checkedAdd(len, 2 + 2 + 2 + 2 * self.policy.named_groups.len); // supported_groups
+        len = try checkedAdd(len, 2 + 2 + 2 + 2 * offered_groups.len); // supported_groups
         len = try checkedAdd(len, 2 + 2 + 2 + 2 * self.policy.signature_schemes.len); // signature_algorithms
-        len = try checkedAdd(len, 2 + 2 + 2 + 2 + 2 + key_share_public_len); // key_share
+        const key_share_len = switch (self.initial_key_share_mode) {
+            .normal => 2 + 2 + 2 + 2 + try groupPublicLen(offered_groups[0]),
```
Jaruso: **[P1] This omits the `client_shares` vector-length field.** The normal encoding is `extension_type(2) + extension_length(2) + client_shares_length(2) + group(2) + key_exchange_length(2) + public_key`, i.e. `10 + public_key_len`, while the empty encoding is `6` bytes. These branches currently return `8 + public_key_len` and `4`.

Because this value feeds `planPskOffer`, a near-limit ClientHello can retain an offer that does not actually fit and then fail only after leases/lifecycle/key-share state have been mutated. Please use:

```zig
.normal => 2 + 2 + 2 + 2 + 2 + try groupPublicLen(offered_groups[0]),
.empty => 2 + 2 + 2,
```

and add an exact-boundary regression for both normal and empty initial-share modes.

src/tls/tls13_backend_tests.zig
```diff
@@ -3456,6 +3640,27 @@ test "#564 native record-mode loopback still negotiates the AES-128-GCM/SHA-256
     try expectNativeSuiteLoopback(.tls_aes_128_gcm_sha256, 32);
 }
 
+test "#335 record-mode loopback covers the complete native suite/group/signature matrix" {
```
Jaruso, line 3643: **[P1] The required P-256 HRR path is still untested.** Every tuple here configures exactly one group and sends that group’s share in ClientHello1, so none of these 18 loopbacks exercises an HRR selecting `secp256r1`, wiping the superseded private key, generating a fresh P-256 ClientHello2 share, and completing with it. The pre-existing HRR tests remain X25519-only.

Please add a deterministic closeout regression where ClientHello1 advertises both groups but carries only the non-selected share (or an empty share), the server HRRs for P-256, ClientHello2 carries a fresh 65-byte P-256 share, and the handshake completes. The same area should cover the audit’s other explicit transition requirement: remove P-256 from the client provider between ClientHello1 and ClientHello2 and prove the client fails closed without emitting an invalid mutation.

src/quic/connection.zig
```diff
@@ -4110,43 +4168,47 @@ test "driver: client and server complete the handshake over protected packets" {
     }
 }
 
-test "driver: negotiated TLS suites protect QUIC packets end to end" {
+test "driver: negotiated TLS suites and groups protect QUIC packets end to end" {
```
Jaruso, line 4210: **[P2] Complete the QUIC matrix assertions required by #335.** This 3×2 loop verifies the negotiated tuple and application/1-RTT keys, but `pair.pump()` has already discarded Initial and Handshake keys before any profile assertions are made. The closeout audit explicitly requires each suite/group case to prove that Initial remains AES-128-GCM/SHA-256 and that negotiated Handshake and 1-RTT protection follow `suite`.

Please instrument the pair/event flow or pause pumping at the relevant epochs so each tuple asserts the Initial profile before replacement, the Handshake profile after secret installation, and the application profile already checked here.

Jaruso (commented):
**Blocking review — changes required before merge.**

I reviewed all 11 changed files against #335’s final closeout audit and checked the successful CI run for this head. The overall design is on the right track: group selection is now tagged and length-aware, provider capability filtering is applied on both roles, RSA-3072/4096 fixtures are real large-key certificates, and the bounded record/QUIC loops cover the intended tuples.

I found two production defects and two acceptance-coverage gaps:

1. `generateEphemeral` does not wipe the helper’s own partially written private scalar when the provider fails.
2. `clientHelloEncodedLen` undercounts both key-share encodings by two bytes, defeating the preflight/PSK-offer sizing boundary.
3. The closeout still lacks the explicitly required P-256 HelloRetryRequest/fresh-ClientHello2 and mid-HRR group-capability-change regressions.
4. The new QUIC 3×2 matrix checks 1-RTT application protection only; it does not make the required per-tuple assertions that Initial remains fixed and Handshake protection follows the negotiated suite.

CI is green, but none of these cases is exercised by the added tests. Details and concrete fixes are inline. GitHub does not permit this account to submit `REQUEST_CHANGES` on its own PR, so this is a comment review, but I consider these findings merge-blocking.

src/tls/tls13_backend.zig
Jaruso: The provider-failure path is now armed correctly with `errdefer`, and the new failing-provider fixture exercises that branch. However, this helper still returns the secret-bearing `EphemeralKeyShare` **by value**. On success, its local `share.private_key` is copied into the caller's result and the helper returns without wiping the helper-owned temporary (`errdefer` does not run). That leaves an extra private-scalar representation in the helper stack frame and conflicts with the closeout's wipe/minimum-copy requirement.

Please make the caller own the only result storage instead of returning the struct:

```zig
fn generateEphemeral(
    self: *Tls13Backend,
    group: tls_algorithms.NamedGroup,
    out: *EphemeralKeyShare,
) HandshakeError!void {
    const provider_group = try providerGroupOrIllegal(group);
    const public_len = provider_group.publicKeyLength();
    const private_len = try groupPrivateLen(group);
    out.* = .{
        .group = group,
        .public_key_len = public_len,
        .private_key_len = private_len,
    };
    errdefer crypto_provider_pkg.secureZero(&out.private_key);
    self.crypto_provider.generateKeyShare(
        provider_group,
        out.public_key[0..public_len],
        out.private_key[0..private_len],
    ) catch return error.SecretExportFailed;
}
```

Then each call site can declare one local, arm its normal `defer`, call `generateEphemeral(..., &share)`, copy it into `self.key_pair`, and wipe the local as it already does.

The new regression also stores a pointer into this helper's stack-local buffer and reads it after the helper/start frame has returned; that pointer is no longer valid storage, so the test can pass accidentally without being a sound zeroization proof. An out-parameter helper allows an in-module test to keep the output object in the test's own live scope and assert it was cleared after failure.

Jaruso (commented):
**Follow-up review — one P1 remains.**

Re-reviewed amended head `243ee586` against #335 and the four prior threads.

Resolved:

- ClientHello key-share preflight now counts the vector-length field correctly, with emitted-length coverage for both normal and empty modes.
- P-256 HRR now has a deterministic ClientHello2/completion regression, plus the required fail-closed test when capability disappears between ClientHello1 and the retry.
- The QUIC 3×2 matrix now checks fixed Initial protection, negotiated Handshake protection, and application/1-RTT protection for every suite/group tuple.

The provider-failure cleanup itself is also fixed, but `generateEphemeral` now returns `EphemeralKeyShare` by value. Its successful helper-local private scalar is not explicitly wiped, and the new test reads a pointer to that helper stack frame after return, so it is not a sound proof of cleanup. The existing thread contains the out-parameter shape needed to keep one caller-owned secret representation and make failure zeroization observable safely.

CI run **1185** is green. I found no additional functional or matrix-coverage issues in the amended diff, but the remaining secret-lifetime finding is merge-blocking. GitHub does not allow this account to submit `REQUEST_CHANGES` on its own PR, so this is recorded as a comment review.

Jaruso (commented):
**Second-pass review of `243ee586` — one blocking issue remains.**

I re-reviewed the amended head against #335’s closeout audit and the four findings from review #4848113220. The new CI run is green.

Resolved and verified:

- ClientHello key-share size accounting now includes the `client_shares` vector-length field for both normal and empty modes, with an emitted-length parity regression.
- P-256 HRR now has a deterministic ClientHello1 → HRR → fresh 65-byte ClientHello2 share → completed-handshake regression, plus a capability-removal-before-ClientHello2 fail-closed test.
- The QUIC 3-suite × 2-group matrix now checks fixed Initial protection, negotiated Handshake protection on both peers/directions, and 1-RTT application protection.

Still blocking:

- `generateEphemeral` retains a named secret-bearing local and returns `EphemeralKeyShare` by value. Its `errdefer` correctly wipes provider-failure paths, but no explicit wipe runs on success before the helper frame exits. The callers then maintain their own local and backend copies. For this closeout’s minimum-copy/explicit-lifetime contract, make the helper write into caller-owned output storage rather than return the private scalar by value.
- The new failure regression stores `private_out.ptr` inside the provider and dereferences it after `start()` and the helper frame have returned. That pointer no longer names live storage, so the assertion is not a sound zeroization proof even though CI passes. An out-parameter helper enables a test to keep the output object alive in the test scope and inspect it legally after the injected failure.

The existing unresolved inline thread contains the expected out-parameter shape. Once that is applied and the regression is moved onto live caller-owned storage, I do not see another blocker in this head.

GitHub does not permit this account to submit `REQUEST_CHANGES` on its own PR, so this is recorded as a blocking comment review.

Jaruso (commented):
**No new revision detected; previous blocking review still applies.**

I re-fetched PR #571 and the live patch. The head is still `243ee58642d2bd7501096221cad13caf9f1a8981`, the same commit reviewed in follow-up review #4848507185, and the remaining inline thread is still unresolved.

`generateEphemeral` continues to return a secret-bearing `EphemeralKeyShare` by value. Its `errdefer` handles provider failure, but not the successful return path, which can leave the helper-local private-scalar representation unwiped. The associated regression still reads through a pointer into expired helper stack storage, so it does not soundly prove zeroization.

No additional code changes were available to evaluate. The PR remains blocked on the existing private-key ownership/zeroization thread; the previously accepted ClientHello length, P-256 HRR, and QUIC matrix fixes remain accepted.

Jaruso (commented):
Re-reviewed current head `ecbad763` against issue #335 and all prior findings. **No blocking code findings remain.**

The final P1 is addressed correctly:
- `generateEphemeral` now writes into caller-owned `EphemeralKeyShare` storage instead of returning a secret-bearing struct by value.
- Its `errdefer` wipes that live output buffer when the provider partially writes and fails.
- Each success-path caller owns and wipes its one local temporary after copying the active share into backend state.
- The replacement regression invokes the helper directly and asserts zeroization on an object whose lifetime remains valid, eliminating the previous dangling-stack-pointer test.

I also rechecked the previously accepted ClientHello length accounting, P-256 HRR/capability-loss coverage, and QUIC Initial/Handshake/1-RTT matrix assertions; they remain intact. The last unresolved inline thread has been resolved.

CI run 1186 is still finishing at review time. All completed jobs are green, including Linux/macOS/ARM unit tests, appliance tests completed so far, TLS dependency audit, security, builds, and H3 HRR/resumption interop. The ARM appliance full integration tail and packaging smoke were still in progress, so merge should remain gated on the workflow’s final success.

GitHub does not permit this account to formally approve its own PR, so this is recorded as a non-blocking COMMENT review.

PR #572, opened 2026-08-03
security: build shared crypto fuzzing contract and provider targets (327-G4)

Jaruso:
## Summary

Issue #376 is a large research story (`327-G`) whose full scope spans a
shared fuzzing contract, harness support, standalone `CryptoProvider`
targets, and CI wiring, with protocol-specific fuzzing explicitly split out
to sibling stories (#491–#494, #247). This PR implements the story's own
scope: the shared contract and the provider-boundary targets it owns.

- Adds `docs/CRYPTO_FUZZ_CONTRACT.md`: the shared fuzzing contract —
  deterministic reproduction, bounded work, arithmetic safety,
  lifetime/borrowed-slice safety, secret-safe diagnostics, and regression
  minimization — that #491–#494's future protocol-level fuzzing stories are
  expected to consume, following the same pattern established by
  `docs/QUIC_H3_FUZZ_MATRIX.md` for #247.
- Adds `tests/crypto_provider_fuzz.zig`: standalone `std.testing.Smith`
  generative fuzz targets plus deterministic regression tests driving the
  real `pure_zig.Provider` directly through `provider.CryptoProvider` (never
  a mock, never a protocol state machine):
  - **AEAD seal/open** — wrong-length key/nonce/tag/output, mismatched
    ciphertext/plaintext lengths, tampered ciphertext/tag/AD mapped to
    `AuthenticationFailed` with the plaintext buffer fully zeroed, and
    zero-length/harness-bounded-maximum plaintext round-trips, for all three
    supported AEADs.
  - **Key exchange** — wrong-length private scalars/peer public
    keys/output buffers for X25519 and secp256r1, plus generative fuzzing of
    `deriveSharedSecret` and `generateKeyShare`.
  - **Signature verification** — malformed public-key and signature
    encodings, structurally-valid single-bit-modified signatures, and
    wrong-message/wrong-key rejection for Ed25519, ECDSA-P256/SHA-256, and
    RSA-PSS-RSAE/SHA-256 (using the existing `rsa.testdata` fixture rather
    than re-deriving key material).
  - **Signing-key boundary** — undersized-output-buffer rejection and
    `deinit` zeroization verification, filling the specific gaps not already
    covered by `src/crypto/pure_zig.zig`'s existing deep per-scheme coverage.
  - **Shared secret helpers** — `FixedSecret`/`BoundedSecret`
    `replace`/`copy`/`eql`/`deinit` property fuzzing across randomized
    capacities and self-overlapping input, including allocator-failure
    injection via `std.testing.FailingAllocator`.
- Wires a new `test-crypto-provider-fuzz` build step and a
  `-Dcrypto-test-filter` option into `build.zig`, added to both
  `test-crypto` and the default `zig build test` (so it runs in existing CI
  with no workflow changes needed), mirroring the QUIC program's
  `-Dquic-test-filter` convention.
- `CHANGELOG.md` entry under a new `### Testing` section.

This intentionally does **not** re-implement TLS handshake/negotiation
(#491), DER/PEM/X.509 (#492), TLS record (#493), resumption/ticket (#494), or
QUIC/H3 (#247) fuzzing — the issue is explicit that #376 owns only the
contract and the provider seam. It also avoids duplicating the substantial
existing deterministic coverage already in `pure_zig.zig`/`rsa.zig`/
`secrets.zig` (documented per-target in the contract doc's matrix), adding
only the missing generative harnesses and a small number of genuine gaps.

## Test plan

- [x] `zig build test-crypto-provider-fuzz --summary all --error-style
      verbose` — 18/18 new tests pass.
- [x] `zig build test-crypto-provider-fuzz -Doptimize=ReleaseFast
      --fuzz=20000` — real coverage-guided fuzzing run (132k+ runs) with no
      crashes.
- [x] `zig build test --summary all --error-style verbose` — full suite,
      3124/3125 passing (1 pre-existing unrelated skip), no regressions.
- [x] `zig fmt --check build.zig src/ tests/` — clean.

Closes #376

### DIFF
--- a/docs/CRYPTO_FUZZ_CONTRACT.md
+++ b/docs/CRYPTO_FUZZ_CONTRACT.md
@@ -63,45 +63,82 @@ growing this one.
 This is the concrete workflow #491–#494 should follow too, rather than each
 inventing their own.
 
-1. **Reproduce and identify the case.** A `--fuzz=<N>` run that finds a
-   failure fails the `zig build` step and prints the panic/error and Zig's
-   own `test "fuzz: <name>"` name plus its run count in the `FUZZING
-   REPORT` block — that name is the target's deterministic case ID
-   (`Deterministic reproduction` above). Re-run the exact same command
-   (same target, same `-Dcrypto-test-filter`, same `-Doptimize`) to confirm
-   the failure reproduces; Zig's fuzz engine is deterministic given the same
-   binary and inputs.
-2. **Choose the minimization form.**
-   - **Prefer a named deterministic `test` block** next to the fuzz target
-     when the failure has clear structure (a specific field's wrong length,
-     a specific tamper, a specific boundary value) — this is what every
-     deterministic regression in `tests/crypto_provider_fuzz.zig` already
-     is, and it stays reviewable years later the way a raw byte blob does
-     not.
-   - **Add a raw entry to the target's `.corpus = &.{...}` array** only when
-     the failure is genuinely opaque (no clean semantic story) and the exact
-     bytes matter. To capture those bytes: temporarily add a
-     `std.debug.print` of the sampled lengths/enum choice/raw slices
-     immediately before the failing `CryptoProvider` (or `secrets`) call
-     inside the target, re-run the same failing case to trigger the print,
-     copy the reported values, then remove the temporary print. Encode the
-     captured bytes as a Zig string literal using `\xNN` escapes for any
-     non-printable byte (see the existing corpus entries in this file for
-     the style) and add it to `.corpus`.
-3. **Never store real secret material.** Every target in this file
+1. **Capture the exact crash input, not a description of it.** A `--fuzz=<N>`
+   run that finds a failure persists the exact failing `Smith` byte input
+   through Zig's own fuzz engine (the memory-mapped/corpus-backed input the
+   coverage-guided runner uses to recover crashing values, per the [Zig
+   0.16 release notes' fuzzer
+   section](https://ziglang.org/download/0.16.0/release-notes.html#Fuzzer)).
+   The `test "fuzz: <name>"` name Zig prints identifies the *target*, and a
+   `FUZZING REPORT` run count is not a stable case ID — re-running the same
+   coverage-guided command is exploration, not an exact replay. Do not
+   substitute a `std.debug.print` of the *values sampled from* the Smith
+   stream (lengths, enum choices, etc.): those are derived data, not the
+   input itself, and cannot be fed back into `.corpus` to reproduce the
+   same run deterministically. Copy the crash input file Zig's fuzzer
+   reports into a stable checked-in fixture instead:
+   `tests/vectors/fuzz/crypto/<target-slug>/<sha256-of-the-file>.bin`, where
+   `<target-slug>` is a short slug for the target (e.g. `aead-open`,
+   `derive-shared-secret`, `verify`, `fixed-secret`, `bounded-secret`). The
+   SHA-256 digest is both the filename and the deterministic case ID/
+   provenance record — identical bytes always hash to the same fixture name,
+   so a duplicate find is a no-op rather than a second copy.
+2. **Wire it back in with `@embedFile`, always.** Every checked-in crash
+   fixture is added to its target's `.corpus` as `@embedFile`, never
+   hand-transcribed as a string literal (transcription is exactly the kind
+   of lossy transcription step 1 rules out):
+   ```zig
+   try std.testing.fuzz({}, fuzzAeadOpen, .{ .corpus = &.{
+       "",
+       // Found 2026-08-04, zig build test-crypto-provider-fuzz
+       // -Doptimize=ReleaseFast -Dcrypto-test-filter="fuzz: AEAD open ..." --fuzz=50M
+       @embedFile("vectors/fuzz/crypto/aead-open/1f3c9a7e...bin"),
+   } });
+   ```
+   Replay it immediately with the *same* target's filtered non-`--fuzz`
+   command (below) to confirm it reproduces deterministically before doing
+   anything else with it.
+3. **Minimize, then decide what to keep.** Reduce the fixture by repeatedly
+   trimming/simplifying its bytes and re-running the same filtered replay
+   command until it stops reproducing the failure, keeping the last input
+   that still did (manual delta-debugging; Zig 0.16 does not ship an
+   automatic minimizer for this workflow). Once the minimal failing input is
+   understood:
+   - If it has a clean semantic story (a specific field's wrong length, a
+     specific tamper, a specific boundary value), translate it into a named
+     deterministic `test` block next to the target — this is what every
+     regression in `tests/crypto_provider_fuzz.zig` already is, and it
+     stays reviewable in a way a byte blob does not. Only remove the raw
+     `.corpus`/`@embedFile` fixture once that named regression has been
+     confirmed to fail against the pre-fix code and pass against the fix
+     (i.e. it exercises the same defect, not a superficially similar one).
+   - If it has no clean semantic story (the exact bytes matter and cannot be
+     reduced to a short description), keep the minimized fixture itself as
+     the permanent regression via `@embedFile`, re-hashing and renaming the
+     file to match its minimized content.
+4. **Never store real secret material.** Every target in this file
    synthesizes or derives its own key material per case from injected
    deterministic entropy — callers never hand it a real production secret —
-   so captured bytes are inherently synthetic/malformed wire-shaped input,
-   never a real private key, certificate, or traffic capture. Corpus entries
-   and regression fixtures must stay that way: hand-crafted or fuzzer-found
-   malformed input only, never copied from a real deployment.
-4. **Record provenance.** Add a one-line comment directly above the new
-   corpus entry or regression test noting when it was found and which
-   command found it (e.g. `-Doptimize=ReleaseFast --fuzz=10M`), so a future
-   reader can tell a hand-written edge case from a fuzzer-discovered one.
-5. **Verify before committing.** Both must pass:
+   so a captured fixture is inherently synthetic/malformed wire-shaped
+   input, never a real private key, certificate, or traffic capture. Corpus
+   fixtures and regression tests must stay that way: hand-crafted or
+   fuzzer-found malformed input only, never copied from a real deployment.
+5. **Record provenance.** Add a one-line comment directly above the new
+   `.corpus`/`@embedFile` entry or regression test (see the example in step
+   2) noting when it was found and which command found it, so a future
+   reader can tell a hand-written edge case from a fuzzer-discovered one;
+   the SHA-256 filename itself is the immutable half of that record.
+6. **Verify before committing, and pick the right command for the artifact.**
+   A raw `.corpus`/`@embedFile` fixture is replayed by the *fuzz test's own*
+   filter; a named deterministic regression is a separate `test` and needs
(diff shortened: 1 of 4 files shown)


tests/crypto_provider_fuzz.zig
```diff
@@ -0,0 +1,594 @@
...
+const pure_zig = crypto_pkg.pure_zig;
+const rsa = crypto_pkg.rsa;
+const secrets = crypto_pkg.secrets;
+
+const Ed25519 = std.crypto.sign.Ed25519;
+const EcdsaP256Sha256 = std.crypto.sign.ecdsa.EcdsaP256Sha256;
+
+fn cryptoProvider() provider.CryptoProvider {
+    const Holder = struct {
+        var entropy = pure_zig.DeterministicEntropy.init(0x376);
```
Jaruso: **Blocking — the entropy stream is shared across every test and fuzz case.** `Holder.entropy` is mutable static state, so `generateKeyShare` and every signing test consume one process-global stream. A minimized case replayed with `-Dcrypto-test-filter` starts at a different stream position than it had during discovery, and identical Smith input is therefore not case-isolated. That contradicts the deterministic-reproduction contract this PR is introducing; it also creates a race if the test runner ever executes cases concurrently.

Construct the entropy and provider inside each test/fuzz callback so their lifetimes cover the `CryptoProvider` use and the seed resets for each case, for example:

```zig
var entropy = pure_zig.DeterministicEntropy.init(0x376);
var provider_instance = pure_zig.Provider.init(entropy.entropy());
const cp = provider_instance.cryptoProvider();
```

For entropy-consuming fuzz targets, an even stronger pattern is to derive the seed from the fuzz input/case data, but it must not depend on earlier cases. Avoid a helper that returns only `CryptoProvider`, because that view borrows the provider/entropy objects and would outlive locals created inside the helper.

tests/crypto_provider_fuzz.zig
```diff
@@ -0,0 +1,594 @@
...
+    @memset(&pt_buf, 0xAA);
+    cp.aeadOpen(
+        aead,
+        key_buf[0..key_len],
+        nonce_buf[0..nonce_len],
+        ad_buf[0..ad_len],
+        ct_buf[0..ct_len],
+        tag_buf[0..tag_len],
+        pt_buf[0..ct_len],
+    ) catch |err| {
```
Jaruso: **Blocking — unexpected typed failures are treated as successful fuzz cases.** This catch accepts every error, but this target selects only AEADs supported by `pure_zig.Provider`. `InvalidInput` and `AuthenticationFailed` are expected peer/caller outcomes; `UnsupportedCapability` is a provider-contract regression and must fail the target. The same catch-all pattern appears in `fuzzDeriveSharedSecret`, `fuzzGenerateKeyShare`, and `fuzzVerify`, where it also hides unexpected `ProviderFailure`/`EntropyFailure` classifications.

Please switch explicitly on the allowed errors and rethrow everything else across all provider fuzz callbacks. For this target, the shape should be similar to:

```zig
cp.aeadOpen(...) catch |err| switch (err) {
    error.InvalidInput => return,
    error.AuthenticationFailed => {
        for (pt_buf[0..ct_len]) |b| {
            if (b != 0) return error.TestUnexpectedResult;
        }
        return;
    },
    else => return err,
};
```

`deriveSharedSecret`/`verify` should likewise enumerate their expected malformed-input/authentication errors, and `generateKeyShare` should only accept `InvalidInput` for invalid output lengths when using a non-failing deterministic entropy source. Otherwise these fuzzers cannot enforce the issue's typed-failure/error-classification requirement.

tests/crypto_provider_fuzz.zig
```diff
@@ -0,0 +1,594 @@
...
+    const value_len = smith.index(capacity + 1);
+    smith.bytes(scratch[0..value_len]);
+
+    // fail_index in {0, 1}: 0 injects failure on BoundedSecret's single
+    // capacity allocation; 1 lets it succeed, exercising the ordinary path.
+    const fail_index = smith.index(2);
+    var failing = std.testing.FailingAllocator.init(fba.allocator(), .{ .fail_index = fail_index });
+
+    var secret = secrets.BoundedSecret{};
+    secret.init(failing.allocator(), capacity, scratch[0..value_len]) catch |err| switch (err) {
```
Jaruso: **Blocking — this target does not exercise the properties its name, PR description, and contract matrix claim.** After `init`, it only compares one equal mirror and lets `deinit` run through `defer`. It never calls `BoundedSecret.replace`, so it does not cover shrink/grow replacement, self-overlap, oversized replacement preserving the prior value, or replacement-tail clearing. It never tests unequal/length-mismatched equality, never explicitly inspects backing memory after `deinit`, and `FixedBufferAllocator` is not the `std.testing.allocator` leak check described in the document. On injected OOM it also returns without asserting that the object remained empty/unowned.

Please add an operation loop comparable to `fuzzFixedSecret`, snapshot the value before rejected replacements, include self-overlapping replacement slices, and explicitly call `deinit` so the FBA backing can be checked for zeroization. Use a separate `FailingAllocator` over `std.testing.allocator` path (or another tracking allocator) to verify allocation-failure cleanup/leak freedom, since the FBA path is useful for byte inspection but not leak detection. Either implement those properties or narrow the test/docs/PR claims, but the current implementation does not satisfy #376's required `BoundedSecret` allocation/replace/equality/deinit property coverage.

Jaruso (commented):
I found four blocking acceptance-criteria gaps. I am submitting this as a comment review because the authenticated account is the PR author and cannot formally request changes on its own pull request.

The overall scope split, provider-boundary focus, build wiring, existing-test reuse, and protocol ownership matrix are sound, and CI is green. However, #376 is not complete until the harness provides case-isolated deterministic reproduction, rejects unexpected provider error classes instead of swallowing them, actually exercises the documented `BoundedSecret` replacement/deinit/allocation properties, and documents the required seed-corpus update procedure.

Please address the inline findings before merging.

docs/CRYPTO_FUZZ_CONTRACT.md
```diff
@@ -0,0 +1,239 @@
...
+zig build test-crypto-provider-fuzz -Doptimize=ReleaseFast -Dcrypto-test-filter="fuzz: FixedSecret" --fuzz=10M --summary all --error-style verbose
+zig build test-crypto-provider-fuzz -Doptimize=ReleaseFast -Dcrypto-test-filter="fuzz: BoundedSecret" --fuzz=10M --summary all --error-style verbose
+```
+
+`-Dcrypto-test-filter` (added alongside the existing `-Dquic-test-filter`)
+gives explicit target selection for the `test-crypto-provider-fuzz` step; the
+`--fuzz=<runs>` limit keeps scheduled runs bounded (`K`/`M`/`G` suffixes
+scale it). #491–#494 should add their own protocol-scoped test steps and
+`-D<area>-test-filter` options following the same pattern rather than
+growing this one.
```
Jaruso, line 59: **Blocking — the required seed-corpus update procedure is missing.** This section documents how to run smoke and long fuzz jobs, and later sections say regressions must be minimized, but #376 explicitly requires both longer-run commands **and seed-corpus update procedures**. Please add a concrete workflow covering: how to capture Zig's minimized input/case ID; when to add it to the inline `.corpus` versus a named deterministic regression; how to encode/store binary input without including secret material; how to record target/provenance; and which filtered replay plus full-suite commands must pass before committing the update. Without that, #491–#494 still have to invent this part of the shared harness rules independently.

docs/CRYPTO_FUZZ_CONTRACT.md
```diff
@@ -0,0 +1,300 @@
...
+
+## Seed-corpus and regression update procedure
+
+This is the concrete workflow #491–#494 should follow too, rather than each
+inventing their own.
+
+1. **Reproduce and identify the case.** A `--fuzz=<N>` run that finds a
+   failure fails the `zig build` step and prints the panic/error and Zig's
+   own `test "fuzz: <name>"` name plus its run count in the `FUZZING
+   REPORT` block — that name is the target's deterministic case ID
```
Jaruso: **Blocking — this still does not provide an exact, deterministic case capture/replay procedure.** In Zig 0.16, the crashing Smith input is saved to a file named in the crash message and is intended to be replayed by adding that file to `FuzzInputOptions.corpus` via `@embedFile`. The test name identifies the target, not the case, and the run count is not a stable case ID. Re-running the same coverage-guided command may rediscover the input, but it is not the exact local reproduction command required by #376.

The temporary `std.debug.print` workflow below is also lossy: it records values generated *from* the Smith stream rather than preserving the exact Smith input, and it does not define a real minimization step. Please make the crash artifact itself authoritative. For example:

```zig
try std.testing.fuzz({}, fuzzAeadOpen, .{ .corpus = &.{
    "",
    @embedFile("vectors/fuzz/crypto/aead-open/<sha256>.bin"),
} });
```

A complete procedure should be roughly:

1. Copy the crash file path printed by Zig into a stable fixture location.
2. Compute its SHA-256 (or similar) and use that digest as the deterministic case ID/provenance.
3. Replay it immediately with the filtered non-fuzz command after adding it through `@embedFile`.
4. Reduce the Smith input while repeatedly running that exact replay, or translate it into a smaller semantic named regression and verify the original crash fixture no longer adds distinct value before removing it.
5. Keep the fixture when the raw decision stream is necessary; otherwise keep the named regression.

Also split the verification command by artifact type. The current

```bash
-Dcrypto-test-filter="fuzz: <exact target name>"
```

replays a corpus fixture, but it will **not run a separately named deterministic regression**. For that path, filter on the regression test's actual name (or run the whole provider-fuzz step) before the full suite. Zig's 0.16 release notes explicitly document crash-file replay through `.corpus` + `@embedFile`: https://ziglang.org/download/0.16.0/release-notes.html#Fuzzer

Jaruso (commented):
Re-review complete against issue #376 and the prior four blockers.

The three code fixes are solid: provider/entropy state is now case-local, provider fuzz callbacks preserve typed error classification, and the rewritten `BoundedSecret` target now exercises replacement, overlap, rejection atomicity, equality variants, allocation failure/leak cleanup, and observable deinit zeroization. CI is green on the new head.

One acceptance-level blocker remains in the newly added seed-corpus procedure. Zig 0.16 writes the exact crashing Smith input to a file named in the crash output; the target name and run count are not a deterministic case ID, and rerunning the same coverage-guided command is not an exact replay mechanism. The workflow also uses the fuzz-test filter even when the chosen artifact is a separately named deterministic regression, so that command would not execute the regression it claims to verify.

Please update the procedure to preserve/hash the emitted crash input, replay it directly through `.corpus`/`@embedFile`, document an actual reduction/minimization step, and select the filtered command based on whether the result is a corpus fixture or a named regression. After that, I do not see another blocker.

Jaruso (commented):
Re-review complete — no further findings.

The remaining blocker is fully addressed:

- the exact Zig crash-input artifact is authoritative rather than a target name/run count or derived debug output;
- checked-in raw regressions use a content-addressed SHA-256 fixture and `@embedFile`;
- the procedure defines exact filtered replay before minimization;
- minimization operates on the original Smith decision stream;
- raw corpus fixtures and named semantic regressions have correctly separated verification commands;
- deterministic case IDs and regression-minimization guidance are consistent throughout the contract.

I also re-checked the earlier fixes and the full #376 acceptance criteria. Case-local entropy, typed failure enforcement, `BoundedSecret` property coverage, build/CI wiring, ownership boundaries, diagnostics guidance, and longer-run commands remain sound. CI for head `318f2c5d46c0e439d1b2818dff9e79def118cae5` is green.

This is ready to merge. GitHub does not allow me to formally approve my own PR, so this COMMENT review represents approval.

PR #573, opened 2026-08-04
perf: [327-I] benchmark CryptoProvider and shared cryptographic machinery

Jaruso:
## Summary

- Adds a repeatable Zig-native benchmark suite (`zig build bench-crypto`) for the shared `CryptoProvider` seam and its direct consumers, driving `pure_zig.Provider` the same own-stack-frame way `tests/crypto_provider_fuzz.zig` does — never a mock, never a protocol state machine.
- Covers: hash (explanatory)/HKDF, AEAD seal/open/auth-failure for all 3 suites, X25519/secp256r1 key exchange (including invalid-peer-key rejection), Ed25519/ECDSA-P256/RSA-PSS sign/verify/reject, `FixedSecret`/`BoundedSecret`/`secureZero`/`constantTimeEqual` overhead, TLS record-layer traffic-key derivation + seal/open + auth-failure via `record_protection`, and ticket/resumption codec size sweeps + `Protector.seal`/`resolve` + resolver-miss paths + keyring/nonce-lease overhead (including a real multi-threaded nonce-reservation contention workload) via `ticket_protection`.
- Every measurement reports Zig/OS/arch/build-mode/provider metadata as JSON, matching the issue's benchmark contract.
- Provider-primitive allocation-freedom is enforced **structurally** at compile time (`CryptoProvider.VTable` methods take no allocator param) rather than by fragile runtime sampling; a handful of core workloads carry generous advisory latency ceilings that warn but never fail the build, per the issue's "budgets may be advisory/trend-based" allowance.
- `test-crypto-bench` runs every workload at a tiny iteration count as a correctness/API-drift smoke check and is part of the default `zig build test`; `bench-crypto` runs full iteration counts and prints the real report.
- CI gains a non-blocking `crypto-bench` job that uploads the JSON report as a diffable trend artifact — it never inspects the numbers or fails on them.
- New doc: `docs/CRYPTO_BENCHMARKS.md`.

**Deliberately out of scope**, per the issue's own ownership boundaries:
- QUIC packet/header-protection crypto benchmarks — blocked on #490 reconciling QUIC provider ownership.
- An OpenSSL reference comparison — the issue frames this as optional ("where useful"); left as follow-on work.
- All whole-protocol (TLS/PKI/QUIC/H3/HTTP) end-to-end benchmarks, which stay with their owning stories (#323–#326/#369/#247/#149/#150 and the root `benchmarks/` HTTP suite).

## Test plan

- [x] `zig build test --summary all` — 3125/3126 passed (1 pre-existing unrelated skip), including the new `test-crypto-bench` smoke suite.
- [x] `zig build test -Dtls-profile=appliance --summary all` — 3094/3095 passed (1 pre-existing unrelated skip).
- [x] `zig fmt --check` on all new/changed `.zig` files and `build.zig`.
- [x] `zig build bench-crypto -Doptimize=ReleaseFast` — runs end to end (~14s) and prints a well-formed JSON report; spot-checked timings for plausibility (found and fixed a compiler loop-invariant-hoisting artifact in the tiniest leaf workloads — see `primitives.zig` comments and `docs/CRYPTO_BENCHMARKS.md`'s "A note on nanosecond-scale numbers").
- [x] `.github/workflows/ci.yml` validated with `python3 -c "import yaml; yaml.safe_load(...)"`.

Closes #378

### DIFF
--- a/src/crypto_bench/harness.zig
+++ b/src/crypto_bench/harness.zig
@@ -1,0 +1,241 @@
+//! Shared measurement/report plumbing for the CryptoProvider benchmark suite
+//! (#378, epic #327-I). Every suite (`primitives.zig`, `record.zig`,
+//! `ticket.zig`) times its own workloads with `monotonicNs` and reports
+//! through the `Measurement`/`Report` types here so `root.zig` can print one
+//! consistent JSON document regardless of which suite produced a result.
+
+const std = @import("std");
+const builtin = @import("builtin");
+const build_options = @import("build_options");
+const compat = @import("zig_compat");
+
+/// Zig 0.16 removed `std.time.Timer`; `std.Io.Clock.awake` is this
+/// toolchain's monotonic (not wall-clock/NTP-adjustable) clock, matching
+/// what `Timer` used to wrap. See the `Clock.awake` doc comment in
+/// `std.Io` and `zig_compat.sleepNs`'s identical choice.
+pub fn monotonicNs() i96 {
+    return std.Io.Clock.awake.now(compat.io()).toNanoseconds();
+}
+
+/// How hard each suite should push. `smoke` runs a handful of iterations so
+/// `zig build test-crypto-bench` finishes in well under a second per
+/// workload and is safe to run under the default `zig build test` — it
+/// exists to catch crashes/panics/API drift, not to produce trustworthy
+/// timings. `full` runs enough iterations for the printed ns/op to be
+/// meaningful and is what `zig build bench-crypto` uses.
+pub const Scale = enum {
+    smoke,
+    full,
+
+    pub fn iterations(self: Scale, full_iterations: usize) usize {
+        return switch (self) {
+            .smoke => @min(full_iterations, 5),
+            .full => full_iterations,
+        };
+    }
+};
+
+/// A minimal allocation counter, wrapping a child allocator to report how
+/// many allocations and bytes a workload performed. Only wired around calls
+/// that take an `std.mem.Allocator` in the first place — the `CryptoProvider`
+/// vtable (HKDF, AEAD, key exchange, verify) never does, so those workloads
+/// are allocation-free by construction rather than by runtime measurement
+/// (see `primitives.assertProviderVtableIsAllocatorFree` in primitives.zig).
+pub const CountingAllocator = struct {
+    child: std.mem.Allocator,
+    allocations: usize = 0,
+    frees: usize = 0,
+    bytes_allocated: usize = 0,
+
+    pub fn init(child: std.mem.Allocator) CountingAllocator {
+        return .{ .child = child };
+    }
+
+    pub fn allocator(self: *CountingAllocator) std.mem.Allocator {
+        return .{ .ptr = self, .vtable = &vtable };
+    }
+
+    fn alloc(ctx: *anyopaque, len: usize, alignment: std.mem.Alignment, ret_addr: usize) ?[*]u8 {
+        const self: *CountingAllocator = @ptrCast(@alignCast(ctx));
+        const ptr = self.child.rawAlloc(len, alignment, ret_addr) orelse return null;
+        self.allocations += 1;
+        self.bytes_allocated += len;
+        return ptr;
+    }
+
+    fn resize(ctx: *anyopaque, memory: []u8, alignment: std.mem.Alignment, new_len: usize, ret_addr: usize) bool {
+        const self: *CountingAllocator = @ptrCast(@alignCast(ctx));
+        return self.child.rawResize(memory, alignment, new_len, ret_addr);
+    }
+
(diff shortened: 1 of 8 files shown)


src/crypto_bench/harness.zig
```diff
@@ -0,0 +1,241 @@
...
+    }
+
+    pub fn items(self: *const Reporter) []const Measurement {
+        return self.list.items;
+    }
+
+    pub fn deinit(self: *Reporter) void {
+        self.list.deinit(self.allocator);
+    }
+};
```
Jaruso, line 188: **Blocking — the benchmark contract requires the exact Tardigrade commit.** `build_options.version` identifies a release/version, not the source revision, so two PR commits with the same version produce metadata that cannot be reliably compared or attributed. Please add a build option populated from Git (for example `git rev-parse HEAD`, with an explicit fallback such as `unknown` for source archives) and serialize it as `tardigrade_commit` in `_meta`.

For example:
```zig
pub const Metadata = struct {
    tardigrade_version: []const u8,
    tardigrade_commit: []const u8,
    // ...
};
```
Then wire the commit value into the benchmark module from `build.zig`. This directly satisfies issue #378's first benchmark-contract bullet.

src/crypto_bench/ticket.zig
```diff
@@ -0,0 +1,644 @@
...
+    defer state.deinit();
+    var protector = ticket_protection.Protector{ .provider = cp, .keyring = &keyring, .limits = session.Limits.default };
+
+    const Worker = struct {
+        fn run(protector_ptr: *ticket_protection.Protector, state_ptr: *const session.ServerRecoverableState, iterations: usize) void {
+            var out_buf: [4096]u8 = undefined;
+            var i: usize = 0;
+            while (i < iterations) : (i += 1) {
+                _ = protector_ptr.seal(std.heap.page_allocator, state_ptr, 1500, &out_buf) catch return;
+            }
```
Jaruso, line 640: **Blocking — worker failures are swallowed, so this benchmark can report success and a full iteration count after doing only partial work.** `catch return` exits a worker on the first error, but the parent has no shared error/completion state and always records `thread_count * per_thread_iterations`. That means nonce exhaustion, an unexpected keyring error, or another regression could make both the smoke test and full benchmark silently produce a misleading measurement.

Please propagate failures and verify completed operations. One approach is a shared atomic completion counter plus a mutex-protected first error (or an atomic failure flag if preserving the exact error is unnecessary):
```zig
const Shared = struct {
    completed: std.atomic.Value(usize) = .init(0),
    failed: std.atomic.Value(bool) = .init(false),
};

fn run(shared: *Shared, protector: *Protector, state: *const State, iterations: usize) void {
    // ...
    protector.seal(...) catch {
        shared.failed.store(true, .release);
        return;
    };
    _ = shared.completed.fetchAdd(1, .monotonic);
}
```
After joining, return an error if `failed` is set or `completed != thread_count * per_thread_iterations`, and use `completed` as the reported iteration count. This is especially important for a nonce-lease benchmark because issue #378 explicitly forbids weakening nonce-uniqueness guarantees for benchmark results.

Jaruso (commented):
I reviewed all eight changed files against issue #378 and the current CI run. The overall structure is strong, the primitive/record/ticket coverage is broad, the documentation is thoughtful, and CI is green. I found two correctness/acceptance-criteria issues that should be addressed before merge:

1. The benchmark contract explicitly requires the Tardigrade commit, but `_meta` reports only `build_options.version`; this makes artifacts from different commits of the same version indistinguishable.
2. The contended nonce-reservation workload silently treats worker failures as successful iterations. A worker returns immediately on the first `Protector.seal` error, but the parent always records `thread_count * per_thread_iterations`, so the smoke test and benchmark can pass while the workload only partially ran. This is particularly important because nonce uniqueness/exhaustion is a security invariant the issue says benchmarks must not weaken.

Please propagate worker failures and record only completed work (or fail the workload), and add the exact commit SHA to the report metadata. I also recommend adding explicit concurrency/state/key-count fields rather than encoding those parameters in `note`/`input_bytes`, but I am not blocking on that if the two correctness issues above are fixed.

Jaruso (commented):
Re-review complete. Both prior blocking findings are fixed correctly:

- Benchmark metadata now includes an exact `tardigrade_commit`, populated from `git rev-parse HEAD` with an explicit `-Dcommit` override and `unknown` fallback for non-Git source trees.
- The contended nonce-reservation workload now records completed operations atomically, propagates worker failure to the parent, verifies the expected operation count after joining, and reports the actual completed count.

I rechecked the affected build wiring, JSON serialization, benchmark correctness path, and issue #378 acceptance criteria. The latest CI run is green, including the crypto benchmark trend job. I found no additional blocking issues in this re-review.

GitHub does not allow approving your own pull request, so this is submitted as a comment review; functionally, this is clear to merge from my review.

PR #574, opened 2026-08-04
security: [494-A] fuzz NewSessionTicket, session codec, PSK, and ticket-envelope parsing

Jaruso:
## Summary

First slice (#494-A, per the issue's own recommended PR decomposition) of #494's fuzz/property coverage for the pure-Zig TLS resumption stack: codecs and allocation-free parsing. Follows the shared contract in `docs/CRYPTO_FUZZ_CONTRACT.md` and #376's `std.testing.fuzz`/`Smith` pattern (already used by `ticket_key_snapshot.zig`/`sni_provider.zig`).

- **`NewSessionTicket` wire codec + owned-state construction** (`src/tls/new_session_ticket.zig`): `decode` never panics or retains a borrow past its input buffer; `encode`/`decode` structurally round-trips generated valid params (`encodedLen(params) == encoded.len` asserted directly); `OutputTooSmall` never partially writes the destination; drives `buildClientTicketState`/`buildServerRecoverableStateNoIdentity` from the round-tripped value with a fresh per-case provider and an RMS sized to the selected suite's digest length, poisoning the source after construction to prove the resulting state owns its ticket/nonce bytes, and asserting the exact `TicketTooLarge`/`InvalidLifetime` outcome when the generated shape should fail.
- **Session-state codec** (`src/tls/session.zig`): raw `decode` never panics on arbitrary bytes and synthetic `Limits` (an explicit zero/one-under/exact/one-over/`0xffff` boundary matrix for every `Limits` field, including deliberately invalid values, which must fail with `error.InvalidLimits`), and a successful decode owns its state (verified via pointer-disjointness from the input, not snapshot/poison); generated client/server records round-trip fully canonically (`clientEncodedLenWithLimits`/`serverEncodedLenWithLimits` match, and re-encoding the *decoded* record reproduces the original bytes exactly, not just a handful of spot-checked fields) and drive `evaluateCompatibility`; a valid client record can never be accepted as a server record or vice versa (flipped record-kind byte).
- **PSK modes/`OfferedPsks`/binder primitives** (`src/tls/pre_shared_key.zig`): `hasMode`/`OfferedPsks.parse` never panic on raw bytes and parsed slices stay within the input; `writeModes`/`writeOffer`'s pre-write structural validation is fully transactional (writer length *and* every output byte unchanged on rejection); generated valid offers round-trip through `parse`; `deriveBinderFromTranscriptHash`/`verifyBinderFromTranscriptHash` are deterministic across both SHA-256 and SHA-384 with PSK/transcript-hash/output/candidate lengths chosen independently around the digest length (so SHA-384's one-over case and wrong-length transcript hashes are actually exercised, not silently collapsed back to a valid length); `deriveBinder`/`verifyBinder` reject one-bit mutations of the transcript hash, PSK, binder, *and* identity bytes (proving the identity is actually included in the transcript at the correct boundary); `obfuscateTicketAge`/`deobfuscateTicketAge`/`observeAgeSkew` arithmetic never panics.
- **Allocation-free ticket-envelope parsing** (`src/tls/ticket_protection.zig`): a *dedicated* `parseEnvelope`-only target (no allocator/provider/keyring/clock — distinct from the existing combined `fuzzTicketIdentity` target, which also drives authenticated `Protector.resolve` and is #494-B's scope) proving every field of `ParsedEnvelope` is deterministic and returns ordered, non-overlapping `header`/`ciphertext`/`tag` views under raw-byte mutation; a single-field-mutation target that asserts by class (typed errors for magic/version/reserved-flags, a guaranteed-changed AEAD id, and required continued success for key-id/nonce/ciphertext/tag mutations, which are opaque until authenticated open) plus synthetic `Limits` boundaries; a deterministic regression exhaustively covering every truncation boundary of a valid envelope.

Wires a dedicated `zig build test-tls-resumption-fuzz` step with `-Dtls-resumption-test-filter` (defaults to the `"fuzz: TLS resumption:"` namespace — not a bare `"fuzz: "` prefix, which would also pull in unrelated pre-existing fuzz targets already living in `tls_core_mod`: `sni_provider.zig`'s two targets, `ticket_key_snapshot.zig`'s, and `ticket_protection.zig`'s own combined identity/resolve target) and documents the step/example reproduction commands, including the named deterministic regressions selected by their own name, in `CRYPTO_FUZZ_CONTRACT.md`.

Existing deep deterministic coverage in each module (truncation matrices, field mutation, allocation-failure sweeps, byte-stability fixtures, etc.) is preserved, not duplicated or weakened — these targets add continuous mutation, cross-operation round trips, and ownership/lifetime invariants on top of it.

Authenticated ticket open/keyring publication, session-cache/lease state machines, and backend/runtime composition remain #494-B/C/D per the issue's own PR decomposition.

## Test plan

- [x] `zig build test-tls-resumption-fuzz --summary all --error-style verbose` (9/9 pass)
- [x] `zig build test-tls --summary all --error-style verbose` (883/883 pass)
- [x] `zig build test --summary all --error-style verbose` (3144/3145 pass, 1 pre-existing skip)
- [x] `zig build test -Dtls-profile=appliance --summary all --error-style verbose` (3113/3114 pass, 1 pre-existing skip)
- [x] `-Dtls-resumption-test-filter` spot-checked to select individual fuzz targets and individual named deterministic regressions

Closes part of #494 (this PR covers only the #494-A slice; #494-B/C/D remain open per the issue).

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -37,6 +37,15 @@ pub fn build(b: *std.Build) void {
     const quic_test_filters: []const []const u8 = if (quic_test_filter) |filter| &.{filter} else &.{};
     const crypto_test_filter = b.option([]const u8, "crypto-test-filter", "Filter tests run by the test-crypto-provider-fuzz step");
     const crypto_test_filters: []const []const u8 = if (crypto_test_filter) |filter| &.{filter} else &.{};
+    // #494: unlike the crypto/QUIC provider-fuzz steps, this step's root
+    // module (tls_core_mod) is the *entire* pure-Zig TLS suite -- also run
+    // unfiltered by `test-tls`/`test` -- not a dedicated fuzz-only root
+    // file, so an unfiltered default here would just re-run the whole
+    // suite under a second name. Default to the "fuzz: " test-name prefix
+    // every #494 target above uses, scoping the step to just those targets;
+    // pass an explicit filter to select one target for a longer local run.
+    const tls_resumption_test_filter = b.option([]const u8, "tls-resumption-test-filter", "Filter tests run by the test-tls-resumption-fuzz step (default: \"fuzz: \", i.e. only the #494 session/PSK/ticket/resumption fuzz targets)") orelse "fuzz: ";
+    const tls_resumption_test_filters: []const []const u8 = &.{tls_resumption_test_filter};
     const tls_profile = b.option(
         TlsProfile,
         "tls-profile",
@@ -217,6 +226,20 @@ pub fn build(b: *std.Build) void {
     const tls_step = b.step("test-tls", "Run pure-Zig TLS core unit tests");
     tls_step.dependOn(&run_tls_core_tests.step);
     test_step.dependOn(&run_tls_core_tests.step);
+
+    // Session/PSK/ticket/resumption fuzz targets (#494, epic #326-K),
+    // following the shared contract in docs/CRYPTO_FUZZ_CONTRACT.md and
+    // #376's test-crypto-provider-fuzz naming pattern. The targets
+    // themselves are inline `test "fuzz: ..."` blocks next to the code
+    // they exercise (new_session_ticket.zig, session.zig,
+    // pre_shared_key.zig, ticket_protection.zig) and already replay their
+    // deterministic seed corpus under plain `test-tls`/`test`; this step
+    // exists to give them a stable, individually filterable/long-runnable
+    // name, matching #376's `-Dcrypto-test-filter` shape.
+    const tls_resumption_fuzz_tests = b.addTest(.{ .root_module = tls_core_mod, .filters = tls_resumption_test_filters });
+    const run_tls_resumption_fuzz_tests = b.addRunArtifact(tls_resumption_fuzz_tests);
+    const tls_resumption_fuzz_step = b.step("test-tls-resumption-fuzz", "Run TLS session/PSK/ticket/resumption fuzz targets (#494)");
+    tls_resumption_fuzz_step.dependOn(&run_tls_resumption_fuzz_tests.step);
 
     const allocation_regression_mod = b.createModule(.{
         .root_source_file = b.path("src/allocation_regression.zig"),
--- a/build.zig
+++ b/build.zig
@@ -41,10 +41,15 @@ pub fn build(b: *std.Build) void {
     // module (tls_core_mod) is the *entire* pure-Zig TLS suite -- also run
     // unfiltered by `test-tls`/`test` -- not a dedicated fuzz-only root
     // file, so an unfiltered default here would just re-run the whole
-    // suite under a second name. Default to the "fuzz: " test-name prefix
-    // every #494 target above uses, scoping the step to just those targets;
-    // pass an explicit filter to select one target for a longer local run.
-    const tls_resumption_test_filter = b.option([]const u8, "tls-resumption-test-filter", "Filter tests run by the test-tls-resumption-fuzz step (default: \"fuzz: \", i.e. only the #494 session/PSK/ticket/resumption fuzz targets)") orelse "fuzz: ";
+    // suite under a second name. `tls_core_mod` also already contains
+    // unrelated pre-existing "fuzz: ..." tests (sni_provider.zig,
+    // ticket_key_snapshot.zig, ticket_protection.zig's own combined
+    // identity/resolve target), so a bare "fuzz: " prefix would silently
+    // pull those in too. Every #494-A target below is therefore named
+    // "fuzz: TLS resumption: ..." and this defaults to that exact
+    // namespace, scoping the step to only those targets; pass an explicit
+    // filter to select one target for a longer local run.
+    const tls_resumption_test_filter = b.option([]const u8, "tls-resumption-test-filter", "Filter tests run by the test-tls-resumption-fuzz step (default: \"fuzz: TLS resumption:\", i.e. only the #494 session/PSK/ticket/resumption fuzz targets)") orelse "fuzz: TLS resumption:";
     const tls_resumption_test_filters: []const []const u8 = &.{tls_resumption_test_filter};
     const tls_profile = b.option(
         TlsProfile,
--- a/src/tls/new_session_ticket.zig
+++ b/src/tls/new_session_ticket.zig
@@ -1009,14 +1009,20 @@ pub fn fuzzNewSessionTicket(allocator: std.mem.Allocator, smith: *std.testing.Sm
     // structural round trips and owned-state construction. Fields are
     // bounded by construction (buffer sizes / modulo) rather than
     // allowing arbitrary unbounded values, per the #376 contract.
+    // `ticket_buf` reaches exactly `session.Limits.default.max_ticket_len`
+    // so generated cases exercise the configured default boundary, not
+    // only small tickets; the absolute wire maximum
+    // (`absolute_ticket_wire_max`) is already deterministically covered
+    // by "server recoverable state honors caller ticket limits" above.
     const direction = smith.index(2);
     const suite = fuzz_cipher_suites[smith.index(fuzz_cipher_suites.len)];
+    const digest_len = algorithms.transcriptHash(suite).digestLength();
 
     var nonce_buf: [max_ticket_nonce_len]u8 = undefined;
     const nonce_len = smith.index(nonce_buf.len + 1);
     smith.bytes(nonce_buf[0..nonce_len]);
 
-    var ticket_buf: [1024]u8 = undefined;
+    var ticket_buf: [session.Limits.default.max_ticket_len]u8 = undefined;
     const ticket_len = 1 + smith.index(ticket_buf.len);
     smith.bytes(ticket_buf[0..ticket_len]);
 
@@ -1041,11 +1047,15 @@ pub fn fuzzNewSessionTicket(allocator: std.mem.Allocator, smith: *std.testing.Sm
         .max_early_data_size = max_early_data_size,
     };
 
-    var encode_buf: [2048]u8 = undefined;
+    // Comfortably above the worst-case encoded length for the generated
+    // shapes above (up to 4096-byte ticket + 255-byte nonce + extension +
+    // fixed fields), so a non-trivial `encode_cap` routinely succeeds
+    // instead of `OutputTooSmall` dominating every large-ticket case.
+    var encode_buf: [8192]u8 = undefined;
     const encode_cap = smith.index(encode_buf.len + 1);
     const scratch = encode_buf[0..encode_cap];
     @memset(scratch, 0xa5);
-    var before_buf: [2048]u8 = undefined;
+    var before_buf: [8192]u8 = undefined;
     @memcpy(before_buf[0..encode_cap], scratch);
 
     const encoded = encode(params, scratch) catch {
@@ -1054,6 +1064,7 @@ pub fn fuzzNewSessionTicket(allocator: std.mem.Allocator, smith: *std.testing.Sm
         try std.testing.expectEqualSlices(u8, before_buf[0..encode_cap], scratch);
         return;
     };
+    try std.testing.expectEqual(try encodedLen(params), encoded.len);
     const round = try decode(encoded);
     try std.testing.expectEqual(params.ticket_lifetime, round.ticket_lifetime);
     try std.testing.expectEqual(params.ticket_age_add, round.ticket_age_add);
@@ -1062,9 +1073,14 @@ pub fn fuzzNewSessionTicket(allocator: std.mem.Allocator, smith: *std.testing.Sm
     try std.testing.expectEqual(params.max_early_data_size, round.max_early_data_size);
 
     // 3. Owned-state construction from the round-tripped `Parsed` value,
-    // exercising both client and server construction directions.
+    // exercising both client and server construction directions. `rms` is
+    // sized to exactly the selected suite's transcript-hash digest length
+    // (`KeySchedule.resumptionPsk` requires an exact match); a fixed
+    // 48-byte slice would make both SHA-256 suites always fail with
+    // `InvalidSecretLength`, silently skipping owned-state construction
+    // for two of the three suites.
     var rms: [session.max_psk_len]u8 = undefined;
-    smith.bytes(&rms);
+    smith.bytes(rms[0..digest_len]);
     var limits_control: [1]u8 = undefined;
     smith.bytes(&limits_control);
     const limits = fuzzLimits(limits_control[0]);
@@ -1078,24 +1094,48 @@ pub fn fuzzNewSessionTicket(allocator: std.mem.Allocator, smith: *std.testing.Sm
     smith.bytes(&received_at_bytes);
     const received_at_unix_ms: i64 = @bitCast(std.mem.readInt(u64, &received_at_bytes, .big));
 
+    // Fresh per-case deterministic entropy/provider pair, on this call's
+    // own stack frame: the #376 contract forbids sharing one
+    // process-global entropy stream across cases (unlike
+    // `testCryptoProvider()` above, which this file's *deterministic*
+    // tests intentionally share).
+    const pure_zig = @import("crypto").pure_zig;
+    var fuzz_entropy = pure_zig.DeterministicEntropy.init(0x1cec_5);
+    var fuzz_provider = pure_zig.Provider.init(fuzz_entropy.entropy());
+    const crypto_provider = fuzz_provider.cryptoProvider();
+
     if (direction == 0) {
-        var maybe_state = buildClientTicketState(
+        const build_result = buildClientTicketState(
             allocator,
-            testCryptoProvider(),
+            crypto_provider,
             round,
             context,
-            &rms,
+            rms[0..digest_len],
             received_at_unix_ms,
             limits,
-        ) catch return;
-        if (maybe_state) |*state| {
-            defer state.deinit();
-            try std.testing.expectEqual(round.ticket.len, state.ticket.slice().len);
+        );
+        if (round.ticket_lifetime == 0) {
+            // `buildClientTicketState` returns `null` (not an error) for a
+            // zero lifetime, before the ticket-size check below ever runs.
+            try std.testing.expectEqual(@as(?session.ClientTicketState, null), try build_result);
+            return;
         }
+        if (round.ticket.len > limits.max_ticket_len) {
+            try std.testing.expectError(error.TicketTooLarge, build_result);
+            return;
+        }
+        var state = (try build_result).?;
+        defer state.deinit();
+        // The state must own its ticket/nonce bytes, not borrow `round`'s
+        // view into `scratch`: poison `scratch` after construction and
+        // confirm the state's content survives.
+        @memset(scratch, 0xee);
+        try std.testing.expectEqualSlices(u8, params.ticket, state.ticket.slice());
+        try std.testing.expectEqualSlices(u8, params.ticket_nonce, state.ticket_nonce.slice());
     } else {
-        var state = buildServerRecoverableStateNoIdentity(
+        const build_result = buildServerRecoverableStateNoIdentity(
             allocator,
-            testCryptoProvider(),
+            crypto_provider,
             .{
                 .ticket_lifetime = round.ticket_lifetime,
                 .ticket_age_add = round.ticket_age_add,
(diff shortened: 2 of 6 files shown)


build.zig
```diff
@@ -37,6 +37,15 @@ pub fn build(b: *std.Build) void {
...
     const crypto_test_filter = b.option([]const u8, "crypto-test-filter", "Filter tests run by the test-crypto-provider-fuzz step");
     const crypto_test_filters: []const []const u8 = if (crypto_test_filter) |filter| &.{filter} else &.{};
+    // #494: unlike the crypto/QUIC provider-fuzz steps, this step's root
+    // module (tls_core_mod) is the *entire* pure-Zig TLS suite -- also run
+    // unfiltered by `test-tls`/`test` -- not a dedicated fuzz-only root
+    // file, so an unfiltered default here would just re-run the whole
+    // suite under a second name. Default to the "fuzz: " test-name prefix
+    // every #494 target above uses, scoping the step to just those targets;
+    // pass an explicit filter to select one target for a longer local run.
+    const tls_resumption_test_filter = b.option([]const u8, "tls-resumption-test-filter", "Filter tests run by the test-tls-resumption-fuzz step (default: \"fuzz: \", i.e. only the #494 session/PSK/ticket/resumption fuzz targets)") orelse "fuzz: ";
```
Jaruso: This default is not scoped to #494. `tls_core_mod` already contains unrelated tests whose names begin with `fuzz: ` (for example the two SNI fuzz targets and the existing ticket-key/ticket-identity targets), so `test-tls-resumption-fuzz` runs those too. The reported 14/14 count is consistent with seven new tests plus seven pre-existing tests, not “only the #494 targets” as the comment/docs claim.

Please give this program a unique namespace, for example:

```zig
const tls_resumption_test_filter = b.option(
    []const u8,
    "tls-resumption-test-filter",
    "Filter TLS resumption fuzz targets",
) orelse "fuzz: TLS resumption:";
```

and rename each new test accordingly, or use a focused test root that imports only the #494 targets. Otherwise long runs and reproduction commands can silently exercise a different fuzz target than this step promises.

src/tls/new_session_ticket.zig
```diff
@@ -949,3 +949,183 @@ fn appendExtension(list: *std.ArrayList(u8), ext_type: u16, payload: []const u8)
...
+    const round = try decode(encoded);
+    try std.testing.expectEqual(params.ticket_lifetime, round.ticket_lifetime);
+    try std.testing.expectEqual(params.ticket_age_add, round.ticket_age_add);
+    try std.testing.expectEqualSlices(u8, params.ticket_nonce, round.ticket_nonce);
+    try std.testing.expectEqualSlices(u8, params.ticket, round.ticket);
+    try std.testing.expectEqual(params.max_early_data_size, round.max_early_data_size);
+
+    // 3. Owned-state construction from the round-tripped `Parsed` value,
+    // exercising both client and server construction directions.
+    var rms: [session.max_psk_len]u8 = undefined;
```
Jaruso, line 1082: `rms` is always a 48-byte slice, but `KeySchedule.resumptionPsk` requires the RMS length to equal the selected hash digest exactly. Therefore both SHA-256 suites (`tls_aes_128_gcm_sha256` and `tls_chacha20_poly1305_sha256`) always return `InvalidSecretLength`, and the surrounding `catch return` silently skips owned-state construction for two of the three suites.

Size the logical slice from the selected suite:

```zig
const digest_len = algorithms.transcriptHash(suite).digestLength();
var rms: [session.max_psk_len]u8 = undefined;
smith.bytes(rms[0..digest_len]);

// ...
rms[0..digest_len],
```

This should be used in both client and server branches, with successful construction assertions for every suite rather than treating this harness error as an uninteresting rejection.

src/tls/session.zig
```diff
@@ -3565,3 +3565,255 @@ test "a failed CompatSnapshot.init replacement leaves the live destination compl
...
+/// Drives `decode` directly on Smith-mutated bytes and synthetic `Limits`:
+/// malformed TLV/header/length input must never panic, wrap, or escape the
+/// cursor, and any successful decode must own its state (checked via
+/// pointer disjointness from the buffer it was decoded from, not by
+/// snapshot/poison, since ownership is exactly non-aliasing).
+pub fn fuzzSessionRawDecode(allocator: std.mem.Allocator, smith: *std.testing.Smith) !void {
+    var limits_control: [1]u8 = undefined;
+    smith.bytes(&limits_control);
+    const limits = fuzzSessionLimits(limits_control[0]);
+    limits.validate() catch return;
```
Jaruso: This discards every invalid synthetic `Limits` value before the production decoder sees it. That means the target cannot verify the #494 requirement that zero/one-over/`0xffff` and other invalid caller limits fail deterministically with `InvalidLimits`; those cases simply become no-op fuzz inputs.

Generate the explicit boundary matrix and assert the decoder's typed result instead:

```zig
const result = decode(allocator, limits, raw);
if (limits.validate()) |_| {
    var decoded = result catch return;
    defer decoded.deinit();
    try expectDecodedOwnership(raw, &decoded);
} else |_| {
    try testing.expectError(error.InvalidLimits, result);
}
```

The generator also needs actual zero, one-under/exact/one-over, `0xffff`, and total-size boundary values rather than only default/hard-cap values.

src/tls/session.zig
```diff
@@ -3565,3 +3565,255 @@ test "a failed CompatSnapshot.init replacement leaves the live destination compl
...
+    const issued_at_unix_ms: i64 = @bitCast(std.mem.readInt(u64, &issued_bytes, .big));
+
+    var lifetime_bytes: [4]u8 = undefined;
+    smith.bytes(&lifetime_bytes);
+    const lifetime_seconds = 1 + std.mem.readInt(u32, &lifetime_bytes, .big) % max_lifetime_seconds;
+
+    var early_max_bytes: [4]u8 = undefined;
+    smith.bytes(&early_max_bytes);
+    const early_data: EarlyDataPolicy = if (smith.index(2) == 1)
+        .{ .early_data_capable = 1 + std.mem.readInt(u32, &early_max_bytes, .big) }
```
Jaruso: This can overflow when the Smith bytes decode to `0xffffffff`: `1 + maxInt(u32)` traps in safety-enabled builds, so the fuzz harness itself violates the target's no-panic property.

Map the arbitrary value into the non-zero `u32` domain without overflowing:

```zig
const raw = std.mem.readInt(u32, &early_max_bytes, .big);
const early_max = (raw % std.math.maxInt(u32)) + 1;
const early_data: EarlyDataPolicy = if (smith.index(2) == 1)
    .{ .early_data_capable = early_max }
else
    .resume_only;
```

Please add a deterministic seed/regression that drives `0xffffffff` through this branch.

src/tls/session.zig
```diff
@@ -3565,3 +3565,255 @@ test "a failed CompatSnapshot.init replacement leaves the live destination compl
...
+        // `common.deinit()` in a `catch` here would double-free it.
+        try state.init(allocator, Limits.default, &common, .{
+            .ticket = ticket_buf[0..ticket_len],
+            .ticket_age_add = std.mem.readInt(u32, &age_bytes, .big),
+            .ticket_nonce = nonce_buf[0..nonce_len],
+            .received_at_unix_ms = @bitCast(std.mem.readInt(u64, &received_bytes, .big)),
+        });
+        defer state.deinit();
+
+        const encoded = encodeClient(&state, Limits.default, encode_buf[0..encode_cap]) catch return;
```
Jaruso, line 3852: The test description says this proves a canonical record round trip, but after decoding it compares only ticket/nonce/suite/lifetime on the client path (and only suite/age-add on the server path). A decoder that drops or changes the PSK, SNI, ALPN, auth binding, timestamps, early-data policy, or any compatibility snapshot would still pass.

Use the exact-length APIs and re-encode the decoded record:

```zig
const needed = try clientEncodedLenWithLimits(&state, Limits.default);
try testing.expectEqual(needed, encoded.len);

var canonical_buf: [4096]u8 = undefined;
const canonical = try encodeClient(&decoded.client, Limits.default, &canonical_buf);
try testing.expectEqualSlices(u8, encoded, canonical);
```

Do the equivalent for the server path and add a helper that compares every common and record-specific field. That is the generated cross-operation property #494-A asks for; the existing fixed byte-stability fixture does not cover the Smith-generated combinations.

src/tls/pre_shared_key.zig
```diff
@@ -1395,3 +1395,189 @@ fn hexBytes(comptime hex: []const u8) [hex.len / 2]u8 {
...
+            .obfuscated_ticket_age = std.mem.readInt(u32, &age_bytes, .big),
+            .digest_len = digest_choices[smith.index(digest_choices.len)],
+        };
+    }
+    const items = items_buf[0..item_count];
+
+    var write_buf: [8192]u8 = undefined;
+    var w = messages.Writer{ .buf = &write_buf };
+    const offer = writeOffer(&w, items) catch {
+        try std.testing.expectEqual(@as(usize, 0), w.len);
```
Jaruso, line 1493: Checking only `w.len == 0` does not prove transactional output. A regression could write or overwrite bytes and then leave/reset the logical length to zero, and this target would pass even though #494 explicitly requires both writer length **and output** to remain unchanged.

Initialize and snapshot the storage before the call:

```zig
var write_buf: [8192]u8 = undefined;
@memset(&write_buf, 0xa5);
const before = write_buf;
var w = messages.Writer{ .buf = &write_buf };

const offer = writeOffer(&w, items) catch {
    try testing.expectEqual(@as(usize, 0), w.len);
    try testing.expectEqualSlices(u8, &before, &write_buf);
    return;
};
```

Please apply the same transaction property to `writeModes`, which is a required API but is not exercised by this fuzz target at all.

src/tls/pre_shared_key.zig
```diff
@@ -1395,3 +1395,189 @@ fn hexBytes(comptime hex: []const u8) [hex.len / 2]u8 {
...
+
+    var psk_buf: [provider.max_digest_len]u8 = undefined;
+    smith.bytes(&psk_buf);
+    var transcript_buf: [provider.max_digest_len]u8 = undefined;
+    smith.bytes(&transcript_buf);
+
+    const psk_len = switch (smith.index(3)) {
+        0 => digest_len,
+        1 => digest_len -| 1,
+        else => @min(digest_len + 1, psk_buf.len),
```
Jaruso: For SHA-384, `digest_len == provider.max_digest_len == 48`, so this supposed one-over branch becomes `min(49, 48) == 48` and exercises the valid length again. The same problem exists for `out_buf`; `transcript_buf` is always passed at the exact length, so wrong transcript-hash lengths are never tested either.

Use `provider.max_digest_len + 1` storage and choose lengths independently for PSK, transcript hash, output, and verification candidate:

```zig
var psk_buf: [provider.max_digest_len + 1]u8 = undefined;
var transcript_buf: [provider.max_digest_len + 1]u8 = undefined;
var out_buf: [provider.max_digest_len + 1]u8 = undefined;

const psk_len = chooseAround(digest_len, smith);
const transcript_len = chooseAround(digest_len, smith);
const out_len = chooseAround(digest_len, smith);
```

Then assert the exact typed behavior for every combination for both SHA-256 and SHA-384.

src/tls/ticket_protection.zig
```diff
@@ -3172,3 +3172,153 @@ test "key windows and ticket lifetime are enforced" {
...
+        1 => identity[4] +%= 1, // version
+        2 => identity[5] = @intCast(smith.index(256)), // AEAD id
+        3 => identity[6 + smith.index(2)] ^= 0xff, // reserved flags
+        4 => identity[8 + smith.index(key_id_len)] ^= 0xff, // key id
+        5 => identity[24 + smith.index(provider.aead_nonce_len)] ^= 0xff, // nonce
+        6 => identity[fixed_header_len + smith.index(ciphertext_len)] ^= 0xff, // ciphertext
+        else => identity[identity.len - 1 - smith.index(tag_len)] ^= 0xff, // tag
+    }
+
+    if (parseEnvelope(identity, session.Limits.default)) |mutated| {
```
Jaruso: For key ID, nonce, ciphertext, and tag, any one-byte mutation is still structurally valid and `parseEnvelope` must continue to succeed; those bytes are opaque until authenticated open. The current `else |_| {}` allows a future parser regression that rejects or interprets those fields to pass this target.

Keep the selected mutation kind and assert by class:

```zig
const mutation = smith.index(8);
// mutate...
if (mutation >= 4) {
    const parsed = try parseEnvelope(identity, session.Limits.default);
    try expectEnvelopeGeometry(identity, parsed);
} else {
    // Assert the expected typed structural result, accounting for an AEAD
    // mutation that deliberately selects a different valid stable ID.
}
```

Also ensure the AEAD mutation actually changes the byte; assigning an arbitrary value can reproduce the original ID and cease to be a mutation.

Jaruso (commented):

I reviewed all six changed files against the recommended **#494-A** slice and the mandatory #376 fuzzing contract. CI is green at `f1f9cf69d7d4e8b5d933539965d287a8098947c7`, and the overall decomposition/build-step direction is sound, but the current harness has two direct correctness defects and several required properties are either skipped or asserted too weakly.

The inline comments cover the concrete defects:

1. the `NewSessionTicket` owned-state target silently skips both SHA-256 suites because it always supplies a 48-byte RMS;
2. the generated session target can overflow while constructing an early-data policy;
3. invalid synthetic limits are discarded before `decode` sees them;
4. the generated session round trip does not actually prove canonical/full-state preservation;
5. transactional PSK rejection checks only writer length, not output bytes;
6. SHA-384 one-over and transcript-length binder cases are not generated;
7. opaque ticket-envelope mutations are allowed to fail structural parsing;
8. the new build step's default `"fuzz: "` filter also selects unrelated pre-existing TLS fuzz tests.

There are also acceptance gaps that are broader than one line:

- **`NewSessionTicket`:** the target does not directly assert `encodedLen(params) == encoded.len`; it never poisons/recycles the source before validating the owned state; generated validation failures are not exercised; the required fuzz-shaped bounded allocation-failure sweep is replaced with an existing fixed deterministic sweep; and the generated/corpus ticket sizes do not reach the configured/default and absolute wire maxima required by #494-A. `testCryptoProvider()` also retains static provider state instead of constructing fresh per-case state as required by the shared contract.
- **Session codec:** the target does not drive `clientEncodedLenWithLimits`, `serverEncodedLenWithLimits`, or `evaluateCompatibility`, and it does not add allocation-failure exploration for fuzz-generated records. The raw limit generator does not cover the issue's zero/one-under/exact/one-over/`0xffff` matrix.
- **PSK:** `writeModes` is not exercised; identity-derived binder mutation is absent; wrong transcript-hash lengths are absent; and the writer transaction check needs to cover bytes as well as `len`.
- **Ticket envelope:** the deterministic second parse only partially compares results, and the target should explicitly assert every successful view's geometry/content on both parses and systematically cover the required truncation boundaries.

Please tighten the harness so every required API/property is observable rather than relying on `catch return` paths or pre-existing fixed tests. Add focused deterministic corpus entries/regressions for each corrected boundary, then rerun the exact filtered non-`--fuzz` commands plus the full #494 validation matrix.

src/tls/session.zig
```diff
@@ -3565,3 +3565,335 @@ test "a failed CompatSnapshot.init replacement leaves the live destination compl
...
+        7 => .{ .max_ticket_len = 1 }, // exact minimum
+        8 => .{ .max_ticket_len = 0xfffe }, // one-under the 0xffff/absolute maximum
+        9 => .{ .max_ticket_len = 0xffff }, // exact maximum (absolute_ticket_wire_max == 0xffff)
+        10 => .{ .max_ticket_len = absolute_ticket_wire_max + 1 }, // invalid: one-over maximum
+        11 => .{ .max_serialized_len = 0 }, // invalid: zero
+        12 => .{ .max_serialized_len = header_len }, // near-minimum, still valid
+        13 => .{ .max_serialized_len = hard_max_serialized_len - 1 }, // one-under maximum
+        14 => .{ .max_serialized_len = hard_max_serialized_len }, // exact maximum
+        15 => .{ .max_serialized_len = hard_max_serialized_len + 1 }, // invalid: one-over maximum
+        16 => .{ .max_transport_compat_len = hard_max_compat_len + 1 }, // invalid: one-over maximum
```
Jaruso: The comment says this is a zero/one-under/exact/one-over matrix for **every** `Limits` field, but the compatibility fields only get the invalid one-over case. `max_transport_compat_len` and `max_application_compat_len` never receive zero, `hard_max_compat_len - 1`, or `hard_max_compat_len`; the fallback arithmetic cannot reach those values from an 8-bit control either. That leaves part of #494-A's explicit synthetic-limit requirement untested.

Please add the same boundary set for both fields, for example:

```zig
18 => .{ .max_transport_compat_len = 0 },
19 => .{ .max_transport_compat_len = hard_max_compat_len - 1 },
20 => .{ .max_transport_compat_len = hard_max_compat_len },
21 => .{ .max_transport_compat_len = hard_max_compat_len + 1 },
22 => .{ .max_application_compat_len = 0 },
23 => .{ .max_application_compat_len = hard_max_compat_len - 1 },
24 => .{ .max_application_compat_len = hard_max_compat_len },
25 => .{ .max_application_compat_len = hard_max_compat_len + 1 },
```

and expand the switch modulus accordingly. Zero is valid for these two fields, so it should reach `decode` as a meaningful tight limit rather than only testing invalid-limit rejection.

src/tls/session.zig
```diff
@@ -3565,3 +3565,335 @@ test "a failed CompatSnapshot.init replacement leaves the live destination compl
...
+/// Maps an arbitrary `u32` into the non-zero `u32` domain
+/// `EarlyDataPolicy.early_data_capable` requires, without `1 +
+/// std.math.maxInt(u32)` overflowing/trapping when `raw ==
+/// 0xffffffff` (safety-build panic; the fuzz harness itself would then
+/// violate the target's own no-panic property).
+fn earlyDataCapableFromRaw(raw: u32) u32 {
+    return (raw % std.math.maxInt(u32)) + 1;
+}
+
+test "fuzz: TLS resumption: earlyDataCapableFromRaw never overflows at the u32 boundary" {
```
Jaruso: This is a named deterministic regression, not a fuzz target: it never calls `std.testing.fuzz` and has no corpus. Naming it `"fuzz: TLS resumption: ..."` places it inside the default fuzz namespace and contradicts both #376's target convention and the documentation added in this PR, which says named deterministic regressions are **not fuzz-tagged** and are selected by their own name.

Rename it to something like:

```zig
test "earlyDataCapableFromRaw never overflows at the u32 boundary" {
```

Then document/run its exact non-`--fuzz` filter just like the `parseEnvelope` truncation regression. The default `test-tls-resumption-fuzz` namespace should contain only actual `std.testing.fuzz` targets.

src/tls/pre_shared_key.zig
```diff
@@ -1395,3 +1395,226 @@ fn hexBytes(comptime hex: []const u8) [hex.len / 2]u8 {
...
+        try std.testing.expectEqualSlices(u8, out_buf[0..digest_len], out2[0..digest_len]);
+
+        try std.testing.expect(try verifyBinderFromTranscriptHash(hash, psk_buf[0..digest_len], transcript_buf[0..digest_len], out_buf[0..digest_len]));
+
+        var mutated_binder = out_buf;
+        var mutated_psk = psk_buf;
+        var mutated_transcript = transcript_buf;
+        const bit_index = smith.index(digest_len * 8);
+        switch (smith.index(3)) {
+            0 => mutated_binder[bit_index / 8] ^= @as(u8, 1) << @intCast(bit_index % 8),
```
Jaruso, line 1579: This mutation switch covers binder, PSK, and the already-computed transcript hash, but #494-A explicitly requires one-bit **identity** mutations not to verify as well. Mutating the digest is useful, but it does not prove the identity bytes are actually included at the correct truncated-ClientHello boundary.

Add a small generated truncated ClientHello/prefix containing a distinct identity region, derive its baseline binder, flip one identity bit, and verify the old binder fails for both hashes. For example:

```zig
var prefix: [128]u8 = undefined;
const identity = prefix[identity_offset .. identity_offset + identity_len];
smith.bytes(prefix[0..prefix_len]);

var binder: [provider.max_digest_len]u8 = undefined;
try deriveBinder(hash, psk_buf[0..digest_len], prefix[0..prefix_len], binder[0..digest_len]);

identity[smith.index(identity.len)] ^= @as(u8, 1) << @intCast(smith.index(8));
try testing.expect(!try verifyBinder(
    hash,
    psk_buf[0..digest_len],
    prefix[0..prefix_len],
    binder[0..digest_len],
));
```

Using `deriveBinder`/`verifyBinder` here (or explicitly recomputing the two transcript hashes before calling the `FromTranscriptHash` APIs) makes the missing identity property observable rather than treating it as indistinguishable from an arbitrary digest mutation.

Jaruso (commented):
## Re-review result: three changes still required

I compared `386dd697f0cb1026adc3f90ddc49d0609fe390e4` directly against the previously reviewed head, re-read all six current file patches, rechecked the #494-A acceptance criteria, and verified the new CI run is green. All eight original inline findings are correctly addressed, and I resolved those threads.

Three issue-level requirements remain incomplete:

1. `fuzzSessionLimits` still does not provide the boundary matrix it claims for the two compatibility-limit fields;
2. the new early-data overflow regression is a deterministic test but is named inside the fuzz-target namespace, contrary to the #376 target/regression contract and this PR's own documentation;
3. the PSK binder target still omits the explicitly required one-bit **identity** mutation property.

The inline comments include concrete fixes. After those are addressed, update the PR description as well: it still says the default filter is bare `"fuzz: "`, reports 14/14 and the old full-suite counts, and describes transactional checking as writer-length-only. Those statements are stale after the follow-up commit.

No production TLS behavior is changed here, and I did not find a regression in the corrected ticket ownership, canonical session round-trip, writer transaction, binder-length, envelope-mutation, or namespaced build-step logic.

Jaruso (commented):
## Re-review result: approved — no blocking findings

I compared `07d8666779d4fcb2ddc0f65f83ad5e3fe590ae94` against the previously reviewed head, inspected all three follow-up file changes, rechecked the complete current PR against the #494-A slice and #376 fuzzing contract, and verified CI run `30929059895` completed successfully.

The three remaining findings are correctly addressed:

- `fuzzSessionLimits` now covers zero, one-under, exact, and one-over boundaries for both compatibility-limit fields, including the valid zero-limit path;
- the early-data overflow regression is outside the `"fuzz: TLS resumption:"` namespace and has a documented exact-filter command;
- the binder target now derives and verifies through the ordinary transcript-hashing APIs, establishes a valid baseline for both hashes, flips one bit inside a bounded identity region within the truncated prefix, and proves the original binder no longer verifies.

The PR description and validation counts are also current. I resolved all three remaining review threads. I found no new correctness, ownership, boundary, transactional-output, or target-scoping regressions in the follow-up commit.

From a code-review perspective, #574 is ready to merge for the recommended #494-A scope. GitHub records this as `COMMENTED` rather than `APPROVED` because the authenticated account is also the PR author.

PR #576, opened 2026-08-05
security: add TLS resumption cache fuzz targets

Jaruso:
## Summary

Implements #494-C for the client session cache and stateful-server cache/resolver operation models.

- Adds two independent `fuzz: TLS resumption:` Smith operation-sequence targets in `src/tls/session_cache.zig` with the exact stable names from the issue.
- Adds full logical state snapshots with domain-separated fingerprints, per-op client/server invariant checks, bounded operation-scoped OOM sweeps, deterministic clocks, and permanent Smith corpus programs.
- Adds test-only zeroization/destruction probes for `destroyClientEntry`, `destroyServerEntry`, and `PublicServerLeaseBox.deinit`.
- Adds the required named deterministic regression for `resolveStatefulServerPsk` lease-box OOM releasing an acquired single-use pin and permitting retry.
- Updates `docs/CRYPTO_FUZZ_CONTRACT.md` with the new target and regression reproduction commands.

## #494-C checklist mapping

- Two independent targets: `fuzzClientCacheSequenceInput` and `fuzzServerCacheSequenceInput`.
- Explicit bounds: `fuzz_cache_limits`, `fuzz_max_operations`, lease slot counts, persisted record cap, OOM fail-index cap, and fixed clock step table.
- Fresh per-case state: each fuzz callback constructs new cache, deterministic clock, random source, lease slots, persistence token state, and snapshots.
- Snapshot coverage: `ClientCacheState` and `ServerCacheState` include counters, generations, persistence tokens, entry metadata, handle/index membership, and secret-safe fingerprints.
- Per-op invariants: `expectClientCacheInvariants` and `expectServerCacheInvariants` run after every opcode path.
- Ownership/transactional assertions: store/insert caller fingerprints, OOM snapshot comparisons, cloned offer/state paths, persistence restore atomicity helpers.
- Permanent corpus: client and server target corpora encode selected/not-selected/abort/drop, reusable/single-use, persistence, OOM, renumbering, handle shape/collision/entropy, adapter OOM, restore, and teardown paths.
- Zeroization/finalization: test-only destruction probes and named probe regression count exact entry/lease-box destruction.
- Out-of-scope policy: target stays in `session_cache.zig`; no ClientHello parsing, backend selection policy, transport parsing, or runtime composition.

## Validation

- `zig build test-tls-resumption-fuzz --summary all --error-style verbose` — 13/13 passed
- `zig build test-tls-resumption-fuzz -Dtls-resumption-test-filter="fuzz: TLS resumption: client session cache operation sequence preserves transactional state, ownership, eviction, and lease semantics" --summary all --error-style verbose` — 3/3 passed
- `zig build test-tls-resumption-fuzz -Dtls-resumption-test-filter="fuzz: TLS resumption: stateful server cache operation sequence preserves transactional indexes, ownership, handle, and lease semantics" --summary all --error-style verbose` — 3/3 passed
- `zig build test-tls-resumption-fuzz -Dtls-resumption-test-filter="resolveStatefulServerPsk lease-box OOM releases an acquired single-use pin and permits retry" --summary all --error-style verbose` — 3/3 passed
- `zig build test-tls --summary all --error-style verbose` — 892/892 passed
- `zig build test --summary all --error-style verbose` — 3165/3166 passed, 1 skipped
- `zig build test -Dtls-profile=appliance --summary all --error-style verbose` — 3134/3135 passed, 1 skipped

Bounded ReleaseFast fuzz smoke:

- `zig build test-tls-resumption-fuzz -Doptimize=ReleaseFast --fuzz=1000 -Dtls-resumption-test-filter="fuzz: TLS resumption: client session cache operation sequence preserves transactional state, ownership, eviction, and lease semantics" --summary all --error-style verbose` — completed 1207 runs
- `zig build test-tls-resumption-fuzz -Doptimize=ReleaseFast --fuzz=1000 -Dtls-resumption-test-filter="fuzz: TLS resumption: stateful server cache operation sequence preserves transactional indexes, ownership, handle, and lease semantics" --summary all --error-style verbose` — completed 1191 runs

Closes #494 after merge of the full issue sequence; this PR implements slice C only.

### DIFF
--- a/docs/CRYPTO_FUZZ_CONTRACT.md
+++ b/docs/CRYPTO_FUZZ_CONTRACT.md
@@ -415,15 +415,8 @@ free on every path -- exercised across all three required AEADs, each fuzz
 case constructing its own fresh `pure_zig.DeterministicEntropy`-backed
 provider rather than sharing static state. The rejection matrix includes
 public envelope-field mutation, authenticated plaintext mutation, key state,
-and lifetime edges. #494-C adds two independent bounded operation-sequence
-targets in `session_cache.zig`: client cache store/lookup/lease/persistence
-state and stateful-server insert/resolve/public-adapter/persistence state.
-Both targets use tiny fixed cache limits, at most 16 opcodes per case,
-deterministic logical clocks, explicit Smith corpus programs, full
-fingerprint snapshots for transactional comparisons, per-op invariant
-checks, bounded operation-scoped `FailingAllocator` sweeps, and test-only
-zeroization/destruction probes for entries and public lease boxes.
-several classes that require successfully authenticating first:
+and lifetime edges, plus several classes that require successfully
+authenticating first:
 `not_yet_valid`/`expired` (forged timestamps); an authenticated state
 extended with an unknown *optional* field, which must still accept and
 round-trip (proving forward compatibility, not just rejection); a
@@ -487,5 +480,14 @@ in the opcode loop, without a redundant assertion. The existing
 `ticket_key_snapshot.zig` persistent-JSON-snapshot fuzz target is unrelated
 (on-disk snapshot file parsing, not the in-memory keyring) and is preserved
 as-is, per the issue's "existing coverage to preserve, not recreate" list.
-Session-cache/lease state machines and backend/runtime composition follow
-in #494-C/D per the issue's PR decomposition.
+#494-C adds two independent bounded operation-sequence targets in
+`session_cache.zig`: client cache store/lookup/lease/persistence state and
+stateful-server insert/resolve/public-adapter/persistence state. Both
+targets use tiny fixed cache limits, at most 16 opcodes per case,
+deterministic logical clocks, explicit Smith corpus programs, full
+fingerprint snapshots for transactional comparisons, transition-specific
+lease oracles, per-op invariant checks, bounded operation-scoped
+`FailingAllocator` sweeps, controlled collision programs, and test-only
+zeroization/destruction probes for entries and public lease boxes.
+Backend/runtime composition follows in #494-D per the issue's PR
+decomposition.
--- a/docs/CRYPTO_FUZZ_CONTRACT.md
+++ b/docs/CRYPTO_FUZZ_CONTRACT.md
@@ -357,9 +357,10 @@ have to go hunting through `build.zig`.
 Unlike this file's own targets, #494's targets are inline `test "fuzz:
 ..."` blocks inside the production modules themselves
 (`src/tls/new_session_ticket.zig`, `src/tls/session.zig`,
-`src/tls/pre_shared_key.zig`, `src/tls/ticket_protection.zig`) — not a
-separate `tests/*.zig` root — so they already replay their deterministic
-seed corpus under plain `zig build test-tls` / `zig build test`. The
+`src/tls/pre_shared_key.zig`, `src/tls/ticket_protection.zig`,
+`src/tls/session_cache.zig`) — not a separate `tests/*.zig` root — so they
+already replay their deterministic seed corpus under plain
+`zig build test-tls` / `zig build test`. The
 `test-tls-resumption-fuzz` step exists to give them a stable,
 individually filterable/long-runnable name, the same shape as
 `-Dcrypto-test-filter` above:
@@ -387,11 +388,14 @@ zig build test-tls-resumption-fuzz -Doptimize=ReleaseFast -Dtls-resumption-test-
 zig build test-tls-resumption-fuzz -Doptimize=ReleaseFast -Dtls-resumption-test-filter="fuzz: TLS resumption: parseEnvelope single-field mutation" --fuzz=10M --summary all --error-style verbose
 zig build test-tls-resumption-fuzz -Doptimize=ReleaseFast -Dtls-resumption-test-filter="fuzz: TLS resumption: Protector.resolve authenticates" --fuzz=10M --summary all --error-style verbose
 zig build test-tls-resumption-fuzz -Doptimize=ReleaseFast -Dtls-resumption-test-filter="fuzz: TLS resumption: ticket keyring install/validate/acquire/retain/release/seal sequence" --fuzz=10M --summary all --error-style verbose
+zig build test-tls-resumption-fuzz -Doptimize=ReleaseFast -Dtls-resumption-test-filter="fuzz: TLS resumption: client session cache operation sequence preserves transactional state, ownership, eviction, and lease semantics" --fuzz=10M --summary all --error-style verbose
+zig build test-tls-resumption-fuzz -Doptimize=ReleaseFast -Dtls-resumption-test-filter="fuzz: TLS resumption: stateful server cache operation sequence preserves transactional indexes, ownership, handle, and lease semantics" --fuzz=10M --summary all --error-style verbose
 
 # Named deterministic regressions (not fuzz-tagged, so outside the
 # default "fuzz: TLS resumption:" namespace — filter on their own name):
 zig build test-tls-resumption-fuzz -Dtls-resumption-test-filter="parseEnvelope rejects every truncated prefix of a valid envelope below the minimum length" --summary all --error-style verbose
 zig build test-tls-resumption-fuzz -Dtls-resumption-test-filter="earlyDataCapableFromRaw never overflows at the u32 boundary" --summary all --error-style verbose
+zig build test-tls-resumption-fuzz -Dtls-resumption-test-filter="resolveStatefulServerPsk lease-box OOM releases an acquired single-use pin and permits retry" --summary all --error-style verbose
 ```
 
 `-Dtls-resumption-test-filter` defaults to `"fuzz: TLS resumption:"` (every
@@ -410,6 +414,15 @@ allocation-failure sweep over every reachable allocation point, all under
 free on every path -- exercised across all three required AEADs, each fuzz
 case constructing its own fresh `pure_zig.DeterministicEntropy`-backed
 provider rather than sharing static state. The rejection matrix includes
+public envelope-field mutation, authenticated plaintext mutation, key state,
+and lifetime edges. #494-C adds two independent bounded operation-sequence
+targets in `session_cache.zig`: client cache store/lookup/lease/persistence
+state and stateful-server insert/resolve/public-adapter/persistence state.
+Both targets use tiny fixed cache limits, at most 16 opcodes per case,
+deterministic logical clocks, explicit Smith corpus programs, full
+fingerprint snapshots for transactional comparisons, per-op invariant
+checks, bounded operation-scoped `FailingAllocator` sweeps, and test-only
+zeroization/destruction probes for entries and public lease boxes.
 several classes that require successfully authenticating first:
 `not_yet_valid`/`expired` (forged timestamps); an authenticated state
 extended with an unknown *optional* field, which must still accept and
--- a/src/tls/session_cache.zig
+++ b/src/tls/session_cache.zig
@@ -4374,8 +4374,161 @@ fn fuzzServerState(allocator: std.mem.Allocator, id: u64) !session.ServerRecover
     return makeServer(allocator, sni);
 }
 
+fn fuzzServerStateWithCompat(allocator: std.mem.Allocator, id: u64) !session.ServerRecoverableState {
+    var sni_buf: [32]u8 = undefined;
+    const sni = try std.fmt.bufPrint(&sni_buf, "sc{d}.example.test", .{id % fuzz_cache_limits.max_origins});
+    var params = commonParams(&([_]u8{0xcd} ** 32), sni);
+    params.transport_compat = .{ .format_id = 7, .format_version = 1, .bytes = "clone-alloc-compat" };
+    var common: session.ResumableSessionCommon = .{};
+    try common.init(allocator, session.Limits.default, params);
+    var state: session.ServerRecoverableState = .{};
+    state.init(&common, 7);
+    return state;
+}
+
+const ClientLeaseTransition = struct {
+    len: usize = 0,
+    present: [pre_shared_key.max_offered_identities]bool = [_]bool{false} ** pre_shared_key.max_offered_identities,
+    entry_ids: [pre_shared_key.max_offered_identities]u64 = [_]u64{0} ** pre_shared_key.max_offered_identities,
+    single_use: [pre_shared_key.max_offered_identities]bool = [_]bool{false} ** pre_shared_key.max_offered_identities,
+    fingerprints: [pre_shared_key.max_offered_identities][32]u8 = undefined,
+};
+
+fn captureClientLeaseTransition(cache: *ClientSessionCache, lease: *pre_shared_key.ClientOfferLease) !ClientLeaseTransition {
+    var out: ClientLeaseTransition = .{ .len = lease.offers.len };
+    cache.mutex.lock();
+    defer cache.mutex.unlock();
+    for (0..lease.offers.len) |i| {
+        const token = lease.tokens[i];
+        out.entry_ids[i] = token.entry_id;
+        out.single_use[i] = token.single_use;
+        const idx = cache.findIndexByEntryId(token.entry_id) orelse {
+            try testing.expect(!token.single_use);
+            continue;
+        };
+        const entry = cache.entries.items[idx];
+        out.present[i] = true;
+        if (token.single_use) {
+            try testing.expectEqual(token.lease_epoch, entry.active_lease_epoch.?);
+        }
+        out.fingerprints[i] = try fingerprintClientTicket(&entry.ticket);
+    }
+    return out;
+}
+
+fn expectClientEntryState(cache: *ClientSessionCache, entry_id: u64, expected_fp: [32]u8, should_exist: bool, should_be_pinned: bool) !void {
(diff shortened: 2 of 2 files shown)


Jaruso (commented):
Changes requested.

This is a strong, well-scoped implementation of the #494-C architecture: the two independent Smith targets are bounded and deterministic, the state snapshots are unusually thorough, the test-only destruction probes are appropriately secret-safe, and CI is green. I found several blocking gaps in the fuzz *oracles*, however. In each case the operation can complete with a structurally valid cache while violating the exact lease/consumption behavior #494-C requires, so the generic post-op invariants cannot detect the regression.

The blocking items are:

1. Add transition-specific assertions for client terminal outcomes (`selected`, `not_selected`, abort/deinit, and `dropOffer`) so the target proves exactly which entry was consumed or preserved and that every other pin was released.
2. Add the equivalent server assertions and make the public-adapter path actually exercise the owned `pre_shared_key.ServerPskLease` commit/release/deinit callbacks. The current public-adapter opcode always deinitializes single-use results, so `PublicServerLeaseBox.commit` is unreachable from the fuzz target.
3. Make each bounded `FailingAllocator` sweep assert that at least one failure was induced and that an eventual exact success was reached within the configured cap. The current `doNotOptimizeAway` calls allow an incomplete or ineffective sweep to pass silently.
4. Make the controlled collision programs require their exact advertised outcomes instead of accepting the same broad result set for both retry-success and retry-exhaustion.
5. Repair the documentation paragraph insertion, which currently splits the #494-B rejection-matrix sentence.

I left inline examples for the expected assertions. After those changes, the target should provide the semantic coverage promised by the issue rather than only structural consistency coverage.

src/tls/session_cache.zig
```diff
@@ -3955,3 +4297,804 @@ test "a removed client entry's inline secrets are wiped from allocator backing m
...
+                        result.deinit();
+                    }
+                } else {
+                    result.deinit();
+                }
+            },
+            .finish_selected => for (&leases) |*lease| {
+                if (lease.* == .hit) {
+                    const selected = if (lease.hit.offers.len == 0) 0 else smith.index(lease.hit.offers.len + 1);
+                    lease.hit.finish(.{ .selected = selected });
```
Jaruso: [P1] Please add a transition oracle here rather than relying only on `expectClientCacheInvariants` after the operation. A bug that consumes every pinned ticket, consumes the wrong selected ticket, or consumes on `.not_selected`/deinit can still leave a perfectly valid empty or smaller cache, so the current target would pass despite violating the #494-C acceptance criteria.

Before terminating a live offer, capture the offered ticket identities/fingerprints and their cache membership/usage. Then assert the exact postcondition:

```zig
const before = try captureClientLeaseTransition(&cache, &lease.hit);
lease.hit.finish(outcome);
lease.deinit();
try expectClientLeaseTransition(&cache, before, outcome);
```

For a valid selected index, exactly the selected `.single_use` entry should disappear; a selected reusable entry and every unselected entry should remain. For an out-of-range selection, `.not_selected`, and abort/deinit, all entries should remain and all pins should be cleared. `dropOffer` should similarly prove that it releases only the dropped offer's pin without consuming its entry. Structural count/index invariants alone cannot establish any of those semantics.

src/tls/session_cache.zig
```diff
@@ -3955,3 +4297,804 @@ test "a removed client entry's inline secrets are wiped from allocator backing m
...
+            .resolve_through_public_adapter => if (handle_count > 0) {
+                var result = try resolveStatefulServerPsk(&cache, testing.allocator, &handles[smith.index(handle_count)], now_ms);
+                if (result == .hit) {
+                    if (result.hit.on_selected) |hook| hook.complete();
+                }
+                result.deinit();
+            },
+            .commit_lease => for (&leases) |*lease| {
+                if (lease.* == .hit) {
+                    lease.hit.lease.commit();
```
Jaruso, line 5219: [P1] The server terminal operations need identity-specific postconditions, and the public adapter must exercise its owned lease callbacks. These internal `commit`/`release`/deinit branches currently perform no before/after membership check, so consuming the wrong handle, consuming on release, or failing to consume on commit can all remain structurally valid and escape the fuzz oracle.

There is also no fuzz path that calls `result.hit.lease.commit()` on the `pre_shared_key.ServerPskResolveResult` returned by `resolveStatefulServerPsk`: `resolve_through_public_adapter` only completes the reusable `on_selected` hook and then deinitializes the result. Consequently `PublicServerLeaseBox.commit` is unreachable, despite the PR claiming public-adapter lease coverage.

Please snapshot the resolved handle/entry before each terminal action and assert:

```zig
// internal ServerLease
commit(single_use) => exact entry removed
commit(reusable)   => entry retained and recency advanced
release/deinit     => entry retained and pin cleared

// public pre_shared_key.ServerPskLease
commit/release/deinit => same membership semantics,
                         plus exactly one lease-box destruction
```

A Smith choice can select commit vs release vs implicit deinit for the public result. That will cover the ownership-transfer box itself, not just the cache's internal lease.

src/tls/session_cache.zig
```diff
@@ -3955,3 +4297,804 @@ test "a removed client entry's inline secrets are wiped from allocator backing m
...
+            var after = try ClientCacheState.capture(&cache);
+            defer after.deinit();
+            try before.expectEqual(&after);
+        } else {
+            try testing.expect(result == .stored or result == .replaced);
+            saw_success = true;
+            break;
+        }
+    }
+    std.mem.doNotOptimizeAway(saw_success);
```
Jaruso: [P1] `doNotOptimizeAway` does not prove that this bounded OOM sweep covered the path. If the operation stops producing allocation failures, or grows beyond `fuzz_max_oom_fail_index`, both an ineffective sweep (`saw_failure == false`) and a truncated sweep (`saw_success == false`) currently pass silently. That directly undermines the PR's “every reachable allocation point” claim.

Please make the sweep fail unless it observed both sides:

```zig
try testing.expect(saw_failure);
try testing.expect(saw_success);
```

Apply the same rule to every client/server OOM helper; `serverResolveOomSweep` also needs a `saw_failure` flag. The first non-failing index should assert the exact expected successful outcome for the fixture (`.stored` or `.hit`, not a broad alternative such as `.miss`, `.replaced`, or `.rejected_capacity`) before breaking. Then the loop's consecutive fail indices actually demonstrate complete coverage up to the first successful run.

src/tls/session_cache.zig
```diff
@@ -3955,3 +4297,804 @@ test "a removed client entry's inline secrets are wiped from allocator backing m
...
+                var a = try fuzzServerState(testing.allocator, 44);
+                var ha: [stateful_identity_len]u8 = undefined;
+                _ = cache.insertMove(&a, now_ms, .reusable, &ha);
+                a.deinit();
+                var b = try fuzzServerState(testing.allocator, 45);
+                defer b.deinit();
+                var hb: [stateful_identity_len]u8 = [_]u8{0xEE} ** stateful_identity_len;
+                const result = cache.insertMove(&b, now_ms, .reusable, &hb);
+                if (op == .handle_collision_then_success) {
+                    if (!(result == .stored or result == .rejected_handle_generation_failed or result == .rejected_capacity or result == .storage_failed))
```
Jaruso: [P1] These two controlled-collision branches currently accept the same broad result set, so neither corpus program proves the contract named by its opcode. On a fresh fixture, a regression that never retries would let `handle_collision_then_success` return `.rejected_handle_generation_failed` and pass; a regression that fails to detect exhaustion would let `handle_collision_exhaustion` return `.stored` and pass. Ignoring the result of the first insertion also means the collision precondition is not guaranteed in an arbitrary live cache.

Please isolate this in a fresh-cache helper (or otherwise require the setup insertion to be `.stored`) and assert exact outcomes:

```zig
try testing.expectEqual(.stored, first_result);
if (expect_retry_success) {
    try testing.expectEqual(.stored, second_result);
    try testing.expect(!std.mem.eql(u8, &ha, &hb));
} else {
    try testing.expectEqual(.rejected_handle_generation_failed, second_result);
    try testing.expectEqualSlices(u8, &sentinel, &hb);
    try before.expectEqual(&after);
    try testing.expectEqualDeep(caller_fingerprint, try fingerprintServerState(&b));
}
```

That makes the permanent seeds catch retry-count, collision-detection, output-atomicity, and caller-ownership regressions.

docs/CRYPTO_FUZZ_CONTRACT.md
```diff
@@ -410,6 +414,15 @@ allocation-failure sweep over every reachable allocation point, all under
...
 provider rather than sharing static state. The rejection matrix includes
+public envelope-field mutation, authenticated plaintext mutation, key state,
+and lifetime edges. #494-C adds two independent bounded operation-sequence
+targets in `session_cache.zig`: client cache store/lookup/lease/persistence
+state and stateful-server insert/resolve/public-adapter/persistence state.
+Both targets use tiny fixed cache limits, at most 16 opcodes per case,
+deterministic logical clocks, explicit Smith corpus programs, full
+fingerprint snapshots for transactional comparisons, per-op invariant
+checks, bounded operation-scoped `FailingAllocator` sweeps, and test-only
+zeroization/destruction probes for entries and public lease boxes.
```
Jaruso: [P2] This insertion splits the existing #494-B sentence. The paragraph now ends with “zeroization/destruction probes...” and immediately continues with the fragment “several classes that require successfully authenticating first:”. Please move the #494-C paragraph after the complete #494-B rejection-matrix discussion (or restore the missing transition) so the contract remains readable and unambiguous.

src/tls/session_cache.zig
```diff
@@ -3955,3 +4297,1025 @@ test "a removed client entry's inline secrets are wiped from allocator backing m
...
+                cache.deinit();
+                cache = try ClientSessionCache.init(testing.allocator, fuzz_cache_limits);
+            },
+            .client_oom_sweep => switch (smith.index(3)) {
+                0 => try clientStoreOomSweep(),
+                1 => try clientLookupOomSweep(),
+                else => try clientPersistenceOomSweep(),
+            },
+        }
+        try expectClientCacheInvariants(&cache);
```
Jaruso, line 4869: [P1] `expectClientCacheInvariants` still does not reconcile the fuzz model's live `ClientOfferLease`s with the cache's active pins. A regression that evicts a pinned entry during store/cleanup, leaves an orphan pin after a failed/incompatible lookup, or permits `beginPersistenceOperation` while a single-use offer is live can leave a perfectly valid-looking cache and pass this check. `restore_clones`, `quiescent_reset`, and final teardown then call `finishAllClientLeases` and destroy/replace the state before the mismatch is observed.

Please add a post-op bijection oracle, for example:

```zig
try expectClientLeaseModel(&cache, &leases, persistence_token);
```

It should prove both directions: every live single-use token resolves to the same `cache_generation`/`entry_id`/`lease_epoch` and remains pinned, and every `active_lease_epoch` belongs to exactly one live token. It should also require `cache.persistence_epoch == 0` whenever any pin exists (and token presence to agree with the tracked persistence token). This catches both forbidden directions: beginning persistence with an outstanding lease and acquiring a pin while persistence is active.

The normal bulk cleanup paths also need transition assertions. Change `finishAllClientLeases` (or add an asserting variant) to capture each hit, deinit it as `.aborted`, and run `expectClientLeaseTransition` before restore/reset/final teardown. Otherwise a bug where implicit deinit consumes a single-use ticket is hidden immediately by the subsequent cache replacement/destruction.

Jaruso (commented):
Follow-up review of `01000d30dc258609aeb9e7deb30bac44f0c389ce`.

The prior five findings are addressed correctly: the terminal client/server transition oracles now assert exact membership behavior, public adapter owned-lease callbacks are reachable and counted, every OOM sweep requires both induced failure and eventual exact success, collision retry/exhaustion fixtures are isolated with exact outcomes, and the fuzz-contract paragraph is repaired. CI is green on this head.

Two P1 gaps remain against #494-C's cross-operation requirements. The per-op checks still validate only cache-internal structure, not the bijection between the state machine's live lease objects and the cache's active pins. They also do not assert the required mutual exclusion between persistence operations and single-use pins, and the bulk `finishAll*Leases` paths erase evidence before restore/reset/teardown. A regression can therefore evict a pinned entry, create an orphan pin, allow persistence to overlap a lease, or consume on implicit deinit while leaving a structurally valid cache.

Please add live-lease/pin reconciliation after every opcode and transition-checked bulk release before restore/reset/final teardown on both caches. Once those are covered, I do not see another blocker in this slice.

src/tls/session_cache.zig
```diff
@@ -3955,3 +4297,1025 @@ test "a removed client entry's inline secrets are wiped from allocator backing m
...
+                handle_count = 0;
+            },
+            .server_oom_sweep => switch (smith.index(4)) {
+                0 => try serverInsertOomSweep(),
+                1 => try serverResolveOomSweep(),
+                2 => try serverPersistenceOomSweep(),
+                else => try publicLeaseBoxOomReleasesPin(),
+            },
+        }
+        expectServerCacheInvariants(&cache) catch return error.ServerFuzzCacheInvariant;
```
Jaruso, line 5350: [P1] The server side has the same remaining cross-operation hole. `expectServerCacheInvariants` validates indexes/accounting, but it does not prove that every live single-use `ServerLease` still owns the matching active pin, that every active pin has exactly one live lease, or that `persistence_epoch != 0` implies zero active pins. Therefore a broken insert/cleanup/resolve/begin-persistence path can evict a leased entry, orphan a pin, or overlap persistence with a lease and still pass until `finishAllServerLeases`/reset erases the evidence.

Add an `expectServerLeaseModel(&cache, &leases, persistence_token)` after every opcode that checks `cache_generation`, `entry_id`, `lease_epoch`, single-use policy, exact pin ownership, and persistence mutual exclusion in both directions. Also make the normal `finishAllServerLeases` calls before restore/reset/final teardown capture each live lease and assert the `.release` transition; keep a non-asserting cleanup-only variant for `errdefer` if necessary.

This is required to establish the #494-C properties that leased entries are never evicted and persistence-busy paths cannot resurrect consumed single-use state; terminal-action oracles alone only catch the subset of cases that later select an explicit commit/release/deinit opcode.

Jaruso (commented):
The two prior P1 findings are fixed: the client/server fuzz models now reconcile live leases bidirectionally with cache pins after every opcode, enforce persistence mutual exclusion, and use checked bulk release/abort transitions before restore, reset, and teardown. CI is green on this head.

One P1 harness issue remains. The `end_persistence_wrong_or_stale` opcode consumes an arbitrary Smith `u64` without ensuring it differs from the currently tracked token. A valid input can therefore end the real persistence operation with the correct token while leaving `persistence_token` populated in the model, causing the new reconciliation oracle to report a false production failure. The same issue exists in the server program. Force the generated token to differ from the active token (and add a permanent seed for active token `1`) or explicitly update the model when the generated token matches.

src/tls/session_cache.zig
```diff
@@ -3955,3 +4297,1141 @@ test "a removed client entry's inline secrets are wiped from allocator backing m
...
+            .begin_persistence => if (persistence_token == null) {
+                persistence_token = cache.beginPersistenceOperation() catch null;
+            } else {
+                try testing.expectError(error.CacheBusy, cache.beginPersistenceOperation());
+            },
+            .end_persistence_correct => if (persistence_token) |token| {
+                cache.endPersistenceOperation(token);
+                persistence_token = null;
+            },
+            .end_persistence_wrong_or_stale => cache.endPersistenceOperation(smith.value(u64)),
```
Jaruso: [P1] This “wrong or stale” token is still an arbitrary Smith value, so it can equal the active token. For example, the deterministic program `begin_persistence` followed by this opcode with value `1` correctly clears `cache.persistence_epoch`, but the fuzz model leaves `persistence_token == 1`; the immediately following `expectClientLeaseModel` then fails on correct production behavior. This makes a valid fuzz input a false positive (and the server branch has the same issue).

Please either force the generated value to differ from the active token or update the tracked model when it matches. Since this opcode is specifically intended to exercise stale/wrong-token no-ops, forcing a mismatch is clearer:

```zig
fn wrongPersistenceToken(smith: *std.testing.Smith, active: ?u64) u64 {
    var candidate = smith.value(u64);
    if (active) |token| {
        if (candidate == token) candidate +%= 1;
    }
    return candidate;
}

.end_persistence_wrong_or_stale =>
    cache.endPersistenceOperation(wrongPersistenceToken(smith, persistence_token)),
```

Apply this to both client and server programs and add permanent seeds equivalent to `{ begin_persistence, end_persistence_wrong_or_stale, 1 }` so the accidental-match case cannot regress.

Jaruso (commented):
## Re-review summary

No remaining findings.

The final persistence-token issue is fixed correctly for both client and server fuzz programs: `wrongPersistenceToken` guarantees the stale/wrong operand cannot equal the active token, including wraparound, and the new permanent corpus seeds cover the accidental-match case that previously produced a false positive.

I rechecked the complete #494-C acceptance surface from the prior reviews: exact client and server lease transitions, public owned-lease commit/release/deinit paths, per-operation lease/pin bijections, persistence mutual exclusion, checked bulk teardown, bounded OOM completeness, controlled collision outcomes, zeroization/destruction probes, and documentation placement. The previous review threads are resolved, the PR is mergeable, and CI is green on `5f20795123db17ab3d6ee5e174c9a404481394bc`.

Because the connected GitHub account is also the PR author, this is submitted as a comment review rather than a formal approval.

PR #577, opened 2026-08-06
Add shared TLS protocol fuzz targets

Jaruso:
## Summary
- add a standalone `test-tls-protocol-fuzz` build step and `-Dtls-protocol-test-filter` namespace for #491
- add seed-backed fuzz targets for TLS handshake framing/extensions/reassembly, ClientHello policy negotiation, transcript HRR rebinding, and HRR/ClientHello2 validation
- document deterministic smoke and longer coverage-guided protocol fuzz commands

## Validation
- `zig build test-tls-protocol-fuzz --summary all --error-style verbose`

Refs #491

### DIFF
--- a/src/tls/engine.zig
+++ b/src/tls/engine.zig
@@ -563,16 +563,19 @@ fn fuzzHandshakeCoreAndDriver(_: void, smith: *std.testing.Smith) !void {
     } else {
         try std.testing.expect(backend.calls > before_calls);
         const repeated = driver.receiveOutcome(.handshake, command);
-        if (repeated.terminal_error) |err| {
-            const calls_after_failure = backend.calls;
-            const sink_len_after_failure = repeated.sink.len;
-            const sink_used_after_failure = repeated.sink.used;
-            const again = driver.receiveOutcome(.handshake, command);
-            try std.testing.expectEqual(err, again.terminal_error.?);
-            try std.testing.expectEqual(calls_after_failure, backend.calls);
-            try std.testing.expectEqual(sink_len_after_failure, again.sink.len);
-            try std.testing.expectEqual(sink_used_after_failure, again.sink.used);
-        }
+        const err = repeated.terminal_error orelse return error.TestExpectedError;
+        try std.testing.expectEqual(error.UnexpectedHandshakeMessage, err);
+
+        const calls_after_failure = backend.calls;
+        const state_after_failure = backend.core.handshake_state;
+        const sink_len_after_failure = repeated.sink.len;
+        const sink_used_after_failure = repeated.sink.used;
+        const again = driver.receiveOutcome(.handshake, command);
+        try std.testing.expectEqual(err, again.terminal_error.?);
+        try std.testing.expectEqual(calls_after_failure, backend.calls);
+        try std.testing.expectEqual(sink_len_after_failure, again.sink.len);
+        try std.testing.expectEqual(sink_used_after_failure, again.sink.used);
+        try std.testing.expectEqual(state_after_failure, backend.core.handshake_state);
     }
 }
 
--- a/src/tls/engine.zig
+++ b/src/tls/engine.zig
@@ -10,6 +10,7 @@ pub const state = @import("state.zig");
 pub const events = @import("events.zig");
 pub const handshake = @import("handshake.zig");
 pub const key_schedule = @import("key_schedule.zig");
+const messages = @import("messages.zig");
 pub const transcript = @import("transcript.zig");
 
 pub const EngineConfig = struct {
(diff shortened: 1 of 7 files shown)


Jaruso (commented):
Not ready to merge yet. The standalone build step, filter namespace, documentation, and CI integration are all in good shape, and the current CI run is green. However, the implementation does not yet satisfy several core #491 acceptance criteria:

- the ClientHello2 target almost never reaches the intended equivalence path and has no semantic success/rejection oracle;
- the reassembly target mutates fragmentation of one valid frame only, not coalesced or attacker-controlled reassembly boundaries;
- negotiation assertions prove policy membership but do not prove peer-offer overlap for all selected values;
- the transcript target does not exercise the actual handshake state machine, malformed ordering, or repeated progression after a terminal error.

Please address the inline comments and rerun the validation commands listed in the PR description. Because this review is being submitted by the PR author account, GitHub only permits a comment review rather than a formal “Request changes” review; these findings are blocking.

src/tls/hello_retry.zig
```diff
@@ -1010,3 +1010,107 @@ test "ClientHello2 validator rejects malformed PSK binder vectors" {
...
+    };
+
+    var first_buf: [768]u8 = undefined;
+    var second_buf: [768]u8 = undefined;
+    const first_share = share_options[smith.index(share_options.len)];
+    const second_share = share_options[smith.index(share_options.len)];
+    const first = try clientHello(&first_buf, @intCast(smith.index(256)), first_share, null, extra_options[smith.index(extra_options.len)]);
+    const second = try clientHello(
+        &second_buf,
+        if (smith.index(4) == 0) 0xbb else first[4],
```
Jaruso: **Blocking — preserve ClientHello1’s actual random in the equivalence branch.**

`first` is a framed handshake message. `first[4]` is the first byte of the ClientHello body (`legacy_version`, always `0x03`), while the 32-byte random starts at raw offset 6. Therefore the nominal “unchanged” branch fills ClientHello2’s random with `0x03` and `validateSecondClientHello` rejects almost every case during the fixed-field comparison, before any HRR extension/cookie/key-share rule is exercised.

Keep the generated byte explicitly instead of reading the framed buffer:

```zig
const first_random: u8 = @intCast(smith.index(256));
const first = try clientHello(
    &first_buf,
    first_random,
    first_share,
    null,
    extra_options[smith.index(extra_options.len)],
);
const second_random = if (smith.index(4) == 0)
    first_random ^ 0xff
else
    first_random;
const second = try clientHello(
    &second_buf,
    second_random,
    second_share,
    second_cookie,
    second_extra,
);
```

This guarantees a real unchanged path and a deliberate mismatched path.

src/tls/hello_retry.zig
```diff
@@ -1010,3 +1010,107 @@ test "ClientHello2 validator rejects malformed PSK binder vectors" {
...
+        cookie_options[smith.index(cookie_options.len)],
+        extra_options[smith.index(extra_options.len)],
+    );
+    const request = Request{
+        .selected_version = .tls13,
+        .cipher_suite = .tls_aes_128_gcm_sha256,
+        .selected_group = share_options[smith.index(share_options.len)],
+        .cookie = cookie_options[smith.index(cookie_options.len)],
+    };
+    validateSecondClientHello(first, second, request) catch |err| switch (err) {
```
Jaruso: **Blocking — give the ClientHello2 validator target a semantic oracle.**

The target currently accepts both success and every listed error. A regression that changes `validateSecondClientHello` to always return success would still pass this fuzz test. The independently sampled `request.selected_group` and `request.cookie` also mean the harness usually cannot know whether the generated pair should be accepted.

Start from a known-valid aligned tuple and assert success, then mutate one rule at a time and assert rejection. For example:

```zig
const request = Request{
    .selected_version = .tls13,
    .cipher_suite = .tls_aes_128_gcm_sha256,
    .selected_group = second_share,
    .cookie = second_cookie,
};
try validateSecondClientHello(valid_first, valid_second, request);

const bad_random = try clientHello(
    &mutated_buf,
    first_random ^ 0xff,
    second_share,
    second_cookie,
    second_extra,
);
try testing.expectError(
    error.IllegalParameter,
    validateSecondClientHello(valid_first, bad_random, request),
);
```

Use separate mutation branches for random/session/cipher changes, stable-extension ordering, requested key share, cookie insertion/removal/value, `early_data`, non-zero padding, and PSK-last rules. That way an unexpected acceptance becomes a fuzz failure rather than another allowed outcome.

src/tls/messages.zig
```diff
@@ -319,3 +319,70 @@ test "reader and writer enforce bounds" {
...
+    var body_storage: [128]u8 = undefined;
+    const body_len = smith.index(body_storage.len + 1);
+    smith.bytes(body_storage[0..body_len]);
+    const kinds = [_]MessageType{ .client_hello, .server_hello, .encrypted_extensions, .certificate, .finished, .message_hash };
+    const encoded = try encode(kinds[smith.index(kinds.len)], body_storage[0..body_len], &encoded_storage);
+    const decoded = try decode(encoded);
+    try testing.expectEqualSlices(u8, encoded, decoded.raw);
+    try testing.expectEqualSlices(u8, body_storage[0..body_len], decoded.body);
+
+    var reassembler = Reassembler(512){};
```
Jaruso, line 377: **Blocking — this does not fuzz coalesced or attacker-controlled reassembly.**

`Reassembler` only receives one frame produced by `encode`, so the fuzzer can vary chunk sizes but cannot exercise two coalesced messages, a chunk crossing a message boundary, malformed u24 lengths, capacity/capacity+1 behavior, partial trailing frames, or invalid discard counts. The existing fixed coalescing unit test is useful, but it is not the mutation-friendly coalescing target required by #491.

Please add at least two paths:

```zig
// Valid coalesced stream: independently generate two frames, concatenate them,
// then choose chunk boundaries across the entire stream.
const expected = [_][]const u8{ first, second };
for (expected) |frame| {
    const message = (try reassembler.peek()).?;
    try testing.expectEqualSlices(u8, frame, message.raw);
    try reassembler.discard(message.raw.len);
}
try testing.expectEqual(@as(usize, 0), reassembler.len);
```

Also feed bounded raw Smith bytes directly through `append`/`peek` using a small capacity and explicit `capacity`, `capacity + 1`, incomplete-header, maximum-length-field, and trailing-partial-frame seeds. Errors should be typed and state/cursor invariants should be checked after failure.

src/tls/negotiation.zig
```diff
@@ -739,3 +739,170 @@ test "malformed required extensions fail deterministically" {
...
+                try testing.expectEqual(selected.named_group, share.group);
+                try testing.expect(parsed.offers.keyShareFor(selected.named_group) != null);
+            },
+            .retry => |group| {
+                try testing.expectEqual(selected.named_group, group);
+                try testing.expect(parsed.offers.key_share_seen);
+                try testing.expect(parsed.offers.keyShareFor(group) == null);
+            },
+        }
+        if (selected.alpn) |alpn| try testing.expect(policy.containsAlpn(alpn.bytes));
```
Jaruso: **Blocking — assert peer-offer overlap, not only policy membership.**

This property would still pass if negotiation regressed to select a policy-approved ALPN that the client never offered, which is exactly the no-overlap failure #491 is meant to protect. The same gap exists for the final signature scheme, selected key-share bytes, and SNI-adjacent result.

Please extend the success invariants along these lines:

```zig
if (selected.alpn) |alpn| {
    try testing.expect(policy.containsAlpn(alpn.bytes));
    try testing.expect(containsAlpn(
        parsed.offers.alpn_protocols[0..parsed.offers.alpn_protocols_len],
        alpn,
    ));
}
```

For `negotiateServer`, also assert that `selected.signature_scheme` occurs in `parsed.offers.signature_schemes`, `selected.alpn` occurs in the offered ALPN list, `selected.key_share` equals `keyShareFor(selected.named_group).?`, and `selected.server_name` matches the offered SNI. Otherwise a policy-valid but peer-unoffered selection remains invisible to this fuzz target.

src/tls/transcript.zig
```diff
@@ -347,3 +347,67 @@ test "an unselected transcript reports no family and a zero-length digest with a
     try testing.expectEqual(@as(usize, 0), digest.len);
     for (digest.bytes) |byte| try testing.expectEqual(@as(u8, 0), byte);
 }
+
+test "fuzz: TLS protocol: transcript update select and HRR rebind sequences are bounded and deterministic" {
```
Jaruso, line 351: **Blocking — add a target for the actual handshake state machine.**

This target exercises only `Transcript`; it never feeds framed handshake messages through `HandshakeState`/`DriverState` or the shared TLS backend. Consequently it cannot detect acceptance of `Finished` before `ServerHello`, duplicate/out-of-order HRR or handshake messages, state advancement after an error, or non-sticky terminal failures. Those are explicit #491 requirements, not transcript-helper properties.

Please add a bounded target beside the engine/backend test harness that generates a short sequence of framed message types and progression calls. It should:

1. include a known-valid ClientHello → ServerHello/HRR → ClientHello2 progression;
2. mutate order, duplicate messages, and insert malformed frames;
3. record the first terminal error;
4. call the same feed/drive/progression entrypoints again and assert the same terminal classification with no new output/events or state advancement.

The existing transcript target can remain, but it does not substitute for this state-machine coverage.

src/tls/messages.zig
```diff
@@ -319,3 +319,122 @@ test "reader and writer enforce bounds" {
...
+    const raw_before_len = raw_reassembler.len;
+    const append_result = raw_reassembler.append(raw);
+    if (append_result) |_| {
+        try testing.expectEqual(raw.len, raw_reassembler.len);
+        const peek_result = raw_reassembler.peek();
+        if (peek_result) |maybe_message| {
+            if (maybe_message) |message| {
+                try testing.expect(message.raw.len <= raw_reassembler.len);
+                try testing.expectEqualSlices(u8, raw_reassembler.data[0..message.raw.len], message.raw);
+                const bad_discard = message.raw.len + 1 + smith.index(raw_reassembler.len - message.raw.len + 1);
```
Jaruso: **Blocking — `bad_discard` is not necessarily outside the buffer.**

When `peek()` returns a complete first frame plus trailing bytes, `raw_reassembler.len - message.raw.len` is positive. `smith.index(...)` can return `0`, making `bad_discard == message.raw.len + 1`, which is still `<= raw_reassembler.len`. `discard` correctly accepts any count within the buffered length, so a coverage-guided run can report a false invariant failure against a correct implementation.

Construct the invalid count from the total buffered length instead:

```zig
const bad_discard = raw_reassembler.len + 1 + smith.index(8);
const before = raw_reassembler.len;
try testing.expectError(
    error.MalformedHandshake,
    raw_reassembler.discard(bad_discard),
);
try testing.expectEqual(before, raw_reassembler.len);
```

Then test partial-but-valid discards separately if that behavior is meant to be covered.

Jaruso (commented):
Follow-up review of `215843f`:

The previous findings were addressed well: ClientHello1's random is now preserved, ClientHello2 has valid and invalid semantic oracles, coalesced/raw reassembly paths were added, negotiation results are checked against peer offers, and the shared handshake core/driver now has HRR, ordering, and terminal-error coverage.

I found two new blocking harness defects in the added fixes. One causes a valid `discard` call to be asserted as invalid when a complete frame has trailing bytes; the other silently accepts a second successful duplicate progression instead of requiring rejection. Both need correction before this target can reliably satisfy #491. CI for this head is still running, with the completed build/test/format/security jobs green so far.

src/tls/engine.zig
```diff
@@ -477,3 +478,146 @@ test "driver deinit is safe after a failed handshake" {
...
+        try std.testing.expectEqual(err, again.terminal_error.?);
+        try std.testing.expectEqual(calls_after_failure, backend.calls);
+        try std.testing.expectEqual(sink_len_after_failure, again.sink.len);
+        try std.testing.expectEqual(sink_used_after_failure, again.sink.used);
+        try std.testing.expectEqual(state_after_failure, backend.core.handshake_state);
+        try std.testing.expectEqual(lifecycle_after_failure, backend.core.handshake_lifecycle);
+    } else {
+        try std.testing.expect(backend.calls > before_calls);
+        const repeated = driver.receiveOutcome(.handshake, command);
+        if (repeated.terminal_error) |err| {
```
Jaruso: **Blocking — require the duplicate operation to fail.**

If the first operation succeeds, this branch only checks stickiness *when* the repeated operation returns an error. If a regression allows the duplicate ClientHello send/receive to succeed twice, `repeated.terminal_error` is null and the fuzz case passes silently.

From the initial state used here, the successful generated cases are the client's first `recordSent(client_hello)` and the server's first `acceptReceived(client_hello)`; repeating either must be rejected. Make that a semantic oracle:

```zig
const repeated = driver.receiveOutcome(.handshake, command);
const err = repeated.terminal_error orelse
    return error.TestExpectedError;
try std.testing.expectEqual(error.UnexpectedHandshakeMessage, err);

const calls_after_failure = backend.calls;
const state_after_failure = backend.core.handshake_state;
const again = driver.receiveOutcome(.handshake, command);
try std.testing.expectEqual(err, again.terminal_error.?);
try std.testing.expectEqual(calls_after_failure, backend.calls);
try std.testing.expectEqual(state_after_failure, backend.core.handshake_state);
```

Without the non-null assertion, the target does not actually detect duplicate-message acceptance.

Jaruso (commented):
Follow-up review complete. Both remaining blocking findings are fixed correctly:

- the reassembly harness now guarantees the invalid discard count is beyond the total buffered length and verifies that the failed discard leaves state unchanged;
- the handshake driver harness now requires duplicate ClientHello progression to fail with `UnexpectedHandshakeMessage`, then verifies sticky terminal behavior and no subsequent state/backend advancement.

I rechecked the updated delta against the previously reviewed #491 acceptance criteria and found no further code issues. All review threads are resolved. The current CI run (#1217) is still in progress, with formatting, lint, security, several build configurations, PKI differential, dependency review, and crypto benchmark checks already passing. From a code-review perspective, this is ready to merge once the remaining CI jobs complete successfully.

PR #575, opened 2026-08-04
security: [494-B] fuzz authenticated ticket open and keyring publication

Jaruso:
## Summary

Second PR of #494's recommended decomposition (#494-B: "ticket security and key publication"). Adds two new `fuzz: TLS resumption: ...` targets to `src/tls/ticket_protection.zig`, following the pattern #494-A (#574) established for `NewSessionTicket`/session codec/PSK/`parseEnvelope`.

- **Authenticated `Protector.resolve` target** — drives the accept path, every typed `ResolveRejectReason` (`unknown_key`, `future_key`, `retired_key`, cross-AEAD `authentication_failed`, and tampered-nonce/ciphertext/tag `authentication_failed`), and a bounded allocation-failure sweep against a structurally valid sealed envelope. Every scenario runs through the file's existing `ZeroCheckingAllocator` to prove the temporary plaintext buffer is wiped before free on the **accept** path too, not just the miss path the existing deterministic `"seal and invalid-plaintext resolve zeroize plaintext before free"` test names.
- **Keyring/snapshot publication operation-sequence target** — a bounded program of `Snapshot.build`/`ReloadableKeyRing.install`/`validateInstallCandidate`/`acquireCurrent`/`Snapshot.retain`/`release` calls, asserting after every step that generation is monotonic, the ledger never shrinks, and a retained old snapshot stays fully intact for as long as it is held — even across installs that replace it as `keyring.current`. The existing `fuzzSnapshotConfig` target already exhaustively fuzzes the *static* properties of a single build/install call; this closes the *temporal* gap a single-call target can't reach.

The existing `ticket_key_snapshot.zig` persistent-JSON-snapshot fuzz target (on-disk snapshot parsing, unrelated to the in-memory `ReloadableKeyRing`) is untouched, per the issue's "existing coverage to preserve, not recreate" list.

`docs/CRYPTO_FUZZ_CONTRACT.md`'s `#494` section is updated with the two new target names/commands and an updated status note (no `build.zig` changes needed — the `test-tls-resumption-fuzz` step already filters generically on the `"fuzz: TLS resumption:"` prefix from #494-A).

## Test plan

- [x] `zig fmt --check build.zig src/ tests/`
- [x] `zig build test-tls-resumption-fuzz --summary all --error-style verbose` (11/11 pass, deterministic seed-corpus replay)
- [x] `zig build test-tls --summary all --error-style verbose` (885/885 pass)
[...]

PR #579, opened 2026-08-06
security: [493-A] fuzz TLS record codec and inner-plaintext framing

Jaruso:
## Summary
- First of three planned PRs for #493 (record/protection/epoch/encrypted-stream fuzzing, epic #325-K), per the issue's own "Implementation plan / agent handoff" comment recommending a 493-A/493-B/493-C split — mirroring how #494 was delivered as #574 (494-A) + #575 (494-B).
- Adds `test-tls-record-fuzz` build step + `-Dtls-record-test-filter` option in `build.zig`, matching the existing `test-tls-resumption-fuzz`/`test-tls-protocol-fuzz` pattern.
- Adds two `std.testing.Smith` fuzz targets in `src/tls/record_codec.zig`:
  - `fuzz: TLS record: codec fragmentation, coalescing, and sink saturation preserve exact consumption` — drives fuzzer-generated valid record streams through `Parser.feedOne` against a single-record-capacity sink, asserting exact per-record byte consumption, correct saturation/`drainReady`/`finish` backpressure semantics (the #408 regression class), and that arbitrary malformed bytes into a fresh parser never panic or exceed the fixed pending buffer.
  - `fuzz: TLS record: inner plaintext framing, padding, and bounds remain transactional` — fuzzes `encodeInnerPlaintext`/`decodeInnerPlaintext` round trips plus raw-byte decoding against an independent last-nonzero-byte oracle, asserting transactional output on every rejected encode and borrowed-slice containment on every accepted decode.
- Removes the shallow, superseded `#327-G` `fuzzRecordInput` helper (unused elsewhere, no real corpus, swallowed all errors).
- Hardens record-length-only arithmetic reachable from fuzzer-chosen scalar lengths to checked `std.math` arithmetic (per the issue's "harden synthetic length arithmetic" requirement), each with its own named boundary test:
  - `record_codec.zig`: new `checkedOwnedLen`/`checkedRecordLen` helpers back `Parser.pendingRecordBytesNeededWith`/`pendingRecordPayloadLenWith`.
  - `record_epoch_bridge.zig`: new `chunkCount` helper (via `std.math.divCeil`) backs `plaintextHandshakeRecordLen`/`protectedHandshakeRecordLen`/`sealHandshakeFragments`/`sealedHandshakeLen`.
  - `encrypted_stream.zig`: `handshakeRecordCount` converted the same way.
- Adds a `#493` section to `docs/CRYPTO_FUZZ_CONTRACT.md` documenting the target names, delivery split, and reproduction commands, and a `CHANGELOG.md` Testing entry.

## Test plan
- [x] `zig build test-tls-record-fuzz --summary all --error-style verbose` (deterministic corpus replay, 4/4 pass)
- [x] `zig build test-tls --summary all --error-style verbose` (905/905 pass)
- [x] `zig build test --summary all --error-style verbose` (3180/3181 pass, 1 pre-existing skip per repo convention — sandboxed UDP smoke test)
- [x] `zig build test -Dtls-profile=appliance --summary all --error-style verbose` (3149/3150 pass, same pre-existing skip)
- [x] Bounded `ReleaseFast` fuzz smoke, 200,000 runs each, no crashes:
  - `zig build test-tls-record-fuzz -Doptimize=ReleaseFast -Dtls-record-test-filter="fuzz: TLS record: codec fragmentation, coalescing, and sink saturation preserve exact consumption" --fuzz=200000`
  - `zig build test-tls-record-fuzz -Doptimize=ReleaseFast -Dtls-record-test-filter="fuzz: TLS record: inner plaintext framing, padding, and bounds remain transactional" --fuzz=200000`

## Follow-on

493-B (record protection + epoch/key lifecycle) and 493-C (scripted in-memory carrier + encrypted-stream progression + docs closeout) are tracked as separate follow-on PRs against the same issue.

Closes: none (partial — #493 stays open for 493-B/493-C)
Ref #493

### DIFF
--- a/src/tls/record_codec.zig
+++ b/src/tls/record_codec.zig
@@ -1115,6 +1115,64 @@ test "inner plaintext encodes content type followed by zero padding" {
     try testing.expectEqual(@as(usize, 3), decoded.padding_len);
 }
 
+// Pins each of `encodeInnerPlaintext`'s four typed outcomes at its exact
+// boundary, in the documented rejection order. The fuzz target below asserts
+// the same classification against generated inputs, but its `.corpus` replay
+// (what plain `zig build test`/CI runs) is not guaranteed to reach every arm,
+// so an error-class regression -- e.g. returning `RecordTooLarge` for an
+// ordinary undersized output -- needs deterministic coverage here too. This
+// was verified by mutation: swapping the `out.len < total` error makes this
+// test fail.
+test "encodeInnerPlaintext classifies every rejection stage at its exact boundary" {
+    var out: [max_ciphertext_fragment_len]u8 = undefined;
+    const content = "finished";
+
+    // 1. change_cipher_spec has no legal inner-plaintext encoding, and is
+    // refused before any length is even considered.
+    try testing.expectError(error.InvalidRecordType, encodeInnerPlaintext(.change_cipher_spec, content, 0, &out));
+    try testing.expectError(error.InvalidRecordType, encodeInnerPlaintext(.change_cipher_spec, "", 0, out[0..0]));
+
+    // 2. content longer than the plaintext fragment maximum.
+    const oversized_content = [_]u8{0} ** (max_plaintext_fragment_len + 1);
+    try testing.expectError(error.RecordTooLarge, encodeInnerPlaintext(.handshake, &oversized_content, 0, &out));
+    // Exactly at the maximum, content alone is fine; it is the total that
+    // then exceeds the ciphertext fragment maximum only once padding pushes
+    // it over (checked in stage 3).
+    const max_content = [_]u8{0} ** max_plaintext_fragment_len;
+    const at_max = try encodeInnerPlaintext(.handshake, &max_content, 0, &out);
+    try testing.expectEqual(max_plaintext_fragment_len + 1, at_max.len);
+
+    // 3. total (content || type || padding) across the ciphertext fragment
+    // maximum: exact-minus-one and exact succeed, exact-plus-one does not.
+    const max_total = max_ciphertext_fragment_len;
+    const under = try encodeInnerPlaintext(.handshake, content, max_total - content.len - 2, &out);
+    try testing.expectEqual(max_total - 1, under.len);
+    const exact = try encodeInnerPlaintext(.handshake, content, max_total - content.len - 1, &out);
+    try testing.expectEqual(max_total, exact.len);
+    try testing.expectError(
+        error.RecordTooLarge,
+        encodeInnerPlaintext(.handshake, content, max_total - content.len, &out),
+    );
+    // And a padding length that overflows `usize` arithmetic outright.
+    try testing.expectError(error.RecordTooLarge, encodeInnerPlaintext(.handshake, content, std.math.maxInt(usize), &out));
+
+    // 4. output capacity, at exact-minus-one / exact / exact-plus-one. This
+    // is the arm that must stay `RecordBufferOverflow` and not collapse into
+    // `RecordTooLarge`.
+    const needed = content.len + 1 + 3;
+    try testing.expectError(
+        error.RecordBufferOverflow,
+        encodeInnerPlaintext(.handshake, content, 3, out[0 .. needed - 1]),
+    );
+    const exact_out = try encodeInnerPlaintext(.handshake, content, 3, out[0..needed]);
+    try testing.expectEqual(needed, exact_out.len);
+    const roomy_out = try encodeInnerPlaintext(.handshake, content, 3, out[0 .. needed + 1]);
+    try testing.expectEqual(needed, roomy_out.len);
+    // A zero-capacity destination is still the output-capacity error, not a
+    // size error, even for the smallest possible encoding.
+    try testing.expectError(error.RecordBufferOverflow, encodeInnerPlaintext(.handshake, "", 0, out[0..0]));
+}
+
 test "inner plaintext rejects all-zero, invalid type, and oversized content" {
     try testing.expectError(error.MalformedInnerPlaintext, decodeInnerPlaintext(&.{ 0, 0, 0 }));
     try testing.expectError(error.MalformedInnerPlaintext, decodeInnerPlaintext(&.{ 1, 99, 0 }));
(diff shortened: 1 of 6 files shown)


src/tls/record_codec.zig
```diff
@@ -1114,6 +1125,311 @@ test "inner plaintext rejects all-zero, invalid type, and oversized content" {
...
+        &[_]u8{ 20, 3, 3, 0, 1, 1 },
+        &[_]u8{ 23, 3, 3, 0, 3, 9, 9, 9 },
+        &([_]u8{0xff} ** 64),
+        &([_]u8{0x01} ** 200),
+    } });
+}
+
+fn fuzzCodecFragmentationInput(_: void, smith: *std.testing.Smith) !void {
+    const mode: RecordMode = if (smith.index(2) == 0) .plaintext else .ciphertext;
+    var parser = Parser.init(mode);
```
Jaruso: **Blocking — the 0x0301 compatibility state never enters this fuzz target.** `Parser.init(mode)` is always strict, so the generated/property path cannot exercise the #408 initial-ClientHello compatibility window that the #493-A handoff explicitly requires: legal `0x0301` on the initial ClientHello (including fragmentation inside its 4-byte handshake header/body), followed by rejection on later records, non-ClientHello handshake messages, non-handshake content, and ciphertext mode. The existing deterministic tests are useful, but #493 specifically asks that those regression classes become persistent fuzz seeds/properties.

Please make the plaintext parser policy part of the Smith-generated case and model the window explicitly. For example:

```zig
const allow_compat = mode == .plaintext and smith.index(2) == 0;
var parser = if (allow_compat)
    Parser.initWithVersionPolicy(.plaintext, .allow_initial_client_hello_compat)
else
    Parser.init(mode);
```

Then generate a real `clientHelloMessage`, mutate the per-record legacy version between `0x0301`/`0x0303`, and assert the exact `InvalidRecordVersion` outcome once the compatibility window is not legal. That gives the fuzzer the stateful transition the deterministic tests currently cover only by fixed examples.

src/tls/record_codec.zig
```diff
@@ -1114,6 +1125,311 @@ test "inner plaintext rejects all-zero, invalid type, and oversized content" {
...
+    // Drive the coalesced oracle stream through `feedOne` with a
+    // single-record-capacity sink, so saturation is normal rather than
+    // exceptional: every source byte must be consumed exactly once, in
+    // order, with no record skipped or duplicated.
+    var sink = RecordSink(1, fuzz_codec_max_payload + header_len){};
+    var offset: usize = 0;
+    var expected_idx: usize = 0;
+    while (offset < stream_len) {
+        try testing.expectEqual(@as(usize, 0), sink.len);
+        const result = try parser.feedOne(stream_buf[offset..stream_len], &sink);
```
Jaruso: **Blocking — this does not actually fuzz caller-visible fragmentation.** Every iteration passes `stream_buf[offset..stream_len]`, i.e. the entire remaining coalesced stream. `feedOne` happens to ingest bytes internally one at a time, but that is not equivalent to repeated API calls with split headers/payloads: it never leaves a Smith-chosen partial header/body buffered across an invocation boundary and then resumes it with a later slice. That is the exact state boundary #493 asks this target to property-test.

Please choose an offered chunk length from Smith and allow `feedOne` to return `emitted == false` while still consuming the offered fragment, e.g.:

```zig
const remaining = stream_len - offset;
const offered_len = 1 + smith.index(remaining);
const result = try parser.feedOne(stream_buf[offset..][0..offered_len], &sink);
offset += result.consumed;

if (!result.emitted) {
    try testing.expectEqual(@as(usize, 0), sink.len);
    continue;
}
// validate exactly one emitted oracle record, then reset/drain as needed
```

Keep the exact-consumption oracle across calls, and include deterministic seed programs that force splits after each of the 5 header bytes plus representative payload boundaries. That makes `fragmentation, coalescing, and sink saturation` true at the public API boundary rather than only inside `feedOne`'s implementation loop.

src/tls/record_codec.zig
```diff
@@ -1114,6 +1125,311 @@ test "inner plaintext rejects all-zero, invalid type, and oversized content" {
...
+    if (inner.len == 0) return true;
+    const outer_start = @intFromPtr(outer.ptr);
+    const outer_end = outer_start + outer.len;
+    const inner_start = @intFromPtr(inner.ptr);
+    const inner_end = inner_start + inner.len;
+    return inner_start >= outer_start and inner_end <= outer_end and inner_end >= inner_start;
+}
+
+const fuzz_codec_max_records = 4;
+const fuzz_codec_max_payload = 48;
```
Jaruso, line 1237: **Blocking — the valid generator cannot reach the record-size boundaries required by 493-A.** With `fuzz_codec_max_payload = 48`, neither plaintext (`16384`) nor ciphertext (`16384 + 256`) exact-maximum records can ever traverse the valid encode/parse/oracle path. The later 256-byte junk buffer also cannot complete an exact-max valid record; it can only exercise malformed/truncated headers, and its errors are swallowed. The issue handoff explicitly calls for zero, exact maximum, and maximum-plus-one record-size properties.

Please add a bounded boundary subcase rather than making every generated record huge. A fixed stack buffer sized to `max_ciphertext_record_len` is enough, with a focused length matrix such as:

```zig
const payload_len = switch (smith.index(5)) {
    0 => 0,
    1 => 1,
    2 => max_len - 1,
    3 => max_len,
    else => small_random_len,
};
```

Then separately mutate a serialized header to declare `max_len + 1` and require `RecordTooLarge` without needing to allocate that payload. The inner-plaintext encode target has the same issue with content/padding capped at 512; add a focused total-length matrix around `max_ciphertext_fragment_len - 1`, exact max, and max+1 there as well so the encoder's maximum boundary is a property, not only a deterministic test.

src/tls/record_codec.zig
```diff
@@ -1114,6 +1125,311 @@ test "inner plaintext rejects all-zero, invalid type, and oversized content" {
...
+
+    var out_buf: [fuzz_inner_plaintext_max_out]u8 = undefined;
+    const out_cap = smith.index(out_buf.len + 1);
+    const out = out_buf[0..out_cap];
+    @memset(out, 0xa5);
+    var before_buf: [fuzz_inner_plaintext_max_out]u8 = undefined;
+    @memcpy(before_buf[0..out_cap], out);
+
+    encode_check: {
+        const encoded = encodeInnerPlaintext(content_type, content_buf[0..content_len], padding_len, out) catch {
```
Jaruso: **Blocking — this accepts the wrong typed error as a passing fuzz case.** These inputs are generated by the harness, so the expected outcome is known. Right now *any* `encodeInnerPlaintext` error is treated as a valid rejection as long as the sentinel remains unchanged. A regression that returns `RecordTooLarge` for an ordinary undersized output (or `InvalidRecordType` for a valid handshake type) would therefore pass this target, which conflicts with #376/#493's requirement to classify failures deterministically by typed stage/error.

Please compute the expected class before the call and assert it explicitly. Roughly:

```zig
const content_and_type = std.math.add(usize, content_len, 1) catch null;
const total = if (content_and_type) |n| std.math.add(usize, n, padding_len) catch null else null;

if (total == null or total.? > max_ciphertext_fragment_len) {
    try testing.expectError(error.RecordTooLarge,
        encodeInnerPlaintext(content_type, content, padding_len, out));
} else if (out_cap < total.?) {
    try testing.expectError(error.RecordBufferOverflow,
        encodeInnerPlaintext(content_type, content, padding_len, out));
} else {
    const encoded = try encodeInnerPlaintext(content_type, content, padding_len, out);
    // round-trip assertions...
}
```

Also occasionally generate `.change_cipher_spec` and require `InvalidRecordType`. Keep the sentinel comparison after each expected failure so transactional output remains independently verified.

Jaruso (commented):
I reviewed all six changed files against #493 and the issue's `Implementation plan / agent handoff`, including the #408 regression classes that 493-A is supposed to promote into persistent fuzz properties. The build-step wiring, checked arithmetic changes in `record_codec.zig` / `record_epoch_bridge.zig` / `encrypted_stream.zig`, documentation structure, and changelog entry all look sound. The current CI run for this head is also fully green across the repository matrix.

**I do not consider this ready to merge yet.** GitHub will not allow this account to submit a formal `REQUEST_CHANGES` review on its own PR, so I am leaving the blocking findings as a comment review instead.

The new codec/inner-plaintext fuzz harness does not yet cover several requirements specifically assigned to 493-A. The existing deterministic tests are good regression coverage, but the issue explicitly says those #408 behaviors must become fuzz seeds/properties rather than remain deterministic-only.

Blocking gaps:

1. The codec target always constructs a strict parser, so the legal initial-ClientHello `0x0301` compatibility state (and its illegal later/ciphertext/non-ClientHello variants) is not fuzzed.
2. The target named `fragmentation...` always gives `feedOne` the entire remaining stream; caller-visible fragmentation across repeated invocations is therefore not randomized/property-tested.
3. Generated valid records are capped at 48-byte payloads, so exact maximum plaintext/ciphertext records are never driven through the valid parser/serialization path. The inner-plaintext encode generator has the analogous <=512-byte content/padding gap around the protocol maximum.
4. The inner-plaintext encoder path treats every error as an acceptable rejection. Since the generated inputs make the expected error computable, this can hide a deterministic error-class regression and conflicts with #376's typed-failure rule.

I left inline comments with concrete implementation shapes for each item. After those are addressed, the rest of this slice looks close to approval. Please also adjust the #493 documentation wording that currently says these boundary/property cases are already exercised so it matches the final harness exactly.

Jaruso (commented):
Re-reviewed the updated head against #493-A and the four blockers from the previous pass. All four are addressed:

- the codec fuzz target now enters and models the initial-ClientHello `0x0301` compatibility window, including fragmented ClientHello state and illegal post-window/non-ClientHello/non-handshake/ciphertext placements;
- `feedOne` is now fuzzed across caller-visible fragment boundaries with exact consumed-byte accounting, plus deterministic split-point coverage;
- plaintext/ciphertext record maxima and max+1 declarations are exercised through the real encode/parse paths, and the inner-plaintext target now covers its content/total/output boundaries;
- `encodeInnerPlaintext` failures are predicted before the call and asserted by exact typed error, with transactional destination checks preserved.

I also rechecked the unchanged build wiring and checked-arithmetic changes from the first pass, and reviewed the expanded docs/changelog for consistency with the implementation. CI run 1223 is green on head `4d036c42`.

No further findings. I consider #493-A ready to merge. GitHub does not allow this account to formally approve its own PR, so this is submitted as a non-blocking COMMENT review.

PR #583, opened 2026-08-08
security: [324-K] fuzz DER, PEM, X.509, and path-validation boundaries

Jaruso:
## Summary

Adds persistent, mutation-driven fuzz targets for the pure-Zig PKI stack under the shared #376 fuzzing contract, plus the `test-pki-fuzz` build step and `-Dpki-test-filter` option that scope them for long coverage-guided runs. Six targets in the `fuzz: PKI:` namespace, inline next to the code they exercise (so their deterministic seed-corpus replay is already part of `zig build test-pki` / `zig build test` — no new CI job).

- **DER cursor/view safety** (`der_tests.zig`) — every borrowed `Element` view stays inside the caller's buffer at the offset it reports; the cursor advances strictly and never leaves its region, including after a failed parse; typed decoders (`readInteger`/`readBitString`/`readObjectIdentifier`/time/string/`readSequence`/`readExplicitContext`) obey the same rule; every prefix of the input is truncation-tested; `readElementDiagnostic` offsets stay in range; `expectEnd` is exact.
- **DER bounds** (`der_tests.zig`) — nesting depth, short/long-form and non-minimal lengths, indefinite and over-wide lengths, `max_element_len` ±1, a declared length one past the bytes present, a `2^64-1` length that must not wrap, trailing data, and the element-count, integer-byte, and OID-component limits, each at exact, one-under, and one-over.
- **PEM/chain loading** (`pem_tests.zig`) — hostile boundaries, labels, base64, block types, duplicate blocks, oversized certificates, and mixed valid/malformed input, with an all-well-formed generator mode giving an exact oracle for the empty-chain and `max_certificates` boundaries. The returned chain is proved to be an owned copy: it is byte-unchanged after the input buffer is overwritten and freed. Full `checkAllAllocationFailures` sweep.
- **X.509 semantic model** (`x509_tests.zig`) — structured generation of Basic Constraints, Key Usage, EKU, SAN (DNS/IP/email/URI/directoryName/registeredID/other), Name Constraints, SKI/AKI, policies, and inhibitAnyPolicy — each optionally critical, optionally replaced with arbitrary bytes — plus malformed signature parameters, duplicate extensions, and bit-flip mutation over the finished DER. Asserts view containment, SAN 4/16 and name-constraint 8/32 IP forms, no repeated extension OID, accessor/extension-list agreement, `hasUnhandledCriticalExtension` consistency, and identical models for identical bytes. `max_extensions` checked at exact and one-over. Full allocation-failure sweep.
- **Path building** (`path_builder_tests.zig`) — adversarial graphs over a four-name alphabet with cycles, cross-signing, ambiguity, byte-identical duplicates across pools, and missing issuers, under Smith-chosen depth/fanout/visit budgets. Asserts leaf-first/anchor-last shape, no repeated certificate in a path, real RFC 5280 §7.1 chaining edges, correct source/input-index reporting, and byte-identical replay including which error is returned.
- **Path validation** (`path_validator_tests.zig`) — Ed25519-signed synthetic chains with hostile extensions, forged signatures, and validity windows on the boundary. Asserts an accepted path is one of the builder's candidates and anchored at a configured anchor by index *and* exact DER, never accepted on a forged signature; verdicts replay identically; and an allocation-failure sweep proves every byte allocated on a failing path is released. A compile-time structural check proves `ValidationFailure` has no pointer-shaped field anywhere in its layout, so a verdict cannot carry certificate bytes or key material out of the validator. A deterministic sub-oracle pins the validity boundary exactly (`notBefore - 1` → `certificate_not_yet_valid`, `notBefore`/`notAfter` → accepted, `notAfter + 1` → `certificate_expired`).

`docs/CRYPTO_FUZZ_CONTRACT.md` gains a `### #492` section recording the stable step/filter names, the per-target property matrix, seed provenance, and the CI model, following the convention #491/#493/#494 established.

### Relationship to #348

#348 stays the owner of semantic differential validation against OpenSSL and a second validator; nothing here re-opens it. Its minimized hostile fixtures (`tests/vectors/pki/reduced/`) seed the DER and X.509 corpora directly, and the PEM surface replays every reduced seed through `loadCertificateDer`/`fuzzLoadChainPem` in a new deterministic regression. The path targets build their own synthetic chains because they need certificates that *parse* to reach path logic at all.

None of these targets performs AIA/OCSP/CRL fetching, and none can: the modules under test contain no network code, and validation takes an injected `validation_time` rather than reading a clock.

## Test plan

- [x] `zig build test` — 3190/3191 pass (1 pre-existing skip), on top of a rebase onto `main` at a22cda79.
- [x] `zig build test-pki-fuzz` — deterministic seed-corpus replay of all six targets.
- [x] `zig fmt --check build.zig src/ tests/` clean.
- [x] `zig build test-tls-record-fuzz` still passes after resolving the `build.zig`/doc merge with #579.
- [x] Coverage-guided runs, `-Doptimize=ReleaseFast --fuzz=<N>`, roughly 2.4M runs total across the six targets (DER decoding 202K, DER bounds 200K, PEM 561K, X.509 781K + 1.19M pre-fix, path building 474K, path validation 144K). **No product defect found.**
- [x] Two invalid fixture combinations in the new harnesses were found by fuzzing and fixed with explanatory comments — an all-zero Key Usage and a `pathLenConstraint` without `cA` — both of which the #341 parser correctly rejects, so they were building unparsable fixtures rather than reaching validation. One latent false-positive oracle (a duplicate-extension assertion that a bit-flip could legitimately invalidate) was tightened to the unmutated case.

Closes #492


Jaruso (commented):
Reviewed all 7 changed files against #492 and the inherited #376 fuzzing contract, including the production DER/X.509/path-validator APIs the new oracles are meant to exercise. CI is green and the overall harness structure is strong: the stable `test-pki-fuzz`/filter wiring is correct, #348 corpus provenance is present, PEM ownership uses a real mutate/free lifetime check plus `checkAllAllocationFailures`, and the path-builder target is meaningfully bounded.

GitHub does not permit the authenticated account to formally request changes on its own PR, but I consider the inline findings below blocking before merge. The most important one is an actual overflow-adjacent DER defect that the new target currently cannot reach: `Reader.childReader` performs an unchecked `content_offset + content_len`. The other findings are acceptance-criteria gaps where the test claims a stronger property than it currently proves (complete borrowed-view containment/model determinism, certificate `pathLenConstraint` enforcement, complete verdict determinism, and a full allocation-failure sweep).

Once these are addressed, I don’t see another architectural blocker in this PR.

src/pki/der_tests.zig
```diff
@@ -457,6 +457,333 @@ fn fuzzDerParse(_: void, smith: *testing.Smith) !void {
...
+        // The header is at least a tag byte and a length byte, the content
+        // ends exactly where the cursor now is, and `content_offset` is an
+        // absolute index into the root input.
+        try testing.expect(elem.encoded.len >= elem.content.len + 2);
+        try testing.expectEqual(after, elem.content_offset + elem.content.len);
+        try testing.expectEqual(input.ptr + elem.content_offset, elem.content.ptr);
+        try testing.expectEqual(input.ptr + before, elem.encoded.ptr);
+
+        if (elem.tag.constructed) {
+            var child = reader.childReader(elem.content_offset, elem.content.len) catch continue;
```
Jaruso, line 503: [P1] This only calls `childReader` with offsets/lengths returned by a successful `readElement`, so those values have already been constrained to the backing input. That means the target cannot exercise #492/#376's required overflow-adjacent **caller offset** cases. There is a real production bug hiding behind that gap: `Reader.childReader` currently computes `const end = content_offset + content_len` unchecked. A regression like this panics in checked builds (and can wrap into an invalid bounded reader in `ReleaseFast`):

```zig
var r = der.Reader.init(&[_]u8{ 0, 0 }, .{});
try testing.expectError(
    error.LengthBeyondInput,
    r.childReader(std.math.maxInt(usize), 2),
);
```

Please add a synthetic overflow-adjacent oracle (not just wire-derived offsets) and harden `childReader`, e.g. with a checked add:

```zig
const end = std.math.add(usize, content_offset, content_len) catch
    return self.fail(error.LengthBeyondInput, content_offset);
```

This is exactly the class the issue calls out as `integer/offset overflow-adjacent cases`, and the shared contract explicitly asks for synthetic caller limits when wire input cannot reach machine-width boundaries.

src/pki/x509_tests.zig
```diff
@@ -1122,3 +1122,393 @@ test "reduced differential seeds keep their recorded parse outcome" {
...
+    if (x509.Certificate.parse(allocator, input, fuzz_x509_limits)) |parsed| {
+        var certificate = parsed;
+        defer certificate.deinit(allocator);
+
+        try testing.expectEqual(input.ptr, certificate.raw.ptr);
+        try testing.expectEqual(input.len, certificate.raw.len);
+        try expectViewContained(certificate.raw, certificate.tbs_raw);
+        try expectViewContained(certificate.raw, certificate.serial_number.content);
+        try expectViewContained(certificate.raw, certificate.signature_algorithm.raw);
+        try expectViewContained(certificate.raw, certificate.subject_public_key_info.raw);
```
Jaruso: [P2] The target is documented as proving **every borrowed view** stays inside `certificate.raw`, but this list only checks a subset of the ownership contract. `Certificate` still exposes borrowed slices that are not covered here, including `signature_algorithm.parameters_raw`, `subject_public_key_info.algorithm.raw/parameters_raw`, `subject_public_key_info.subject_public_key.data`, issuer/subject RDN attribute `value` slices, optional issuer/subject unique-ID bitstrings, AIA `location` GeneralNames, and CRL distribution-point `full_names`.

A parser regression that accidentally points any of those at temporary/recycled storage would still pass this fuzz target, so the #492 lifetime criterion is not fully enforced. Please factor this into recursive containment helpers for `AlgorithmIdentifier`, `Name`, `SubjectPublicKeyInfo`, `GeneralName`, etc., and walk every borrowed field in the parsed model.

src/pki/x509_tests.zig
```diff
@@ -1122,3 +1122,393 @@ test "reduced differential seeds keep their recorded parse outcome" {
...
+        defer replay.deinit(allocator);
+        try testing.expectEqual(certificate.version, replay.version);
+        try testing.expectEqual(certificate.extensions.len, replay.extensions.len);
+        try testing.expectEqualSlices(u8, certificate.tbs_raw, replay.tbs_raw);
+        try testing.expectEqualSlices(u8, certificate.issuer.chaining_key, replay.issuer.chaining_key);
+        try testing.expectEqualSlices(u8, certificate.subject.chaining_key, replay.subject.chaining_key);
+        for (certificate.extensions, replay.extensions) |mine, other| {
+            try testing.expect(mine.oid.eql(&other.oid));
+            try testing.expectEqual(mine.critical, other.critical);
+            try testing.expectEqual(@as(std.meta.Tag(x509.Extension.Parsed), mine.parsed), @as(std.meta.Tag(x509.Extension.Parsed), other.parsed));
```
Jaruso: [P2] This does not yet prove the comment's “identical bytes produce an identical model” property. The replay comparison checks the extension OID/critical bit/**union tag**, but not the parsed extension payload. It also omits other semantic fields such as serial number, algorithms/parameters, validity, SPKI/key type, unique IDs, and signature value. Two parses could therefore disagree on `BasicConstraints`, KU/EKU, SAN/Name Constraints, policy data, etc. while this oracle still passes.

Please compare the complete semantic model (or add a deterministic model fingerprint/equality helper that includes every public semantic field and parsed extension payload), rather than only the structural tags.

src/pki/path_validator_tests.zig
```diff
@@ -3042,3 +3042,309 @@ test "policy mappings apply through an asserted anyPolicy" {
...
+        .rejected => |mine| switch (second) {
+            .accepted => return error.TestUnexpectedResult,
+            .rejected => |other| {
+                try testing.expectEqual(mine.reason, other.reason);
+                try testing.expectEqual(mine.certificate_index, other.certificate_index);
+                try testing.expectEqual(mine.constraint_certificate_index, other.constraint_certificate_index);
+                try testing.expectEqual(mine.name_constraint_kind, other.name_constraint_kind);
+                try testing.expectEqual(mine.name_form, other.name_form);
+                try testing.expectEqual(mine.policy_stage, other.policy_stage);
+                try testing.expectEqual(mine.policy_graph_depth, other.policy_graph_depth);
```
Jaruso, line 3333: [P2] `expectSameVerdict` omits `extension_oid` and `policy_oid`, even though both are part of `ValidationFailure`. Two identical validation runs could return the same reason/stage/index metadata but different OIDs and this determinism oracle would still pass.

Please compare both optional OIDs by value as part of the verdict, for example with a small helper that handles `null`/non-`null` and uses `ObjectIdentifier.eql` for the populated case. The #492 acceptance criterion is deterministic **failure output**, so every field in the failure value needs to participate.

src/pki/path_validator_tests.zig
```diff
@@ -3042,3 +3042,309 @@ test "policy mappings apply through an asserted anyPolicy" {
...
+            .subject_key = 2,
+            .issuer_key = if (forged_intermediate) 9 else 3,
+            .not_before = intermediate_window[0],
+            .not_after = intermediate_window[1],
+            .ca = intermediate_ca,
+            // RFC 5280 §4.2.1.9 only permits pathLenConstraint alongside
+            // cA TRUE, and the #341 parser enforces that, so pairing it with
+            // a non-CA here would build an unparsable fixture instead of an
+            // interesting validation case.
+            .path_len = if (intermediate_ca == true and smith.index(3) == 0) @intCast(smith.index(3)) else null,
```
Jaruso, line 3640: [P1] This generator never actually exercises RFC 5280 certificate `pathLenConstraint` enforcement. There is at most one intermediate. In `validatePath`, the intermediate at leaf-first index 1 counts CA certificates in `path.elements[1..certificate_index]`, which is an empty slice at index 1, so its `path_len` can never be exceeded. The root's `path_len` also does not help because trust-anchor extensions are intentionally not validation inputs.

That leaves #492's required “path-length constraints” property uncovered despite generating the field. Please add a shape with at least two intermediates (e.g. `leaf -> I1 -> I2 -> root`) or a deterministic sub-oracle. A useful pin would be `I2: cA=true, pathLenConstraint=0` => `path_length_exceeded` at I2, while the otherwise-identical `pathLenConstraint=1` case accepts.

src/pki/path_validator_tests.zig
```diff
@@ -3042,3 +3042,309 @@ test "policy mappings apply through an asserted anyPolicy" {
...
+    // Identical bytes, policy, and time must produce an identical verdict.
+    var replay = validator.validateCandidates(allocator, candidates, validation_policy, cp);
+    defer replay.deinit(allocator);
+    try expectSameVerdict(result, replay);
+
+    // Allocation failure at every reachable point: the verdict degrades to a
+    // structured rejection and every byte allocated on the failing path is
+    // released, so no partially owned accepted path escapes.
+    var fail_index: usize = 0;
+    while (fail_index < 32) : (fail_index += 1) {
```
Jaruso: [P2] The comment says “every reachable point”, but the hard `fail_index < 32` cap makes the sweep silently incomplete. If a generated validation path performs 33+ allocations, indices 0–31 can all induce failures and the loop exits because of the cap without ever proving there are no later failure points.

Please run until the first attempt where `has_induced_failure == false`, or make the work bound explicit and assert the sweep actually reached that terminal attempt. For example:

```zig
var fail_index: usize = 0;
while (true) : (fail_index += 1) {
    var failing = testing.FailingAllocator.init(allocator, .{ .fail_index = fail_index });
    var attempt = validator.validateCandidates(failing.allocator(), candidates, validation_policy, cp);
    attempt.deinit(failing.allocator());
    if (!failing.has_induced_failure) break;
    try testing.expectEqual(failing.allocated_bytes, failing.freed_bytes);
}
```

The generated path/policy dimensions are already bounded, so this remains finite while actually satisfying the allocation-failure acceptance criterion.

src/pki/path_validator_tests.zig
```diff
@@ -3042,3 +3042,428 @@ test "policy mappings apply through an asserted anyPolicy" {
...
+        try testing.expect(lhs.eql(&rhs));
+    } else {
+        try testing.expect(actual == null);
+    }
+}
+
+fn expectSameVerdict(first: validator.ValidationResult, second: validator.ValidationResult) !void {
+    switch (first) {
+        .accepted => |mine| switch (second) {
+            .accepted => |other| {
```
Jaruso, line 3302: [P2] The rejected branch now compares every `ValidationFailure` field, but the accepted branch still does not compare the complete `ValidationResult`. `AcceptedPath` includes `policies`, and `PolicyResult` contains `authority_constrained`, `user_constrained`, and deterministic `resource_usage`; two identical runs could return the same path but different policy outputs and this oracle would still pass. #492's inherited property is deterministic **result** for identical bytes/policy/time, not just deterministic path selection/failure metadata.

Please compare the policy result as well. For example, after the accepted-path loop, compare `resource_usage` and both OID slices by length/order/value:

```zig
try testing.expectEqual(mine.policies.resource_usage, other.policies.resource_usage);
try testing.expectEqual(mine.policies.authority_constrained.len, other.policies.authority_constrained.len);
for (mine.policies.authority_constrained, other.policies.authority_constrained) |lhs, rhs| {
    try testing.expect(lhs.eql(&rhs));
}
try testing.expectEqual(mine.policies.user_constrained.len, other.policies.user_constrained.len);
for (mine.policies.user_constrained, other.policies.user_constrained) |lhs, rhs| {
    try testing.expect(lhs.eql(&rhs));
}
```

That makes replay cover the entire accepted semantic output just as the rejected branch now covers the entire failure value.

docs/CRYPTO_FUZZ_CONTRACT.md
```diff
@@ -641,3 +647,86 @@ behavior, and the legal/illegal initial-ClientHello `0x0301` window) into
...
...cy_oid`; the allocation-failure sweep runs to the first fail index past the validator's last allocation -- not to a fixed cap -- asserting at each induced failure that every allocated byte was released, so no partially owned accepted path escapes. A compile-time structural check proves `ValidationFailure` has no pointer-shaped field anywhere in its layout, so a verdict cannot carry certificate bytes or key material out of the validator. Two deterministic sub-oracles pin the boundaries the generator can reach but not fix. The validity window: `notBefore - 1` is `certificate_not_yet_valid`, `notBefore` and `notAfter` accept, `notAfter + 1` is `certificate_expired`. RFC 5280 path length: this needs a `leaf -> I1 -> I2 -> root` shape to be reachable at all -- with a single intermediate the validator counts CA certificates in an empty slice, so its `pathLenConstraint` can never be exceeded, and the anchor's own constraint is not a validation input -- so the generator emits zero, one, or two intermediates and the sub-oracle pins both branches in every case: `I2` with `pathLenConstraint=0` is `path_length_exceeded` at index 2, while the otherwise-identical `pathLenConstraint=1` accepts. |
```
Jaruso: [P3] This row still describes the randomized validator input as a “two- or three-certificate shape”, but the follow-up fix now intentionally generates zero, one, **or two** intermediates, i.e. 2/3/4-certificate paths. The same row later correctly explains the four-certificate `leaf -> I1 -> I2 -> root` case. Please update the earlier phrase to “two- to four-certificate shape” (or “zero to two intermediates”) so the documented fuzz contract matches the harness.

Jaruso (commented):
Follow-up review on `6044cd4d`:

I verified all six findings from the previous pass are addressed and resolved those threads. In particular, the DER `childReader` overflow is fixed in production code with checked addition and has both fuzz/property coverage and a deterministic regression; the X.509 ownership/model oracles now walk the relevant borrowed state and compare parsed payloads; the validator now exercises a four-certificate path-length boundary; failure OIDs participate in replay; and the allocation-failure sweep runs through the terminal non-failing allocation index. CI is green on this head.

I found one remaining acceptance-criteria gap in accepted-result determinism, plus one documentation mismatch. The determinism issue should be addressed before merge. GitHub does not permit this account to submit a formal request-changes review on its own PR, so this is posted as a COMMENT review.

src/pki/path_validator_tests.zig
```diff
@@ -3042,3 +3042,488 @@ test "policy mappings apply through an asserted anyPolicy" {
...
+    // pathLenConstraint on it can never be exceeded and RFC 5280 path-length
+    // enforcement is unreachable. The anchor's own constraint does not help
+    // either, since trust-anchor extensions are deliberately not validation
+    // inputs. Only a `leaf -> I1 -> I2 -> root` shape reaches the rule.
+    const intermediate_count = smith.index(3);
+    const leaf_window = windows[smith.index(windows.len)];
+    const intermediate_window = windows[smith.index(windows.len)];
+    const root_window = windows[smith.index(windows.len)];
+    // A signature forged with a key nobody in the chain holds.
+    const forged_leaf = smith.index(6) == 0;
```
Jaruso: [P2] #492 explicitly requires the bounded path builder/validator target to fuzz **“algorithm/signature failure classes.”** This Smith program currently varies only the signing key (`forged_leaf` / `forged_intermediate`), so among `signatureFailure`'s five validator outcomes it can reach only `.signature_invalid`. The production mapper also has `.signature_algorithm_unsupported`, `.signature_key_mismatch`, `.signature_malformed`, and `.issuer_public_key_malformed`; those are covered by older deterministic tests, but not by this fuzz target, so interactions with the generated path/policy/time state remain outside coverage-guided exploration.

This does **not** need to duplicate #376's provider-primitive malformed-key/signature fuzzing. You can mutate the already-parsed synthetic views just enough to drive the validator classification seam, then assert the exact expected `FailureReason` before replay. For example, make one Smith-chosen mutually-exclusive signature mode after the fixtures are built:

```zig
const signature_case = smith.index(5);
var expected_signature_failure: ?validator.FailureReason = null;

switch (signature_case) {
    0 => {}, // valid signature path
    1 => { // well-formed shape, wrong signer
        // use the existing issuer_key = 9 generation
        expected_signature_failure = .signature_invalid;
    },
    2 => { // malformed Ed25519 signature length
        fx.certs.items[0].signature_value.data =
            fx.certs.items[0].signature_value.data[0..63];
        expected_signature_failure = .signature_malformed;
    },
    3 => { // unsupported signature algorithm
        fx.certs.items[0].signature_algorithm.oid =
            try oid.ObjectIdentifier.fromComponents(&[_]u32{ 1, 2, 3, 4, 5 });
        expected_signature_failure = .signature_algorithm_unsupported;
    },
    4 => { // supported algorithm paired with the wrong issuer key type
        fx.certs.items[0].signature_algorithm.oid =
            try oid.ObjectIdentifier.fromComponents(&oid.well_known.ecdsa_with_sha256);
        fx.certs.items[0].signature_algorithm.parameters_raw = null;
        expected_signature_failure = .signature_key_mismatch;
    },
    else => unreachable,
}
```

A corresponding issuer-view mutation such as truncating the selected Ed25519 issuer's `subject_public_key_info.subject_public_key.data` to 31 bytes can cover `.issuer_public_key_malformed` as another arm. Because signature verification runs before validity/constraints/policy, these modes can have exact reason oracles even while the rest of the Smith-generated chain remains adversarial. Please add at least the algorithm + malformed/key classification arms so this target actually satisfies the issue's stated fuzz matrix.

Jaruso (commented):
Third-pass review on the current head after the follow-up fixes and merge from `main`.

I verified the two outstanding findings from the prior review are fixed and resolved those threads: accepted-result replay now compares the complete RFC 9618 policy result (both constrained OID slices plus `resource_usage`), and the fuzz-contract documentation now accurately describes the two- through four-certificate path shapes. The merge-conflict resolution in `CRYPTO_FUZZ_CONTRACT.md` also preserved the newer #493 content while retaining the #492 section.

On a fresh acceptance-criteria pass against #492, I found one remaining gap in the path-validator fuzz target: #492 explicitly lists **algorithm/signature failure classes** among the cases the bounded synthetic-chain fuzzer must exercise. The Smith program currently only varies which Ed25519 key signs the certificate, so the verification layer can reach `signature_invalid`, but not the validator's other signature-classification outputs. Existing deterministic tests cover those mappings, but they are not driven by this fuzz target and therefore do not exercise their interactions with the generated path/policy/time state.

I left one inline P2 with a concrete way to add those cases without duplicating #376's provider-primitive fuzzing. I would address that before closing #492/merging this PR. Current CI is still running on the post-merge head; all completed jobs I checked are green so far.

docs/CRYPTO_FUZZ_CONTRACT.md
```diff
@@ -801,3 +807,89 @@ legal/illegal initial-ClientHello `0x0301` window in #493-A, and epoch
...
+zig build test-pki-fuzz -Doptimize=ReleaseFast -Dpki-test-filter="fuzz: PKI: path building terminates inside its budgets and is deterministic on adversarial graphs" --fuzz=10M --summary all --error-style verbose
+zig build test-pki-fuzz -Doptimize=ReleaseFast -Dpki-test-filter="fuzz: PKI: path validation is deterministic, bounded, and sanitized on adversarial chains" --fuzz=10M --summary all --error-style verbose
+
+# Named deterministic regressions (outside the "fuzz: PKI:" namespace —
+# filter on their own name, or run the whole `test-pki` step):
+zig build test-pki-fuzz -Dpki-test-filter="a validation failure carries only reason and stage metadata, never borrowed memory" --summary all --error-style verbose
+zig build test-pki-fuzz -Dpki-test-filter="childReader rejects an offset and length whose sum overflows machine width" --summary all --error-style verbose
+zig build test-pki -Dpki-test-filter="validation classifies every algorithm and signature failure class" --summary all --error-style verbose
```
Jaruso: [P3] These two new “filter on their own name” commands use `test-pki`, but `-Dpki-test-filter` is only applied to the separate `pki_fuzz_tests` artifact behind `test-pki-fuzz`; the ordinary `test-pki` step is built unfiltered (`b.addTest(.{ .root_module = pki_mod })`). So these commands silently run the entire PKI suite instead of the named test.

Please keep the same step used by the two examples immediately above:

```bash
zig build test-pki-fuzz -Dpki-test-filter="validation classifies every algorithm and signature failure class" --summary all --error-style verbose
zig build test-pki-fuzz -Dpki-test-filter="validation pins the validity window and pathLenConstraint boundaries" --summary all --error-style verbose
```

Alternatively, the docs could simply say to run `zig build test-pki` with no filter. This is documentation-only; the tests themselves and the fuzz wiring are correct.

Jaruso (commented):
Follow-up review on `2a17df98`: the remaining P2 algorithm/signature-classification gap is fully addressed, and I resolved that thread. The new `SignatureCase` axis reaches all five `signatureFailure` mappings, the named deterministic test proves every class is actually reachable, and the fuzz oracle is consistent with `validatePath` ordering. CI is green on this head.

I found one small documentation-only issue in the newly-added commands: the two deterministic examples use `test-pki` with `-Dpki-test-filter`, but that filter is only wired into the `test-pki-fuzz` artifact. I left one P3 inline comment with the correction. I found no remaining implementation or acceptance-criteria blocker in the code itself.

PR #584, opened 2026-08-08
security: [493-B] fuzz TLS record protection and epoch/key lifecycle

Jaruso:
## Summary

Second of #493's three planned slices (per the issue's implementation-plan comment): **record protection** and **epoch/key lifecycle**. Adds two `std.testing.Smith` targets under the existing `test-tls-record-fuzz` step that #493-A wired up, plus named deterministic companions for the arms CI's seed-corpus replay cannot reliably reach.

**This slice carries a production fix.** Review of the epoch oracle uncovered a real lifecycle defect in `record_epoch_bridge.zig`: **teardown was not terminal**, and the reference model agreed with the bug rather than detecting it.

The record contract has two session-teardown paths, and both reset `handshake_complete` to false — which was the *only* condition `installTrafficSecret(.zero_rtt, ...)` checked:

1. `Bridge.deinit()`, the unconditional wipe for an abandoned or failed handshake;
2. `discardEpoch(.application)`, the **orderly** teardown of a completed session, which that function's own documentation calls session teardown.

After either one, a bridge could have fresh 0-RTT read and write keys installed and resume sealing and opening records on that epoch. Both were confirmed with throwaway repros before fixing (seal → open → plaintext recovered), and both are now closed by a `torn_down` flag that every teardown path sets and nothing clears; `installTrafficSecret` rejects every epoch and direction once set. 0-RTT *discard* is deliberately unaffected — it is epoch-scoped, not session teardown. No production caller reused a bridge across either path, so this only closes the resurrection paths.

The record-protection target itself found no defect.

### `fuzz: TLS record: protection tamper and sequence boundaries preserve authentication state`

In `src/tls/record_protection.zig`.

- Classifies every `WriteState.seal` outcome **before** the call (`expectedSeal`, mirroring `seal`'s documented rejection order: exhausted sequence → `encodeInnerPlaintext`'s stages → sealed payload size → output capacity) and asserts that exact typed error, with the destination sentinel and write sequence re-verified after each rejection.
- Matrices over suite, traffic-secret length (exact-minus-one/exact/exact-plus-one per hash), content, padding, output capacity, and initial sequence (`{0, 1, 255, 256, maxInt-1, maxInt}`). Padding carries the size boundaries because it is a scalar the encoder zero-fills, so the exact-maximum sealed payload, one past it, and a synthetic `maxInt(usize)` need no oversized content buffer.
- On the accepting path the serialized header (what the AEAD authenticates as associated data) is re-derived independently and compared field by field; a write at `maxInt(u64)` is proven usable exactly once before the state latches exhausted; returned views are checked for containment in their backing buffer.
- Tamper battery against **one shared read state**: ciphertext bit flip, tag bit flip, tag truncation at every representative boundary, associated-data content-type and version mutation, wrong read sequence, wrong traffic secret, wrong suite-derived keys from the same secret, authenticated-but-malformed all-zero inner plaintext, and an exact-minus-one output buffer. Each must fail with its exact typed error and leave sequence/exhaustion untouched; failures reaching the AEAD must leave the attempted plaintext span **zeroed**, while validation/preflight failures must leave the buffer **entirely untouched**. Because they share one read state, the legitimate open that follows also proves a failed open leaves the state usable — and, once it succeeds, that replaying at the advanced sequence fails closed.
- A test-only `FaultProvider` wrapper makes the documented `UnsupportedCapability` (capability mask, caught by `TrafficKeys.derive`'s preflight) and `InvalidInput` (injected at the `aeadSeal`/`aeadOpen`/HKDF vtable seam) classes reachable deterministically, rather than inventing errors the record API cannot return. The fixture zeroes `plaintext` on a forced open failure so the "no unauthenticated plaintext" property is not proven against a fixture weaker than the interface it stands in for.
- Teardown is asserted directly on the public `FixedSecret` backing arrays: key/IV bytes zero, lengths zero, sequence reset, state exhausted, repeated teardown safe.

### `fuzz: TLS record: epoch operation sequences preserve one-way key lifecycle`

In `src/tls/record_epoch_bridge.zig`.

- Bounded programs of up to 48 operations — traffic-secret install at every epoch/direction with a length matrix, epoch discard, handshake completion, negotiated-suite updates including unknown code points, seal and open at each epoch, and `deinit` teardown at an arbitrary point followed by continued operation — driven against a `Bridge` and an independent `EpochModel` written from the documented transition rules rather than from the implementation.
- After **every** operation, accepted or rejected, a full `BridgeSnapshot` must equal the model's prediction: both direction phases, all three discard flags, completion, the negotiated suite, and for each of the six key slots an optional holding its sequence. Folding presence and sequence into one optional catches both a key that should have been wiped and a sequence that advanced when it should not have; the snapshot also cross-checks `hasReadKeys`/`hasWriteKeys` against the private slots at every step.
- The model predicts not just each gate but whether a record can authenticate at all — it tracks the suite each slot was installed under plus both sequence counters, and the target uses one fixed secret per epoch, so identical suite + identical sequence means byte-identical keys. A bridge that opened a record under the wrong epoch's keys, or consumed a read sequence number on a failed open, fails here.
- Covers the #408 transition classes as seeds: early discard, duplicate discard, missing direction, application install before handshake, partial completion, abandoned handshake, full progression, and teardown at every boundary.

TLS-over-TCP KeyUpdate/replacement remains #357 — the bridge exposes no such surface, so the operation set is documented as the extension point rather than a fabricated API.

### Docs

`docs/CRYPTO_FUZZ_CONTRACT.md` gains the #493-B narrative, the new filtered/long-run commands, a named-regression replay command, an explicit bounds table for all four #493 targets so far, and a statement that no #493 target uses network I/O, ambient entropy, real secrets, or external peers. `CHANGELOG.md` gains a Testing entry.

## Test plan

- [x] `zig build test-tls-record-fuzz --summary all` — 6/6
- [x] `zig build test-tls --summary all` — 915/915
- [x] `zig build test --summary all` — 3190/3191 (1 pre-existing skip)
- [x] `zig build test -Dtls-profile=appliance --summary all` — 3159/3160 (1 pre-existing skip)
- [x] Bounded `ReleaseFast` coverage-guided smoke, both new targets at `--fuzz=2M`, re-run after each production fix:
  - `protection tamper and sequence boundaries…` → 2 200 294 runs, clean
  - `epoch operation sequences…` → 2 000 286 runs after fix 1, and re-run again after fix 2, clean
- [x] **Both production defects confirmed by throwaway repro before fixing** — deinit (or application discard) → reinstall both 0-RTT directions → seal → open → plaintext recovered. Each repro passed against the unfixed bridge, then was deleted and replaced by a named regression.
- [x] **Oracle bug found by the coverage-guided run**, not the corpus: the epoch model did not predict `ReadState.open`'s `InvalidRecordType` preflight, so replaying an initial-epoch *plaintext* record at a protected epoch mispredicted `AuthenticationFailed`. The model now mirrors the full preflight order.
- [x] **Mutation-validated in both directions**, following #493-A's standard. Production-side mutations prove the tests catch the defect; oracle-side mutations prove the model cannot regress to mirroring the implementation:

| Mutation | Caught by |
| --- | --- |
| Drop `discardEpoch(.application)`'s `handshake_complete = false` | scripted companion (fuzz target in 16 runs) |
| Swap `seal`'s `RecordBufferOverflow` for `RecordTooLarge` | `seal classifies every rejection stage…` |
| Remove `installTrafficSecret`'s `torn_down` guard | both named teardown regressions (fuzz target in 17 runs) |
| Remove production `torn_down` on application discard only | `…after orderly application discard` + companion (fuzz target in 38 runs) |
| Remove the **model's** `torn_down` rule (either path) | scripted companion |

  The seed-corpus replay alone caught none of these, which is exactly why each has a named companion.

Ref #493 (closed by #493-C).

### DIFF
--- a/src/tls/record_epoch_bridge.zig
+++ b/src/tls/record_epoch_bridge.zig
@@ -1102,3 +1102,668 @@ test "record epoch bridge shuttles protocol-neutral driver events through record
     try testing.expectEqual(@as(u64, 1), server_bridge.write_application.?.sequence);
     try testing.expectEqual(@as(u64, 1), client_bridge.read_application.?.sequence);
 }
+
+// -----------------------------------------------------------------------
+// #493 epoch/key-lifecycle fuzz target. See docs/CRYPTO_FUZZ_CONTRACT.md's
+// "#493" section for the shared rules this follows (fresh provider and
+// bridge per case, fixed buffers, explicit operation bounds, typed-error
+// classification, sanitized diagnostics).
+// -----------------------------------------------------------------------
+
+const fuzz_epoch_max_operations = 48;
+const fuzz_epoch_content_len = 24;
+
+const all_epochs = [_]events.EncryptionEpoch{ .initial, .zero_rtt, .handshake, .application };
+const all_suites = [_]algorithms.CipherSuite{
+    .tls_aes_128_gcm_sha256,
+    .tls_aes_256_gcm_sha384,
+    .tls_chacha20_poly1305_sha256,
+};
+
+fn epochIndex(epoch: events.EncryptionEpoch) usize {
+    return switch (epoch) {
+        .initial => 0,
+        .zero_rtt => 1,
+        .handshake => 2,
+        .application => 3,
+    };
+}
+
+/// The traffic secret this target always uses for `epoch`, so a read slot and
+/// a write slot installed for the same epoch under the same suite hold
+/// byte-identical key material. That makes "does this record authenticate?"
+/// a function of the model's tracked suite and sequence alone, rather than of
+/// key bytes the model would otherwise have to reproduce.
+fn fuzzEpochSecret(epoch: events.EncryptionEpoch, len: usize, out: []u8) []const u8 {
+    const tag: u8 = 0x40 +% @as(u8, @intCast(epochIndex(epoch)));
+    for (out[0..len], 0..) |*byte, i| byte.* = tag ^ @as(u8, @truncate(i));
+    return out[0..len];
+}
+
+/// The externally observable state of a `Bridge`: direction phases, the
+/// discard/completion flags, the negotiated suite, and -- for each of the six
+/// key slots -- whether a key is installed and, if so, its sequence counter.
+/// Folding presence and sequence into one optional means a comparison catches
+/// both a key that should have been wiped and a sequence that advanced when
+/// it should not have.
+const BridgeSnapshot = struct {
+    cipher_suite: algorithms.CipherSuite,
+    read_phase: DirectionPhase,
+    write_phase: DirectionPhase,
+    read_handshake: ?u64,
+    write_handshake: ?u64,
+    read_zero_rtt: ?u64,
+    write_zero_rtt: ?u64,
+    read_application: ?u64,
+    write_application: ?u64,
+    initial_discarded: bool,
+    handshake_discarded: bool,
+    application_discarded: bool,
+    handshake_complete: bool,
+};
+
+fn bridgeSnapshot(bridge: *const Bridge) !BridgeSnapshot {
+    const snapshot = BridgeSnapshot{
+        .cipher_suite = bridge.cipher_suite,
+        .read_phase = bridge.read_phase,
+        .write_phase = bridge.write_phase,
+        .read_handshake = if (bridge.read_handshake) |state| state.sequence else null,
+        .write_handshake = if (bridge.write_handshake) |state| state.sequence else null,
+        .read_zero_rtt = if (bridge.read_zero_rtt) |state| state.sequence else null,
+        .write_zero_rtt = if (bridge.write_zero_rtt) |state| state.sequence else null,
+        .read_application = if (bridge.read_application) |state| state.sequence else null,
+        .write_application = if (bridge.write_application) |state| state.sequence else null,
+        .initial_discarded = bridge.initial_discarded,
+        .handshake_discarded = bridge.handshake_discarded,
+        .application_discarded = bridge.application_discarded,
+        .handshake_complete = bridge.handshake_complete,
+    };
+    // The public key-presence predicates must agree with the private slots
+    // they report on, at every point in the program.
+    try testing.expectEqual(snapshot.read_handshake != null, bridge.hasReadKeys(.handshake));
+    try testing.expectEqual(snapshot.write_handshake != null, bridge.hasWriteKeys(.handshake));
+    try testing.expectEqual(snapshot.read_zero_rtt != null, bridge.hasReadKeys(.zero_rtt));
+    try testing.expectEqual(snapshot.write_zero_rtt != null, bridge.hasWriteKeys(.zero_rtt));
+    try testing.expectEqual(snapshot.read_application != null, bridge.hasReadKeys(.application));
+    try testing.expectEqual(snapshot.write_application != null, bridge.hasWriteKeys(.application));
+    // The initial epoch never has record-protection keys of its own.
+    try testing.expect(!bridge.hasReadKeys(.initial));
+    try testing.expect(!bridge.hasWriteKeys(.initial));
+    return snapshot;
+}
+
+fn expectSnapshotEqual(expected: BridgeSnapshot, actual: BridgeSnapshot) !void {
+    try testing.expectEqual(expected.cipher_suite, actual.cipher_suite);
+    try testing.expectEqual(expected.read_phase, actual.read_phase);
+    try testing.expectEqual(expected.write_phase, actual.write_phase);
+    try testing.expectEqual(expected.read_handshake, actual.read_handshake);
+    try testing.expectEqual(expected.write_handshake, actual.write_handshake);
+    try testing.expectEqual(expected.read_zero_rtt, actual.read_zero_rtt);
+    try testing.expectEqual(expected.write_zero_rtt, actual.write_zero_rtt);
+    try testing.expectEqual(expected.read_application, actual.read_application);
+    try testing.expectEqual(expected.write_application, actual.write_application);
+    try testing.expectEqual(expected.initial_discarded, actual.initial_discarded);
+    try testing.expectEqual(expected.handshake_discarded, actual.handshake_discarded);
+    try testing.expectEqual(expected.application_discarded, actual.application_discarded);
+    try testing.expectEqual(expected.handshake_complete, actual.handshake_complete);
+}
+
+/// A compact reference model of the epoch lifecycle, written directly from
+/// the documented transition rules rather than from the `Bridge` code, so a
+/// rule the implementation silently relaxes shows up as a disagreement.
+///
+/// Per slot it tracks the suite the key was installed under (which decides
+/// whether a sealed record can authenticate against it) and the sequence
+/// counter, in addition to mere presence.
+const EpochModel = struct {
+    cipher_suite: algorithms.CipherSuite,
+    read_phase: DirectionPhase = .initial,
+    write_phase: DirectionPhase = .initial,
+    read_suite: [4]?algorithms.CipherSuite = .{null} ** 4,
+    write_suite: [4]?algorithms.CipherSuite = .{null} ** 4,
+    read_sequence: [4]u64 = .{0} ** 4,
+    write_sequence: [4]u64 = .{0} ** 4,
+    initial_discarded: bool = false,
+    handshake_discarded: bool = false,
+    application_discarded: bool = false,
+    handshake_complete: bool = false,
+
+    fn snapshot(self: *const EpochModel) BridgeSnapshot {
+        return .{
+            .cipher_suite = self.cipher_suite,
+            .read_phase = self.read_phase,
+            .write_phase = self.write_phase,
+            .read_handshake = self.slotSequence(.read, .handshake),
+            .write_handshake = self.slotSequence(.write, .handshake),
+            .read_zero_rtt = self.slotSequence(.read, .zero_rtt),
+            .write_zero_rtt = self.slotSequence(.write, .zero_rtt),
+            .read_application = self.slotSequence(.read, .application),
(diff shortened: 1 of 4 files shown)


src/tls/record_epoch_bridge.zig
```diff
@@ -1102,3 +1102,668 @@ test "record epoch bridge shuttles protocol-neutral driver events through record
...
+                if (phase != .initial or self.handshake_discarded) return error.InvalidEpochTransition;
+            },
+            .application => {
+                const phase = if (direction == .read) self.read_phase else self.write_phase;
+                if (phase != .handshake or self.handshake_discarded or self.application_discarded) {
+                    return error.InvalidEpochTransition;
+                }
+            },
+            .zero_rtt => {
+                if (self.handshake_complete) return error.InvalidEpochTransition;
```
Jaruso, line 1333: **Blocking:** this rule makes the reference model reproduce the production teardown bug instead of detecting it. `applyTeardown()` sets `handshake_complete = false` and clears the slots, so after a `deinit` operation this branch predicts that a fresh `.zero_rtt` install is valid. `Bridge.installTrafficSecret(.zero_rtt, ...)` has the same gate, which means a fuzz program can deinit, reinstall 0-RTT read/write keys, and then use that epoch again while model and implementation still compare equal. That violates this target's stated one-way/terminal teardown property. Please give the model an independent terminal state (and fix the production bridge to reject installs after teardown), then add a named post-deinit 0-RTT regression; otherwise the operation that can actually revive the bridge is precisely the one the oracle blesses.

Jaruso (commented):
Blocking finding: the epoch lifecycle oracle currently agrees with a real post-teardown resurrection path, so I would not merge this yet even though CI is green.

After `Bridge.deinit()`, production clears all key slots and marks the initial/handshake/application discard flags, but `installTrafficSecret(.zero_rtt, ...)` only gates on `handshake_complete`. Because `deinit()` resets `handshake_complete = false`, a fresh 0-RTT key can be installed after teardown. `EpochModel.predictInstall(.zero_rtt)` has the same rule, so the fuzz target accepts the resurrection instead of finding it. Once both 0-RTT directions are reinstalled, `sealProtected(.zero_rtt, ...)` / `openProtected(.zero_rtt, ...)` become usable again.

That conflicts directly with #493-B's implementation-plan requirement that teardown at an arbitrary point followed by continued operation remain terminal, and with this PR's own scripted-test comment that "nothing can revive the bridge." The scripted companion currently tries only handshake install/discard/complete after teardown, so it never exercises the one epoch that can actually come back.

Please make teardown terminal in the independent model **and** in `Bridge`, then pin the exact regression deterministically. For example:

```zig
// Model: do not derive this rule from the current Bridge gates.
torn_down: bool = false,

fn predictInstall(self: *const EpochModel, epoch: events.EncryptionEpoch, ...) ?Error {
    if (self.torn_down) return error.InvalidEpochTransition;
    ...
}

fn applyTeardown(self: *EpochModel) void {
    ...
    self.torn_down = true;
}
```

Production should have an equivalent terminal guard (or another explicit state that makes every traffic-secret install after `deinit()` reject). Then add a named regression specifically attempting both 0-RTT directions after teardown:

```zig
test "record epoch bridge cannot reinstall zero-rtt keys after teardown" {
    const cp = testProvider();
    var bridge = Bridge.init(cp, .tls_aes_128_gcm_sha256);
    const early = secret(0x91);

    bridge.deinit();

    try testing.expectError(
        error.InvalidEpochTransition,
        bridge.installTrafficSecret(.zero_rtt, .read, &early),
    );
    try testing.expectError(
        error.InvalidEpochTransition,
        bridge.installTrafficSecret(.zero_rtt, .write, &early),
    );
    try testing.expect(!bridge.hasReadKeys(.zero_rtt));
    try testing.expect(!bridge.hasWriteKeys(.zero_rtt));
}
```

I would also add the same post-teardown 0-RTT attempts to the scripted `EpochModel` companion so mutation validation proves the oracle itself cannot regress back to mirroring this implementation hole.

Aside from this blocker, the record-protection target is strong: exact typed rejection classification, sequence-exhaustion boundaries, shared-state tamper battery, provider-error injection, direct zeroization assertions, and the deterministic companion tests all line up well with the #493-B plan. CI for head `e9ddd354` is green across the repository workflow.

src/tls/record_epoch_bridge.zig
```diff
@@ -109,6 +119,9 @@ pub const Bridge = struct {
         direction: events.SecretDirection,
         traffic_secret: []const u8,
     ) Error!void {
+        // Teardown is terminal for every epoch, including the ones whose
+        // ordinary gates `deinit` happens to leave open.
+        if (self.torn_down) return error.InvalidEpochTransition;
```
Jaruso, line 133: **Blocking:** this guard only makes `deinit()` terminal, but `discardEpoch(.application)` is also documented above as **session teardown**. That branch clears application keys and sets `handshake_complete = false` without setting `torn_down`, so it reopens the 0-RTT gate here: complete session → `discardEpoch(.application)` → install fresh 0-RTT read/write secrets → seal/open again. `EpochModel.applyDiscard(.application)` currently mirrors the same state change, so the fuzz oracle blesses the resurrection just as it did for the original `deinit()` bug.

Please make orderly application discard enter the same terminal state (or otherwise independently forbid all later traffic-secret installs), and pin the exact window with a deterministic regression before calling `deinit()`. For example, after successful completion + application discard, both of these should fail and leave no keys installed:

```zig
try testing.expectError(error.InvalidEpochTransition,
    bridge.installTrafficSecret(.zero_rtt, .read, &early));
try testing.expectError(error.InvalidEpochTransition,
    bridge.installTrafficSecret(.zero_rtt, .write, &early));
try testing.expect(!bridge.hasReadKeys(.zero_rtt));
try testing.expect(!bridge.hasWriteKeys(.zero_rtt));
```

The model should state the same terminal policy independently and the scripted companion should attempt these installs immediately after `.application` discard, before `deinit`, so removing either production or oracle protection is mutation-detectable.

Jaruso (commented):
Re-review of current head `c7a6e9cc`.

The original post-`deinit()` 0-RTT resurrection finding is fixed correctly: `Bridge` now has an irreversible `torn_down` bit, the model carries an independent terminal rule, the snapshot compares it, and the named regression exercises both 0-RTT directions.

I found one remaining **blocking** lifecycle hole in the same policy: orderly `.application` discard is explicitly documented in `discardEpoch` as session teardown, but that branch does not set `torn_down`. It clears application keys and resets `handshake_complete = false`, which reopens the 0-RTT install path. The model's `.application` `applyDiscard` does the same, so a program can complete a session, `discardEpoch(.application)`, reinstall 0-RTT read/write keys, and use them while model and implementation still compare equal.

Please make orderly application discard terminal as well (preferably by setting the same terminal state in both production and the independent model), and add a named regression that attempts both 0-RTT installs **after application discard but before `deinit`**. The scripted model companion should exercise that exact window too, so mutation validation proves the oracle does not bless it.

The current CI failure on `Test (macos-14)` is unrelated infrastructure: Zig installation failed because `curl` could not resolve `ziglang.org`; Linux tests and the other completed jobs are green. CI is still in progress overall.

Minor follow-up: the PR description still says **“No production behavior changed”** and “the new properties found no defect,” which is now stale after the teardown fix and should be updated before merge.

Jaruso (commented):
Re-review at `c7a6e9cc`: no new commits have landed since my previous review, so the blocking application-discard lifecycle finding is unchanged.

`discardEpoch(.application)` is documented as orderly session teardown and clears the application keys, resets both phases, and sets `handshake_complete = false`, but it does not set `torn_down`. That leaves `installTrafficSecret(.zero_rtt, ...)` open again immediately after application discard. The reference model mirrors the same transition, so this resurrection path still compares equal instead of failing the property.

The prior inline comment has the requested fix and deterministic regression shape; I am not duplicating a second inline thread for the same defect. The earlier `deinit()` resurrection path is fixed correctly.

CI for this head is now green. This PR remains blocked on the unresolved application-discard/0-RTT lifecycle issue. Also, the PR summary is still stale in saying no production behavior changed / no defect was found, since the previous review already resulted in a production lifecycle fix.

Jaruso (commented):
Re-review of `f9cc4837af8ace7efcff96c5dd0fe68a70a137ff`: no remaining blockers found.

The follow-up correctly closes the second teardown resurrection path: `discardEpoch(.application)` now enters the same irreversible `torn_down` state as `deinit()`, so resetting `handshake_complete` can no longer reopen 0-RTT traffic-secret installation. The `EpochModel` states that policy independently, and the scripted companion exercises the vulnerable window after orderly application discard but before `deinit()`, so production and oracle regressions are mutation-detectable separately. The dedicated deterministic regression also checks both 0-RTT directions, absence of reinstalled keys, unusability of the torn-down epochs, and preservation of the intentionally idempotent/non-terminal 0-RTT discard behavior.

I re-checked this against the #493-B lifecycle requirements and the previously reviewed record-protection half; I do not see another acceptance-criteria gap or lifecycle escape introduced by the fix. The PR description/docs have also been updated to acknowledge the production defects rather than claiming no behavior change.

Current-head CI run 31238177227 is green, including Linux/macOS/ARM unit tests, appliance-profile tests, format/lint/security, ReleaseSafe/ReleaseFast builds, packaging/perf smoke, and H3 resumption/0-RTT interop.

No changes requested.

Jaruso (commented):
Post-merge verification: no new findings. PR #584 was merged at the same `f9cc4837` head from my last clean review. I compared the final merge commit (`5b64eed8`) against that reviewed head for every PR-touched file (`src/tls/record_epoch_bridge.zig`, `src/tls/record_protection.zig`, `docs/CRYPTO_FUZZ_CONTRACT.md`, and `CHANGELOG.md`); all four blob SHAs are identical, so the merge introduced no conflict-resolution or content delta. Both previously reported teardown/0-RTT lifecycle defects remain fixed, the #493-B acceptance scope remains satisfied, and the last current-head CI run was green. No changes requested.

PR #587, opened 2026-08-08
security: [493-C] fuzz encrypted-stream progression and terminal cleanup

Jaruso:
## Summary

Final slice of #493 (epic #325-K): the scripted in-memory carrier, the two
`encrypted_stream.zig` fuzz targets, seven named deterministic companions, and
the docs closeout. Follows #493-A and #493-B under the same
`test-tls-record-fuzz` step and `-Dtls-record-test-filter` namespace.

- **`ScriptedCarrier`** — a private, test-only `Carrier` built entirely from
  fixed arrays: bounded inbound queue, capture buffer, independent cyclic read
  and write action scripts (`transfer at most N`, `WouldBlock`, zero-byte
  progress, `EndOfStream`, one typed carrier error per direction), and
  read/write/close counters. No fuzz callback opens a socket, and neither
  target calls the module's existing socket-pair helpers. A `.transfer` with
  nothing to move reports `WouldBlock` rather than zero, because a zero-byte
  carrier *read* is EOF to `drive()`; zero-byte writes are scripted separately.

- **`fuzz: TLS record: encrypted stream scripted carrier progression preserves
  bytes and terminal state`** — bounded programs (≤24 ops) of drive / read /
  write / close / enqueue / repeated-drive / teardown against a real
  `PureZigRecordStream` pair. The peer is a full record stream, so the oracle
  is checked against the real record path. Both application byte streams are
  *generated*, so "no loss or duplication" is "the nth delivered byte equals
  f(n)" plus a monotone delivered-≤-accepted bound — which keeps a 16 KiB
  fragment write expressible without a 16 KiB oracle buffer.

  Around every `drive()`: exact outbound byte conservation
  (`queued_before == queued_after + bytes the carrier accepted`, so a partial
  write discards only the written prefix); the per-drive carrier byte budgets;
  a calls-vs-bytes bound that rules out spinning; progress is *real* (a drive
  claiming progress moved a byte or changed observable state, and a
  no-progress drive moved nothing); and reported readiness matches the
  stream's. Per operation: queue watermarks and fixed capacities,
  plaintext/provenance agreement, at-most-once owned-carrier release, and
  exact backpressure alternation (`pauses - resumes == (paused ? 1 : 0)`).
  Plaintext-I/O errors are classified, not swallowed: an un-latched stream may
  only refuse I/O for one of the two deferred terminal conditions and must
  report it unchanged.

- **`fuzz: TLS record: encrypted stream cleanup preserves root errors across
  alerts and epoch transitions`** — four scenario families: deferred
  fatal-alert flush under partial / blocked / zero-progress / failing writes;
  a record authentication failure behind already-delivered genuine plaintext;
  a `.handshake` epoch discard landing on a partially buffered record; and
  teardown with every buffer, both parsers, carrier input, and a pending
  terminal alert populated at once. Every terminal outcome asserts each owned
  buffer is not merely empty but *zeroed* in its backing storage — both
  parsers' pending arrays and the plaintext provenance shadow included — and
  that no key material survives at any epoch in either direction.

## Production change

`fail()` and `deinit()` each release the handshake driver and wipe every bridge
key, but a stream that completed a clean `close_notify` exchange only dropped
its owned buffers and latched `.closed`. Two holders of secret-bearing state
survived until the caller happened to call `deinit()`:

1. the bridge's **application traffic keys**; and
2. the owned **handshake driver**, whose borrowed `EventSink` may still hold
   copied traffic-secret scratch until `Driver.deinit()` — a retention the
   transport contract documents explicitly.

Nothing could use either (`.closed` rejects every entry point), so this is
retention without a purpose rather than an exploitable path, but the record
contract's rule is that a torn-down session keeps no secret-bearing state. The
five orderly close-completion sites now share a `finishClose()` helper that
runs the same teardown order as `deinit()`/`fail()`: driver, queues, bridge.
`authStillPending()` gained the matching `driver_torn_down` guard, because
`drive()` can reach it later in the same call after an in-loop
`queueCloseNotify()` completed the close.

Found by this slice's teardown property; pinned by `encrypted stream close is
terminal and idempotent from every lifecycle state` and `encrypted stream
orderly close releases the handshake driver and its secret scratch`.

## Named companions

Per the #493-A/B standard, the scripted classes the seed replay cannot
reliably reach get named deterministic tests:

| Companion | Pins |
| --- | --- |
| `…delivers every byte across one-byte reads and writes` | full bidirectional exchange through single-byte transfers |
| `…settles without spinning under repeated would-block and zero-progress carriers` | 64 repetitions: no progress, no byte moved, queue unchanged, readiness stable, constant carrier work per drive |
| `…carrier EOF at every record boundary preserves truncation and delivery order` | EOF before a record, mid-header, mid-payload, one byte short, and exactly after a complete record; buffered plaintext delivered first, then the preserved `TruncatedStream` |
| `…partial carrier writes preserve the exact unwritten suffix` | prefix and suffix compared byte for byte on every drive, then reassembled and opened at the peer |
| `…output saturation pauses plaintext writes and resumes below the low watermark` | one pause per crossing, a rejected retry consuming nothing, one resume, no byte lost |
| `…inbound plaintext saturation pauses carrier reads and resumes after draining` | the same, read side |
| `…close is terminal and idempotent from every lifecycle state` | close during handshaking, open, closing, failed; carrier released once; the `finishClose` key wipe |
| `…orderly close releases the handshake driver and its secret scratch` | completed session + mid-handshake cancellation: driver/backend released exactly once *at close*, sink scratch zeroed, no second release on `deinit()` |

## Review fixes (second commit)

All five blockers from the review are addressed:

1. **Orderly close now tears down the driver too**, with a new regression
   covering both `finishClose` paths.
2. **The progression oracle no longer swallows peer errors** — only
   `WouldBlock` and the clean-`close_notify` branch are expected outcomes.
   Fixing that exposed a second gap: the seed corpus did not reliably reach the
   write-then-deliver path, so a corrupted-emission mutation still passed. Every
   case now carries a mandatory floor of real traffic in both directions, and a
   two-stage deterministic epilogue drains what the scripts obstructed (stage
   one write-side only, so every drive there is a write-only drive).
3. **`made_progress == false` now asserts** zero carrier bytes in either
   direction and `!state_changed`, not just the queue snapshot.
4. **The teardown family dirties every state it zeroizes** — both parsers
   simultaneously, inbound plaintext with provenance, handshake bytes, carrier
   input, outbound ciphertext, pending terminal alert — set up with `try` and
   asserted nonzero immediately before `deinit()`.
5. **The authentication-failure family is deterministic** and requires
   `AuthenticationFailed` rather than accepting "no terminal error happened".

## Mutation validation

| Mutation | Caught by |
| --- | --- |
| Drop `teardownDriver()` from `finishClose` | `…orderly close releases the handshake driver and its secret scratch` |
| Drop `bridge.deinit()` from `finishClose` | `encrypted stream close is terminal and idempotent…` |
| Subject emits one corrupted ciphertext byte toward the peer | progression target, seed replay |
| A carrier write is not reported as `made_progress` | progression target, seed replay (epilogue's write-only stage) |
| `drive()` latches `SocketWriteFailed` instead of the preserved root error | cleanup fuzz target, seed replay |
| `ByteQueue.clear` stops zeroing its backing storage | cleanup fuzz target + `byte queues wipe discarded and cleared storage` + `terminal cleanup clears parser-owned ciphertext and queued plaintext` |
| `drive()`'s write loop consumes one byte more than the carrier accepted | `…partial carrier writes preserve the exact unwritten suffix` + 4 existing close/handshake tests |

The corruption mutation has to be scoped to the subject's own role: corrupting
*both* directions ends the case early on a bad **inbound** record and never
reaches the outbound accounting — a reminder that a mutation which merely turns
some test red is not the same as one that proves the property under test bites.

## Test plan

- [x] `zig fmt --check build.zig src/ tests/` — clean
- [x] `zig build test-tls-record-fuzz --summary all` — 8/8 (deterministic corpus replay)
- [x] `zig build test-tls --summary all` — 925/925
- [x] `zig build test --summary all` — 3200/3201 (1 skipped)
- [x] `zig build test -Dtls-profile=appliance --summary all` — 3169/3170 (1 skipped)
- [x] Coverage-guided `ReleaseFast` smoke, progression target — **151 544 runs, 1 531 unique, clean**
      (`zig build test-tls-record-fuzz -Doptimize=ReleaseFast -Dtls-record-test-filter="fuzz: TLS record: encrypted stream scripted carrier progression preserves bytes and terminal state" --fuzz=150K`)
- [x] Coverage-guided `ReleaseFast` smoke, cleanup target — **250 208 runs, 196 unique, clean**
      (`zig build test-tls-record-fuzz -Doptimize=ReleaseFast -Dtls-record-test-filter="fuzz: TLS record: encrypted stream cleanup preserves root errors across alerts and epoch transitions" --fuzz=250K`)
- [x] Mutation validation, six mutations, all caught (table above)

No new CI step: the targets are inline under `tls_core_mod`, so their seed
corpora already replay under the existing `zig build test` job.

## Notes

- No `@embedFile` corpus fixtures were needed — the one defect this slice found
  is expressible as a named semantic regression, matching #493-A/B.
- TLS-over-TCP KeyUpdate/replacement remains #357; the stream targets' operation
  sets are documented as the extension point rather than a fabricated API.

Closes #493.

### DIFF
--- a/src/tls/encrypted_stream.zig
+++ b/src/tls/encrypted_stream.zig
@@ -756,12 +756,16 @@ pub const PureZigRecordStream = struct {
     }
 
     /// Complete an orderly close. Beyond dropping every owned buffer this also
-    /// wipes the bridge's key material: a `.closed` stream rejects every entry
-    /// point, so no record can ever be sealed or opened again, and leaving
-    /// application traffic keys resident until an eventual `deinit()` is
-    /// retention without a purpose. `fail()` already wipes them on the terminal
-    /// path; this is the same rule for the orderly one.
+    /// releases the two other holders of secret-bearing state -- the handshake
+    /// driver (whose borrowed `EventSink` may still hold copied traffic-secret
+    /// scratch until `Driver.deinit()`, per the transport contract) and the
+    /// bridge's key material. A `.closed` stream rejects every entry point, so
+    /// no record can ever be sealed or opened again and nothing here can be
+    /// needed again; holding either until an eventual outer `deinit()` is
+    /// retention without a purpose. This is the same teardown order `deinit()`
+    /// and `fail()` use, for the orderly path.
     fn finishClose(self: *PureZigRecordStream) void {
+        self.teardownDriver();
         self.clearOwnedQueues();
         self.bridge.deinit();
         self.lifecycle = .closed;
@@ -1767,7 +1771,11 @@ pub const PureZigRecordStream = struct {
         return @as(u64, plaintext_limit) + records * overhead_per_record;
     }
 
+    /// A torn-down driver is never polled again: `finishClose`/`fail`/`deinit`
+    /// release the backend, and `drive()` can reach this predicate later in the
+    /// same call after an in-loop `queueCloseNotify` completed the close.
     fn authStillPending(self: *const PureZigRecordStream) bool {
+        if (self.driver_torn_down) return false;
         if (self.handshake_driver) |*driver| return driver.authPending();
         return false;
     }
@@ -4229,6 +4237,9 @@ const ScriptedRecordBackend = struct {
     auth_pending: bool = false,
     auth_pending_polls: usize = 0,
     auth_poll_count: usize = 0,
+    /// How many times the stream released this backend. `Driver.deinit` is not
+    /// idempotent, so this must never exceed one.
+    deinit_count: usize = 0,
 
     const hs_c2s = secret(0x11);
     const hs_s2c = secret(0x22);
@@ -4236,7 +4247,12 @@ const ScriptedRecordBackend = struct {
     const app_s2c = secret(0x44);
 
     fn recordBackend(self: *ScriptedRecordBackend) RecordHandshakeBackend {
-        return .{ .ptr = self, .startFn = start, .receiveFn = receive, .authPendingFn = authPending, .resumeFn = resumeAuth };
+        return .{ .ptr = self, .startFn = start, .receiveFn = receive, .authPendingFn = authPending, .resumeFn = resumeAuth, .deinitFn = backendDeinit };
+    }
+
+    fn backendDeinit(ptr: *anyopaque) void {
+        const self: *ScriptedRecordBackend = @ptrCast(@alignCast(ptr));
+        self.deinit_count += 1;
     }
 
     fn start(ptr: *anyopaque, role: tls_state.Role, _: void, sink: *RecordTransport.EventSink) RecordHandshakeError!void {
@@ -5042,8 +5058,12 @@ const StreamOracle = struct {
     peer_sent: usize = 0,
     /// Plaintext bytes the subject delivered to its application.
     subject_received: usize = 0,
-    /// False once the peer saw `close_notify` or failed: it then swallows
-    /// further input, so it can no longer witness the subject's byte stream.
+    /// False once the peer has seen the subject's `close_notify`: from then on
+    /// it swallows further input by contract, so it can no longer witness the
+    /// subject's byte stream. This is the *only* thing that retires the peer --
+    /// a record the peer cannot open or parse means the subject emitted bytes
+    /// the real record path rejects, which is precisely the corruption this
+    /// target exists to catch, so those errors propagate and fail the case.
     peer_usable: bool = true,
     /// The subject's latched terminal error, once one exists.
     terminal: ?anyerror = null,
@@ -5056,11 +5076,11 @@ fn drainPeerPlaintext(peer: *PureZigRecordStream, oracle: *StreamOracle) !void {
     var rounds: usize = 0;
     while (rounds < fuzz_stream_sync_rounds) : (rounds += 1) {
         const n = peer.readPlaintext(&buf) catch |err| switch (err) {
+            // Nothing buffered, or a peer that has already seen `close_notify`.
             error.WouldBlock, error.EndOfStream => return,
-            else => {
-                oracle.peer_usable = false;
-                return;
-            },
+            // Anything else means the plaintext the subject produced could not
+            // be delivered by the real record path.
+            else => return err,
         };
         if (n == 0) return;
         for (buf[0..n], 0..) |byte, i| {
@@ -5083,13 +5103,14 @@ fn syncCapture(carrier: *ScriptedCarrier, peer: *PureZigRecordStream, oracle: *S
     }
     var rounds: usize = 0;
     while (carrier.captured_len > 0 and rounds < fuzz_stream_sync_rounds) : (rounds += 1) {
+        // Only `WouldBlock` is an expected outcome here: it is the peer's own
+        // buffer pressure, and the same bytes are retried next round. Every
+        // other error means the subject emitted ciphertext the real record path
+        // could not authenticate or frame, so it must fail the case rather than
+        // quietly retire the oracle.
         const consumed = peer.feedCiphertext(carrier.captureSlice()) catch |err| switch (err) {
             error.WouldBlock => 0,
-            else => {
-                oracle.peer_usable = false;
-                carrier.captured_len = 0;
-                return;
-            },
+            else => return err,
         };
         if (consumed > 0) carrier.consumeCapture(consumed);
         try drainPeerPlaintext(peer, oracle);
@@ -5124,18 +5145,17 @@ fn enqueuePeerRecord(
 
     var chunk: [fuzz_stream_peer_chunk]u8 = undefined;
     for (chunk[0..len], 0..) |*byte, i| byte.* = peerStreamByte(oracle.peer_sent + i);
+    // `WouldBlock` is the peer's own outbound backpressure and simply means
+    // "not this round"; anything else is a real defect in the record path.
     const n = peer.writePlaintext(chunk[0..len]) catch |err| switch (err) {
         error.WouldBlock => return,
-        else => {
-            oracle.peer_usable = false;
-            return;
-        },
+        else => return err,
     };
 
     var buf: [256]u8 = undefined;
     while (peer.queuedCiphertextLen() > 0) {
-        const moved = peer.drainCiphertext(&buf) catch break;
-        if (moved == 0) break;
+        const moved = try peer.drainCiphertext(&buf);
+        try testing.expect(moved > 0);
         try testing.expect(carrier.pushInbound(buf[0..moved]));
     }
     oracle.peer_sent += n;
@@ -5246,8 +5266,13 @@ fn driveChecked(
     if (result.made_progress) {
         try testing.expect(read_bytes + write_bytes > 0 or state_changed);
     } else {
-        // ...and a drive that reports none must not have moved a byte in or
-        // out of any owned queue.
+        // ...and a drive that reports none moved no carrier byte, changed no
+        // observable state, and left every owned queue exactly as it was. This
+        // is the full "no spin under repeated zero-progress readiness"
+        // property, not just one projection of it.
+        try testing.expectEqual(@as(usize, 0), read_bytes);
+        try testing.expectEqual(@as(usize, 0), write_bytes);
+        try testing.expect(!state_changed);
         try testing.expect(std.meta.eql(queues_before, subject.bufferSnapshot().current));
     }
     return result;
@@ -5414,9 +5439,17 @@ fn fuzzEncryptedStreamProgressionInput(_: void, smith: *std.testing.Smith) !void
     var write_src: [record_codec.max_plaintext_fragment_len]u8 = undefined;
     var read_buf: [fuzz_stream_max_read_buf]u8 = undefined;
 
-    for (0..smith.index(4)) |_| {
+    // Every case carries real bytes in both directions before the fuzzer's
+    // program starts: one mandatory inbound record and one mandatory
+    // application write, plus a fuzzer-chosen number of extra inbound records.
+    // Without that floor a program whose scripts never move a byte would reach
+    // the accounting assertions with nothing to account for.
+    for (0..1 + smith.index(3)) |_| {
         try enqueuePeerRecord(&peer, &carrier_state, &oracle, 1 + smith.index(fuzz_stream_peer_chunk));
     }
+    try subjectWriteOnce(&subject, &oracle, write_src[0 .. 1 + smith.index(64)]);
+    try testing.expect(oracle.peer_sent > 0);
+    try testing.expect(oracle.subject_sent > 0);
     try checkStreamInvariants(&subject, &carrier_state);
 
     var torn_down = false;
@@ -5508,6 +5541,52 @@ fn fuzzEncryptedStreamProgressionInput(_: void, smith: *std.testing.Smith) !void
         try checkStreamInvariants(&subject, &carrier_state);
         if (oracle.terminal != null) break;
         if (!result.made_progress and oracle.subject_received == before_received) break;
+    }
+
+    if (oracle.terminal) |err| {
+        try expectStableTerminal(&subject, &carrier_state, err);
+        if (carrier_state.owns_handle) try testing.expectEqual(@as(usize, 1), carrier_state.close_calls);
+        return;
+    }
+
+    // Deterministic epilogue. The scripts have already done their fragmentation
+    // work; lifting the obstruction lets everything still queued or in flight
+    // drain, so the byte accounting below is actually reached instead of being
+    // skipped for any case whose script happened never to let a byte through.
+    // This is what makes the seed-corpus replay a real end-to-end check on
+    // every record the subject emitted and every record it consumed.
+    if (subject.lifecycle != .closed and !subject.carrier_eof) {
+        // Stage one opens the write side only. Every drive here is a
+        // *write-only* drive, so the carrier bytes it moves are the sole thing
+        // that can justify `made_progress` -- which is what makes
+        // `driveChecked`'s no-progress branch a real check rather than one a
+        // concurrent read could satisfy on its behalf.
+        carrier_state.read_script_len = 1;
+        carrier_state.read_script[0] = .would_block;
+        carrier_state.write_script_len = 1;
+        carrier_state.write_script[0] = .{ .transfer = scripted_capture_capacity };
+        var drain_rounds: usize = 0;
+        while (oracle.terminal == null and drain_rounds < fuzz_stream_flush_rounds) : (drain_rounds += 1) {
+            const result = try driveChecked(&subject, &carrier_state, &oracle) orelse break;
+            try syncCapture(&carrier_state, &peer, &oracle);
+            try checkStreamInvariants(&subject, &carrier_state);
+            if (oracle.terminal != null or !result.made_progress) break;
+        }
+        try testing.expect(drain_rounds < fuzz_stream_flush_rounds);
+
+        // Stage two opens the read side too and settles the whole stream.
+        carrier_state.read_script[0] = .{ .transfer = scripted_inbound_capacity };
+        var epilogue_rounds: usize = 0;
+        while (oracle.terminal == null and epilogue_rounds < fuzz_stream_flush_rounds) : (epilogue_rounds += 1) {
+            const before_received = oracle.subject_received;
+            const result = try driveChecked(&subject, &carrier_state, &oracle) orelse break;
+            try syncCapture(&carrier_state, &peer, &oracle);
+            try subjectReadOnce(&subject, &oracle, &read_buf);
+            try checkStreamInvariants(&subject, &carrier_state);
+            if (oracle.terminal != null) break;
+            if (!result.made_progress and oracle.subject_received == before_received) break;
+        }
+        try testing.expect(epilogue_rounds < fuzz_stream_flush_rounds);
     }
 
     if (oracle.terminal) |err| {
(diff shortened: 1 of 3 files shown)


src/tls/encrypted_stream.zig
```diff
@@ -755,6 +755,18 @@ pub const PureZigRecordStream = struct {
...
 
+    /// Complete an orderly close. Beyond dropping every owned buffer this also
+    /// wipes the bridge's key material: a `.closed` stream rejects every entry
+    /// point, so no record can ever be sealed or opened again, and leaving
+    /// application traffic keys resident until an eventual `deinit()` is
+    /// retention without a purpose. `fail()` already wipes them on the terminal
+    /// path; this is the same rule for the orderly one.
+    fn finishClose(self: *PureZigRecordStream) void {
+        self.clearOwnedQueues();
+        self.bridge.deinit();
```
Jaruso, line 770: [P1] **Tear down the driver on orderly close too.** `finishClose()` now wipes `Bridge`, but a production stream created with `initWithCarrierAndBackend` can still own `handshake_driver`. `Driver.complete()` does not tear it down, and `tls/transport.zig` explicitly documents that `EventSink` may retain copied traffic-secret scratch until `Driver.deinit()`/`EventSink.reset()`. So a clean close can still retain secret-bearing driver/backend state until some later outer `PureZigRecordStream.deinit()`—the same retention class this helper is meant to eliminate.

Please use the same ordering already used by `deinit()`/`fail()`:
```zig
fn finishClose(self: *PureZigRecordStream) void {
    self.teardownDriver();
    self.clearOwnedQueues();
    self.bridge.deinit();
    self.lifecycle = .closed;
}
```
Then add a regression that closes a stream with a real or instrumented handshake backend (not `CompanionPair`, which has no driver), and assert driver/backend teardown occurs exactly once and the sink's used secret scratch is zeroed. `teardownDriver()` is already idempotence-guarded, so a later `deinit()` remains safe.

src/tls/encrypted_stream.zig
```diff
@@ -4801,3 +4826,1455 @@ test "driver-owned handshake latches on the flush deadline when a fatal alert ca
...
+    if (!oracle.peer_usable) {
+        carrier.captured_len = 0;
+        return;
+    }
+    var rounds: usize = 0;
+    while (carrier.captured_len > 0 and rounds < fuzz_stream_sync_rounds) : (rounds += 1) {
+        const consumed = peer.feedCiphertext(carrier.captureSlice()) catch |err| switch (err) {
+            error.WouldBlock => 0,
+            else => {
+                oracle.peer_usable = false;
```
Jaruso: [P1] **Do not disable the oracle on unexpected peer errors.** This peer is the validator for ciphertext produced by the subject. If `feedCiphertext()` returns `AuthenticationFailed`, `InvalidRecordType`, etc., that means the subject emitted bytes that the real record path could not consume. Marking `peer_usable = false`, dropping `captured_len`, and returning makes that corruption disappear; the final `subject_sent == peer_received` assertion is then skipped because `peer_usable` is false.

Only the explicit clean-`close_notify` branch below should make the peer unusable. Unexpected record errors should fail the fuzz case, for example:
```zig
const consumed = peer.feedCiphertext(carrier.captureSlice()) catch |err| switch (err) {
    error.WouldBlock => 0,
    else => return err,
};
```
Please apply the same rule to `drainPeerPlaintext()` and `enqueuePeerRecord()`—their unexpected errors/`catch break` paths should be propagated or explicitly classified, not swallowed. This is also required by #493's shared rule to classify expected failures by exact typed error.

src/tls/encrypted_stream.zig
```diff
@@ -4801,3 +4826,1455 @@ test "driver-owned handshake latches on the flush deadline when a fatal alert ca
...
+        carrier_eof_before != subject.carrier_eof or
+        peer_closed_before != subject.peer_closed or
+        !std.meta.eql(deferred_before, subject.pending_terminal_read_error) or
+        !std.meta.eql(queues_before, subject.bufferSnapshot().current);
+    if (result.made_progress) {
+        try testing.expect(read_bytes + write_bytes > 0 or state_changed);
+    } else {
+        // ...and a drive that reports none must not have moved a byte in or
+        // out of any owned queue.
+        try testing.expect(std.meta.eql(queues_before, subject.bufferSnapshot().current));
```
Jaruso, line 5276: [P2] **The no-progress branch does not prove that nothing moved.** The PR/docs say a drive reporting `made_progress == false` moved no bytes and changed no observable state, but this branch only compares the queue snapshot. A drive could move carrier bytes while ending with the same queue lengths, or change lifecycle/EOF/peer-close/deferred-error state, and this would still pass.

You already computed everything needed above; assert it here:
```zig
} else {
    try testing.expectEqual(@as(usize, 0), read_bytes);
    try testing.expectEqual(@as(usize, 0), write_bytes);
    try testing.expect(!state_changed);
}
```
Keeping the queue equality as an additional assertion is fine. This closes the actual `no spin under repeated zero-progress readiness` property instead of only checking one projection of it.

src/tls/encrypted_stream.zig
```diff
@@ -4801,3 +4826,1455 @@ test "driver-owned handshake latches on the flush deadline when a fatal alert ca
...
+    // Carrier input the read budget has not reached yet.
+    var filler: [128]u8 = undefined;
+    for (&filler, 0..) |*byte, i| byte.* = @truncate(i);
+    _ = carrier_state.pushInbound(&filler);
+    subject.inbound_carrier.append(&filler) catch {};
+
+    // ...and a pending terminal alert flush that the blocked carrier is holding.
+    if (smith.index(2) == 0) subject.deferHandshakeFailure(error.CertificateInvalid, null);
+
+    subject.deinit();
```
Jaruso, line 6016: [P2] **Dirty every state you claim to zeroize before calling `deinit()`.** This scenario is documented as teardown with *every owned buffer, both parsers, and a pending terminal state populated at once*, but `initial_parser` is never populated, `inbound_plaintext`/its provenance shadow are never populated, and `pending_terminal` exists only in half the cases. Several setup operations also use `catch {}` / `catch break`, so even the states you intend to dirty can silently remain empty. `expectTeardownMetadataCleared()` can therefore pass vacuously for untouched backing arrays.

Please make the setup deterministic (`try` rather than swallowing setup failures) and assert the preconditions immediately before teardown, e.g. nonzero inbound carrier/plaintext/provenance/handshake/outbound queues, nonzero data in both parser backing states, and `pending_terminal != null`. If normal lifecycle rules cannot leave both parsers dirty simultaneously, direct same-file test setup is appropriate—the property under test here is that teardown zeroes *nonzero* owned storage. Then call `deinit()` and perform the existing zeroization assertions.

src/tls/encrypted_stream.zig
```diff
@@ -4801,3 +4826,1455 @@ test "driver-owned handshake latches on the flush deadline when a fatal alert ca
...
+        try subjectReadOnce(&subject, &oracle, &read_buf);
+        try checkStreamInvariants(&subject, &carrier_state);
+        if (oracle.terminal != null) break;
+        if (!result.made_progress and carrier_state.inboundPending() == 0) break;
+    }
+    try testing.expect(drives < 256);
+
+    // The tampered record reached the bridge only if the carrier script let it
+    // through; when it did, the failure is exactly `AuthenticationFailed`.
+    if (oracle.terminal) |err| {
```
Jaruso: [P2] **Require this scenario to actually reach `AuthenticationFailed`.** The setup above has several `catch return`/early-return paths, and the assertion here is conditional on `oracle.terminal` being non-null. That means a fuzz input can select the authentication-failure family, fail to construct/deliver the tampered record, and still count as a successful case without testing the property named by the function.

This setup should be deterministic once selected. Prefer something like:
```zig
_ = try peer.writePlaintext(content[0..content_len]);
const record_len = try peer.drainCiphertext(&record_buf);
try testing.expect(record_len > record_codec.header_len);
try testing.expect(carrier_state.pushInbound(record_buf[0..record_len]));
// drive until the forced first transfer reaches the tampered record...
try testing.expectEqual(
    @as(?anyerror, error.AuthenticationFailed),
    oracle.terminal,
);
```
If a generated prelude can legitimately make the scenario unreachable, normalize that prelude before entering this family rather than silently returning. #493 specifically requires authentication failure to be classified and to expose no unauthenticated plaintext, so selecting this arm must prove that outcome.

Jaruso (commented):
**Changes required before merge.** GitHub will not allow this account to submit `REQUEST_CHANGES` on its own PR, so I am recording the blocking review as a comment review instead.

I reviewed this as the final #493-C slice against issue #493's acceptance criteria and the implementation-plan comment, including the already-merged #493-A (#579) and #493-B (#584) coverage. The PR is structurally on the right path: the in-memory carrier is fixed-capacity/no-network, the required progression/cleanup targets and deterministic companions are present, the documentation/commands/bounds are filled in, and CI on this head is green.

There are five acceptance-level blockers:

1. **Orderly close wipes the bridge but not the owned handshake driver/backend.** The generic transport contract explicitly requires `Driver.deinit()` at teardown because its `EventSink` can retain copied traffic-secret scratch. `finishClose()` therefore fixes one secret-retention path while leaving another until a later outer `deinit()`.
2. **The progression oracle swallows unexpected peer record errors.** If bytes emitted by the subject fail authentication/parsing at the peer, the harness marks the peer unusable, discards the captured bytes, and later skips the end-to-end byte equality. That turns exactly the kind of subject corruption this target should catch into a successful fuzz case.
3. **`made_progress == false` does not actually prove no movement/state change.** The current branch checks only the queue snapshot, not carrier bytes or the other observable state already summarized by `state_changed`.
4. **The teardown scenario does not dirty every buffer/both parsers before asserting zeroization.** `initial_parser`, inbound plaintext, and provenance are not populated at all; pending terminal state is optional; and several setup failures are swallowed. Those post-teardown zero checks can therefore pass vacuously.
5. **The cleanup authentication-failure family can return/pass without ever observing `AuthenticationFailed`.** Setup/reachability failures are silently returned and the terminal assertion is conditional.

These map directly to #493's required properties: exact byte preservation, no spin under zero progress, terminal/teardown secret cleanup, and exact typed failure classification rather than swallowing errors. Concrete fixes are in the inline comments.

After addressing them, rerun the stated `test-tls-record-fuzz`, `test-tls`, full/default, appliance-profile, and ReleaseFast fuzz smokes. I would also mutation-check at least (a) an invalid byte emitted toward the peer to prove the progression oracle fails, and (b) removal of `teardownDriver()` from orderly close to prove the new driver-retention regression catches it.

Procedural closeout: the PR body currently says `Ref #493`; if this merge is intended to close the final story as #493-A/B describe, change that to `Closes #493` (or close the issue explicitly after merge) once these blockers are resolved.

src/tls/encrypted_stream.zig
Jaruso: The deterministic setup is much better and fixes the empty-parser/plaintext/pending-terminal parts of this finding, but the provenance zeroization check is still partially vacuous. `appendInboundPlaintext(&filler, smith.index(2) == 0)` writes `false` into the provenance buffer for half the cases, which is indistinguishable from its initial/cleared state; the precondition only proves `len > 0`, not that the backing values are dirty. A broken `PlaintextProvenanceQueue.clear()` could therefore still pass those cases.

Please force this teardown family to write nonzero provenance state (e.g. `appendInboundPlaintext(&filler, true)`) and assert at least one `true` exists before `deinit()`, then keep the existing post-teardown all-`false` assertion. That makes every storage-zeroization check in this scenario genuinely non-vacuous.

src/tls/encrypted_stream.zig
```diff
@@ -4801,3 +4842,1633 @@ test "driver-owned handshake latches on the flush deadline when a fatal alert ca
...
+    var peer = PureZigRecordStream.init(.client, cp, suite);
+    defer peer.deinit();
+    var subject = PureZigRecordStream.initWithCarrier(.server, cp, suite, carrier_state.carrier());
+    defer subject.deinit();
+    try establishForSuite(&peer, &subject, suite);
+
+    var oracle = StreamOracle{};
+    // Optionally put genuine plaintext ahead of the tampered record. Bounded so
+    // the fixed inbound queue always keeps room for the tampered record itself.
+    for (0..smith.index(3)) |_| {
```
Jaruso: [P2] **Stage the genuine prelude so it is actually delivered before the authentication failure.** This currently queues the optional genuine records and the tampered record before the first `drive()`. Because the first read action is forced to transfer the full inbound queue and `drive_record_budget` is 8 while this case has at most three records, that one `drive()` processes the genuine records and then hits the tampered one. `fail(AuthenticationFailed)` clears `inbound_plaintext`, and the caller's `driveChecked(... ) orelse break` means `subjectReadOnce()` never runs. So `oracle.subject_received` is still zero: this does not exercise the documented/#493-C scenario of an authentication failure **behind already-delivered genuine plaintext**.

Make the prelude mandatory and stage it separately: enqueue at least one genuine record, drive/read until `oracle.subject_received == oracle.peer_sent` and assert that value is > 0, then construct/enqueue the tampered next-sequence record and require `AuthenticationFailed`. For example:
```zig
try enqueuePeerRecord(&peer, &carrier_state, &oracle, 32);
_ = (try driveChecked(&subject, &carrier_state, &oracle)).?;
try subjectReadOnce(&subject, &oracle, &read_buf);
try testing.expectEqual(oracle.peer_sent, oracle.subject_received);
try testing.expect(oracle.subject_received > 0);

// now seal/tamper the next record and drive to AuthenticationFailed
```
That also proves the stream's read sequence advanced successfully on genuine plaintext before the later record fails authentication.

Jaruso (commented):
**Second-pass review — two changes still required before merge.** The second commit fixes the two P1s, the no-progress assertion, and the original “authentication arm can silently pass” issue; I resolved those threads. The driver teardown regression is especially solid: it verifies backend release exactly once at orderly close, sink scratch zeroization, bridge key removal, and later `deinit()` idempotence.

Two P2 coverage gaps remain:

1. The teardown family now dirties almost every state correctly, but `appendInboundPlaintext(&filler, smith.index(2) == 0)` writes an all-`false` provenance shadow in half the cases. Since `false` is already the cleared value, `len > 0` does not prove that backing storage was dirtied. I left the existing teardown thread open and replied with the concrete fix: force `true` provenance and assert it is present before `deinit()`.
2. The authentication-failure family now deterministically reaches `AuthenticationFailed`, but it still does **not** cover the scenario it and the #493 implementation plan claim: failure *behind already-delivered genuine plaintext*. All prelude records and the tampered record are queued before the first `drive()`. The forced first carrier read can take the entire queue, and `drive_record_budget` (8) exceeds the maximum three records in the case, so the same `drive()` opens the genuine records and then fails on the tampered one. `fail()` clears `inbound_plaintext`, and `driveChecked(... ) orelse break` prevents `subjectReadOnce()` from running. The prelude therefore never reaches the application. See the inline comment for the staged fix.

Once these are addressed, I do not see another code-level blocker against #493's acceptance criteria. Procedurally, the branch is currently **2 commits behind `main`**, GitHub reports `mergeable=false`, and the connector shows **no Actions run attached to head `e02f893`**. Rebase after the fixes and rerun the full gate set on the rebased head before merging.

src/tls/encrypted_stream.zig
```diff
@@ -4801,3 +4842,1676 @@ test "driver-owned handshake latches on the flush deadline when a fatal alert ca
...
+/// of pre-queued handshake records draining at one byte per writable
+/// notification, behind a script that only becomes writable one step in eight.
+const fuzz_cleanup_max_drives = 4096;
+
+test "fuzz: TLS record: encrypted stream cleanup preserves root errors across alerts and epoch transitions" {
+    try testing.fuzz({}, fuzzEncryptedStreamCleanupInput, .{
+        .corpus = &.{
+            "",
+            &[_]u8{0},
+            // One entry per scenario family, plus a spread of script and
```
Jaruso, line 5650: [P3, non-blocking] This comment is stale after the selector was removed. These entries are no longer “one per scenario family”; every corpus entry now runs all four families. Please reword this to describe the intended Smith/script/root-error spread instead, so the checked-in corpus documentation matches the new execution model.

Jaruso (commented):
Third-pass review: **no blocking findings remain.** GitHub does not allow this account to approve its own PR, so I am recording the final status as a comment review.

I re-checked the current rebased head against the two remaining P2s and against #493's acceptance criteria. Both blockers are fixed correctly:

- The teardown family now dirties the provenance backing storage with at least one `true` value and asserts that it is genuinely non-cleared before `deinit()`, so the post-teardown all-`false` check is no longer vacuous.
- The authentication-failure family now stages genuine application plaintext first, proves it reached the caller and advanced the application read sequence, and only then injects the tampered next record. It requires `AuthenticationFailed` and verifies the already-delivered byte count remains exact across failure.

The additional change to execute all four cleanup families in every fuzz callback is also a good fix: it closes the discovered deterministic seed-replay coverage hole instead of relying on a Smith selector that the checked-in corpus never spread across.

Across #493-A (#579), #493-B (#584), and this final #493-C slice, the linked issue's acceptance criteria are now covered: standalone commands, fragmentation/saturation byte accounting, authentication/sequence behavior, epoch and teardown cleanup, memory-only partial/would-block/close/cancellation/repeated-drive coverage, bounded deterministic fuzzing, CI-safe replay plus documented heavy runs, and checked-in regressions/mutation checks. The PR now uses `Closes #493` as expected.

Current head `63d9ec8` is mergeable and CI run #1247 completed successfully across unit tests, appliance tests, build matrices, format, security, dependency review, and the other repository gates.

I left one **non-blocking P3** inline cleanup note: the corpus comment still describes entries as one-per-scenario-family even though every family now runs for every entry. The PR summary also still says “seven named deterministic companions” while the table/code now contain eight. Neither affects correctness.

From the code/security/acceptance standpoint, **this is ready to merge and close #493**.

PR #588, opened 2026-08-08
tls: [323-J] add shared TLS-engine interoperability and conformance suite

Jaruso:
## Summary

Automated external interop/conformance matrix for the shared TLS 1.3 engine, per #338 (research story 323-J).

- **Both roles, both transports, one engine.** `tls_interop_tool` (new) drives the record transport over real TCP; `h3_interop_tool` gains matrix flags and policy-configurable QUIC constructors. Both build their `tls_core.policy.Policy` from one shared vocabulary (`tests/tls_interop_matrix.zig`, derived from the engine's own `native_capabilities`), so a matrix row means the same engine configuration on either transport — and a newly supported suite/group/signature becomes a new row automatically instead of being silently skipped.
- **Independent implementations, out of process.** OpenSSL and GnuTLS, both roles. Nothing foreign is linked in; peer package names stay inside `scripts/interop/`, the audited external-peer boundary.
- **Positive rows assert the negotiated tuple**, not just that a handshake completed. A row pinning one value per dimension fails if the peer landed elsewhere.
- **Negative rows pin the RFC 8446 §6 failure class**, checking the alert whichever side raised it (`alert_origin=local|peer`): ALPN / cipher / group / signature no-overlap, TLS 1.2 downgrade, absent SNI, wrong pinned certificate, and two raw malformed-ordering cases.
- **HelloRetryRequest in both directions**, consuming the backend behaviour #333/#484/#485 landed.
- **Secret-free transcripts + reduced failure fixtures** for every row, captured *below* the record-protection boundary so redaction is structural rather than a filtering step that has to be trusted.
- **Profiles + CI.** `--profile ci` keeps both roles, both transports, every cipher suite and every negative row, thinning only the positive group/signature sweep. New CI job runs it on every PR and uploads logs/transcripts (never `certs/`) on failure.

## Defects this suite found (fixed here)

1. **HelloRetryRequest could not complete against a compatibility-mode client.** RFC 8446 §5.1 has a client send `change_cipher_spec` immediately after receiving an HRR, before ClientHello2. An HRR flight derives no traffic secrets, so the record transport's "has a ClientHello been accepted yet?" test — which inferred the answer from encryption-epoch movement — still said no and rejected the legal record with `UnexpectedRecordContent`. In practice the native TLS listener could not complete an HRR handshake with essentially any real client. It now asks the backend whether it has sent an HRR (`Backend.helloRetryRequestSent`), which is likewise only true after a complete, accepted ClientHello — so a `change_cipher_spec` spliced into a *partial* ClientHello is still rejected.

2. **No-overlap failures sent the wrong alert.** No mutually supported cipher suite, group, or signature scheme was reported as `illegal_parameter`, telling the peer its ClientHello was malformed when every value in it was legal. RFC 8446 §4.1.1 requires `handshake_failure`. Adds the distinct `NoMutualParameters` failure plus `UnsupportedProtocolVersion` → `protocol_version` per §4.2.1. The mapping is deliberately **role-aware**: a client rejecting a ServerHello selection it never offered keeps `illegal_parameter` per §4.1.3, since that is the server violating the protocol rather than the two sides sharing nothing.

## Test plan

- [x] `zig build test` — green (includes the new `test-tls-interop-matrix` step).
- [x] `scripts/interop/run-tls-interop.sh --profile full` — **101/101 rows pass** against OpenSSL 3.6.2 and GnuTLS, both roles, both transports, all 18 tuples.
- [x] `scripts/interop/run-tls-interop.sh --profile ci` — **51/51 rows pass**.
- [x] Redaction contract verified empirically: all 83 transcripts from a full run were parsed and checked to contain only content-type 21/22 records, with no private-key material present.
- [x] `scripts/audit-dependencies.sh` passes — peer names stay inside the external-peer boundary.
- [x] HRR regressions in both directions, plus the companion case proving a `change_cipher_spec` before any ClientHello is still rejected.
- [x] Peer-alert accessor covered for ALPN rejection, certificate rejection, and a clean `close_notify` (which must not be recorded as a fatal class).

## Notes

`UnsupportedProtocolVersion` → `protocol_version` has no external row: it needs a ClientHello whose `supported_versions` is present but lists no version we accept, and neither peer can be persuaded to send one (a TLS 1.2 client omits the extension entirely, which is the `tls12_downgrade` row). It is covered by unit tests instead, and this is called out in the docs.

The external QUIC/H3 *peer* matrix (ngtcp2, quiche, aioquic) already exists in `scripts/interop/run-interop.sh`; this suite links to it rather than duplicating it, and its own QUIC rows prove each pinned tuple traverses the QUIC transport of the same engine.

Closes #338

### DIFF
--- a/docs/TLS_INTEROP_MATRIX.md
+++ b/docs/TLS_INTEROP_MATRIX.md
@@ -55,9 +55,22 @@ exits non-zero if any row failed and lists them at the end.
 
 The tuple vocabulary lives in exactly one place,
 [`tests/tls_interop_matrix.zig`](../tests/tls_interop_matrix.zig), and is
-derived from the engine's own `native_capabilities` rather than restated. A
-suite, group, or signature scheme added to the engine therefore becomes a new
-matrix row automatically instead of being silently skipped.
+derived from the engine's own `native_capabilities` rather than restated.
+
+That derivation reaches the shell runner too: it builds its dimension lists
+from `tls_interop_tool list-capabilities` instead of keeping its own copy, so
+there is no second registry to drift. A suite, group, or signature scheme added
+to the engine therefore becomes a new matrix row automatically — and if the
+runner has no OpenSSL/GnuTLS spelling (or no credential) for it yet, preflight
+fails loudly naming exactly what is missing, rather than quietly enumerating
+the old set.
+
+Parsing is likewise scoped to `native_capabilities`, not the whole protocol
+registry. `secp384r1` and `rsa_pkcs1_sha256` have matrix names but are not
+natively negotiated, and are rejected: `Config.policy` hands its selection
+straight to `Policy.fromCapabilities`, which does not filter, so accepting them
+would let a row *widen* the engine policy and turn a green cell into evidence
+about a test-only capability.
 
 | Dimension | Values | Matrix name |
 | --- | --- | --- |
@@ -106,18 +119,42 @@ result line reports `alert_origin=local` when the engine rejected the peer and
 | `cipher_no_overlap` | `handshake_failure` | RFC 8446 §4.1.1 |
 | `group_no_overlap` | `handshake_failure` | RFC 8446 §4.1.1 |
 | `signature_no_overlap` | `handshake_failure` | RFC 8446 §4.4.2.2 |
-| `tls12_downgrade` | `missing_extension` | RFC 8446 §4.2.1 — a TLS 1.2 hello carries no `supported_versions` |
+| `tls12_downgrade` | `protocol_version` | RFC 8446 App. D.2 + §4.2.1 — an absent `supported_versions` is a legal *legacy* hello |
 | `sni_absent` | `missing_extension` | server configured with `require_sni` |
 | `wrong_pinned_certificate` | `bad_certificate` | RFC 8446 §6 |
 | `malformed_ordering` | `UnexpectedRecordContent` | application data before any ClientHello |
 | `ccs_before_clienthello` | `UnexpectedRecordContent` | RFC 8446 §5.1 — the compatibility window is not open yet |
 
-`UnsupportedProtocolVersion` → `protocol_version` (RFC 8446 §4.2.1) has no
-external row: it needs a ClientHello whose `supported_versions` is present but
-lists no version we accept, and neither OpenSSL nor GnuTLS can be persuaded to
-send one — a TLS 1.2 client omits the extension entirely, which is the
-`tls12_downgrade` row instead. The mapping is covered by unit tests in
-`src/tls/alerts.zig`.
+Every negative row pins a typed local error as well as the wire alert. A row
+that asserted only "the handshake failed" would stay green through a timeout,
+a record-ordering bug, or any unrelated failure, while the behaviour it names
+went untested.
+
+The `supported_versions`-present-but-unsatisfiable variant of §4.2.1 has no
+external row — neither OpenSSL nor GnuTLS can be persuaded to send such a
+hello, since a TLS 1.2 client omits the extension entirely (which is the
+`tls12_downgrade` row). It is covered by unit tests in
+`src/tls/tls13_backend_tests.zig` alongside the absent-extension case.
+
+### Certificate selection
+
+The positive rows hand the server exactly one identity, so they prove the
+*preselected* certificate works — they cannot prove the engine would have
+chosen it. A separate section gives the server all three credentials at once
+through the production SNI/signature-algorithm credential provider
+(`--identity PATTERN:CERT:KEY`, repeatable) and lets the peer's ClientHello
+drive the choice:
+
+| Row | Selector | Expectation |
+| --- | --- | --- |
+| `sni_ed25519` / `sni_ecdsa_p256` / `sni_rsa` | peer SNI | the matching credential |
+| `unknown_sni_uses_default` | peer SNI matching nothing | the default (first) credential |
+| `sigalgs_ed25519` / `sigalgs_rsa_pss` | peer `signature_algorithms` | two credentials share one host name, so only the offer can disambiguate |
+| `no_applicable_credential` | peer offers a scheme no credential holds | `handshake_failure` (RFC 8446 §4.4.2.2) |
+
+Each row asserts `--expect-signature`, read back from the engine's own
+`negotiated_signature_scheme` — what the engine actually signed
+CertificateVerify with — not from what the harness passed in.
 
 ### HelloRetryRequest
 
@@ -156,16 +193,16 @@ reproduces the negotiation without the peer being present.
 
 ## Profiles
 
-`full` runs every tuple: 3 suites × 2 groups × 3 signatures = 18 tuples, each
-in both roles against both peers on the record transport, plus the same 18 on
-QUIC, plus HRR and the negative rows.
+**Both profiles run every supported tuple.** 3 suites × 2 groups × 3 signatures
+= 18 tuples, each in both record roles against OpenSSL, plus all 18 on QUIC,
+plus HRR, every negative row, and every certificate-selection row. #338
+requires CI to prove positive interop for all supported tuples and both roles,
+so tuple coverage is not something a profile may reduce.
 
-`ci` keeps **both roles, both transports, every cipher suite, and every
-negative row**, and reduces only the positive group/signature sweep: the
-baseline suite (`aes128-gcm-sha256`) still walks all six group/signature
-pairings, while the other two suites run one representative pairing each. A
-regression in any single group or signature is therefore still caught, at a
-fraction of the wall-clock cost.
+What `ci` reduces is *peer multiplicity*: GnuTLS cross-checks one
+representative tuple per suite rather than all 18. Only a regression that is
+simultaneously tuple-specific **and** GnuTLS-specific could reach the nightly
+`full` profile without CI catching it.
 
 ## Adding a row
 
@@ -175,10 +212,12 @@ fraction of the wall-clock cost.
   fail until you do) and add the OpenSSL/GnuTLS spellings to
   `run-tls-interop.sh`.
 - **A new negative case**: add a `run_negative_server_row` call naming the
-  expected alert. Leave the alert empty only when no standard defines one, and
-  pin `--expect-error` instead so the row still asserts something specific — a
-  negative row that only asserts "something went wrong" will pass for the wrong
-  reason.
+  expected alert, and always pin `--expect-error` too. Leave the alert empty
+  only when no standard defines one — but never leave both empty, or the row
+  will pass for the wrong reason.
+- **A new raw-bytes case** (no TLS client involved): use `run_raw_record_row`,
+  which writes the bytes from the runner itself rather than through a child
+  shell.
 
 ## Findings
 
@@ -198,5 +237,15 @@ it:
    suite, group, or signature scheme was reported as `illegal_parameter`,
    telling the peer its ClientHello was malformed when every value in it was
    legal. RFC 8446 §4.1.1 requires `handshake_failure`. The engine now has a
-   distinct `NoMutualParameters` failure for this, and a distinct
-   `UnsupportedProtocolVersion` mapping to `protocol_version` per §4.2.1.
+   distinct `NoMutualParameters` failure for this. The mapping is role-aware:
+   a *client* rejecting a ServerHello selection it never offered keeps
+   `illegal_parameter` per §4.1.3, since that is the server violating the
+   protocol rather than the two sides sharing nothing.
+
+3. **A legacy ClientHello was rejected as malformed.** A ClientHello with no
+   `supported_versions` extension was reported as `missing_extension`. RFC 8446
+   Appendix D.2 is explicit that this is a legal pre-1.3 hello: the server
+   negotiates `min(legacy_version, TLS 1.2)`, and a TLS-1.3-only endpoint then
+   aborts with `protocol_version` per §4.2.1. The engine now has a distinct
+   `UnsupportedProtocolVersion` failure covering both that case and a
+   `supported_versions` naming nothing we support.
--- a/docs/TLS_INTEROP_MATRIX.md
+++ b/docs/TLS_INTEROP_MATRIX.md
@@ -1,0 +1,202 @@
+# Shared TLS-engine interoperability and conformance matrix
+
+Tracking issue: [#338](https://github.com/Bare-Systems/Tardigrade/issues/338)
+(research story 323-J).
+
+This suite proves that the TLS 1.3 engine Tardigrade ships — one engine, two
+transports — interoperates with independent implementations across every
+cipher-suite/group/signature tuple it claims to support, in **both** the client
+and the server role, and that it fails in the way the standards require when
+there is nothing to negotiate.
+
+It complements rather than replaces:
+
+- [`RESUMPTION_TEST_PLAN.md`](RESUMPTION_TEST_PLAN.md) (#369) — resumption and
+  0-RTT behaviour through the full gateway process.
+- `scripts/interop/run-interop.sh` (#247) — the HTTP/3 peer matrix against
+  ngtcp2, quiche, and aioquic.
+
+## Running it
+
+```bash
+zig build build-tls-interop build-h3-interop
+```
+
+```bash
+scripts/interop/run-tls-interop.sh
+```
+
+```bash
+scripts/interop/run-tls-interop.sh --profile ci
+```
+
+```bash
+scripts/interop/run-tls-interop.sh --list --profile ci
+```
+
+`openssl` is required. `gnutls-cli`/`gnutls-serv` are optional; without them
+the independent-implementation rows report `SKIP` rather than failing, so the
+suite still runs on a machine that only has OpenSSL.
+
+| Variable | Meaning |
+| --- | --- |
+| `TLS_INTEROP_PROFILE` | `ci` or `full` (default `full`; `--profile` wins) |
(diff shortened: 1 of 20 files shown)


scripts/interop/run-tls-interop.sh
```diff
@@ -0,0 +1,544 @@
...
+    rsa-pss-rsae-sha256) echo "$certs/rsa2048" ;;
+  esac
+}
+
+# The reduced CI profile: every suite and both roles survive, but the
+# group/signature sweep collapses to one representative pairing per suite. The
+# baseline suite still walks the full sweep, so a regression in any single
+# group or signature is still caught somewhere in CI.
+row_in_ci_profile() { # suite group signature
+  [ "$1" = "aes128-gcm-sha256" ] && return 0
```
Jaruso: [P1] This reduced predicate does not meet #338's explicit acceptance criterion that **CI proves positive interop for all supported tuples and both roles**. With the current 3×2×3 capability set it keeps all 6 AES-128 pairings, but only `x25519/ed25519` for AES-256 and ChaCha20: 8/18 tuples. A regression specific to e.g. AES-256 + P-256 + ECDSA can therefore merge green.

Please keep all 18 tuple combinations in `--profile ci`. If wall-clock is the concern, reduce peer multiplicity or shard the matrix instead of deleting tuples. For example, CI could run every tuple against OpenSSL in both record roles plus every QUIC tuple, while keeping a smaller representative GnuTLS cross-check (or split the 18 tuples across parallel jobs). The important invariant is that every supported `(suite, group, signature)` combination appears in required CI.

scripts/interop/run-tls-interop.sh
```diff
@@ -0,0 +1,544 @@
...
+say "== record transport: negative conformance =="
+
+# RFC 7301 §3.2: no mutually supported protocol is fatal no_application_protocol.
+run_negative_server_row "record/negative/alpn_no_overlap" no_application_protocol \
+  --alpn h2 -- \
+  'printf x | '"$openssl_bin"' s_client -connect 127.0.0.1:$PORT -servername tardigrade.test -tls1_3 -alpn http/1.1 -CAfile $CERT'
+
+# RFC 8446 §4.2.1: a TLS 1.2 ClientHello carries no supported_versions, so a
+# TLS-1.3-only server rejects it before any version can be negotiated.
+run_negative_server_row "record/negative/tls12_downgrade" missing_extension \
```
Jaruso: [P1] The expected failure class here is backwards for a genuine TLS 1.2 ClientHello. RFC 8446 Appendix D.2 says that when `supported_versions` is absent, the server treats the hello as a pre-1.3 ClientHello and negotiates `min(ClientHello.legacy_version, TLS 1.2)`; if the server only supports versions greater than that legacy version, it **MUST abort with `protocol_version`**.

Right now `parseClientHelloObserved()` returns `MissingSupportedVersions` for this input and `mapNegotiationError()` folds that into `MissingExtension`, so this green external row is codifying the behavior the conformance suite should catch.

Please distinguish a valid legacy ClientHello from a malformed TLS-1.3 ClientHello and exercise the new typed version failure here, e.g. conceptually:

```sh
run_negative_server_row "record/negative/tls12_downgrade" protocol_version \
  --alpn http/1.1 --expect-error UnsupportedProtocolVersion -- \
  '... openssl s_client ... -tls1_2 ...'
```

The parser/backend will need to preserve enough legacy-version context for that row to reach `UnsupportedProtocolVersion` rather than `MissingExtension`.

scripts/interop/run-tls-interop.sh
```diff
@@ -0,0 +1,544 @@
...
+  'printf x | '"$openssl_bin"' s_client -connect 127.0.0.1:$PORT -servername tardigrade.test -tls1_3 -groups P-384 -alpn http/1.1 -CAfile $CERT'
+
+# No signature scheme the server's Ed25519 credential can satisfy
+# (RFC 8446 §4.4.2.2 -> handshake_failure).
+run_negative_server_row "record/negative/signature_no_overlap" handshake_failure \
+  --signature ed25519 --alpn http/1.1 -- \
+  'printf x | '"$openssl_bin"' s_client -connect 127.0.0.1:$PORT -servername tardigrade.test -tls1_3 -sigalgs rsa_pss_rsae_sha256 -alpn http/1.1 -CAfile $CERT'
+
+# A ClientHello with no SNI against a server that requires one.
+run_negative_server_row "record/negative/sni_absent" "" \
```
Jaruso: [P1] This row only proves that *something* failed. Passing an empty expected alert means `run_negative_server_row` installs only `--expect fail`; there is also no `--expect-error`. A timeout, record-ordering bug, or unrelated handshake failure would make this row pass even though the intended SNI policy was never exercised.

The docs currently state this case is `missing_extension`, and the engine maps `MissingServerName` to `MissingExtension`, so pin both the wire class and local error instead of accepting any failure, for example:

```sh
run_negative_server_row "record/negative/sni_absent" missing_extension \
  --require-sni --alpn http/1.1 --expect-error MissingExtension -- \
  '... -noservername ...'
```

If the project wants a different typed error for required-SNI policy, that is fine too; the important part is that this negative row asserts the intended cause/class.

tests/tls_interop_matrix.zig
```diff
@@ -0,0 +1,296 @@
...
+
+pub fn parseCipherSuite(name: []const u8) ParseError!CipherSuite {
+    inline for (comptime std.enums.values(CipherSuite)) |suite| {
+        if (std.mem.eql(u8, name, cipherSuiteName(suite))) return suite;
+    }
+    return error.UnknownCipherSuite;
+}
+
+pub fn parseNamedGroup(name: []const u8) ParseError!NamedGroup {
+    inline for (comptime std.enums.values(NamedGroup)) |group| {
```
Jaruso: [P1] This parser violates the module's stated invariant that matrix rows can only **narrow** `native_capabilities`. It iterates every enum value, so it accepts `secp384r1` even though `named_groups` is the native-capability slice. `parseSignatureScheme()` has the same issue with `rsa_pkcs1_sha256`. `Config.policy()` then passes the selected slice directly to `Policy.fromCapabilities()`, which does not filter it, so a test caller can widen the engine policy to a non-native capability.

Please parse against the exported native slices, not the whole enum registry (and do the same for every dimension so future enum additions cannot accidentally become test-only capabilities):

```zig
pub fn parseNamedGroup(name: []const u8) ParseError!NamedGroup {
    for (named_groups) |group| {
        if (std.mem.eql(u8, name, namedGroupName(group))) return group;
    }
    return error.UnknownNamedGroup;
}

pub fn parseSignatureScheme(name: []const u8) ParseError!SignatureScheme {
    for (signature_schemes) |scheme| {
        if (std.mem.eql(u8, name, signatureSchemeName(scheme))) return scheme;
    }
    return error.UnknownSignatureScheme;
}
```

You can still keep the separate whole-enum name-collision tests for registry hygiene.

scripts/interop/run-tls-interop.sh
```diff
@@ -0,0 +1,544 @@
...
+    sleep 0.1
+  done
+  return 1
+}
+
+# ── matrix vocabulary ───────────────────────────────────────────────────────
+# These names are the ones tests/tls_interop_matrix.zig parses; the OpenSSL and
+# GnuTLS spellings beside them are how each peer names the same identifier.
+
+suites=(aes128-gcm-sha256 aes256-gcm-sha384 chacha20-poly1305-sha256)
```
Jaruso: [P1] These arrays create a second, unchecked capability registry. The PR/docs say a suite/group/signature added to `native_capabilities` automatically becomes a new matrix row, but the actual runner will silently keep enumerating only these hard-coded values. `test-tls-interop-matrix` cannot detect that drift because it never observes this shell list.

Please make the runner consume a machine-readable list derived from `tests/tls_interop_matrix.zig`/`native_capabilities`, or add an explicit cross-language consistency gate. One straightforward shape would be a tool mode such as:

```text
tls_interop_tool list-capabilities --format tsv
cipher-suite\taes128-gcm-sha256
...
group\tx25519
...
signature\ted25519
...
```

and have the script populate `suites`, `groups`, and `signatures` from that output. Peer spelling maps can remain in this script, and an unknown spelling should fail preflight loudly. That preserves the claimed "new capability => new row (or immediate CI failure)" invariant.

tests/h3_interop_tool.zig
```diff
@@ -92,8 +93,59 @@ const Args = struct {
...
+        matrix.namedGroupName(backend.engine.negotiated_named_group),
+        backend.alpn,
+        saw_hrr,
+    });
+}
+
+/// A row that pinned exactly one value along a dimension is asserting the
+/// peer landed on it, not merely that some handshake succeeded -- the same
+/// check `tls_interop_tool` applies on the record transport.
+fn matrixTupleHolds(args: Args, backend: *const tls_backend.Tls13Backend) bool {
```
Jaruso, line 137: [P2] #338's remaining-work review explicitly asks for **one shared TLS expectation/conformance model used by record and QUIC carriers**. This reimplements the negotiated-tuple assertions separately from `tls_interop_tool.evaluate()`, which is exactly the drift-prone duplication that requirement was trying to remove. There is already drift: the record evaluator also handles the expected ALPN, while this helper checks only suite/group.

Please move the common expectation into `tls_interop_matrix.zig`, e.g. a transport-neutral result/validator:

```zig
pub const Negotiated = struct {
    cipher_suite: ?CipherSuite,
    named_group: ?NamedGroup,
    alpn: []const u8,
};

pub fn validateNegotiated(self: *const Config, got: Negotiated, expected_alpn: ?[]const u8) ValidationError!void { ... }
```

Both tools should translate their backend state into that shared value and call the same validator; carrier-specific I/O/outcome handling can remain separate.

scripts/interop/run-tls-interop.sh
```diff
@@ -0,0 +1,544 @@
...
+# ── row runners ─────────────────────────────────────────────────────────────
+
+# native TLS server <- external client
+run_server_row() { # name peer suite group signature [extra native args...]
+  local name="$1" peer="$2" suite="$3" group="$4" sig="$5"
+  shift 5
+  local p log cert peer_groups
+  next_port; p="$port"
+  log="$logs/${name//\//_}"
+  cert="$(cert_for_signature "$sig")"
```
Jaruso, line 254: [P1] This means the harness is choosing the certificate *before* the TLS engine runs. The server process then receives one fixed identity, so a passing signature row demonstrates that the preselected certificate works; it does not exercise the engine's credential-selection path. #338 explicitly includes **certificate selection**, and the remaining-work note calls out credential-selection failures. The `wrong_pinned_certificate` row is client verification, which is a different behavior.

Please add a server mode backed by multiple credentials/the credential-provider seam and rows that force the engine to select among them from peer SNI + signature-algorithm offers, plus a no-applicable-credential negative case. For example, the tool could accept repeated identities:

```text
--identity tardigrade.test:ed25519-cert.pem:ed25519-key.pem
--identity p256.test:p256-cert.pem:p256-key.pem
--identity rsa.test:rsa2048-cert.pem:rsa2048-key.pem
```

Then have external clients vary `-servername`/`-sigalgs` and assert the leaf/fingerprint or selected credential reported by the engine. That would cover the actual #334 credential-selection behavior rather than scripting the expected answer outside TLS.

scripts/interop/run-tls-interop.sh
```diff
@@ -0,0 +1,544 @@
...
+  return 1
+}
+
+# Wait on the external server's own readiness output. Deliberately *not* a
+# `nc -z` connect probe: `openssl s_server -naccept 1` serves exactly one
+# connection, and the probe would consume it, leaving the row's real client
+# with nothing to talk to.
+wait_for_peer_listener() { # logfile pattern
+  local i
+  for i in $(seq 1 150); do
```
Jaruso: [P1] The PR currently fails the repository's required `Lint` job on this new script. Shellcheck reports `SC2034` here because `i` is unused, and `SC2016` on the later single-quoted child commands containing `$PORT`/`$CERT`.

This loop can simply use `_`:

```sh
for _ in $(seq 1 150); do
```

For the later `SC2016` sites, the delayed expansion inside `bash -c` is intentional, so either restructure the command construction or add narrowly scoped `# shellcheck disable=SC2016` comments explaining that `$PORT`/`$CERT` are intentionally expanded by the child shell. Please get the top-level Lint job green before merge.

Jaruso (commented):
**Blocking review — changes required before #338 should be closed.**

I reviewed all 20 changed files against #338's acceptance criteria and the 2026-07-27 remaining-work note. GitHub does not permit this account to submit `REQUEST_CHANGES` on its own PR, so this is posted as a `COMMENT` review, but the findings below are merge-blocking.

The production-side fixes in this PR are strong: the HRR/compatibility-mode CCS window is tied to accepted backend state rather than epoch movement, the server-side no-overlap mapping is now role-aware, peer fatal alerts are retained for conformance diagnostics, and the focused regression coverage is good. The dedicated TLS conformance CI job also passes on this head.

The harness does not yet satisfy #338 closely enough to close it:

1. The reduced CI profile exercises only 8 of the 18 supported cipher/group/signature tuples. #338 explicitly requires CI to prove positive interop for **all supported tuples and both roles**. Reducing duplicate peer coverage or sharding is fine; dropping tuple combinations is not.
2. The TLS 1.2 downgrade row codifies `missing_extension`, but RFC 8446 Appendix D.2 requires a TLS-1.3-only server receiving a legacy ClientHello without `supported_versions` to abort with `protocol_version`. The current parser/backend path maps absence to `MissingExtension`, so the green row is currently blessing the wrong standards behavior.
3. The absent-SNI negative row asserts only that *some* failure occurred, despite the docs naming a specific failure class.
4. The matrix parser can select enum values outside `native_capabilities`, so `Config.policy()` can widen the test policy despite its documented "narrow only" invariant.
5. The shell runner separately hard-codes all suite/group/signature arrays. A new native capability can therefore still be silently omitted from the actual matrix even though the PR/docs claim it automatically becomes a row.
6. Record and QUIC have separate negotiated-tuple assertion implementations. The issue's remaining-work note explicitly calls for one shared expectation/conformance model, and these implementations already have different checks.
7. Certificate **selection** is not exercised: the shell harness selects the one certificate before starting the engine, and the tool gives the server one fixed identity. The wrong-pin case covers client certificate verification, not the server credential-selection/SNI/signature path requested by #338.
8. The overall CI run is currently red because the new runner fails shellcheck (`SC2034` plus the intentionally delayed `$PORT`/`$CERT` expansions triggering `SC2016`).

I've left inline comments with concrete fixes/examples for each blocker. Once these are addressed, the underlying engine changes and the general harness shape look like a good foundation for closing #338.

docs/TLS_INTEROP_MATRIX.md
```diff
@@ -0,0 +1,251 @@
...
+| `cipher_no_overlap` | `handshake_failure` | RFC 8446 §4.1.1 |
+| `group_no_overlap` | `handshake_failure` | RFC 8446 §4.1.1 |
+| `signature_no_overlap` | `handshake_failure` | RFC 8446 §4.4.2.2 |
+| `tls12_downgrade` | `protocol_version` | RFC 8446 App. D.2 + §4.2.1 — an absent `supported_versions` is a legal *legacy* hello |
+| `sni_absent` | `missing_extension` | server configured with `require_sni` |
+| `wrong_pinned_certificate` | `bad_certificate` | RFC 8446 §6 |
+| `malformed_ordering` | `UnexpectedRecordContent` | application data before any ClientHello |
+| `ccs_before_clienthello` | `UnexpectedRecordContent` | RFC 8446 §5.1 — the compatibility window is not open yet |
+
+Every negative row pins a typed local error as well as the wire alert. A row
```
Jaruso, line 128: [P3, non-blocking] The two raw-record rows intentionally have no TLS wire alert (`UnexpectedRecordContent` is a record-layer failure and `emittedAlertFor` returns `null`), so “Every negative row pins a typed local error **as well as the wire alert**” is slightly too strong. I’d phrase this as: “Every negative row pins a typed local error and, where the protocol defines/emits one, the wire alert.” The implementation itself is correct; this is just keeping the documented contract exact.

Jaruso (commented):
Re-review of `382699ce71affc53b2036d615b634d36318d2523`: the eight blockers from my previous review are addressed, and I do not see a remaining merge-blocking issue.

Verified against #338 and the previous review findings:

- CI now covers all 18 supported `(cipher, group, signature)` tuples in both record roles against OpenSSL and all 18 over QUIC; the reduced profile trims only GnuTLS peer multiplicity.
- The runner derives its dimension lists from `native_capabilities` through `list-capabilities`, and native parsing can no longer widen policy to registry-only algorithms.
- Record and QUIC now use the same `Config.validateNegotiated` expectation model.
- The TLS 1.2/absent-`supported_versions` path now produces `UnsupportedProtocolVersion` / `protocol_version` as required by RFC 8446 Appendix D.2 + §4.2.1, with focused unit coverage and a real OpenSSL negative row.
- Negative rows now pin typed causes rather than accepting arbitrary failures.
- Certificate selection is now exercised through the production `ReloadableProvider` / `CredentialProvider` seam with multiple simultaneously available identities, including SNI selection, signature-algorithm disambiguation, default fallback, and no-applicable-credential failure.
- The HRR/compatibility-CCS and role-aware alert fixes from the original commit still look correct and retain focused regression coverage.
- The repository CI run is fully green. The TLS conformance job itself executed `pass=78 fail=0 skip=0`; Lint/ShellCheck is green as well.

**Disposition: approved in substance / okay to merge and close #338.** GitHub does not allow this authenticated account to formally approve its own PR, so this is necessarily a `COMMENT` review.

Only minor housekeeping remains: I left one non-blocking documentation wording comment below, and the PR description still has the old `101/101` / `51/51` counts and says CI thins tuple coverage even though the new implementation runs `108/108` full and `78/78` CI and preserves every tuple.

PR #590, opened 2026-08-09
pki: [324-K] define revocation, OCSP, CRL, and stapling policy seam

Jaruso:
## Summary

Defines an explicit, testable certificate-status policy for the pure-Zig PKI path — CRLs, OCSP responses, stapling, and must-staple — without introducing online revocation fetching.

- **`src/pki/revocation.zig`** — four modes (`disabled` default, `stapled_only`, `soft_fail`, `strict`); evidence model for stapled responses, cached status, and CRL sets with freshness bounds, defect reasons, and an authentication flag; deterministic source precedence where a usable `revoked` from any source wins.
- **Explicit default** — every accepted path carries a `revocation.Report` naming, per certificate, what was checked and what was not. Under the default policy the report says nothing was checked, so an acceptance never implies revocation was verified. `disabled` also never *rejects*: evidence is not examined at all, so a peer cannot fail a handshake against a gateway with revocation switched off.
- **Evidence is keyed by certificate identity** (SHA-256 over the exact DER), never by path position. One evidence set is shared across every candidate, and candidates differ in shape and depth, so a position-derived key would both misdirect evidence onto the wrong certificate and let a longer candidate's assertions invalidate a shorter one. Assertions for certificates not on the path being evaluated are skipped.
- **Must-staple (RFC 7633)** — new TLS Feature extension parsing in the X.509 model. The **end-entity** form is the stapling obligation, enforced in every mode that consults status. The **issuer** form is a different rule: RFC 7633 §4.2.2 makes a TLS-Feature-carrying CA constrain what it signs — children must assert the same feature set or a superset, for every advertised feature — enforced during path validation as `tls_feature_constraint_violation`, independently of the status mode. The end-entity obligation is scoped by RFC 7633 §§4.1/4.3.3 to features the ClientHello actually offered, carried in as `offered_status_request` (false today, since this stack does not yet send `status_request`). A certificate advertising feature 5 or RFC 6961 feature 17 is therefore outside the intersection and reported as `not_offered_by_client` rather than rejected, while both features still participate in the §4.2.2 issuer constraint.
- **Reported, not assumed** — `Report.must_staple` distinguishes `not_required`, `enforced`, `not_offered_by_client`, `unenforced_status_disabled`, and `unenforced_by_configuration`, so an acceptance never reads as "honored" when the assertion was merely not applied.
- **Status data surfaced** — typed `authorityInfoAccess`, `crlDistributionPoints`, and `hasOcspResponder` accessors; `strict` distinguishes "certificate publishes no revocation mechanism" (`status_source_unsupported`) from "we looked and found nothing".
- **No ambient I/O** — evidence is caller-supplied. `revocation.Provider` runs before validation and returns a `CollectedEvidence` that owns its per-call assertion array, so concurrent handshakes cannot alias provider scratch storage and adding a real fetch/cache provider changes no TLS-handshake or path-validation signature.
- **Trust boundary** — unauthenticated evidence never counts as a completed check by default (`require_signature_verified`). The one exception is a *peer-stapled* revocation, which can only condemn the peer itself; an unauthenticated `revoked` from a fetched responder or CRL is refused, since trusting it would be a one-packet denial of service. Revocation is otherwise monotonic, except that a stale `certificateHold` does not outrank fresher evidence of its RFC 5280 §5.3.1 release.
- **[docs/PKI_REVOCATION.md](docs/PKI_REVOCATION.md)** — modes, trust boundary, freshness, and the privacy / availability / operational tradeoffs, plus what is deliberately not implemented yet.

## Test plan

- [x] `zig build test` — full suite green, including `src/pki/revocation_tests.zig` and end-to-end coverage in `src/pki/path_validator_tests.zig`
- [x] Stapled-good, stapled-revoked (with CRL reason), missing, stale, malformed, unauthenticated, unknown, must-staple, and unsupported-source cases across all four modes
- [x] Evidence identity: foreign evidence never applied by position in either direction; the same certificate consumes its own evidence at a different depth; a longer candidate's assertions cannot reject a shorter candidate, unit and end-to-end through `validateCandidates`
- [x] RFC 7633 §4.2.2 chain constraint: featureless child rejected even with a good staple present, child asserting a different feature rejected, same-set and superset children accepted
- [x] ClientHello intersection: feature 5 with `offered_status_request = false` accepts across every consulting mode and reports `not_offered_by_client`; the same certificate and absent staple with the flag set rejects; a certificate without the extension reports `not_required` either way
- [x] `status_request_v2`: a feature-17 leaf is accepted with `not_required` bare and with a staple; `{5,17}` reports `enforced` off an ordinary staple and still rejects when that staple is missing
- [x] Trust boundary: forged cached/CRL revocation cannot deny service, the peer's own unverified staple still self-revokes, a stale `certificateHold` loses to fresher evidence while a fresh hold and a stale permanent revocation still stand
- [x] `disabled` accepts despite oversized and misfiled evidence
- [x] Freshness boundaries: skew on both sides, future-dated evidence, absent `nextUpdate`, incoherent `nextUpdate < thisUpdate`, age bound
- [x] Provider contract: two simultaneously-live `CollectedEvidence` results, neither aliasing nor freeing the other's storage
- [x] Determinism oracle extended to cover the status report and new failure metadata; `FailingAllocator` sweep proves the status path is leak-free and degrades to a structured rejection
- [x] `zig build test-pki`, `test-pki-fuzz`, `test-pki-reduce`, `test-pki-openssl` green; `zig fmt --check` clean

Closes #349

### DIFF
--- a/src/pki/path_validator.zig
+++ b/src/pki/path_validator.zig
@@ -92,6 +92,7 @@ pub const FailureReason = enum {
     policy_constraints_invalid,
     inhibit_any_policy_invalid,
     certificate_policy_resource_limit_exceeded,
+    tls_feature_constraint_violation,
     certificate_revoked,
     revocation_status_unavailable,
     revocation_status_stale,
@@ -236,6 +237,10 @@ pub fn validatePath(
         if (!usage.digital_signature) return rejectOid(.key_usage_violation, 0, &wk.key_usage);
     }
 
+    if (checkTlsFeatureConstraints(path)) |validation_failure| {
+        return .{ .rejected = validation_failure };
+    }
+
     if (policy.require_server_auth_eku) {
         // A present EKU restricts every non-anchor certificate in the path;
         // absence is unrestricted. Trust-anchor purpose is configuration.
@@ -371,6 +376,39 @@ fn checkStructure(path: path_builder.Path, policy: ValidationPolicy) ?Validation
         }
     }
     return null;
+}
+
+/// RFC 7633 §4.2.2: a certificate-signing certificate carrying the TLS Feature
+/// extension constrains what it signs — every certificate it issues must assert
+/// the same feature set or a superset of it. This is a chain constraint on the
+/// extension's contents, not a stapling obligation that propagates downward:
+/// the operational must-staple requirement is read from the end-entity
+/// certificate, after this constraint holds (see `revocation.zig`).
+///
+/// The terminal trust anchor is excluded for the same reason its other
+/// extensions are: anchor restrictions are local trust configuration.
+fn checkTlsFeatureConstraints(path: path_builder.Path) ?ValidationFailure {
+    const anchor_index = path.elements.len - 1;
+    for (path.elements[1..anchor_index], 1..) |issuer, issuer_index| {
+        const required = issuer.certificate.tlsFeatures() orelse continue;
+        const child_index = issuer_index - 1;
+        const child = path.elements[child_index].certificate.tlsFeatures() orelse
+            return tlsFeatureFailure(child_index, issuer_index);
+        for (required.features) |feature| {
+            if (!child.contains(feature)) return tlsFeatureFailure(child_index, issuer_index);
+        }
+    }
+    return null;
+}
+
+fn tlsFeatureFailure(certificate_index: usize, issuer_index: usize) ValidationFailure {
+    return .{
+        .reason = .tls_feature_constraint_violation,
+        .certificate_index = certificate_index,
+        // Compile-time well-known OID, below the fixed component bound.
+        .extension_oid = oid.ObjectIdentifier.fromComponents(&wk.tls_feature) catch unreachable,
+        .constraint_certificate_index = issuer_index,
+    };
 }
 
 fn checkExtensions(certificate: *const x509.Certificate, certificate_index: usize) ?ValidationFailure {
--- a/src/pki/path_validator.zig
+++ b/src/pki/path_validator.zig
@@ -17,7 +17,11 @@
 //!   directoryName, dNSName, rfc822Name, URI, and IP forms;
 //! - certificate policies use the bounded RFC 9618 policy graph. The default
 //!   user policy is anyPolicy with no initial inhibition, preserving ordinary
-//!   validation compatibility while returning deterministic constrained sets.
+//!   validation compatibility while returning deterministic constrained sets;
+//! - certificate-status policy (revocation, OCSP, stapling, must-staple) is
+//!   evaluated from caller-supplied evidence only. It defaults to disabled and
+//!   records that fact in the accepted result rather than implying a check
+//!   happened. See `revocation.zig` for the seam and its trust boundary.
 
 const std = @import("std");
 const certificate_policy = @import("certificate_policy.zig");
@@ -26,6 +30,7 @@ const identity = @import("identity.zig");
 const name_constraints = @import("name_constraints.zig");
 const oid = @import("oid.zig");
 const path_builder = @import("path_builder.zig");
+const revocation = @import("revocation.zig");
 const verify = @import("verify.zig");
 const x509 = @import("x509.zig");
 
@@ -43,6 +48,14 @@ pub const ValidationPolicy = struct {
     maximum_extensions_per_certificate: usize = 64,
     name_constraints: name_constraints.Limits = .{},
     certificate_policy: certificate_policy.Configuration = .{},
+    /// Certificate-status policy. Defaults to disabled: no status is consulted
+    /// and the accepted result says so.
+    revocation: revocation.Configuration = .{},
+    /// Status evidence for this validation, collected by the caller before the
+    /// call. Borrowed. The validator performs no network or file I/O of any
+    /// kind, so a future fetch/cache provider fills this in without changing
+    /// this signature.
+    revocation_evidence: revocation.Evidence = .{},
     /// Anchors passed to `path_builder.build`, in the same order. Borrowed for
     /// the call only. The terminal element's input index and DER must match.
     trust_anchors: []const x509.Certificate,
@@ -79,6 +92,15 @@ pub const FailureReason = enum {
     policy_constraints_invalid,
     inhibit_any_policy_invalid,
     certificate_policy_resource_limit_exceeded,
+    certificate_revoked,
+    revocation_status_unavailable,
+    revocation_status_stale,
+    revocation_status_malformed,
+    revocation_status_unauthenticated,
+    revocation_must_staple_not_satisfied,
+    revocation_source_unsupported,
+    revocation_evidence_invalid,
+    revocation_resource_limit_exceeded,
     identity_mismatch,
     invalid_identity_reference,
     validation_resource_limit_exceeded,
@@ -98,6 +120,9 @@ pub const ValidationFailure = struct {
     policy_oid: ?oid.ObjectIdentifier = null,
     policy_graph_depth: ?usize = null,
     policy_stage: ?certificate_policy.Stage = null,
+    revocation_source: ?revocation.Source = null,
+    revocation_defect: ?revocation.Defect = null,
+    revocation_reason: ?revocation.CrlReason = null,
 };
 
 pub const AcceptedPath = struct {
@@ -106,6 +131,10 @@ pub const AcceptedPath = struct {
     accepted_path: []const path_builder.Element,
     /// Owned RFC 9618 policy outputs. Qualifiers are intentionally omitted.
     policies: certificate_policy.PolicyResult,
+    /// Owned record of what certificate-status evidence was checked, and what
+    /// was not. Present on every acceptance, including the default disabled
+    /// policy, so callers never have to assume revocation was verified.
+    revocation: revocation.Report,
 };
 
 pub const ValidationResult = union(enum) {
@@ -118,6 +147,8 @@ pub const ValidationResult = union(enum) {
                 allocator.free(accepted_value.accepted_path);
                 var policies = accepted_value.policies;
                 policies.deinit(allocator);
+                var report = accepted_value.revocation;
+                report.deinit(allocator);
             },
             .rejected => {},
         }
@@ -226,24 +257,47 @@ pub fn validatePath(
         .failure => |policy_failure| return .{ .rejected = certificatePolicyFailure(policy_failure) },
     };
 
+    // Certificate-status policy runs on caller-supplied evidence only; it
+    // never reaches a responder, a CRL endpoint, or the filesystem.
+    var status_report = switch (revocation.evaluatePath(
+        allocator,
+        path,
+        policy.revocation,
+        policy.revocation_evidence,
+        policy.validation_time,
+    )) {
+        .accepted => |report| report,
+        .rejected => |status_failure| {
+            policies.deinit(allocator);
+            return .{ .rejected = revocationFailure(status_failure) };
+        },
+    };
+
     // Identity is intentionally last: no path is accepted based on a name
     // match before its signatures and RFC 5280 policy checks succeed.
     if (policy.expected_dns_name) |expected| {
         const verdict = identity.verifyHost(leaf, expected) catch {
             policies.deinit(allocator);
+            status_report.deinit(allocator);
             return reject(.invalid_identity_reference, 0);
         };
         if (!verdict.isMatch()) {
             policies.deinit(allocator);
+            status_report.deinit(allocator);
             return reject(.identity_mismatch, 0);
         }
     }
 
     const owned = allocator.dupe(path_builder.Element, path.elements) catch {
         policies.deinit(allocator);
+        status_report.deinit(allocator);
         return rejectWithoutCertificate(.out_of_memory);
     };
-    return .{ .accepted = .{ .accepted_path = owned, .policies = policies } };
+    return .{ .accepted = .{
+        .accepted_path = owned,
+        .policies = policies,
+        .revocation = status_report,
+    } };
 }
 
 /// Validate candidates in builder order and return the first accepted path.
@@ -354,7 +408,10 @@ fn criticalExtensionHandled(extension_oid: *const oid.ObjectIdentifier) bool {
         extension_oid.eqlComponents(&wk.policy_constraints) or
         extension_oid.eqlComponents(&wk.inhibit_any_policy) or
         extension_oid.eqlComponents(&wk.subject_key_identifier) or
-        extension_oid.eqlComponents(&wk.authority_key_identifier);
+        extension_oid.eqlComponents(&wk.authority_key_identifier) or
+        // RFC 7633 permits a critical TLS Feature extension; `revocation.zig`
+        // now processes must-staple, so it is no longer unhandled.
+        extension_oid.eqlComponents(&wk.tls_feature);
 }
 
 fn certificatePolicyFailure(policy_failure: certificate_policy.Failure) ValidationFailure {
@@ -375,6 +432,28 @@ fn certificatePolicyFailure(policy_failure: certificate_policy.Failure) Validati
         .policy_oid = policy_failure.policy_oid,
         .policy_graph_depth = policy_failure.graph_depth,
         .policy_stage = policy_failure.stage,
+    };
+}
+
+fn revocationFailure(status_failure: revocation.Failure) ValidationFailure {
+    const reason: FailureReason = switch (status_failure.reason) {
+        .certificate_revoked => .certificate_revoked,
+        .status_unavailable => .revocation_status_unavailable,
+        .status_stale => .revocation_status_stale,
+        .status_malformed => .revocation_status_malformed,
+        .status_unauthenticated => .revocation_status_unauthenticated,
+        .must_staple_not_satisfied => .revocation_must_staple_not_satisfied,
+        .status_source_unsupported => .revocation_source_unsupported,
+        .evidence_index_invalid => .revocation_evidence_invalid,
+        .resource_limit_exceeded => .revocation_resource_limit_exceeded,
+        .out_of_memory => .out_of_memory,
+    };
+    return .{
+        .reason = reason,
+        .certificate_index = status_failure.certificate_index,
+        .revocation_source = status_failure.source,
+        .revocation_defect = status_failure.defect,
+        .revocation_reason = status_failure.revocation_reason,
     };
 }
 
--- a/src/pki/path_validator.zig
+++ b/src/pki/path_validator.zig
@@ -99,7 +99,6 @@ pub const FailureReason = enum {
     revocation_status_malformed,
     revocation_status_unauthenticated,
     revocation_must_staple_not_satisfied,
-    revocation_must_staple_feature_unsupported,
     revocation_source_unsupported,
     revocation_resource_limit_exceeded,
     identity_mismatch,
@@ -481,7 +480,6 @@ fn revocationFailure(status_failure: revocation.Failure) ValidationFailure {
         .status_malformed => .revocation_status_malformed,
         .status_unauthenticated => .revocation_status_unauthenticated,
         .must_staple_not_satisfied => .revocation_must_staple_not_satisfied,
-        .must_staple_feature_unsupported => .revocation_must_staple_feature_unsupported,
         .status_source_unsupported => .revocation_source_unsupported,
         // `checkStructure` already guarantees a well-formed candidate path, so
         // this is unreachable from `validatePath`; it stays mapped for direct
--- a/src/pki/path_validator.zig
+++ b/src/pki/path_validator.zig
@@ -99,8 +99,8 @@ pub const FailureReason = enum {
     revocation_status_malformed,
     revocation_status_unauthenticated,
     revocation_must_staple_not_satisfied,
+    revocation_must_staple_feature_unsupported,
     revocation_source_unsupported,
-    revocation_evidence_invalid,
     revocation_resource_limit_exceeded,
     identity_mismatch,
     invalid_identity_reference,
@@ -481,8 +481,12 @@ fn revocationFailure(status_failure: revocation.Failure) ValidationFailure {
         .status_malformed => .revocation_status_malformed,
         .status_unauthenticated => .revocation_status_unauthenticated,
         .must_staple_not_satisfied => .revocation_must_staple_not_satisfied,
+        .must_staple_feature_unsupported => .revocation_must_staple_feature_unsupported,
         .status_source_unsupported => .revocation_source_unsupported,
-        .evidence_index_invalid => .revocation_evidence_invalid,
+        // `checkStructure` already guarantees a well-formed candidate path, so
+        // this is unreachable from `validatePath`; it stays mapped for direct
+        // `revocation.evaluatePath` callers.
+        .malformed_path => .malformed_path,
         .resource_limit_exceeded => .revocation_resource_limit_exceeded,
         .out_of_memory => .out_of_memory,
     };
--- a/src/pki/revocation.zig
+++ b/src/pki/revocation.zig
@@ -1,0 +1,661 @@
+//! Certificate-status policy seam for CRLs, OCSP, stapling, and must-staple
+//! (#349).
+//!
+//! This module decides what a path's *certificate-status evidence* means. It
+//! never fetches anything: no socket, no file, no clock. Evidence arrives as
+//! caller-supplied `StatusAssertion` values, and the validator turns them into
+//! an explicit `Report` recording, per certificate, exactly what was checked
+//! and what was not. Nothing here silently pretends revocation was checked.
+//!
+//! ## Why the seam looks like this
+//!
+//! Online revocation fetching is not implemented yet. Rather than leave the
+//! question undefined, the policy is explicit and the default (`Mode.disabled`)
+//! says so in the result: every entry is `not_checked_policy_disabled`, and a
+//! must-staple certificate accepted under it sets `Report.must_staple_unenforced`.
+//! An operator can therefore tell a "no revocation check happened" acceptance
+//! from a "status was checked and is good" acceptance.
+//!
+//! A future OCSP/CRL fetch-and-cache provider produces `StatusAssertion`
+//! values before validation runs — see `Provider`. That keeps network I/O,
+//! caching, and responder-signature verification outside both the TLS
+//! handshake and the path-validation API, which is why adding one later
+//! changes no signature in `path_validator`.
+//!
+//! ## Trust boundary
+//!
+//! `StatusAssertion.signature_verified` is asserted by whoever produced the
+//! evidence, not by this module: verifying an OCSP responder signature needs
+//! its own delegated-responder path validation, and CRL verification needs the
+//! CRL issuer's key. With `Configuration.require_signature_verified` (the
+//! default), unverified evidence never counts as a completed check — a peer
+//! cannot manufacture a "good" status by stapling bytes nobody authenticated.
+//! A *revoked* assertion is honored even when unverified: the only party that
+//! can supply attacker-chosen stapled bytes is the peer being authenticated,
+//! and revoking itself is not an attack worth defending against.
+
+const std = @import("std");
+const path_builder = @import("path_builder.zig");
+const x509 = @import("x509.zig");
+
+/// Certificate-status policy. Ordered from least to most demanding.
+pub const Mode = enum {
+    /// No status is consulted. Explicitly recorded in the report; must-staple
+    /// is not enforced because no evidence channel exists in this mode.
+    disabled,
+    /// Only stapled evidence is consulted. Absent stapling is accepted (and
+    /// recorded); stapled evidence that is present but unusable is rejected.
+    stapled_only,
+    /// All supplied evidence is consulted. Revoked rejects; missing, stale,
+    /// defective, or unknown status is recorded and accepted.
+    soft_fail,
+    /// Every non-anchor certificate needs usable, good status evidence.
+    strict,
+};
+
+/// Where a status assertion came from. Precedence for selection is the
+/// declaration order: fresher, peer-delivered stapled evidence outranks a
+/// cache, which outranks a periodically distributed CRL set.
+pub const Source = enum {
+    stapled_ocsp,
+    cached_ocsp,
+    crl_set,
+};
+
+pub const Status = enum { good, revoked, unknown };
+
+/// RFC 5280 §5.3.1 CRLReason values, carried through for logging and for
+/// operators who treat `certificate_hold` differently from a key compromise.
+pub const CrlReason = enum(u8) {
+    unspecified = 0,
+    key_compromise = 1,
+    ca_compromise = 2,
+    affiliation_changed = 3,
+    superseded = 4,
+    cessation_of_operation = 5,
+    certificate_hold = 6,
+    remove_from_crl = 8,
+    privilege_withdrawn = 9,
+    aa_compromise = 10,
+};
+
+/// Why supplied evidence cannot be used at all. Reported by the producing
+/// provider; this module never decodes an OCSP response or a CRL.
+pub const Defect = enum {
+    malformed_encoding,
+    unsupported_response_type,
+    responder_not_authorized,
+    signature_invalid,
+    /// The evidence covers a different certificate than it was filed under.
+    identity_mismatch,
+    /// The responder answered with an error status (RFC 6960 §2.3).
+    responder_error,
+};
+
+/// One certificate-status assertion supplied by the caller. Timestamps are
+/// Unix seconds on the same scale as `validation_time`; all are optional
+/// because CRL sets and cache entries do not all carry every field.
+pub const StatusAssertion = struct {
+    /// Leaf-first index of the certificate this assertion covers.
+    certificate_index: usize,
+    source: Source,
+    status: Status = .unknown,
+    /// Set when the evidence could not be interpreted. A defective assertion
+    /// never contributes a status, whatever `status` says.
+    defect: ?Defect = null,
+    revocation_reason: ?CrlReason = null,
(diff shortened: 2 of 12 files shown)


src/pki/path_validator.zig
```diff
@@ -43,6 +48,14 @@ pub const ValidationPolicy = struct {
...
     name_constraints: name_constraints.Limits = .{},
     certificate_policy: certificate_policy.Configuration = .{},
+    /// Certificate-status policy. Defaults to disabled: no status is consulted
+    /// and the accepted result says so.
+    revocation: revocation.Configuration = .{},
+    /// Status evidence for this validation, collected by the caller before the
+    /// call. Borrowed. The validator performs no network or file I/O of any
+    /// kind, so a future fetch/cache provider fills this in without changing
+    /// this signature.
+    revocation_evidence: revocation.Evidence = .{},
```
Jaruso, line 58: **[P1] Bind status evidence to certificate identity before reusing it across candidate paths.**

`StatusAssertion` is keyed only by a leaf-first `certificate_index`, but `validateCandidates()` passes the same `policy.revocation_evidence` to every candidate. Alternate/cross-signed candidates can have different certificates at the same index. If evidence says index 1 is `good` for intermediate A, candidate A can fail a later RFC 5280 check and candidate B can then consume A's `good` assertion at index 1 and pass `strict` revocation. The inverse can falsely revoke B.

That is both a validation correctness bug and a violation of the acceptance criterion that a future provider can be added without changing this API: a single evidence set cannot safely describe multiple candidate paths by position alone.

Please bind each assertion to a stable certificate identity and verify it against the current path before using it (or make evidence explicitly per-candidate). For example:

```zig
pub const StatusAssertion = struct {
    certificate_index: usize,
    certificate_sha256: [32]u8,
    // ...
};

fn assertionMatches(assertion: StatusAssertion, cert: *const x509.Certificate) bool {
    var digest: [32]u8 = undefined;
    std.crypto.hash.sha2.Sha256.hash(cert.raw, &digest, .{});
    return std.mem.eql(u8, &assertion.certificate_sha256, &digest);
}
```

Then reject/ignore a mismatched binding rather than applying it by index. Please add an alternate-path test where candidate 1 and candidate 2 have different intermediates at the same index, evidence belongs to candidate 1, candidate 1 fails another validation rule, and candidate 2 must not inherit that evidence.

src/pki/revocation.zig
```diff
@@ -0,0 +1,661 @@
...
+    accepted: Report,
+    rejected: Failure,
+};
+
+/// Interface for a future fetch-and-cache status provider. Implementations run
+/// *before* validation and hand back evidence; the validator itself never
+/// calls one, which is what keeps ambient network I/O out of path validation.
+/// The returned assertions borrow storage the implementation owns and must
+/// outlive the validation call.
+pub const Provider = struct {
```
Jaruso, line 369: **[P1] The provider seam needs an ownership/release contract before it can satisfy the future-provider acceptance criterion.**

`collect()` receives an allocator, but its return value is documented as provider-owned borrowed storage and there is no `deinit`/`release` hook. A real provider will usually have to compose a per-validation assertion array from stapled data + cache + CRL data. If it allocates that array with the supplied allocator, nothing in this interface can free it. If it instead keeps one provider-owned scratch array, concurrent handshakes can overwrite/alias each other's evidence. Retaining every per-call array until provider teardown is an unbounded-growth workaround, not a viable contract.

Please make lifetime/ownership explicit now so adding the real provider does not require changing this seam. One workable shape is:

```zig
pub const CollectedEvidence = struct {
    evidence: Evidence,
    owned_assertions: []StatusAssertion,

    pub fn deinit(self: *CollectedEvidence, allocator: std.mem.Allocator) void {
        allocator.free(self.owned_assertions);
        self.* = undefined;
    }
};

// collect(...) error{OutOfMemory}!CollectedEvidence
```

The raw OCSP/CRL blobs can still borrow cache/provider storage if desired; the per-call assertion list needs a deterministic lifetime. A provider-vtable `release` callback would also work. Please cover two simultaneously-live collection results so the contract is safe for concurrent validations.

src/pki/revocation.zig
```diff
@@ -0,0 +1,661 @@
...
+    config: Configuration,
+    evidence: Evidence,
+    validation_time: i64,
+) Outcome {
+    if (path.elements.len < 2) {
+        return .{ .rejected = .{ .reason = .evidence_index_invalid, .certificate_index = null } };
+    }
+    const certificate_count = path.elements.len - 1;
+
+    if (evidence.assertions.len > config.limits.maximum_assertions) {
```
Jaruso: **[P2] `.disabled` can still reject a path based on revocation evidence.**

The mode contract says disabled consults nothing, but assertion-count, raw-size, and certificate-index validation all run before the disabled early return. That means a caller can leave the default policy disabled and still get `resource_limit_exceeded` or `evidence_index_invalid` solely because status data was present. In particular, evidence collected opportunistically from an untrusted peer can turn "revocation off" into fail-closed behavior.

Please construct the explicit disabled report and return it before evidence validation/selection. Conceptually:

```zig
const entries = try buildEntries(...);
if (config.mode == .disabled) {
    return .{ .accepted = .{
        .mode = .disabled,
        .entries = entries,
        .must_staple_unenforced = must_staple_required,
    } };
}

// Only modes that consult status validate/use evidence from here on.
```

Also add a regression test with disabled mode plus out-of-range/oversized assertions; it should still accept and report `not_checked_policy_disabled`.

src/pki/revocation.zig
```diff
@@ -0,0 +1,661 @@
...
+        if (assertion.certificate_index >= certificate_count) {
+            return .{ .rejected = .{
+                .reason = .evidence_index_invalid,
+                .certificate_index = assertion.certificate_index,
+                .source = assertion.source,
+            } };
+        }
+    }
+
+    // RFC 7633 §4.2.3: a CA asserting the feature obliges the certificates it
```
Jaruso: **[P1] This implements the CA form of RFC 7633 incorrectly.**

RFC 7633 §4.2.2 says a Certificate Signing Certificate carrying TLS Feature constrains each certificate it signs: the child **must itself contain the same set or a superset** of the issuer's features. It does not say that seeing feature 5/17 anywhere in the CA chain directly turns the leaf into a must-staple certificate. See: https://www.rfc-editor.org/rfc/rfc7633.html#section-4.2.2

With the current code, `Intermediate=[status_request]`, `Leaf=[]` can be accepted if the leaf happens to provide a good staple, even though the chain violates the CA's extension constraint. Conversely, a CA feature other than 5/17 is ignored entirely, even though the same/superset constraint applies to every advertised feature. This is especially important because `path_validator` now declares TLS Feature a handled critical extension.

Please validate the issuer/child feature-set constraint during path validation, then derive the operational must-staple requirement from the end-entity certificate after that constraint succeeds. For example:

```zig
for (path.elements[1..anchor_index], 1..) |issuer, issuer_index| {
    if (issuer.certificate.tlsFeatures()) |required| {
        const child = path.elements[issuer_index - 1].certificate.tlsFeatures() orelse
            return rejectTlsFeatureConstraint(issuer_index - 1);
        for (required.features) |feature| {
            if (!child.contains(feature))
                return rejectTlsFeatureConstraint(issuer_index - 1);
        }
    }
}

const must_staple_required = path.elements[0].certificate.mustStaple();
```

Please add tests for (1) CA feature 5 + child missing feature => chain rejected even if a staple is present, (2) CA arbitrary feature + child missing it => rejected, and (3) child same/superset => normal leaf feature processing.

src/pki/revocation.zig
```diff
@@ -0,0 +1,661 @@
...
+            };
+            allocator.free(entries);
+            return .{ .rejected = failure };
+        }
+    }
+
+    return .{ .accepted = .{
+        .mode = config.mode,
+        .entries = entries,
+        .must_staple_unenforced = false,
```
Jaruso: **[P2] The report says must-staple was enforced when the configuration explicitly disabled it.**

If a leaf asserts must-staple, `mode = .soft_fail`, `enforce_must_staple = false`, and no staple is present, this path reaches this return and sets `must_staple_unenforced = false`. The acceptance therefore reports the opposite of what happened. The existing test for disabling enforcement only checks that the path accepts; it misses the reporting regression.

Please make the report encode explicit non-enforcement too, e.g.:

```zig
.must_staple_unenforced = must_staple_required and !config.enforce_must_staple,
```

(or replace the bool with an enum that distinguishes `not_required`, `enforced`, `disabled_by_status_mode`, and `disabled_by_configuration`). Then assert the report value in the existing `must-staple enforcement can be disabled explicitly` test.

src/pki/revocation.zig
```diff
@@ -0,0 +1,661 @@
...
+                blocked.blocker = .defective;
+                blocked.defect = defect;
+            }
+            continue;
+        }
+
+        // Revocation is honored even from stale or unauthenticated evidence:
+        // ignoring it could only ever help a certificate its own issuer has
+        // withdrawn. See the trust-boundary note at the top of this file.
+        if (assertion.status == .revoked) {
```
Jaruso: **[P1] A `revoked` status is not safe to trust before authentication/freshness for every source.**

This short-circuit applies to `cached_ocsp` and `crl_set` as well as peer-stapled data, but the trust-boundary rationale only justifies the self-stapled case. A future HTTP OCSP/CRL provider can parse an attacker-supplied response as `revoked` while `signature_verified == false`; this code hard-rejects it before the authentication rule, giving an on-path attacker a trivial denial of service.

There is a second correctness problem with bypassing freshness universally: RFC 5280 explicitly allows `certificateHold` to be released (`removeFromCRL`), so a stale authenticated hold is not monotonic and must not forever override a newer good/released status. RFC 5280 describes release-from-hold as a status change: https://www.rfc-editor.org/rfc/rfc5280#section-5.3.1

Please make the exception source/reason aware. At minimum, authenticate non-stapled evidence before trusting any status, and do not let stale `certificate_hold` override newer evidence. For example:

```zig
const self_stapled_revocation =
    assertion.source == .stapled_ocsp and assertion.status == .revoked;

if (config.require_signature_verified and
    !assertion.signature_verified and
    !self_stapled_revocation)
{
    // record unauthenticated blocker
    continue;
}

if (assertion.status == .revoked) {
    if (assertion.revocation_reason == .certificate_hold and
        !isFresh(assertion, config.freshness, validation_time))
    {
        // stale hold cannot be sticky; let fresher evidence decide
        continue;
    }
    return revokedSelection(assertion);
}
```

Please add tests that unauthenticated cached/CRL `revoked` cannot produce `certificate_revoked` under the default trust boundary, and that a stale `certificate_hold` does not beat a fresh authenticated good/release indication.

Jaruso (commented):
Changes requested before merge. I reviewed all 12 changed files against #349 and its dependencies (#341, #345, #346 are closed), and CI for this head is green.

Acceptance-criteria status:
- Explicit default/reporting: largely implemented, but disabled mode can still reject based on evidence and the must-staple report can misstate explicit non-enforcement.
- Required status cases: coverage is broad, but the missing cases below expose correctness gaps (alternate paths, CA TLS Feature propagation, unauthenticated non-stapled revocation, temporary hold/release, disabled mode with bad evidence).
- Result records checked evidence: structurally yes, but evidence is position-bound rather than certificate-bound, so the report can describe evidence for the wrong certificate on alternate paths.
- No ambient network I/O: satisfied.
- Future fetch/cache provider without API changes: not satisfied yet because evidence identity and provider ownership/lifetime are not safe for real multi-candidate/concurrent use.

The implementation direction is good and the test work is substantial, but the certificate-status seam is a security boundary; the issues inline need to be fixed before this can safely close #349.

src/pki/path_validator.zig
Jaruso: Fixed in 9240f5cc. `StatusAssertion` now carries a `CertificateIdentity` (SHA-256 over the exact DER), and `select()` matches on identity, not position — a mismatched assertion is ignored rather than applied. Added `an alternate candidate cannot inherit another candidate's status evidence` in `path_validator_tests.zig`: two intermediates with the same subject and key, candidate A expired, evidence bound to A. Candidate B validated in isolation rejects (its index-1 certificate has no evidence of its own), and re-binding the same evidence to B's intermediate accepts — which proves the rejection was the identity check rather than an unrelated failure.

src/pki/revocation.zig
Jaruso: Fixed in 9240f5cc. `collect` now returns a `CollectedEvidence` that owns its per-call assertion array (`deinit(allocator)`, `evidence()` to borrow it for validation); only the encoded `raw` blobs may borrow longer-lived cache storage. Added `collected evidence owns its assertions and survives concurrent collections`, which drives a stub provider that composes a per-path array, keeps two results live at once, validates against both, and asserts the first is untouched by the second.

src/pki/revocation.zig
Jaruso: Fixed in 9240f5cc. The disabled report is built and returned before any evidence is examined — assertion-count, raw-size, and index validation now run only in modes that consult status. Added `disabled mode never rejects on evidence, however malformed`: out-of-range index, oversized `raw`, and a defective foreign assertion, all against limits of 1 assertion / 8 bytes, still accept with `not_checked_policy_disabled`.

src/pki/revocation.zig
Jaruso: Fixed in 9240f5cc. Replaced the bool with `MustStapleOutcome` — `not_required`, `enforced`, `unenforced_status_disabled`, `unenforced_by_configuration` — so both unenforced cases are distinct and an acceptance never reads as honored. The `must-staple enforcement can be disabled explicitly` test now asserts `unenforced_by_configuration`, and the other must-staple tests assert their expected outcome too.

src/pki/revocation.zig
Jaruso: Fixed in 9240f5cc — you were right, I had conflated the two forms. The issuer form is now a chain constraint in `path_validator.checkTlsFeatureConstraints`: every certificate a TLS-Feature-carrying CA signs must assert the same set or a superset, for *every* advertised feature, enforced always and independently of the status mode (new `tls_feature_constraint_violation`, carrying the constraining issuer's index). `must_staple_required` is now read from the end-entity certificate alone, after that constraint holds. Tests: CA feature 5 + featureless child rejects even with a good staple present; CA feature + child asserting a *different* feature rejects; same-set and superset children accept and then get normal leaf processing. The old `an issuing CA can assert must-staple...` test was inverted and is replaced by one asserting the CA form is not a leaf obligation.

src/pki/revocation.zig
Jaruso: Fixed in 9240f5cc. The unauthenticated exception is now narrowed to a peer-stapled revocation only — a `revoked` from `cached_ocsp` or `crl_set` must be authenticated before it is trusted, so a forged responder answer is a blocked `unauthenticated` assertion rather than a hard rejection. Separately, a stale `certificate_hold` is now treated as unusable so fresher evidence of an RFC 5280 §5.3.1 release decides; a fresh hold still revokes and a stale permanent revocation still stands. Tests cover all four: forged cached/CRL revocation cannot deny service (soft-fail records it, strict rejects as `status_unauthenticated`), the peer's own unverified staple still self-revokes, a stale hold loses to a fresh good, and a stale hold alone is `status_stale` rather than a revocation.

src/pki/revocation.zig
```diff
@@ -0,0 +1,796 @@
...
+    for (evidence.assertions) |assertion| {
+        if (assertion.raw.len > config.limits.maximum_evidence_bytes) {
+            allocator.free(entries);
+            return .{ .rejected = .{
+                .reason = .resource_limit_exceeded,
+                .certificate_index = assertion.certificate_index,
+                .source = assertion.source,
+            } };
+        }
+        if (assertion.certificate_index >= certificate_count) {
```
Jaruso: **[P1] Certificate identity is now stable, but the stored path index can still poison alternate candidates.**

This check runs over the *entire shared evidence set* before identity matching. `validateCandidates()` intentionally reuses one `revocation_evidence` value for every candidate, but `certificate_index` is path-local. A perfectly valid assertion for a longer candidate can therefore reject a shorter candidate as `evidence_index_invalid` before that shorter candidate's own matching assertions are considered.

Concrete example:

```text
candidate A: leaf -> A1 -> A2 -> root   // non-anchor indexes 0,1,2
candidate B: leaf -> B1 -> root         // non-anchor indexes 0,1

evidence union:
  leaf good            index 0
  A1 good              index 1
  A2 good              index 2
  B1 good              index 1
```

If A fails some unrelated path rule and validation advances to B, the `A2` assertion hits this branch (`2 >= 2`) and B is rejected even though B has complete status evidence. There is a related false-negative in `select()`: it requires both identity *and* the original index, so the same exact certificate appearing at a different depth in another cross-signed/rollover path cannot consume its own evidence.

Now that `CertificateIdentity` exists, it should be authoritative. The simplest shape is to remove `certificate_index` from `StatusAssertion` entirely (or retain it only as diagnostic/advisory metadata), match assertions by identity, and derive the report/failure index from the path currently being evaluated:

```zig
fn select(
    evidence: Evidence,
    certificate: *const x509.Certificate,
    config: Configuration,
    validation_time: i64,
) Selection {
    const identity = CertificateIdentity.of(certificate);
    for (evidence.assertions) |assertion| {
        if (!assertion.certificate.eql(identity)) continue;
        if (!consults(config.mode, assertion.source)) continue;
        // ...
    }
}
```

Then `evaluatePath()` should not reject foreign identities merely because their old path index is outside the current candidate; those assertions belong to other candidates and are expected in a shared set. Keep the global count/raw-size bounds, but make candidate applicability identity-based.

Please add an end-to-end regression with a longer candidate A and shorter candidate B, one union evidence set containing A's index-2 assertion plus complete B evidence, A failing another validation rule, and `validateCandidates()` accepting B. A second case where the same certificate moves depth between candidates would lock down the remaining index coupling too.

src/pki/x509.zig
```diff
@@ -352,6 +353,28 @@ pub const DistributionPoint = struct {
...
+    features: []const u16,
+
+    pub fn contains(self: *const TlsFeatures, feature: u16) bool {
+        return std.mem.indexOfScalar(u16, self.features, feature) != null;
+    }
+
+    /// RFC 7633 §4.2.1: asserting `status_request` obliges the server to
+    /// deliver a stapled certificate-status response.
+    pub fn requiresStapledStatus(self: *const TlsFeatures) bool {
+        return self.contains(status_request) or self.contains(status_request_v2);
```
Jaruso: **[P2] `status_request_v2` cannot be treated as equivalent to ordinary OCSP stapling in this TLS 1.3-only stack.**

Feature values in RFC 7633 name specific TLS extensions. This helper collapses `status_request` (5) and `status_request_v2` (17) into the same must-staple predicate, while `revocation.Source.stapled_ocsp` carries no mechanism/version information. As a result, a leaf that asserts **only feature 17** can receive an ordinary TLS 1.3 `status_request` OCSP response and be reported as `MustStapleOutcome.enforced`.

RFC 8446 §4.4.2.1 explicitly obsoletes RFC 6961 for TLS 1.3: `status_request_v2` is deprecated, and TLS 1.3 servers **MUST NOT** act on or send that extension. Tardigrade's native TLS path is TLS 1.3-only, so a v1-style staple cannot prove compliance with a feature-17 declaration. See https://www.rfc-editor.org/rfc/rfc8446.html#section-4.4.2.1.

Please make the protocol distinction explicit. Two reasonable approaches are:

```zig
pub fn requiresStapledStatus(self: *const TlsFeatures) bool {
    // TLS 1.3 can satisfy status_request (5), not RFC 6961 v2 (17).
    return self.contains(status_request);
}

pub fn requiresUnsupportedStatusRequestV2(self: *const TlsFeatures) bool {
    return self.contains(status_request_v2);
}
```

and reject/report feature 17 as unsupported for this TLS profile; or extend the evidence/handshake plumbing so the report can distinguish which advertised TLS feature was actually negotiated/satisfied. Either way, do not report `enforced` for feature 17 merely because `.stapled_ocsp + .checked_good` exists.

Please add a regression where the leaf contains only `{17}` and a normal good `.stapled_ocsp` assertion is supplied: the result must **not** be `MustStapleOutcome.enforced`. Also keep a `{5}` case proving ordinary TLS 1.3 stapling still satisfies must-staple.

Jaruso (commented):
Second pass on 9240f5cc. I re-checked the linked #349 acceptance criteria, every file changed by the follow-up commit, the six prior review threads, and CI (run 1255 is green).

The follow-up correctly fixes the provider ownership contract, disabled-mode rejection, RFC 7633 issuer/child constraint, must-staple outcome reporting, and the unauthenticated/stale revocation trust-boundary issue. Certificate identity binding also fixes the same-index/different-certificate false acceptance I called out previously, but the implementation still uses the path-local index as part of assertion validity/matching, so the evidence set is not actually reusable across candidate paths of different shape. I left that as a P1 below.

I also found one TLS 1.3-specific correctness issue in the original TLS Feature handling: `status_request_v2` (17) is treated as equivalent to `status_request` (5) and a generic staple can report it as enforced. RFC 8446 §4.4.2.1 obsoletes RFC 6961 for TLS 1.3 and says servers MUST NOT act on/send `status_request_v2`, so this TLS-1.3-only stack cannot claim feature 17 was satisfied by ordinary status_request stapling. I left a P2 with concrete options/tests.

Non-blocking documentation cleanup: the PR description itself is now stale after 9240f5cc. It still says the CA form "binds the leaf", still uses the old `must_staple_unenforced` wording, and still says a revoked assertion is honored even when unverified/stale without the new source/reason qualifications. Please update the PR body/test-plan bullets so they match the corrected implementation before merge.

With the two inline findings below, I still would not merge this yet.

src/pki/revocation.zig
Jaruso: Fixed in 5d18748b — you were right that identity alone should be authoritative. `certificate_index` is removed from `StatusAssertion` entirely rather than kept advisory, so there is no position-derived key left to poison anything: `select()` matches on identity, `evaluatePath` no longer index-checks the shared set, and the report/failure index comes from the path being evaluated. `evidence_index_invalid` is gone with it (the <2-element guard now reports `malformed_path`), and the anchor case is simply an identity that never matches a prospective certificate. Three regressions: `a longer candidate's evidence cannot invalidate a shorter candidate` (union set with A's depth-2 assertion, shorter candidate B accepts with complete evidence), `the same certificate consumes its own evidence at any depth`, and end-to-end `a longer candidate's evidence cannot reject a shorter candidate end to end` through `validateCandidates` where candidate A fails on validity and B is accepted.

src/pki/x509.zig
Jaruso: Fixed in 5d18748b. `requiresStapledStatus()` now covers feature 5 only, with `assertsStatusRequestV2()` reporting the RFC 6961 declaration separately. A leaf declaring feature 17 is rejected with `must_staple_feature_unsupported` in any mode that consults status, and recorded as `MustStapleOutcome.unsatisfiable_status_request_v2` under `disabled` (which must never reject). I went with fail-closed rather than ignoring feature 17: a certificate demanding a mechanism RFC 8446 §4.4.2.1 forbids this stack from negotiating cannot have its assertion honored, and silently accepting is the exact failure mode this policy exists to prevent. `{5,17}` is rejected too — 17 stays unsatisfiable regardless of what else is declared. Regressions cover feature-17-only with a good staple (not `enforced`), `{5,17}`, `{5}` still satisfied by ordinary TLS 1.3 stapling, and the disabled-mode recording.

src/pki/revocation.zig
```diff
@@ -0,0 +1,813 @@
...
+            allocator.free(entries);
+            return .{ .rejected = failure };
+        }
+    }
+
+    var must_staple: MustStapleOutcome = .not_required;
+    if (must_staple_required or unsatisfiable_feature) {
+        if (!config.enforce_must_staple) {
+            must_staple = .unenforced_by_configuration;
+        } else if (unsatisfiable_feature) {
```
Jaruso: **[P2] Do not reject a TLS 1.3 certificate solely because it advertises `status_request_v2`.**

This is still too strong, but for the opposite reason from the previous implementation. RFC 8446 §4.4.2.1 says a TLS 1.3 server must not act on/send `status_request_v2`; it does **not** make a certificate containing feature 17 invalid. RFC 7633 is explicit about the client side:

- §4.1: a client is not required to offer or support a TLS feature merely because the certificate names it.
- §4.3.3: invalidity is based on features present in **both** the client's `ClientHello` and the certificate that the server then fails to satisfy.

So for this TLS-1.3-only client, if we do not offer `status_request_v2` (which we should not need to), feature 17 is outside that intersection and must not by itself fail path validation. The current branch turns a perfectly usable TLS-1.3 certificate that also supports v2 for TLS 1.2 peers into `must_staple_feature_unsupported`.

The simplest fix for this story is to keep feature 17 parsed (and keep it in the RFC 7633 §4.2.2 CA same/superset constraint), but make end-entity must-staple enforcement depend only on feature 5:

```zig
const must_staple_required = path.elements[0].certificate.mustStaple(); // status_request (5)

// Do not reject solely for status_request_v2 (17). It is not offered/acted on
// by this TLS 1.3 profile, so it creates no end-entity obligation here.
if (must_staple_required) {
    if (!config.enforce_must_staple) {
        must_staple = .unenforced_by_configuration;
    } else {
        const leaf_entry = entries[0];
        if (leaf_entry.source != .stapled_ocsp or
            leaf_entry.determination != .checked_good)
        {
            // must_staple_not_satisfied
        }
        must_staple = .enforced;
    }
}
```

If the long-term goal is to implement RFC 7633 for arbitrary TLS features, carry the actual ClientHello feature set into this policy and enforce the intersection described by §4.3.3; that would also make the handshake dependency explicit instead of inferring it from certificate contents alone.

Please invert the new v2 regressions: `{17}` under TLS 1.3 should not fail solely because of 17, and `{5,17}` with a valid ordinary `status_request` staple should satisfy feature 5 rather than be rejected because 17 is also present. The CA same/superset test for 17 should remain.

Jaruso (commented):
Re-review of `5d18748b` complete. The identity-only evidence fix is correct: the path-local index is gone, sibling-candidate evidence is skipped by identity, the same certificate can consume its evidence at a different depth, and the new unit/end-to-end regressions cover the exact failure modes from the previous pass. The provider ownership, disabled-mode behavior, issuer TLS-Feature constraint, must-staple reporting, and authenticated-revocation fixes also remain intact. CI is green.

I found one remaining standards/interoperability issue in the new `status_request_v2` handling. RFC 8446 makes v2 unusable as a TLS 1.3 response mechanism, but RFC 7633 explicitly does *not* require a client to offer/support every feature named in the certificate and only makes the certificate invalid when a feature is present in both the client's ClientHello and the certificate and the server fails to satisfy it. A TLS 1.3-only client that does not offer feature 17 therefore should not reject a certificate merely because the certificate contains 17. I left the detailed fix/test shape inline.

Because this connected GitHub identity is the PR author, this is a COMMENT review rather than GitHub's formal REQUEST_CHANGES state. I would still hold merge for the remaining feature-17 behavior.

src/pki/revocation.zig
Jaruso: Fixed in ffa53937 — your reading of RFC 7633 §§4.1/4.3.3 is correct and I over-corrected. A certificate advertising feature 17 is no longer rejected: this TLS 1.3 profile never offers `status_request_v2`, so it is outside the ClientHello∩certificate intersection and creates no end-entity obligation. `must_staple_feature_unsupported` and the `unsatisfiable_status_request_v2` outcome are removed along with it, since nothing produces them now.

Feature 17 stays parsed, stays reported by `assertsStatusRequestV2` (documented as the hook for a future implementation that carries the negotiated ClientHello feature set into policy and enforces §4.3.3 properly), and still participates in the §4.2.2 issuer same/superset constraint, which iterates every advertised feature.

Tests inverted as asked: `{17}` alone now accepts with `MustStapleOutcome.not_required` both bare and with a staple; `{5,17}` with an ordinary `status_request` staple reports `enforced`, and dropping that staple still rejects with `must_staple_not_satisfied` so feature 5 enforcement is proven intact. The §4.2.2 test keeps its superset `{5,17}` case and the disjoint case where a child asserts a different feature than its issuer requires.

src/pki/revocation.zig
```diff
@@ -0,0 +1,792 @@
...
+    // during path validation (`path_validator.checkTlsFeatureConstraints`), not
+    // an assertion that propagates down as a stapling obligation of its own.
+    //
+    // Only `status_request` (5) creates that obligation. RFC 6961
+    // `status_request_v2` (17) is not offered by this TLS 1.3 profile, and
+    // RFC 7633 §§4.1/4.3.3 scope the certificate's demand to features present
+    // in *both* the ClientHello and the certificate — so a certificate that
+    // also advertises 17 for TLS 1.2 peers is perfectly usable here and must
+    // not be rejected over it.
+    const must_staple_required = path.elements[0].certificate.mustStaple();
```
Jaruso: **[P1] Apply the ClientHello-intersection rule to feature 5 too.**

The feature-17 correction is right, but the exact RFC 7633 §§4.1/4.3.3 rule cited immediately above applies to *every* TLS feature, including `status_request` (5): a client treats the certificate as invalid only when a feature is present in both the certificate and the client's ClientHello and the server fails to satisfy it.

Today Tardigrade's native TLS 1.3 `sendClientHello()` does **not** emit extension 5 at all. Its extension list contains `supported_versions`, `supported_groups`, `signature_algorithms`, `key_share`, ALPN/SNI/transport, and the early-data/PSK extensions as applicable, but there is no `status_request`. That means this unconditional `certificate.mustStaple()` requirement can reject a feature-5 certificate under `.stapled_only`/`.soft_fail`/`.strict` even though this client never requested the feature on that connection.

Please make the handshake intersection explicit instead of special-casing only 17. For example, carry the offered TLS-feature set (or at minimum `offered_status_request: bool`) into the status policy:

```zig
pub const Configuration = struct {
    // ...
    offered_status_request: bool = false,
};

const must_staple_required = config.offered_status_request and
    path.elements[0].certificate.mustStaple();
```

Then the TLS caller sets that bit from the actual ClientHello it emitted. Alternatively, wire `status_request` into this TLS profile now and still carry/derive that negotiated fact so direct path-validation callers do not silently assume it.

Please add regressions for both sides: feature 5 + `offered_status_request = false` must not fail solely for missing stapling, while feature 5 + `offered_status_request = true` must still require a usable staple. This also makes the future TLS-handshake/provider seam truthful rather than encoding a handshake fact the validator cannot currently know.

Jaruso (commented):
One remaining change requested before merge. The prior feature-17 finding is fixed correctly in `ffa53937`, and I rechecked the follow-up diff, the earlier review fixes, the #349 acceptance criteria, and the current CI run (green).

The remaining issue is the other half of the same RFC 7633 §4.3.3 rule: end-entity feature enforcement must be scoped to the features actually present in the client's ClientHello. The current native TLS 1.3 ClientHello emits supported_versions, supported_groups, signature_algorithms, key_share, ALPN/SNI/transport/early-data/PSK as applicable, but no `status_request` extension. Therefore the present unconditional feature-5 check in `revocation.evaluatePath()` can reject a must-staple certificate even though this client never requested feature 5 on that connection.

Acceptance-criteria status otherwise looks complete: default/reporting is explicit; required status cases are covered; accepted results record consulted evidence; no ambient validator I/O is introduced; evidence identity/provider ownership now support the future provider seam safely. Once ClientHello feature context is represented (or status_request is actually wired and the policy is driven by that fact), I do not see another blocker in this PR.

src/pki/revocation.zig
Jaruso: Fixed in 57612bb1. `Configuration.offered_status_request` now carries the handshake fact into policy and gates the end-entity obligation:

```zig
const must_staple_asserted = path.elements[0].certificate.mustStaple();
const must_staple_required = must_staple_asserted and config.offered_status_request;
```

It defaults to **false**, which matches what `sendClientHello()` actually emits today — I confirmed there is no `status_request` anywhere in `src/tls`. So a must-staple certificate on a connection that never asked for stapling is now accepted rather than rejected.

Rather than fold that into `not_required`, it reports a distinct `MustStapleOutcome.not_offered_by_client`, so the assertion stays visible in the record instead of disappearing — same reasoning as the other unenforced states. `Entry.must_staple` still reports that the certificate carries the extension.

Regressions both ways: feature 5 with `offered_status_request = false` accepts across `stapled_only`/`soft_fail`/`strict` and reports `not_offered_by_client` (strict still wants ordinary status evidence, asserted separately so the must-staple obligation is the only thing under test); the same certificate and the same absent staple with the flag set rejects as `must_staple_not_satisfied`, and with a good staple reports `enforced`. A second test pins that a certificate *without* the extension reports `not_required` whether or not the client offered, so `not_offered_by_client` can only ever describe a real assertion. The validator-level fixtures now set the flag explicitly too.

Jaruso (commented):
Re-review complete on 57612bb1. No remaining findings.

The last RFC 7633 issue is fixed correctly: `Configuration.offered_status_request` now carries the actual ClientHello fact, defaults to `false` to match the current TLS 1.3 backend, and gates the feature-5 end-entity obligation. The report preserves the certificate assertion with `not_offered_by_client` rather than collapsing it into `not_required`, while `offered_status_request = true` still requires a usable stapled-good response. The validator-level and unit regressions cover both sides of that boundary.

I also rechecked the full #349 acceptance criteria against the current implementation:
- default behavior is explicit/documented and reports unchecked status;
- stapled good/bad, missing, stale, malformed, must-staple, unsupported, authentication, and freshness cases are covered;
- accepted validation results carry per-certificate status evidence/reporting;
- validator code performs no ambient network I/O;
- the provider ownership/identity seam and `ValidationPolicy.revocation` configuration allow future fetch/cache wiring without changing path-validation signatures.

The previous alternate-path identity, provider ownership, disabled-mode, RFC 7633 issuer constraint, reporting, revocation-authentication, feature-17, and ClientHello-intersection findings are all resolved in the current code. CI for this head is green.

From this review, PR #590 is ready to merge.

PR #591, opened 2026-08-09
tls: [325-H] support TLS-over-TCP KeyUpdate

Jaruso:
## Summary

Implements RFC 8446 §4.6.3 post-handshake `KeyUpdate` for record-mode TLS, advancing one direction's application traffic secret at a time along §7.2's `"traffic upd"` chain.

- **Transition points are where the RFC puts them, and event ordering enforces them.** Receiving keys advance after the peer's `KeyUpdate` has been processed; sending keys advance only *after* our own has been sealed — so the message announcing a transition stays readable under the generation it retires. Each transition restarts that direction's record sequence at zero and zeroizes the keys and secret it replaced.
- **No update loops.** `update_requested` draws exactly one reciprocal update, which always carries `update_not_requested`. Simultaneous updates crossing on the wire converge at two generations per direction and stop.
- **Record-mode only.** RFC 9001 §6 removes `KeyUpdate` from TLS-over-QUIC, which carries its own key phase in the packet header, so a QUIC-profile engine neither offers a way to send one nor accepts one that arrives.
- **Usage thresholds are a hook, not a trigger.** `PureZigRecordStream.requestKeyUpdate` initiates one locally; `setKeyUpdateLimits`/`keyUpdateDue` expose a records-sealed threshold. Nothing fires an update on its own, and it is disabled by default, so existing connections' wire behaviour is unchanged until a caller opts in.

New module [key_update.zig](src/tls/key_update.zig) owns the wire form and the proactive-update policy; the secret step is `KeySchedule.nextTrafficSecret`; applying it to live keys is `Bridge.updateTrafficSecret`.

## Defect found and fixed while building this

Post-handshake reassembly always allocated, and the allocator is only installed by a client that configured a session-ticket consumer. Since **any** peer may send a `KeyUpdate` on **any** completed connection, an endpoint that simply did not use tickets would have failed the connection the moment a conforming peer rotated its keys.

Messages that fit inline (today, exactly `KeyUpdate`) now reassemble without an allocator, and a `KeyUpdate` declaring any other body length is rejected as a framing error *before* a buffer is chosen — so the failure class no longer depends on whether tickets were configured, and an attacker-chosen length is never buffered for a five-byte message.

## Test plan

- [x] `zig build test` — full suite green, including the crypto-boundary audit
- [x] `zig build test -Dtls-profile=appliance`, `test-integration-native-tls`, `test-tls-interop-matrix`
- [x] One-way, reciprocal, repeated, fragmented (every split point), and simultaneous updates over real socket pairs with two real engines
- [x] Chain arithmetic checked against an independent HKDF computation, not against the implementation
- [x] Sequence restarts at zero; old-generation records fail to open; a direction that skipped an update cannot read the next generation
- [x] No retired secret, key, or IV survives anywhere in the bridge's footprint after an update or teardown
- [x] Deterministic rejection of illegal request values (`illegal_parameter`, alert observed on the peer), wrong-length and over-long bodies (`decode_error`, identical with and without a post-handshake allocator), pre-handshake updates in both roles, and QUIC-mode messages
- [x] **OpenSSL 3.6.2 interop in both roles** — 3 new matrix rows, all passing:
  ```
  record/server/openssl/key-update/reciprocal   PASS  key_updates_sent=1 key_updates_received=1
  record/client/openssl/key-update/reciprocal   PASS  key_updates_sent=1 key_updates_received=1
  record/client/openssl/key-update/one-way      PASS  key_updates_sent=1 key_updates_received=0
  ```
- [x] **Mutation-checked**: inverting the emit ordering so the `KeyUpdate` is sealed under the *new* keys fails 7 unit tests and is rejected by OpenSSL with a fatal alert — the rows bite rather than passing on an ignored message.

Closes #357

### DIFF
--- a/src/tls/key_update.zig
+++ b/src/tls/key_update.zig
@@ -80,22 +80,44 @@ pub fn encode(request: Request, out: []u8) EncodeError![]const u8 {
 /// owner of the write state calls; whether to act on it is that owner's
 /// policy, which keeps this module free of any transport concern.
 pub const UsageLimits = struct {
-    /// Advance sending keys once this many records have been sealed under the
-    /// current generation. `null` disables proactive updates entirely, which
-    /// is the default: an endpoint is always free to never send a `KeyUpdate`,
-    /// and turning it on by default would change the wire behavior of every
-    /// existing connection.
+    /// The most records one generation of sending keys may protect, *counting
+    /// the `KeyUpdate` that retires it*. `null` disables proactive updates
+    /// entirely, which is the default: an endpoint is always free to never
+    /// send a `KeyUpdate`, and turning it on by default would change the wire
+    /// behavior of every existing connection.
+    ///
+    /// This is a cap on the generation, not a count at which to start
+    /// thinking about one — see `reached`, which fires a record early so that
+    /// the `KeyUpdate` fits underneath the cap rather than exceeding it.
     records: ?u64 = null,
 
     /// RFC 8446 §5.5's AES-GCM record limit (2^24.5, rounded down to a whole
     /// number of records). Callers opting into proactive updates should pick a
     /// value at or below this; it is offered as a named constant rather than a
     /// default so the choice stays explicit at the call site.
+    ///
+    /// Safe to pass straight to `records`: `reached` reserves the control
+    /// record, so a caller writing the RFC's own number gets the RFC's own
+    /// guarantee without having to subtract anything themselves.
     pub const aes_gcm_record_limit: u64 = 23_726_566;
 
+    /// Whether `records_sealed` records under the current generation is enough
+    /// to act on.
+    ///
+    /// Fires one record *before* the configured cap, because acting on it is
+    /// not free: `requestKeyUpdate` must seal the `KeyUpdate` itself under the
+    /// keys it is retiring (RFC 8446 §4.6.3 — the message announcing a
+    /// transition is protected by the generation it ends), so that message is
+    /// one more record against the old keys. Firing at the cap itself would
+    /// let the retiring generation protect `records + 1` records, which is
+    /// precisely the bound the caller asked not to cross.
+    ///
+    /// Saturating rather than wrapping: a nonsensical `records = 0` is due
+    /// immediately instead of underflowing into a limit of `2^64 - 1`, which
+    /// would silently disable the policy it was meant to enforce.
     pub fn reached(self: UsageLimits, records_sealed: u64) bool {
         const limit = self.records orelse return false;
-        return records_sealed >= limit;
+        return records_sealed >= limit -| 1;
     }
 };
 
@@ -136,8 +158,42 @@ test "usage limits stay inert until configured" {
     try std.testing.expect(!off.reached(0));
     try std.testing.expect(!off.reached(std.math.maxInt(u64)));
 
+    // A cap of 4 becomes due at 3 sealed records: the KeyUpdate is then the
+    // 4th and final record the retiring generation protects.
     const on: UsageLimits = .{ .records = 4 };
-    try std.testing.expect(!on.reached(3));
+    try std.testing.expect(!on.reached(2));
+    try std.testing.expect(on.reached(3));
     try std.testing.expect(on.reached(4));
-    try std.testing.expect(on.reached(5));
+}
+
+test "the AES-GCM cap leaves room for the KeyUpdate that retires the generation" {
+    // The property the constant exists for: a caller who configures RFC 8446
+    // §5.5's own number must never let the retiring keys protect more records
+    // than it names. Since `requestKeyUpdate` seals the KeyUpdate under those
+    // same old keys, becoming due at the cap itself would make that message
+    // the (limit + 1)-th record -- one past the bound.
+    const policy: UsageLimits = .{ .records = UsageLimits.aes_gcm_record_limit };
+    try std.testing.expect(!policy.reached(UsageLimits.aes_gcm_record_limit - 2));
+    try std.testing.expect(policy.reached(UsageLimits.aes_gcm_record_limit - 1));
+    // Acting the moment it is due seals the KeyUpdate as record number
+    // `aes_gcm_record_limit`, the last one permitted -- not the first one
+    // beyond it.
+    try std.testing.expectEqual(
+        UsageLimits.aes_gcm_record_limit,
+        (UsageLimits.aes_gcm_record_limit - 1) + 1,
+    );
+}
+
+test "a degenerate zero cap is due immediately rather than underflowing" {
+    // `0 -| 1` saturates to 0, so this reports due at once. Wrapping would
+    // have produced a limit of 2^64 - 1 and silently disabled the policy.
+    const zero: UsageLimits = .{ .records = 0 };
+    try std.testing.expect(zero.reached(0));
+
+    const one: UsageLimits = .{ .records = 1 };
+    try std.testing.expect(one.reached(0));
+
+    const huge: UsageLimits = .{ .records = std.math.maxInt(u64) };
+    try std.testing.expect(!huge.reached(std.math.maxInt(u64) - 2));
+    try std.testing.expect(huge.reached(std.math.maxInt(u64) - 1));
 }
--- a/src/tls/key_update.zig
+++ b/src/tls/key_update.zig
@@ -1,0 +1,143 @@
+//! TLS 1.3 post-handshake `KeyUpdate` (RFC 8446 §4.6.3).
+//!
+//! This module owns only the message's wire form and the local policy for
+//! *when* an endpoint should ask for a proactive update. It has no record,
+//! socket, QUIC, or key-schedule types: the secret advancement itself is
+//! `key_schedule.KeySchedule.nextTrafficSecret`, and applying that to live
+//! traffic keys is `record_epoch_bridge`.
+//!
+//! `KeyUpdate` is record-mode-only. RFC 9001 §6 removes it from TLS-over-QUIC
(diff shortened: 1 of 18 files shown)


src/tls/tls13_backend.zig
```diff
@@ -146,33 +176,60 @@ const PostHandshakeInput = struct {
                 const body_len: usize = @intCast(std.mem.readInt(u24, self.header[1..4], .big));
                 const frame_len = handshake_header_len + body_len;
                 if (frame_len > max_new_session_ticket_message_len) return error.HandshakeBufferOverflow;
```
Jaruso: [P2] Classify `KeyUpdate` length before the generic ticket-size cap

This guard runs before the KeyUpdate-specific exact-length check below. A peer can send just a 4-byte handshake header with `type = key_update` and a declared body large enough that `frame_len > max_new_session_ticket_message_len`; this returns `HandshakeBufferOverflow` instead of `MalformedHandshake`. In record mode, `mappedFatalAlert` does not map `HandshakeBufferOverflow`, so this path can terminate locally without the RFC `decode_error` alert. It also contradicts the PR's stated invariant that any non-1-byte KeyUpdate body is rejected as a framing error *before* a buffer is chosen.

Please perform the type-specific length validation first, then apply the generic post-handshake cap, e.g.:

```zig
const body_len: usize = @intCast(std.mem.readInt(u24, self.header[1..4], .big));
const frame_len = handshake_header_len + body_len;
const is_key_update = self.header[0] == @intFromEnum(MessageType.key_update);

if (is_key_update and frame_len != key_update.message_len) {
    return error.MalformedHandshake;
}
if (frame_len > max_new_session_ticket_message_len) {
    return error.HandshakeBufferOverflow;
}
```

Please also add a regression using only a KeyUpdate header whose declared body makes `frame_len` exceed `max_new_session_ticket_message_len`, and assert `MalformedHandshake` locally plus `decode_error` at the peer. Running it with and without the post-handshake allocator would preserve the invariant this PR is explicitly trying to establish.

src/tls/key_update.zig
```diff
@@ -0,0 +1,143 @@
...
+    /// is the default: an endpoint is always free to never send a `KeyUpdate`,
+    /// and turning it on by default would change the wire behavior of every
+    /// existing connection.
+    records: ?u64 = null,
+
+    /// RFC 8446 §5.5's AES-GCM record limit (2^24.5, rounded down to a whole
+    /// number of records). Callers opting into proactive updates should pick a
+    /// value at or below this; it is offered as a named constant rather than a
+    /// default so the choice stays explicit at the call site.
+    pub const aes_gcm_record_limit: u64 = 23_726_566;
```
Jaruso, line 102: [P2] Reserve the retiring-key `KeyUpdate` record before this AES-GCM limit

The hook currently becomes due only after `records_sealed >= records`. If a caller follows this comment and configures `.records = UsageLimits.aes_gcm_record_limit`, the stream can seal 23,726,566 records under the current keys before `keyUpdateDue()` fires. `requestKeyUpdate()` then has to seal the KeyUpdate handshake record under those *same old keys* before advancing them, so the retiring generation protects one more record than the configured RFC-derived cap.

Please make the actionable threshold reserve one old-key record for the KeyUpdate itself. Two reasonable shapes are:

```zig
pub const aes_gcm_record_limit: u64 = 23_726_566;
pub const aes_gcm_update_threshold: u64 = aes_gcm_record_limit - 1;
```

and document that callers configure the threshold, or change the predicate semantics to account for the required next control record (using saturating/overflow-safe arithmetic).

Add a boundary test that the AES-GCM policy reports due at `aes_gcm_record_limit - 1`; that guarantees the KeyUpdate itself is at most the final permitted record under the retiring generation rather than the first record beyond it.

Jaruso (commented):
The core #357 implementation is strong: I traced the directional transition ordering through the backend event batch into the record bridge, the reciprocal/no-loop behavior, fragmentation/reassembly, sequence reset, retired-key zeroization, QUIC rejection, and the OpenSSL client/server rows. Those acceptance criteria are covered, and CI is green.

I found two boundary issues that I consider must-fix before merge:

1. A `KeyUpdate` declaring a body large enough to exceed the generic post-handshake ticket cap returns `HandshakeBufferOverflow` before the KeyUpdate-specific length validation runs. That bypasses the intended `MalformedHandshake`/`decode_error` path and contradicts this PR's guarantee that *any* non-1-byte KeyUpdate body is rejected as a framing error before buffer selection.
2. The documented AES-GCM usage-limit constant is usable directly as the `records` trigger, but `keyUpdateDue()` only becomes true after that many records have already been sealed. Because the KeyUpdate record itself must still be sealed under the retiring keys, acting on the hook at that point consumes one additional old-key record. The trigger needs to reserve that control record.

I left detailed inline comments with concrete fixes and regression-test boundaries for both. GitHub does not permit the authenticated author to submit a formal REQUEST_CHANGES review on their own PR, so this is posted as COMMENT, but my review disposition is **changes required**.

src/tls/tls13_backend.zig
Jaruso: Fixed in c7c380cf — you were right on both counts, including the alert consequence.

Confirmed before changing anything: `mappedFatalAlert` ([encrypted_stream.zig:1265](https://github.com/Bare-Systems/Tardigrade/blob/c7c380cf/src/tls/encrypted_stream.zig#L1265)) has no `HandshakeBufferOverflow` arm, so it falls to `else => null` and that path terminated locally with nothing on the wire.

Applied your ordering — type-specific framing validation first, then the generic cap. Every non-1-byte body is now a framing error at every declared length.

Regression added: `#357 a KeyUpdate header declaring a body past the ticket cap is still a framing error`. It sends **only** the 4-byte header with `body_len = 0xFFFFFF` (asserting up front that this exceeds `max_new_session_ticket_message_len`, so the test can't silently stop covering the case if that constant grows), and checks `MalformedHandshake` locally plus `decode_error` observed at the peer via `peerAlert()`, both with and without the post-handshake allocator.

Mutation-checked: restoring the old ordering fails that test.

src/tls/key_update.zig
Jaruso: Fixed in c7c380cf. Agreed on the defect — the retiring generation did protect `records + 1`.

I took your second option (change the predicate) rather than the paired-constant one, for a reason worth flagging in case you'd rather have the first: with a separate `aes_gcm_update_threshold`, `.records = aes_gcm_record_limit` stays spellable and stays wrong, so the footgun survives next to its own fix. Making `reached` reserve the control record means a caller who writes the RFC's number gets the RFC's guarantee, and there's only one constant to reach for. `records` is now documented as a cap on the whole generation *including* the `KeyUpdate` that retires it.

Uses saturating `-|` as you asked: a degenerate `records = 0` is due immediately instead of underflowing to `2^64 - 1`, which would have silently disabled the very policy it configured.

Tests added:
- `the AES-GCM cap leaves room for the KeyUpdate that retires the generation` — the boundary you asked for: not due at `limit - 2`, due at `limit - 1`.
- `a degenerate zero cap is due immediately rather than underflowing`, covering `0`, `1`, and `maxInt(u64)`.
- The stream-level test now pins the boundary against `applicationRecordsSealed()` rather than a loop counter, so a short write can't drift the count.

One thing I couldn't assert the way I first tried, worth recording: the KeyUpdate being the cap-th record isn't observable *after* `requestKeyUpdate` returns. It seals the message and applies the write-side advance in one event batch, so the sequence has already restarted by then. The `cap - 1` assertion taken beforehand is what pins it; I left a comment saying so rather than leaving a weaker assertion that looked stronger.

Mutation-checked: restoring `>= limit` fails three tests.

src/tls/tls13_backend.zig
Jaruso: Re-reviewed `c7c380cf`. Verified fixed: the KeyUpdate-specific exact-length classification now runs before the generic post-handshake size cap, and the new header-only regression exercises a declared `0xFFFFFF` body with and without the allocator, asserting both local `MalformedHandshake` and peer-observed `decode_error`. This closes the failure-class/alert hole without changing the valid KeyUpdate path.

src/tls/key_update.zig
Jaruso: Re-reviewed `c7c380cf`. Verified fixed: `UsageLimits.reached` now reserves the retiring-key control record with saturating subtraction, the AES-GCM boundary is pinned at `limit - 1`, and the stream-level test checks the actual write sequence before `requestKeyUpdate` resets it. The revised `records` semantics are documented clearly and remove the direct-use footgun I called out.

Jaruso (commented):
Re-review of `c7c380cf` complete. Both prior P2 findings are fixed and I resolved their threads.

I rechecked the fixes against the original #357 acceptance criteria and the surrounding KeyUpdate paths: directional secret advancement and event ordering are unchanged; reciprocal/repeated/simultaneous behavior still terminates correctly; sequence reset and retired-key zeroization remain intact; malformed/pre-handshake/QUIC rejection stays deterministic; and the OpenSSL interop coverage is still present.

The framing fix now classifies every non-1-byte KeyUpdate before the generic post-handshake cap, including a header-only `0xFFFFFF` regression that verifies peer-observed `decode_error`. The usage-limit fix now treats `records` as a generation cap including the retiring KeyUpdate record, fires at `cap - 1` with saturating arithmetic, and pins that boundary both at the policy and stream levels.

No new code findings on this pass. The new-head CI run is still in progress at review time, but the unit-test jobs and TLS interop/conformance matrix are already green. **Code review disposition: ready to merge once the remaining CI jobs finish successfully.**

GitHub does not allow the authenticated author to approve their own PR, so this is submitted as COMMENT rather than APPROVE.

PR #592, opened 2026-08-09
perf(proxy): complete streaming request-upload and transport scope (#139)

### LINKED ISSUE #139: perf: implement streaming reverse proxy mode with bounded buffering

Jaruso:
## Summary

Implements the remaining gaps in #139. PR 1 of the issue's recommended sequence (route-level streaming policy, typed fallback reasons, fallback metrics) already landed on `main`, so this covers **PR 2 (request-streaming completion)** and **PR 3 (gateway-level conformance tests)** plus the documentation contract from item 5.

**This does not close #139.** The issue's acceptance criteria still require "Memory/backpressure behavior satisfies the enforcement and metrics requirements in #140", and #140 is open. The bounds relied on here are structural (one fixed relay buffer per direction for HTTP/1, the per-stream receive window for HTTP/2), which `PROXY_STREAMING.md` states explicitly rather than claiming they substitute for #140's enforcement. Closing #139 should wait for that criterion, or for a deliberate revision of the criteria.

### Chunked client uploads now stream

New `src/http/chunked_upload.zig` decodes downstream chunked framing incrementally — no heap allocation, fixed scratch space, borrows the bytes already read with the request head, and pulls from the connection only as the caller drains. The relay re-frames rather than forwarding raw octets:

- **HTTP/1.1 upstream** — `Transfer-Encoding: chunked`, re-chunked from the decoded payload.
- **HTTP/2 upstream** — flow-controlled DATA frames terminated by `END_STREAM`, with no synthesized `Content-Length`.

Chunk extensions and trailers are consumed and dropped. The decoded payload is budgeted against `TARDIGRADE_MAX_BODY_SIZE` *during* the relay: oversized → `413`, malformed → `400`, neither charged to origin health. Because the fault is only detectable once the offending bytes arrive, the origin has received a partial request by then — an HTTP/1.1 connection is closed rather than pooled, while HTTP/2 resets only the affected stream and keeps the shared connection.

### Unix-socket and mTLS upstreams now stream

The streaming relay owns AF_UNIX connect and keep-alive pooling under a `unix:<path>` key, via a new `compat.connectBoundedUnix` so the connect honours `TARDIGRADE_UPSTREAM_CONNECT_TIMEOUT_MS` like every other transport. A Unix-socket upstream always uses HTTP/1.1 — no origin for the h2 pool to key or ALPN-negotiate. The mTLS gate was conservatism: that path already passed the client certificate through.

### Bug found while building the mTLS test

The streaming relay constructed `UpstreamTlsOptions` **without** `alpn_policy`, silently taking the `require_http1` default. An HTTPS origin was therefore only ever offered `http/1.1`, so `TARDIGRADE_UPSTREAM_PROTOCOL=h2`/`auto` had no effect once streaming was enabled — every TLS exchange fell to the HTTP/1.1 relay on a fresh, unpooled connection. Now derived from the configured protocol like the buffered path. Cleartext `h2c` was unaffected, which is why the h2c test alone would not have caught it.

### Fallback reasons

The reason enum now lives in `http.metrics` and `recordProxyStreamingFallback` switches over it exhaustively instead of matching strings — a new reason cannot compile until it has a counter (verified by confirming an added case fails to build at the recorder). Three reasons that can no longer occur are removed: `unix_socket_target`, `upstream_mtls_target`, `chunked_request_upload`. `early_data_retry_semantics` was being emitted with no counter behind it; that gap is fixed.

**Breaking for dashboards:** anything referencing the three removed label series needs updating. Noted under `### Removed` in the changelog.

### Docs

New `docs/PROXY_STREAMING.md`: policy and aliases, the supported matrix for responses / uploads / transports, every fallback reason, and the retry and body-dependent-middleware contract (item 5). `OBSERVABILITY.md` and `UPSTREAM_POOLING.md` updated to match.

## Test plan

- [x] `zig build test` — full unit suite green (+9 for the chunked reader: split reads, head-buffered prefix, extensions, trailers, connection left positioned for the next request, malformed framing, truncation, budget enforcement, unbounded trailers)
- [x] `zig build test-integration` — 84 pass / 0 fail, 9 new gateway-level tests
- [x] `zig build test-failure` — 9/9
- [x] `zig fmt --check` clean

New integration coverage (fixture extended with an AF_UNIX listener, a prior-knowledge h2c origin with HPACK request capture, close-delimited responses, 204/304 reason phrases, and chunked request-body capture):

- [x] 512 KiB chunked upload with extensions and a trailer to an **HTTP/1.1** upstream — origin sees `Transfer-Encoding: chunked`, no `Content-Length`, exact decoded payload
- [x] 384 KiB chunked upload to an **h2c** upstream — origin sees the exact payload as DATA, with neither `content-length` nor `transfer-encoding`, delimited by `END_STREAM`
- [x] streaming response from an **mTLS** origin (`openssl s_server -Verify 1`), sized past the buffered cap so a buffered relay would fail it; runs only in the general/OpenSSL-adapter profile and is skipped when `openssl` is absent
- [x] malformed chunk framing → `400`; payload past the body maximum → `413`
- [x] close-delimited upstream response relayed and re-chunked downstream
- [x] client abort during a 4 MiB response, and mid-upload → `tardigrade_proxy_client_aborts_total`
- [x] Unix-socket upstream streams 256 KiB past the buffered cap with no fallback event
- [x] retries configured → `retries_configured` fallback and a buffered request

Already covered on `main` and still passing: large fixed-length and chunked responses, HEAD/204/304, upstream truncation after the head is written, large `Content-Length` upload, buffered fallback, route override vs global policy.

Refs #139


src/gateway_proxy.zig
```diff
@@ -1978,7 +2082,10 @@ pub fn executeStreamingHttpProxyRequest(
         }
         if (!reused) {
             const connect_start = http.event_loop.monotonicMs();
-            const new_fd = compat.connectBoundedTcp(host, port, connect_timeout_ms) catch |err| {
+            const new_fd = (if (unix_socket_path) |socket_path|
+                compat.connectBlockingUnix(socket_path)
```
Jaruso: [P2] The newly enabled Unix streaming path drops the configured connect deadline here. TCP calls `connectBoundedTcp(host, port, connect_timeout_ms)`, but this branch calls the explicitly blocking `connectBlockingUnix`, so a blocked AF_UNIX `connect()` can hold the worker outside `TARDIGRADE_UPSTREAM_CONNECT_TIMEOUT_MS`. That contradicts this executor's stated per-phase timeout contract and makes Unix streaming less bounded than the other newly supported transports.

Please add a bounded Unix helper using the same `connectFdBounded` machinery, e.g. conceptually:
```zig
pub fn connectBoundedUnix(path: []const u8, timeout_ms: u32) !std.posix.fd_t {
    var addr = std.mem.zeroes(std.c.sockaddr.un);
    addr.family = std.posix.AF.UNIX;
    if (path.len >= addr.path.len) return error.NameTooLong;
    @memcpy(addr.path[0..path.len], path);
    addr.path[path.len] = 0;

    const len: std.posix.socklen_t = @intCast(@offsetOf(std.c.sockaddr.un, "path") + path.len + 1);
    const sock = std.c.socket(std.posix.AF.UNIX, std.posix.SOCK.STREAM, 0);
    if (sock < 0) return error.SocketFailed;
    errdefer _ = std.c.close(sock);
    try connectFdBounded(sock, @ptrCast(&addr), len, timeout_ms);
    return sock;
}
```
then use `compat.connectBoundedUnix(socket_path, connect_timeout_ms)` here. Please add a timeout/saturated-listener test for this branch as well.

src/http/metrics.zig
```diff
@@ -2266,28 +2254,32 @@ test "#368 Slice 3: early-data replay counters appear in Prometheus output with
 
 test "Metrics records proxy streaming fallback reasons" {
     const allocator = std.testing.allocator;
+    // Must stay in sync with gateway_proxy_runtime.StreamingFallbackReason: a
+    // reason with no counter here is emitted but silently unobservable.
+    const reasons = [_][]const u8{
```
Jaruso: [P2] This does **not** actually guarantee the synchronization claimed by the comment/PR summary. `reasons` is another manually maintained string list, independent of `gateway_proxy_runtime.StreamingFallbackReason`. A future enum case can be added, `metricLabel()` updated, and production can call `recordProxyStreamingFallback()` with that new label while this test still passes and the string-chain recorder silently drops it.

Please make the reason type the single source of truth and make the metric mapping exhaustive. One clean layering is to put the enum in a low-level shared module (or `http.metrics` if that dependency direction works):
```zig
pub const ProxyStreamingFallbackReason = enum {
    policy_disabled,
    retries_configured,
    missing_content_length,
    body_too_large,
    body_dependent_middleware,
    unsupported_route_type,
    early_data_retry_semantics,

    pub fn label(self: @This()) []const u8 {
        return @tagName(self);
    }
};

pub fn recordProxyStreamingFallback(self: *Metrics, reason: ProxyStreamingFallbackReason) void {
    switch (reason) {
        .policy_disabled => self.proxy_streaming_fallback_policy_disabled += 1,
        .retries_configured => self.proxy_streaming_fallback_retries_configured += 1,
        // ...every case, exhaustively...
    }
}
```
Then the test can iterate `std.enums.values(ProxyStreamingFallbackReason)` and a new enum case forces the recorder switch to be updated at compile time.

docs/PROXY_STREAMING.md
```diff
@@ -0,0 +1,152 @@
...
+| `Transfer-Encoding: chunked` | Decoded incrementally and re-framed as `Transfer-Encoding: chunked`. | Decoded incrementally and sent as DATA frames terminated by `END_STREAM`; no `Content-Length` is synthesized. |
+
+Chunked uploads are re-framed rather than forwarded byte-for-byte, so chunk
+extensions and trailers sent by the client are consumed and dropped, not passed
+to the origin. The decoded payload is counted against the configured
+request-body maximum while it is relayed: an upload that outgrows the maximum is
+rejected with `413` mid-relay, and malformed chunk framing is rejected with
+`400`. Neither is attributed to the origin's health. Because the failure is only
+detectable once the offending bytes arrive, the origin has by then received a
+partial request; that upstream connection is torn down rather than pooled.
```
Jaruso: [P2] This is not true for the HTTP/2 path documented immediately above. On a chunk framing/size error, `streamViaH2Pool` calls `conn.finishStreaming(stream)`, which sends `RST_STREAM(CANCEL)` for the unfinished stream; it only evicts the shared H2 connection when `!conn.healthy()`. Closing the whole H2 connection would unnecessarily kill unrelated multiplexed requests.

Please make the protocol distinction explicit, e.g.:

> Because the failure is detected after partial forwarding, an HTTP/1.1 upstream connection is closed and not pooled. For HTTP/2, only the affected stream is reset with CANCEL; the shared connection remains reusable unless it is independently unhealthy.

That also makes the docs accurately describe the implementation's desirable multiplexing behavior.

tests/integration.zig
```diff
@@ -12117,6 +12167,427 @@ test "proxy full streaming upload works for server block route override" {
     try std.testing.expectEqualStrings(payload, captured);
 }
 
+test "proxy full streaming mode relays a chunked upload without buffering it" {
```
Jaruso, line 12426: [P2] This validates only the new HTTP/1.1 re-chunking branch. The same PR adds `relayStreamingUploadToHttp2` and removes both the chunked-upload and upstream-mTLS fallback barriers, but there is no gateway-level test executing a chunked downstream upload against an H2/H2C upstream or an mTLS streaming upstream. A bug in either newly claimed combination can therefore ship while all of these tests remain green.

Please add conformance coverage before removing those fallback reasons. For the H2 case, the test should send a multi-chunk body (including extension/trailer), force an H2/H2C upstream, and assert the origin receives the exact decoded payload via DATA with end-of-stream semantics and no synthesized `Content-Length`/`Transfer-Encoding`, plus the streaming metric/no-fallback result. For mTLS, stand up the existing TLS test origin requiring a client certificate, send a body/response larger than the buffered cap with streaming enabled, and assert the exchange succeeds and no `upstream_mtls_target` fallback is possible. If the integration fixture cannot expose H2 request capture yet, add the fixture support rather than limiting this to the H1 path—the new H2 upload helper itself needs an executable regression test.

Jaruso (commented):
**Blocking findings** — submitted as a COMMENT because GitHub does not allow an author to request changes on their own PR.

This is close, and CI is green, but I found three implementation/test blockers plus one documentation contract mismatch.

The biggest runtime issue is the new Unix-socket streaming branch bypassing `connect_timeout_ms`: TCP uses `connectBoundedTcp`, while Unix uses the explicitly blocking `connectBlockingUnix`. That makes a newly supported streaming target violate the per-phase timeout contract the executor documents. Please add/use a bounded Unix connect helper rather than allowing the worker to block in `connect()` outside the configured deadline.

The fallback-metric regression test also does not provide the protection described by its comment/PR summary. `reasons` is a second hand-maintained string list; adding a new `StreamingFallbackReason` can still compile, emit a label, and be silently ignored by `Metrics.recordProxyStreamingFallback`. Please make the reason type the shared source of truth and make metric dispatch exhaustive.

The new gateway tests prove the HTTP/1 chunked relay and Unix response path, but they do not exercise the newly added `relayStreamingUploadToHttp2` path or the newly unblocked upstream-mTLS streaming combination. Because this PR removes those fallback barriers and #139 requires the supported combinations to be tested/observable, I want executable coverage for those branches before we rely on them.

There is also a docs/runtime mismatch: malformed/oversized partial uploads close the H1 connection, but on H2 `finishStreaming` resets the affected stream and preserves a healthy shared connection. The new streaming contract should state that distinction accurately.

One issue-level point: the PR currently says `Closes #139`, but #139's acceptance criteria still explicitly require the memory/backpressure enforcement and metrics owned by #140. #140 is still open and its current audit says those high/low-watermark and aggregate-cap requirements remain incomplete. I am **not** asking this PR to absorb #140; instead, either remove the closing keyword and leave #139 open until that criterion is satisfied, or update #139's acceptance criteria/relationship first if the maintainer intent is that #140 is no longer blocking closure.

Once these are addressed I can re-review the resulting commit against the same #139 criteria.

Jaruso (commented):
Second-pass review: the PR head is still `e837109cb2cb6bb088703541cbb5a932296f3aa0`, the same commit reviewed previously. I re-checked the current issue contract and CI state; CI remains green, but there are no code changes addressing the prior review and all four inline threads remain unresolved.

The prior blocking findings therefore still stand:

1. Unix-socket streaming uses `connectBlockingUnix()` and does not honor the configured connect deadline.
2. The fallback-metric regression test is still independently hard-coded and does not make `StreamingFallbackReason` -> metrics accounting exhaustive at compile time.
3. `PROXY_STREAMING.md` still incorrectly says a malformed/oversized chunked upload tears down the upstream connection for all protocols; H2 correctly resets only the affected stream unless the shared connection is unhealthy.
4. Gateway-level conformance coverage is still missing for the new chunked-upload-to-H2/H2C path and the newly enabled mTLS streaming path.

I also re-checked #139. Its acceptance criteria still explicitly require: `Memory/backpressure behavior satisfies the enforcement and metrics requirements in #140.` #140 remains open. That does not mean this PR should absorb #140, but as written this PR should not `Closes #139` unless #139's acceptance criteria are intentionally revised to make #140 an independent follow-up.

No additional defects found beyond the prior review. I would re-review once the head commit changes.

Jaruso (commented):
Final re-review on `7b5d291e`: clean from the PR-scope perspective. I re-checked the four prior blockers and the changes made afterward.

- Unix-socket streaming now uses `connectBoundedUnix(..., connect_timeout_ms)` and shares the bounded-connect machinery with TCP.
- `ProxyStreamingFallbackReason` is now the single typed source of truth in metrics; the recorder switch is exhaustive and the test iterates the enum values.
- `PROXY_STREAMING.md` now correctly distinguishes H1 connection close from H2 stream reset/reuse.
- Gateway conformance coverage now exercises chunked downstream upload to h2c and mTLS streaming, with the mTLS test correctly gated to the general/OpenSSL-adapter profile.
- The follow-up H2 fix correctly prevents a connection EOF that arrives after `END_STREAM` from retroactively invalidating a completed response while still leaving the shared connection unhealthy for future reuse.
- Final CI run #1277 is green across the complete workflow matrix.

All previous review threads are resolved. I found no remaining PR-level blocker; **PR #592 is ready to merge.**

Issue-level note remains intentional: #139 itself should stay open after this merge because its final acceptance criterion still depends on #140, which is open. The PR already says `Refs #139` rather than closing it, so the current PR metadata is correct.

PR #594, opened 2026-08-09
tls: extend OpenSSL TCP interop coverage

Jaruso:
## Summary

- add an explicit OpenSSL TCP `h2` entrypoint row that negotiates ALPN `h2` and verifies the full HTTP/2 client preface plus SETTINGS frame arrives as decrypted application data
- expand `run-tls-interop.sh --list` to show OpenSSL client/server, negative, KeyUpdate, selection, QUIC, shutdown, and HTTP entrypoint rows
- document issue #358 coverage, OpenSSL 3.x prerequisites, HTTP/1.1/HTTP/2 entrypoint behavior, and clean-close/truncation expectations

Closes #358

## Validation

- `bash -n scripts/interop/run-tls-interop.sh`
- `./scripts/interop/run-tls-interop.sh --list --profile ci`
- `zig build build-tls-interop build-h3-interop --summary all --error-style verbose`
- `zig build test-tls-interop-matrix --summary all --error-style verbose`
- `shellcheck scripts/interop/run-tls-interop.sh`
- `TLS_INTEROP_WORKDIR="$PWD/.tmp-tls-interop-594-latest" ./scripts/interop/run-tls-interop.sh --profile ci` (`pass=84 fail=0 skip=0`)

### DIFF
--- a/scripts/interop/run-tls-interop.sh
+++ b/scripts/interop/run-tls-interop.sh
@@ -171,6 +171,50 @@ check_peer_spellings() {
   fi
 }
 
+list_matrix_rows() {
+  say "profile: $profile"
+  say "engine capabilities: ${#suites[@]} suites x ${#groups[@]} groups x ${#signatures[@]} signatures"
+  for suite in "${suites[@]}"; do
+    for group in "${groups[@]}"; do
+      for sig in "${signatures[@]}"; do
+        tuple="$suite/$group/$sig"
+        say "positive  record/server/openssl/$tuple"
+        say "positive  record/client/openssl/$tuple"
+        if [ "$have_gnutls" -eq 0 ]; then
+          say "skip      record/server/gnutls/$tuple"
+          say "skip      record/client/gnutls/$tuple"
+        elif gnutls_row_in_profile "$suite" "$group" "$sig"; then
+          say "positive  record/server/gnutls/$tuple"
+          say "positive  record/client/gnutls/$tuple"
+        fi
+        say "positive  quic/loopback/$tuple"
+      done
+    done
+  done
+  say "positive  record/server/openssl/http2_entrypoint"
+  say "positive  record/server/openssl/hrr"
+  say "positive  record/client/openssl/hrr"
+  say "positive  record/server/openssl/key-update/reciprocal"
+  say "positive  record/client/openssl/key-update/reciprocal"
+  say "positive  record/client/openssl/key-update/one-way"
+  say "negative  record/negative/alpn_no_overlap"
+  say "negative  record/negative/tls12_downgrade"
+  say "negative  record/negative/cipher_no_overlap"
+  say "negative  record/negative/group_no_overlap"
+  say "negative  record/negative/signature_no_overlap"
+  say "negative  record/negative/sni_absent"
+  say "negative  record/negative/malformed_ordering"
+  say "negative  record/negative/ccs_before_clienthello"
+  say "negative  record/negative/wrong_pinned_certificate"
+  say "selection record/selection/sni_ed25519"
+  say "selection record/selection/sni_ecdsa_p256"
+  say "selection record/selection/sni_rsa"
+  say "selection record/selection/unknown_sni_uses_default"
+  say "selection record/selection/sigalgs_ed25519"
+  say "selection record/selection/sigalgs_rsa_pss"
+  say "selection record/selection/no_applicable_credential"
+}
+
 openssl_suite() {
   case "$1" in
     aes128-gcm-sha256) echo TLS_AES_128_GCM_SHA256 ;;
@@ -263,15 +307,19 @@ peer_request_stdin() { # once|keep_writing
       printf 'GET / HTTP/1.0\r\nX-Key-Update-Probe: 1\r\n\r\n'
       sleep 2
       ;;
+    h2_preface)
+      printf 'PRI * HTTP/2.0\r\n\r\nSM\r\n\r\n'
+      printf '\x00\x00\x00\x04\x00\x00\x00\x00\x00'
+      ;;
     *) printf 'GET / HTTP/1.0\r\n\r\n' ;;
   esac
 }
 
 # native TLS server <- external client
-run_server_row() { # name peer suite group signature [extra native args...]
-  local name="$1" peer="$2" suite="$3" group="$4" sig="$5"
-  shift 5
-  local p log cert peer_groups peer_stdin
+run_server_alpn_row() { # name peer suite group signature alpn peer_stdin [extra native args...]
+  local name="$1" peer="$2" suite="$3" group="$4" sig="$5" alpn="$6" peer_stdin="$7"
+  shift 7
+  local p log cert peer_groups
   next_port; p="$port"
   log="$logs/${name//\//_}"
   cert="$(cert_for_signature "$sig")"
@@ -279,12 +327,11 @@ run_server_row() { # name peer suite group signature [extra native args...]
   # native server does not accept; ordinary rows offer the row's own group.
   peer_groups="$(openssl_group "$group")"
   case " $* " in *" --expect-hrr "*) peer_groups="P-384:$(openssl_group "$group")" ;; esac
-  peer_stdin=once
   case " $* " in *" --key-update "*) peer_stdin=keep_writing ;; esac
 
   "$tls_tool" server --port "$p" --cert "$cert-cert.pem" --key "$cert-key.pem" \
     --cipher-suite "$suite" --group "$group" --signature "$sig" \
-    --alpn http/1.1 --transcript "$transcripts/${name//\//_}.txt" \
+    --alpn "$alpn" --transcript "$transcripts/${name//\//_}.txt" \
     --timeout-ms 20000 "$@" > "$log.native" 2>&1 &
   local native_pid=$!
   if ! wait_for_native_listener "$log.native"; then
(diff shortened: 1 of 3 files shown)


scripts/interop/run-tls-interop.sh
```diff
@@ -171,6 +171,50 @@ check_peer_spellings() {
...
+  say "positive  record/client/openssl/key-update/one-way"
+  say "negative  record/negative/alpn_no_overlap"
+  say "negative  record/negative/tls12_downgrade"
+  say "negative  record/negative/cipher_no_overlap"
+  say "negative  record/negative/group_no_overlap"
+  say "negative  record/negative/signature_no_overlap"
+  say "negative  record/negative/sni_absent"
+  say "negative  record/negative/malformed_ordering"
+  say "negative  record/negative/ccs_before_clienthello"
+  say "negative  record/negative/wrong_pinned_certificate"
```
Jaruso, line 210: **Blocking — #358's truncation acceptance criterion is still uncovered.** The issue explicitly requires expected truncation cases to fail predictably, and its scope also calls out clean close. This matrix still has no external-OpenSSL clean-shutdown or abrupt-EOF/truncation row.

The existing positive rows do not implicitly prove this: in `tls_interop_tool.Runner.runOne`, once `delivered` becomes true, a later `stream.drive()` error is deliberately ignored (`if (delivered) break`), so an EOF without `close_notify` after the application payload cannot fail the row even though `PureZigRecordStream` correctly surfaces that condition as `TruncatedStream`.

Please add explicit peer-close expectations and rows before closing #358. A concrete shape would be:

```zig
const PeerCloseExpectation = enum { none, clean, truncated };
// Args: --expect-peer-close clean|truncated

// Once app data is delivered, keep driving when a close expectation exists.
// clean: require peer close_notify / clean EOF and no terminal error.
// truncated: require error.TruncatedStream after an external peer drops TCP
//            without sending close_notify.
```

Then add/list two OpenSSL-backed rows, e.g. `record/server/openssl/close_notify` using the peer's normal TLS shutdown and `record/server/openssl/truncation` using an OpenSSL peer kept alive after its request and then terminated without TLS shutdown. The truncation row should assert the typed `TruncatedStream` result, not merely "the peer disconnected." This also gives the issue's clean-close scope an explicit interop fixture instead of relying on lower-level unit coverage.

scripts/interop/run-tls-interop.sh
```diff
@@ -525,6 +570,16 @@ for suite in "${suites[@]}"; do
...
 
+say ""
+say "== record transport: HTTP protocol entrypoints =="
+# #358: the positive tuple sweep proves HTTP/1.1 application bytes in both
+# roles. This row pins the HTTP/2 ALPN entrypoint against OpenSSL too: the
+# peer sends the RFC 9113 client connection preface and an empty SETTINGS
+# frame after negotiation, and the native server verifies those bytes crossed
+# under the negotiated TLS 1.3 application keys.
+run_server_alpn_row "record/server/openssl/http2_entrypoint" openssl aes128-gcm-sha256 x25519 ed25519 \
+  h2 h2_preface --expect-contains "PRI * HTTP/2.0" --send "tardigrade-h2-interop"
```
Jaruso: **The new h2 row can pass on a partial client preface.** `--expect-contains "PRI * HTTP/2.0"` only requires the first 14 bytes. `Runner.matched()` then marks the application exchange delivered, and once the reply ciphertext drains `runOne` can exit without ever reading the remaining `\r\n\r\nSM\r\n\r\n` bytes or the 9-byte empty SETTINGS frame that `h2_preface` writes.

That is weaker than the row/comment's intended proof. Please require the complete 24-byte HTTP/2 connection preface and enough received bytes to include the SETTINGS frame. For example, add a small length expectation to `tls_interop_tool`:

```zig
// Args
expect_min_bytes: usize = 0, // --expect-min-bytes N

// evaluate()
if (outcome.received_len < args.expect_min_bytes) failed = true;
```

and drive the row as:

```bash
h2 h2_preface \
  --expect-contains $'PRI * HTTP/2.0\r\n\r\nSM\r\n\r\n' \
  --expect-min-bytes 33 \
  --send "tardigrade-h2-interop"
```

24 bytes for the connection preface + 9 bytes for the empty SETTINGS frame makes the assertion match what this row says it exercises.

Jaruso (commented):
Blocking review (submitted as COMMENT because this is the authenticated user's own PR).

I reviewed both changed files, the full `run-tls-interop.sh`/`tls_interop_tool` behavior on this head, and issue #358 against its stated scope and acceptance criteria. The existing #588 matrix work still looks sound for the supported algorithm sweep, both TCP roles, SNI/certificate selection, certificate rejection, ALPN mismatch, malformed/error alerts, KeyUpdate, secret-safe transcripts, and out-of-process OpenSSL isolation.

Two gaps remain before this should `Closes #358`:

1. #358 explicitly requires truncation cases to fail predictably and also calls out clean-close coverage. There is still no external OpenSSL shutdown/truncation row, and `tls_interop_tool` currently suppresses transport errors after the application exchange is marked delivered, so the existing positive rows cannot substitute for that assertion.
2. The new HTTP/2 row can succeed after receiving only the first 14 bytes (`PRI * HTTP/2.0`) of the 24-byte client connection preface; it does not require the full preface or the following 9-byte SETTINGS frame to have arrived before the native side declares the exchange complete.

I left detailed inline suggestions for both. The focused TLS interop/conformance CI job and lint are green on this head; the overall CI workflow still has the Ubuntu appliance-profile job in progress at the time of review.

scripts/interop/run-tls-interop.sh
```diff
@@ -315,6 +364,57 @@ run_server_row() { # name peer suite group signature [extra native args...]
...
+  "$openssl_bin" s_client -connect "127.0.0.1:$p" \
+    -servername tardigrade.test -tls1_3 \
+    -ciphersuites "$(openssl_suite "$suite")" -groups "$(openssl_group "$group")" \
+    -sigalgs "$(openssl_sigalg "$sig")" -alpn http/1.1 \
+    -CAfile "$cert-cert.pem" < "$fifo" > "$log.peer" 2>&1 &
+  peer_pid=$!
+
+  exec 3>"$fifo"
+  printf 'GET / HTTP/1.0\r\nConnection: keep-alive\r\n\r\n' >&3
+  sleep 1
```
Jaruso, line 405: **Blocking — make the application exchange a required precondition for the forced truncation.** Right now the one-second sleep is the only thing separating “OpenSSL connected” from `kill -KILL`, while the native invocation only asserts `TruncatedStream`. If this runner is slow enough that the peer is killed before the HTTP request reaches the native stream, EOF during/after the handshake can still produce `TruncatedStream` and the row will pass without testing the post-application truncation described in the docs.

The current CI run happened to receive the full request (`app_bytes=42`), but that is not part of the row's pass condition. At minimum, make those 42 bytes mandatory:

```bash
"$tls_tool" server ... \
  --expect fail --expect-error TruncatedStream \
  --expect-peer-close truncated \
  --expect-min-bytes 42 \
  ...
```

(`42` is the exact length of `GET / HTTP/1.0\r\nConnection: keep-alive\r\n\r\n`.) That converts an early kill into a red row instead of a false positive.

Better still, avoid using elapsed time as the synchronization primitive: keep the FIFO writer open, wait until there is positive evidence that the application exchange has progressed (for example a dedicated progress signal from `tls_interop_tool`, or a reliably unbuffered observation that the peer received `tardigrade-tls-interop`), and only then `SIGKILL` the OpenSSL process. The important invariant is: **handshake completed + application data was decrypted, then TCP disappears without `close_notify`, then native reports `TruncatedStream`.**

Jaruso (commented):
Follow-up review on b103ee4.

The previous H2 blocker is fixed: the row now requires the full 24-byte client preface plus the 9-byte SETTINGS frame, and CI confirms `app_bytes=33`. The shutdown work also correctly adds explicit `close_notify` and `TruncatedStream` expectations; the focused TLS interop job is green at `pass=84 fail=0 skip=0`, with the current truncation run reporting `app_bytes=42`.

There is one remaining blocking test-validity issue before I think this should close #358. The truncation row kills OpenSSL after an unconditional one-second sleep, but it never asserts that the 42-byte application request was actually decrypted before the kill. Because `--expect fail --expect-error TruncatedStream --expect-peer-close truncated` is sufficient by itself, a slow run that is killed during/just after the handshake can still report `TruncatedStream` and pass. That would make the negative shutdown case timing-dependent and capable of going green without exercising post-handshake application-data truncation, contrary to #358's requirement for predictable truncation coverage.

I left the concrete fix inline. Once the row makes successful application-data exchange a required precondition for the forced TCP abort, I do not see another acceptance-criteria gap in this PR. The PR description's local validation count is also now stale (`82` vs the current `84` rows), but that is documentation cleanup rather than a merge blocker.

The overall CI workflow is still in progress at review time, but the changed-area checks are green: TLS interop/conformance matrix, unit tests, lint/shellcheck, formatting, dependency audit, and the normal builds/tests that have completed so far.

Jaruso (commented):
Re-review complete on head `2637f08505351a6e3aca18a859cefe6c69db83ff`. I do not have any remaining code or acceptance-criteria findings.

The last blocker is fixed: the OpenSSL truncation row now requires `--expect-min-bytes 42` in addition to `TruncatedStream`, so killing the peer before the HTTP request reaches the native stream can no longer produce a false-positive pass. The current TLS interop CI run exercised that exact path and reported `record/server/openssl/truncation PASS error=TruncatedStream app_bytes=42`.

The earlier blocker is also still correctly addressed: the HTTP/2 entrypoint requires the complete 24-byte client preface plus the 9-byte SETTINGS frame (`app_bytes=33` in CI), and explicit OpenSSL clean-close/truncation rows cover the shutdown requirements.

Against #358, this now covers both TCP roles and supported tuples, SNI/certificate validation, ALPN `http/1.1` and `h2`, HTTP application data, malformed/error/alert cases, clean close, truncation, KeyUpdate, secret-safe reproducible transcripts, out-of-process OpenSSL isolation, local/CI profiles, and documented OpenSSL prerequisites.

Focused `TLS interop/conformance matrix` is green at `pass=84 fail=0 skip=0`; lint, format, unit tests, security, dependency review, and the completed build jobs are also green. The overall CI workflow is still in progress only because several broader appliance/packaging jobs have not finished yet; I see no current failures.

This is effectively LGTM from code review. GitHub does not allow me to approve the authenticated user's own PR, so this is submitted as COMMENT rather than APPROVE.

PR #595, opened 2026-08-09
perf(proxy): enforce HTTP/2 watermark and aggregate buffer backpressure (#140)

### LINKED ISSUE #140: perf: add watermark-based backpressure for proxy and connection buffers

Jaruso:
## Summary

Implements the HTTP/2 response-enforcement slice of #140 (the issue's "PR 2"), plus the aggregate policy from its PR 4. The shared accounting type, config, and metrics from PR 1 already existed; this makes them *enforce* rather than merely observe.

**This PR does not close #140.** The issue's PR-3 slice — request-direction accounting and HTTP/1 pause/resume observability, with its slow-origin, large-upload, and client-abort coverage — is still outstanding, and #140's acceptance criteria require it.

### HTTP/2 per-stream bounds

- **The advertised receive window is derived from configuration.** `SETTINGS_INITIAL_WINDOW_SIZE` was a fixed 1 MiB constant unrelated to `TARDIGRADE_PROXY_BUFFER_PER_STREAM_HIGH_WATERMARK_BYTES`. It is now that value, so protocol flow control stops a well-behaved origin exactly where the accounting model says the queue is full.
- **Credit is returned with hysteresis.** While a queue sits at or above the high watermark, consumed bytes accumulate uncredited; the total goes out as one `WINDOW_UPDATE` when the queue drains below the low watermark. Crediting each chunk let the origin refill immediately, so the band never paused anything. `tardigrade_buffer_read_pauses_total{side="upstream"}` and its resume counterpart are now real signals instead of always-zero series.

### Aggregate bounds

- **Per-origin and global hard limits are enforced**, not just validated — N concurrent slow streams could previously retain N windows.
- **Reservations track retained allocation, not logical length.** `shrinkRetainingCapacity` keeps the backing buffer, so a drained queue could return its bytes to the aggregates while still owning ~a window of memory; RSS could grow by `N × window` with the counters near zero. Queues now allocate precisely and hand storage back the moment they drain empty.
- Reservations are taken before memory is committed, and a refusal at one scope rolls the other back.
- Both limits still default to `0` (unlimited), so existing deployments are unchanged until they opt in.

### Failure semantics

Deterministic on both sides of the downstream commitment boundary:

- **Pre-commit** — including when the reader rejects DATA between decoding the origin's headers and the worker relaying them — the request fails `503 proxy_buffer_saturated` and is not retried.
- **Post-commit** the stream is reset upstream, the response truncated, reservations released, and the reason logged.

Neither counts against upstream health: local memory pressure must not trip a healthy origin's failure policy. A refused stream becomes discard-only, so DATA still in flight cannot re-reserve, re-count the limit event, or queue bytes behind an error the consumer has not seen. Unrelated streams, including others on the same connection, are untouched. Allocation failure while setting up origin accounting is a `503` rather than silently dropping the scope.

### Configuration and reload

Each connection pins the whole per-stream policy it advertised, so a peer is judged by the credit it was actually granted rather than a newer snapshot. The pool owns per-origin limits and applies a reload to existing origins; requests no longer write their own config snapshot into shared state. Aggregate limits reload immediately; the per-stream policy reaches connections opened afterwards, since SETTINGS are negotiated once per connection. A high watermark too large to advertise as an HTTP/2 window is rejected at startup and reload.

Docs updated in `PROXY_STREAMING.md`, `OBSERVABILITY.md`, and the changelog.

**Out of scope**, per the issue's sequencing: the HTTP/1 relay keeps its fixed buffers, gains no queue, and takes no aggregate reservation — its buffers do not grow with concurrency the way an h2 stream queue does.

## Test plan

- [x] `zig build test` — full unit suite passes
- [x] `zig build test-integration` — passes (65 skips are the pre-existing socket-sandbox skips)
- [x] `zig build -Doptimize=ReleaseFast` — clean
- [x] **Concurrent-stream aggregate case** with limits that pass `validate()` (`origin_hard == per_stream_hard`): one slow stream holds its full window, a second stream's DATA is refused, and the test asserts the *refused* stream is the one reset upstream, the healthy stream stays usable, origin bytes never exceed the cap, exactly one refusal is counted, and every scope returns to zero
- [x] **Hysteresis**: the first stream-level `WINDOW_UPDATE` the peer sees carries the coalesced 32 bytes drained across the band, not the 16 of the first chunk — an eager implementation fails this
- [x] **Retained storage**: after a fill/drain cycle both `body.capacity` and the aggregate counters are zero
- [x] **Reload**: an existing origin's in-place limit changes and immediately refuses an over-cap reservation, a newly seen origin starts under the reloaded policy, and an open connection keeps the window and policy it advertised
- [x] Window derivation from config; a peer flooding past a *small configured* window is cut off at exactly that window
- [x] Prometheus rendering covers the origin/global `scope` labels, including a zero-valued scope

Refs #140

### DIFF
--- a/docs/OBSERVABILITY.md
+++ b/docs/OBSERVABILITY.md
@@ -54,10 +54,17 @@ logs are written through `src/http/logger.zig`.
   URLs, request IDs, or stream IDs. Current byte gauges cover bounded buffered
   responses, HTTP/1 streaming relay buffers, and HTTP/2 streaming response
   queues. `scope="stream"` reports logical queue occupancy; `scope="global"`
-  reports retained allocation, which is the quantity the global hard limit is
-  enforced against, so that gauge is directly comparable to
-  `tardigrade_buffer_config_limit_bytes{scope="global"}`. The two differ while a
-  queue is partially drained but still owns its buffer.
+  reports retained allocation, and the two differ while a queue is partially
+  drained but still owns its buffer. The `scope="global"` series is a roll-up
+  across *all* proxy paths and is **not** the quantity the configured global
+  hard limit is checked against — see the next entry.
+- the aggregate bytes the global hard limit is enforced against:
+  `tardigrade_proxy_buffer_aggregate_bytes_current{direction,scope="global"}`,
+  read directly from the account that performs the enforcement. Today that
+  covers HTTP/2 streaming response queues only, so this is the series to compare
+  with `tardigrade_buffer_config_limit_bytes{scope="global",limit="hard"}`;
+  comparing the broader `tardigrade_buffered_bytes_current{scope="global"}`
+  roll-up with that limit overstates pressure under mixed traffic.
   `tardigrade_buffer_limit_exceeded_total` carries `scope="origin"` and
   `scope="global"` alongside `scope="stream"`: an aggregate refusal means the
   upstream origin (or the process) had no room for bytes the stream itself could
@@ -94,8 +101,14 @@ logs are written through `src/http/logger.zig`.
   handshake low/high/hard watermarks. These are not described as enforced by
   OpenSSL.
 - reverse-proxy abort counters:
-  `tardigrade_proxy_client_aborts_total` and
-  `tardigrade_proxy_upstream_aborts_total`
+  `tardigrade_proxy_client_aborts_total`,
+  `tardigrade_proxy_upstream_aborts_total`, and
+  `tardigrade_proxy_local_capacity_aborts_total`. The three are mutually
+  exclusive causes for one truncated transfer: the client went away, the origin
+  went away, or this proxy ran out of buffer capacity after the response head
+  was committed. The last is local pressure and is deliberately kept out of the
+  upstream series, and out of upstream health and circuit-breaker state, so it
+  cannot be mistaken for an origin fault.
 - reverse-proxy streaming fallback event counter:
   `tardigrade_proxy_streaming_fallback_total{reason=...}` with fixed reasons
   `policy_disabled`, `retries_configured`, and `early_data_retry_semantics` for
--- a/docs/OBSERVABILITY.md
+++ b/docs/OBSERVABILITY.md
@@ -53,8 +53,20 @@ logs are written through `src/http/logger.zig`.
   to protocol-independent directions, scopes, and sides; they never include
   URLs, request IDs, or stream IDs. Current byte gauges cover bounded buffered
   responses, HTTP/1 streaming relay buffers, and HTTP/2 streaming response queues.
-  Pause/resume counters are exported as reserved zero-valued series until a later
-  backpressure slice adds production pause/resume transition sites.
+  `tardigrade_buffer_limit_exceeded_total` carries `scope="origin"` and
+  `scope="global"` alongside `scope="stream"`: an aggregate refusal means the
+  upstream origin (or the process) had no room for bytes the stream itself could
+  have held. `tardigrade_buffer_read_pauses_total{side="upstream"}` and its
+  resume counterpart record HTTP/2 streaming response queues crossing their high
+  and low watermarks — a pause means stream credit is being withheld from the
+  origin, a resume means the coalesced `WINDOW_UPDATE` went out. The
+  `side="downstream"` series remain reserved at zero until the HTTP/1 relay
+  gains pause/resume transition sites.
+- per-origin HTTP/2 buffer accounting:
+  `tardigrade_upstream_h2_pool_buffered_bytes{upstream}` and
+  `tardigrade_upstream_h2_pool_buffer_limit_exceeded_total{upstream}`. Label
+  cardinality is bounded by the number of distinct configured origins, matching
+  the other `tardigrade_upstream_h2_pool_*` series.
 - configured proxy buffer limits:
   `tardigrade_buffer_config_limit_bytes{direction,scope,limit}` for the
   per-stream low/high/hard watermarks plus per-origin/global hard-limit
@@ -300,11 +312,11 @@ Two outcome shapes exist:
 | Single header too large | `TARDIGRADE_MAX_HEADER_SIZE` | 8 KiB | `request_limits.validateHeaderSize` | `431`-class rejection |
 | All headers too large | `TARDIGRADE_MAX_HEADERS_TOTAL_SIZE` | 32 KiB | `request_limits.validateHeadersTotalSize` | `431` before body allocation |
 | Request body too large | `TARDIGRADE_MAX_BODY_SIZE` | 1 MiB | `request_limits.validateBodySize` | `413`-class rejection |
-| Proxy per-stream buffer low watermark | `TARDIGRADE_PROXY_BUFFER_PER_STREAM_LOW_WATERMARK_BYTES` | 256 KiB | proxy buffer accounting | Paired with high watermark for pause/resume decisions; must satisfy `low < high <= hard` |
-| Proxy per-stream buffer high watermark | `TARDIGRADE_PROXY_BUFFER_PER_STREAM_HIGH_WATERMARK_BYTES` | 768 KiB | proxy buffer accounting | High-watermark transition is observable through `tardigrade_buffer_high_watermark_events_total` |
+| Proxy per-stream buffer low watermark | `TARDIGRADE_PROXY_BUFFER_PER_STREAM_LOW_WATERMARK_BYTES` | 256 KiB | proxy buffer accounting; HTTP/2 stream credit | Must satisfy `low < high <= hard`. An HTTP/2 stream queue that reached the high watermark is credited nothing until it drains below this, then receives one coalesced `WINDOW_UPDATE` |
+| Proxy per-stream buffer high watermark | `TARDIGRADE_PROXY_BUFFER_PER_STREAM_HIGH_WATERMARK_BYTES` | 768 KiB | proxy buffer accounting; HTTP/2 `SETTINGS_INITIAL_WINDOW_SIZE` | High-watermark transition is observable through `tardigrade_buffer_high_watermark_events_total`. This value is advertised verbatim as the streaming HTTP/2 receive window, so a well-behaved origin stops sending once a stream holds this many unconsumed bytes. A reload applies to connections opened afterwards (SETTINGS are per connection) |
 | Proxy per-stream buffer hard limit | `TARDIGRADE_PROXY_BUFFER_PER_STREAM_HARD_LIMIT_BYTES` | 1 MiB | proxy buffer accounting | Hard-limit exceedance is observable through `tardigrade_buffer_limit_exceeded_total`; enforcement lands per proxy path as backpressure work expands |
-| Proxy per-origin buffer hard limit | `TARDIGRADE_PROXY_BUFFER_PER_ORIGIN_HARD_LIMIT_BYTES` | 0 (not enforced yet) | future aggregate proxy buffer accounting | When non-zero, must be at least the per-stream hard limit |
-| Proxy global buffer hard limit | `TARDIGRADE_PROXY_BUFFER_GLOBAL_HARD_LIMIT_BYTES` | 0 (not enforced yet) | future aggregate proxy buffer accounting | When non-zero, must be at least the per-stream hard limit |
+| Proxy per-origin buffer hard limit | `TARDIGRADE_PROXY_BUFFER_PER_ORIGIN_HARD_LIMIT_BYTES` | 0 (unlimited) | HTTP/2 streaming response queues | When non-zero, must be at least the per-stream hard limit. Caps the retained response-queue memory all streams to one origin hold together. Refused before commitment → `503 proxy_buffer_saturated`; refused after → that stream is reset and truncated. Never counted against upstream health. Reload applies immediately |
+| Proxy global buffer hard limit | `TARDIGRADE_PROXY_BUFFER_GLOBAL_HARD_LIMIT_BYTES` | 0 (unlimited) | HTTP/2 streaming response queues | When non-zero, must be at least the per-stream hard limit. Process-wide ceiling across every origin, with the same pre/post-commitment behavior; reload applies immediately |
 | Native TLS inbound ciphertext watermarks | `TARDIGRADE_TLS_INBOUND_CIPHERTEXT_LOW_WATERMARK_BYTES`, `TARDIGRADE_TLS_INBOUND_CIPHERTEXT_HIGH_WATERMARK_BYTES`, `TARDIGRADE_TLS_INBOUND_CIPHERTEXT_HARD_LIMIT_BYTES` | TLS core defaults | pure-Zig TLS record stream | Invalid ordering, capacity overflow, or reserve violations reject startup/reload |
 | Native TLS inbound plaintext watermarks | `TARDIGRADE_TLS_INBOUND_PLAINTEXT_LOW_WATERMARK_BYTES`, `TARDIGRADE_TLS_INBOUND_PLAINTEXT_HIGH_WATERMARK_BYTES`, `TARDIGRADE_TLS_INBOUND_PLAINTEXT_HARD_LIMIT_BYTES` | TLS core defaults | pure-Zig TLS record stream | High pauses carrier reads; resume occurs only after all inbound queues drain to low |
 | Native TLS outbound ciphertext watermarks | `TARDIGRADE_TLS_OUTBOUND_CIPHERTEXT_LOW_WATERMARK_BYTES`, `TARDIGRADE_TLS_OUTBOUND_CIPHERTEXT_HIGH_WATERMARK_BYTES`, `TARDIGRADE_TLS_OUTBOUND_CIPHERTEXT_HARD_LIMIT_BYTES` | TLS core defaults | pure-Zig TLS record stream | High pauses plaintext writes; hard-limit rejection does not advance write state |
--- a/docs/OBSERVABILITY.md
+++ b/docs/OBSERVABILITY.md
@@ -52,7 +52,12 @@ logs are written through `src/http/logger.zig`.
   `tardigrade_buffer_limit_exceeded_total{direction,scope}`. Labels are fixed
   to protocol-independent directions, scopes, and sides; they never include
   URLs, request IDs, or stream IDs. Current byte gauges cover bounded buffered
-  responses, HTTP/1 streaming relay buffers, and HTTP/2 streaming response queues.
+  responses, HTTP/1 streaming relay buffers, and HTTP/2 streaming response
+  queues. `scope="stream"` reports logical queue occupancy; `scope="global"`
+  reports retained allocation, which is the quantity the global hard limit is
+  enforced against, so that gauge is directly comparable to
+  `tardigrade_buffer_config_limit_bytes{scope="global"}`. The two differ while a
+  queue is partially drained but still owns its buffer.
   `tardigrade_buffer_limit_exceeded_total` carries `scope="origin"` and
   `scope="global"` alongside `scope="stream"`: an aggregate refusal means the
   upstream origin (or the process) had no room for bytes the stream itself could
--- a/src/gateway_proxy.zig
+++ b/src/gateway_proxy.zig
@@ -191,6 +191,12 @@ pub const StreamingProxyResult = struct {
     response_body_bytes: usize,
     upstream_ttfb_ms: u64,
     upstream_aborted: bool = false,
+    /// The relay was truncated because *local* proxy buffer capacity ran out,
+    /// not because the origin failed. The response head was already committed
+    /// so the status cannot change, but the origin must not be blamed for it —
+    /// counting this against upstream health would let local memory pressure
+    /// trip a healthy origin's circuit breaker.
+    local_capacity_aborted: bool = false,
 };
 
 pub fn uriComponentBytes(component: std.Uri.Component) []const u8 {
@@ -801,6 +807,7 @@ fn streamViaH2Pool(
     cancel_token: ?*const CancellationToken,
     proxy_buffer_limits: proxy_buffer_account.Limits,
     proxy_buffer_observer: proxy_buffer_account.Observer,
+    proxy_buffer_global: ?*proxy_buffer_account.Aggregate,
 ) !StreamingProxyResult {
     const deadline_ms: u32 = if (read_deadline_ms > 0)
         read_deadline_ms
@@ -833,6 +840,25 @@ fn streamViaH2Pool(
             .h2 => |conn| {
                 if (h1_pool) |p| p.recordProtocol(true);
 
+                // Aggregate capacity for this origin's queued response bytes
+                // (#140). Looked up only on the h2 path, so an ALPN-h1 origin
+                // never creates an h2 origin entry. The pointer is stable for
+                // the pool's life, surviving retries and reconnects.
+                //
+                // A failure here means the accounting itself is unavailable
+                // (allocation failure) — precisely when these limits matter
+                // most. Dropping the scope would let the request retain
+                // response bytes outside the configured per-origin bound, so
+                // this is a deterministic pre-commit rejection instead.
+                const origin_buffer_account = h2_pool.originBufferAccount(key) catch {
+                    h2_pool.release(conn);
+                    return error.ProxyBufferCapacityUnavailable;
+                };
+                const proxy_buffer_capacity = proxy_buffer_account.AggregateCapacity{
+                    .origin = origin_buffer_account,
+                    .global = proxy_buffer_global,
+                };
+
                 var authority_buf: [300]u8 = undefined;
                 const authority = if (port == default_port)
                     host
@@ -857,8 +883,9 @@ fn streamViaH2Pool(
                     .headers = extra_headers,
                     .body = buffered_body,
                     .body_mode = if (streaming_body == null) .complete else .streaming,
-                    .proxy_buffer_limits = proxy_buffer_limits,
+                    .proxy_buffer_accounting = true,
                     .proxy_buffer_observer = proxy_buffer_observer,
+                    .proxy_buffer_capacity = proxy_buffer_capacity,
                 }) catch |err| {
                     // Nothing has reached the client yet: evict the dead
                     // connection so new requests do not pick it, and retry
@@ -893,6 +920,10 @@ fn streamViaH2Pool(
                     conn.finishStreaming(stream);
                     if (!conn.healthy()) h2_pool.evict(key, conn);
                     h2_pool.release(conn);
+                    // Local capacity, not the origin: nothing has reached the
+                    // client, so this becomes a clean 503 and never counts
+                    // against the origin. Retrying would hit the same wall.
+                    if (err == error.BufferLimitExceeded) return error.ProxyBufferCapacityUnavailable;
                     if (streaming_body == null and attempt == 0 and (err == error.Http2GoAway or err == error.Http2ConnectionClosed or err == error.Http2StreamReset)) {
                         continue;
                     }
@@ -921,6 +952,7 @@ fn streamViaH2Pool(
 
                 var body_bytes: usize = 0;
                 var aborted = false;
+                var local_capacity_aborted = false;
                 if (body_allowed) {
                     while (true) {
                         if (cancelStopped(cancel_token)) {
@@ -928,12 +960,19 @@ fn streamViaH2Pool(
                             h2_pool.release(conn);
                             return error.RequestCancelled;
                         }
-                        const n = conn.readStreamingBody(stream, read_buf) catch {
-                            // Upstream failed mid-body after the head went
-                            // downstream: report an aborted relay (the client
-                            // sees the truncated chunked body); other streams
-                            // on the connection are unaffected unless the
-                            // whole connection died (handled below).
+                        const n = conn.readStreamingBody(stream, read_buf) catch |err| {
+                            // Failed mid-body after the head went downstream:
+                            // report an aborted relay (the client sees the
+                            // truncated chunked body); other streams on the
+                            // connection are unaffected unless the whole
+                            // connection died (handled below).
+                            //
+                            // The status can no longer change, but the *cause*
+                            // still matters: a local buffer-capacity abort is
+                            // this proxy running out of room, and blaming the
+                            // origin for it would let local memory pressure
+                            // trip a healthy origin's failure policy.
+                            local_capacity_aborted = err == error.BufferLimitExceeded;
                             aborted = true;
                             break;
                         };
@@ -969,6 +1008,7 @@ fn streamViaH2Pool(
                     .response_body_bytes = body_bytes,
                     .upstream_ttfb_ms = ttfb_ms,
                     .upstream_aborted = aborted,
+                    .local_capacity_aborted = local_capacity_aborted,
                 };
             },
         }
@@ -1974,6 +2014,10 @@ pub fn executeStreamingHttpProxyRequest(
     sticky_set_cookie: ?[]const u8,
     cancel_token: ?*const CancellationToken,
     proxy_buffer_observer: proxy_buffer_account.Observer,
+    /// Process-wide proxy buffer aggregate (#140). Streams reserve against it
+    /// before retaining queued body bytes, so aggregate memory stays bounded no
+    /// matter how many origins or concurrent streams are slow.
+    proxy_buffer_global: ?*proxy_buffer_account.Aggregate,
     pool: ?*http.upstream_pool.UpstreamPool,
     /// Optional per-origin HTTP/2 multiplexing pool (#145).
     h2_pool: ?*http.upstream_h2.H2ConnPool,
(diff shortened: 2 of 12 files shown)


src/gateway_proxy.zig
```diff
@@ -812,6 +813,17 @@ fn streamViaH2Pool(
     var key_buf: [300]u8 = undefined;
     const key = std.fmt.bufPrint(&key_buf, "{s}:{s}:{d}", .{ if (is_tls) "h2" else "h2c", host, port }) catch host;
 
+    // Aggregate capacity for this origin's queued response bytes (#140). The
+    // origin entry is created once and lives for the pool's life, so the
+    // pointer stays valid across retries and connection churn. Failing to
+    // create it (OOM) only costs origin-scope enforcement — the global scope
+    // and the per-stream bound still apply.
+    const origin_buffer_account = h2_pool.originBufferAccount(key, proxy_buffer_limits.per_origin_hard_limit) catch null;
```
Jaruso: **[P1] Do not fail open when the configured origin cap cannot be created.**

`catch null` silently disables `.origin` enforcement for this request. Under allocator pressure—exactly when this protection matters—the request can therefore retain response bytes without the configured per-origin bound. #140's failure policy says unavailable local proxy capacity before commitment should fail deterministically (503), not weaken the configured limit.

Please propagate a distinct local-capacity error instead of dropping the scope, e.g. conceptually:

```zig
const origin_buffer_account = h2_pool.originBufferAccount(
    key,
    proxy_buffer_limits.per_origin_hard_limit,
) catch |err| switch (err) {
    error.OutOfMemory => return error.ProxyBufferCapacityUnavailable,
    else => return err,
};
```

Then map `ProxyBufferCapacityUnavailable` to a pre-commit 503 without blaming the upstream. Moving this lookup until after `.h2` acquisition would also avoid allocating an h2-origin entry for ALPN-h1/failing connections.

src/gateway_shutdown.zig
```diff
@@ -630,6 +630,10 @@ pub fn applyReloadedRuntimeConfig(cfg: *const edge_config.EdgeConfig, state: *Ga
     state.max_total_connection_memory_bytes = cfg.max_total_connection_memory_bytes;
     state.connection_memory_estimate_bytes = if (cfg.max_connection_memory_bytes > 0) cfg.max_connection_memory_bytes else MAX_REQUEST_SIZE;
     state.proxy_buffer_limits = cfg.proxy_buffer_limits;
+    // Aggregate enforcement follows the reload immediately; the per-connection
+    // HTTP/2 receive window cannot (SETTINGS were already sent) and applies to
+    // connections opened after the reload.
```
Jaruso: **[P1] The reloaded high watermark never reaches later connections.**

The comment says the new window applies to connections opened after reload, but only `state.proxy_buffer_limits` and the global aggregate are updated here. `H2ConnPool.config.stream_recv_window` remains the startup value, and `acquire()` passes that unchanged value to every future `PooledH2Conn.init`, so reconnecting after reload still advertises the old window.

There is a second consistency problem: new streams opened on an *existing* connection get the reloaded `proxy_buffer_limits`, while `stream.recv_window` remains that connection's old advertised window. If the operator lowers the per-stream hard limit below the old window, a compliant peer can send within its negotiated credit and be aborted by the newer local hard limit.

Please make the reload policy explicit and internally consistent—e.g. update the pool's policy for future connections and pin the complete per-stream low/high/hard policy to each connection, or recycle/update connections when the per-stream policy changes. Add a reload test that lowers the window/limits and proves both old-connection and new-connection behavior.

src/http/upstream_h2.zig
```diff
@@ -1489,6 +1530,19 @@ pub const H2ConnPool = struct {
...
+    /// The origin-scope buffer aggregate for `key`, refreshed with the current
+    /// configured hard limit so a reload takes effect on the next reservation.
+    /// The pointer is stable for the pool's lifetime.
+    pub fn originBufferAccount(
+        self: *H2ConnPool,
+        key: []const u8,
+        hard_limit: usize,
+    ) !*proxy_buffer_account.Aggregate {
+        const counters = try self.originCounters(key);
+        counters.buffer.setHardLimit(hard_limit);
```
Jaruso: **[P1] This does not make an origin-limit reload effective on the next reservation.**

`setHardLimit()` runs only when another request calls `originBufferAccount()`. Existing streams can continue reserving against the old origin limit indefinitely after reload if no new request starts for that origin. Also, because the new limit is supplied from a request-local config snapshot, a request that started before reload can arrive here later and write the *old* value back over the new one.

The shared policy should have one writer: store the current per-origin limit in `H2ConnPool` and update it from the reload path, applying it to every existing `H2OriginCounters.buffer` and to the default used for future origins. `streamViaH2Pool` should only fetch the stable aggregate pointer; it should not mutate shared configuration from a request snapshot.

src/http/upstream_h2.zig
```diff
@@ -479,33 +489,43 @@ pub const Stream = struct {
...
+                after.limit_exceeded_events > before.limit_exceeded_events,
+            );
         }
     }
 
     fn releaseQueuedBody(self: *Stream, bytes: usize) void {
         if (bytes == 0) return;
         if (self.proxy_body_account) |*account| {
             account.release(bytes) catch unreachable;
+            self.proxy_buffer_capacity.release(account.direction, bytes);
```
Jaruso: **[P1] Returning aggregate capacity here does not return the memory it is supposed to bound.**

`acknowledgeStreamingBody()` calls this release and only afterward calls `compactAcknowledgedBody()`, which uses `shrinkRetainingCapacity`. That shrinks `body.items.len` but deliberately keeps the ArrayList backing allocation. A long-lived stream can therefore fill a 768 KiB window, drain to zero, return ~768 KiB to the origin/global aggregates, and still own roughly that allocation. Repeating this across many active streams lets RSS grow by `N × window` while `currentBytes()` is near zero—the exact concurrency multiplication #140 is intended to prevent.

Please make the aggregate reservation track retained storage, not only logical queued length, or actually relinquish/shrink backing storage before returning aggregate capacity. A bounded/ring queue is another option. The regression test should fill/drain multiple long-lived streams and assert retained queue capacity (or allocator bytes) stays within the aggregate policy; checking only that counters return to zero will miss this failure mode.

src/http/upstream_h2.zig
```diff
@@ -479,33 +489,43 @@ pub const Stream = struct {
...
-            };
+        const account = if (self.proxy_body_account) |*a| a else return;
+        const direction = account.direction;
+        // Aggregate capacity first: a stream may not retain bytes its origin or
+        // the process as a whole has no room for.
+        self.proxy_buffer_capacity.reserve(direction, bytes) catch |err| {
+            if (self.proxy_buffer_observer) |observer| {
+                observer.recordAggregateLimitExceeded(direction, proxy_buffer_account.aggregateFailureScope(err));
+            }
+            return error.BufferLimitExceeded;
```
Jaruso: **[P1] An aggregate refusal must make the stream terminal immediately.**

After this returns, `handleData()` sets `s.err`/`local_abort`, but the stream remains in `self.streams` and later DATA frames still go through `recordQueuedBody()` because `handleData()` does not short-circuit errored/local-aborted streams. If another stream drains and aggregate capacity becomes available before the worker calls `finishStreaming()`, a later frame on this already-aborted stream can successfully reserve and append bytes. `readStreamingBody()` checks queued `avail` before `stream.err`, so those post-failure bytes can even be delivered downstream before the error finally surfaces.

On the first hard-limit refusal, send/queue `RST_STREAM(CANCEL)` promptly and make all subsequent DATA for that stream discard-only with no new reservation/accounting. That also prevents repeated limit-event increments while the worker is blocked on a slow downstream write.

src/gateway_proxy.zig
```diff
@@ -859,6 +871,7 @@ fn streamViaH2Pool(
                     .body_mode = if (streaming_body == null) .complete else .streaming,
                     .proxy_buffer_limits = proxy_buffer_limits,
                     .proxy_buffer_observer = proxy_buffer_observer,
+                    .proxy_buffer_capacity = proxy_buffer_capacity,
```
Jaruso, line 893: **[P1] Preserve local-capacity failures across the downstream commitment boundary.**

The new capacity failure can occur *before* the response is committed: the h2 reader may process upstream HEADERS, signal the worker, then process/reject DATA before the worker writes the head downstream. `waitStreamingResponseHead()` currently returns success whenever `status != null`, even if `stream.err` is already `BufferLimitExceeded`, so this race commits the origin's status and then truncates instead of producing the required pre-commit 503.

The same loss of cause also hurts post-commit behavior: the relay collapses the failure into `upstream_aborted = true`, and `gateway_proxy_runtime` treats every `upstream_aborted` as `recordUpstreamFailure(...)`. A healthy origin can therefore be penalized / trip failure policy because *local* origin/global memory capacity was exhausted.

Please carry an explicit cause and commitment state. For example:

```zig
const AbortCause = enum { none, upstream, local_capacity };
```

Before the worker marks the downstream response committed, `OriginBufferLimitExceeded` / `GlobalBufferLimitExceeded` should surface as a local-capacity error mapped to 503 and must not count against origin health. After commitment, the same cause should truncate + RST + increment the limit metric/log reason, but still must not be recorded as an upstream-origin failure.

src/http/upstream_h2.zig
```diff
@@ -886,7 +914,7 @@ pub fn H2Conn(comptime Transport: type) type {
                 .id = id,
                 .send_window = self.peer_initial_window,
                 .streaming = streaming,
-                .recv_window = if (streaming) @as(i64, OUR_INITIAL_WINDOW) else 0,
+                .recv_window = if (streaming) @as(i64, self.stream_recv_window) else 0,
```
Jaruso, line 1107: **[P2] Setting the initial window to `high` does not enforce the configured low-watermark resume behavior.**

The receive window now gives the correct upper bound, but `acknowledgeStreamingBody()` still increments `recv_window` and sends a stream `WINDOW_UPDATE` for *every* downstream-consumed chunk. Once the queue reaches `high`, the peer therefore resumes after the first few bytes drain, not after the queue falls to `low`. The low watermark currently only re-arms the accounting boolean; it has no backpressure effect. That misses #140's PR-2 criterion to pause at high and resume below low, and can let one slow stream continuously reclaim aggregate capacity instead of allowing the queue to drain through the hysteresis band.

Please accumulate drained-but-not-yet-credited bytes while the account remains above-high, and send the accumulated `WINDOW_UPDATE` when `Account.release()` crosses the low watermark. Add a deterministic test proving no stream credit is emitted between high and low, followed by one resume when low is crossed (and hook the pause/resume metric transition as appropriate).

src/http/upstream_h2.zig
```diff
@@ -2665,6 +2738,130 @@ test "streaming stream fails when the peer overruns the advertised window" {
...
+
+test "aggregate exhaustion aborts the stream and leaves no scope holding bytes" {
+    const fds = try makeSocketpair();
+    var saw_rst = std.atomic.Value(bool).init(false);
+    const server = try std.Thread.spawn(.{}, cannedLimitExceededStreamingServer, .{ fds[1], &saw_rst });
+
+    var observer = TestProxyBufferObserver{};
+    // Room per stream, but the origin as a whole is already nearly full — the
+    // case a large fan-out of slow streams creates.
+    var origin = proxy_buffer_account.Aggregate.init(.origin, 8);
```
Jaruso: **[P2] This aggregate exhaustion test cannot occur with a valid production configuration.**

The test sets `origin` hard limit to 8 while `smallProxyBufferLimits(64)` sets the per-stream hard limit to 64. `Limits.validate()` explicitly rejects any non-zero per-origin limit smaller than the per-stream hard limit. Also, despite the comment, this aggregate starts at `current == 0`; it is not "already nearly full." The test is therefore proving a synthetic single-stream cap ordering that production cannot configure, not the issue's many-slow-stream aggregate case.

Please test the real scenario with valid limits and shared concurrency: e.g. `stream_hard = 64`, `origin_hard >= 64`, hold most of the origin reservation with one slow stream, then have another stream's DATA cross the remaining origin capacity. Ideally use two or more concurrent streams on the same h2 connection and assert the refusing stream is reset, the already-buffered/other stream remains usable, aggregate bytes never exceed the configured cap, and all scopes return to zero after drain/teardown.

Jaruso (commented):
Blocking review findings. I reviewed this against #140's stated PR-2/PR-4 scope and the full issue invariants. The core direction is good—configured h2 window, lock-free aggregate reservations, rollback, scope metrics, and cleanup are all sensible—but the current implementation does not yet provide the memory/failure/reload guarantees the PR description claims.

The inline comments cover the blocking issues: aggregate capacity is released while the ArrayList still retains the backing allocation; aggregate refusal is not terminal; local-capacity failures are not preserved across the downstream commitment boundary and can be counted as origin failures; per-origin and per-stream reload behavior is stale/inconsistent; configured origin enforcement fails open on allocation failure; and the h2 flow-control path never actually uses the low watermark for resume hysteresis. The new aggregate socket test also uses a configuration that production validation forbids, so it does not satisfy the concurrent-slow-stream acceptance case.

One process issue as well: please remove `Closes #140` (use `Refs #140` or equivalent). This PR explicitly leaves the issue's PR-3 request-direction / HTTP/1 pause-resume work out of scope, while #140's acceptance criteria still require that work and its slow-origin/upload/pause-resume coverage. Merging this PR should not auto-close the parent issue before that remaining slice lands.

CI is green for the unit/build matrix on this head. The GitHub Actions `Integration tests` job is skipped; the PR reports a local `zig build test-integration` pass, but the acceptance gaps above are behavioral and are not covered by the current tests.

Note: GitHub will not allow the connected account to submit `REQUEST_CHANGES` on its own PR, so this is posted as a `COMMENT` review; the findings below are blocking and should be treated as change requests.

Jaruso (commented):
Re-review of the current head (`9081026`): there are no source changes since my previous review, so the eight existing inline blockers remain applicable and I am not duplicating those threads.

The PR description has been rewritten to say those findings are fixed (hysteresis, retained-storage accounting, deterministic pre/post-commit handling, discard-only refused streams, reload propagation, OOM -> 503, valid concurrent aggregate coverage, and `Refs #140`), but the branch still contains the original implementation. For example, the current source still uses `originBufferAccount(..., hard_limit) catch null`, releases aggregate bytes before `shrinkRetainingCapacity`, sends stream credit on each acknowledgement, and only updates the global aggregate during reload. The head commit itself also still says `Closes #140`.

Please push the implementation changes described in the updated PR body. Once the head SHA moves, I can re-evaluate each existing blocker against the new diff. CI is green for this unchanged commit, but it is the same CI run as the prior review and does not validate the claimed-but-unpushed fixes.

src/gateway_proxy.zig
```diff
@@ -893,6 +920,10 @@ fn streamViaH2Pool(
                     conn.finishStreaming(stream);
                     if (!conn.healthy()) h2_pool.evict(key, conn);
                     h2_pool.release(conn);
+                    // Local capacity, not the origin: nothing has reached the
+                    // client, so this becomes a clean 503 and never counts
+                    // against the origin. Retrying would hit the same wall.
+                    if (err == error.BufferLimitExceeded) return error.ProxyBufferCapacityUnavailable;
```
Jaruso, line 940: **[P1] The pre-commit capacity cause is still not preserved across every path or across the actual commit boundary.**

This mapping only runs after `waitStreamingResponseHead()`. There are two remaining holes:

1. With a full streaming upload, the origin can send HEADERS/DATA while `relayStreamingUploadToHttp2()` is still sending the request body. If the response queue hits origin/global capacity, the reader stores `stream.err = BufferLimitExceeded`; the next `writeStreamingRequestBody()` returns that error through the *earlier upload catch*, which currently just `return err`. `gateway_proxy_runtime` then treats it as an upstream failure and returns 502 instead of the required local-capacity 503.
2. Even after this wait succeeds, it unlocks `state_mutex` before `writeStreamedUpstreamResponseHeadFromHeaders()` starts. The reader can reject DATA in that gap. The worker has already passed this check, so it still writes the origin status and converts what was physically a pre-head refusal into a post-commit truncation.

Please preserve a distinct response-capacity abort cause on the stream (rather than generic `BufferLimitExceeded`) and add a mutex-protected commitment state transition immediately before head relay begins. Conceptually:

```zig
const DownstreamState = enum { precommit, committing, committed };
const AbortCause = enum { none, upstream, local_capacity };

// under state_mutex, immediately before starting the head write
if (stream.abort_cause == .local_capacity)
    return error.ProxyBufferCapacityUnavailable;
stream.downstream_state = .committing;
```

Then any local-capacity error observed while the state is still `.precommit` must become `503 proxy_buffer_saturated`, including errors that surface through request-upload writes; once `.committing` begins, truncate/reset without blaming the origin. Add a deterministic test for an early h2 response arriving while a streaming upload is still in progress, plus a barrier-controlled test for the HEADERS→head-write race.

src/http/upstream_h2.zig
```diff
@@ -1199,8 +1347,15 @@ pub fn H2Conn(comptime Transport: type) type {
             const maybe_stream = self.streams.get(fr.stream_id);
             var replenish_stream = false;
             var flow_violation = false;
+            var reset_stream = false;
```
Jaruso: **[P1] A stream flow-control violation still never sends the required upstream reset.**

`reset_stream` is only set by the `BufferLimitExceeded` branch below. When `recv_window < 0`, this code sets `s.err = Http2FlowControlError` and `flow_violation = true`, but the post-unlock path only broadcasts a condition. `finishStreaming()` will *also* skip its RST because its predicate requires `stream.err == null or stream.local_abort`; a flow-control error has an error and is not a local-capacity abort.

So an origin that overruns the configured window is never told to stop the stream. Because the connection-level window is replenished for every subsequent DATA frame, a violating peer can keep sending frames on that errored stream until the worker eventually finishes it. That misses #140's requirement to stop/reset the offending upstream stream and the HTTP/2 stream-error semantics this code is modeling.

Please make the reset action carry an error code and emit `RST_STREAM(FLOW_CONTROL_ERROR)` on the first window overrun, e.g.:

```zig
var reset_code: ?u32 = null;
...
if (s.recv_window < 0) {
    s.err = error.Http2FlowControlError;
    if (!s.rst_sent and s.wire_opened) {
        s.rst_sent = true;
        reset_code = 0x3; // FLOW_CONTROL_ERROR
    }
    deliver = false;
}
```

Then keep later DATA discard-only and add a socketpair assertion that the peer actually receives the RST for the violating stream.

src/http/upstream_h2.zig
```diff
@@ -1218,21 +1373,24 @@ pub fn H2Conn(comptime Transport: type) type {
...
                 if (deliver) {
-                    s.recordQueuedBody(fr.payload.len) catch |err| {
-                        if (err == error.BufferLimitExceeded) {
+                    s.enqueueBody(self.allocator, fr.payload) catch |err| switch (err) {
+                        // Local capacity is exhausted at some scope. Make the
+                        // stream terminal here and now: reset it upstream so
+                        // the origin stops sending, and leave it discard-only
+                        // for any DATA already in flight.
+                        error.BufferLimitExceeded => {
                             if (s.err == null) s.err = err;
```
Jaruso, line 1478: **[P1] The new reader-side `wire_opened` check introduces an unsynchronized cross-thread read/write.**

`wire_opened` is still assigned in `sendRequest()` as a plain `stream.wire_opened = true` *after* the request HEADERS write and without `state_mutex`; its field comment still says it is written/read only by the owning worker. This new branch reads it from the reader thread under `state_mutex`.

A fast origin can respond as soon as the HEADERS syscall completes, so the reader can reach this line concurrently with (or just before) the worker's plain assignment. Besides the Zig data race, it can observe `false` and skip the immediate RST even though the stream is already on the wire.

Please move `wire_opened` into the shared state discipline. One safe pattern is to mark it under `state_mutex` immediately before starting the HEADERS write; a write failure already poisons `conn_err`, so `finishStreaming()` will not emit a bogus reset for a request the peer never received. Alternatively make it atomic, but the mutex approach keeps the stream state machine coherent. Update the comment and add a fast-response test that forces DATA to arrive as the request-head write completes.

src/http/upstream_h2.zig
```diff
@@ -512,7 +593,18 @@ pub const Stream = struct {
...
 
-    fn compactAcknowledgedBody(self: *Stream, bytes: usize) void {
+    /// Give the queue's backing allocation back once it drains empty, and with
+    /// it the aggregate reservation that covered it. Without this the queue
+    /// would keep its peak allocation for the stream's whole life while the
+    /// aggregate counters read zero — the concurrency multiplication the
+    /// aggregate limits exist to prevent.
+    fn releaseRetainedStorage(self: *Stream) void {
+        if (self.proxy_capacity_reserved == 0) return;
+        self.proxy_buffer_capacity.release(self.accountDirection(), self.proxy_capacity_reserved);
```
Jaruso: **[P2] The global current-byte gauge now under-reports the memory that the global hard limit actually reserves.**

This retained-capacity accounting is the right fix for the original RSS hole, but the observer path still has the old logical-byte semantics: `releaseQueuedBody()` calls `observer.releaseReservation(..., bytes)` on every downstream acknowledgement, and `Metrics.releaseProxyBufferReservation()` decrements both the stream and `scope="global"` current gauges by that logical amount. The aggregate reservation here is intentionally *not* released until the backing allocation is freed.

Example: a queue allocates 64 bytes, then the client drains 16. The actual global aggregate remains 64 (correct, because the ArrayList still owns 64), but `tardigrade_buffered_bytes_current{direction="upstream_to_downstream",scope="global"}` becomes 48. Operators therefore cannot compare the advertised global current gauge with `TARDIGRADE_PROXY_BUFFER_GLOBAL_HARD_LIMIT_BYTES`, even though the issue explicitly requires current-byte accounting to follow owned bytes and exposes config limits so those gauges are interpretable.

Please separate logical stream occupancy from aggregate retained-capacity deltas. The global aggregate gauge should move when `proxy_buffer_capacity.reserve/release` moves, not when the stream account drains. A dedicated h2 aggregate-current series is also viable if preserving the existing cross-path global gauge requires compatibility, but there must be a stable gauge that reports the same retained bytes the hard limit enforces. Add a partial-drain test (not just drain-to-zero) that proves the reported aggregate/global current value stays at the retained capacity until `clearAndFree()`.

src/http/upstream_h2.zig
```diff
@@ -2665,6 +2971,298 @@ test "streaming stream fails when the peer overruns the advertised window" {
...
+
+    const refused_id: u32 = refused.id;
+    conn.finishStreaming(refused);
+    conn.finishStreaming(slow);
+    conn.deinit();
+    server.join();
+    _ = std.c.close(fds[1]);
+
+    // The refused stream — not the healthy one — was reset upstream.
+    try testing.expectEqual(refused_id, rst_stream_id.load(.acquire));
```
Jaruso, line 3292: **[P1] This new concurrency regression test currently fails CI because the test itself resets the supposedly healthy stream.**

The current CI failure is `expected 3, found 1` here on Ubuntu/macOS and the appliance suites. `cannedTwoStreamServer()` records the *first* RST it sees and returns. But the server sends the slow stream's DATA without `END_STREAM`, and this test later calls `finishStreaming(slow)`. Since that stream is still wire-open, `finishStreaming()` legitimately sends `RST_STREAM(CANCEL)` for stream 1. That CANCEL races the reader thread's reset for refused stream 3 and can be the first frame the server observes — exactly the red-CI result.

Please make the healthy stream actually complete before finishing it (for example send its first DATA with `END_STREAM`) or have the server collect reset IDs until it observes the refused stream rather than returning on the first unrelated RST. The stronger assertion should be: refused stream receives the capacity-reset, the healthy stream reaches clean END_STREAM and is never reset for the capacity event, aggregate bytes never exceed the cap, and all reservations return to zero.

Jaruso (commented):
Re-review of `485425e` after the review-fix rewrite.

The prior findings were mostly addressed correctly: origin accounting no longer fails open, the pool owns reload policy, connections pin their advertised stream limits, retained queue storage remains aggregate-reserved until actually freed, locally-aborted streams become discard-only, response-capacity failures are kept out of upstream health after commitment, hysteresis now withholds/re-coalesces stream credit, the concurrency test uses a production-valid aggregate shape, and `Closes #140` was changed to `Refs #140`.

This is still **not merge-ready**, though. There are five actionable findings inline below. The most immediate objective blocker is CI: the new `origin capacity bounds concurrent streams...` test fails on Ubuntu, macOS, and both appliance-profile unit suites (`expected 3, found 1` at `upstream_h2.zig:3121`). In addition, the downstream commitment boundary is still not linearized, the early-response/full-upload path can still misclassify local response-capacity exhaustion as an upstream 502/failure, the new reader-side `wire_opened` read introduces an unsynchronized cross-thread access, stream flow-control violations are not actually reset upstream, and the global current-byte gauge no longer represents the retained allocation that the global hard limit enforces.

Minor cleanup while touching metrics: the Prometheus HELP strings for `tardigrade_buffer_read_pauses_total` / `...resumes_total` still say "Reserved for future" even though this PR now emits real upstream transitions.

GitHub does not allow this account to submit `REQUEST_CHANGES` on its own PR, so this COMMENT review should be treated as blocking until the inline findings and red CI are cleared.

Jaruso (commented):
Re-review of head `3b8e9a2e0461bd707394bbc9de05ab034f53e154`: I re-checked the five blockers from the prior pass and the changes that addressed them. I do not have any new blocking findings.

Verified:
- the concurrent-stream aggregate test no longer races an unrelated healthy-stream CANCEL, and CI now passes it;
- pre-commit local-capacity failures are preserved through both the streaming-upload path and the HEADERS→downstream-head commitment boundary via an explicit abort cause plus `beginDownstreamCommit`;
- stream receive-window overruns now emit `RST_STREAM(FLOW_CONTROL_ERROR)` and have socket-level coverage;
- `wire_opened` is now published under `state_mutex` before the HEADERS write, eliminating the reader/worker data race;
- aggregate/global current-byte observability now follows retained allocation rather than logical queue occupancy, including a partial-drain regression test;
- the Prometheus pause/resume HELP text now describes the production behavior rather than calling the series reserved.

The earlier architecture fixes (pinned per-connection buffer policy, reload propagation, fail-closed origin accounting, terminal discard-only aborts, retained-storage release, valid aggregate concurrency coverage, and high/low hysteresis) remain intact.

CI run 31347214119 is green across unit tests on Ubuntu/macOS/ARM, appliance-profile tests/integration coverage, builds, lint/format, security, packaging, perf, and protocol suites. The standalone GitHub Actions job named `Integration tests` is still skipped, but the appliance-profile full integration suites ran successfully where enabled.

This PR correctly uses `Refs #140` / explicitly states it does not close #140 because the PR-3 request-direction + HTTP/1 slice remains outstanding.

Verdict: merge-ready for the scoped PR-2/aggregate slice of #140.

src/gateway_proxy.zig
```diff
@@ -969,6 +1034,7 @@ fn streamViaH2Pool(
                     .response_body_bytes = body_bytes,
                     .upstream_ttfb_ms = ttfb_ms,
                     .upstream_aborted = aborted,
+                    .local_capacity_aborted = local_capacity_aborted,
```
Jaruso, line 1037: **[P2] Keep local-capacity truncation out of the upstream-abort metric.**

`aborted` is set for every body-read failure, including `BufferLimitExceeded`, so this result currently carries both `upstream_aborted = true` and `local_capacity_aborted = true`. `gateway_proxy_runtime` correctly skips `recordUpstreamFailure()` when the local flag is set, but it still unconditionally calls `metricsRecordProxyUpstreamAbort()` whenever `upstream_aborted` is true. The Prometheus HELP text defines that counter as "Total proxied transfers aborted by upstream origins", so local memory pressure still looks like an origin abort in observability.

Please make the classifications mutually exclusive, for example:

```zig
.upstream_aborted = aborted and !local_capacity_aborted,
.local_capacity_aborted = local_capacity_aborted,
```

or gate `metricsRecordProxyUpstreamAbort()` on `!streamed.local_capacity_aborted`. Add a regression proving a post-commit aggregate refusal increments the buffer-limit metric/log path but **not** `tardigrade_proxy_upstream_aborts_total`.

docs/OBSERVABILITY.md
```diff
@@ -52,9 +52,26 @@ logs are written through `src/http/logger.zig`.
...
   to protocol-independent directions, scopes, and sides; they never include
   URLs, request IDs, or stream IDs. Current byte gauges cover bounded buffered
-  responses, HTTP/1 streaming relay buffers, and HTTP/2 streaming response queues.
-  Pause/resume counters are exported as reserved zero-valued series until a later
-  backpressure slice adds production pause/resume transition sites.
+  responses, HTTP/1 streaming relay buffers, and HTTP/2 streaming response
+  queues. `scope="stream"` reports logical queue occupancy; `scope="global"`
+  reports retained allocation, which is the quantity the global hard limit is
+  enforced against, so that gauge is directly comparable to
+  `tardigrade_buffer_config_limit_bytes{scope="global"}`. The two differ while a
```
Jaruso: **[P2] The generic global gauge is not the same population the global hard limit currently enforces.**

The retained-byte observer is now used by HTTP/1/request-side `ProxyBufferReservation` and the buffered compatibility path as well as HTTP/2, so `tardigrade_buffered_bytes_current{scope="global"}` is an all-accounted-path retained-memory gauge. But this PR only passes `proxy_buffer_global_account` into the HTTP/2 streaming response path; the HTTP/1 fixed buffers are explicitly not reserved against aggregate scopes. A mixed workload can therefore legitimately report this gauge above `TARDIGRADE_PROXY_BUFFER_GLOBAL_HARD_LIMIT_BYTES` even while the actual enforced H2 aggregate is below its cap.

That makes the new "directly comparable" claim incorrect. Please either expose a dedicated current gauge sourced from the same aggregate that enforces the H2 global cap (e.g. `proxy_buffer_global_account.currentBytes(.upstream_to_downstream)`) and compare *that* to the limit, or broaden the hard-limit reservation to every path counted here. If the latter stays in PR 3, keep this generic gauge as all accounted retained bytes but document that the configured global hard limit in this slice only governs the H2 response aggregate. The same direct-comparability wording in `PROXY_STREAMING.md` should be adjusted too.

Jaruso (commented):
Follow-up on `3b8e9a2`: the previous transport/state-machine blockers are fixed. I verified the mutex-linearized downstream commitment boundary, early-response capacity propagation during uploads, `RST_STREAM(FLOW_CONTROL_ERROR)` on receive-window overruns, the `wire_opened` synchronization fix, retained-allocation accounting, the concurrent-stream regression, and the updated pause/resume behavior. The current CI run is green across the unit/build/appliance matrix.

Two P2 observability issues remain before I would merge this slice:

1. A post-commit local-capacity truncation is still returned with `upstream_aborted = true`, so `gateway_proxy_runtime` increments `tardigrade_proxy_upstream_aborts_total` even though the metric explicitly means "aborted by upstream origins." The health/circuit-breaker classification is fixed, but the exported abort classification is still wrong.
2. `tardigrade_buffered_bytes_current{scope="global"}` now correctly tracks retained allocation, but it includes buffered/HTTP/1/request-side retained bytes while `TARDIGRADE_PROXY_BUFFER_GLOBAL_HARD_LIMIT_BYTES` is enforced only by the HTTP/2 response aggregate in this PR. Therefore the documentation's new claim that the generic global gauge is directly comparable with that hard limit is false under mixed traffic. Either expose an H2 aggregate-current gauge sourced from the same `proxy_buffer_global_account` that enforces the limit, or extend the aggregate reservation to the other paths (which is likely PR-3 scope). Until then, document the distinction rather than equating the two.

No new P1 transport/backpressure correctness findings on this head. #140 should remain open as the PR already states because PR-3 request-direction/HTTP/1 work is still outstanding.

Jaruso (commented):
Re-reviewed the new head (`5892114`) against the two remaining P2 findings from the previous pass and re-checked the surrounding #140 HTTP/2 response/aggregate behavior for regressions.

Verified fixed:

- Local proxy-capacity truncations are no longer attributed to `tardigrade_proxy_upstream_aborts_total`; they now have the separate `tardigrade_proxy_local_capacity_aborts_total` path, while the upstream-health classification remains protected from local saturation.
- The generic `tardigrade_buffered_bytes_current{scope="global"}` roll-up is no longer documented as the enforced aggregate. A dedicated `tardigrade_proxy_buffer_aggregate_bytes_current{direction,scope="global"}` gauge is sourced directly from `proxy_buffer_global_account`, and the docs correctly distinguish the all-path retained-memory roll-up from the H2 response aggregate governed by the current global hard limit.
- Production startup initializes the aggregate from the configured global hard limit, reload updates it in place, and the new metrics test helpers initialize the account before rendering it.
- The previous transport/state-machine fixes remain intact: deterministic pre/post-commit classification, early-response/upload capacity mapping, immediate flow-control RST, synchronized `wire_opened`, retained-storage accounting, high/low credit hysteresis, aggregate refusal isolation, and clean reservation teardown.
- The PR still correctly uses `Refs #140` / explicitly does not close #140 because PR-3 request-direction and HTTP/1 observability work remains outstanding.

CI for this head is green.

I found no new blocking issues in this pass. From a code-review standpoint, this is ready to merge.

Issue #144, opened 2026-06-06
perf: close out final A/B evidence for merged writev response writer


Jaruso:
## Goal

Finish the **last performance-evidence obligation** for the response-writer work already merged in PR #598.

The writev/scatter-gather implementation itself is complete. Do **not** rebuild the response writer, add another abstraction, or reopen already-resolved correctness work. This ticket remains open only because the final proxy-performance fix changed the PR head *after* the previously posted A/B comparison.

---

### Implementation — complete

PR #598 (`perf: add writev response writer`) merged as commit `a7dfc9cfe156cb67b5d06842dc0bb26a200054ea` from final head:

```text
e302ca96046ad724d541be026c17e904d4d6e90d
```

It provides the intended #144 architecture:

- [x] plaintext HTTP/1 raw-socket gathered-write support;
- [x] bounded/allocation-free common header assembly with oversized-header fallback;
- [x] partial `writev` retry/cursor handling, including correctness coverage for short writes/error paths;
- [x] static buffered responses use the tracked response writer;
- [x] common buffered reverse-proxy responses use the optimized response writer;
- [x] TLS/native-encrypted/generic writers retain correct non-writev fallback behavior;
- [x] HEAD/empty-body paths use the correct single-write behavior;
- [x] response-write observability exists:
  - `tardigrade_response_write_mode_total{mode=...}`;
  - `tardigrade_response_writev_iovecs_total`;
  - `tardigrade_response_write_errors_total{mode=...}`;
- [x] docs describe `writev`, `single_write`, `tls_buffered`, and fallback semantics;
- [x] unit/integration/perf-smoke CI was green on the merged head.

### Benchmark history — why one row is still open

The PR did receive same-host before/after evidence comparing base:

```text
d39353267f51116e0d30bdccec3614b3019a12db
```

to intermediate head:

```text
f4f0dc19d9d08227a233423cac4032f55ed6c550
```

That evidence covered static/proxy/keepalive, header-heavy behavior, allocation measurements, CPU sampling, and documented the local environment's TLS/syscall-tool limitations.

Crucially, the A/B did its job: it exposed a **~20% regression in the small buffered proxy path** because the intermediate implementation replaced an already-optimal one-write small response with a two-iovec `writev`.

The final PR commit `e302ca9` fixed that regression structurally:

- if proxy head + body fit in the existing 8 KiB scratch buffer, retain the original coalesced **single `writeAll()`** fast path;
[...]

PR #598, opened 2026-08-10
perf: add writev response writer

Jaruso:
## Summary
- add a gathered plaintext response write path for HTTP/1 responses via a raw-socket writev helper
- keep TLS and generic writers on buffered/fallback writes, with documented response-write metrics
- wire tracked response writes through gateway/static/proxy response paths and add response writer tests

Refs #144

## Validation
- zig build -Doptimize=ReleaseFast --summary all --error-style verbose
- zig build bench-allocations --summary all --error-style verbose
- ./benchmarks/ci-smoke.sh --duration 5 --connections 10 --threads 2

## Benchmark smoke notes
Local loopback smoke on Apple M4, 5s duration, 10 connections, 2 threads:
- static-http1: 97,796.78 req/s, p99 13.063 ms, errors 0
- proxy-http1: 23,922.26 req/s, p99 9.532 ms, errors 0
- keepalive: 85,529.06 req/s, p99 2.618 ms, errors 0
- upstream connection reuse: 122036/122046 (99%)

### DIFF
--- a/src/gateway_proxy_response.zig
+++ b/src/gateway_proxy_response.zig
@@ -237,6 +237,12 @@ fn writeBufferedUpstreamResponseMeasured(
             try writer.writeAll(head.bytes);
             return .{ .mode = .single_write };
         }
+        const total_len = head.bytes.len + upstream_response.body.len;
+        if (head.allocator == null and total_len <= scratch.len) {
+            @memcpy(scratch[head.bytes.len..total_len], upstream_response.body);
+            try writer.writeAll(scratch[0..total_len]);
+            return .{ .mode = .single_write };
+        }
         const fragments = [_][]const u8{ head.bytes, upstream_response.body };
         return .{
             .mode = .writev,
@@ -601,7 +607,7 @@ test "writeBufferedUpstreamResponse serializes a single forwarded response head"
     try std.testing.expect(std.mem.endsWith(u8, output, "\r\n\r\npong"));
 }
 
-test "writeBufferedUpstreamResponseWithMetrics uses gathered write for data-plane body" {
+test "writeBufferedUpstreamResponseWithMetrics keeps small data-plane body on single-write fast path" {
     const allocator = std.testing.allocator;
     const body = try allocator.dupe(u8, "pong");
     var upstream_headers = [_]TestUpstreamHeader{
@@ -633,11 +639,53 @@ test "writeBufferedUpstreamResponseWithMetrics uses gathered write for data-plan
         &metrics_mutex,
     );
 
+    try std.testing.expectEqual(@as(usize, 0), writer.writev_calls);
+    try std.testing.expectEqual(@as(usize, 0), writer.writev_iovecs);
+    try std.testing.expectEqual(@as(usize, 1), writer.write_all_calls);
+    try std.testing.expect(std.mem.startsWith(u8, writer.output.items, "HTTP/1.1 200 OK\r\n"));
+    try std.testing.expect(std.mem.endsWith(u8, writer.output.items, "pong"));
+    try std.testing.expectEqual(@as(u64, 1), metrics.response_write_mode_total[1]);
+    try std.testing.expectEqual(@as(u64, 0), metrics.response_writev_iovecs_total);
+}
+
+test "writeBufferedUpstreamResponseWithMetrics uses gathered write when body does not fit scratch" {
+    const allocator = std.testing.allocator;
+    const body = try allocator.alloc(u8, 10 * 1024);
+    @memset(body, 'x');
+    var upstream_headers = [_]TestUpstreamHeader{
+        .{ .name = "Content-Type", .value = "text/plain" },
+    };
+    var response = TestBufferedUpstreamResponse{
+        .metadata_arena = std.heap.ArenaAllocator.init(allocator),
+        .status_code = 200,
+        .reason = "OK",
+        .headers = upstream_headers[0..],
+        .body = body,
+    };
+    defer response.deinit(allocator);
+
+    var writer = TestGatheredProxyWriter.init(allocator);
+    defer writer.deinit();
+    var metrics = http.metrics.Metrics.init();
+    var metrics_mutex = compat.Mutex{};
+
+    try writeBufferedUpstreamResponseWithMetrics(
+        &writer,
+        &response,
+        true,
+        "req-gathered-large",
+        &http.security_headers.SecurityHeaders.api,
+        null,
+        null,
+        &metrics,
+        &metrics_mutex,
+    );
+
     try std.testing.expectEqual(@as(usize, 1), writer.writev_calls);
     try std.testing.expectEqual(@as(usize, 2), writer.writev_iovecs);
     try std.testing.expectEqual(@as(usize, 0), writer.write_all_calls);
     try std.testing.expect(std.mem.startsWith(u8, writer.output.items, "HTTP/1.1 200 OK\r\n"));
-    try std.testing.expect(std.mem.endsWith(u8, writer.output.items, "pong"));
+    try std.testing.expect(std.mem.endsWith(u8, writer.output.items, body));
     try std.testing.expectEqual(@as(u64, 1), metrics.response_write_mode_total[0]);
     try std.testing.expectEqual(@as(u64, 2), metrics.response_writev_iovecs_total);
 }
--- a/src/gateway_proxy_response.zig
+++ b/src/gateway_proxy_response.zig
@@ -298,7 +298,7 @@ pub fn sendApiError(allocator: std.mem.Allocator, writer: anytype, status: http.
         gph.setRequestIdHeaders(&response, rid);
     }
     applyResponseHeaders(state, &response);
-    try response.write(writer);
+    try response.writeWithMetrics(writer, &state.metrics, &state.metrics_mutex);
     state.metricsRecord(@intFromEnum(status));
     state.metricsRecordErrorCode(code);
 }
--- a/src/gateway_proxy_response.zig
+++ b/src/gateway_proxy_response.zig
@@ -180,6 +180,130 @@ pub fn writeBufferedUpstreamResponse(
     try writer.writeAll(response_stream.getWritten());
 }
 
+pub fn writeBufferedUpstreamResponseWithMetrics(
+    writer: anytype,
+    upstream_response: anytype,
+    keep_alive: bool,
+    correlation_id: []const u8,
+    security: *const http.security_headers.SecurityHeaders,
+    alt_svc: ?[]const u8,
+    sticky_set_cookie: ?[]const u8,
+    metrics: *http.metrics.Metrics,
+    metrics_mutex: *compat.Mutex,
+) !void {
+    const result = writeBufferedUpstreamResponseMeasured(
+        writer,
+        upstream_response,
+        keep_alive,
+        correlation_id,
+        security,
+        alt_svc,
+        sticky_set_cookie,
+    ) catch |err| {
+        recordResponseWriteMetric(metrics, metrics_mutex, attemptedWriterMode(@TypeOf(writer), upstream_response.body.len == 0), 0, true);
+        return err;
+    };
+    recordResponseWriteMetric(metrics, metrics_mutex, result.mode, result.iovecs, false);
+}
+
+const BufferedWriteMeasurement = struct {
+    mode: http.metrics.ResponseWriteMode,
+    iovecs: usize = 0,
+};
+
+fn writeBufferedUpstreamResponseMeasured(
+    writer: anytype,
+    upstream_response: anytype,
+    keep_alive: bool,
+    correlation_id: []const u8,
+    security: *const http.security_headers.SecurityHeaders,
+    alt_svc: ?[]const u8,
+    sticky_set_cookie: ?[]const u8,
+) !BufferedWriteMeasurement {
+    const Writer = @TypeOf(writer);
+    if (comptime writerSupportsGatheredWrite(Writer)) {
+        var scratch: [8192]u8 = undefined;
+        const head = try buildBufferedUpstreamResponseHeadBounded(
+            &scratch,
+            upstream_response,
+            keep_alive,
+            correlation_id,
+            security,
+            alt_svc,
+            sticky_set_cookie,
+        );
+        defer head.deinit();
+        if (upstream_response.body.len == 0) {
+            try writer.writeAll(head.bytes);
+            return .{ .mode = .single_write };
+        }
+        const fragments = [_][]const u8{ head.bytes, upstream_response.body };
+        return .{
+            .mode = .writev,
+            .iovecs = try writer.writeGatheredAll(&fragments),
+        };
+    }
+
+    try writeBufferedUpstreamResponse(
+        writer,
+        upstream_response,
+        keep_alive,
+        correlation_id,
+        security,
+        alt_svc,
+        sticky_set_cookie,
+    );
+    return .{ .mode = classifyWriterMode(Writer, upstream_response.body.len == 0) };
+}
+
+const HeaderBytes = struct {
+    allocator: ?std.mem.Allocator = null,
+    bytes: []u8,
+
+    fn deinit(self: HeaderBytes) void {
+        if (self.allocator) |allocator| allocator.free(self.bytes);
+    }
+};
+
+fn buildBufferedUpstreamResponseHeadBounded(
+    scratch: []u8,
+    upstream_response: anytype,
+    keep_alive: bool,
+    correlation_id: []const u8,
+    security: *const http.security_headers.SecurityHeaders,
+    alt_svc: ?[]const u8,
+    sticky_set_cookie: ?[]const u8,
+) !HeaderBytes {
+    var stream = compat.fixedBufferStream(scratch);
+    writeBufferedUpstreamResponseHead(
+        stream.writer(),
+        upstream_response,
+        keep_alive,
+        correlation_id,
+        security,
+        alt_svc,
+        sticky_set_cookie,
+    ) catch {
+        const fallback_allocator = std.heap.page_allocator;
+        var allocating: std.Io.Writer.Allocating = .init(fallback_allocator);
+        errdefer allocating.deinit();
+        try writeBufferedUpstreamResponseHead(
+            &allocating.writer,
+            upstream_response,
+            keep_alive,
+            correlation_id,
+            security,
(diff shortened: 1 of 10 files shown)


src/http/response.zig
```diff
@@ -196,38 +173,68 @@ pub const Response = struct {
...
+    const WriteMeasurement = struct {
+        mode: metrics_mod.ResponseWriteMode,
+        iovecs: usize = 0,
+    };
+
+    fn writeMeasured(self: *Response, writer: anytype) !WriteMeasurement {
+        const body_bytes = self.body orelse "";
+        const Writer = @TypeOf(writer);
+        if (comptime writerSupportsGatheredWrite(Writer)) {
+            const header_bytes = try self.buildHeaderBytes(self.allocator);
```
Jaruso: **[P1] Avoid adding heap allocations/copies to every gathered response.** `buildHeaderBytes()` uses `std.Io.Writer.Allocating` and then `allocator.dupe(...)`, so the new hot plaintext path allocates a backing buffer and then allocates/copies the finished header bytes again before `writev`. That works functionally, but it directly cuts against #144's copy/allocation goal and is not covered by the PR's before/after evidence. Please make the common header serialization allocation-free (for example a bounded `fixedBufferStream` scratch buffer with an oversized-header fallback), then submit `{ header_bytes, body_bytes }` as the two iovecs.

src/http/response.zig
```diff
@@ -196,38 +173,68 @@ pub const Response = struct {
...
+        self: *Response,
+        writer: anytype,
+        metrics: *metrics_mod.Metrics,
+        metrics_mutex: *compat.Mutex,
+    ) !void {
+        self.writeHead(writer) catch |err| {
+            recordResponseWriteMetric(metrics, metrics_mutex, .single_write, 0, true);
+            return err;
+        };
+        recordResponseWriteMetric(metrics, metrics_mutex, .single_write, 0, false);
```
Jaruso: **[P2] `single_write` is not the operation actually performed here.** `writeHead()` delegates to `writeHeaders()`, which issues multiple `print`/`writeAll` calls. On a raw socket that is multiple writes/syscalls, yet this records one `single_write` response and the observability docs say header-only writes use that mode. Please either serialize the head once and do one `writeAll`, or classify/describe this mode accurately. A writer-call-count test for HEAD would catch this.

src/http/response.zig
```diff
@@ -196,38 +173,68 @@ pub const Response = struct {
...
+        if (comptime writerSupportsGatheredWrite(Writer)) {
+            const header_bytes = try self.buildHeaderBytes(self.allocator);
+            defer self.allocator.free(header_bytes);
+            if (body_bytes.len == 0) {
+                try writer.writeAll(header_bytes);
+                return .{ .mode = .single_write };
+            }
+            const fragments = [_][]const u8{ header_bytes, body_bytes };
+            try writer.writeGatheredAll(&fragments);
+            return .{ .mode = .writev, .iovecs = fragments.len };
```
Jaruso: **[P2] This undercounts `tardigrade_response_writev_iovecs_total` on partial writes.** `fragments.len` is the logical vector size (normally 2), but `writevAllFd()` can submit multiple vectors when the kernel returns a short write. The metric HELP text says it counts total iovecs *submitted*, so a header-partial retry could submit 2 + 2 iovecs while this records only 2. Please count the actual `iov_len` values passed to `writev`, or change the metric semantics/name/help text to explicitly mean logical response fragments.

src/gateway_static_runtime.zig
```diff
@@ -251,7 +251,7 @@ pub fn writeStaticServedResponse(
     }
 
     gp.applyResponseHeaders(state, &response);
-    try response.writeHead(writer);
+    try response.writeHeadWithMetrics(writer, &state.metrics, &state.metrics_mutex);
```
Jaruso, line 266: **[P1] The buffered static path still bypasses the gathered writer.** This unconditionally writes the head first; when `served.file_path == null`, the body is then written separately with `writer.writeAll(body)` below. That means the static-file fallback specifically called out in #144 never gets `writev`. Branch here so non-HEAD buffered responses use `response.writeWithMetrics(...)` and return; keep `writeHeadWithMetrics(...)` only for HEAD and file-backed `sendfile` responses.

src/gateway_proxy_runtime.zig
```diff
@@ -436,7 +436,7 @@ fn executeVersionedApiProxyRoute(
                 _ = response.setHeader("Set-Cookie", cookie);
             }
             applyResponseHeaders(state, &response);
-            try response.write(writer);
+            try response.writeWithMetrics(writer, &state.metrics, &state.metrics_mutex);
```
Jaruso, line 441: **[P1] This covers the versioned/control-plane buffered response, but not the common data-plane proxy path required by #144.** `DataPlaneProxyResponse.writeHttp1()` still routes through `writeBufferedUpstreamResponse()`, which builds an 8 KiB combined head/body buffer and calls `writeAll()`. Please route that general buffered `proxy_pass` response through the gathered abstraction too (serialized head + body iovecs for plaintext, TLS/fallback otherwise), and record the same response-write metrics there.

src/zig_compat.zig
```diff
@@ -184,6 +192,64 @@ pub const NetStream = struct {
...
+            }
+            iovecs[iov_len] = .{
+                .base = data[i].ptr + first_offset,
+                .len = data[i].len - first_offset,
+            };
+            iov_len += 1;
+            first_offset = 0;
+        }
+
+        const written = try writevFd(fd, iovecs[0..iov_len]);
```
Jaruso: **[P1] Please directly test short `writev` results before merging this correctness-critical loop.** The new `TestVectoredWriter` always consumes all fragments, so none of this index/offset advancement is exercised. The issue itself identifies partial-write handling as the main risk. Please add deterministic cases for a short write inside the first iovec, exactly at the header boundary, and inside the body, plus EINTR/zero/error behavior. Factoring the cursor advancement or syscall call behind a test hook/helper is fine.

Jaruso (commented):
**Blocking review — GitHub does not allow this account to submit `REQUEST_CHANGES` on its own PR, so this is filed as a COMMENT review instead. I would not merge #598 yet.**

The core `writev` primitive is promising, but PR #598 does not yet satisfy several explicit acceptance criteria from #144.

What is already good:
- Raw plaintext `NetStream.Writer` now has a gathered-write path with retry/offset logic.
- TLS/generic writers retain a non-writev fallback and the intended modes are documented.
- Basic response tests cover small/large/empty/error cases, and existing HEAD coverage remains.
- Current CI is green on Linux/macOS, x86_64/ARM, including the perf-smoke job.

Blocking gaps:

1. **The common static buffered response path still does `head` + separate body writes.** `writeStaticServedResponse` always calls `writeHeadWithMetrics()`, then calls `writer.writeAll(body)` for the buffered path. That means one of the explicitly requested static fallback paths never uses the new gathered writer.

   Expected shape:
   ```zig
   if (head_only) {
       try response.writeHeadWithMetrics(writer, &state.metrics, &state.metrics_mutex);
       return status;
   }

   if (served.file_path == null) {
       try response.writeWithMetrics(writer, &state.metrics, &state.metrics_mutex);
       return status;
   }

   // File-backed path: write head, then sendfile.
   try response.writeHeadWithMetrics(writer, &state.metrics, &state.metrics_mutex);
   try sendFileFd(...);
   ```

2. **The general data-plane proxy response path is still unchanged.** `DataPlaneProxyResponse.writeHttp1()` still calls `writeBufferedUpstreamResponse()`, which coalesces the response head and as much of the body as possible into an 8 KiB buffer before a normal `writeAll()`. The new `Response.writeWithMetrics()` call in this PR covers the versioned/control-plane response branch, but not the common buffered `proxy_pass` path named in #144. Please convert `writeBufferedUpstreamResponse()` (or its abstraction) to gather a serialized head plus body on supported plaintext writers, while retaining TLS/fallback behavior.

3. **The new writev hot path introduces per-response heap allocation/copy overhead that is not accounted for.** `writeMeasured()` calls `buildHeaderBytes()`, which uses an allocating writer and then duplicates the completed header bytes. For the common case this adds heap allocation(s) plus an additional header copy before the syscall optimization. A bounded fixed scratch buffer (with a safe oversized-header fallback) would avoid this on the common path, e.g. serialize headers into a `fixedBufferStream`, then submit `{ headers, body }` as the two iovecs.

4. **Partial-write correctness needs direct tests.** `writevAllFd()` contains the highest-risk logic in this ticket, but `TestVectoredWriter` always consumes the complete vector. Existing partial-write tests exercise the older response state machine, not this new offset-advancement loop. Please add deterministic coverage for at least: short write inside the header, exact header boundary, short write in the body, EINTR retry, and zero/error handling. Factoring the cursor advancement or syscall hook into a testable helper is fine.

5. **The new metrics are not fully truthful under the current implementation.** `writeHeadWithMetrics()` records `single_write`, but it delegates to `writeHeaders()`, which performs multiple `print`/`writeAll` calls. Also `response_writev_iovecs_total` currently records the logical fragment count once (normally 2), not the actual number of iovecs submitted if `writevAllFd()` must retry after partial writes. Either change the metric semantics/help text or instrument the actual submissions.

6. **The benchmark acceptance criteria are not demonstrated.** The PR body provides absolute smoke numbers only. #144 explicitly asks for before/after copy/allocation/syscall impact and a matrix including tiny static, small proxy/dynamic, header-heavy, and TLS/plaintext, comparing req/s, p99, CPU/request, allocations/request, and syscalls/request. This is especially important here because the implementation changes allocation behavior. Please attach the relevant before/after measurements (and Beelink results if available, per the issue comment) before merge.

Once the static + common proxy paths actually use the abstraction, the allocation strategy is addressed/measured, partial writes are directly tested, and the required benchmark comparison is documented, this should be ready for another pass.

Jaruso (commented):
Re-review of `f4f0dc1` after the first round of feedback.

All six code-level findings from my previous review are addressed, and I resolved those six threads:

- the common gathered response path now serializes headers into a bounded 8 KiB stack scratch buffer instead of allocating on every response;
- gathered-capable HEAD/empty-body responses now actually perform one `writeAll`, so `single_write` is truthful;
- `writevAllWith` returns the actual cumulative iovec submissions, including retry submissions;
- the buffered static path now uses `Response.writeWithMetrics(...)` for non-HEAD/non-sendfile responses;
- the common buffered data-plane `proxy_pass` path now uses a gathered head + body write with metrics;
- short writes inside the first iovec, exactly at the boundary, inside the body, EINTR, zero-write, and error behavior are now directly tested.

I also re-read the new `writevAllWith` cursor/retry implementation and the new proxy/static wiring and did not find another correctness defect in this delta.

**One acceptance-criteria blocker remains before I would merge #598: the benchmark evidence required by #144 still is not present.** The PR body still contains only absolute after-change M4 smoke numbers. Issue #144 explicitly asks for before/after copy/allocation/syscall impact and the benchmark matrix covering tiny static responses, small dynamic/proxy responses, header-heavy responses, and TLS/plaintext variants, comparing req/s, p99, CPU/request, allocations/request, and syscalls/request. The issue discussion also specifically asks to re-measure on the Beelink before drawing the performance conclusion if that machine is available.

Please attach a concise before/after table (or link to reproducible artifacts) for at least:

```text
scenario                 baseline     writev
static plaintext req/s   ...          ...
static plaintext p99     ...          ...
proxy plaintext req/s    ...          ...
proxy plaintext p99      ...          ...
header-heavy             ...          ...
TLS                      ...          ...
allocations/request      ...          ...
syscalls/request         ...          ...
CPU/request              ...          ...
```

If a metric is not feasible with the existing harness, document that explicitly and provide the closest measurable proxy rather than silently omitting it.

Current CI for `f4f0dc1` is still running, but the main unit tests on Linux/macOS/ARM and the perf-smoke job have already passed at the time of this review. Assuming the remaining CI finishes green, I do not see another code change required beyond satisfying/documenting the benchmark acceptance criterion.

Jaruso (commented):
Third pass on #598. There is no new commit or new benchmark evidence since my previous review, so the code verdict is unchanged: I still do not see another implementation defect in `f4f0dc1`, and all six prior code threads remain correctly resolved.

CI has now completed fully green for this head commit, including the perf-smoke job. I also inspected the completed perf-smoke output to make sure it did not implicitly satisfy the remaining acceptance criterion. It reports only the post-change run (for example, static HTTP/1 ~107,933 req/s / p99 0.361 ms / avg CPU 237.98%, proxy HTTP/1 ~9,480 req/s / p99 2.245 ms / avg CPU 50.16%). It does not run or report the baseline commit, header-heavy or TLS variants, allocations/request, syscalls/request, or a before/after CPU/request comparison.

So the sole blocker from the previous review still stands: issue #144 explicitly requires before/after benchmark evidence for the writev change. The existing issue comment about the earlier single-write prototype is useful historical context, but it is not an A/B measurement of this implementation.

Once that benchmark comparison is attached (or the issue acceptance criterion is explicitly amended), I consider the code itself ready to merge. No new inline findings on this pass.

src/gateway_proxy_response.zig
```diff
@@ -180,6 +180,130 @@ pub fn writeBufferedUpstreamResponse(
...
+        );
+        defer head.deinit();
+        if (upstream_response.body.len == 0) {
+            try writer.writeAll(head.bytes);
+            return .{ .mode = .single_write };
+        }
+        const fragments = [_][]const u8{ head.bytes, upstream_response.body };
+        return .{
+            .mode = .writev,
+            .iovecs = try writer.writeGatheredAll(&fragments),
```
Jaruso, line 249: **[P1] The posted A/B benchmark shows this unconditional gathered write regresses the common `proxy-http1` path by 20.3%.** Throughput falls from 22,465.85 to 17,911.16 req/s with essentially identical p99. The CPU numbers are also not neutral once normalized: `0.818 / 22465.85 ≈ 36.4 µs CPU/request` before vs `0.7853 / 17911.16 ≈ 43.8 µs` after, about **+20.4% CPU/request**.

That is consistent with the base implementation: for small proxy responses, `writeBufferedUpstreamResponse()` already packed head + body into the 8 KiB stack buffer and performed one `writeAll()`. This changes a one-syscall, allocation-free fast path into a `writev` call without removing any syscall or allocation.

Please keep the coalesced single-write path when the complete response fits in the bounded scratch buffer, and use gathered I/O when the body does not fit (or otherwise when it actually avoids an additional write). For example, conceptually:

```zig
if (head.allocator == null and head.bytes.len + upstream_response.body.len <= scratch.len) {
    const total_len = head.bytes.len + upstream_response.body.len;
    @memcpy(scratch[head.bytes.len..total_len], upstream_response.body);
    try writer.writeAll(scratch[0..total_len]);
    return .{ .mode = .single_write };
}

const fragments = [_][]const u8{ head.bytes, upstream_response.body };
return .{ .mode = .writev, .iovecs = try writer.writeGatheredAll(&fragments) };
```

If you want to keep writev even for tiny proxy responses, then please repeat the proxy A/B enough times to show the current 20% delta is noise before merging. A performance change should not knowingly make the primary proxy path ~20% more CPU-expensive per request.

Jaruso (commented):
Fourth pass after the requested benchmark evidence was posted.

The benchmark documentation gap from my previous reviews is now addressed. The post compares the PR base (`d3935326`) to the current head (`f4f0dc1`) on the same M4 host, includes static/proxy/keepalive and header-heavy scenarios, allocation counts, CPU sampling, and explicitly documents why TLS and syscall/request could not be measured in that local environment. CI is also fully green for this head.

However, the new A/B data exposes a blocker that was not visible before the benchmarks existed: **the common buffered `proxy-http1` path regresses materially.** Throughput drops from 22,465.85 req/s to 17,911.16 req/s (-20.3%) while p99 is unchanged. More importantly, the posted CPU percentages are not actually neutral when normalized by work done: baseline is roughly `0.818 / 22465.85 = 36.4 µs CPU/request`, while the PR is `0.7853 / 17911.16 = 43.8 µs CPU/request`, or about **+20.4% CPU/request**.

That lines up with the implementation change: the baseline proxy path already serialized the head and small body into an 8 KiB stack buffer and issued one `writeAll()`. For a tiny `/proxy/health` response, replacing that one-write path with a two-iovec `writev` does not eliminate a syscall or heap allocation; it only adds vectored-syscall setup. The benchmark result therefore looks plausible rather than obviously noisy.

I left one inline P1 on the `writeGatheredAll` call. Please either preserve the existing single-write fast path when head + body fit in the bounded scratch buffer and use writev when gathering actually avoids an extra write/copy, or run enough repeated A/B proxy trials to demonstrate that the observed ~20% loss is measurement noise. Given this is a performance PR and `proxy_pass` is a primary data-plane path, I would not merge a reproducible 20% proxy regression even though the other scenarios improve substantially.

GitHub does not allow this account to request changes on its own PR, so this is filed as a blocking COMMENT review.

Jaruso (commented):
Fifth pass on `e302ca9`, which contains one targeted fix in `src/gateway_proxy_response.zig` for the proxy regression found in the posted A/B data.

The code change addresses the P1 correctly, and I resolved that thread:

- when the serialized proxy head and body fit in the existing 8 KiB scratch buffer, the code now copies the body into that buffer and performs one `writeAll()`, preserving the baseline small-response fast path;
- when the response does not fit the scratch buffer, it still uses the new `{ head, body }` gathered write, so the ticket's writev optimization remains active where it actually avoids the old extra write/copy;
- tests now assert both branches and their metrics (`single_write` for the small response, `writev`/2 iovecs for the large response).

I re-read the boundary/ownership logic around `head.allocator`, `total_len`, the scratch slice, and the large-body fallback and do not see a correctness defect in this delta. CI for `e302ca9` has also completed fully green across the matrix, including unit tests and perf smoke.

**One final performance-validation item remains before I would merge this performance PR:** the benchmark comment currently compares baseline `d3935326` against the previous head `f4f0dc1`. The current head is `e302ca9`, and this last commit exists specifically to change the performance result that previously showed a 20.3% proxy regression. The shared-runner perf smoke for `e302ca9` passes, but it is not an A/B comparison and ran on different hardware from the previous CI smoke, so it cannot demonstrate that the regression is gone.

Please rerun the same-host targeted A/B for at least `proxy-http1`, comparing `d3935326` to `e302ca9`, and update the benchmark comment. Re-running the full matrix is not necessary because this commit only changes the buffered proxy response branch. Ideally also include one body larger than 8 KiB so the final evidence covers both the restored single-write fast path and the retained writev path.

If the small proxy result is baseline-neutral or better and the >8 KiB proxy path does not regress materially, I consider #598 merge-ready. No additional code changes requested on this pass.

PR #599, opened 2026-08-10
perf(proxy): account and bound request-direction buffers, surface HTTP/1 stalls (#140)

Jaruso:
Slice 3 of #140 — the issue's **PR 3: request direction and HTTP/1 observability**. Builds on #595 (HTTP/2 response enforcement).

## Summary

**Request-direction aggregate accounting.** Upload relay buffers were accounted per stream but never charged to the per-origin or global aggregates, so `TARDIGRADE_PROXY_BUFFER_GLOBAL_HARD_LIMIT_BYTES` bounded responses while concurrent uploads could multiply past it. Both upload relays now reserve against those scopes (HTTP/2 uploads against origin + global; HTTP/1 against global — per-origin accounting for h1 origins is the issue's PR 4, and an h1 origin's relay memory is already bounded per request by fixed buffers). The HTTP/1 response relay buffer clears the global scope too, so the enforcing gauge is no longer narrower than the paths it governs.

**Reserve once, not per frame.** An HTTP/2 upload reserved and released `n` bytes per DATA frame. That tracked nothing retained and gave the aggregate scopes a flicker no concurrent stream could ever have been bounded by. It now reserves its relay buffer once for the upload's life, matching the HTTP/1 model, plus a second reservation for body bytes that arrived with the request head (a genuinely separate buffer).

**Deterministic pre-commitment failure semantics.** Capacity refusals during an upload now split by what ran out:

| Refused at | Status | Meaning |
| --- | --- | --- |
| per-stream hard limit | `413 payload_too_large` | *this* upload's in-flight bytes exceed its allowed bound |
| per-origin / global hard limit | `503 proxy_buffer_saturated` | the proxy is out of room for anybody |

Both are raised before the response head is committed, and neither is recorded against upstream health. Previously an HTTP/1 upload refusal returned the raw error and surfaced as a `502` charged to a healthy origin.

**HTTP/1 stall observability.** `tardigrade_buffer_read_pauses_total{side="downstream"}` and its resume counterpart could only ever read zero. The HTTP/1 upload relay deliberately has no queue whose depth could cross a watermark — it reads the client again only once the upstream write goes through — so a blocked upstream write *is* its backpressure. A zero-timeout writability `poll` before each chunk turns that into the pause/resume series without adding a queue, per the issue's explicit constraint. A healthy origin pays one non-blocking `poll` per chunk and moves no counters.

Only transitions are recorded, and a relay torn down mid-stall still reports its resume, so the difference between the two counters reads as "how many uploads are stalled right now" rather than drifting by one per abort. A write that begins blocking only *after* the check passes is deliberately not counted — an origin that has genuinely stopped consuming keeps the buffer full across iterations and is caught, while a momentary block is not a backpressure event.

## Test plan

`zig build test` — 3354/3355 pass (1 skipped: the known UDP smoke test under this sandbox). Run 5× to confirm the timing-sensitive test is stable. `zig build test-failure` fails identically on `main` (`error.SocketUnconnected`, a sandbox socket restriction) — unrelated to this change.

New tests:

- [x] **Slow origin → downstream read pause.** The upstream send buffer is filled to `EAGAIN` before the relay starts, so the stall is a fact of the socket's state at the moment the relay looks rather than a race against a background reader (the first draft was flaky for exactly that reason). Asserts a pause fired, pauses == resumes, and peak reserved is the fixed 16 KiB buffer — not the 64 KiB body.
- [x] **Client abort mid-upload** (HTTP/1 and HTTP/2) — every gauge and the global aggregate return to zero; the pause counters stay untouched because nothing stalled.
- [x] **Per-stream hard cap → `RequestBufferLimitExceeded`**, charged to `scope="stream"` and not to an aggregate scope.
- [x] **Aggregate exhausted by a concurrent relay → `ProxyBufferCapacityUnavailable`**, labelled `scope="global"`; the refused relay retains nothing and the reservation that did fit is intact.
- [x] **HTTP/2 upload reserves once** — 16 DATA frames go out, one buffer is ever owned.
- [x] `ReadStall` transition unit tests: no counters when nothing stalls, one pair per stall rather than one per write, balanced at teardown.

Docs (`PROXY_STREAMING.md`, `OBSERVABILITY.md`) and `CHANGELOG.md` updated, including the stale claims that the aggregate limit governs HTTP/2 response queues only and that the `side="downstream"` series is reserved at zero.

## Left to PR 4

Per-origin buffer accounting for HTTP/1 origins, and the tuning-guidance documentation — both explicitly the issue's PR 4 scope. Noted inline at the two sites that pass a global-only capacity.

Refs #140

### DIFF
--- a/docs/OBSERVABILITY.md
+++ b/docs/OBSERVABILITY.md
@@ -83,10 +83,14 @@ logs are written through `src/http/logger.zig`.
   difference between pauses and resumes reads as "how many relays are stalled
   right now".
 - per-origin HTTP/2 buffer accounting:
-  `tardigrade_upstream_h2_pool_buffered_bytes{upstream}` and
-  `tardigrade_upstream_h2_pool_buffer_limit_exceeded_total{upstream}`. Label
-  cardinality is bounded by the number of distinct configured origins, matching
-  the other `tardigrade_upstream_h2_pool_*` series.
+  `tardigrade_upstream_h2_pool_buffered_bytes{upstream,direction}` and
+  `tardigrade_upstream_h2_pool_buffer_limit_exceeded_total{upstream,direction}`.
+  Response queues and request-direction upload relay buffers are enforced
+  against the *same* per-origin account, so both directions are reported: a
+  request-only pressure would otherwise read as zero while the limit was
+  actively bounding it. Label cardinality stays bounded — two fixed directions
+  per distinct configured origin — and `direction` takes the same values as the
+  process-wide buffer family.
 - configured proxy buffer limits:
   `tardigrade_buffer_config_limit_bytes{direction,scope,limit}` for the
   per-stream low/high/hard watermarks plus per-origin/global hard-limit
--- a/docs/OBSERVABILITY.md
+++ b/docs/OBSERVABILITY.md
@@ -52,19 +52,23 @@ logs are written through `src/http/logger.zig`.
   `tardigrade_buffer_limit_exceeded_total{direction,scope}`. Labels are fixed
   to protocol-independent directions, scopes, and sides; they never include
   URLs, request IDs, or stream IDs. Current byte gauges cover bounded buffered
-  responses, HTTP/1 streaming relay buffers, and HTTP/2 streaming response
-  queues. `scope="stream"` reports logical queue occupancy; `scope="global"`
+  responses, HTTP/1 streaming relay buffers in both directions, HTTP/2 streaming
+  response queues, and request-direction upload relay buffers.
+  `scope="stream"` reports logical queue occupancy; `scope="global"`
   reports retained allocation, and the two differ while a queue is partially
   drained but still owns its buffer. The `scope="global"` series is a roll-up
   across *all* proxy paths and is **not** the quantity the configured global
   hard limit is checked against — see the next entry.
 - the aggregate bytes the global hard limit is enforced against:
   `tardigrade_proxy_buffer_aggregate_bytes_current{direction,scope="global"}`,
-  read directly from the account that performs the enforcement. Today that
-  covers HTTP/2 streaming response queues only, so this is the series to compare
-  with `tardigrade_buffer_config_limit_bytes{scope="global",limit="hard"}`;
-  comparing the broader `tardigrade_buffered_bytes_current{scope="global"}`
-  roll-up with that limit overstates pressure under mixed traffic.
+  read directly from the account that performs the enforcement. That covers
+  HTTP/2 stream queues, HTTP/1 relay buffers in both directions, and
+  request-direction upload buffers on both protocols, so this is the series to
+  compare with
+  `tardigrade_buffer_config_limit_bytes{scope="global",limit="hard"}`; comparing
+  the broader `tardigrade_buffered_bytes_current{scope="global"}` roll-up with
+  that limit overstates pressure under mixed traffic, because the roll-up also
+  includes the bounded buffered compatibility path, which has its own cap.
   `tardigrade_buffer_limit_exceeded_total` carries `scope="origin"` and
   `scope="global"` alongside `scope="stream"`: an aggregate refusal means the
   upstream origin (or the process) had no room for bytes the stream itself could
@@ -72,8 +76,12 @@ logs are written through `src/http/logger.zig`.
   resume counterpart record HTTP/2 streaming response queues crossing their high
   and low watermarks — a pause means stream credit is being withheld from the
   origin, a resume means the coalesced `WINDOW_UPDATE` went out. The
-  `side="downstream"` series remain reserved at zero until the HTTP/1 relay
-  gains pause/resume transition sites.
+  `side="downstream"` series record the HTTP/1 upload relay finding the origin's
+  send buffer full: that path has no queue to cross a watermark, so a blocked
+  upstream write *is* its backpressure, and it reads the client again only once
+  the write goes through. Both sides count transitions, not writes, so the
+  difference between pauses and resumes reads as "how many relays are stalled
+  right now".
 - per-origin HTTP/2 buffer accounting:
   `tardigrade_upstream_h2_pool_buffered_bytes{upstream}` and
   `tardigrade_upstream_h2_pool_buffer_limit_exceeded_total{upstream}`. Label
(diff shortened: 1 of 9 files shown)


src/gateway_proxy.zig
```diff
@@ -727,24 +774,33 @@ fn relayStreamingUploadToHttp2(
...
     proxy_buffer_limits: proxy_buffer_account.Limits,
     proxy_buffer_observer: proxy_buffer_account.Observer,
+    proxy_buffer_capacity: proxy_buffer_account.AggregateCapacity,
 ) !void {
+    // The relay owns `read_buf` for the whole upload, so reserve it once rather
+    // than per chunk: the reservation then tracks memory actually retained, and
+    // the aggregate scopes see one stable claim instead of a reserve/release
+    // flicker per DATA frame that no concurrent stream could ever be bounded by.
+    var upload_reservation = ProxyBufferReservation.init(.downstream_to_upstream, proxy_buffer_limits, proxy_buffer_observer, proxy_buffer_capacity);
+    try reserveUploadBytes(&upload_reservation, read_buf.len);
```
Jaruso: **[P1] Account the request-head bytes for chunked uploads too.** This reserves `read_buf`, and the `.length` branch separately reserves its initial body bytes, but the `.chunked` branch passes `sb.initial_bytes` into `chunked_upload.Reader` without ever reserving them. Those raw framing/body bytes remain a second live buffer while `read_buf` is owned, so concurrent chunked uploads can retain memory outside the stream/origin/global limits. Please reserve `sb.initial_bytes.len` for the chunked case (raw retained bytes, not decoded payload), then release that portion when the reader no longer owns it. A focused test should populate `initial_bytes`, exhaust the aggregate with that second buffer, and prove all scopes return to zero on success/error.

src/gateway_proxy.zig
```diff
@@ -1852,36 +1927,56 @@ fn sendStreamingProxyRequest(
...
     cancel_token: ?*const CancellationToken,
     proxy_buffer_limits: proxy_buffer_account.Limits,
     proxy_buffer_observer: proxy_buffer_account.Observer,
+    proxy_buffer_capacity: proxy_buffer_account.AggregateCapacity,
 ) !void {
     var relay: [16 * 1024]u8 = undefined;
-    var upload_reservation = ProxyBufferReservation.init(.downstream_to_upstream, proxy_buffer_limits, proxy_buffer_observer);
-    try upload_reservation.reserve(relay.len);
+    var upload_reservation = ProxyBufferReservation.init(.downstream_to_upstream, proxy_buffer_limits, proxy_buffer_observer, proxy_buffer_capacity);
+    try reserveUploadBytes(&upload_reservation, relay.len);
```
Jaruso: **[P1] The HTTP/1 chunked path has the same accounting hole.** `sb.initial_bytes` is retained by `chunked_upload.Reader` in addition to this 16 KiB relay buffer, but only `relay.len` is reserved. That means request-head remainder bytes from a chunked client are invisible to the stream/global accounting. Please reserve the raw initial slice on top of the relay buffer and release it when its ownership ends; add the H1 mirror test, including malformed/aborted chunked input so teardown proves the reservation cannot leak.

src/gateway_proxy.zig
```diff
@@ -1838,11 +1911,13 @@ fn sendStreamingProxyRequest(
     if (streaming_body) |sb| {
         try relayStreamingUploadToHttp1(
             transport,
+            fd,
             sb,
             downstream_conn,
             cancel_token,
             proxy_buffer_limits,
             proxy_buffer_observer,
+            proxy_buffer_capacity,
```
Jaruso: **[P1] Move this capacity decision before the upstream request head is written.** `sendStreamingProxyRequest()` currently executes `transport.writeAll(req_aw.written())` first and only then enters this relay, where the 16 KiB/initial-byte reservation is attempted. If the stream/global limit refuses the upload, Tardigrade returns a local 413/503 but the origin has already received an incomplete, potentially side-effecting request. #140's PR-3 sequence explicitly calls for deterministic capacity behavior *before request forwarding*. Please preflight the complete known upload footprint before the request-head write, `errdefer` it on send failure, and pass the live reservation into the relay rather than constructing it after forwarding.

src/gateway_proxy.zig
```diff
@@ -913,6 +969,7 @@ fn streamViaH2Pool(
                         cancel_token,
                         proxy_buffer_limits,
                         proxy_buffer_observer,
+                        proxy_buffer_capacity,
```
Jaruso: **[P1] HTTP/2 also reserves too late.** `conn.openStreaming()` explicitly sends the request HEADERS (and any `req.body` DATA) before returning; only afterward does this relay attempt the upload reservation. An origin therefore sees a request before a stream/origin/global capacity refusal is turned into 413/503. Please reserve the known upload footprint before `openStreaming()`, roll it back if opening fails, and transfer the reservation into `relayStreamingUploadToHttp2`. Add a test where capacity is already exhausted and assert the peer receives no request HEADERS.

src/gateway_proxy.zig
```diff
@@ -1363,6 +1420,20 @@ fn pollFdReadable(fd: std.posix.fd_t, timeout_ms: u32) !bool {
...
+/// buffer is full because the peer has stopped draining — which for the
+/// synchronous HTTP/1 relay is the whole of its backpressure, since it has no
+/// queue whose depth could cross a watermark instead. Never waits.
+fn pollFdWritable(fd: std.posix.fd_t) !bool {
+    var pfds = [_]std.posix.pollfd{.{
+        .fd = fd,
+        .events = std.posix.POLL.OUT,
+        .revents = 0,
+    }};
+    const ready = try std.posix.poll(&pfds, 0);
```
Jaruso: **[P2] `false` does not uniquely mean "send buffer full" here.** POSIX `poll()` can return `ready != 0` because `revents` contains `POLLERR`, `POLLHUP`, or `POLLNVAL` even when `POLLOUT` is absent. This function then returns `false`, `noteUpstreamWriteStall()` records a pause, the subsequent write fails, and the teardown emits a matching resume — making a broken upstream look like a slow-origin backpressure event. Please inspect error/hangup bits separately and treat only a genuine zero-ready writable check as a stall. For example, return an enum/optional result rather than a boolean, and add a closed-peer test asserting pause/resume remain unchanged.

docs/OBSERVABILITY.md
```diff
@@ -332,9 +340,9 @@ Two outcome shapes exist:
...
+| Proxy per-stream buffer hard limit | `TARDIGRADE_PROXY_BUFFER_PER_STREAM_HARD_LIMIT_BYTES` | 1 MiB | proxy buffer accounting | Hard-limit exceedance is observable through `tardigrade_buffer_limit_exceeded_total`. On the request direction a refusal here is *this* upload outgrowing its allowed in-flight bound → `413 payload_too_large`, distinct from an aggregate refusal's `503` |
+| Proxy per-origin buffer hard limit | `TARDIGRADE_PROXY_BUFFER_PER_ORIGIN_HARD_LIMIT_BYTES` | 0 (unlimited) | HTTP/2 stream queues (responses and uploads) | When non-zero, must be at least the per-stream hard limit. Caps the retained queue memory all streams to one origin hold together. Refused before commitment → `503 proxy_buffer_saturated`; refused after → that stream is reset and truncated. Never counted against upstream health. Reload applies immediately. HTTP/1 origins have no per-origin account; their relay memory is bounded per request by fixed buffers |
```
Jaruso, line 348: **[P2] This now documents request+response enforcement, but the per-origin H2 metrics still expose responses only.** `H2ConnPool.snapshotOrigins()` reads `buffer.currentBytes(.upstream_to_downstream)` and `limitExceededEvents(.upstream_to_downstream)` exclusively. With this PR, an upload can consume or refuse the same origin aggregate in `downstream_to_upstream` while `tardigrade_upstream_h2_pool_buffered_bytes{upstream=...}` / `_buffer_limit_exceeded_total` remain zero. Please extend the bounded-cardinality origin metrics to expose both directions (e.g. a fixed `direction` label or separate request/response fields) and add a request-direction origin-metric test before making this documentation claim.

Jaruso (commented):
## Review result: changes needed

I reviewed all 7 changed files against #140's PR-3 scope. CI is green, and the overall direction is good: relay-buffer lifetime reservations, 413-vs-503 mapping without charging upstream health, teardown cleanup, and the `ReadStall` transition object all fit the intended slice.

There are four blocking correctness/acceptance-criteria gaps:

1. **Chunked request-head bytes are not accounted.** `StreamingRequestBody.initial_bytes` can contain raw chunk framing/body bytes that were read with the request head. Both chunked relays keep that slice alive while also owning the relay buffer, but only the relay buffer is reserved. Those bytes therefore bypass stream/origin/global accounting.
2. **Capacity is checked after request forwarding.** HTTP/1 writes the upstream request head before entering the relay; HTTP/2 `openStreaming()` sends HEADERS before the relay reservation. A local 413/503 can therefore be returned only after the origin has already received an incomplete request, which misses #140's PR-3 requirement for deterministic capacity behavior before request forwarding.
3. **H2 per-origin observability is still response-only.** Uploads now consume/refuse the origin aggregate's `downstream_to_upstream` slot, but `H2ConnPool.snapshotOrigins()` exports only `upstream_to_downstream`, so the origin-level buffer gauge/refusal counter can report zero while request-direction pressure is actually enforcing the limit.
4. **`pollFdWritable` can misclassify socket failure as backpressure.** `poll()` may return ready for `POLLERR`/`POLLHUP`/`POLLNVAL` with no `POLLOUT`; the current boolean result turns that into a pause followed by a resume when the write fails.

Recommended implementation shape for the first two findings is to preflight one upload reservation before any upstream HEADERS/request head is written, then transfer that live reservation into the relay:

```zig
fn preflightUploadReservation(
    limits: proxy_buffer_account.Limits,
    observer: proxy_buffer_account.Observer,
    capacity: proxy_buffer_account.AggregateCapacity,
    relay_bytes: usize,
    initial_bytes: usize,
) !ProxyBufferReservation {
    var r = ProxyBufferReservation.init(
        .downstream_to_upstream,
        limits,
        observer,
        capacity,
    );
    errdefer r.releaseAll();

    try reserveUploadBytes(&r, relay_bytes);
    if (initial_bytes != 0) try reserveUploadBytes(&r, initial_bytes);
    return r;
}
```

For `Content-Length`, `initial_bytes` should be `min(sb.initial_bytes.len, content_length)`; for chunked framing it should be the raw `sb.initial_bytes.len`, because those are the bytes actually retained. Release the initial-byte portion when `Reader`/the direct initial write no longer owns it, and release the fixed relay reservation on relay teardown.

Please add focused tests for chunked initial-byte accounting/refusal and release, pre-forward rejection (origin sees no request head/HEADERS), a closed/error upstream socket not incrementing pause/resume, and request-direction H2 per-origin metrics.

Jaruso (commented):
Second-pass review against #140 PR-3 acceptance criteria and all prior findings.

The six findings from the first review are addressed: chunked request-head bytes are now accounted/released on both protocols, H1/H2 capacity is preflighted before forwarding, broken sockets no longer increment stall counters, and the H2 per-origin series now expose both directions with bounded cardinality. The new focused tests cover those fixes well, and the main unit-test jobs are green on this head.

One blocking issue remains from the preflight refactor: the H2 upload reservation now has the lifetime of the entire H2 exchange rather than the upload phase. After `relayStreamingUploadToHttp2()` succeeds, the fixed relay-buffer reservation remains charged as `downstream_to_upstream` while the code waits for and relays the response. A long response therefore consumes request-direction origin/global capacity even though no upload bytes are retained anymore, so otherwise-admissible concurrent uploads can be rejected with `503 proxy_buffer_saturated`; the direction metrics are also wrong for that interval. This conflicts with #140's ownership rule that accounting follows bytes only while that path owns them, and with the updated docs saying the upload reservation is released on completion.

Please end or transfer the request-direction reservation at the upload/response boundary. If the same `read_buf` remains owned as a response relay buffer, account that ownership in `upstream_to_downstream` rather than keeping the stale upload claim. Add a production-lifetime test where upload A completes but waits on a slow response, then upload B can reserve request-direction capacity that A no longer owns.

CI is still in progress overall at the time of this review, but Linux/macOS/ARM unit tests, format, builds, lint, security, perf smoke, and several other jobs are already green.

**Verdict: changes still needed; one blocker remains.**

src/gateway_proxy.zig
```diff
@@ -879,6 +980,25 @@ fn streamViaH2Pool(
                     try path_buf.writer.writeAll(uriComponentBytes(q));
                 }
 
+                // Claim the upload's buffer footprint before `openStreaming`,
+                // which sends HEADERS (#140). Reserving after it would let a
+                // local 413/503 be raised only once the origin had already
+                // received a request it will never see the body of.
+                var upload_reservation: ?ProxyBufferReservation = null;
+                defer if (upload_reservation) |*reservation| reservation.releaseAll();
```
Jaruso, line 988: **[P1] Release or transfer this reservation when the H2 upload finishes.** This `defer` now lives at the whole `.h2` exchange scope, so after `relayStreamingUploadToHttp2()` succeeds the base `read_buf.len` reservation stays charged to `downstream_to_upstream` through `waitStreamingResponseHead()` and the entire response relay. That is no longer request-direction ownership: a slow/large response can keep origin/global upload capacity occupied and cause unrelated uploads to get a false `503 proxy_buffer_saturated`, while the direction gauges continue reporting upload bytes that are no longer being uploaded. Before this refactor the relay-local defer ended at upload completion. Please release the request-direction claim immediately after the upload phase, or transfer the still-owned `read_buf` to an `upstream_to_downstream` reservation if it remains the response relay buffer. A regression test should complete upload A, hold its response open, and prove upload B can reserve the request-direction capacity A no longer owns.

Jaruso (commented):
## Re-review result: clean

Reviewed the new head (`fb947e13`) against the prior review findings and issue #140's PR 3 scope.

The remaining H2 reservation-lifetime blocker is fixed correctly: request-direction capacity is explicitly released immediately after the upload relay completes, before waiting on/relaying the response. The new regression test holds the first response open and proves a second upload can reserve the released request-direction capacity while the first exchange is still active, which directly covers the false-503 failure mode from the previous review.

I also re-checked the earlier fixes: request-head remainder accounting for chunked H1/H2 uploads, pre-forward capacity reservation on both protocols, H1 broken-socket vs backpressure classification, bidirectional H2 per-origin metrics, release/error paths, and the documented 413/503 semantics. I did not find a new correctness regression in the touched code.

CI for this head completed successfully.

**Verdict:** no blocking findings for this PR. The PR 3 request-direction/HTTP/1-observability scope from #140 is satisfied. The explicitly deferred HTTP/1 per-origin/tuning work remains PR 4 scope.

Issue #150, opened 2026-06-06
perf: add performance regression CI and dashboard artifacts


Jaruso:
## Summary
Add a lightweight performance regression workflow that can catch obvious throughput, latency, allocation, and memory regressions before they land.

## Research basis
Industry-grade performance work requires continuous measurement. Envoy, HAProxy, Caddy, and Varnish all emphasize observability. For Tardigrade, benchmark and telemetry work should feed into regression detection so performance improvements do not silently decay.

## Proposed design
- Add a reduced CI-safe benchmark profile.
- Store benchmark outputs as workflow artifacts.
- Compare PR results to a rolling baseline where feasible.
- Gate only on large regressions to avoid flaky failures.
- Track static tiny-file and proxy small-response benchmarks first.
- Add optional nightly/full benchmark workflow for heavier scenarios.
- Generate a Markdown summary for PRs if practical.

## Acceptance criteria
- CI can run a short benchmark smoke test without excessive flakiness.
- Benchmark artifacts include machine metadata and raw results.
- Regression thresholds are documented.
- Failures are actionable rather than noisy.
- Full benchmark suite can run manually or on schedule.

## Metrics to track
- req/s
- p50/p95/p99 latency
- RSS
- CPU time if available
- allocations/request if instrumentation exists
- proxy upstream reuse ratio if applicable

## Complexity
Medium.

## Risk
Medium. CI performance data can be noisy; thresholds must be conservative.

## Dependencies
Depends on #136 and benefits from #149.

Jaruso:
Closeout reconciliation — 2026-08-15

The original #150 acceptance is now implemented on `main`; this issue's GitHub state simply did not catch up with the repository.

Evidence checked against current `main`:

- `.github/workflows/ci.yml` contains a PR/push **Perf smoke** job on `ubuntu-latest`.
- It builds `tardi` with `-Doptimize=ReleaseFast`.
- It runs `benchmarks/ci-smoke.sh` and uploads the raw JSON result as the `perf-smoke-results` artifact even on failure.
- `benchmarks/ci-smoke.sh` explicitly identifies itself as a shared-runner coarse regression guard rather than release/marketing evidence.
- The bounded smoke exercises `static-http1`, `proxy-http1`, and `keepalive`.
- It fails on conservative throughput floors, non-zero errors, p99 ceilings, missing p95/p999/CPU/RSS fields, and an upstream-connection-reuse regression (<80%).
[...]

PR #600, opened 2026-08-10
perf(proxy): bound HTTP/1 relay buffers per origin (#140)

### LINKED ISSUE #140: perf: add watermark-based backpressure for proxy and connection buffers

Jaruso:
## Summary

Closes out the remaining #140 scope after #418, #595, and #599: per-origin aggregate accounting for HTTP/1, the tests that prove it, and the tuning guidance.

- **Per-origin enforcement for HTTP/1 relay buffers, both directions.** A fixed relay buffer bounds one request, but nothing bounded how many requests one origin had in flight — a slow origin could hold `concurrency × buffer` with every individual request looking well behaved, and only the process-wide limit stood between that and the box. `TARDIGRADE_PROXY_BUFFER_PER_ORIGIN_HARD_LIMIT_BYTES` now applies to HTTP/1 exactly as it already did to HTTP/2 stream queues.
- **Origin identity.** Keyed as the connection pool keys origins. An origin reached over TLS whose ALPN negotiated HTTP/1.1 is charged under its HTTP/1 key rather than the HTTP/2 pool key it was acquired through — otherwise one origin's memory was split across two limits. An origin gets an account whenever it is proxied to, including with connection pooling disabled: the account bounds memory, which pooling has no bearing on.
- **Bounded-cardinality metrics.** `tardigrade_upstream_pool_buffered_bytes{upstream,direction}` and `tardigrade_upstream_pool_buffer_limit_exceeded_total{upstream,direction}`, the HTTP/1 counterparts of the h2 pool series and read from the account that does the enforcing, so either is directly comparable with the configured limit. Two fixed directions per configured origin. They come from their own map rather than the `tardigrade_upstream_pool_*` connection series, because the two memberships genuinely differ.
- **Failure semantics.** An HTTP/1 response can never meet a capacity refusal after commitment — its whole reservation is taken before the response head is written — so a refusal is always a clean pre-commitment `503 proxy_buffer_saturated`, never charged to upstream health. A reloaded limit applies immediately, including to origins already holding reservations.
- **Concurrency cost.** The per-request account lookup takes its own lock rather than the shared pool mutex, so it neither joins checkout/release traffic nor pollutes the pool's lock-contention benchmark metrics.
- **Tuning guidance** in `docs/PROXY_STREAMING.md`: what each of the five limits trades against, how to size the per-origin limit from expected concurrency, a symptom → series → meaning table, and which gauge *not* to compare with the global limit. Benchmark scenarios (#149) and CI regression thresholds (#150) are documented as staying with those issues.

## Test plan

- [x] `zig build test` — all unit tests pass.
- [x] Reservations return to zero on **success, upstream abort mid-body, read timeout, cancellation, and downstream (client) abort** — five new tests driving the real `streamProxyOverTransport` response relay over a socketpair, each asserting the stream, origin, and global scopes all drain.
- [x] **Concurrent HTTP/1 traffic respects the per-origin cap** — three relays park mid-response against an origin sized for two; the third is refused at the `origin` scope specifically, before its head is committed, without the cap ever being exceeded, and every scope drains once the admitted relays finish.
- [x] A cancelled HTTP/1 upload releases its origin reservation.
- [x] Pool-level tests: pointer stability across rehashes, per-origin isolation, reload applying to existing *and* new origins, snapshot reporting both directions and refusals, and an account existing with pooling disabled.
- [x] `zig build test-integration` — new end-to-end test: a parked upload holds 16 KiB against its origin, a second concurrent upload to that origin gets `503 proxy_buffer_saturated`, the parked upload still completes, and the origin's gauge returns to zero. The existing full-streaming-upload test gained mid-flight and post-exchange per-origin gauge assertions.
- [x] `zig fmt --check src/ tests/`.

Pre-existing failure, unrelated and confirmed failing on `main` at `efa3db66` before any change here: `integration.test.reload while serving short static requests drops no in-flight request (#170)`.

Closes #140


docs/PROXY_STREAMING.md
```diff
@@ -324,3 +357,99 @@ write that begins to block only *after* the check passes is deliberately not
...
+`tardigrade_upstream_h2_pool_buffered_bytes{upstream,direction}` (HTTP/2). Both
+come from the enforcing account, so they are directly comparable with the
+limit. A non-zero
+`…_buffer_limit_exceeded_total{upstream,direction}` on one origin while others
+stay quiet is the signal this limit is designed to produce: that origin is
+absorbing memory and is being contained rather than allowed to spread.
+
+### 3. Global: what is this process's ceiling?
+
+`TARDIGRADE_PROXY_BUFFER_GLOBAL_HARD_LIMIT_BYTES` is the process-wide backstop
```
Jaruso, line 518: **Blocking — this process-wide ceiling is not complete yet, so #140's first acceptance criterion is still unmet.** `executeStreamingHttpProxyRequest` allocates one `read_buf` per request. HTTP/1 explicitly reserves `read_buf.len` for the response, but on the h2 path the upload reservation is released at the upload/response boundary and the same `read_buf` is then used by `readStreamingBody()` without a response-side reservation. `readStreamingBody()` copies bytes out of `stream.body`, and `acknowledgeStreamingBody()` does not release the queue reservation until *after* the downstream write, so a blocked downstream temporarily leaves two application-owned copies (queue + relay buffer) while only the queue is represented in stream/origin/global accounting. N slow h2 responses can therefore add roughly `N * read_buf.len` beyond these configured ceilings.

The document itself still says a few paragraphs above that this buffer's HTTP/2 response-phase ownership is unaccounted. Since this PR says it closes #140, that gap needs to be fixed here (or #140 must stay open). I would expect the h2 response relay buffer to be charged before downstream commitment, against the same stream/origin/global policy, held for the response relay lifetime, and released on every exit. Please add a concurrent-h2 test that proves the cap includes both queue storage and the relay buffer and that all scopes return to zero.

src/http/upstream_pool.zig
```diff
@@ -186,9 +246,82 @@ pub const UpstreamPool = struct {
...
+    /// Apply a reloaded proxy buffer policy. The per-origin hard limit takes
+    /// effect immediately for every origin, including ones already carrying
+    /// reservations: bytes already reserved above the new limit stay reserved,
+    /// and the next reservation is refused until the origin drains.
+    pub fn setProxyBufferLimits(self: *UpstreamPool, limits: proxy_buffer_account.Limits) void {
+        self.origin_buffers_mutex.lock();
+        defer self.origin_buffers_mutex.unlock();
+        self.config.proxy_buffer_limits = limits;
+        var it = self.origin_buffers.valueIterator();
+        while (it.next()) |account| account.*.setHardLimit(limits.per_origin_hard_limit);
```
Jaruso, line 292: **Blocking — the documented immediate-reload guarantee is racy.** Holding `origin_buffers_mutex` here serializes map/config updates, but relays already hold stable `Aggregate*` pointers and reserve without this mutex. `Aggregate.reserve()` currently loads `hard_limit` once *before* its CAS loop. A reservation can therefore (1) read the old larger limit, (2) race this `setHardLimit()` store of a smaller limit, and then (3) successfully CAS its byte count using the stale policy. Multiple already-running reservers can exceed the new cap after reload, despite the comment saying the next reservation must be refused.

Please make reload vs. reserve linearizable in the shared `Aggregate` primitive and add a concurrent reload test, not only the current sequential one. One workable shape is a generation/revalidation protocol around a successful reservation (rolling the reservation back if a newer limit makes it invalid); serializing reserve/set-limit on a scope-local lock would also be correct, though it changes the existing lock-free design.

docs/PROXY_STREAMING.md
```diff
@@ -324,3 +357,99 @@ write that begins to block only *after* the check passes is deliberately not
...
+`TARDIGRADE_PROXY_BUFFER_PER_ORIGIN_HARD_LIMIT_BYTES` bounds concurrency ×
+per-stream for a single origin, on both protocols. Set it from the concurrency
+you actually want to sustain to that origin:
+
+```
+per_origin ≈ expected concurrent slow requests × per_stream_high_watermark
+```
+
+It must be at least the per-stream hard limit — one request has to fit — and
+that is also the floor at which it admits exactly one request at a time.
```
Jaruso: This tuning statement is incorrect for the newly supported HTTP/1 path (and is too strong for h2 generally). The minimum valid per-origin limit is the per-stream hard limit, but an HTTP/1 request reserves only its fixed relay buffer (`max(proxy_stream_buffer_size, 16 KiB)` for responses / 16 KiB for uploads). With the defaults, a 1 MiB per-origin floor can admit about 64 concurrent 16 KiB HTTP/1 relay reservations in one direction — not exactly one request. H2 can also admit multiple streams when each retains less than the hard limit.

Please reword this as something like “this floor guarantees that one maximum-size stream can fit,” and make the sizing guidance protocol-aware: h2 concurrency is driven primarily by the configured stream/window budget, while h1 concurrency is driven by the fixed relay-buffer size.

Jaruso (commented):
I reviewed all 10 changed files against #140's acceptance criteria and the prior #418/#595/#599 slices. The HTTP/1 per-origin work itself is well structured: stable origin-account pointers, separate lookup locking, pooling-disabled coverage, ALPN-h1 key normalization, release-path tests, metrics, and integration coverage all line up with the intended PR-4 slice.

I don't think #140 can be closed yet, though. There are two correctness gaps:

1. The HTTP/2 response relay's per-request `read_buf` is still not charged to the stream/origin/global accounting after the upload reservation is released. `readStreamingBody` copies queued bytes into that buffer, and queue accounting remains until `acknowledgeStreamingBody` after the downstream write, so while a slow downstream blocks there are two application-owned copies but only the h2 queue is represented by the hard-limit accounting. With N concurrent h2 responses, memory can therefore exceed the configured aggregate ceilings by roughly `N × read_buf.len`. This directly misses #140's acceptance criterion that every application-owned proxy body buffer be accounted; the document itself still acknowledges this gap earlier in the request-direction section.

2. `UpstreamPool.setProxyBufferLimits()` claims the new limit applies immediately to existing accounts, but reservations do not take `origin_buffers_mutex`, and `Aggregate.reserve()` snapshots `hard_limit` before its CAS loop. A reservation can read the old limit, race a reload that stores the new lower limit, then successfully CAS under the old policy. The sequential reload test cannot catch this. The shared aggregate primitive needs a reload/reserve linearization mechanism (or the guarantee/docs need to be weakened, but that would not match this PR's stated semantics).

I also left a documentation correction for the claim that the minimum valid per-origin limit admits exactly one request; that is not true for HTTP/1 fixed relay buffers and is not generally true for partially occupied h2 streams.

Please address these before merging/closing #140. CI is otherwise looking healthy at review time: format, unit/build/security/perf jobs are green; the ubuntu appliance-profile job is still running its full integration stage.

Jaruso (commented):
Re-review: there are no new commits since my previous review; the PR is still at `963b5880e2b396e4257dd36391cd31c72e4d48cb`, and all three prior inline threads remain unresolved. CI run 1313 has now completed successfully, which is good, but it does not change the two correctness blockers: (1) the HTTP/2 response relay `read_buf` remains outside the stream/origin/global accounting while the queue reservation is still live, so #140's "every application-owned proxy body buffer is accounted" acceptance criterion is not yet met; and (2) `Aggregate.reserve()` can still complete against a stale hard-limit load after a concurrent reload lowers the limit, so the documented immediate-reload guarantee is not linearizable. The per-origin tuning statement also remains inaccurate for HTTP/1. I would still not merge this or close #140 until the existing threads are addressed.

docs/PROXY_STREAMING.md
Jaruso: Fixed in f69e76b4 — you were right, and the gap was exactly as you described.

The response relay buffer is now reserved in `streamViaH2Pool` for the response phase, in `.upstream_to_downstream` against the same stream/origin/global policy. Placement details:

- **Taken before the commitment boundary** — immediately after `body_allowed` is computed and before `beginDownstreamCommit`, so a refusal is a clean pre-commit `503 proxy_buffer_saturated` rather than a committed status that then has to be truncated.
- **Released on every exit** via a `defer` scoped to the `.h2` block, so the retry `continue` paths release it too.
- **Bodiless responses are not charged** — they never touch the buffer. That is the same line HTTP/1 draws, and it keeps a `HEAD`/`204` from holding a relay buffer's worth of budget for nothing.

New test: `an http2 response relay buffer is charged to the aggregate scopes while it relays`. It parks the relay in exactly the window you identified — the origin sends a head plus one DATA frame without `END_STREAM`, and a gated downstream writer blocks inside the first chunk write, which is after the bytes have been copied into the relay buffer and before `acknowledgeStreamingBody` releases the queue. While parked it asserts:

- `global.currentBytes(.upstream_to_downstream) >= read_buf.len` and the same on the pool's real origin account (fetched via `originBufferAccount`, not a stand-in). Before the fix this read only the queue's retained bytes for a two-byte body, so the 64 KiB buffer is the discriminator.
- A concurrent reservation of `read_buf.len` against a ceiling sized for two buffers is **refused** — i.e. the cap genuinely counts both the queue storage and the relay buffer, which is the property that was missing.
- Every scope returns to zero after the exchange completes, origin account read before `pool.deinit()` frees it.

The stale paragraph in the request-direction section is gone; that doc now says the buffer is re-reserved in the response direction, and the HTTP/2 response-bounds section explains the double-copy and why charging only the queue was insufficient.

src/http/upstream_pool.zig
Jaruso: Fixed in f69e76b4, in the shared `Aggregate` primitive as you suggested, keeping the lock-free design.

Two changes to `reserve`:

1. **The limit is re-read on every attempt** rather than once before the loop. This was the worse half: a reserver spinning in the CAS retry loop carried its stale policy through every retry, so contention widened the window rather than narrowing it.
2. **The commit is revalidated.** After winning the CAS the reserver re-reads the limit, and if a shrink beat it there it rolls its own bytes back with a `fetchSub` and returns `BufferLimitExceeded` (counted as a limit-exceeded event).

`setHardLimit`'s store, the policy loads, and the CAS are all `seq_cst`, which puts them in a single total order. So a commit either precedes the shrink — the documented "already reserved above the new limit stays reserved" case — or follows it and is refused. There is no interleaving left where a reservation reads the old limit, is overtaken, and still commits.

The fast path is unchanged in shape: uncontended with no concurrent reload it is one CAS plus two limit loads, and the rollback is only ever reached by a reservation that genuinely raced a reload.

**On testing it.** A concurrent test alone cannot actually pin this, and I want to flag that rather than imply otherwise. An outside observer sampling the scope after a shrink cannot distinguish a violation from the guarantee, because bytes admitted *before* the reload legitimately stay reserved for as long as their holder keeps them — so "current > new limit" is a valid state. So there are two tests:

- `a reload landing inside the commit window refuses the racing reservation` — deterministic. A test-only seam (`reserve_race_hook`, gated on `builtin.is_test` so production builds do not contain the branch) fires after the policy has been read and approved and before the commit lands; the hook shrinks the limit there. This is placed to catch **both** halves: the pre-check used the stale limit, so only revalidation can refuse it, and if the limit read were moved back out of the loop the test would still fail. Plus `a reload that lands after the commit leaves the reservation alone` for the mirror case, so revalidation cannot be "reject whenever the limit changed".
- `concurrent reservations settle under a limit reloaded beneath them` — 8 threads hammering the CAS and rollback paths while the limit is flipped. Its assertion is one an observer soundly can make: it waits until every thread has completed a full reserve/release cycle after the shrink (so nothing admitted under the old policy can still be outstanding) and only then requires the scope to be within the new limit. It also asserts both admissions and refusals actually occurred, so the run is known to have exercised the window, and that nothing leaked or double-released.

`docs/PROXY_STREAMING.md`'s reload section now states the linearization explicitly instead of just claiming "immediately".

docs/PROXY_STREAMING.md
Jaruso: Corrected in f69e76b4 — the claim was wrong and your arithmetic is right.

The floor is now stated as "it guarantees that one maximum-size stream fits, **not** that only one request fits", with your worked example: with the shipped defaults a 1 MiB per-origin floor admits roughly 64 concurrent 16 KiB HTTP/1 relay reservations in one direction, and it admits several HTTP/2 streams whenever each retains less than the full per-stream hard limit. I also noted what you'd actually do to admit exactly one HTTP/1 request — set the limit to the relay buffer size, which means lowering the per-stream hard limit too, since the floor still applies.

Sizing is now protocol-aware, as a table rather than one formula:

| Protocol | What one request holds | Sizing |
| --- | --- | --- |
| HTTP/1 | A fixed relay buffer: `max(proxy_stream_buffer_size, 16 KiB)` for a response, 16 KiB for an upload. Does not grow with the body. | `per_origin ≈ concurrent requests × relay buffer` |
| HTTP/2 | The stream's queue, filling toward the per-stream high watermark under a slow consumer, plus the response relay buffer while relaying. | `per_origin ≈ concurrent slow streams × (per_stream_high_watermark + relay buffer)` |

The HTTP/2 row now includes the relay buffer, since that is charged as of this revision. And I added the consequence an operator would want stated outright: because HTTP/1's per-request cost is fixed and small, an HTTP/1 origin sustains far more concurrency per byte of budget than an HTTP/2 origin whose streams each fill a window — so size against whichever protocol the origin actually negotiates.

src/http/proxy_buffer_account.zig
```diff
@@ -264,10 +296,27 @@ pub const Aggregate = struct {
...
+            }
+            if (slot.cmpxchgWeak(current, next, .seq_cst, .seq_cst)) |actual| {
                 current = actual;
                 continue;
             }
+            // Committed. A reload may have landed between the policy read
+            // above and this commit; if it did, these bytes were never
+            // admissible, so give them back rather than leave the scope over a
+            // limit that is already in force.
+            const limit_now = self.hard_limit.load(.seq_cst);
```
Jaruso: **Blocking — the post-CAS rollback is still not linearizable, because the tentative reservation is visible to other reservers before this operation decides whether it succeeded.** `seq_cst` gives the atomics a total order, but it does not make `current + hard_limit` one atomic transaction.

There are two indistinguishable cases at this line:

1. `read old limit -> reload shrink -> CAS -> read new limit`: this reservation must fail.
2. `read old limit -> CAS -> reload shrink -> read new limit`: the CAS happened before the reload, so under the documented semantics this is the **already-reserved** case and should remain admitted.

This code rolls back in both cases. More importantly, the CAS has already published those bytes. A concrete externally-visible bad history is:

```text
current=0, old_limit=100
A: reserve(80) -> CAS publishes current=80
reload: setHardLimit(50)
B: reserve(10) -> sees current=80/new_limit=50 -> REFUSED
A: post-CAS revalidation sees 50 -> rolls back to 0 -> REFUSED
```

Both A and B return failure and final `current` is 0, but B was a 10-byte reservation made under a 50-byte limit and should have succeeded. There is no sequential history of successful/failed reservations that explains that refusal, so the implementation is not linearizable and can surface a spurious `503 proxy_buffer_saturated` during reload.

The existing deterministic test only forces the shrink **before** the CAS. The "reload after commit" test calls `setHardLimit` after `reserve()` has fully returned, so it does not cover the crucial `CAS -> shrink -> revalidation` window.

Please make reload exclude in-flight *reservation attempts* rather than publishing then rolling back. A simple correctness baseline is a scope-local lock around check+commit and reload. If keeping the normal path lock-free matters, one workable shape is a rare-reload admission gate:

```zig
// sketch only
while (true) {
    while (self.reloading.load(.acquire)) std.Thread.yield() catch {};
    _ = self.inflight_reservers.fetchAdd(1, .acq_rel);
    if (self.reloading.load(.acquire)) {
        _ = self.inflight_reservers.fetchSub(1, .release);
        continue;
    }
    defer _ = self.inflight_reservers.fetchSub(1, .release);

    // The limit is stable for this entire check/CAS attempt.
    const limit = self.hard_limit.load(.acquire);
    // ...existing current CAS loop, with no post-publication rollback...
}

pub fn setHardLimit(self: *Aggregate, new_limit: usize) void {
    self.reloading.store(true, .release);
    while (self.inflight_reservers.load(.acquire) != 0) std.Thread.yield() catch {};
    self.hard_limit.store(new_limit, .release);
    self.reloading.store(false, .release);
}
```

A mutex/RW-lock or a genuinely single atomic state is also fine. Please add a deterministic `CAS -> shrink -> second reserver -> revalidation` test; it should prove the second small reservation is never falsely refused and that a reservation whose commit precedes the reload is treated consistently.

src/gateway_proxy.zig
```diff
@@ -1083,6 +1090,43 @@ fn streamViaH2Pool(
...
+                // Taken before the commitment boundary below, so a refusal is
+                // still a clean pre-commit 503 rather than a committed status
+                // that has to be truncated, and released on every exit from
+                // this block — including the retry `continue` paths. A
+                // bodiless response never touches the buffer and is not
+                // charged for it, which is the line HTTP/1 draws too.
+                var response_reservation: ?ProxyBufferReservation = null;
+                defer if (response_reservation) |*reservation| reservation.releaseAll();
+                if (body_allowed) {
+                    var reservation = ProxyBufferReservation.init(
```
Jaruso, line 1226: **Blocking — this accounts the missing H2 relay buffer, but it gives it a second independent `scope=.stream` hard-limit account rather than sharing the stream's existing budget.** `ProxyBufferReservation.init()` constructs a fresh `Account.init(..., .stream, proxy_buffer_limits)`, while the H2 queue already has `stream.proxy_body_account`. The queue and relay therefore each get to consume up to `PER_STREAM_HARD_LIMIT_BYTES` independently.

That breaks #140's acceptance criterion that the configurable **per-stream hard limit is enforced** once these two buffers coexist. For example, this is a valid policy today:

```text
low  = 256 KiB
high = 768 KiB
hard = 768 KiB   // high <= hard is valid
relay buffer = 16 KiB
origin/global = unlimited
```

A compliant H2 peer may fill the stream queue toward the advertised 768 KiB high watermark. This fresh account then independently admits another 16 KiB relay reservation, so one stream can own ~784 KiB while its configured hard limit is 768 KiB. Neither local account reports a hard-limit exceedance. If `per_origin` is set to its currently-valid/documented floor of 768 KiB instead, the same queue+relay overlap can turn into a timing-dependent pre-commit 503, so the doc claim that the floor guarantees one maximum-size stream fits is also not generally true after this change.

Please enforce one shared hard budget across the simultaneous queue + relay ownership while keeping the queue's high/low hysteresis separate. Conceptually I'd expect something like:

```zig
const Stream = struct {
    proxy_queue_account: ?proxy_buffer_account.Account, // logical queue/high-low
    proxy_stream_hard_budget: ?proxy_buffer_account.Aggregate, // total owned response bytes
};
```

The H2 queue's retained storage and this response relay buffer would both reserve/release against `proxy_stream_hard_budget`; only `proxy_queue_account` should drive high/low transitions and WINDOW_UPDATE behavior. Origin/global reservations remain additional scopes. Since the per-stream policy is connection-pinned, this hard budget can be initialized from the connection's pinned `per_stream_hard_limit` and does not need reload mutation.

Please add a test with `high == hard` and a live relay buffer: drive the queue near high while the downstream write is gated and prove the total stream-owned queue + relay bytes cannot exceed `hard` (and that any refusal happens deterministically before commitment). The current new test proves origin/global accounting, but it never asserts the per-stream hard budget across both copies.

Jaruso (commented):
Re-review of f69e76b4 against #140. The previous missing-H2-relay-buffer finding is materially addressed, the protocol-aware tuning correction is good, and CI run #1314 is green. I still don't think #140 can close yet: the reload fix is not actually linearizable because it publishes tentative `current` before deciding whether the reservation succeeds, and the new H2 relay reservation creates a second independent stream account instead of enforcing the configured hard limit across the whole stream's simultaneous queue + relay ownership. Details and concrete fix shapes are inline.

src/http/proxy_buffer_account.zig
Jaruso: You're right, and my reasoning was wrong in a way worth naming: I argued that `seq_cst` let the post-CAS load distinguish the two orderings. It doesn't. The load tells you the limit *now*, not whether the store was ordered before or after your CAS — the store can sit between them — so case 1 and case 2 are indistinguishable from there and I rolled back both. And your second point is the more damaging one anyway: the tentative commit is visible to other reservers before this operation decides, so your A/B history produces a refusal with no sequential explanation.

Replaced with the gate, in 36a95fe1, essentially your sketch:

```zig
fn enterReservation(self: *Aggregate) void {
    while (true) {
        while (self.reloading.load(.acquire)) std.Thread.yield() catch {};
        _ = self.inflight_reservers.fetchAdd(1, .acq_rel);
        if (!self.reloading.load(.acquire)) return;   // closes the set-between-check-and-increment race
        _ = self.inflight_reservers.fetchSub(1, .acq_rel);
    }
}
```

`setHardLimit` claims `reloading` with a CAS (so concurrent reloads can't interleave one's wait with another's store), waits for `inflight_reservers == 0`, publishes, then clears. `reserve` reads the limit once inside the gate and it is stable for the whole attempt; no tentative publication, no rollback. It can't stall behind a blocked reserver, since an in-flight attempt is a CAS loop that waits on nothing.

Tests, replacing the two that no longer describe the design:

- `a reload cannot publish while a reservation attempt is in flight` — deterministic, and it needed the seam repositioned. The hook now *holds* an attempt open from inside the gate rather than mutating the limit (a hook calling `setHardLimit` would deadlock against its own in-flight count, so the doc comment says so). A reserver is parked mid-attempt, a reloader runs on another thread, and the test spins asserting `hardLimit()` still reads the old value and the reload has not completed. Then it releases the reserver and checks the reservation **committed and was not rolled back** — your case 2 — and only then does the new limit appear.
- `a refused reservation never strands a smaller one that fits` — your A/B history directly: 80 refused under a 50 limit leaves `current == 0`, and the 10-byte reservation is then admitted.

The concurrent stress test still passes unchanged, and I left it explicitly documented as covering "doesn't leak, double-release or livelock under contention" rather than pretending it covers the ordering.

src/gateway_proxy.zig
Jaruso: Correct, and it's a gap my own previous commit introduced — before it the relay buffer wasn't accounted at all, so there was no second budget to double up. Fixed in 36a95fe1 along the lines you sketched.

`AggregateCapacity` gains a `stream` scope, and `streamViaH2Pool` creates one `Aggregate` per stream sized from the connection-pinned `per_stream_hard_limit`. Both the queue's retained storage and the response relay buffer reserve against it, so one stream's total owned bytes are bounded once. Two details worth noting:

- **`Aggregate` already separates directions** (`current[2]`), so a single budget object gives per-direction totals for free — an upload and a response on one stream stay independent budgets rather than sharing one, which is what you'd want.
- **The queue's `Account` is untouched**, so high/low hysteresis and `WINDOW_UPDATE` still run purely on logical queue length, as you asked. The relay reservation keeps its own `Account` too, but only for the `scope="stream"` gauge and watermark events — the shared budget is what binds.

Lifetime: the budget is block-scoped in `streamViaH2Pool`. Every path out of that block calls `finishStreaming` first, which removes the stream from the connection map under `state_mutex` and destroys it, so no reader can reach the budget afterwards; a retry gets a fresh one, which is right since it's a fresh stream. As you note, the connection-pinned policy means it needs no reload mutation.

**On your `high == hard` example — I went further than sharing the budget, and this is the part I'd most like you to check.** Sharing alone doesn't make that config work; it makes it *truncate*, because a compliant peer filling its advertised 768 KiB window plus a 16 KiB relay buffer exceeds a 768 KiB budget, and the overrun lands on the queue post-commitment. Turning a previously-working config into a truncating one silently seemed worse than the bug. So `validateProxyBufferLimitsCoverRelayAllocations` now also requires:

```
per_stream_hard_limit >= per_stream_high_watermark + max(proxy_stream_buffer_size, 16 KiB)
```

That makes the policy self-consistent — the hard limit provably covers everything one stream can own — and restores your point about the per-origin floor: with the rule in force, a floor of `per_stream_hard_limit` really does guarantee one maximum-size stream fits.

**This can reject a config that previously started**, which is a deliberate user-visible change and flagged in the CHANGELOG. The shipped defaults satisfy it with room to spare (768 + 16 ≤ 1024 KiB), and the only in-tree config that needed adjusting was the integration test I added last round (16384 → 18432). If you'd rather not tighten validation in this PR, the alternative is to accept post-commit truncation for tight policies and weaken the floor claim in the docs instead — say the word and I'll switch.

Tests:

- `an http2 stream cannot own more than its per-stream hard limit across queue and relay` — h2 exchange with `hard == high + relay` exactly (the tightest policy validation now permits), downstream write gated mid-relay, asserting the total owned is `>= relay buffer` (so the buffer is genuinely in there) and `<= per_stream_hard_limit` (so it isn't two budgets), then zero afterwards.
- `an http2 stream budget refuses the relay buffer before the response is committed` — a queue holding most of the budget, relay reservation refused at `scope="stream"` specifically with no other scope configured, rolled back cleanly with the queue's claim intact.
- `per-stream hard limit must cover a full window plus its relay buffer` — the validation rule, including that the defaults pass.

src/gateway_proxy.zig
```diff
@@ -961,7 +968,32 @@ fn streamViaH2Pool(
...
+                // the connection's map under its state lock and destroys it, so
+                // no reader can reach this budget afterwards. A retry starts a
+                // fresh one, as it should — it is a fresh stream.
+                //
+                // The per-stream policy is connection-pinned, so this needs no
+                // reload mutation: the stream is judged by the policy its
+                // connection advertised.
+                var stream_hard_budget = proxy_buffer_account.Aggregate.init(
+                    .stream,
+                    proxy_buffer_limits.per_stream_hard_limit,
```
Jaruso: **Blocking — this is not actually connection-pinned.** `proxy_buffer_limits` is the current request/config snapshot: `executeStreamingHttpProxyRequest()` passes `cfg.proxy_buffer_limits` into `streamViaH2Pool`. But a reused H2 connection intentionally keeps the policy it negotiated when it was opened: `H2Conn.openStreaming()` initializes `stream.proxy_body_account` from `self.buffer_limits`, and the reload code explicitly says the per-stream policy reaches *new connections only*.

So after a hot reload, one stream can now be governed by two generations of the same policy:

```text
connection opened under A: high=768 KiB, hard=1 MiB
reload to B:             high=1 MiB,   hard=2 MiB
new stream reuses old connection

queue Account            -> A (self.buffer_limits)
shared queue+relay budget -> B (this line)
relay local Account       -> B (ProxyBufferReservation.init(... proxy_buffer_limits ...))
```

On a raise, the shared budget can permit queue+relay ownership above the connection-pinned hard limit that the docs say the stream is measured against. On a shrink, the opposite happens: the new budget can reject a relay with a pre-commit `503` even though the peer is still operating under the larger receive window/policy that this connection advertised. That makes behavior depend on whether the request happened to land on a pre-reload pooled connection.

Please source **all response-side per-stream policy for an H2 stream from the acquired connection**, not from `cfg`. One reasonable shape is to expose the connection's immutable pinned limits before `openStreaming`:

```zig
const pinned = conn.proxyBufferLimits(); // returns self.buffer_limits
var stream_hard_budget = Aggregate.init(.stream, pinned.per_stream_hard_limit);
...
const response_reservation = ProxyBufferReservation.init(
    .upstream_to_downstream,
    pinned,
    observer,
    capacity,
);
```

An even cleaner ownership model is to put the shared hard budget on `Stream` and initialize it inside `openStreaming()` from `self.buffer_limits`, alongside `proxy_body_account`; then the queue cannot accidentally receive a caller-snapshot stream budget at all.

There is one related cross-version detail to handle in the test/design: `read_buf` is sized from the **current** `cfg.proxy_stream_buffer_size`. Now that validation requires `hard >= high + relay`, increasing the relay size on reload can make a new buffer too large for an old connection's pinned `hard/high` pair. Either pin the H2 relay-buffer footprint with the connection as part of this policy, cap the H2 relay slice to what the pinned policy can cover, or retire/reconnect H2 connections when that footprint changes.

Please add a reload regression test that opens/reuses an H2 connection under policy A, reloads to policy B, then starts a new stream on that same connection and proves the queue, shared hard budget, and response relay are all judged by A until a new connection is established.

Jaruso (commented):
Re-review of 36a95fe1: the two blockers from the previous round are fixed. The reload gate is now linearizable (no tentative publication/rollback), and the H2 queue + relay share one per-stream aggregate budget with sensible validation and focused tests. CI #1315 is green.

One blocking reload-policy mismatch remains: the new shared stream budget is described as connection-pinned, but it is initialized from the current request's `cfg.proxy_buffer_limits`, while `H2Conn.openStreaming()` deliberately initializes the queue account from the connection's pinned `self.buffer_limits`. After a hot reload, a reused H2 connection can therefore enforce two different generations of the per-stream policy on the same stream. I left the concrete interleavings and an implementation/test shape inline.

Once that is fixed, I do not currently see another blocker to closing #140.

src/gateway_proxy.zig
Jaruso: Right on both counts, and I should not have written "connection-pinned" in that comment when the value came from `cfg`. Fixed in fa2a152a.

**Policy source.** Added `H2Conn.proxyBufferLimits()` returning the immutable `self.buffer_limits`, and `streamViaH2Pool` now takes `const pinned_limits = conn.proxyBufferLimits()` immediately after acquiring the connection. Everything per-stream is sized from it: the shared queue+relay budget, the response `ProxyBufferReservation`, and the upload preflight too — the upload charges the same shared budget, so leaving it on `cfg` would have reproduced the split in the request direction.

I went with the accessor rather than moving the budget onto `Stream`. Your second shape is cleaner in the sense you describe, but the budget also has to be reachable by the relay reservation in `gateway_proxy`, so putting it on `Stream` means exposing it back out anyway; the accessor keeps one owner and the block-scoped lifetime already tied to `finishStreaming`. Happy to move it inside `openStreaming` if you'd rather have the queue structurally unable to receive a caller-snapshot budget — say so and I'll do it.

**The `read_buf` sizing detail.** Good catch, and it is a consequence of the validation rule I added last round. I took your "cap the H2 relay slice" option:

```zig
const pinned_relay_bytes = @min(
    read_buf.len,
    pinned_limits.per_stream_hard_limit - pinned_limits.per_stream_high_watermark,
);
const relay_buf = read_buf[0..pinned_relay_bytes];
```

`relay_buf` is what the h2 path reserves and relays through, in both directions. The cap can never bite below 16 KiB, since any validated policy was checked against a relay of at least that, so `hard - high >= 16 KiB` for whatever policy pinned the connection. Excess allocation just goes unused until the connection is replaced. I preferred this to retiring connections on a footprint change: the buffer is per request rather than per connection, so a `proxy_stream_buffer_size` change would otherwise churn every pooled connection for a reason the peer cannot observe.

**Reload regression test:** `a stream on a pooled http2 connection is judged by the policy that connection pinned`. It needs one connection to survive the reload, so the origin fixture serves two sequential streams on a single connection. Exchange one opens under policy A (`high = 16 KiB`, `hard = 32 KiB` — exactly one window plus one relay buffer, the tightest validation now allows) and completes, returning the connection to the pool. The pool is reloaded to policy B (four times larger in every dimension, `hard = 128 KiB`), and exchange two runs with B as its config snapshot and a 64 KiB `read_buf`, reusing the A-pinned connection and parking mid-relay. The assertion is that everything the stream owns stays `<= policy_a.per_stream_hard_limit` and `< policy_b.per_stream_hard_limit`. Before the fix the relay alone reserved B's 64 KiB, which is twice A's entire hard limit, so it fails loudly.

src/gateway_proxy.zig
```diff
@@ -961,11 +968,63 @@ fn streamViaH2Pool(
...
+                // goes unused for the life of this connection.
+                //
+                // Never restrictive in practice: the pinned policy was
+                // validated against a relay of at least 16 KiB, so this is
+                // always at least that.
+                const pinned_relay_bytes = @min(
+                    read_buf.len,
+                    pinned_limits.per_stream_hard_limit - pinned_limits.per_stream_high_watermark,
+                );
+                const relay_buf = read_buf[0..pinned_relay_bytes];
```
Jaruso: **Blocking — this caps the H2 *slice*, not the backing allocation, so the latest reload fix still leaves application-owned relay memory outside #140's limits.** `executeStreamingHttpProxyRequest()` allocates `read_buf` up front at `max(cfg.proxy_stream_buffer_size, 16 KiB)` and keeps that allocation alive for the whole exchange. On a reused older H2 connection, this code then slices it down to `pinned_relay_bytes` and reserves only `relay_buf.len`; the remainder is still real per-request memory owned by the proxy, just unused.

That matters because this accounting model explicitly treats **retained allocation** as the quantity aggregate hard limits protect, not merely bytes currently populated. For example, if connection A pinned a 16 KiB relay allowance, then a reload grows `proxy_stream_buffer_size` to 1 MiB, every request reusing A allocates 1 MiB here but charges only 16 KiB to the stream/origin/global scopes. N concurrent streams can therefore add roughly `N × (1 MiB - 16 KiB)` beyond the configured ceilings. The new reload test actually constructs the same shape with a 64 KiB `read_buf`; it proves the *used slice* is pinned, but never checks the larger allocation it leaves behind.

Please make the allocation itself match the pinned relay size after the H2 connection is acquired (while keeping the HTTP/1 fallback's current-config allocation/charge semantics), or use another design that charges the full backing allocation without violating the connection-pinned stream budget. The regression test should observe the production allocation size (e.g. a counting allocator / executor-level test) and prove `allocated relay bytes == accounted relay bytes` across a buffer-size reload, not only `relay_buf.len <= pinned hard - high`.

Jaruso (commented):
Re-reviewed current head `fa2a152a` against #140 and the previous review threads. The reload gate in `Aggregate` now gives the reservation/reload ordering the documented linearization point, the shared H2 stream hard budget correctly prevents queue + relay from consuming two independent hard limits, and sourcing the response-side stream policy from the acquired connection fixes the mixed-generation policy bug.

There is still one blocking issue in the latest pinned-relay fix: it caps the **slice used by H2**, but not the backing per-request allocation. That leaves reload-grown `read_buf` capacity outside the stream/origin/global accounting even while the response relay is active, which is the exact physical-memory class #140 is intended to bound. I left the concrete case inline.

CI for this head is green, but I would not merge/close #140 until the allocated relay capacity and the charged capacity are the same (or the full allocation is otherwise covered by the same limits), with a regression test that observes the production allocation rather than only the smaller slice.

src/gateway_proxy.zig
Jaruso: Correct — I capped the slice and left the allocation, which is the thing that actually costs memory. Fixed in d3a391be.

`streamViaH2Pool` now takes `requested_relay_bytes: usize` instead of a `[]u8`, and allocates on the h2 path only once the connection — and therefore the policy that has to account for the buffer — is known:

```zig
const pinned_relay_bytes = @min(
    requested_relay_bytes,
    pinned_limits.per_stream_hard_limit - pinned_limits.per_stream_high_watermark,
);
const relay_buf = try allocator.alloc(u8, pinned_relay_bytes);
defer allocator.free(relay_buf);
```

Allocated and accounted are now the same number. `executeStreamingHttpProxyRequest` computes the size up front but defers its own allocation until *after* the h2 branch, so nothing is allocated for a path that won't use it — previously the h2 path was carrying a buffer allocated by the caller for the whole exchange. The ALPN-h1 fallback inside `streamViaH2Pool` allocates at the current config's size too, since a fresh unpooled TLS connection pins no policy of its own; that keeps your "HTTP/1 fallback's current-config allocation/charge semantics" requirement exactly as it was.

**Test observes the production allocation**, as you asked: `an http2 relay buffer allocates exactly what it accounts across a buffer-size reload`. It wraps the allocator in a `LargestAllocationTracker` that records the largest single allocation — the relay buffer dwarfs everything else a streaming exchange takes from the heap — then opens a connection under a policy with a 16 KiB relay allowance, reloads to one asking for 1 MiB, and reuses that connection. Assertions:

```zig
try std.testing.expectEqual(pinned_relay_bytes, tracker.largest);   // 16 KiB, not 1 MiB
try std.testing.expect(tracker.largest < reloaded_relay_bytes);
try std.testing.expect(counters.peak_reserved >= pinned_relay_bytes);
try std.testing.expect(counters.peak_reserved < reloaded_relay_bytes);
```

Slicing a 1 MiB allocation down to 16 KiB leaves `tracker.largest` reading 1 MiB, so the previous revision fails this outright.

One honest note on the accounted-side assertion: it's a band rather than an equality because the observer also sees the stream queue's couple of body bytes on top of the relay reservation, so `peak_reserved` is `relay + queue`. The equality that matters — allocation size — is exact.

src/gateway_proxy.zig
```diff
@@ -961,11 +981,66 @@ fn streamViaH2Pool(
...
+                // and what is accounted are deliberately the same number.
+                //
+                // Never restrictive in practice: the pinned policy was
+                // validated against a relay of at least 16 KiB, so this is
+                // always at least that.
+                const pinned_relay_bytes = @min(
+                    requested_relay_bytes,
+                    pinned_limits.per_stream_hard_limit - pinned_limits.per_stream_high_watermark,
+                );
+                const relay_buf = try allocator.alloc(u8, pinned_relay_bytes);
```
Jaruso: **Blocking — the allocation now has the right *size*, but it still happens before the capacity reservation, so the limits do not actually bound the memory being allocated.** #140 explicitly requires “reservation before allocating or accepting queued bytes.” Here every H2 exchange allocates `pinned_relay_bytes` immediately after acquiring the connection. For a response-only request, no matching reservation exists until *after* `waitStreamingResponseHead()` and `body_allowed` are known; for a streaming upload, `preflightUploadReservation()` is also below this allocation.

A concrete bad case is a global/origin cap sized for one 64 KiB relay buffer and N concurrent GETs whose origin is slow to produce response heads. All N requests reach this line and allocate 64 KiB first; none has charged stream/origin/global yet, so the aggregate cap cannot stop the process from holding `N × 64 KiB`. They are only refused later, after the memory peak the limit is supposed to prevent has already happened. A `HEAD`/`204` path can allocate the buffer and never charge it at all.

There is also a new cleanup bug at the same site: `conn` has already been acquired, but `try allocator.alloc(...)` can return before any `h2_pool.release(conn)` path runs, leaking that acquired pool reference on OOM. Previously the relay allocation happened before `h2_pool.acquire`, so this failure mode did not exist.

Please make allocation follow successful reservation, not merely match its eventual size. The cleanest lifecycle I see is phase-specific ownership:

```zig
// streaming upload
var upload_reservation = try preflightUploadReservation(..., pinned_relay_bytes, ...);
errdefer upload_reservation.releaseAll();
const upload_buf = allocator.alloc(u8, pinned_relay_bytes) catch |err| {
    h2_pool.release(conn);
    return err;
};
defer allocator.free(upload_buf);
// relay upload, then release/free at the upload boundary

// response
try conn.waitStreamingResponseHead(stream);
if (body_allowed) {
    var response_reservation = ProxyBufferReservation.init(...);
    try response_reservation.reserve(pinned_relay_bytes); // still pre-commit
    errdefer response_reservation.releaseAll();
    const response_buf = allocator.alloc(u8, pinned_relay_bytes) catch |err| {
        response_reservation.releaseAll();
        conn.finishStreaming(stream);
        h2_pool.release(conn);
        return err;
    };
    // relay, then free/release
}
```

That also avoids keeping an empty upload buffer allocated-but-unaccounted while waiting for the response head after the request-direction reservation has been released. Please extend the allocation regression test to prove **reservation precedes allocation** under a tight origin/global cap (e.g. multiple requests parked before the response head cannot allocate more relay memory than the admitted capacity), and exercise the allocator-failure path so the acquired H2 connection/ref is not stranded.

Jaruso (commented):
Re-reviewed head `d3a391be` against #140 and the prior threads. The previous two unresolved findings are fixed and I resolved those threads: all per-stream H2 policy now comes from the connection-pinned limits, and the relay backing allocation is now actually sized to that policy rather than allocating the current-config size and slicing it down. The reload gate/shared stream budget/tuning-doc fixes from the earlier rounds also still look correct. CI #1317 is green.

There is still one blocking #140 issue in the new allocation placement: the H2 relay buffer is allocated before the stream/origin/global reservation that is supposed to bound that allocation. That means the configured caps can reject the request only *after* the memory has already been consumed, so many concurrent H2 exchanges waiting for a response head can multiply relay allocations outside the aggregate ceilings. It also creates a new acquired-connection leak on allocator failure at this site. I left the concrete lifecycle/fix shape inline.

I would still keep #140 open until the relay allocation itself is reservation-first (and the allocation-failure path releases the acquired connection/reservations). After that, this looks very close to merge-ready.

src/gateway_proxy.zig
Jaruso: Both correct. Fixed in 44d75d3d — and chasing them turned up a pre-existing memory-safety bug, so this reply is longer than usual.

**Reservation now precedes allocation.** Each phase reserves first and allocates second:

- Upload: `preflightUploadReservation` → `allocator.alloc` → relay → free *and* release at the upload/response boundary (so an empty upload buffer is no longer held through the response, as you noted).
- Response: `waitStreamingResponseHead` → reserve → allocate → relay. Still pre-commitment, so a refusal is a clean 503.
- A bodiless response allocates nothing at all now.

**The cleanup bug** you spotted is fixed too: both allocation sites hand the connection reference back before returning. While there I found the same shape one line up — the `:path` build used a bare `try` after `acquire` — and gave it the same treatment.

**I did not keep the shared budget object, and I want to be explicit about why**, since it departs from what you asked for. Your `high == hard` fix from last round is intact, but implemented as reserved headroom: `openStreaming` now takes `proxy_relay_reserved_bytes` and subtracts it from the queue's own hard limit, so queue + relay ≤ `per_stream_hard_limit` with no object shared between them.

The reason is that the shared object was unsafe, and my "block-scoped is fine, every path calls `finishStreaming` first" argument in the previous round was wrong. `finishStreaming` removes the stream from the map under `state_mutex` but calls `stream.destroy` *outside* it, so a reader already inside `enqueueBody` can still be holding that stream — and with the budget on the worker's stack, it would write through a pointer into a dead stack frame. Headroom bounds the sum identically and has no lifetime coupling at all. `AggregateCapacity.stream` and `StreamBufferLimitExceeded` are gone again; validation's `hard >= high + relay` still guarantees the queue a full window.

**Pre-existing bug found on the way:** `H2ConnPool.evict` frees the key the connection map owns *before* `conns.remove(key)`. The removal then probes with a key whose stored counterpart is freed memory; on a hash collision it can fail to remove the entry at all, leaving a map entry whose key is already freed — double-freed by `deinit`, whose `release()` on the connection then faults. `acquire` and `reapIdle` already used `fetchRemove`-then-free; `evict` now matches. My new concurrent test reproduced it as a segfault plus a `[DebugAllocator] Double free detected` under parallel load. Called out separately in the CHANGELOG's Fixed section since it is not really part of #140.

**Tests, per your ask:**

- `http2 relay memory is not allocated until a reservation admits it` — three concurrent exchanges parked on an origin that withholds response heads, under a global cap sized for one relay buffer. Asserts relay-sized live allocation is **0** while they wait (it read 3 × 32 KiB before), and never exceeds one admitted reservation afterwards. The allocator tracks live relay-sized bytes, since "did N requests each take a buffer" is about concurrent ownership, which a largest-single-allocation figure can't answer. The pool is warmed with one exchange first so the three genuinely reuse one connection rather than racing to create it.
- `a failed relay allocation hands the http2 connection back` — an allocator that fails at the relay size; the exchange errors, and a subsequent exchange on the same pool still succeeds, with the leak checker covering the connection object.

Two honest notes. My first version of the concurrency test shared one `UploadBufferObserver` across threads — its counters are plain integers, so that was itself a data race; each exchange now owns one. And the threaded tests now tear down through `defer`s in the right order, because the original crash surfaced when an assertion failed early and left `pool.deinit()` running against live exchange threads.

src/gateway_proxy.zig
```diff
@@ -2363,12 +2531,11 @@ pub fn executeStreamingHttpProxyRequest(
...
-    // Aggregate capacity for this connection's relay buffers, in both
-    // directions (#140). The HTTP/1 pool has no per-origin buffer account — an
-    // h1 origin's relay memory is bounded per request by the fixed relay
-    // buffers, so the scope that concurrency can actually multiply is the
-    // process; per-origin accounting for h1 origins is the issue's PR 4.
-    const h1_buffer_capacity = proxy_buffer_account.AggregateCapacity{ .global = proxy_buffer_global };
+    // The HTTP/1 relay buffer, allocated and charged at the current config's
+    // size. An h1 connection pins no policy of its own, so unlike the h2 path
+    // there is no older generation for this to have to fit inside.
+    const read_buf = try allocator.alloc(u8, requested_relay_bytes);
```
Jaruso: **Blocking — the full #140 pass shows HTTP/1 still violates the same reserve-before-allocation invariant that this round fixed for H2.** This `read_buf` is allocated before the HTTP/1 origin account is even looked up, and `streamProxyOverTransport()` then passes it to `readUpstreamHead()` before creating the `upstream_to_downstream` `ProxyBufferReservation`. `readUpstreamHead()` can also read past `\r\n\r\n`, so initial response-body bytes can already be sitting in this application-owned buffer before origin/global capacity is checked.

A concrete failure case is an origin/global cap sized for one 64 KiB HTTP/1 response relay and N concurrent requests whose origin is slow to produce a complete response head. All N requests allocate 64 KiB here while the enforcing aggregate is still zero. With fast heads, each can additionally accept body bytes into that buffer before the later reservation refuses it. The cap therefore rejects *after* the memory peak it is supposed to prevent, directly violating #140's required property: **“Reservation before allocating or accepting queued bytes.”** The ALPN→HTTP/1 fallback in `streamViaH2Pool` has the same ordering (`alloc(h1_read_buf)` → `streamProxyOverTransport` → reserve after head parsing).

Please make HTTP/1 response ownership phase-specific as well. After the request/upload phase completes, but **before** allocating the response read/relay buffer and before calling `readUpstreamHead`, reserve the response footprint against the stream/origin/global policy; only then allocate and read. Keep that reservation/allocation scoped to each retry attempt and release/free it on every exit, including allocation failure and stale-connection retry. If you want bodiless responses not to hold a relay reservation, that requires a separately bounded header-read strategy; the current buffer cannot be considered unowned while it is being used to discover whether a body exists.

Please add a production-path regression analogous to the new H2 test: multiple HTTP/1 exchanges parked before a complete response head under a one-buffer origin/global cap, with a counting allocator proving allocated response relay memory never exceeds admitted capacity and no response-body bytes are accepted before reservation.

src/gateway_proxy.zig
```diff
@@ -4267,3 +4454,1484 @@ test "connectBlockingUnix + exchange round-trips a Unix-socket origin" {
...
+        self.mutex.lock();
+        defer self.mutex.unlock();
+        return self.peak_large_live;
+    }
+
+    fn alloc(ctx: *anyopaque, len: usize, alignment: std.mem.Alignment, ret_addr: usize) ?[*]u8 {
+        const self: *LargestAllocationTracker = @ptrCast(@alignCast(ctx));
+        if (self.fail_large and len >= self.large_threshold) return null;
+        const ptr = self.child.rawAlloc(len, alignment, ret_addr) orelse return null;
+        self.largest = @max(self.largest, len);
```
Jaruso: The new concurrent allocation test still has a data race in its tracker. `http2 relay memory is not allocated until a reservation admits it` shares one `LargestAllocationTracker` across three exchange threads. `large_live`/`peak_large_live` are protected by `mutex`, but every `alloc` writes `self.largest` here without that mutex; `resize` and `remap` do the same. The fact that this particular test reads `peakLargeLive()` rather than `largest` does not make concurrent unsynchronized writes to `largest` safe.

Please put `largest` under the same lock (or make it atomic), including resize/remap updates. For example, fold the max update into a small locked helper used by all three vtable operations. This is the same class of test-only race the latest reply correctly fixed by giving each exchange its own `UploadBufferObserver`; the allocator regression should be race-free too.

docs/PROXY_STREAMING.md
```diff
@@ -239,13 +291,74 @@ never leaves bytes reserved at the global scope.
...
+
+The per-stream policy — and the receive window derived from it —
 applies to connections opened after the reload: `SETTINGS_INITIAL_WINDOW_SIZE`
 is negotiated once per connection, and a peer already holding credit is still
 judged by what it was granted. Each connection therefore pins the complete
 low/high/hard policy it advertised, and every stream on it is measured against
 that, never against a newer snapshot.
 
+That is *every* per-stream decision, not just the queue: the shared per-stream
+budget and the response relay buffer are sized from the connection's pinned
```
Jaruso: This is stale after `44d75d3d`: there is no longer a shared per-stream budget object. The implementation now pins the relay size from the connection policy and subtracts that reserved headroom from the queue's own per-stream hard limit. Please describe that mechanism here, and update the matching “shared budget” wording in the CHANGELOG. Also, the CHANGELOG's earlier claim that the H2 relay change closes the *last* application-owned-buffer gap should wait until the HTTP/1 response allocation-before-reservation issue is fixed.

Jaruso (commented):
Full #140 ownership-lifecycle pass on `44d75d3d`.

The latest H2 fixes check out: request and response relay memory now follows reserve → allocate → use → free/release ordering, upload ownership ends at the phase boundary, allocator failures return the acquired connection reference, the connection-pinned policy is used consistently, and replacing the stack-shared stream budget with queue hard-limit headroom preserves `queue + relay <= per_stream_hard_limit` without a cross-thread lifetime hazard. The `H2ConnPool.evict` fetch-remove-before-free fix is also correct. CI #1321 is green, and I resolved the previous H2 thread.

I did find one remaining **production blocker** in the full protocol-wide pass: HTTP/1 response relay memory still allocates — and can accept initial response-body bytes while parsing the head — before the response-direction origin/global reservation exists. That is the same explicit #140 invariant we just fixed on H2 (`reservation before allocating or accepting queued bytes`), and means the new HTTP/1 per-origin/global limits do not actually bound the memory peak they claim to bound. The ALPN→H1 fallback has the same lifecycle. Detailed fix shape is in the inline comment.

There is also a race in the new concurrent allocation regression itself: the shared tracker protects `large_live`/`peak_large_live` but still writes `largest` unsynchronized from every exchange thread. Please make the tracker fully thread-safe so the regression proving the ordering invariant does not introduce undefined behavior of its own.

Finally, the docs/CHANGELOG still contain wording from the removed shared-stream-budget implementation and the CHANGELOG's “last gap” claim is premature until the HTTP/1 lifecycle is corrected.

So: the H2 work from this round is accepted, but I still would **not merge / close #140** until the HTTP/1 response reserve-before-allocation gap and the concurrent test race are fixed.

src/gateway_proxy.zig
Jaruso: Correct, and I should have applied the ordering rule to both paths in the same round instead of fixing H2 and leaving H1 with the identical defect. Fixed in 4522d08d.

`streamProxyOverTransport` now takes `relay_bytes: usize` instead of a buffer. After `sendStreamingProxyRequest` returns (so the upload's reservation is already back), it reserves the response footprint against stream/origin/global, and only then allocates and calls `readUpstreamHead`. Per attempt, so a stale-connection retry re-reserves rather than carrying the previous attempt's claim, and released on every exit including allocation failure. The ALPN→HTTP/1 fallback goes through the same function, so it inherits this rather than needing its own copy — `executeStreamingHttpProxyRequest` no longer allocates a relay buffer at all.

**No bodiless exemption on this path**, per your point: the buffer is what discovers whether a body exists, so it can't be treated as unowned while that question is open. HTTP/2 keeps its exemption because there the head arrives before the buffer is needed.

Both refusal modes return `ProxyBufferCapacityUnavailable` — per-stream and aggregate alike are local capacity before commitment, so both are a clean 503 rather than a 502 charged to the origin.

**Regression test:** `http1 response memory is not allocated until a reservation admits it`, the direct analogue of the H2 one. Three exchanges against an origin that sends `HTTP/1.1 200 OK\r\n` and stops — an incomplete head, so an admitted relay parks inside `readUpstreamHead` holding its reservation — under an origin cap sized for one buffer, with one shared counting allocator across all three. It asserts peak relay-sized live allocation is exactly one buffer (three before the reordering), that the two refusals are `ProxyBufferCapacityUnavailable` with nothing written downstream, and that every scope and the tracker drain to zero afterwards.

That test replaces `concurrent http1 responses to one origin cannot exceed its buffer cap`, which asserted a strict subset of it — the cap holding under concurrency — without the allocation evidence.

src/gateway_proxy.zig
Jaruso: Right — and it's the same class I'd just fixed one layer over, which makes missing it worse rather than better. Fixed in 4522d08d.

Every field update now goes through one locked helper: `noteAlloc` takes the lock and updates `largest` and the live/peak fields together, `noteFree` takes it for the live decrement, and `alloc`/`resize`/`remap`/`free` all route through them — `resize` and `remap` now also account the old length they replace, which they were previously ignoring. Reads go through `largestAllocation()` and `peakLargeLive()`, both locked.

Your framing is the right one: "this test happens to read the other field" is not a reason for a write to be unsynchronized. Worth noting the shared tracker is now genuinely shared on purpose — the new HTTP/1 test uses one across three exchange threads to measure process-wide relay memory, which only works because every update is under the lock.

docs/PROXY_STREAMING.md
Jaruso: Both stale spots corrected in 4522d08d.

`docs/PROXY_STREAMING.md` reload section now says the *headroom held back for the relay buffer* — and the buffer itself — are sized from the connection's pinned policy, instead of referring to a shared budget object that no longer exists. (The response-bounds section a few paragraphs up already described the headroom mechanism; this paragraph was the one I missed.)

CHANGELOG: the "shared budget" phrasing in the connection-pinning paragraph is now "relay headroom", and the claim that the HTTP/2 relay change closed the *last* application-owned-buffer gap is gone — you were right that it couldn't stand while HTTP/1 still allocated before reserving. The allocation-ordering paragraph now covers both protocols and spells out why HTTP/1 has no bodiless exemption.

src/gateway_proxy.zig
```diff
@@ -2189,6 +2358,34 @@ fn streamProxyOverTransport(
...
+        .upstream_to_downstream,
+        proxy_buffer_limits,
+        proxy_buffer_observer,
+        proxy_buffer_capacity,
+    );
+    defer response_reservation.releaseAll();
+    // Either refusal is local capacity before anything is committed
+    // downstream, so both are the caller's clean 503 rather than a status
+    // blamed on the origin.
+    response_reservation.reserve(relay_bytes) catch return error.ProxyBufferCapacityUnavailable;
```
Jaruso, line 2385: Verified on re-review: this is now the correct #140 ownership boundary. The response reservation is committed before the relay allocation on the next line and before `readUpstreamHead` can accept any response bytes; defer ordering also keeps the reservation live until after the buffer is freed.

Jaruso (commented):
Re-reviewed the delta from `44d75d3d` through `4522d08d` and rechecked the affected paths against #140's acceptance criteria.

No blocking code findings remain.

The HTTP/1 response path now has the ownership ordering #140 requires: the upload phase has released its claim, response capacity is reserved before the relay allocation exists, allocation happens only after admission, and `readUpstreamHead` cannot accept response/body bytes until both are in place. The reservation/allocation are attempt-scoped, so stale-connection retry, allocation failure, timeout, upstream abort, and client abort all unwind cleanly. The ALPN→HTTP/1 fallback uses the same path, so it no longer has a separate ordering gap.

I also verified the concurrent allocation tracker no longer contains the unsynchronized `largest` writes, and the stale shared-budget wording has been corrected to describe the current reserved-headroom design. All review threads are now resolved.

From a code/acceptance-criteria perspective, this is merge-ready and #140 can close with it. I cannot submit an APPROVE review because the authenticated GitHub user is the PR author, so this is a COMMENT review instead.

CI run #1324 is still in progress at the time of this review. Every completed job is green; the remaining `Test appliance profile (ubuntu-latest)` job was still running. Merge once that run finishes green.

Issue #593, opened 2026-08-09
perf: execute competitive benchmark suite and document results


Jaruso:
## Summary
Run the competitive benchmark suite added by #589 against Tardigrade, NGINX, HAProxy, and Caddy in a controlled environment, **run the accept-batching/fairness benchmark suite added by #601 for #146**, and **run the event-loop backend benchmark/profiling suite required by #148 against Tardigrade's default `epoll` backend and the opt-in `io_uring` prototype landed by #602**. Archive the raw outputs, document the results, and turn any material findings into focused follow-up performance work.

This is the validation follow-up to #149 / #589, the execution home for the remaining benchmark acceptance work from #146 / #601, and now also the real-hardware evidence pass for #148 / #602. The benchmark harnesses and backend prototype can be merged independently, but the full non-smoke four-server suite, the accept-batching suite, and the `epoll` vs `io_uring` comparison still need to be executed end-to-end on suitable real hardware.

## Goals
- Prove that the complete competitive suite works against all four servers, not only the reduced Tardigrade smoke path.
- Produce a reproducible baseline for future performance work.
- Document where Tardigrade wins, loses, or is not directly comparable.
- Record the upstream-pool contention/sharding conclusion from measured high-core results rather than leaving the harness unexercised.
- Validate the accept-batching implementation from #601 with real measurements rather than inferred behavior.
- Determine whether the default accept batch size improves connection-churn throughput without regressing mixed-traffic tail latency.
- Verify that the configured fairness control is actually exercised under load and that listener-shard fairness remains acceptable where sharding is supported.
- Measure whether the readiness-only `io_uring` backend from #602 provides a material improvement over the default Linux `epoll` backend for Tardigrade's actual shared-`EventLoop` workloads.
- Determine whether any `io_uring` delta comes from reduced syscalls/wakeups or simply measurement noise, and whether the shared event loop is a measurable bottleneck at all.
- Compare `epoll` vs `io_uring` for throughput, tail latency, CPU/request, syscalls/request, and wakeups/request under the workloads required by #148.
- Produce an evidence-based recommendation to retain the `io_uring` prototype as opt-in, remove it if it provides no meaningful benefit, or open focused follow-up work if the measured improvement justifies the added Linux-specific maintenance cost.
- Turn any material regressions or gaps discovered by the run into focused follow-up issues.

## Test environment
Use a dedicated, otherwise-idle machine where practical. Before running, capture and retain the suite metadata for:
[...]

PR #601, opened 2026-08-10
perf: add accept batching controls

### LINKED ISSUE #146: perf: add accept batching and fairness controls
## Summary
Add tunable accept batching and fairness controls so Tardigrade can drain bursts efficiently without starving other workers, listener shards, or active connections.

## Research basis
HAProxy exposes accept and listener sharding controls because batching and fairness can materially affect throughput and tail latency. High-performance servers often batch readiness work, but uncontrolled batching can create unfairness.

## Proposed design
- Add an accept batch limit per loop/shard.
- Track how many connections are accepted per loop iteration.
- Add fairness/yield behavior after a configurable number of accepts.
- Ensure overload handling still rejects early when worker queues or connection limits are saturated.
- Make defaults conservative.
- Integrate with listener sharding if #137 is implemented.

## Acceptance criteria
- Accept loop can drain multiple ready accepts per iteration up to a configured cap.
- Metrics expose accept batch sizes and fairness yields.
- Benchmark shows whether batching improves throughput under connection churn.
- Tail latency does not regress under mixed traffic.
- Defaults are documented.

## Metrics to add or verify
- `tardigrade_accept_batch_size` histogram
- `tardigrade_accept_batches_total`
- `tardigrade_accept_fairness_yields_total`
- `tardigrade_accept_rejected_due_to_worker_queue_total`

## Benchmark requirements
Run:
- connection churn benchmark with keepalive disabled
- burst benchmark
- mixed active keepalive + new connection benchmark

Compare req/s, p99/p999, accept errors, worker queue latency, and shard fairness if applicable.

## Complexity
Small to medium.

## Risk
Medium. Too-aggressive batching can improve throughput while harming fairness and tail latency.

## Dependencies
Depends on #136. Related to #137.

Jaruso:
**What changed and why**
Adds bounded accept batching controls so listener readiness turns can drain bursts without unbounded monopolization. The new env knobs are `TARDIGRADE_ACCEPT_BATCH_LIMIT` (default `64`) and `TARDIGRADE_ACCEPT_FAIRNESS_YIELD_EVERY` (default `0`, disabled).

Also adds Prometheus metrics for accept batch sizes, accepted batches, and fairness yields, and extends the listener-sharding benchmark parser to include those series.

**Related issues**
Closes #146

**Tests**
- Added/updated metrics unit coverage for `tardigrade_accept_batch_size`, `tardigrade_accept_batches_total`, and `tardigrade_accept_fairness_yields_total`.
- Updated listener-sharding integration assertions to verify the new metrics are exported per shard.

Validation run locally:
- `zig fmt src/edge_config.zig src/gateway_accept.zig src/gateway_state.zig src/edge_gateway.zig src/http/metrics.zig tests/integration.zig`
- `zig build test --summary all --error-style verbose` passed: 3378 passed, 1 skipped.
- `bash -n benchmarks/listener-sharding.sh` passed.
- `git diff --check` passed.
- `zig build test-integration` did not complete: repeated failure in existing `reload while serving short static requests drops no in-flight request (#170)` after 84 passed / 65 skipped, with `hammer.fail` nonzero and upstream `SocketUnconnected` / `InvalidHttpResponse` errors. The failing test is not the listener-sharding accept-metrics test changed here.

**Security impact**
Affects the accept hot path only. Overload handling still performs connection-slot checks before worker submission and still returns the deterministic accept-time `503` for global, per-IP, memory, and worker-queue saturation.

**Performance impact**
Affects the listener accept loop. Expected impact is better throughput under connection churn when batching is raised, with bounded fairness controls and metrics to watch tail-latency/fairness tradeoffs.

**Checklist**
- [ ] `zig fmt --check build.zig src/ tests/` passes
- [ ] `zig build test-integration` green (required for protocol, proxy, and TLS changes)


src/gateway_accept.zig
```diff
@@ -82,6 +106,7 @@ pub fn acceptReadyConnectionsShard(listen_fd: std.posix.fd_t, shard_id: u16, wor
         }
 
         state.metricsRecordAccept(shard_id);
+        accepted_this_batch += 1;
```
Jaruso, line 119: `accepted_this_batch` now counts this socket, but the `tryAcquireConnectionSlot(...) catch` a few lines below closes the fd and `continue`s without ever calling `acceptBatchShouldYield()`. Repeated tracking/allocation failures can therefore drain more than both configured limits in one readiness turn. Please make every path after a successful `accept()` honor the batch decision. For example:

```zig
const slot_result = state.tryAcquireConnectionSlot(client_fd, ip_key) catch |err| {
    state.logger.warn(null, "connection slot tracking error: {}", .{err});
    _ = std.c.close(client_fd);
    if (acceptBatchShouldYield(state, shard_id, accepted_this_batch, batch_options, &recorded_batch)) return;
    continue;
};
```

Even better, structure the loop so there is one common post-accept batch check and exceptional branches cannot accidentally bypass it.

benchmarks/listener-sharding.sh
```diff
@@ -622,7 +659,7 @@ jq -n \
             keepalive_path: $keepalive_path,
             save_dir: $save_dir,
             workloads: ["static-http1", "proxy-http1", "keepalive", "connection-churn-http1", "mixed-short-static-proxy"],
-            metrics: ["req/s", "p50_ms", "p95_ms", "p99_ms", "p999_ms", "cpu_pct_avg", "rss_mb_peak", "accept_errors_total", "worker_queued_jobs", "worker_queue_wait_us_count", "worker_queue_wait_us_sum", "per-shard accepts_total"]
+            metrics: ["req/s", "p50_ms", "p95_ms", "p99_ms", "p999_ms", "cpu_pct_avg", "rss_mb_peak", "accept_errors_total", "accept_batch_size_count", "accept_batch_size_sum", "accept_batches_total", "accept_fairness_yields_total", "worker_queued_jobs", "worker_queue_wait_us_count", "worker_queue_wait_us_sum", "per-shard accepts_total"]
```
Jaruso, line 816: Adding the batch series to the output is useful, but this benchmark still cannot establish the #146 batching acceptance criteria. `run_profile` only varies `TARDIGRADE_LISTENER_SHARDS`; it never defines a baseline/default/fairness batching profile. The workload list also has no burst case, and `mixed-short-static-proxy` is run with `Connection: close`, so it is not the required mixed **active keepalive + new connection** workload. Please extend the benchmark to vary `TARDIGRADE_ACCEPT_BATCH_LIMIT` / `TARDIGRADE_ACCEPT_FAIRNESS_YIELD_EVERY`, add the missing workloads, and preserve those settings/results in the summary so req/s and p99/p999 can be compared.

tests/integration.zig
```diff
@@ -14673,6 +14673,9 @@ test "listener sharding starts real gateway with reuse-port shards and exports p
 
         try std.testing.expect(prometheusLabeledMetricValue(metrics.body, "tardigrade_accept_errors_total", &.{ shard_label, "reason=\"poll\"" }) != null);
         try std.testing.expect(prometheusLabeledMetricValue(metrics.body, "tardigrade_accept_errors_total", &.{ shard_label, "reason=\"accept\"" }) != null);
+        try std.testing.expect(prometheusLabeledMetricValue(metrics.body, "tardigrade_accept_batch_size_count", &.{shard_label}) != null);
```
Jaruso, line 14830: This verifies that the metric exists, but none of the added coverage verifies the new accept behavior itself. Please add a focused test that queues more sockets than the configured limit and asserts one readiness turn accepts only `limit`, then a later turn drains the remainder; likewise exercise `fairness_yield_every` and assert the fairness counter increments. Without that, regressions in the boundedness logic (including the error-path bypass noted above) remain invisible to the suite.

Jaruso (commented):
Blocking findings before merge. The core shape is good: the new config is wired through both the main listener and sharded accept paths, the Prometheus histogram/counters are bounded per shard, overload rejection remains before worker submission, and the defaults are documented. CI is also green on this head.

Issue #146 is not fully satisfied yet:

1. **The configured cap/fairness limit can be bypassed after a successful `accept()`.** `accepted_this_batch` is incremented before connection-slot tracking. If `tryAcquireConnectionSlot()` returns an error, that catch branch closes the fd and immediately `continue`s without calling `acceptBatchShouldYield()`. Under repeated allocator/state-tracking errors, a single readiness turn can therefore accept past both `TARDIGRADE_ACCEPT_BATCH_LIMIT` and `TARDIGRADE_ACCEPT_FAIRNESS_YIELD_EVERY`, which defeats the boundedness this PR is introducing. The inline comment has a concrete fix shape.

2. **The benchmark acceptance criteria are still missing.** #146 requires a connection-churn comparison, a burst benchmark, and mixed **active keepalive + new connection** traffic, with req/s, p99/p999, accept errors, queue latency, and fairness compared. This PR extends the listener-sharding parser, but the benchmark still varies listener shard count rather than batching settings; its workload list has no burst case; and `mixed-short-static-proxy` explicitly sends `Connection: close`, so it does not keep active keepalive traffic alive while new connections churn. The PR description also contains no batching baseline/default/fairness results, so the acceptance criteria “benchmark shows whether batching improves throughput” and “tail latency does not regress under mixed traffic” cannot currently be verified.

A minimal direction would be to make batching part of the benchmark profile, for example:

```bash
run_profile() {
    local label="$1"
    local shard_count="$2"
    local batch_limit="$3"
    local fairness_yield_every="$4"
    # ...
    export TARDIGRADE_LISTENER_SHARDS="$shard_count"
    export TARDIGRADE_ACCEPT_BATCH_LIMIT="$batch_limit"
    export TARDIGRADE_ACCEPT_FAIRNESS_YIELD_EVERY="$fairness_yield_every"
    # ...
}

baseline_json="$(run_profile "batch-1" 1 1 0)"
default_json="$(run_profile "batch-64" 1 64 0)"
fair_json="$(run_profile "batch-64-fair-8" 1 64 8)"
```

Then add an explicit burst case and a mixed case that runs sustained keepalive traffic concurrently with connection-churn traffic instead of forcing every mixed request closed. Record the batch settings in summary metadata and include the measured comparison in the PR (or a checked-in benchmark artifact) so throughput and p99/p999 regressions are reviewable.

3. **Behavioral coverage should exercise the batching controls themselves.** The added tests verify that the new metrics are emitted, but they do not drive a ready listener and prove that `limit=N` accepts no more than N sockets in one turn, that `fairness_yield_every` ends a turn and increments its counter, or that a subsequent readiness turn drains what remains. That gap is why the error-path escape above passes the current suite. I would add focused accept-loop coverage before marking “New behavior covered by tests” complete.

Once the cap is enforced on every post-accept path and the ticket’s benchmark/behavioral validation is added, I don’t see another blocker in the metrics/config/docs portions of this diff.

Note: GitHub does not allow the connected author account to submit a formal `REQUEST_CHANGES` review on its own PR, so this is posted as a COMMENT review; the findings above are still blockers from this review.

benchmarks/listener-sharding.sh
```diff
@@ -578,19 +653,27 @@ EOF
...
+        --argjson batch_limit "$batch_limit" \
+        --argjson fairness_yield_every "$fairness_yield_every" \
+        '{label: $label, requested_shards: $shards, actual_shards: $metrics.listener_shards, accept_batch_limit: $batch_limit, accept_fairness_yield_every: $fairness_yield_every, artifacts: {standard_json: $run_json, standard_log: $run_log, startup_metrics_prom: $startup_metrics_file, metrics_prom: $metrics_file, connection_churn_wrk: $churn_raw, connection_churn_queue_samples: $churn_queue_samples, connection_burst_wrk: $burst_raw, connection_burst_queue_samples: $burst_queue_samples, mixed_keepalive_churn_keepalive_wrk: $mixed_keepalive_raw, mixed_keepalive_churn_churn_wrk: $mixed_churn_raw, mixed_keepalive_churn_queue_samples: $mixed_queue_samples}, standard: $standard, "connection-churn-http1": $churn, "connection-burst-http1": $burst, "mixed-keepalive-churn": $mixed, metrics: $metrics}'
 
     cleanup_gateway
 }
 
-single_json="$(run_profile "single" 1)"
-sharded_json="$(run_profile "sharded" "$SHARDS")"
+batch_1_json="$(run_profile "batch-1" 1 1 0)"
```
Jaruso: The benchmark matrix itself now matches the requested #146 scenarios, but the ticket says to **run** them and compare the resulting throughput/tail latency. I don't see a `summary.json` or measured results attached anywhere on the PR yet. Please run these profiles on the same host/build and include at least connection-churn/burst req/s, mixed keepalive+churn p99/p999, accept errors, worker-queue wait, and fairness/shard distribution. Without those measurements we still cannot verify the acceptance criteria that batching improves churn throughput and does not regress mixed-traffic tail latency.

Jaruso (commented):
Re-review of `a30f6e0` against issue #146 and the previous review.

The implementation feedback from the first pass is addressed:

- **Cap/fairness bypass: fixed.** The `tryAcquireConnectionSlot()` error path now runs `acceptBatchShouldYield()` after a successful `accept()`, so every post-accept exit/continue path honors the configured bound.
- **Behavioral coverage: fixed.** The new socket-level tests prove that `limit=N` bounds one readiness turn, a subsequent turn drains the remainder, and `fairness_yield_every` terminates the turn and increments the fairness counter.
- **Benchmark harness shape: fixed.** The script now runs distinct `batch-1`, `batch-64`, `batch-64-fair-8`, and sharded profiles; adds a high-connection burst workload; runs sustained keepalive traffic concurrently with connection churn; and records req/s, p99/p999, accept metrics, queue metrics, and profile settings in the summary.

I did not find another code-level blocker in the new commit.

**One acceptance item is still outstanding before #146 can be closed:** the issue explicitly requires the benchmarks to be *run* and requires the result to establish both whether batching improves throughput under connection churn and that tail latency does not regress under mixed traffic. The PR description/comments still contain no measured `batch-1` vs `batch-64` vs fairness results or checked-in benchmark artifact, so those two acceptance criteria remain unverifiable.

Please run the new suite on the same host/build for all profiles and attach the resulting `summary.json` (or summarize the relevant numbers in the PR): connection-churn req/s, burst req/s, mixed keepalive/churn p99 and p999, accept errors, worker queue wait, and fairness/shard distribution. If the default `64` profile materially regresses mixed-traffic tail latency, the default should be adjusted before merge; otherwise the measurements close the remaining acceptance gap.

Current CI for `a30f6e0` is still in progress. Several build/unit/security jobs have already passed, but the overall run is not complete yet.

As before, GitHub will not allow the connected author account to submit `REQUEST_CHANGES` on its own PR, so this is a COMMENT review; the missing benchmark evidence is still a merge blocker for issue #146 acceptance.

Jaruso (commented):
Re-review of the current head `a30f6e0`.

There are no new commits or PR comments since the prior review, so the code state is unchanged. I rechecked the prior findings and thread state:

- The post-accept cap/fairness bypass is fixed; that thread is now resolved.
- The focused socket-level batching/fairness tests are present and cover the requested behavior; that thread is now resolved.
- The benchmark harness now varies batch/fairness settings and includes churn, burst, and concurrent keepalive + new-connection workloads; that older harness-shape thread is now resolved.
- CI run 1323 has now completed successfully across build, unit, appliance/integration, lint, security, packaging, and perf-smoke jobs.

One blocker remains open: the actual #146 accept-batching benchmark results still have not been produced or attached. I also checked the completed CI perf-smoke job: it runs `./benchmarks/ci-smoke.sh` and only measures the generic static/proxy/keepalive scenarios. It does not execute `benchmarks/listener-sharding.sh`, does not compare `batch-1` vs `batch-64` vs `batch-64-fair-8`, and therefore cannot establish the issue's required churn-throughput or mixed-traffic p99/p999 result.

So the current status is: **implementation looks merge-ready, CI is green, but issue #146 acceptance is still incomplete until the new benchmark matrix is actually run and the measurements are attached/summarized.** The existing open inline thread at `benchmarks/listener-sharding.sh:673` remains the only blocker; I am not adding a duplicate inline comment for the same finding.

benchmarks/results/2026-08-10/accept-batching-local-fallback-final/README.md
```diff
@@ -0,0 +1,55 @@
...
+  --connections 30 \
+  --burst-connections 120 \
+  --threads 4 \
+  --static-path /health \
+  --proxy-path /proxy/health \
+  --keepalive-path /health \
+  --save-dir benchmarks/results/2026-08-10/accept-batching-local-fallback-final
+```
+
+`summary.json` is the authoritative machine-readable artifact. The raw wrk
```
Jaruso: This says `summary.json` is authoritative, but the committed `summary.json` is a zero-byte file. Several `sharded-batch-64-*` artifacts are also zero-byte while the table below still reports sharded churn/mixed values. Please only commit a completed, self-consistent run: generate the table from the final `summary.json` (ideally with a script) and fail the evidence-generation step if the summary or any required profile artifact is empty.

benchmarks/results/2026-08-10/accept-batching-local-fallback-final/batch-64-fair-8-metrics.prom
```diff
@@ -0,0 +1,548 @@
...
+tardigrade_accept_batch_size_bucket{shard="0",le="1"} 538424
+tardigrade_accept_batch_size_bucket{shard="0",le="2"} 538424
+tardigrade_accept_batch_size_bucket{shard="0",le="4"} 538424
+tardigrade_accept_batch_size_bucket{shard="0",le="8"} 538424
+tardigrade_accept_batch_size_bucket{shard="0",le="16"} 538424
+tardigrade_accept_batch_size_bucket{shard="0",le="+Inf"} 538424
+tardigrade_accept_batch_size_sum{shard="0"} 538424
+tardigrade_accept_batch_size_count{shard="0"} 538424
+tardigrade_accept_batches_total{shard="0"} 538424
+tardigrade_accept_fairness_yields_total{shard="0"} 0
```
Jaruso: This profile never exercised the fairness control—or batching at all. `accepts_total`, `accept_batches_total`, `accept_batch_size_count`, and `accept_batch_size_sum` are all exactly `538424`, and the `le="1"` bucket already contains all 538424 samples. So every readiness turn accepted exactly one connection and `fairness_yield_every=8` could never fire. A performance comparison between this profile and batch-1/batch-64 cannot tell us anything about batching. Please rerun with enough simultaneous pending connections to produce batches >1 and require `fairness_yields_total > 0` for the fairness profile before treating the run as valid.

benchmarks/results/2026-08-10/accept-batching-local-fallback-final/batch-64-connection-burst.wrk.txt
```diff
@@ -0,0 +1,16 @@
...
+  Thread Stats   Avg      Stdev     Max   +/- Stdev
+    Latency     0.00us    0.00us   0.00us     nan%
+    Req/Sec     0.00      0.00     0.00       nan%
+  Latency Distribution
+     50%    0.00us
+     75%    0.00us
+     90%    0.00us
+     99%    0.00us
+  0 requests in 10.03s, 0.00B read
+  Socket errors: connect 120, read 0, write 0, timeout 0
```
Jaruso: This burst sample is invalid: it completed **0 requests** and all 120 connections failed to connect. Because wrk still exits successfully, the harness currently promotes a failed client workload into the summary. Please make `run_wrk_workload` reject samples with zero throughput or non-zero socket errors (at minimum for these acceptance runs), and add recovery/isolation between the preceding connection-churn workload and burst so client ephemeral-port/TIME_WAIT exhaustion cannot dominate the next scenario.

Jaruso (commented):
Re-review of `23be944` against issue #146 and the prior benchmark-evidence blocker.

The implementation itself is still in good shape and CI run 1329 is green. The new commit does add benchmark artifacts, but I still cannot accept them as evidence for the performance criteria.

There are three blocking problems with this run:

1. **The run never exercised accept batching.** In `batch-64-fair-8-metrics.prom`, `accepts_total == accept_batches_total == accept_batch_size_count == accept_batch_size_sum == 538424`, and the entire histogram is already contained in `le="1"`; `tardigrade_accept_fairness_yields_total` is also `0`. The plain `batch-64` profile has the same pattern (`425372` accepts / `425372` batches, all size 1). That means every readiness turn accepted exactly one socket. Therefore the batch-1, batch-64, and fairness profiles did not differ in actual batching behavior, so their req/s and latency differences cannot establish the effect of this feature.

2. **The burst/mixed phases are client-failure dominated.** The batch-64 burst performed 0 requests and reported 120 connect errors; the mixed churn side performed only 1 request and 28 connect errors. The keepalive side also reports hundreds of socket errors. These are not usable p99/p999 or throughput comparisons. Given the workload order (high close-connection churn immediately followed by burst/mixed), this strongly looks like the client/host exhausting sockets/TIME_WAIT rather than measuring Tardigrade. The harness should reject such a sample instead of writing it into the result table.

3. **The committed evidence is incomplete/inconsistent.** `README.md` says `summary.json` is authoritative, but `summary.json` is empty. Several sharded-profile raw artifacts are also empty, while the README still contains sharded churn/mixed numbers that cannot be reproduced from those committed files. The README itself correctly notes that macOS fell back to one effective shard, so this also provides no shard-fairness result.

Before merge, rerun on a host/workload that actually creates queued accepts and validate the run before accepting it. I would add explicit benchmark guards along these lines:

```bash
# Reject workloads dominated by client errors / no useful sample.
if ! jq -e '.rps > 0 and .errors == 0' <<<"$summary_json" >/dev/null; then
    echo "${label}: invalid benchmark sample (zero throughput or client errors)" >&2
    return 1
fi

# For batch profiles, prove batching was actually exercised.
accepted="$(jq '[.accept_metrics.accept_batch_size_sum_delta[].value] | add // 0' <<<"$result")"
batches="$(jq '[.accept_metrics.accept_batch_size_count_delta[].value] | add // 0' <<<"$result")"
if [[ "$batch_limit" -gt 1 ]] && (( accepted <= batches )); then
    echo "${label}: no readiness turn accepted more than one connection; batching was not exercised" >&2
    return 1
fi

# For the fairness profile, require at least one actual fairness cutoff.
if [[ "$fairness_yield_every" -gt 0 ]]; then
    yields="$(jq '[.accept_metrics.accept_fairness_yields_total_delta[].value] | add // 0' <<<"$result")"
    (( yields > 0 )) || { echo "${label}: fairness cutoff was never exercised" >&2; return 1; }
fi
```

Also insert recovery between close-connection workloads (or isolate them into fresh client/network namespaces / use a host with enough ephemeral-port headroom) so the burst and mixed phases are not inheriting socket exhaustion from the churn phase.

Once we have a completed `summary.json` from a valid run where batch-64 actually observes batches >1, the fairness profile records real yields, and the burst/mixed workloads complete without client socket failures, the remaining #146 acceptance criteria can be evaluated. As before, GitHub does not allow the connected author account to submit a formal REQUEST_CHANGES review on its own PR, so this is posted as COMMENT, but these findings are merge blockers.

benchmarks/listener-sharding.sh
```diff
@@ -578,55 +725,81 @@ EOF
...
```
Jaruso: These `run_profile` calls execute the entire function inside command substitution. In Bash, `set -e` is normally cleared in `$(...)`, so the new validators inside `run_profile`/`run_wrk_workload` do **not** reliably abort when they return 1. For example, `validate_wrk_sample ...` can fail, then `run_wrk_workload` continues and emits JSON; likewise `validate_profile_accept_batching` can fail and `run_profile` can continue through its final `jq` and `cleanup_gateway`, returning success. That means the exact zero-error/batching/fairness guards added in this commit can still admit an invalid run.

Please propagate failures explicitly instead of relying on `errexit` across command substitution, e.g.:

```bash
summary_json="$(wrk_summary_json_from_file "$label" "$output")" || return 1
validate_wrk_sample "$label" "$summary_json" "$output" || return 1

churn_json="$(run_wrk_workload ...)" || return 1
burst_json="$(WRK_CONNECTIONS="$BURST_CONNECTIONS" run_wrk_workload ...)" || return 1
mixed_json="$(run_mixed_keepalive_churn_workload ...)" || return 1
validate_profile_accept_batching ... || return 1
validate_profile_artifacts ... || return 1
```

Then make the top-level captures explicit as well (`batch_1_json="$(run_profile ...)" || exit 1`, etc.), or otherwise restructure profiles to write results without wrapping the whole function in `$(...)`. `shopt -s inherit_errexit` could also address this on newer Bash versions, but explicit propagation is safer for the script's macOS portability.

Jaruso (commented):
Re-review of `c87bea3` against issue #146 and the previous benchmark-evidence findings.

The latest commit moves the benchmark harness in the right direction:

- The invalid macOS fallback evidence has been removed rather than retained as acceptance data.
- Workload validation now intends to reject zero-throughput or client/socket-error samples.
- Recovery pauses were added between close-heavy workloads.
- Profile validation now intends to require actual batch sizes >1 and a real fairness yield for the fairness profile.
- Required artifacts and the final summary are checked for non-empty/valid JSON output.
- CI run 1332 is green.

However, there is one new blocking harness bug: **the validation failures can be swallowed because every `run_profile` invocation executes inside command substitution.** Bash normally disables `errexit` inside `$(...)`, so `set -e` at the top of the script does not make a failed `validate_wrk_sample`, `validate_profile_accept_batching`, or `validate_profile_artifacts` abort `run_profile`. The function can continue after a validation returns 1, emit JSON, run `cleanup_gateway`, and ultimately return success. I reproduced the shell behavior with a minimal `set -e; x="$(func_with_failed_validation)"` case: the failed validation was ignored and the assignment exited 0. The inline comment has a concrete fix shape.

This matters directly to the previous review: the new code is specifically supposed to prevent another zero-request / zero-fairness run from being accepted, but in its current form those guards are advisory stderr messages rather than hard gates.

Separately, issue #146 still requires an actual valid benchmark result: it explicitly requires measured connection-churn throughput and mixed-traffic tail latency, plus the burst workload and comparison metrics. There is currently no replacement `summary.json` or PR comment containing a valid run after the bad evidence was removed. So once the error propagation is fixed, the suite still needs to be run on a host/load that actually exercises batches >1 and fairness yields, with zero client workload errors, and those measurements need to be attached/summarized before #146 can be closed.

I did not find a new defect in the accept-loop/config/metrics/test implementation itself. The remaining work is benchmark correctness + valid evidence. As before, GitHub will not allow the connected author account to submit `REQUEST_CHANGES` on its own PR, so this is a COMMENT review; the finding is still blocking for #146 acceptance.

benchmarks/listener-sharding.sh
```diff
@@ -457,43 +567,104 @@ run_wrk_workload() {
...
         export TARDIGRADE_RATE_LIMIT_RPS="0"
         export TARDIGRADE_PID_FILE="${pid_file}"
         export TMPDIR="${profile_tmp}"
         exec sh -c "$START_COMMAND"
     ) &
     gateway_pid="$!"
     echo "$gateway_pid" > "$pid_file"
-    wait_for_ready
-    verify_listener_shards "$shard_count" "$startup_metrics_file"
+    wait_for_ready || return 1
```
Jaruso: Now that failures are explicitly propagated, every `|| return 1` after the gateway starts can bypass the only normal `cleanup_gateway` call at the bottom of `run_profile`. Because this whole function is invoked inside `$(run_profile ...)`, the `gateway_pid` assignment happens in the command-substitution subshell; the parent shell's EXIT cleanup cannot see that PID, and Bash does not reliably preserve the parent's caught EXIT trap as a cleanup handler for this child context. A validation failure can therefore leave `tardi` running/bound to the port.

Please make cleanup unconditional on every profile exit. One straightforward shape is to centralize failure cleanup:

```bash
profile_fail() {
    local rc="${1:-1}"
    cleanup_gateway
    return "$rc"
}

wait_for_ready || { rc=$?; profile_fail "$rc"; return "$?"; }
verify_listener_shards "$shard_count" "$startup_metrics_file" || { rc=$?; profile_fail "$rc"; return "$?"; }
# ...same for workload/validation failures...
```

Or better, structure `run_profile` so all work records an `rc`, then executes `cleanup_gateway` exactly once before returning that status. The key requirement is that a deliberately rejected benchmark sample cannot orphan the gateway process.

Jaruso (commented):
Re-review of `bc5d1a6` against issue #146 and the previous harness finding.

The prior `errexit`/command-substitution bug is fixed. The new commit explicitly propagates failures through `run_wrk_workload`, the mixed workload, `run_profile`, and the top-level profile captures, so zero-throughput/socket-error samples, missing artifacts, lack of actual batching, and lack of fairness yields can no longer be silently accepted. CI run 1334 is also green.

I found one new lifecycle problem introduced by those early returns: **a failed profile can leave its gateway process running.** `run_profile` still executes inside `$(...)`, so its `gateway_pid` assignment lives in the command-substitution subshell. The script's top-level `trap cleanup_gateway EXIT` does not provide reliable cleanup for that child context, and the parent shell cannot clean the gateway afterward because its own `gateway_pid` remains unchanged. Any of the new `|| return 1` paths after the gateway starts can therefore exit `run_profile` before the normal `cleanup_gateway` at the end. The inline comment has a concrete fix shape.

This matters especially now because the validation paths are deliberately expected to fail bad benchmark runs. A rejected batch/fairness sample should leave the host clean enough to correct the workload and rerun immediately, rather than leaving `tardi` bound to the benchmark port.

Separately, the final issue #146 performance acceptance criteria are still outstanding. The previously invalid benchmark evidence was correctly removed, but there is still no replacement valid `summary.json` or PR comment with measured results. After fixing failure cleanup, the suite still needs a valid run demonstrating:

- `batch-64` actually observes batches > 1;
- the fairness profile records real fairness yields;
- churn/burst/mixed workloads complete without client/socket errors;
- connection-churn req/s is compared across batch profiles;
- mixed keepalive + churn p99/p999 does not regress materially;
- accept errors, worker-queue wait, and shard fairness are included where applicable.

I did not find another defect in the already-reviewed accept loop, configuration, metrics, docs, or behavioral tests. As before, GitHub will not allow the connected author account to submit `REQUEST_CHANGES` on its own PR, so this is a COMMENT review; cleanup correctness and the valid benchmark evidence remain blockers for closing #146.

src/gateway_accept.zig
```diff
@@ -50,15 +65,24 @@ pub fn runShardAcceptLoop(ctx: ShardAcceptContext) void {
         };
         if (ready == 0) continue; // timeout — loop and recheck shutdown flag
         if ((pfd[0].revents & std.posix.POLL.IN) != 0) {
-            acceptReadyConnectionsShard(ctx.listen_fd, ctx.shard_id, ctx.worker_pool, ctx.state);
+            acceptReadyConnectionsShard(ctx.listen_fd, ctx.shard_id, ctx.worker_pool, ctx.state, ctx.batch_options);
```
Jaruso: This is not an actual fairness *yield* in the sharded path. When `acceptReadyConnectionsShard()` stops because `fairness_yield_every` was reached, this dedicated thread immediately loops back into level-triggered `poll()` on the same listener. If the backlog is still non-empty, `poll()` returns immediately and this shard can start the next accept batch without giving worker/sibling threads an intentional scheduling opportunity. That means the fairness knob has real event-loop fairness in the single-listener path, but in sharded mode it is effectively just another batch boundary.

Please make the stop reason observable and perform an actual scheduler yield when fairness caused the stop. For example:

```zig
const AcceptTurnResult = enum { drained, batch_cap, fairness_yield, shutdown };

const result = acceptReadyConnectionsShard(
    ctx.listen_fd,
    ctx.shard_id,
    ctx.worker_pool,
    ctx.state,
    ctx.batch_options,
);
if (result == .fairness_yield) {
    std.Thread.yield() catch {};
}
```

`acceptBatchShouldYield` can return a stop reason (or the accept function can otherwise report whether fairness vs the hard cap ended the turn). The existing fairness unit test proves that the inner function returns at N and increments the metric; it does not prove that the outer dedicated shard loop actually yields before immediately accepting the next N.

Jaruso (commented):
Full end-to-end review of `bc5d1a6` against every requirement in #146. I re-read all 9 changed files, the surrounding event-loop implementation, the sharded listener path, worker/overload flow, metrics exposition, behavioral tests, docs, and the complete benchmark script rather than only the latest commit. CI run 1334 is green.

The config/main-loop/metrics/docs/tests portions are in good shape. I verified the main event loop uses level-triggered epoll/kqueue, so returning after a capped batch does not strand an already-readable listener. The earlier cap-bypass, behavioral-test, invalid-artifact, socket-error validation, and swallowed-errexit findings are all addressed.

There are **three remaining code blockers, total**:

1. **Benchmark failure cleanup/hang (existing open thread at the current `run_profile` failure paths).** Once the gateway is started, any `|| return 1` can bypass the normal `cleanup_gateway` at the bottom of `run_profile`. Because the profile itself runs inside `$(run_profile ...)`, the parent shell does not receive the child `gateway_pid`. In the common case where the background gateway inherits the command-substitution stdout pipe, an early return can do worse than leak the process: the command substitution can wait indefinitely for that still-running gateway to close the pipe. Make cleanup unconditional for every profile exit, preferably with a single cleanup epilogue/status path (or a cleanup trap installed inside the command-substitution context), not the parent EXIT trap.

2. **The fairness control does not actually yield in the dedicated sharded accept loop.** The inner accept function returns at the fairness threshold, but `runShardAcceptLoop` immediately calls level-triggered `poll()` on the same still-readable listener and can resume accepting immediately. The new inline comment gives a concrete stop-reason/`std.Thread.yield()` shape. The single-listener main event loop already gets real fairness by returning to process other events; the dedicated shard thread needs an equivalent scheduler yield when the fairness limit is what stopped the turn.

3. **Benchmark validation proves batching only across the aggregate of all workloads.** `validate_profile_accept_batching` sums churn + burst + mixed. A profile can therefore pass because the burst workload produced batches >1 while the connection-churn and mixed workloads stayed entirely at batch size 1. That would still leave the two performance acceptance criteria untested. Validate batching separately for each required workload (at minimum churn and mixed; I recommend burst too). The inline comment includes a concrete helper shape.

After those three code fixes, there is one non-code acceptance step: run the corrected benchmark matrix and attach/summarize the valid results. For #146 to close, the evidence should show on the same build/host:

- `batch-1` vs `batch-64` connection-churn req/s with the `batch-64` churn sample actually observing batches >1;
- the burst workload with useful throughput and zero client/socket errors;
- mixed active-keepalive + churn p99 and p999 for both halves, with batching actually exercised during mixed traffic;
- accept errors and worker-queue wait/count/sum (or derived average) for the same workloads;
- `batch-64-fair-8` recording real fairness yields;
- per-shard accept distribution on a host where the requested shard count is actually active (Linux is the obvious choice given the previous macOS fallback);
- enough profile/workload isolation or recovery to avoid the TIME_WAIT/ephemeral-port exhaustion that invalidated the prior run. If deltas are small, repeat runs rather than relying on a single noisy sample.

Once those three code items and the resulting valid measurements are present, I do **not** have another blocker in the current PR. This review is posted as COMMENT because GitHub does not allow the connected author account to submit REQUEST_CHANGES on its own PR.

benchmarks/listener-sharding.sh
```diff
@@ -402,11 +449,80 @@ accept_delta_json() {
...
+        sleep "$RECOVERY_SECONDS"
+    fi
+}
+
+validate_profile_accept_batching() {
+    local label="$1" batch_limit="$2" fairness_yield_every="$3"
+    shift 3
+
+    local accepted batches yields
+    accepted="$(printf '%s\n' "$@" | jq -s '[.[].accept_metrics.accept_batch_size_sum_delta[]?.value] | add // 0')" || return 1
```
Jaruso: This validation is too broad for #146 because it aggregates **churn + burst + mixed** before checking `accepted > batches`. A `batch-64` run can pass this guard if only the burst workload ever sees a batch >1 while `connection-churn-http1` and `mixed-keepalive-churn` remain entirely size-1. In that case the suite still cannot establish either required claim: batching improved churn throughput, or batching did not regress mixed-traffic tail latency, because those workloads never exercised batching.

Validate the required workloads independently. For example:

```bash
validate_workload_batching() {
    local profile="$1" workload="$2" result="$3"
    local accepted batches
    accepted="$(jq '[.accept_metrics.accept_batch_size_sum_delta[]?.value] | add // 0' <<<"$result")" || return 1
    batches="$(jq '[.accept_metrics.accept_batch_size_count_delta[]?.value] | add // 0' <<<"$result")" || return 1
    if (( accepted <= batches )); then
        echo "${profile}/${workload}: batching was not exercised (accepted=${accepted}, batches=${batches})" >&2
        return 1
    fi
}

if (( batch_limit > 1 )); then
    validate_workload_batching "$label" "connection-churn" "$churn_json" || return 1
    validate_workload_batching "$label" "connection-burst" "$burst_json" || return 1
    validate_workload_batching "$label" "mixed-keepalive-churn" "$mixed_json" || return 1
fi
```

Keep the fairness-yield check as well, but don't let activity in one workload certify that an unrelated workload exercised the feature.

Jaruso (commented):
Re-review of `7237aff` against the three blockers from the comprehensive pass and issue #146.

All three code blockers are fixed on this head:

1. **Failure cleanup is now reliable.** `run_profile` routes every post-start failure through `profile_fail`, which calls `cleanup_gateway` before returning the original nonzero status. The prior command-substitution/orphaned-gateway problem is addressed.
2. **Sharded fairness now performs a real scheduler yield.** `acceptReadyConnectionsShard` returns an `AcceptTurnResult`; the dedicated shard loop calls `std.Thread.yield()` specifically when the turn ended via `.fairness_yield`. The batch/fairness tests were updated to assert the stop reason.
3. **Benchmark batching validation is now workload-specific.** `validate_workload_batching` is applied independently to connection churn, burst, and mixed traffic for every `batch_limit > 1` profile, so activity in one workload can no longer certify another.

I rechecked the surrounding accept-loop behavior and the new failure paths and did not find another code-level blocker. CI run 1339 on this head completed successfully.

The **only remaining blocker for closing #146 is acceptance evidence, not code**. The PR still has no replacement valid `summary.json` / measured benchmark result after the earlier invalid macOS run was removed. #146 explicitly requires the suite to be run and the results to establish: connection-churn throughput effect, burst behavior, mixed keepalive+churn p99/p999 without regression, accept errors, worker-queue latency, and shard fairness where applicable. The corrected harness now has the right guards to reject invalid samples; it just needs one valid run whose results are attached or summarized.

Once that valid benchmark run is present and the numbers satisfy the ticket, I have no remaining review comment on the implementation itself. As before, GitHub does not allow the connected author account to submit a formal approval/request-changes review on its own PR, so this is posted as COMMENT.

Jaruso (commented):
Re-review after merging `origin/main` into the #146 implementation branch and resolving the `src/edge_config.zig` conflict.

I compared the post-merge head `f48c2e4` against the previously reviewed clean head and then re-read the final PR diff against the new base (`7692088`, which includes #602's opt-in `io_uring` event-loop backend).

**Conflict resolution is clean.** `edge_config.zig` preserved both sides correctly: the #146 fields (`accept_batch_limit`, `accept_fairness_yield_every`), env parsing/defaults, and returned config values coexist with #148's new event-loop backend fields/parsing. I do not see any dropped or duplicated config state.

I also checked the new-base interaction that could have mattered for accept batching. The unsharded listener now may run on the opt-in readiness-only `io_uring` backend, whose one-shot `POLL_ADD` completions are re-armed inside `EventLoop.wait()` specifically to preserve the level-triggered persistence callers rely on. Therefore returning from `acceptReadyConnectionsShard` after `batch_cap` / `fairness_yield` does not strand a still-readable listener. The sharded accept path remains on its dedicated `poll()` threads and is unaffected by the shared event-loop backend.

All previously reviewed #146 fixes are still present after the merge:
- every successful accept path honors the configured cap/fairness boundary;
- sharded fairness reports a distinct `fairness_yield` result and calls `std.Thread.yield()`;
- focused socket-level tests cover the cap, subsequent draining, fairness stop reason, and metric increments;
- benchmark validation propagates failures, cleans up the gateway on every profile failure, rejects client/socket-error samples, and validates batching independently for churn, burst, and mixed traffic;
- metrics/config/docs remain consistent.

CI run **1343** is green, the PR is mergeable, and all prior review threads remain resolved. I found **no new code or merge-conflict blocker**.

The real-hardware benchmark execution originally listed in #146 is now explicitly tracked in #593, which contains the full `batch-1` / `batch-64` / `batch-64-fair-8` / sharded execution matrix and validity criteria. Given that scope split, I am not treating the absence of those measurements as a blocker for this implementation PR.

No further changes requested from this review.

PR #602, opened 2026-08-10
perf(event-loop): readiness-only io_uring backend behind a feature flag (#148)

Jaruso:
## Summary

Phase 0 (the design doc) plus **Phase 1: the readiness-only `io_uring`
prototype that doc specified now exists**, behind
`TARDIGRADE_EVENT_LOOP_BACKEND=io_uring`.

`epoll`/`kqueue` remains the default on every platform. Nothing changes
unless an operator explicitly opts in.

## What the backend is

It sits behind `event_loop.zig`'s existing `add`/`modify`/`remove`/`wait`
seam — the seam's shape is unchanged — and uses only `IORING_OP_POLL_ADD`,
`IORING_OP_POLL_REMOVE`, and `IORING_OP_TIMEOUT` (the last solely to
implement `wait(timeout_ms)`'s bounded block).

It is **not** an operation/completion abstraction: no `IORING_OP_ACCEPT`, no
registered buffers/files, no multishot variants. `accept()` and every
per-request read/write stay exactly where they were, as blocking calls on
worker threads. `gateway_accept.zig`'s sharded accept `poll()` loops,
`keepalive_park.zig`'s `ParkedRegistry` reaper and maintenance tick, and
`gateway_shutdown.zig`'s drain are all untouched — as this doc argued after
review, a readiness-only boundary exposes no per-connection
timer/cancellation op, so there is nothing there it could simplify.

It serves all three shared-`EventLoop` roles identically to `epoll`:
unsharded listener accept, parked HTTP/1 keepalive, and native-TLS
`ActiveRegistry`/`ManagedConnection` handshake+idle readiness.

**Minimum kernel: Linux 5.5**, established by runtime feature/opcode probing
rather than a version check. `IORING_OP_TIMEOUT` alone would imply 5.4;
`IORING_FEAT_NODROP` raises it to 5.5 (see difference 3 below).

## The parts most worth reviewing

Four places where the ring is genuinely not a drop-in for level-triggered
`epoll`:

1. **`POLL_ADD` is one-shot.** Callers depend on level-triggered persistence
   (the listener fd is registered once and never re-added), so every
   delivered completion is re-armed inside `wait` before it returns.
2. **A completion carries the mask from when the poll fired**, not readiness
   at `wait` time — so read and write readiness for one fd can arrive on
   separate iterations where `epoll` would coalesce them. No readiness is
   lost. Closing the gap needs `IORING_POLL_ADD_LEVEL`, which is 5.13+ and
   above this prototype's floor.
3. **A lost or failed completion strands a registration permanently.**
   Because a one-shot poll is re-armed only after its CQE is consumed,
   anything that consumes the poll without yielding a usable completion
   leaves the table claiming the fd is armed when nothing will ever re-arm
   it — silently fatal for the listener. Two guards: startup requires
   `IORING_FEAT_NODROP` (kernels before 5.5 discard completions on CQ
   overflow), and a negative `cqe.res` on a live token fails the loop instead
   of being skipped.
4. **Closing a fd does not unregister it, so unregister-before-close is an
   invariant.** `epoll` drops a closed fd from every set; `POLL_ADD` instead
   pins the underlying file, so a poll left armed across a close survives it
   and can be reported against — and re-armed on — whatever descriptor
   inherits that number. `reapActiveConnections` already removed before
   `deinit`; `parkedConnectionCloseHook` now removes before
   `ParkedRegistry.freeConn` closes, covering the idle reaper, drain and
   `closeAll`. `add` keeps strict `EPOLL_CTL_ADD`/`EEXIST` semantics —
   inferring staleness from a duplicate fd number was tried and is unsound,
   since it cannot cover a stale completion reaped *before* the re-add.

Plus one internal concern: `epoll_ctl` is kernel-synchronized and
`removeReadFd` documents worker threads calling it while the loop thread is
in `epoll_wait`. The io_uring SQ is plain shared memory, so every userspace
ring mutation is mutex-guarded and submitting threads call `io_uring_enter`
themselves with `min_complete = 0`, so a poll armed from a worker takes
effect even while the loop thread is blocked in its own `io_uring_enter`.
Registrations carry a monotonic token in the high half of `user_data` so a
completion racing a `POLL_REMOVE` is discarded rather than reported against
a fd the caller has already removed.

## Failure handling

**Startup fails closed.** An explicitly requested backend that cannot be
established — unsupported kernel, missing opcode, missing
`IORING_FEAT_NODROP`, seccomp-blocked `io_uring_setup`,
`RLIMIT_MEMLOCK`/ring allocation failure — aborts startup with a clear error
rather than silently falling back to `epoll`. No runtime hot-switching.
`auto` is rejected as a config value, because the auto-detect-with-fallback
mode the doc sketches is deliberately not implemented.

**Runtime failures are not swallowed.** A poll or timer counts as armed only
once the kernel accepts it; `EINTR` from `io_uring_enter` with
`min_complete = 0` is not treated as success. If the bounded-wait timer
cannot be queued under pressure, `wait` degrades to a non-blocking poll
rather than blocking with no timer to break it. Anything worse is sticky and
surfaces as `error.EventLoopUnrecoverable`, on which `edge_gateway` requests
shutdown and drains — deliberately stricter than the epoll path's
log-and-continue, because a wedged ring can stop delivering listener
readiness while the process still looks healthy.

## Configuration

| Variable | Default | Meaning |
|---|---|---|
| `TARDIGRADE_EVENT_LOOP_BACKEND` | `default` | `default` keeps the platform backend. `io_uring` selects the prototype. `epoll`/`kqueue` may be named explicitly and fail closed if not this platform's backend. |
| `TARDIGRADE_EVENT_LOOP_IO_URING_ENTRIES` | `256` | SQ depth, bounded 64–4096. The modest default keeps ring memory (~24 KiB) inside the 64 KiB `RLIMIT_MEMLOCK` soft limit common on stock hosts, since kernels before 5.12 charge ring memory against it. |

## Validation status — please read before citing this

- **Eight Linux-only tests:** one ring-size fail-closed test plus seven that
  drive a live ring. They mirror the `epoll`/`kqueue` behaviour tests (write
  readiness, `modify`, re-arm persistence, removal, duplicate rejection) and
  add three regressions for the failure modes above — close/fd-number-reuse,
  CQ pressure (150 ready fds against a 128-entry CQ), and a failed completion
  on a live token surfacing as `EventLoopUnrecoverable`. Live-ring tests skip
  themselves where `io_uring_setup` is blocked.
- **Executed on Linux by CI, not by me.** My development host was
  macOS/arm64, where this path compiles out entirely; I validated it there
  only by cross-compiling for `x86_64-linux-gnu` and by review. CI is the
  first actual execution, and the io_uring tests run rather than skip on both
  `ubuntu-latest` (x86_64) and `ubuntu-24.04-arm` (aarch64) — confirmed by
  skip differential: macOS skips 9 where Linux skips 1, a difference of
  exactly 8.
- **Two branches remain unexercised.** CI kernels always have
  `IORING_REGISTER_PROBE`, so the functional `-ETIME` fallback (reachable
  only at exactly 5.5) never runs; and the `IORING_FEAT_NODROP` refusal path
  cannot run on a kernel that has the feature. The CQ-pressure test proves
  the overflow backlog works *with* NODROP, not that the refusal works
  without it.
- **No benchmark evidence of any kind.** No profiling pass has shown the
  shared `EventLoop`'s readiness/idle-wait role to be a bottleneck, and no
  throughput/latency comparison against `epoll` has been run. Per
  `CONCURRENCY.md` and `UPSTREAM_POOLING.md`, such numbers must come from
  real hardware. Nothing here should be cited as evidence `io_uring` is
  faster, slower, or equivalent for this workload.

The evidence precondition in the doc's Recommendation is therefore still
**not** met. The prototype was built ahead of that evidence deliberately, to
make the option concrete and measurable — not to argue for changing the
default.

## Docs

`docs/EVENT_LOOP_BACKENDS.md` gains a "Phase 1 prototype" section covering
all of the above, and its status/recommendation/matrix are reconciled.
`docs/CONCURRENCY.md`'s standing "any new backend must be
benchmark-justified" rule is reconciled explicitly: it still stands, and
this backend is opt-in precisely because it does not yet meet it.

## Follow-ups still open under #148

This PR does not close #148. Still outstanding:

- Benchmark/profiling evidence on real hardware against the `benchmarks/`
  harness (#136). If it shows no measurable win, the honest outcome is to
  remove the prototype rather than carry it.
- `tardigrade_event_loop_iterations_total{backend=...}` label.

Refs #148.

### DIFF
--- a/docs/EVENT_LOOP_BACKENDS.md
+++ b/docs/EVENT_LOOP_BACKENDS.md
@@ -90,10 +90,19 @@ detail in [`docs/CONCURRENCY.md`](CONCURRENCY.md):
   handles roles 2 and 3 above. Accepted fds from every shard still enter
   the same bounded worker pool.
 - Accepted connections are dispatched to a bounded `WorkerPool`
-  (`src/http/worker_pool.zig`) of OS threads. Each worker owns one
-  connection's full synchronous lifecycle — TLS handshake, HTTP parse,
-  proxy/serve, response write — using **blocking** socket calls
-  (`read`/`write`/`poll`/`SO_RCVTIMEO`/`SO_SNDTIMEO`). `worker_pool.zig`
+  (`src/http/worker_pool.zig`) of OS threads, but "one worker owns the whole
+  connection lifecycle uninterrupted" is only accurate for the legacy
+  OpenSSL/plaintext path — there a worker does own one connection's full
+  synchronous lifecycle (TLS handshake, HTTP parse, proxy/serve, response
+  write) using **blocking** socket calls
+  (`read`/`write`/`poll`/`SO_RCVTIMEO`/`SO_SNDTIMEO`) with no handoff back to
+  `EventLoop` until an idle keepalive gap. For the native-TLS path, a worker
+  instead owns *active request/handshake processing*: a connection can yield
+  out of `native_handshake` (or the idle gap after a response) back to
+  `EventLoop`/`ActiveRegistry` and be re-dispatched to a different worker
+  later, so no single worker owns that connection's end-to-end lifecycle —
+  only each checked-out slice of it, which is itself still blocking work.
+  `worker_pool.zig`
   documents why this is deliberate: `std.Io.Group` (Zig 0.16) and the removed
   `std.Thread.Pool` both assume non-blocking, async-style work items, and
   Tardigrade needs direct control over thread count, work-stealing, CPU
@@ -126,10 +135,37 @@ exactly the surface `src/http/event_loop.zig` already exposes:
 - block for a bounded time and return ready events (`wait`)
 - report a stable backend identifier for logs/metrics (`backendName`)
 
+This is a **readiness-poller** abstraction, not an operation/completion
+abstraction: it has no submission-token model, no way to submit an
+accept/read/write/timeout operation, and no way to return an operation
+result (an accepted fd, a byte count). That shape is a deliberate choice for
+Phase 0, not an oversight — see the explicit call-out at the end of this
+section.
+
 Any alternative backend considered below is evaluated against whether it can
 sit behind this same four-operation boundary without forcing a rewrite of the
 blocking, thread-per-request handler code that currently owns TLS, HTTP
 framing, and proxy streaming.
+
+**This constrains what a future `io_uring` backend may be, and this doc picks
+the narrower of the two honest options:** a future `io_uring` backend
+targeting this boundary is a **readiness-only poller replacement** —
+`IORING_OP_POLL_ADD`/`IORING_OP_POLL_REMOVE` (plus a ring-native timeout for
+`wait`'s bounded-block behavior) standing in for `epoll_wait`/`kqueue`, with
+`accept()` and all request I/O staying exactly where they are today: outside
+the ring, as blocking calls on worker threads. It is **not** a wider
+operation/completion abstraction using `IORING_OP_ACCEPT`, registered
+buffers/files, or multishot variants — adopting those would require
+widening this boundary to typed submissions/completions (e.g. `submitAccept`,
+`submitRead`, `submitWrite`, `submitTimer`, `cancel`, `waitCompletions` with
+opaque tokens) and defining how `epoll`/`kqueue` implement or adapt that
+wider contract, which is a materially larger design than Phase 0 scopes and
+is explicitly out of scope here. A later doc revision would need to make
+that adoption case explicitly, including why the readiness-only version was
+insufficient. Under the readiness-only choice, the sharded per-shard accept
+`poll()` loops (`gateway_accept.zig`) are **not** exercised by this boundary
+at all unless a separate future change first migrates them onto the shared
+`EventLoop` — they stay on `std.posix.poll()` either way.
 
 ## Alternatives compared
 
@@ -245,11 +281,16 @@ framing, and proxy streaming.
   scheduling, not the request hot path itself: per-request reads/writes stay
   synchronous blocking calls on worker threads (including within the
   native-TLS path's own `WaitingEncryptedHttpConnection.waitFor`, which polls
-  locally rather than through the shared loop). Useful `io_uring`
-  features (e.g. `IORING_OP_ACCEPT`, fixed buffers, registered fds) require
-  reasonably recent kernels (5.5+ minimum, several features need 5.11+ or
-  5.19+), which is a real portability constraint the current epoll fallback
-  does not have.
+  locally rather than through the shared loop). Per the "Backend abstraction
+  boundary" section above, a backend targeting this boundary is
+  **readiness-only** (`IORING_OP_POLL_ADD`/`IORING_OP_POLL_REMOVE` plus a
+  ring-native timeout) — it does not use `IORING_OP_ACCEPT`, registered
+  buffers/files, or multishot variants, since those require widening the
+  boundary to an operation/completion abstraction, which is explicitly out
+  of scope for Phase 0. The readiness-only operation set is available from
+  the original `io_uring` interface (Linux 5.1); see "Failure modes and
+  fallback policy" below for how kernel-requirement documentation should be
+  structured if a wider abstraction is ever proposed instead.
 - **macOS support:** None — `io_uring` is Linux-only, so this would always
   need to stay behind a compile-time/feature-flag gate with `epoll`/`kqueue`
   as the portable default, exactly as #148 already proposes.
@@ -380,14 +421,22 @@ no prototype exists yet:
   runtime conditions (log, drain, exit), consistent with how the current
   `EventLoop.wait()` error path already just logs and continues rather than
   trying to self-heal by rebuilding the poller.
-- **Kernel/toolchain requirements, if pursued:** modern `io_uring` opcodes
-  useful here (`IORING_OP_ACCEPT`, fixed buffers/files, multishot variants)
-  need meaningfully newer kernels than the epoll baseline requires — commonly
-  cited minimums range from 5.5 (basic ring support) through 5.11/5.19 for
-  specific opcodes/features relevant to a listener-readiness use case. A
-  prototype must pin and document the exact minimum kernel version it
-  targets and probe for it explicitly rather than assuming availability from
-  a compile-time Linux check.
+- **Kernel/toolchain requirements, if pursued:** separate the **ring
+  baseline** from the **prototype's exact operation set**, rather than citing
+  one number for "basic support." `io_uring` itself — the
+  `io_uring_setup`/`io_uring_enter`/`io_uring_register` syscalls, and with
+  them `IORING_OP_POLL_ADD`/`IORING_OP_POLL_REMOVE` (the readiness-only
+  operation set this doc scopes a prototype to, per "Backend abstraction
+  boundary" above) — landed in Linux **5.1**. Individual capabilities this
+  doc explicitly excludes from a readiness-only prototype have their own,
+  later floors if a future wider abstraction ever adopts them:
+  `IORING_OP_ACCEPT` (5.5), multishot poll (5.13), and multishot accept
+  (5.19, per upstream `liburing` docs) are commonly cited examples, not an
+  exhaustive list. A prototype must pin and document the exact minimum
+  kernel version it targets **for the specific opcodes/features it actually
+  uses**, and probe for those capabilities explicitly at startup rather than
+  assuming availability from a compile-time Linux check or from the 5.1 ring
+  baseline alone.
 
 This policy applies to whichever `EventLoop` role(s) a future prototype
 targets (unsharded accept, parked keepalive, and/or native-TLS active
--- a/docs/EVENT_LOOP_BACKENDS.md
+++ b/docs/EVENT_LOOP_BACKENDS.md
@@ -563,25 +563,28 @@ loop.
 
 ### Validation status — read this before citing the prototype
 
-- **Unit tests:** the `epoll`/`kqueue` behaviour tests in `event_loop.zig` are
-  mirrored for the `io_uring` backend (write readiness, `modify` replacing
-  interest, re-arm persistence, removal, duplicate-registration rejection),
-  plus fail-closed tests for an unavailable backend and an unusable ring size,
-  a close/fd-number-reuse test that verifies no completion from a closed
-  registration is reported against or re-armed on the descriptor that inherits
-  its number, and a CQ-pressure regression that makes 150 fds ready
-  simultaneously against a 128-entry completion queue and requires every one
-  to surface. The behaviour tests skip themselves when `io_uring_setup` is
-  unavailable — container seccomp profiles, Docker's default among them, block
-  it outright — since that is an environment property, not a defect in the
-  backend.
+- **Unit tests: eight Linux-only tests** — one ring-size fail-closed test plus
+  seven that drive a live ring. They mirror the `epoll`/`kqueue` behaviour
+  tests (write readiness, `modify` replacing interest, re-arm persistence,
+  removal, duplicate-registration rejection) and add three regressions for the
+  failure modes above: a close/fd-number-reuse test verifying no completion
+  from a closed registration is reported against or re-armed on the descriptor
+  that inherits its number; a CQ-pressure test making 150 fds ready against a
+  128-entry completion queue and requiring every one to surface; and an
+  operation-error test asserting that a negative completion on a live token
+  makes `wait` return `error.EventLoopUnrecoverable` rather than silently
+  dropping the registration. There is also a platform fail-closed test for an
+  unavailable backend. The live-ring tests skip themselves when
+  `io_uring_setup` is unavailable — container seccomp profiles, Docker's
+  default among them, block it outright — since that is an environment
+  property, not a defect in the backend.
 - **Executed on Linux by CI, not by the author.** The development host for
   this change was macOS/arm64, where the io_uring path compiles out entirely;
   it was validated there only by cross-compiling for `x86_64-linux-gnu` and by
   review. The Linux CI jobs were the first actual execution, and all six
   io_uring tests ran and passed there on both `ubuntu-latest` (x86_64) and
   `ubuntu-24.04-arm` (aarch64) — confirmed by skip-count differential: the
-  macOS job skips 7 tests where the Linux jobs skip 1, and the difference of 6
+  macOS job skips 9 tests where the Linux jobs skip 1, and the difference of 8
   is exactly this backend's tests. So `io_uring_setup` is not blocked on those
   runners, the capability probe succeeds, and the readiness, `modify`,
   re-arm, removal, and recycled-fd paths all behave as intended on a live
--- a/docs/EVENT_LOOP_BACKENDS.md
+++ b/docs/EVENT_LOOP_BACKENDS.md
@@ -1,0 +1,317 @@
+# Event-loop backend evaluation (#148)
+
+Status: **Phase 0 — design doc only.** No runtime behavior changes. This
+document satisfies the design-doc acceptance criteria from #148 and from the
+consolidated architecture-evaluation scope folded in from #213. It does not
+implement, prototype, or benchmark an `io_uring` (or other) backend.
+
+## Why
+
+#148 asks Tardigrade to evaluate an optional Linux `io_uring` backend behind
+an event-loop abstraction, and #213 broadened that into a comparison of the
+current epoll/kqueue model against `libxev`, `std.Io` evented/proactor
+facilities, and direct `io_uring`, for the repository's pinned Zig toolchain
+(`minimum_zig_version = 0.16.0`, see `build.zig.zon`).
+
+#148 explicitly scopes itself as a second-order optimization: "research
+suggests this is a second-order optimization after listener sharding,
+buffering/backpressure, and upstream connection economics," and states it
+"should be tackled after #137, #138, #139, #140, and #141 provide stronger
+baseline wins."
+
+## Dependency status (checked 2026-08-10)
+
+| Issue | Title | Status |
+|---|---|---|
+| #136 | Benchmark harness and telemetry | Closed |
+| #137 | Per-core listener sharding (`SO_REUSEPORT`) | Closed |
+| #138 | Park idle keepalive connections outside workers | Closed |
+| #139 | Streaming reverse proxy with bounded buffering | Closed |
+| #140 | Watermark-based backpressure for proxy/connection buffers | **Open** — PR #600 (HTTP/1 relay buffer bounding) in flight |
+| #141 | Upstream connection pooling redesign | Closed |
+
+Four of the five prerequisite baseline-wins issues are closed; #140 is not.
+Per #148's own stated ordering, this is not yet the point at which a
+prototype or migration should start. This doc proceeds only as far as the
+design/comparison work, and defers the prototype/benchmark decision until
+#140 lands (see "Recommendation" below).
+
+## Current architecture and poller assumptions
+
+Tardigrade uses a **blocking I/O, thread-per-request** model, audited in
+detail in [`docs/CONCURRENCY.md`](CONCURRENCY.md):
+
+- `src/http/event_loop.zig` (`EventLoop`) wraps `epoll` on Linux and `kqueue`
+  on BSD/macOS behind a small `add`/`modify`/`remove`/`wait` interface keyed
+  on raw fds, with `Interest{ read, write }` and a fixed-capacity `Event`
+  output array. `EventLoop.init()` picks the backend from `builtin.os.tag` at
+  compile/runtime; there is no runtime backend override today.
+- The event loop's job is narrow: watch the listener fd(s) for `accept`
+  readiness and watch **parked idle keepalive connections**
+  (`src/http/keepalive_park.zig`, #138) for the next request byte. It is not
+  a general async-I/O runtime — active request handling never touches it.
+- Accepted connections are dispatched to a bounded `WorkerPool`
+  (`src/http/worker_pool.zig`) of OS threads. Each worker owns one
+  connection's full synchronous lifecycle — TLS handshake, HTTP parse,
+  proxy/serve, response write — using **blocking** socket calls
+  (`read`/`write`/`poll`/`SO_RCVTIMEO`/`SO_SNDTIMEO`). `worker_pool.zig`
+  documents why this is deliberate: `std.Io.Group` (Zig 0.16) and the removed
+  `std.Thread.Pool` both assume non-blocking, async-style work items, and
+  Tardigrade needs direct control over thread count, work-stealing, CPU
+  affinity, and drain-before-shutdown semantics for a production reverse
+  proxy.
+- Listener sharding (#137) runs one independent blocking `poll()`-based
+  accept loop per shard thread (`src/gateway_accept.zig`) when
+  `TARDIGRADE_LISTENER_SHARDS > 1` and `SO_REUSEPORT`/`SO_REUSEPORT_LB` is
+  available; accepted fds still enter the same bounded worker pool.
+- `docs/CONCURRENCY.md` already states the project's standing position on
+  this exact question: *"Do not start an `io_uring` or async-runtime rewrite
+  from this boundary. Any new backend must be benchmark-justified and fit
+  behind the same socket/protocol ownership rules."* This design doc does not
+  change that position; it documents the comparison the maintainer status
+  block on #148 asked for so the position is backed by a written rationale
+  rather than only a one-line rule.
+- `tardigrade_event_loop_iterations_total` (a counter of timer-tick
+  iterations) is already exported in Prometheus and JSON metrics
+  (`src/http/metrics.zig`); the event-loop backend name (`"epoll"` /
+  `"kqueue"`) is already logged once at startup
+  (`src/edge_gateway.zig:576`) but is not yet attached as a metric label.
+  Adding a `backend` label to the existing counter (mirroring the
+  `{backend=...}` pattern already used by the TLS buffer metrics) is a small,
+  low-risk follow-up independent of the rest of this evaluation — see
+  "Small independent follow-up" below.
+
+## Backend abstraction boundary
+
+For the purposes of this comparison, the "event-loop backend" boundary is
+exactly the surface `src/http/event_loop.zig` already exposes:
+
+- register/modify/remove readiness interest for a fd (`add`, `modify`,
+  `remove`)
+- block for a bounded time and return ready events (`wait`)
+- report a stable backend identifier for logs/metrics (`backendName`)
+
+Any alternative backend considered below is evaluated against whether it can
+sit behind this same four-operation boundary without forcing a rewrite of the
+blocking, thread-per-request handler code that currently owns TLS, HTTP
+framing, and proxy streaming.
+
+## Alternatives compared
+
+### A. Current model — direct `epoll`/`kqueue`, thread-per-request workers
+
+- **Linux fast path:** Mature, well-understood, level-triggered epoll used
+  only for accept + parked-keepalive readiness (a small fraction of total
+  I/O — active request I/O is synchronous blocking calls on worker threads).
+  No known syscall-count problem has been measured in this narrow role;
+  #148's own research basis frames `io_uring`'s syscall-reduction pitch as
+  most valuable for I/O-heavy async runtimes, which this narrow accept/park
+  loop is not.
+- **macOS/kernel requirements:** `kqueue` is available on every supported
+  BSD/macOS target with no minimum-version gymnastics. No portability
+  fallback logic needed — it already covers both `SUPPORT_MATRIX.md`
+  platforms.
+- **Complexity / maintenance:** Already implemented, tested, and in
+  production use (398 lines, unit-tested for both backends via CI matrix).
+  Zero incremental maintenance cost to keep it.
+- **TLS/encrypted-stream integration:** Already fully integrated —
+  `keepalive_park.zig` moves `TlsConnection` state off the worker stack
+  across the idle gap and the event loop only tracks fd readiness, never TLS
+  record state.
+- **Proxy streaming / backpressure:** Already integrated with #139's
+  streaming relay and #140's (in-progress) watermark work, which are both
+  built on synchronous blocking reads/writes, not on event-loop readiness
+  callbacks.
+- **Timers, cancellation, keepalive parking, graceful drain:** Already
+  implemented (`keepalive_park.zig` idle reaper on the maintenance tick,
+  `gateway_shutdown.zig` drain, per-phase `poll(2)`/`SO_*TIMEO` deadlines on
+  worker threads).
+- **Prototype needed?** No — this is the shipped baseline.
+
+### B. Cleaner direct epoll/kqueue abstraction (refactor only, same backends)
+
+- Would not change Linux/macOS support or kernel requirements at all — same
+  syscalls, same platforms.
+- Only credible motivation is code-quality (e.g., unifying edge- vs
+  level-triggered handling, or widening the event loop's role beyond
+  accept/park). #148 does not identify a concrete pain point in the current
+  abstraction that blocks other roadmap work, and `CONCURRENCY.md`'s
+  "Follow-up work" table (h1/h2/h3 stream transport boundary, QUIC UDP
+  endpoint) tracks the actual seams that need boundary work next (#241,
+  #242, #248), not `event_loop.zig` itself.
+- **Verdict:** No action item here beyond normal refactoring hygiene; not a
+  distinct backend decision.
+
+### C. `libxev`
+
+- Not vendored or referenced anywhere in this repository today (`libxev`
+  search returns no hits).
+- Would require adopting `libxev`'s completion-based (io_uring-on-Linux,
+  kqueue-on-BSD/macOS, IOCP-on-Windows) async model, which is fundamentally
+  callback/completion-oriented — a different concurrency shape than
+  Tardigrade's blocking-thread-per-request handler. Adopting it for real gain
+  would mean rewriting the request handler to be non-blocking end-to-end
+  (TLS, HTTP parsing, proxy relay, FastCGI/SCGI/uWSGI transports), which
+  `worker_pool.zig`'s design rationale explicitly warns against doing without
+  first replacing the blocking I/O model throughout the connection handler
+  and capturing an expected throughput/latency delta.
+- Pulling in `libxev` as a dependency also cuts against the project's
+  documented "core Zig only" direction for the data plane (see
+  `docs/UPSTREAM_POOLING.md`'s HTTP/2 upstream work, which built h2 framing
+  in-repo rather than take an external dependency, keeping OpenSSL as the
+  one intentional exception).
+- **Verdict:** Not justified now. Revisit only if/when the handler model
+  itself is being redesigned around non-blocking I/O for other reasons.
+
+### D. `std.Io` evented/proactor facilities
+
+- `zig_compat.zig` and `CONCURRENCY.md` already draw this exact line:
+  high-level, non-socket-path code uses `std.Io`/`zig_compat`; the
+  listener/gateway/proxy/H2/H3 socket path deliberately stays on raw
+  `std.posix`/`std.c` because Zig 0.16's `std.Io` does not yet expose
+  stable, production-grade semantics for `accept`, `poll`, non-blocking
+  toggles, `SO_*TIMEO`, and peer-address extraction on the data plane.
+- `worker_pool.zig` already tried and rejected `std.Io.Group` for the exact
+  reason relevant here: it expects non-blocking, async-style work items, and
+  a blocking call on a Group-managed thread stalls the whole group. FastCGI
+  pooling (`docs/UPSTREAM_POOLING.md` Phase 2) hit this directly in
+  production — FastCGI/SCGI/uWSGI over `std.Io` event-loop streams **hung**
+  against `php-fpm`'s blocking request/response exchange, and had to be
+  switched to raw blocking sockets to fix it.
+- **Verdict:** Not viable as a socket-path backend on the pinned Zig 0.16
+  toolchain today, independent of `io_uring` specifically. This is a
+  toolchain-maturity blocker, not a design preference, and should be
+  re-evaluated only if a future Zig release stabilizes proactor-style I/O for
+  the primitives the data plane needs (`accept`, bounded `poll`/timeout
+  reads and writes, peer address extraction).
+
+### E. Direct `io_uring`
+
+- **Linux fast path / kernel requirements:** Real syscall-batching and
+  completion-based I/O upside exists, but it is most valuable for
+  high-fan-out non-blocking I/O (many concurrent reads/writes multiplexed
+  through one submission/completion queue), which is exactly the shape
+  Tardigrade's worker pool does *not* have — each worker already does one
+  blocking `read`/`write` at a time on its own thread. The event loop's
+  actual `io_uring`-addressable surface today is narrow: listener accept and
+  parked-keepalive readiness, not the request hot path. Useful `io_uring`
+  features (e.g. `IORING_OP_ACCEPT`, fixed buffers, registered fds) require
+  reasonably recent kernels (5.5+ minimum, several features need 5.11+ or
+  5.19+), which is a real portability constraint the current epoll fallback
+  does not have.
+- **macOS support:** None — `io_uring` is Linux-only, so this would always
+  need to stay behind a compile-time/feature-flag gate with `epoll`/`kqueue`
+  as the portable default, exactly as #148 already proposes.
+- **Complexity / maintenance:** High. A correct `io_uring` integration means
+  new submission/completion-queue lifecycle code, a second set of
+  cancellation/timeout semantics, and (per Envoy's and Monoio's own
+  documented experience, cited in #148's research basis) materially more
+  operational surface than epoll for a comparatively narrow accept/park role
+  in this codebase's current architecture.
+- **TLS/encrypted-stream integration:** Would not change — TLS record
+  handling happens inside the blocking worker, not in the event loop, so
+  `io_uring` adoption at the accept/park boundary would not touch TLS state
+  machines either way.
+- **Proxy streaming and backpressure:** No expected interaction with #139
+  (streaming relay) or #140 (watermark backpressure) as currently designed,
+  since both operate on synchronous blocking reads/writes inside worker
+  threads, not on event-loop-driven readiness.
+- **Timers, cancellation, keepalive parking, graceful drain:** `io_uring`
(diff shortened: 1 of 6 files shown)


docs/EVENT_LOOP_BACKENDS.md
```diff
@@ -0,0 +1,317 @@
...
+
+- `src/http/event_loop.zig` (`EventLoop`) wraps `epoll` on Linux and `kqueue`
+  on BSD/macOS behind a small `add`/`modify`/`remove`/`wait` interface keyed
+  on raw fds, with `Interest{ read, write }` and a fixed-capacity `Event`
+  output array. `EventLoop.init()` picks the backend from `builtin.os.tag` at
+  compile/runtime; there is no runtime backend override today.
+- The event loop's job is narrow: watch the listener fd(s) for `accept`
+  readiness and watch **parked idle keepalive connections**
+  (`src/http/keepalive_park.zig`, #138) for the next request byte. It is not
+  a general async-I/O runtime — active request handling never touches it.
```
Jaruso: [P2] The current runtime no longer matches this description. `edge_gateway.zig` checks `active.contains(ev.fd)` inside the main `EventLoop.wait()` loop and dispatches `submitActiveSocketReady`; the native-TLS path inserts a `ManagedConnection` into `ActiveRegistry` and calls `ctx.event_loop.add(client_fd, interest)`; `rearmActiveConnection()` registers it again as the native handshake / HTTP state machine advances; and timer-driven `active_drive_poll` work is coordinated through the same event-loop/registry path. At the same time, when listener sharding is enabled, accepts are *not* watched by this `EventLoop` at all — `gateway_accept.runShardAcceptLoop()` uses a separate blocking `poll()` loop per shard.

Please describe the actual split explicitly, e.g. current `EventLoop` roles are (1) unsharded listener readiness, (2) parked HTTP/1 keepalive readiness, and (3) active managed/native downstream readiness/drive scheduling; sharded listener accept loops are separate; the legacy plaintext/OpenSSL request path remains predominantly blocking on workers. Then propagate that correction through the `io_uring` section, TLS row, comparison matrix, and long-term prototype recommendation. As written, the later claim that the only `io_uring`-addressable surface is accept/park is based on this incorrect premise.

docs/EVENT_LOOP_BACKENDS.md
```diff
@@ -0,0 +1,317 @@
...
+  blocking `read`/`write` at a time on its own thread. The event loop's
+  actual `io_uring`-addressable surface today is narrow: listener accept and
+  parked-keepalive readiness, not the request hot path. Useful `io_uring`
+  features (e.g. `IORING_OP_ACCEPT`, fixed buffers, registered fds) require
+  reasonably recent kernels (5.5+ minimum, several features need 5.11+ or
+  5.19+), which is a real portability constraint the current epoll fallback
+  does not have.
+- **macOS support:** None — `io_uring` is Linux-only, so this would always
+  need to stay behind a compile-time/feature-flag gate with `epoll`/`kqueue`
+  as the portable default, exactly as #148 already proposes.
```
Jaruso, line 316: [P2] #148 explicitly requires failure modes, kernel/toolchain requirements, and fallback behavior to be documented. This tells us the portable default, but not what a future feature-flagged `io_uring` backend does when setup/capability probing fails (unsupported kernel/opcodes, seccomp/container restrictions, resource exhaustion) or when the ring fails after startup.

Please make the policy explicit in Phase 0 before a prototype is written. One reasonable shape would be: keep epoll/kqueue as the default; an explicitly requested `io_uring` mode fails startup if required capabilities cannot be established rather than silently changing backend; if an `auto` mode is ever added, initialization failure may log/metric the reason and fall back to epoll *before accepting traffic*; and runtime SQ/CQ errors do not hot-switch a live process between backends, but surface through normal error/drain semantics. The exact policy can differ, but the design note needs to choose and document it to satisfy this acceptance criterion.

docs/EVENT_LOOP_BACKENDS.md
```diff
@@ -0,0 +1,317 @@
...
+  future async I/O model, starting an `io_uring` prototype before it lands
+  risks throwaway work.
+- **Small independent follow-up (optional, not gated on the rest of this
+  doc):** attach a `backend` label to the existing
+  `tardigrade_event_loop_iterations_total` counter (the backend name is
+  already computed via `EventLoop.backendName()` and already logged once at
+  startup), so operators can see the active backend in
+  `/status/metrics` without parsing startup logs. This is a few-line,
+  low-risk metrics change independent of the architecture question and can
+  land whenever convenient; it is not required to close #148.
```
Jaruso: [P2] This PR does not need to implement the metric, but saying it is “not required to close #148” changes the scope defined by the issue. #148's **Metrics to add or verify** section explicitly calls for an event-loop backend identifier in logs/metrics and `tardigrade_event_loop_iterations_total{backend=...}`. Please keep that work tracked under #148 unless the issue itself is amended. Suggested wording: “This can land independently of the architecture prototype; this Phase 0 PR does not implement it, and it remains follow-up work under #148 before closure.”

docs/EVENT_LOOP_BACKENDS.md
```diff
@@ -0,0 +1,317 @@
...
+  #242, #248), not `event_loop.zig` itself.
+- **Verdict:** No action item here beyond normal refactoring hygiene; not a
+  distinct backend decision.
+
+### C. `libxev`
+
+- Not vendored or referenced anywhere in this repository today (`libxev`
+  search returns no hits).
+- Would require adopting `libxev`'s completion-based (io_uring-on-Linux,
+  kqueue-on-BSD/macOS, IOCP-on-Windows) async model, which is fundamentally
```
Jaruso: [P3] Please update this platform claim to match current upstream `libxev` documentation. The current `mitchellh/libxev` README advertises Linux (`io_uring` + epoll), macOS (kqueue), and WASM, and says Windows support is planned/coming soon; `IOCP-on-Windows` therefore reads as a currently shipped backend when upstream does not document it that way. Likewise, avoid broadening kqueue support to generic BSD unless the current upstream support matrix/source actually guarantees those targets. This does not change the recommendation, but the architecture comparison should be precise about the alternative being evaluated.

Jaruso (commented):
**Review result: changes needed before merge.** GitHub does not permit the PR author to submit `REQUEST_CHANGES` on their own PR, so this is recorded as a COMMENT review, but I would treat the P2 findings below as merge-blocking.

Phase 0 as an incremental deliverable is reasonable, and the README docs-index change is fine. I reviewed the full patch against #148 plus the consolidated #213 criteria, then checked the document's architecture claims against the current event-loop, worker-pool, active-connection, keepalive, and sharded-listener code.

The main blocker is that the design note models the current event loop too narrowly. Current `main` does not use it only for listener accept + parked keepalive: native TLS `ManagedConnection`s are registered in the `ActiveRegistry`, armed on `EventLoop`, dispatched as `active_socket_ready`/`active_drive_poll`, and re-armed as their handshake/HTTP state machines progress. Conversely, when listener sharding is enabled, accept readiness is handled outside `EventLoop` by `gateway_accept.runShardAcceptLoop` using `poll()`. That distinction materially changes the TLS discussion, the claimed `io_uring`-addressable surface, the comparison matrix, and the proposed future prototype boundary, so those sections need to be corrected consistently rather than just changing one sentence.

I also left two #148-scope corrections: the Phase 0 design should document concrete initialization/runtime failure and fallback semantics for a future feature-flagged backend, and it should not redefine the issue's requested backend-labeled event-loop metric as unnecessary for #148 closure. There is also one minor factual correction for current `libxev` platform support.

CI note: the PR's CI run is currently red only in `Test appliance profile (ubuntu-24.04-arm)`, where the existing `src/http/upstream_h2.zig` test `origin capacity bounds concurrent streams and refuses only the overflowing one` reported `expected 3, found 0`. Format, normal Linux/macOS builds/tests, perf smoke, lint, security, and the other appliance jobs passed. Since this PR changes only Markdown, I do not consider that ARM-only runtime test failure caused by this patch, but the required check is still visibly red.

Once the current event-loop responsibilities and the downstream recommendation are corrected, the rest of the Phase 0 direction looks consistent with the issue's intended sequencing: no backend migration before #140, no premature follow-up implementation issue, and measurement before choosing an additional backend.

docs/EVENT_LOOP_BACKENDS.md
Jaruso: Fixed in 4e07d03: "Current architecture and poller assumptions" now enumerates the three actual `EventLoop` roles (unsharded listener accept, parked HTTP/1 keepalive, and active managed/native-TLS handshake+idle readiness via `ActiveRegistry`/`ManagedConnection`/`rearmActiveConnection`/`dueDrivePollFds`), notes sharded accept runs entirely outside `EventLoop` via `gateway_accept.runShardAcceptLoop`'s own `poll()`, and clarifies that even native-TLS per-request I/O still blocks locally (`WaitingEncryptedHttpConnection.waitFor`) rather than looping through the shared loop. Propagated through the TLS bullets, the io_uring-addressable-surface claim, the comparison matrix, and the long-term recommendation.

---

docs/EVENT_LOOP_BACKENDS.md
Jaruso: Fixed in 4e07d03: added a new "Failure modes and fallback policy for a future `io_uring` backend" section using essentially the shape you suggested — default stays epoll/kqueue; an explicitly requested `io_uring` mode fails startup (not silent fallback) if capabilities can't be established; a future `auto` mode would probe and fall back before accepting traffic; no runtime hot-switching, SQ/CQ errors surface through normal drain/error semantics; plus documented kernel-version minimums (5.5 baseline, 5.11/5.19 for specific opcodes) that a prototype must pin and probe for explicitly.

---

docs/EVENT_LOOP_BACKENDS.md
Jaruso: Fixed in 4e07d03, using your suggested wording almost verbatim: it now says this "can land independently of the architecture prototype work; this Phase 0 PR does not implement it, and it remains follow-up work under #148 before closure," citing #148's Metrics to add or verify section.

---

docs/EVENT_LOOP_BACKENDS.md
Jaruso: Fixed in 4e07d03: corrected to Linux (`io_uring` or epoll), macOS (`kqueue`), and WASM per upstream `mitchellh/libxev` docs, with Windows/IOCP noted as upstream-documented planned/upcoming rather than shipped, and dropped the generic-BSD kqueue broadening.

---

docs/EVENT_LOOP_BACKENDS.md
```diff
@@ -0,0 +1,425 @@
...
+
+For the purposes of this comparison, the "event-loop backend" boundary is
+exactly the surface `src/http/event_loop.zig` already exposes:
+
+- register/modify/remove readiness interest for a fd (`add`, `modify`,
+  `remove`)
+- block for a bounded time and return ready events (`wait`)
+- report a stable backend identifier for logs/metrics (`backendName`)
+
+Any alternative backend considered below is evaluated against whether it can
```
Jaruso, line 161: [P2] This boundary is too narrow for the direct `io_uring` operations the rest of the design says a future prototype may use. `add`/`modify`/`remove`/`wait` can express **fd readiness** (an `io_uring` implementation could emulate this with `IORING_OP_POLL_ADD`/poll cancellation), but there is nowhere in this API to submit an accept/read/write/timeout operation, associate ownership with a submission token, or return an operation result such as an accepted fd or byte count. That means `IORING_OP_ACCEPT`, registered-buffer reads/writes, multishot accept, etc. cannot actually sit behind the “same four-operation boundary” without changing it.

Please make the design choose one of these explicitly:

- **Readiness-only prototype:** keep this exact seam, state that the `io_uring` backend is only a poller replacement (`POLL_ADD`/`POLL_REMOVE` + timeout/wait); `accept()` and request I/O remain outside it. Then qualify/remove the later `IORING_OP_ACCEPT`/fixed-buffer claims and make clear sharded listener `poll()` loops are not exercised unless separately migrated.
- **Operation/completion abstraction:** widen the proposed backend boundary to typed submissions/completions for accept/read/write/timer/cancel (for example `submitAccept`, `submitRead`, `submitWrite`, `submitTimer`, `cancel`, `waitCompletions` with opaque tokens), and explain how epoll/kqueue implement/adapt that contract versus `io_uring`.

#148 explicitly asks for an abstraction boundary for accept/read/write/timer readiness. As written, the document promises evaluation of direct completion operations while constraining the implementation to a readiness interface that cannot expose them.

docs/EVENT_LOOP_BACKENDS.md
```diff
@@ -0,0 +1,425 @@
...
+  (e.g. a ring entering a bad state after `io_uring_enter` failures) are not
+  a live-migration trigger back to epoll; they surface through the same
+  error/drain semantics `gateway_shutdown.zig` already uses for other fatal
+  runtime conditions (log, drain, exit), consistent with how the current
+  `EventLoop.wait()` error path already just logs and continues rather than
+  trying to self-heal by rebuilding the poller.
+- **Kernel/toolchain requirements, if pursued:** modern `io_uring` opcodes
+  useful here (`IORING_OP_ACCEPT`, fixed buffers/files, multishot variants)
+  need meaningfully newer kernels than the epoll baseline requires — commonly
+  cited minimums range from 5.5 (basic ring support) through 5.11/5.19 for
```
Jaruso: [P2] `5.5` is not the baseline for “basic ring support.” The `io_uring_setup`/`io_uring_enter`/`io_uring_register` syscalls landed in Linux **5.1**. The later versions belong to individual capabilities: `IORING_OP_ACCEPT` arrived later (5.5), multishot poll is 5.13, and upstream liburing documents multishot accept as available since 5.19. Registered files/buffers also existed in the early interface.

Because kernel requirements are an explicit #148 acceptance criterion, please separate **ring baseline** from **prototype operation requirements**. For example: “io_uring exists from 5.1; the prototype will require Linux >= X because it uses operations/features A/B/C, and will probe those capabilities at startup.” If you do not want to commit to exact per-op numbers in Phase 0, that is safer than calling 5.5 basic support. The important thing is that the eventual minimum comes from the exact SQE/features selected by the prototype.

docs/EVENT_LOOP_BACKENDS.md
```diff
@@ -0,0 +1,425 @@
...
+  handled entirely outside the shared `EventLoop`: each shard thread runs
+  its own independent blocking `poll()` loop
+  (`gaccept.runShardAcceptLoop`/`acceptReadyConnectionsShard` in
+  `src/gateway_accept.zig`) against its own listener fd, and the shared
+  `EventLoop` registers no listener fd at all in that mode — it still
+  handles roles 2 and 3 above. Accepted fds from every shard still enter
+  the same bounded worker pool.
+- Accepted connections are dispatched to a bounded `WorkerPool`
+  (`src/http/worker_pool.zig`) of OS threads. Each worker owns one
+  connection's full synchronous lifecycle — TLS handshake, HTTP parse,
```
Jaruso: [P3] This is still internally inconsistent with the corrected native-TLS description immediately above. A native-TLS connection can yield during `native_handshake`, be stored in `ActiveRegistry`, and later be dispatched/rearmed from `EventLoop`; a single worker therefore does **not** own its TLS handshake as one uninterrupted synchronous lifecycle. The blocking-per-request statement is still valid for the legacy OpenSSL/plaintext path, and the native path becomes locally blocking for request work once checked out, but this blanket sentence reintroduces the model the preceding section just corrected. Please qualify it by path (or say workers own active request processing rather than the entire connection/TLS-handshake lifecycle).

Jaruso (commented):
Second-pass review of 4e07d03. The four findings from my previous review are addressed, and CI is green on this head. I found two remaining design-level issues that should be fixed before merge, plus one documentation consistency issue:

1. **P2 — the declared backend abstraction cannot express the direct `io_uring` operations the document proposes evaluating.** The current `add`/`modify`/`remove`/`wait` seam is a readiness-poller abstraction; it has no operation submission/completion model for accept/read/write/timer results. The design needs to explicitly choose between a readiness-only `io_uring` adapter (e.g. poll SQEs behind the existing seam) and a wider operation/completion abstraction. Otherwise the later discussion of `IORING_OP_ACCEPT`, registered buffers/files, multishot accept, etc. cannot actually be implemented behind the boundary this section declares.
2. **P2 — the kernel-version requirements are inaccurate.** `io_uring` itself landed in Linux 5.1, not 5.5. Registered files/buffers also existed from the early interface; individual networking operations have later floors (for example accept in 5.5, multishot poll in 5.13, multishot accept in 5.19). Because #148 explicitly requires kernel requirements/fallback behavior, the design should separate the ring baseline from the exact operation set the eventual prototype requires instead of calling 5.5 “basic ring support.”
3. **P3 — one stale “full synchronous lifecycle” sentence still contradicts the corrected native-TLS architecture immediately above it.** Qualify that statement to the legacy OpenSSL/plaintext path, or otherwise distinguish native handshake scheduling from the blocking per-request worker path.

Once those are corrected, I don’t see another Phase-0 blocker: keeping #148 open for the prototype/benchmarks/metrics is appropriate, #140 is still open via PR #600, and the updated failure/fallback policy and prior architecture/libxev/metrics corrections are good.

docs/EVENT_LOOP_BACKENDS.md
Jaruso: Fixed in da73233: chose the readiness-only prototype option explicitly. Added a new paragraph stating a future `io_uring` backend targeting this boundary is `IORING_OP_POLL_ADD`/`IORING_OP_POLL_REMOVE` + a ring-native timeout only, standing in for `epoll_wait`/`kqueue`, with `accept()`/request I/O staying outside the ring; qualified/removed the `IORING_OP_ACCEPT`/fixed-buffer/multishot claims elsewhere (they'd require the wider operation/completion abstraction, explicitly out of scope); and stated the sharded per-shard `poll()` loops aren't exercised by this boundary unless separately migrated.

---

docs/EVENT_LOOP_BACKENDS.md
Jaruso: Fixed in da73233: corrected to Linux 5.1 for the ring baseline (`io_uring_setup`/`_enter`/`_register`, which covers the readiness-only `POLL_ADD`/`POLL_REMOVE` operation set this doc now scopes a prototype to), with `IORING_OP_ACCEPT` (5.5), multishot poll (5.13), and multishot accept (5.19) called out as separate, later, op-specific floors that only matter if a future wider abstraction adopts them — not "basic ring support."

---

docs/EVENT_LOOP_BACKENDS.md
Jaruso: Fixed in da73233: qualified to "workers own active request/handshake processing," with the full-uninterrupted-lifecycle claim restricted to the legacy OpenSSL/plaintext path, and an explicit note that a native-TLS connection can yield out of `native_handshake` back to `EventLoop`/`ActiveRegistry` and be re-dispatched to a different worker later.

---

docs/EVENT_LOOP_BACKENDS.md
```diff
@@ -0,0 +1,474 @@
...
+  synchronous blocking calls on worker threads (including within the
+  native-TLS path's own `WaitingEncryptedHttpConnection.waitFor`, which polls
+  locally rather than through the shared loop). Per the "Backend abstraction
+  boundary" section above, a backend targeting this boundary is
+  **readiness-only** (`IORING_OP_POLL_ADD`/`IORING_OP_POLL_REMOVE` plus a
+  ring-native timeout) — it does not use `IORING_OP_ACCEPT`, registered
+  buffers/files, or multishot variants, since those require widening the
+  boundary to an operation/completion abstraction, which is explicitly out
+  of scope for Phase 0. The readiness-only operation set is available from
+  the original `io_uring` interface (Linux 5.1); see "Failure modes and
```
Jaruso: [P2] The newly chosen prototype is not actually a Linux 5.1 operation set as written. The boundary above explicitly says `wait(timeout_ms)` is implemented with **a ring-native timeout** in addition to `IORING_OP_POLL_ADD`/`IORING_OP_POLL_REMOVE`. In the upstream Linux UAPI, v5.1 contains `POLL_ADD` and `POLL_REMOVE` but does **not** contain `IORING_OP_TIMEOUT`; the v5.4 UAPI does contain `IORING_OP_TIMEOUT`.

Please distinguish the ring baseline from the minimum for the *prototype you just selected*. With the design currently written, the simplest accurate contract is something like:

> `io_uring` and `IORING_OP_POLL_ADD`/`POLL_REMOVE` exist from Linux 5.1, but this readiness-only prototype requires Linux >= 5.4 because `EventLoop.wait(timeout_ms)` is implemented with `IORING_OP_TIMEOUT`. Startup probes all three required operations/capabilities.

Then make the failure-policy section state 5.1 as historical ring/poll baseline and >=5.4 as this prototype's actual minimum. Alternatively, keep a 5.1 prototype only if you explicitly choose and document a non-`IORING_OP_TIMEOUT` mechanism for bounded waits. #148 requires the kernel requirement for the proposed prototype itself, not just the oldest kernel on which some of its opcodes exist.

docs/EVENT_LOOP_BACKENDS.md
```diff
@@ -0,0 +1,474 @@
...
+- **TLS/encrypted-stream integration:** Would not change — TLS record
+  handling (both the legacy OpenSSL path and the native pure-Zig path)
+  happens inside the blocking worker once a connection is checked out, not
+  in the event loop, so `io_uring` adoption at the readiness/idle-wait
+  boundary described above would not touch TLS state machines either way.
+- **Proxy streaming and backpressure:** No expected interaction with #139
+  (streaming relay) or #140 (watermark backpressure) as currently designed,
+  since both operate on synchronous blocking reads/writes inside worker
+  threads, not on event-loop-driven readiness.
+- **Timers, cancellation, keepalive parking, graceful drain:** `io_uring`
```
Jaruso: [P2] This tradeoff no longer follows from the readiness-only boundary selected earlier in the same commit. The retained API exposes `add`/`modify`/`remove` plus one bounded `wait(timeout_ms)`; it has no per-connection timer submission/cancellation API. Therefore a ring-native timeout can implement the **global bounded wait**, but it cannot replace `keepalive_park.zig`'s per-connection expiration/reaper model through this abstraction.

Please make this section and the comparison-matrix row match the chosen scope. Expected shape:

> **Timers, cancellation, keepalive parking, graceful drain:** unchanged at the readiness-only boundary. `IORING_OP_TIMEOUT` is used only to preserve `EventLoop.wait(timeout_ms)` semantics; parked-connection expiry continues to use the existing `ParkedRegistry` maintenance reaper, and drain behavior remains unchanged. Moving per-connection timers/cancellation into the ring would require the wider operation/completion abstraction explicitly deferred above.

That distinction matters because #148 specifically requires the comparison to cover timers, cancellation, keepalive parking, and drain. As written, this bullet attributes a potential benefit to the Phase 0 prototype that its own backend interface cannot expose.

Jaruso (commented):
**Review result: two P2 design inconsistencies remain before merge.**

The three findings from my previous pass are addressed correctly in `da73233`: the worker-ownership language is now path-specific, the backend boundary explicitly chooses a readiness-only `io_uring` prototype, and the 5.1 ring baseline is separated from later completion-oriented operations. I rechecked the updated document against #148's acceptance criteria and the current CI run; all applicable CI jobs are green.

Two consequences of the newly chosen readiness-only design still need to be made internally consistent:

1. The prototype as currently specified is **not actually Linux 5.1-compatible** because preserving `EventLoop.wait(timeout_ms)` is explicitly assigned to a ring-native timeout. Linux v5.1 has `IORING_OP_POLL_ADD`/`IORING_OP_POLL_REMOVE`, but not `IORING_OP_TIMEOUT`; that opcode is present by v5.4. Either set the chosen prototype floor to >=5.4, or describe a different bounded-wait mechanism for 5.1.
2. Once Phase 0 chooses the existing `add`/`modify`/`remove`/`wait` readiness seam, per-connection timer/cancellation behavior is not exposed through that abstraction. The statement that `io_uring` could simplify the keepalive reaper therefore no longer describes the prototype selected by this document; at this boundary, the ring-native timeout only implements the bounded `wait`, while the `ParkedRegistry` reaper/drain behavior remains unchanged.

Because this is the PR author's connected account, GitHub does not allow me to submit `REQUEST_CHANGES`; this COMMENT review should still be treated as changes-needed for the P2s below.

docs/EVENT_LOOP_BACKENDS.md
Jaruso: Fixed in db84081, using essentially your suggested wording: `POLL_ADD`/`POLL_REMOVE` exist from 5.1, but this readiness-only prototype's actual minimum is Linux **5.4** because `wait(timeout_ms)` is implemented with `IORING_OP_TIMEOUT`. Corrected in the backend-abstraction section, the io_uring section, and the failure-modes/kernel-requirements section, which now states 5.1 as the historical ring/poll baseline and 5.4 as this prototype's real minimum.

---

docs/EVENT_LOOP_BACKENDS.md
Jaruso: Fixed in db84081, using your suggested wording almost verbatim, in both the prose bullet and the comparison-matrix row: unchanged at the readiness-only boundary, `IORING_OP_TIMEOUT` only preserves the single bounded `wait()`, `ParkedRegistry`'s reaper and drain are untouched, and a real reaper simplification would need the wider operation/completion abstraction this doc already defers.

---

src/http/event_loop.zig
```diff
@@ -223,6 +265,381 @@ pub const EventLoop = struct {
...
+        // codebase relies on that: both `parkedConnectionCloseHook` and
+        // `activeConnectionCloseHook` only do accounting, so the keepalive idle
+        // reaper and the active-connection reaper close registered fds without
+        // unregistering them first. io_uring has no equivalent implicit
+        // cleanup, so an entry still present here belongs to a closed fd whose
+        // number the kernel has since recycled. Cancel that stale poll and take
+        // the fd over, which is what the caller would have observed under
+        // epoll — rather than reporting EEXIST for a connection it considers
+        // brand new and failing to register it at all.
+        if (self.registrations.getPtr(fd)) |stale| {
```
Jaruso: [P1] A duplicate numeric fd cannot safely be treated as proof that the existing registration is stale. `POLL_ADD` holds a kernel reference to the original file, so `close(fd)` does not cancel that request. Today `ParkedRegistry.reapIdle()`/`closeAll()` can close a registered parked fd without calling `EventLoop.remove` first. If the process reuses that integer for an unrelated descriptor and the *old* poll completes before another `add(fd, ...)` happens, `reap()` still finds this map entry/token as live and immediately re-arms **the recycled numeric fd**, attaching the poll to the unrelated new file. An always-ready replacement socket can then produce a spurious/busy readiness loop even though it is not in the parked/active registries.

Please make unregister-before-close an invariant instead of trying to infer staleness here. For example, give the parked close hook access to the loop and remove before the fd is closed:

```zig
const ParkedCloseCtx = struct {
    state: *GatewayState,
    event_loop: *http.event_loop.EventLoop,
};

fn parkedConnectionCloseHook(raw: *anyopaque, fd: std.posix.fd_t) void {
    const ctx: *ParkedCloseCtx = @ptrCast(@alignCast(raw));
    ctx.event_loop.remove(fd) catch {}; // before freeConn closes fd
    ctx.state.releaseConnectionSlot(fd);
}
```

Then this `add` path should preserve normal duplicate-registration semantics rather than silently taking over any existing entry. Also replace the current “recycled fd” test: it only calls `add` twice on the same still-open socket, so it never exercises close/reuse. Force a real close + fd-number reuse (or test the parked close hook end-to-end) and verify no old completion can be reported or re-armed against the replacement descriptor. The PR body/docs should also be corrected: the active expiry reaper already calls `event_loop.remove(conn.fd)` before `conn.deinit()`; the parked reaper is the path relying on epoll's implicit close cleanup.

src/http/event_loop.zig
```diff
@@ -223,6 +265,381 @@ pub const EventLoop = struct {
...
+            // `io_uring_setup` is commonly blocked outright by container
+            // seccomp profiles (Docker's default profile denies it), which
+            // surfaces as EPERM/ENOSYS rather than as a missing opcode.
+            error.PermissionDenied, error.SystemOutdated => return error.IoUringUnavailable,
+            error.SystemResources => return error.IoUringRingAllocationFailed,
+            else => return error.IoUringUnavailable,
+        };
+        errdefer ring.deinit();
+
+        try probeCapabilities(&ring);
```
Jaruso, line 425: [P1] Opcode probing is not sufficient for the documented Linux 5.4 floor. Linux 5.4 can **drop CQEs when the completion queue is full** (it increments `cq_overflow` instead); `IORING_FEAT_NODROP` and the overflow-list preservation behavior arrive in 5.5. That is correctness-critical for this adapter because each `POLL_ADD` is one-shot and is only re-armed after its CQE is reaped. If a burst of ready fds overflows the CQ on 5.4, the kernel can consume a poll and discard its completion while `registrations` still says that fd is armed, so that fd can disappear from readiness forever.

Please either (a) require the setup feature `IORING_FEAT_NODROP` at startup and document the prototype floor as 5.5+ (feature-check it rather than relying only on a version number), or (b) explicitly detect/recover from CQ overflow by reconstructing/re-arming affected live registrations. The feature-gate is much simpler and safer for a prototype. I would also add a stress regression with more simultaneous one-shot completions than CQ capacity so the overflow/backlog path is exercised; the current handful-of-socketpair tests cannot expose this class of loss.

src/http/event_loop.zig
```diff
@@ -223,6 +265,381 @@ pub const EventLoop = struct {
...
+                };
+                // `count = 1` makes the timer complete as soon as one other
+                // completion is posted, or when `ts` elapses — the same shape
+                // as `epoll_wait(timeout)`, and self-completing either way so
+                // no dangling timer has to be cancelled afterwards.
+                if (self.ring.timeout(timeout_user_data, &ts, 1, 0)) |_| {
+                    timer_armed = true;
+                } else |_| {}
+            }
+            self.submitLocked() catch {};
```
Jaruso: [P1] Do not swallow this submission failure. `timer_armed` currently means only that `ring.timeout(...)` obtained/filled an SQE. `submitLocked()` still has to flush and successfully enter the kernel. If that submission fails, this `catch {}` leaves `timer_armed = true`, so the code below may call `io_uring_enter(0, 1, GETEVENTS)` believing a bounded timeout exists even though the timeout was never submitted; on an otherwise idle ring the event-loop thread can then block indefinitely past its maintenance/shutdown deadline. `submitLocked()` also treats `SignalInterrupt` as success, which has the same bookkeeping problem when no SQEs were actually submitted.

This is part of a broader runtime-error issue in the same adapter: `reap()` also does `self.armLocked(fd, reg) catch continue` and `self.submitLocked() catch {}`. If that happens while re-arming the unsharded listener, there is no per-connection deadline reaper to recover it—the server can simply stop receiving listener readiness while staying up.

Please make successful kernel submission part of the state transition and propagate failures instead of suppressing them, e.g. conceptually:

```zig
if (timeout_ms > 0) {
    try self.ring.timeout(...);
    try self.submitLocked();
    timer_submitted = true;
}
...
try self.armLocked(fd, reg);
try self.submitLocked();
```

Then carry an unrecoverable ring error into the fail/drain/exit policy documented for this backend. Merely returning it from `wait()` is not quite enough today because `edge_gateway` logs `event_loop.wait` errors and `continue`s; the io_uring path needs a defined fatal transition rather than silently running with missing polls. If you intentionally degrade a bounded wait to non-blocking on queue pressure, set the flag only after submission failure is known and do not enter a blocking wait.

Jaruso (commented):
**Review result: changes needed before merge.** The two P2s from my prior pass are fixed correctly, but the PR has since expanded from a design-doc change into the actual Phase 1 `io_uring` runtime backend, so I re-reviewed the implementation as new scope against #148. The default backend remains unchanged, the opt-in/fail-closed configuration is wired correctly, and I confirmed the Linux CI tests really execute rather than all skipping. However, the new adapter has three correctness failures I would treat as P1 merge blockers:

1. Closing a registered fd without `EventLoop.remove` can let a late completion re-arm a completely unrelated descriptor that later reused the same fd number. The current duplicate-`add` takeover does not close that race. The parked idle reaper/`closeAll` path needs an explicit unregister-before-close invariant (the active expiry reaper already does this), and the recycled-fd test must actually close and reuse a descriptor.
2. The documented Linux 5.4 floor is unsafe for this one-shot/re-arm design because 5.4 can drop CQEs when the CQ is full. Since a delivered `POLL_ADD` is re-armed only after its CQE is consumed, a dropped CQE can permanently strand a registration. Require `IORING_FEAT_NODROP` (effectively raising the floor to 5.5 unless recovery is implemented) or handle CQ overflow explicitly, and add a CQ-pressure regression test.
3. Runtime submission/re-arm errors are swallowed. In particular, `wait()` can mark a timeout as armed, fail to submit it, ignore that failure, then block in `io_uring_enter(..., min_complete=1)` with no kernel timeout to wake it. The same pattern can silently lose a listener re-arm. Runtime ring failures need to propagate into the backend's defined fatal/drain behavior rather than leave the process alive with a wedged event loop.

The rest of the new configuration/wiring and docs are directionally consistent with #148. Benchmark/profiling evidence and the backend-labelled metric remain correctly tracked as follow-up work under #148; this PR still does not close the issue. CI is green on the current head, including Linux x86_64/arm and macOS.

src/http/event_loop.zig
```diff
@@ -223,6 +265,435 @@ pub const EventLoop = struct {
...
+            // Stale completion for a registration that was removed or
+            // re-armed under a new interest.
+            if (reg.token != decodeToken(cqe.user_data)) continue;
+            // A negative result is a negated errno, not a poll mask. The
+            // common case is -ECANCELED, though a cancel normally also
+            // removes or re-tokenizes the registration and so is already
+            // filtered above. Anything else is a poll the kernel refused;
+            // like the epoll path's error handling, this reports no readiness
+            // and does not try to self-heal by re-arming.
+            if (cqe.res < 0) continue;
```
Jaruso: [P1] A negative CQE for the **current live registration token** cannot be ignored here. `POLL_ADD` is single-shot: once its CQE is posted, that request is terminated, and liburing documents operation failures as a negated errno in `cqe.res`. The stale/cancel cases are already handled above: `remove` deletes the map entry, and `modify` changes the token before the old cancellation CQE can be accepted. Therefore, if execution reaches this line with a matching live token and `res < 0`, the registration table says the fd is armed even though no poll remains. The listener is the worst case: the process can stay healthy-looking while never receiving another accept readiness notification.

This should follow the same fail-fast invariant introduced for submission/re-arm failures. For example:

```zig
if (cqe.res < 0) {
    _ = self.registrations.remove(fd);
    return self.markFatal();
}
```

(or classify a specifically recoverable error and explicitly re-arm it, but silently continuing is not safe). Please add a regression for the asynchronous-operation-error path as well. A simple deterministic shape is to close a socket fd, submit a poll for that now-invalid fd, let `io_uring_enter` accept the SQE, and assert that the resulting `-EBADF` CQE makes `wait()` return `error.EventLoopUnrecoverable` rather than `0` while leaving a live registration behind.

Jaruso (commented):
**Review result: one remaining merge-blocking runtime issue.** The new commit substantially fixes all three P1 findings from my previous pass, and I resolved those three threads: parked connections now unregister before close with a real fd-reuse regression; startup requires `IORING_FEAT_NODROP` and the CQ-pressure test exercises the overflow backlog; and submission/re-arm failures now become a sticky `EventLoopUnrecoverable` state that the gateway turns into shutdown/drain instead of logging and continuing.

I re-read the modified backend/fatal paths against #148 and Zig 0.16's `IoUring` queue semantics. One P1 remains in the completion path: a negative CQE for the **current live poll token** is silently discarded. At that point the one-shot `POLL_ADD` has terminated, so leaving the registration in the map without re-arming or failing the loop recreates the same healthy-looking-but-deaf state the latest commit was designed to prevent. See the inline comment for the expected fix/test.

I also left two P3 consistency comments. The source-level architecture comment still describes the now-removed stale-fd takeover and a 5.4 floor, and the validation section still says six Linux io_uring tests; the current Linux/macOS skip differential is 8 (Linux: 1 skip, macOS: 9 skips) after the two new regressions. The PR body has the same stale 5.4/takeover/six-test wording and should be reconciled too.

#148 itself should remain open after this PR: the issue still requires real-hardware benchmark/profiling evidence and the backend-labelled event-loop metric before closure. This PR correctly keeps the default epoll/kqueue path unchanged.

CI on this head is currently red only in `Test appliance profile (ubuntu-latest)`, where the failing test is `http.upstream_h2.test.response capacity exhausted during an upload surfaces before commitment`. `src/http/upstream_h2.zig` is unchanged by this PR, while the normal Linux and ARM test jobs, macOS tests, builds, format/lint, perf smoke, packaging, TLS/H3 interop, and the ARM appliance job all pass. I do not attribute that existing/flaky upstream_h2 failure to #602.

src/http/event_loop.zig
```diff
@@ -223,6 +265,435 @@ pub const EventLoop = struct {
...
+/// Emulates the level-triggered `add`/`modify`/`remove`/`wait` contract the
+/// rest of the gateway already programs against, using only
+/// `IORING_OP_POLL_ADD`, `IORING_OP_POLL_REMOVE` and `IORING_OP_TIMEOUT`.
+/// Two behaviours of the ring have to be papered over to match `epoll`:
+///
+///  1. **`POLL_ADD` is one-shot.** `epoll` in level-triggered mode keeps
+///     reporting a fd until the caller drains or removes it, and callers rely
+///     on that (the listener fd is registered once and never re-added). Every
+///     delivered completion is therefore immediately re-armed inside `wait`.
+///  2. **Closing a fd does not unregister it.** `epoll` drops a fd from every
```
Jaruso: [P3] This top-level implementation comment is now stale after the fd-lifetime fix. It still says the close hooks only release accounting and that `add` treats an existing fd as stale and takes it over, but the new invariant is the opposite: parked teardown unregisters before close, active expiry already unregisters before `deinit`, and `add` now preserves duplicate-registration/EEXIST semantics. The same block later still calls 5.4 the documented floor even though `IORING_FEAT_NODROP` raises the implemented floor to 5.5. Please reconcile this comment with the updated design doc so future maintainers do not implement against the superseded behavior.

docs/EVENT_LOOP_BACKENDS.md
```diff
@@ -0,0 +1,718 @@
...
+  simultaneously against a 128-entry completion queue and requires every one
+  to surface. The behaviour tests skip themselves when `io_uring_setup` is
+  unavailable — container seccomp profiles, Docker's default among them, block
+  it outright — since that is an environment property, not a defect in the
+  backend.
+- **Executed on Linux by CI, not by the author.** The development host for
+  this change was macOS/arm64, where the io_uring path compiles out entirely;
+  it was validated there only by cross-compiling for `x86_64-linux-gnu` and by
+  review. The Linux CI jobs were the first actual execution, and all six
+  io_uring tests ran and passed there on both `ubuntu-latest` (x86_64) and
```
Jaruso, line 585: [P3] The validation count did not get updated when the two new Linux regressions were added. On the current head, the normal Linux job reports `992 pass, 1 skip` in this test group while macOS reports `984 pass, 9 skip`, so the differential is **8**, not 6. That matches the Linux-only ring-size test plus the seven live-io_uring behavior/regression tests. Please update this paragraph to say eight tests / macOS 9 skips vs Linux 1 skip, and reconcile the PR body as well (it still says six tests, a 5.4 floor, and describes the removed stale-fd takeover behavior).

Jaruso: Partially fixed in `ebbe23d`: the heading now correctly says eight Linux-only tests, the skip differential is 9 vs 1 (=8), and the PR body is reconciled to the 5.5 floor / unregister-before-close behavior. One sentence in this same paragraph still says **“all six io_uring tests ran and passed”**. Please change that to **“all eight io_uring tests ran and passed”** so the paragraph is internally consistent. No code change needed for this remaining P3.

Jaruso (commented):
Re-review of `ebbe23d` against the remaining findings and #148 acceptance criteria.

The code-level P1 is fixed correctly. A negative CQE on the current live registration token now removes the registration and transitions the ring to sticky `EventLoopUnrecoverable`, and the new closed-fd regression exercises the asynchronous failed-completion path. I also re-checked the surrounding cancellation/token logic; `remove` deletes the registration before cancellation completes and `modify` re-tokenizes before the old completion can be accepted, so the new live-token test is scoped correctly rather than misclassifying normal cancellation CQEs.

The stale implementation-comment P3 is also fixed, and I resolved both of those prior threads.

One P3 documentation consistency issue remains in the existing validation-count thread: the section now correctly says eight Linux-only tests and correctly reports the 9-vs-1 skip differential, but the next sentence still says “all six io_uring tests ran and passed.” I replied on that thread with the exact one-word-number correction. This is not a code blocker.

I re-checked #148: this PR still intentionally does not close it. The design/prototype/default-runtime/failure-mode requirements addressed by this PR are represented, while the issue remains open for the required real-hardware benchmark/profiling evidence and `tardigrade_event_loop_iterations_total{backend=...}` metric work. That is consistent with the issue rather than silently redefining closure criteria.

CI is still running on this head. The ARM appliance-profile job is currently red in the unchanged `src/http/upstream_h2.zig` test `origin capacity bounds concurrent streams and refuses only the overflowing one` (`expected 3, found 0`), the same unrelated flaky area seen in earlier rounds. Builds, Format, Lint, Security, macOS tests, H3/TLS checks and other completed jobs are green so far; the normal Linux/ARM test jobs were still in progress when this review was submitted.

**Review result:** no remaining code-level findings from this pass. Fix the one P3 docs count typo, and treat CI independently—if the remaining relevant tests finish green / the known unrelated appliance flake is rerun successfully, I have no current implementation blocker to merging this prototype PR.

Jaruso (commented):
Re-reviewed at `cfab69c` against the previous review and #148 scope. The only change since `ebbe23d` is the requested documentation correction from “all six io_uring tests” to “all eight io_uring tests”; the validation paragraph is now internally consistent with the 9-vs-1 skip differential and the PR body already reflects the Linux 5.5 floor, NODROP requirement, unregister-before-close invariant, and fail-fast negative-CQE handling.

I resolved the final outstanding review thread. No code changed in this commit, and I found no remaining implementation or documentation findings in the PR. The prior code-level blockers remain correctly addressed: fd teardown unregisters before close, CQ completion loss is gated by `IORING_FEAT_NODROP`, submission/re-arm failures become sticky/fatal, and a negative completion for a live one-shot poll now fails fast with a regression covering that path.

#148 is correctly left open: this PR still does not provide the benchmark/profiling evidence or backend-labeled event-loop metric required for issue closure, and the PR explicitly tracks those as follow-up work rather than claiming to close the issue.

**Review result: merge-ready from code-review perspective.** I cannot submit APPROVE on my own PR, so this is recorded as COMMENT. The fresh CI run for this docs-only head is currently queued; merge should still wait for required checks to finish.

PR #603, opened 2026-08-11
feat(tls): negotiate record_size_limit and add bounded record padding (#359)

Jaruso:
## Summary

Implements RFC 8449's `record_size_limit` extension on the native TLS record transport, plus a local, bounded RFC 8446 §5.4 padding policy. New module: [record_size.zig](src/tls/record_size.zig) owns the wire form, the range rules, and the sizing/padding arithmetic — no I/O, no keys.

**Negotiated sizing.** Each side advertises the largest `TLSInnerPlaintext` it will *receive*. Client offers in ClientHello (identically positioned in CH1 and CH2 so the HRR "legal mutation" check still passes); server answers in EncryptedExtensions on both the full and PSK-resumed flights.

- Outbound: application data, handshake fragments, and post-handshake messages are all sized from `Bridge.outboundContentMax()`. Sealing past it fails closed with `RecordSizeLimitExceeded` rather than silently truncating.
- Inbound: a record whose inner plaintext exceeds our own advertisement is refused **before** the AEAD open — the point of the bound is to cap the work an unauthenticated peer can cause — and maps to `record_overflow` (newly added to `AlertDescription`).
- Default is the TLS 1.3 protocol maximum (`policy.Policy.record_size_limit`), so the extension is offered but constrains nothing until an operator lowers it. Peers that omit it mean "protocol maximum", per the RFC.
- Out-of-range handling follows RFC 8449 §4's asymmetry exactly: `< 64` is `illegal_parameter` for both roles; above the maximum a **server clamps** (a client may be advertising a size a future version enables) and a **client rejects** (the version is already settled).
- Unprotected records stay exempt. TLS-over-QUIC has no TLS records and neither offers nor accepts the extension (`unsupported_extension`); the extension ID is also reserved against transport-profile collision.

**Padding.** `PureZigRecordStream.setRecordPadding` rounds each application record's inner plaintext up to a block boundary so length no longer reveals content length. Off by default; never applied to handshake or alert records; clamped by the negotiated limit rather than the reverse, so it can never push a record past what the peer accepts — including an already-at-cap record, which gets none.

**Observability.** `recordSizeLimits()` and `recordSizeCounters()` report the negotiated state and the effects (peer-limited writes, padded records, padding bytes, oversize records rejected).

**Docs.** [TLS_RECORD_TRANSPORT.md](docs/TLS_RECORD_TRANSPORT.md) gains a section separating privacy padding from the negotiated bound, and states explicitly that neither is record coalescing (not implemented; `writePlaintext` seals exactly what it accepts). [TLS_INTEROP_MATRIX.md](docs/TLS_INTEROP_MATRIX.md) records why there is no interop row — traced against OpenSSL 3.6.2, it does not implement RFC 8449 in either role, so the positive rows cover the "peer omits the extension" path instead.

### Goldens repinned

The record-profile transcript goldens and the #485 HRR KAT moved: both ClientHellos now carry the new extension, so every transcript-derived secret below them legitimately changed. The **extension-profile goldens are untouched**, which is exactly the property that test exists to hold. Both call sites carry a comment explaining the repin.

## Test plan

- [x] `zig build test` — full suite green
- [x] `zig build test-tls-record-fuzz`, `zig build test-quic`
- [x] `record_size.zig`: decode round-trip, 2-byte framing, `< 64` for both roles, clamp-vs-reject above max, exhaustive proof that `content + type + padding <= cap` across every block size × content length × cap
- [x] `record_epoch_bridge.zig`: boundary and off-by-one on both send and receive, read sequence unmoved by a pre-AEAD rejection, handshake flight fragmenting at the negotiated limit, padding bounded at the cap, at-capacity record gets no padding, config validation
- [x] `tls13_backend.zig` / `tls13_backend_tests.zig`: ClientHello carries a well-formed advertisement, QUIC profile does not, client accepts both boundaries and rejects out-of-range and mis-framed bodies, server clamps above-max and rejects below-min, absent advertisement, QUIC server rejects the extension, end-to-end negotiation of two distinct limits
- [x] `encrypted_stream.zig`: driver-owned handshake caps every write at the peer's limit and the peer accepts all of them, padding round-trips within the cap, oversize inbound fails with `record_overflow`
- [x] `zig build test-integration` — one pre-existing failure (`reload while serving short static requests…`, #170), confirmed identical on the base commit `fada721c` and unrelated to TLS

Closes #359

### DIFF
--- a/docs/TLS_INTEROP_MATRIX.md
+++ b/docs/TLS_INTEROP_MATRIX.md
@@ -171,6 +171,28 @@ negotiated suite's own HKDF, which the positive matrix already sweeps for
 every suite. What is peer-specific here is the message exchange, not the
 arithmetic.
 
+### Record size limit (RFC 8449)
+
+There is deliberately **no** dedicated row for `record_size_limit` (#359).
+OpenSSL does not implement RFC 8449 — traced against OpenSSL 3.6.2, its
+ClientHello and EncryptedExtensions carry no extension 28 in either role — so a
+row asserting a negotiated limit would assert something no peer in this harness
+can do.
+
+What the existing positive rows do cover is the case the RFC defines for
+exactly that situation: a peer that never sends the extension means "the
+protocol maximum", not "no limit". Every OpenSSL row therefore exercises the
+native side offering the extension (its default advertisement is the protocol
+maximum) to a peer that ignores it, and completing normally — which is the
+interoperability property that matters for enabling it by default.
+
+Coverage for a genuinely negotiated limit lives in-tree, where both sides can
+be driven: wire encode/decode and the role-asymmetric out-of-range handling in
+`src/tls/tls13_backend.zig` and `src/tls/tls13_backend_tests.zig`, and the
+sealing/opening bounds in `src/tls/record_epoch_bridge.zig` and
+`src/tls/encrypted_stream.zig`. If a future peer in this matrix implements RFC
+8449, a row belongs here.
+
 ### Negative conformance
 
 Each of these asserts the failure **class**, using the RFC 8446 §6 alert where
--- a/docs/TLS_INTEROP_MATRIX.md
+++ b/docs/TLS_INTEROP_MATRIX.md
@@ -173,25 +173,43 @@ arithmetic.
 
 ### Record size limit (RFC 8449)
 
-There is deliberately **no** dedicated row for `record_size_limit` (#359).
-OpenSSL does not implement RFC 8449 — traced against OpenSSL 3.6.2, its
-ClientHello and EncryptedExtensions carry no extension 28 in either role — so a
-row asserting a negotiated limit would assert something no peer in this harness
-can do.
+Two GnuTLS rows (#359) assert *negotiated* record sizing. OpenSSL does not
+implement RFC 8449 — traced against OpenSSL 3.6.2, its ClientHello carries no
+extension 28 — but GnuTLS does, and `gnutls-cli`/`gnutls-serv` expose
+`--recordsize` to advertise a non-default value. That makes these the rows
+where the extension is exercised against an independent implementation rather
+than against ourselves.
 
-What the existing positive rows do cover is the case the RFC defines for
-exactly that situation: a peer that never sends the extension means "the
-protocol maximum", not "no limit". Every OpenSSL row therefore exercises the
-native side offering the extension (its default advertisement is the protocol
-maximum) to a peer that ignores it, and completing normally — which is the
-interoperability property that matters for enabling it by default.
+One caveat worth knowing when reading them: GnuTLS's `--recordsize N`
+advertises **N+1** on the wire, because RFC 8449 measures the complete
+`TLSInnerPlaintext` and GnuTLS adds the content-type byte itself. So
+`--recordsize 1024` negotiates a 1025-byte limit, which is what the rows pin.
 
-Coverage for a genuinely negotiated limit lives in-tree, where both sides can
-be driven: wire encode/decode and the role-asymmetric out-of-range handling in
-`src/tls/tls13_backend.zig` and `src/tls/tls13_backend_tests.zig`, and the
-sealing/opening bounds in `src/tls/record_epoch_bridge.zig` and
-`src/tls/encrypted_stream.zig`. If a future peer in this matrix implements RFC
-8449, a row belongs here.
+| Row | Peer | What it proves |
+| --- | --- | --- |
+| `record/server/gnutls/record-size-limit` | `gnutls-cli --recordsize 1024` | **we honor the peer's limit**: a 3000-byte response leaves as ≥5 protected records, none with an inner plaintext above 1025 |
+| `record/client/gnutls/record-size-limit` | `gnutls-serv --recordsize 1024` | **the peer honors ours**: we advertise 512, and GnuTLS fragments its ~780-byte response to fit while still delivering every byte |
+
+The two directions are deliberately split across the rows because neither
+stands in for the other. A row that only checked our own sending would pass
+against a peer that ignored the extension completely; a row that only checked
+what arrived would pass without our sizing logic ever running.
+
+Both assert on real record sizes rather than on a completed handshake. The
+server row pins the peer's advertised value, a per-record upper bound, and a
+minimum record count — so a build that negotiated the extension and then
+ignored it fails on fragmentation, and one that stopped sending fails on the
+record count. The client row pins the received-record bound together with a
+minimum byte count, so a peer that "honored" the limit by truncating its
+response fails too. Both bounds also refuse to pass vacuously: a row where no
+protected record was sealed (or none arrived) is a failure, not a silent
+zero-versus-limit comparison.
+
+Coverage for the parts no peer can drive — the role-asymmetric handling of an
+out-of-range advertisement, the request/response rule for the server's
+EncryptedExtensions answer, and the offer-versus-negotiated distinction — lives
+in-tree, in `src/tls/tls13_backend.zig`, `src/tls/tls13_backend_tests.zig`,
+`src/tls/record_epoch_bridge.zig`, and `src/tls/encrypted_stream.zig`.
 
 ### Negative conformance
 
--- a/docs/TLS_RECORD_TRANSPORT.md
+++ b/docs/TLS_RECORD_TRANSPORT.md
@@ -113,3 +113,91 @@ immediate application work; raw socket registration is derived only from
 `wants_read` and `wants_write`. A plaintext read blocked on a TLS write retry
 therefore registers only write interest, and a plaintext write blocked on a TLS
 read retry registers only read interest.
+
+## Record sizing and padding (#359)
+
+Two independent knobs sit on the record path. They are easy to confuse and do
+opposite things, so this section names them apart explicitly.
+
+### `record_size_limit` — negotiated, bounds memory
+
+RFC 8449's `record_size_limit` extension (type 28) is a *negotiated* bound.
+Each endpoint advertises the largest `TLSInnerPlaintext` it is willing to
+**receive**; the peer must not send a protected record larger than that.
+The value covers the complete inner plaintext — content, the content-type
+byte, and any padding — which is why the TLS 1.3 maximum is 2^14+1 (16385)
+rather than 2^14, and why usable content is always one byte less than the
+advertised limit.
+
+* Record transport only. TLS-over-QUIC has no TLS records, so the QUIC
+  profile neither offers the extension nor accepts one: a peer that sends it
+  gets `unsupported_extension`.
+* The client offers it in ClientHello (in both ClientHello1 and ClientHello2,
+  at the same position, so an HRR's "ClientHello2 is a legal mutation of
+  ClientHello1" check still passes). The server answers in
+  EncryptedExtensions, on the full and PSK-resumed flights alike.
+* Configured by `policy.Policy.record_size_limit`. The default is the
+  protocol maximum: the extension is offered but constrains nothing, which is
+  the most interoperable choice and avoids inflating per-record overhead for
+  connections that never needed smaller records. Lowering it is a deliberate
+  memory/latency tradeoff.
+* Values below 64 are rejected as `illegal_parameter` by both roles. Above
+  the protocol maximum the handling is asymmetric, exactly as RFC 8449 §4
+  requires: a **server clamps** (a client may be advertising a size a future
+  version enables, and a server must not enforce the restriction), while a
+  **client rejects** — by EncryptedExtensions the negotiated version is
+  already settled, so no such reading is available.
+* Enforcement is symmetric in `record_epoch_bridge`. Outbound, every
+  protected write and every protected handshake fragment is sized by
+  `Bridge.outboundContentMax()`; sealing more than that fails closed with
+  `RecordSizeLimitExceeded` rather than silently truncating. Inbound, a
+  record whose inner plaintext would exceed our own advertisement is refused
+  *before* the AEAD open — the point of the bound is to cap the work an
+  unauthenticated peer can cause — and maps to the `record_overflow` alert.
+* Unprotected records are exempt (RFC 8449 §4). The initial-epoch plaintext
+  ClientHello/ServerHello keeps the full protocol fragment.
+
+A very small advertised limit interacts with the fixed outbound queue: a large
+server flight (certificate chain plus CertificateVerify) split into 63-byte
+fragments pays per-record overhead on every fragment. The exact requirement is
+preflighted by `sealedHandshakeLen`, so the failure mode is a deterministic
+refusal to progress, never an overflow — but an operator choosing a limit near
+RFC 8449's floor should size the certificate chain accordingly.
+
+### Padding — local, buys privacy, costs bandwidth
+
+RFC 8446 §5.4 record padding is *not* negotiated and has no wire signal. It
+appends zero bytes to the inner plaintext so that an on-path observer cannot
+read the true content length out of the record length. It is a
+traffic-analysis countermeasure, and it is the only thing here that is.
+
+* Configured per stream with `PureZigRecordStream.setRecordPadding`. Off by
+  default: every padding byte costs bandwidth and AEAD work and buys nothing
+  except length uniformity.
+* The only policy is `.block = N`, which rounds each inner plaintext up to a
+  multiple of `N`. A fixed target size was rejected because it cannot pad a
+  record that is already larger and inflates a small-record stream by a
+  constant factor; rounding degrades smoothly.
+* Applied only to `application_data`. Handshake messages have protocol-fixed
+  lengths that are already visible in the transcript, so padding them inflates
+  a latency-critical flight for no privacy; alerts are two bytes and their size
+  should not vary at all.
+* Padding is bounded by the negotiated limit, never the other way around: the
+  target length is clamped to `record_size_limit` before the padding is
+  derived from it, so `content + type + padding` is within the cap for every
+  input, including a record that is already at the cap (which gets none).
+
+Neither knob is record *coalescing*. Coalescing — merging several small
+application writes into one record to amortize the 22-byte header/tag
+overhead — is a throughput optimization that changes when bytes leave, and it
+is not implemented here: `writePlaintext` seals exactly what it accepts, one
+record per call. Padding makes records *larger* on purpose and never merges
+them; a smaller `record_size_limit` makes them *more numerous*. Reading either
+as a performance feature gets the tradeoff backwards.
+
+### Observability
+
+`PureZigRecordStream.recordSizeLimits()` reports the negotiated state and
+`recordSizeCounters()` the effects: writes narrowed by the peer's limit,
+records that carried padding, total padding bytes, and inbound records refused
+for exceeding our own advertisement.
--- a/docs/TLS_RECORD_TRANSPORT.md
+++ b/docs/TLS_RECORD_TRANSPORT.md
@@ -161,10 +161,22 @@ advertised limit.
   to fails with `record_overflow`. When the extension does not negotiate, the
   bound stays at the protocol maximum and the mark can never exceed it, so the
   measurement clears itself.
-* The retained mark covers handshake-epoch records only. A server activates when
-  it *writes* its answer, and 0-RTT records can already be in flight by then —
-  a client cannot honor a bound it has not received, so failing those
-  retroactively would reject legitimate early data.
+* **0-RTT is exempt from the bound entirely**, not merely from the retroactive
+  settle. RFC 8449 §4 renegotiates the limit on resumption and governs records
+  by the limits of the handshake that produced their protection keys; RFC 8446
+  §4.2.10 puts early data in the client's first flight, so it is created — and
+  may already be buffered in flight — before the server's answer exists. A
+  server activates its bound when it *writes* EncryptedExtensions, which is
+  before the handshake completes, so an early record can legitimately arrive at
+  a server already enforcing 512. Judging it against that value would reject
+  early data a conforming client had no way to size correctly.
+
+  The fuller answer would be to honor the *previous* session's bound, but
+  `ResumableSessionCommon` persists no record-size limit today, so there is no
+  PSK-associated value to apply. Until one exists, early records keep the
+  protocol maximum, which `record_codec` already enforces. The exemption is
+  scoped to the epoch, not the connection: the same record size is refused as
+  soon as it arrives at the handshake or application epoch.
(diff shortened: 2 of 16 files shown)


src/tls/tls13_backend.zig
```diff
@@ -4943,6 +5060,12 @@ pub const Tls13Backend = struct {
             try w.u16_(@intCast(payload.len));
             try w.bytes(payload);
         }
+        // #359: RFC 8449 §4 — "In TLS 1.3, the server sends the
+        // `record_size_limit` extension in the EncryptedExtensions message."
+        // Sent unconditionally when the profile offers it, not only in reply
+        // to a client that asked: the value bounds what the *server* accepts,
+        // and the client needs it whether or not it advertised one itself.
+        try self.writeRecordSizeLimitOffer(&w);
```
Jaruso: **Blocking:** this server response cannot be unconditional. RFC 8446 §4.2 requires extension responses to correspond to a client request; RFC 8449 puts the server's response in EE, but does not override that rule. If the ClientHello omitted extension 28, sending it here can make a conforming client abort with `unsupported_extension`.

`peer_record_size_limit` already gives us the offer-presence signal, so this should be conditional in **both** server EE builders, e.g.:

```zig
if (self.peer_record_size_limit != null) {
    try self.writeRecordSizeLimitOffer(&w);
}
```

Please add full-handshake and PSK-resumption tests that inspect EE and prove extension 28 is absent when the client did not offer it.

src/tls/tls13_backend.zig
```diff
@@ -3087,6 +3118,65 @@ pub const Tls13Backend = struct {
...
+    fn recordSizeLimitEncodedLen(self: *const Tls13Backend) usize {
+        return if (self.offeredRecordSizeLimit() == null) 0 else 2 + 2 + record_size.body_len;
+    }
+
+    /// The connection's negotiated record-size state (#359), for the record
+    /// transport. `local` is what we advertised — the bound we enforce on
+    /// arriving records — and `peer` is what the peer advertised, or null
+    /// until (or unless) it does.
+    pub fn recordSizeLimits(self: *const Tls13Backend) record_size.Limits {
+        return .{
```
Jaruso, line 3206: **Blocking:** `local` here is an *offer*, not necessarily an active negotiated receive bound. With `policy.record_size_limit = 512`, a Tardigrade client talking to a peer that ignores RFC 8449 will currently refresh the bridge to `local=512` and reject a perfectly legal >512-byte protected record. RFC 8449 §4 says that if the extension is not negotiated, normal protocol record sizes remain permitted.

Please track whether extension 28 actually became active and return the configured local value only then, for example conceptually:

```zig
.local = if (self.record_size_limit_negotiated)
    self.policy.record_size_limit
else
    record_size.default_limit,
```

The role timing matters: a client offer alone must not activate its inbound bound; the server EE response confirms it. On the server, once the CH offered extension 28 and we are going to reply, the peer-advertised send bound still needs to be available early enough to fragment the protected server handshake flight. Add an omitted-extension regression using a small configured local value so this cannot regress silently.

docs/TLS_INTEROP_MATRIX.md
```diff
@@ -171,6 +171,28 @@ negotiated suite's own HKDF, which the positive matrix already sweeps for
 every suite. What is peer-specific here is the message exchange, not the
 arithmetic.
 
+### Record size limit (RFC 8449)
+
+There is deliberately **no** dedicated row for `record_size_limit` (#359).
+OpenSSL does not implement RFC 8449 — traced against OpenSSL 3.6.2, its
+ClientHello and EncryptedExtensions carry no extension 28 in either role — so a
+row asserting a negotiated limit would assert something no peer in this harness
```
Jaruso: This is not true for the harness we already ship. `scripts/interop/install-tls-peer-deps-ci.sh` installs GnuTLS as the #338 independent record-TLS peer, and GnuTLS supports RFC 8449 (`gnutls_record_set_max_recv_size` / `gnutls_record_set_max_size`); both `gnutls-cli` and `gnutls-serv` expose `--recordsize`.

#359's first acceptance criterion explicitly requires negotiated record-size tests against supported peers, so please add dedicated GnuTLS rows instead of documenting the criterion away. A useful pair is native-server ↔ `gnutls-cli --recordsize 1024` and native-client ↔ `gnutls-serv --recordsize 1024`, with >1 KiB application data (or transcript assertions) so the test proves records are actually bounded/fragmented, not just that the handshake completed.

Jaruso (commented):
### Review result: blockers found

I reviewed all 13 changed files and checked the implementation against #359's acceptance criteria. I cannot submit `REQUEST_CHANGES` because this PR is authored by the authenticated GitHub account, so this is a comment review, but I consider the three items below merge-blocking.

The record arithmetic/padding implementation is otherwise in good shape: `TLSInnerPlaintext` accounting includes the content-type byte, protected handshake/application output is split at the peer bound, padding is capped by the effective inner limit, oversize input maps to `record_overflow`, and the boundary/off-by-one coverage is extensive. The docs also clearly distinguish privacy padding from performance coalescing.

#### 1. Server sends an unsolicited `record_size_limit` response

TLS 1.3 extension responses are request/response: an EncryptedExtensions response must not be sent when the client omitted the corresponding ClientHello extension. A conforming client can abort with `unsupported_extension`. Gate extension 28 in both the full-auth and PSK-resumption EE paths on the client actually having offered it.

#### 2. Configured local limit is enforced even when RFC 8449 was not negotiated

`recordSizeLimits()` currently exposes `policy.record_size_limit` as the active inbound bound as soon as the record profile is used. That makes a configured client (for example 512) reject a legal larger protected record from a peer such as OpenSSL that ignores extension 28, even though RFC 8449 explicitly says that when the extension is not negotiated the normal protocol maximum applies.

Track whether the extension became active and expose the configured local bound only then. Keep the directionality: on the server, a ClientHello offer can immediately establish the peer *send* bound for the server's protected handshake flight because the server is going to answer the offer; on the client, its advertised local receive bound cannot become effective merely because it was offered — the server's EE response is what confirms negotiation.

A minimal shape would be along these lines (the exact state representation is up to you):

```zig
record_size_limit_negotiated: bool = false,

pub fn recordSizeLimits(self: *const Tls13Backend) record_size.Limits {
    return .{
        .local = if (self.record_size_limit_negotiated)
            self.policy.record_size_limit
        else
            record_size.default_limit,
        .peer = self.peer_record_size_limit,
    };
}
```

Then set/clear the active state at the role-appropriate CH/EE transitions, including resumption/HRR, and ensure the server only emits EE extension 28 when `peer_record_size_limit != null`.

Required regression coverage should include:

- server receives a ClientHello with no extension 28: both full and PSK EE omit extension 28;
- server with a small configured local limit accepts normal protocol-sized protected records when the client did not offer extension 28;
- client with a small configured local limit talking to a peer that omits extension 28 accepts records above that configured value (up to the TLS protocol max);
- once the server EE confirms extension 28, the configured local receive limit is enforced and an over-limit record produces fatal `record_overflow`;
- existing peer-bound fragmentation still applies to the protected server flight when the client *does* offer a small limit.

#### 3. External negotiated-limit acceptance criterion is not met

#359 explicitly requires negotiated record-size tests against supported peers, and the #338 record-TLS harness already installs/runs GnuTLS in both roles. GnuTLS supports RFC 8449 and `gnutls-cli` / `gnutls-serv` expose `--recordsize`. The new matrix documentation therefore should not state that no peer in the harness can negotiate extension 28.

Add dedicated GnuTLS rows (for example `--recordsize 1024`) in both native roles and make the exchange large enough to prove the negotiated bound actually affects records rather than merely completing a handshake. This will independently exercise the negotiation path and catch the server-response issue above.

#### CI

The current head CI run is red in the appliance ARM HTTP/2 job and the H3 resumption/0-RTT job. Those failures look outside the changed record-TLS surface, so I am not attributing them to this PR here, but they still need to be reconciled/green before merge.

Once the three blockers above are fixed, the rest of the implementation looks close to satisfying #359.

docs/TLS_RECORD_TRANSPORT.md
```diff
@@ -113,3 +113,96 @@ immediate application work; raw socket registration is derived only from
...
+rather than 2^14, and why usable content is always one byte less than the
+advertised limit.
+
+* Record transport only. TLS-over-QUIC has no TLS records, so the QUIC profile
+  never offers the extension and never reads one. A QUIC **server** ignores a
+  client that offers it — without even validating the value — because RFC 8446
+  §4.1.2 has a server ignore an unsupported ClientHello extension, and
+  GnuTLS-based QUIC clients do send it. A **client** still rejects one in
+  EncryptedExtensions with `unsupported_extension`, which is the opposite
+  case and the one RFC 8446 §4.2 requires aborting on: a server must not answer
```
Jaruso, line 138: This new text confirms the earlier interop blocker rather than removing it: GnuTLS sends `record_size_limit`, and the repository's record-TLS conformance harness already installs/runs GnuTLS. `docs/TLS_INTEROP_MATRIX.md` still says no peer in the harness can negotiate extension 28, which is now directly contradicted by this PR. Please add explicit GnuTLS record-mode rows using a non-default `--recordsize` in both roles and assert actual bounded/fragmented traffic, then update the matrix documentation accordingly.

Jaruso (commented):
Re-review of `0852390e`:

The new QUIC fix is correct: a QUIC-profile server should ignore a ClientHello `record_size_limit` it does not support rather than parse/reject it. The H3 resumption/0-RTT interop job is now green, along with the TLS interop/conformance matrix and the appliance jobs that had failed on the previous head.

However, the three blocking findings from the prior review remain unresolved on this head:

1. Both record-mode server EncryptedExtensions builders still emit extension 28 unconditionally. TLS 1.3 forbids an extension response when the client did not send the corresponding request; the server must emit `record_size_limit` only when the ClientHello offered it. The full-handshake and PSK-resumption paths both need the same condition plus regression tests that inspect EE when the CH omits extension 28.

2. `recordSizeLimits().local` still returns the configured advertisement even when the extension was not negotiated. A client configured with a small local limit therefore treats that offer as an active receive bound even when the server omits extension 28, contrary to RFC 8449's rule that, if the extension is not negotiated, ordinary protocol record sizes remain legal. Please separate "configured/offered" from "active negotiated" state and test a small-limit client against an omitted-extension server.

3. #359 explicitly requires negotiated record-size tests against supported peers. The repository's existing record-TLS harness already installs GnuTLS, whose CLI exposes `--recordsize`, and this new commit itself confirms GnuTLS sends extension 28. Dedicated native-server ↔ GnuTLS-client and native-client ↔ GnuTLS-server rows should assert a non-default negotiated limit and force enough application/protected-handshake data to prove the limit is actually enforced/fragmented.

The rest of the #359 acceptance surface still looks solid from the first review: boundary/off-by-one/fragmented/over-limit behavior, padding arithmetic/caps, and the privacy-padding vs coalescing documentation are well covered.

I would not merge yet. GitHub still prevents `REQUEST_CHANGES` because this account owns the PR, so this is submitted as a COMMENT review.

Jaruso (commented):
Re-review at `0852390e`: there are no commits since the previous review, so the previously identified blockers remain unchanged and all four inline threads are still current/non-outdated. CI is now fully green, including H3 resumption/0-RTT interop and the ARM jobs, confirming the QUIC regression fix is good. However, this is still not merge-ready until the protocol/acceptance issues are addressed: (1) only send `record_size_limit` in server EncryptedExtensions when the client offered it, in both full and PSK-resumed handshakes; (2) do not activate the configured local receive bound unless extension 28 was actually negotiated; and (3) satisfy #359's external negotiated-record-size acceptance criterion with GnuTLS record-mode coverage and correct the contradictory interop documentation. No additional defects found beyond the existing review threads.

src/tls/tls13_backend.zig
Jaruso: Fixed in ad138f1e. You are right that RFC 8449 does not override RFC 8446 §4.2 — I had reasoned that the value is useful to the client regardless, which is true but not a licence to send an unsolicited response. Both EE builders now go through `writeRecordSizeLimitResponse`, which returns early unless `peer_record_size_limit != null`.

Tests added in `tls13_backend_tests.zig`:
- `#359 a server omits record_size_limit from EncryptedExtensions when the client did not offer` — inspects the EE message in the full-handshake flight.
- `#359 a server answers in EncryptedExtensions exactly when the client offered`.
- `#359 a PSK-resumed server flight follows the same request/response rule` — loops over offered/not-offered so the resumed builder is proven in both directions rather than only the negative one.

src/tls/tls13_backend.zig
Jaruso: Fixed in ad138f1e. Added `record_size_limit_active`, set at the response rather than the offer: the client sets it when the EE answer arrives, the server when it writes that answer. `recordSizeLimits().local` returns the configured value only when active and `record_size.default_limit` otherwise.

I deliberately did not gate the `peer` half the same way, for the timing reason you raised — the server sets it while parsing the ClientHello, and the stream refreshes limits at the top of `applyDriverOutcome`, before that batch serialises its `handshake_bytes`. So the send bound is in place to fragment the flight that carries the response.

The omitted-extension regression you asked for is `#359 a client that offers but is ignored keeps the protocol-maximum receive bound`: configured 512, EE with ALPN and no extension 28, asserts `local` is the protocol maximum and `inboundInnerMax()` is 16385. Its positive counterpart (`... only once the server answers`) checks the same client flips to 512 once the answer arrives.

docs/TLS_INTEROP_MATRIX.md
Jaruso: You are right, and I got this wrong by checking OpenSSL only — the same GnuTLS behaviour had already broken the H3 interop job earlier in this branch, which should have told me the harness had a capable peer. Rewrote the section in ad138f1e around two real rows rather than documenting the criterion away.

- `record/server/gnutls/record-size-limit` — `gnutls-cli --recordsize 1024`; a 3000-byte response must leave as ≥5 protected records with no inner plaintext above 1025.
- `record/client/gnutls/record-size-limit` — we advertise 512 and GnuTLS must fragment its ~780-byte response to fit while still delivering all 782 bytes.

I split the directions across the rows on purpose: a row that only checked our own sending would pass against a peer that ignored the extension, and one that only checked what arrived would pass without our sizing logic running. Supporting that, the tool gained `--record-size-limit`, `--expect-peer-record-size-limit`, `--expect-max-record-bytes`, `--expect-min-records-sent`, and `--expect-max-received-record-bytes`; both bounds fail rather than pass vacuously when nothing was sealed or nothing arrived.

One detail worth recording: GnuTLS `--recordsize 1024` advertises **1025**, because RFC 8449 counts the content-type byte and GnuTLS adds it. The rows pin the wire value, and the matrix doc calls this out.

Verified locally with GnuTLS 3.8.13: `pass=86 fail=0 skip=0`.

docs/TLS_RECORD_TRANSPORT.md
Jaruso: Corrected in ad138f1e. The paragraph now says a QUIC *server* ignores the extension (RFC 8446 §4.1.2) while a client still rejects an unsolicited EE response, and I added the request/response and offer-vs-negotiated rules alongside it. `TLS_INTEROP_MATRIX.md` no longer claims the harness has no capable peer — it documents the two GnuTLS rows instead, including the `--recordsize N` → `N+1` detail. See the reply on that file for what the rows assert.

src/tls/record_epoch_bridge.zig
```diff
@@ -464,22 +542,54 @@ pub const Bridge = struct {
...
+    /// even hold a tag is left for that path to name.
+    fn checkInboundRecordSize(
+        self: *Bridge,
+        read: *const record_protection.ReadState,
+        record: record_codec.Record,
+    ) Error!void {
+        const tag_len = read.keys.profile.aead.tagLength();
+        if (record.payload.len < tag_len) return;
+        const inner_len = record.payload.len - tag_len;
+        if (inner_len <= self.record_size_limits.inboundInnerMax()) return;
```
Jaruso, line 630: **Blocking:** the new `record_size_limit_active` split fixes the ignored-extension case, but it creates a one-flight enforcement gap on the client. The first protected server record is opened here while the bridge still has `local = default_limit`; only after `driveReceive()` parses EncryptedExtensions does the backend set `record_size_limit_active = true` and `applyDriverOutcome()` refresh the bridge. If a client advertised 512 and the server responds with extension 28 in a 513+ byte `TLSInnerPlaintext`, this check accepts that very record and it is never revalidated.

RFC 8449 §4 explicitly applies the limit to protected handshake records and requires fatal `record_overflow` when a TLS endpoint receives a record larger than its advertised negotiated limit. We cannot safely pre-reject at 512 before knowing whether the server answered the extension, so the first-record exception has to be resolved *after* EE becomes known.

Please retain the inner lengths of protected server-handshake records received while the client's offer is pending; when EE completes:

```zig
if (record_size_limit_just_activated and pending_pre_ee_max_inner_len > negotiated.local) {
    return error.RecordSizeLimitExceeded;
}
```

Then clear the pending measurement if EE omits extension 28. Track the max across all pending records rather than only the record that completes EE, because EncryptedExtensions itself can be fragmented across records. Add a regression with client limit 512 and an oversized protected record carrying/fragmenting the EE response, and assert `record_overflow`.

Jaruso (commented):
The four prior blockers are addressed in ad138f1e and I resolved those threads. The conditional EE response, offer-vs-negotiated state, full/PSK regressions, and two GnuTLS record-size rows all look correct; the external matrix is green (86/86, including both RFC 8449 rows), and the rest of CI is green as well.

I found one remaining RFC 8449 boundary bug in the new activation model, detailed inline: on the client, the protected record that *contains* the server's EncryptedExtensions response is opened while `local` is still the protocol maximum. `record_size_limit_active` flips only while processing that already-opened record, so if that record itself exceeds the smaller value the client advertised, it is accepted and never revalidated. RFC 8449 applies the limit to protected handshake records and requires a TLS endpoint receiving a record larger than its advertised negotiated limit to send fatal `record_overflow`.

This needs a small deferred-enforcement path for the pre-EE protected record(s): retain their inner lengths while the client's offer is pending, and if EE confirms extension 28, validate those records against the now-active local limit before continuing. This needs to handle an EE fragmented across records, so tracking the maximum pending protected-record inner length until EE is complete is safer than checking only the single record that happens to complete parsing.

Once that edge is covered with a regression test, I don't see anything else blocking #359.

src/tls/record_epoch_bridge.zig
Jaruso: Fixed in 537d021e — you are right, and the window is exactly as you describe: the record is opened against `default_limit`, `driveReceive()` then activates, and nothing ever went back to judge the record that carried the activation.

Implemented as you suggested. `Bridge.pending_handshake_inner_max` retains a high-water mark of handshake-epoch inner lengths (a mark across the flight, not just the completing record — EE can fragment), and `setRecordSizeLimits` settles it when the bound becomes known:

```zig
if (self.pending_handshake_inner_max > limits.local) {
    self.record_size_counters.oversize_records_rejected +|= 1;
    return error.RecordSizeLimitExceeded;
}
```

Two deliberate choices worth your eye:

**No explicit clear is needed when EE omits extension 28.** `local` then stays at the protocol maximum, which no record can exceed, so the comparison is inert. I preferred that to an explicit reset because it cannot get out of step with the activation flag. There is a regression for it (`an unanswered extension leaves pre-activation handshake records alone`, using a full 16384-byte flight).

**The mark is handshake-epoch only, which I think is load-bearing.** A server activates when it *writes* its EE answer, and 0-RTT records can already be in flight by then. A client cannot honor a bound it has not received, so retaining 0-RTT records and judging them at activation would reject legitimate early data — a false positive I would rather not introduce. On a server no handshake-epoch record arrives before it writes that answer (the client's Finished follows it), so the check is naturally client-only. Covered by `0-RTT records are never settled retroactively against the server's own bound`.

Related: the stream now `deferHandshakeFailure`s this rather than `fail`s it, so the fatal alert is actually queued — `fail()` tears down without emitting one, which would have left the `record_overflow` requirement unobservable.

Regressions added, including the one you asked for: client limit 512, server answers extension 28 and sends its Finished in a single 2002-byte protected record, asserting `error.RecordSizeLimitExceeded` on the client **and** `record_overflow` observed on the peer. Its counterpart drives the identical oversized flight against a server that never answers and asserts the handshake still completes.

`zig build test` green; interop matrix still `pass=86 fail=0 skip=0`.

One thing I did not change and want to flag rather than decide unilaterally: live enforcement still applies to `.zero_rtt` records once the bound is active, so an operator who lowers `record_size_limit` could reject early data from a client that had no way to know it. Happy to exempt 0-RTT from live enforcement too, or to remember the limit across resumption, if you would prefer either — it seemed like a separate call from this fix.

Jaruso (commented):
Follow-up review at `537d021e`.

The pre-activation handshake-record gap from my previous review is fixed correctly and I resolved that thread. The high-water mark covers fragmented EncryptedExtensions, activation settles the pre-EE records, the ignored-extension path remains at the protocol maximum, and the stream now queues the required `record_overflow` alert before latching the failure.

I found one remaining blocking resumption issue in the 0-RTT path (inline). The current handshake's newly negotiated local limit is installed when the server writes EncryptedExtensions, but buffered/in-flight 0-RTT records were sent before the client could possibly know that value. Applying that new bound to `.zero_rtt` can therefore reject conforming early data when an operator lowers `record_size_limit`. The resumable session model does not currently retain a prior record-size limit, so there is no PSK-associated bound available to apply instead.

CI on this head is green, including unit/appliance tests, the TLS interop/conformance matrix, and the resumption/0-RTT jobs. Apart from the 0-RTT epoch-boundary issue, the #359 acceptance criteria now look covered.

Jaruso (commented):
Re-reviewed at `133dbef6`. The remaining 0-RTT blocker is addressed correctly: `.zero_rtt` no longer applies the current handshake's negotiated `record_size_limit`, while handshake/application epochs still enforce it. The new bridge-level regression pins the epoch boundary directly, and the resumption regression proves the server actually negotiates 512, uses a real resumed 0-RTT secret, accepts a >512 protocol-legal early record in the buffered/pre-EE case, and then enforces the exact 512-inner boundary at 1-RTT.

I also re-checked the full #359 acceptance criteria against the implementation already reviewed: external negotiated GnuTLS rows exist in both roles; boundary/off-by-one/fragmentation/over-limit coverage is present; padding remains bounded by negotiated/protocol limits; defaults remain interoperable; and the docs distinguish padding from coalescing. RFC 8449 §4's resumption rule is now handled without incorrectly applying the newly negotiated receive bound to early records.

CI run #1348 is fully green, including unit tests, TLS interop/conformance, H3 resumption/0-RTT interop, appliance integration, formatting, security, and all builds. I found no additional blocking defects in the new delta. From the code-review/acceptance-criteria standpoint, this is merge-ready. (Using COMMENT because this is the PR author's account and GitHub does not permit approving one's own PR.)

Issue #255, opened 2026-07-02
quic: add qlog, keylog, and transport observability for pure Zig QUIC/H3


Jaruso:
## Summary

Add the observability required to debug pure Zig QUIC and HTTP/3: qlog, keylog support, protocol metrics, and structured logs.

## Motivation

QUIC and HTTP/3 are difficult to debug from normal application logs. Mature implementations rely on qlog/qvis, key logs, pcaps, and detailed transport metrics to understand handshake failures, loss recovery, PTO, migration, stream resets, and QPACK blocking.

## Scope

- qlog event emission for core QUIC transport events:
  - connection started/closed
  - packet sent/received/lost
  - ACK processed
  - PTO/loss detection
  - congestion state changes
  - stream opened/closed/reset
  - flow-control blocked/unblocked
  - path validation/migration
- HTTP/3/qlog events where practical:
  - control stream
  - SETTINGS
  - HEADERS/DATA
  - GOAWAY
  - QPACK blocked/unblocked
- Optional TLS key log support for packet decryption during local debugging.
- Prometheus metrics for:
  - active QUIC connections
  - handshake failures by stage
  - Retry count
  - anti-amplification blocked sends
  - PTO count
  - packets lost
  - bytes sent/received
  - stream resets
  - flow-control blocked events
  - QPACK blocked streams
  - h3 requests and latency by protocol
- Operator docs for enabling/disabling qlog/keylog safely.

## Non-goals

- Full qvis UI integration.
- High-cardinality per-client metrics.
- Making qlog always-on in production.

## Acceptance criteria

- qlog output can be enabled for local/debug runs.
- Key log output can be enabled for local/debug runs if supported by the TLS adapter.
- Transport metrics make handshake/loss/flow-control failures distinguishable.
- qlog/keylog paths are disabled by default and documented as sensitive/debug-only.
- Interop/failure harnesses can save qlog/keylog artifacts.

## Testing strategy

- Unit tests for qlog event serialization where applicable.
- Integration test verifies qlog file is produced and contains expected event classes.
- Metrics snapshot tests for common error paths.
- Manual validation with qvis once enough transport exists.

## References

- qlog drafts
- qvis tooling
- QUIC Interop Runner artifact patterns
- #240 Pure Zig QUIC and HTTP/3 Foundation
- #247 interop/fuzz/benchmark harness

PR #605, opened 2026-08-11
Wire H3 qlog events and QPACK metrics

Jaruso:
## Summary
- add a typed HTTP/3 EventSink and emit SETTINGS, stream classification, frame create/parse, GOAWAY, and priority update events at H3 state-machine boundaries
- bridge accepted runtime H3 events into the injected http3 qlog sink, keeping the default no-op behavior
- expose aggregate QPACK blocked/unblocked/cancelled/table/decode-failure Prometheus metrics and document the static-decoder caveat

Refs #255 (slice 3)

## Tests
- zig build test-quic
- zig build test

### DIFF
--- a/docs/QUIC_QLOG.md
+++ b/docs/QUIC_QLOG.md
@@ -112,19 +112,19 @@ Names below are `namespace:event`. Transport events (`src/quic/qlog.zig`):
 
 | Requirement (#255)        | qlog event                          | Key data |
 |---------------------------|-------------------------------------|----------|
-| handshake                 | `quic:connection_started`           | odcid/scid/dcid lengths |
+| handshake                 | `quic:connection_started`           | required `local` / `remote` endpoint objects, plus Tardigrade CID-length diagnostics |
 |                           | `tardigrade:quic_handshake_progressed` | `stage` (started -> confirmed / failed) |
 |                           | `quic:connection_closed`            | `reason`, optional `error_code` |
-|                           | `quic:key_updated`                  | `key_phase` |
-| packet sent               | `quic:packet_sent`                  | type, number, length, ack-eliciting |
-| packet received           | `quic:packet_received`              | type, number, length |
-| packet lost               | `quic:packet_lost`                  | type, number |
+|                           | `quic:key_updated`                  | required `key_type`, optional full `key_phase` and `trigger` |
+| packet sent               | `quic:packet_sent`                  | `header`, `raw`, Tardigrade ack-eliciting diagnostic |
+| packet received           | `quic:packet_received`              | `header`, `raw` |
+| packet lost               | `quic:packet_lost`                  | `header` |
 | recovery metrics          | `quic:recovery_metrics_updated`     | RTT/PTO/cwnd/bytes-in-flight updates |
 | **deprotection failure**  | `quic:packet_dropped`               | `trigger:"decryption_failure"` |
 | PATH_CHALLENGE/RESPONSE   | `tardigrade:quic_path_validation`   | `phase` (challenge/response sent/received, validated, failed) |
-| migration                 | `quic:migration_state_updated`      | `trigger` (nat_rebinding/active), `state` (accepted/blocked) |
+| migration                 | `quic:migration_state_updated`      | required `new` migration state, optional `old` |
 | stream reset              | `tardigrade:quic_stream_reset`      | `direction` (reset/stop-sending, sent/received), stream id, error code |
-| flow-control blocked      | `quic:connection_data_blocked_updated` / `quic:stream_data_blocked_updated` | optional stream id, limit |
+| flow-control blocked      | `quic:connection_data_blocked_updated` / `quic:stream_data_blocked_updated` | required `new` blocked state, optional `old` / `reason`; stream variant requires `stream_id` |
 
 Application events (`src/http3/qlog.zig`):
 
@@ -132,13 +132,13 @@ Application events (`src/http3/qlog.zig`):
 |---------------------------|-------------------------------------|----------|
 | SETTINGS                  | `http3:parameters_set`              | max field section size, QPACK table cap, blocked streams |
 | control stream            | `http3:stream_type_set`             | stream id, stream type |
-| HEADERS / DATA / GOAWAY   | `http3:frame_created` / `http3:frame_parsed` | frame type, stream id, raw length |
+| HEADERS / DATA / GOAWAY   | `http3:frame_created` / `http3:frame_parsed` | variant-specific frame payloads; HEADERS carries `headers`, SETTINGS carries `settings`, GOAWAY carries `id` |
 | **QPACK blocked**         | `tardigrade:qpack_stream_state_updated` | `state` (blocked/unblocked), stream id |
 
 `quic:packet_dropped` with `trigger:"decryption_failure"` is the
 canonical qlog encoding of an AEAD deprotection failure, satisfying the #255
 requirement that deprotection failures are reported deterministically. It is
-distinct from a normal drop (`unknown_connection_id`, `key_unavailable`, …).
+distinct from a normal drop (`connection_unknown`, `key_unavailable`, ...).
 
 ### Serialization format
 
@@ -158,11 +158,13 @@ Each record is one JSON-SEQ line:
   `file_schema: urn:ietf:params:qlog:file:sequential`,
   `serialization_format: application/qlog+json-seq`, `trace`,
   `event_schemas`, vantage point, and ODCID `group_id` as the first JSON-SEQ
-  record. The header spans both packages — the `group_id` ties transport and
+  record. The trace declares `clock_type: monotonic` and `epoch: unknown`
+  because event timestamps are monotonic process-relative values, not Unix
+  timestamps. The header spans both packages — the `group_id` ties transport and
   H3 events to one connection — so the **composition root** fills it in and
   writes it once, then appends event records from both `quic` and `http3`. The
   `tests/quic_h3_smoke.zig` harness exercises exactly this shape (header +
-  transport record + H3 record) so the merged-file contract is locked.
+  representative QUIC/H3 records) so the merged-file contract is locked.
 - JSON-SEQ qlog artifacts use the `.sqlog` suffix. Reserve `.qlog` for the
   normal contained JSON qlog form.
 
--- a/docs/QUIC_QLOG.md
+++ b/docs/QUIC_QLOG.md
@@ -22,7 +22,7 @@ layering, and the safety rules up front so those call-sites are mechanical.
   from a captured trace, not just from ad-hoc `std.log` lines.
 - Provide the primitives (trace-header + event-line + keylog-line writers) so
   that, once the composition root assembles them, the interop/failure harnesses
-  (#247) save `*.qlog` / `*.keys` artifacts that qvis / Wireshark consume
+  (#247) save `*.sqlog` / `*.keys` artifacts that qvis / Wireshark consume
   directly. This PR ships those writers and the merged-shape test; the root
   wiring that emits real flows is follow-up.
 - Keep everything **off by default** and cheap when off.
@@ -43,7 +43,7 @@ respects that boundary rather than punching through it.
 ```
                  composition root (gateway h3 listener / smoke harness)
                  ┌───────────────────────────────────────────────────┐
-                 │  owns the *.qlog file + *.keys file writers        │
+                 │  owns the *.sqlog file + *.keys file writers       │
                  │  installs one quic.qlog.Sink, one http3.qlog.Sink, │
                  │  one quic.keylog.Sink; interleaves all three       │
                  └───────────────▲───────────────▲───────────────▲────┘
@@ -67,7 +67,7 @@ Rules:
    `Sink{}` is a no-op, so the seam costs nothing until a root wires it.
 4. The **concrete file writers** live at the composition root, which already
    owns both packages. It timestamps and interleaves the streams into one
-   JSON-SEQ `.qlog` file, so a single trace still shows transport and H3 events
+   JSON-SEQ `.sqlog` file, so a single trace still shows transport and H3 events
    side by side without either package depending on the other.
 
 This is why there is no single shared "qlog writer" module: sharing one would
@@ -89,32 +89,53 @@ into one valid file at the root instead.
 
 ## qlog event catalogue
 
-Names below are `category:event`. Transport events (`src/quic/qlog.zig`):
+The qlog scaffold tracks the July 2026 submitted drafts:
+
+- `draft-ietf-quic-qlog-main-schema-14`
+- `draft-ietf-quic-qlog-quic-events-13`
+- `draft-ietf-quic-qlog-h3-events-13`
+
+Because these are still drafts, the sequential trace header declares
+draft-qualified event schema URIs:
+
+- `urn:ietf:params:qlog:events:quic-13`
+- `urn:ietf:params:qlog:events:http3-13`
+- `https://bare.systems/tardigrade/qlog/events/debug-1`
+
+The Tardigrade URI covers debug events that are useful for #255 but are not
+standardized qlog events. Most importantly, current HTTP/3 qlog no longer
+standardizes QPACK events, so QPACK blocked-stream visibility is either a
+Prometheus/structured diagnostic signal or the explicit
+`tardigrade:qpack_stream_state_updated` extension below.
+
+Names below are `namespace:event`. Transport events (`src/quic/qlog.zig`):
 
 | Requirement (#255)        | qlog event                          | Key data |
 |---------------------------|-------------------------------------|----------|
-| handshake                 | `connectivity:connection_started`   | odcid/scid/dcid lengths |
-|                           | `connectivity:handshake`            | `stage` (started → confirmed / failed) |
-|                           | `connectivity:connection_closed`    | `reason`, optional `error_code` |
-|                           | `security:key_updated`              | `key_phase` |
-| packet sent               | `transport:packet_sent`             | type, number, length, ack-eliciting |
-| packet received           | `transport:packet_received`         | type, number, length |
-| packet lost               | `recovery:packet_lost`              | type, number, bytes-in-flight, cwnd |
-| **deprotection failure**  | `transport:packet_dropped`          | `trigger:"payload_decrypt_error"` |
-| PATH_CHALLENGE/RESPONSE   | `transport:path_validation`         | `phase` (challenge/response sent/received, validated, failed) |
-| migration                 | `connectivity:connection_migrated`  | `kind` (nat_rebinding/active), `outcome` (accepted/blocked) |
-| stream reset              | `transport:stream_reset`            | `direction` (reset/stop-sending, sent/received), stream id, error code |
-| flow-control blocked      | `transport:data_blocked`            | `scope` (connection/stream), optional stream id, limit |
+| handshake                 | `quic:connection_started`           | odcid/scid/dcid lengths |
+|                           | `tardigrade:quic_handshake_progressed` | `stage` (started -> confirmed / failed) |
+|                           | `quic:connection_closed`            | `reason`, optional `error_code` |
+|                           | `quic:key_updated`                  | `key_phase` |
+| packet sent               | `quic:packet_sent`                  | type, number, length, ack-eliciting |
+| packet received           | `quic:packet_received`              | type, number, length |
+| packet lost               | `quic:packet_lost`                  | type, number |
+| recovery metrics          | `quic:recovery_metrics_updated`     | RTT/PTO/cwnd/bytes-in-flight updates |
+| **deprotection failure**  | `quic:packet_dropped`               | `trigger:"decryption_failure"` |
+| PATH_CHALLENGE/RESPONSE   | `tardigrade:quic_path_validation`   | `phase` (challenge/response sent/received, validated, failed) |
+| migration                 | `quic:migration_state_updated`      | `trigger` (nat_rebinding/active), `state` (accepted/blocked) |
+| stream reset              | `tardigrade:quic_stream_reset`      | `direction` (reset/stop-sending, sent/received), stream id, error code |
+| flow-control blocked      | `quic:connection_data_blocked_updated` / `quic:stream_data_blocked_updated` | optional stream id, limit |
 
 Application events (`src/http3/qlog.zig`):
 
 | Requirement (#255)        | qlog event                          | Key data |
 |---------------------------|-------------------------------------|----------|
-| SETTINGS                  | `http:parameters_set`               | max field section size, QPACK table cap, blocked streams |
-| control stream / HEADERS / DATA / GOAWAY | `http:frame_created` / `http:frame_parsed` | frame type, stream id, length |
-| **QPACK blocked**         | `qpack:stream_state_updated`        | `state` (blocked/unblocked), stream id |
+| SETTINGS                  | `http3:parameters_set`              | max field section size, QPACK table cap, blocked streams |
+| control stream            | `http3:stream_type_set`             | stream id, stream type |
+| HEADERS / DATA / GOAWAY   | `http3:frame_created` / `http3:frame_parsed` | frame type, stream id, raw length |
+| **QPACK blocked**         | `tardigrade:qpack_stream_state_updated` | `state` (blocked/unblocked), stream id |
 
-`transport:packet_dropped` with `trigger:"payload_decrypt_error"` is the
+`quic:packet_dropped` with `trigger:"decryption_failure"` is the
 canonical qlog encoding of an AEAD deprotection failure, satisfying the #255
 requirement that deprotection failures are reported deterministically. It is
 distinct from a normal drop (`unknown_connection_id`, `key_unavailable`, …).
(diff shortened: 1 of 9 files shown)


Jaruso (commented):
**Blocking review — changes needed before merge.**

I reviewed this against #255's 2026-08-10 reconciliation, especially Slice 3 and its prerequisite. The typed H3 seam is a good direction, and the new unit coverage exercises the happy path, but there are several blocking correctness/acceptance gaps:

1. H3 qlog routing loses connection identity: every accepted H3 connection points at one runtime-wide sink, so concurrent connections cannot be routed to the same per-connection trace as their QUIC events.
2. Request streams are not represented by the stream-classification event model at all.
3. Request frame parsing is observed through `saw_headers`/`saw_data` state transitions instead of the actual frame decode boundary, so frame events are incomplete and can have incorrect lengths.
4. Client request creation similarly emits one HEADERS event for a combined HEADERS+DATA buffer and never emits the DATA frame.
5. Local SETTINGS do not emit `parameters_set`, despite Slice 3 requiring local/peer SETTINGS application visibility.
6. The Prometheus QPACK current gauges are not based on current per-connection state; the blocked gauge can leak after teardown, and table bytes are overwritten rather than aggregated.
7. This also wires new live events into the legacy `http:*`/`qpack:*` qlog scaffold even though #255 explicitly says the `http3:*` schema refresh/custom-QPACK namespace is a prerequisite before adding live emission call sites.

CI note: unit tests/builds and the H3 resumption/0-RTT interop job are green. The overall CI failure is in the unrelated TLS peer dependency install (`apt` gets 403s from packages.microsoft.com), so I am not treating that external runner failure as a PR defect.

Please address the inline comments and add tests that cover two concurrent H3 connections, request-stream classification, multiple/coalesced/fragmented DATA frames, client request bodies, local+peer SETTINGS, and QPACK gauge teardown/multi-connection aggregation.

Jaruso (commented):
Follow-up review on `7b80cb2`.

The previous blockers around per-connection sink routing, exact request-frame parse observation, body DATA creation, local SETTINGS emission, and QPACK gauge aggregation have been materially addressed. I re-read the new head against #255 Slice 3 and the current July 2026 HTTP/3 qlog draft.

There are still **3 blockers before merge**:

1. local client-created request streams still never emit `stream_type_set(request)`;
2. `frame_created` is emitted after transport delivery rather than at the actual H3 framing boundary, so write/backpressure failures lose creation events;
3. the branch still wires live events into a serializer that only renamed the namespace but does not implement the current draft-13 event data shapes. PR #606 is now the dedicated Slice 1 refresh and should land first; rebase this PR onto it and adapt the H3 event mapper to those types rather than duplicating a partial schema refresh here.

CI on the new head is currently still running, but all completed unit-test/build/H3 interop/TLS interop jobs I checked are green so far.

Jaruso (commented):
Blocking follow-up review — changes are still needed before merge.

The three blockers from my previous pass are fixed in this head: local request streams are now classified, `frame_created` is emitted at the H3 framing boundary before transport writes, and the current draft-shaped qlog model has been integrated. I also re-read all nine changed files and the Slice 3 requirements from #255.

I found six remaining correctness issues:

1. The runtime mapper is still lossy: `Conn.Event` carries only frame type/stream/length, so `h3FrameToQlog()` fabricates required frame data (`id = 0`, `push_id = 0`, `element_id = 0`) and emits empty HEADERS/SETTINGS payloads. The typed H3 event must carry the actual frame-specific data at the state-machine boundary rather than reconstructing placeholders later.
2. `parameters_set` drops whether the settings were local or remote by hard-coding `initiator = null`. Carry initiator in the typed event and emit `.local` / `.remote` at the two transition points.
3. Request/control `frame_parsed` events are emitted only after semantic processing succeeds. A syntactically decoded frame that causes the protocol error is exactly the frame qlog needs to show, but it currently disappears from the trace.
4. The HTTP/3 serializer writes header names/values (and priority text) directly into JSON without escaping. Peer-controlled `"` or `\\` bytes can therefore corrupt the JSON-SEQ record.
5. PRIORITY_UPDATE frame serialization still uses an `element_id` field; the current H3 qlog shape distinguishes request `stream_id` from push `push_id`.
6. SETTINGS frame tests/serialization use free-form uppercase names such as `SETTINGS_QPACK_BLOCKED_STREAMS`; the draft vocabulary is the lowercase qlog setting-name enum. Please make this typed rather than allowing arbitrary strings.

The QPACK process-gauge delta/teardown fix now looks sound, the per-connection sink routing is correct, and the exact frame-length observer work is good. Once real HEADERS payloads are carried through, also update the docs/comments claiming a 512-byte output buffer covers every event — decoded header lists make record size data-dependent.

CI had no failures among the completed jobs I checked on this head; several jobs were still finishing at review time.

Jaruso (commented):
Follow-up review on `e47c559`.

The six blockers from my previous pass are materially addressed in this head: frame payloads are now typed instead of fabricated, SETTINGS ownership is carried through, request/control rejected-frame observation was moved before semantic application, JSON text is escaped with a byte fallback, PRIORITY_UPDATE preserves request-vs-push identity, and SETTINGS serialization uses the current lower-case qlog names.

I re-checked the full 9-file PR against #255 Slice 3 and the refreshed H3 qlog model. There are still **3 correctness blockers before merge**:

1. SETTINGS `frame_created`/`frame_parsed` still use semantic `frame.Settings`, which cannot faithfully represent the wire SETTINGS list (explicit zero values, unknown settings, or rejected reserved/duplicate entries).
2. The client response path still returns on QPACK decode failure before emitting the already-parsed HEADERS frame.
3. The new frame payloads contain borrowed slices, but neither the H3 event sink nor qlog sink defines callback-only lifetime/ownership. Several producers pass stack-backed arrays/slices, so a sink that retains/enqueues the event can dereference dangling memory.

Details and expected fixes are inline. CI on this head is still running; all completed unit/build/H3/TLS/security jobs I checked are green so far.

Jaruso (commented):
## Final exhaustive review — `e47c559`

This review **supersedes my earlier incremental reviews**. I re-reviewed all 9 changed files from scratch against the 2026-08-10 reconciliation on #255, the Slice 1 prerequisite, the complete Slice 3 acceptance surface, the current July 2026 qlog main/QUIC/H3 schemas, H3 success and failure paths, payload ownership/lifetimes, and the completed CI run.

The earlier findings around per-connection routing, request classification, exact request-frame observation, DATA creation, framing-boundary creation events, local/peer SETTINGS ownership, QPACK gauge aggregation, payload escaping, and typed priority targets are materially addressed. The branch is **still not mergeable**, however. The list below is the complete remaining change set I found on this head.

### 1. Make Slice 1 a single source of truth before merging Slice 3

PR #606 is still the dedicated Slice 1 qlog serializer refresh, while #605 currently embeds its own copy of the same serializer work. Do not maintain/fix two divergent copies. Fix the serializer/schema items below in **#606**, complete its required qvis/manual validation, merge #606, then rebase #605 onto it and drop the duplicated Slice 1 diff. GitHub currently reports #605 as non-mergeable, so this also gives the cleanest branch-resolution order.

### 2. Finish the current-draft QUIC qlog model

The refresh still has three schema-level gaps:

- `connection_closed` models internal close categories as `reason` and permits a bare `error_code`; the current schema separates the standardized close `trigger`, transport/application error identity, conditional raw error code, and free-form reason. Map the internal close cause to those standard fields rather than repurposing `reason`.
- `BlockedReason` serializes Tardigrade/internal names such as `data_blocked_frame` and `stream_data_blocked_frame`; these are not legal qlog blocked-reason values. Use the standard vocabulary and map internal causes into it (`connection_flow_control`, `stream_flow_control`, etc.).
- `packet_sent`/`packet_received` require a packet number even for Retry and Version Negotiation. Make packet number optional/variant-constrained and only serialize it for packet types that carry one. While finishing this new current-draft type surface, include the current 0-RTT key types as well.

### 3. Preserve SETTINGS as wire data, not only semantic state

`frame.Settings` is the wrong diagnostic representation for a SETTINGS frame. It loses unknown setting IDs, explicit zero values, explicit false values, wire ordering, and presence. `eventFrameFromRaw()` also gives the diagnostic decoder only 8 entries while the semantic control parser accepts 32.

Carry the actual decoded setting entries (`id/id_value/value`) in the frame event. Map unknown IDs to `name:"unknown"` plus `name_bytes`, retain `reserved`/`unknown` in the qlog setting model, and preserve zero/false/order/presence. Keep `parameters_set` as a separate presence-aware semantic event and include the current `extended_connect` / `h3_datagram` fields.

Tests required: explicit QPACK value `0`, explicit boolean `0`, an unknown setting ID, and a valid SETTINGS list with more than 8 entries must survive into the qlog event faithfully.

### 4. Never fabricate or misclassify a parsed H3 frame

`eventFrameFromRaw()` still manufactures valid-looking values when required payload decoding fails (`GOAWAY id=0`, push IDs `0`, empty SETTINGS), and the request observer rewrites any unsupported/unknown frame into DATA through its `orelse .data` fallback.

Add explicit `unknown`/`reserved` diagnostic frame variants carrying `frame_type_bytes`, or deliberately omit unsupported standard details. Never substitute zero and never change the frame type. If Push Promise remains in the standard qlog frame model, populate its required decoded header field section; otherwise do not claim a complete standard Push Promise payload yet.

Tests required: an unknown numeric frame type is logged as unknown, never DATA; malformed ID-bearing frames are never reported with zero; supported non-zero IDs round-trip unchanged.

### 5. Put the client response observer at the actual parse boundary

The server/request path now observes `RawFrame` before semantic validation, but `pollResponse()` still emits HEADERS only after duplicate-HEADERS checks and successful QPACK decode. A syntactically parsed HEADERS frame therefore disappears when it is a second HEADERS frame or when its QPACK block is malformed.

Use the same invariant in both directions: as soon as `decodeFrameWithLimit()` succeeds, record the H3 frame boundary; then apply QPACK/message semantics. If decoded HTTP fields are unavailable, represent that honestly instead of silently omitting the frame or reporting an empty successful field section.

Tests required: malformed-QPACK HEADERS, duplicate/second HEADERS, and DATA-before-response-HEADERS must still leave the appropriate parsed-frame event before the protocol error is returned.

### 6. Define and enforce the borrowed-payload lifetime contract

`EventFrame.headers.fields` and PRIORITY_UPDATE `field_value` can point into stack/scratch/raw-frame memory, while `conn.EventSink` and `http3.qlog.Sink` currently expose value types that look retainable. A non-blocking sink that queues the Event/Record object can dereference dangling memory after the callback returns.

Either own/deep-copy payloads, or make the synchronous borrowing contract explicit on **both** sink APIs: borrowed slices are valid only for the duration of the callback; consumers must serialize/deep-copy before returning and must never retain or queue the Event/Record itself. Test recorders that retain payload-bearing events should deep-copy them.

A safe eventual writer shape is conceptually:

```zig
fn emit(ctx: ?*anyopaque, record: http3.qlog.Record) void {
    var buf: [max_qlog_record_len]u8 = undefined;
    const line = http3.qlog.writeJson(record, &buf) catch {
        state.dropped += 1;
        return;
    };
    state.enqueueCopy(line); // owns bytes; never retains `record`
}
```

### 7. Make all qlog text serialization safe for peer-controlled/malformed bytes

Two gaps remain after the JSON escaping fix:

- HTTP field text-vs-bytes selection uses only UTF-8 validity. The H3 qlog schema requires byte-oriented fields when characters are outside the HTTP-safe ranges; valid UTF-8 alone is not sufficient.
- PRIORITY_UPDATE bypasses the text/bytes helper and feeds the raw `field_value` directly to the JSON-string writer. `priority.decodePayload()` returns arbitrary bytes, so malformed peer input can still create invalid JSON before semantic priority parsing rejects it.

Use HTTP field validity to select `name/value` versus `name_bytes/value_bytes`. For malformed PRIORITY_UPDATE bytes, do not place arbitrary bytes into the standard text field; keep them only in a documented raw/custom diagnostic path or omit the typed value while preserving the parsed-frame boundary. Add complete-record `std.json` parse tests for invalid UTF-8, control bytes, and HTTP-invalid-but-UTF-8 field data.

Also escape/validate every public text value in `writeTraceHeader()` (`title`, `description`, `group_id`) instead of interpolating caller strings verbatim.

### 8. Establish a real bounded-record policy

The docs and QUIC serializer comments still say 512 bytes covers every event, but H3 HEADERS now serialize decoded field arrays and the H3 tests already had to move to 2048-byte buffers. A legitimate field section can exceed either bound, especially after JSON escaping or byte expansion.

Define a supported `max_serialized_record_len` from the H3 bounds, an intentional/traced truncation policy, or a growing/streaming composition-root serializer. Ordinary valid HEADERS must not become silently dropped records merely because a sink followed the documented 512-byte guidance. Restore/add a large-header test that locks the chosen contract and update every stale 512-byte statement.

### 9. State the QPACK production caveat explicitly

The process-level QPACK delta/gauge aggregation now looks correct, including actual current blocked count, multi-connection aggregation, and teardown subtraction. But #255 explicitly notes that production `conn.zig` still uses the static decoder and does not compose `DynamicDecoder.decodeOrBlock()` today. Therefore the dynamic blocked/table/decode series and the custom blocked/unblocked event remain zero/unemitted in production.

The docs must state that clearly: a zero currently means **“dynamic decoder not composed”**, not “runtime QPACK blocking was observed and absent.” Do not claim the dynamic-QPACK acceptance item from standalone `qpack.zig` unit metrics.

### Required validation after these fixes

Keep/add the negative and fidelity tests named above, then run `zig fmt --check`, `zig build test-quic`, `zig build test`, and the qlog composition smoke tests. CI for `e47c559` is currently green, so there is no unrelated CI failure to chase. For Slice 1, complete #255's manual compatibility requirement by loading one generated `.sqlog` in qvis; #606 currently documents that this has not been performed.

**After #606 is corrected/validated/merged and #605 is rebased with the H3 fixes above, this is the complete remaining change set I found on this head. The next review should be verification of this checklist and regressions introduced by those fixes, not another open-ended discovery pass.**

Jaruso (commented):
## Verification review of the exhaustive checklist — `93c2271`

This is a verification pass against my consolidated review #4908428840, not a new exploratory review. I rechecked every item in that checklist against the rebased branch and current #255 Slice 3 acceptance surface.

A substantial amount is now correct:

- #606 is merged and #605 is rebased onto it, so Slice 1 has a single source of truth.
- Standard blocked-reason vocabulary and 0-RTT key types are fixed.
- SETTINGS frame events now preserve wire entries, including explicit zero values, unknown/reserved IDs, order, and >8 entries.
- Local/remote SETTINGS ownership, request/control/QPACK stream classification, exact frame lengths, framing-boundary `frame_created`, and rejected-frame observation remain intact.
- The client response path now observes malformed-QPACK HEADERS, duplicate HEADERS, and DATA-before-HEADERS before returning the protocol error.
- Both H3 sink APIs now explicitly define synchronous borrowed-slice lifetimes, and retaining test recorders deep-copy payloads.
- HTTP field text-vs-bytes selection and trace-header JSON escaping are fixed.
- The variable-size record-buffer policy/large-record test and the production static-QPACK caveat are now documented.
- CI run 31513371949 is green.

There are still **blocking items before this is merge-ready**:

### 1. Two previously-listed Slice 1 schema items are still present in the merged #606 base

These were already called out in #4908428840 and remain unchanged after the rebase:

**`quic:connection_closed` still uses the internal close category as the standard `reason` field.** In quic-events-13, `reason` is free-form text; the standardized categorical field is `trigger` (`idle_timeout`, `application`, `error`, `version_mismatch`, `stateless_reset`, `aborted`, `unspecified`). Keep the new `CloseError` union for tying raw codes to `connection_error:"unknown"` / `application_error:"unknown"`, but add a standard trigger mapping instead of serializing `CloseReason` into `reason`:

```zig
pub const CloseTrigger = enum {
    idle_timeout,
    application,
    error,
    version_mismatch,
    stateless_reset,
    aborted,
    unspecified,
};

connection_closed: struct {
    trigger: CloseTrigger,
    close_error: CloseError = .none,
    reason: ?[]const u8 = null, // optional free-form diagnostic text
},
```

A bounded mapping from the existing internal causes is straightforward: `idle_timeout -> idle_timeout`, `application_close -> application`, `transport_error/handshake_failure -> error`, `stateless_reset -> stateless_reset`.

**`packet_sent` / `packet_received` still require `packet_number: u64` for every packet type.** The current PacketHeader schema only permits a packet number for Initial, Handshake, 0-RTT, and 1-RTT; Retry and Version Negotiation do not carry one. Make it optional/variant-constrained and omit the JSON field for non-numbered packet types. Add full-record tests for Retry and Version Negotiation proving no fabricated packet number appears.

### 2. `http3:parameters_set` has the wrong wire/schema types for two standardized fields

`extended_connect` and `h3_datagram` are `uint16` in h3-events-13, not JSON booleans. The inline comment has the concrete change. Keep absence as `null` if desired, but when emitted the standard field must be numeric (`1`, or `0` if presence is tracked), not `true`/`false`.

### 3. SETTINGS still violates the `reserved` / `name_bytes` invariant

The draft permits `name_bytes` only when `name == "unknown"`. A reserved setting is represented as `name:"reserved"` without `name_bytes`. The current serializer emits both. If retaining the exact reserved numeric ID is useful, put it in a clearly documented Tardigrade extension field rather than the standardized `name_bytes` field.

### 4. The default-disabled H3 qlog path is no longer cheap

Every server request installs `RequestFrameObserver`, and every HEADERS frame is QPACK-decoded once in that observer and then decoded again by normal `RequestStream` processing. This occurs even when `EventSink.emitFn == null`, so the default no-op qlog configuration pays a second QPACK decode on every request HEADERS block.

At minimum, pass a `null` observer when no sink is installed. Better, when diagnostics are enabled, refactor the request decode path so QPACK is decoded once and decoded fields are exposed to the observer before semantic header validation.

### 5. Known/malformed frames still get mislabeled as `unknown`, and failed HEADERS are represented as successfully-decoded empty HEADERS

The zero-ID fabrication is fixed, but `eventFrameFromRaw()` now uses the qlog `unknown` frame for malformed *known* GOAWAY/PRIORITY_UPDATE/CANCEL_PUSH/MAX_PUSH_ID/SETTINGS payloads and for every PUSH_PROMISE. In h3-events-13, `HTTP3UnknownFrame` is specifically for an unknown numeric frame type; PUSH_PROMISE and GOAWAY are known standard frame types.

Only `raw.typ == .unknown` should become the standard qlog `unknown` frame. For a known frame whose required payload cannot be decoded, preserve the frame boundary via a documented Tardigrade malformed/unsupported-frame diagnostic rather than changing its identity. For PUSH_PROMISE, either decode its required push ID + QPACK field section before emitting the standard `push_promise` frame or use the custom diagnostic until push support exists.

The same honesty rule applies to client/server QPACK failure: a standard `headers` frame with `headers:[]` asserts that an empty HTTP field section was successfully decoded. If QPACK decode failed, preserve the boundary with a decode-failure/malformed diagnostic rather than manufacturing an empty successful field section.

Required tests: malformed GOAWAY remains identifiable as a malformed **GOAWAY**, not unknown; valid PUSH_PROMISE is not reported as unknown; QPACK-failed HEADERS does not serialize a standard empty `headers` array.

### 6. `priority_field_value_bytes` is not a standard H3 qlog field

The standard PRIORITY_UPDATE shape has required text `priority_field_value`; there is no standardized `priority_field_value_bytes`. Preserving malformed bytes is useful, but this must be a documented custom field/event (for example `tardigrade_priority_field_value_bytes`) rather than an unqualified field that looks like part of the advertised http3-13 schema. The parsed frame boundary still needs to remain observable when the priority value is malformed.

### Acceptance validation still outstanding

PR #606 was merged with its description still stating that qvis was unavailable. #255 Slice 1 explicitly requires manually loading one produced `.sqlog` in qvis as the final compatibility check. That is not a code change, but it still needs to be performed and recorded before the Slice 1 prerequisite is actually complete.

### Non-blocking naming cleanup

`http3.qlog.max_serialized_record_len = 64 KiB` is documented as a recommended buffer, not a true maximum; legal field sections can exceed it. The surrounding docs now correctly tell the composition root to size to its configured field-section budget / stream records / retain `NoSpaceLeft`, so the policy itself is acceptable. I would rename the constant to `recommended_record_buffer_len` (or derive a real maximum) so a future concrete sink does not accidentally treat 64 KiB as a protocol bound.

Because this is the author's own PR, GitHub will not accept a formal REQUEST_CHANGES review from this account, so this is submitted as a blocking COMMENT review. Once the six code items above plus the qvis validation are addressed, the consolidated checklist is satisfied; I do not have another untouched review category waiting behind these.

Jaruso (commented):
## Follow-up verification — `9d43f3d`

This pass verifies the exact remaining items from review #4908686873 against the one new commit (`Address qlog verification feedback`). Most of that checklist is now fixed correctly: `connection_closed` uses the standard trigger vocabulary; `extended_connect` / `h3_datagram` are numeric; reserved SETTINGS no longer carry `name_bytes`; malformed known frames are no longer rewritten as `unknown`; malformed PRIORITY_UPDATE bytes are explicitly marked with a Tardigrade extension field; and the prior lifetime/settings/parse-boundary fixes remain intact.

I still see **three code blockers**:

1. **Packet-number validity is not enforced.** `packet_number` is optional now, but the serializer emits it whenever the caller supplies one. The current QUIC qlog PacketHeader permits `packet_number` only for Initial/Handshake/0RTT/1RTT, so `.retry` or `.version_negotiation` can still serialize an invalid standard event. Gate serialization on packet type (or encode the relationship in the type) and strengthen the test by deliberately supplying a packet number to Retry/VN and asserting it is suppressed/rejected.

2. **The default-disabled runtime still pays the second QPACK decode.** `Conn.pumpRequests()` now skips the observer only when `self.events.emitFn == null`, but the production runtime unconditionally installs `h3ConnectionEvent` even when `h3_sink` is the default no-op sink. Therefore the condition is always true in production and request HEADERS are still decoded once for diagnostics and once for semantics with qlog disabled. Only install the H3 EventSink when the selected qlog sink actually has an `emit_fn`, or carry an explicit enabled bit through the observer.

3. **Every PUSH_PROMISE is still classified as malformed.** `eventFrameFromRaw()` returns `.malformed` unconditionally for `.push_promise`, so a syntactically valid server PUSH_PROMISE on a client request stream is logged as `tardigrade_malformed_payload`. h3-events-13 defines PUSH_PROMISE as a known frame with required `push_id` and decoded `headers`. Either decode those fields and emit the standard frame, or use a distinct custom `unsupported/not_decoded` diagnostic; do not label a valid-but-unimplemented payload as malformed. Add a valid PUSH_PROMISE test, not only the intentionally incomplete payload test.

### Validation

CI run #1364 is still finishing, but every completed build/unit/H3/TLS/security/packaging job I checked is green; the remaining appliance jobs were still in progress at review time.

The non-code acceptance gate is also still outstanding: I found no new PR/issue record showing that a generated `.sqlog` was manually loaded in qvis. #255 Slice 1 explicitly lists that compatibility check, and #606 was merged while documenting that qvis was unavailable.

**Once these three code items are fixed and the qvis load is recorded, the exhaustive checklist is satisfied.** I do not have another unresolved category behind these findings. GitHub does not allow this account to request changes on its own PR, so this is submitted as a blocking COMMENT review.

Jaruso (commented):
## Verification follow-up — `ba16273`

The three code blockers from review #4909028618 are fixed on this head:

- packet RX/TX now suppresses `packet_number` for Retry / Version Negotiation even if a caller supplies one;
- the runtime no longer installs the H3 event observer when the selected qlog sink is the default no-op, so disabled qlog avoids the diagnostic QPACK decode path;
- PUSH_PROMISE now preserves `push_id` plus decoded fields and the runtime/qlog adapter carries them through.

I checked the one-commit delta for regressions in those fixes. **One code blocker remains, introduced by the new PUSH_PROMISE decoder:** every `qpack.decode()` failure is converted to `.malformed`, but this helper is explicitly the static-only decoder and also has fixed 128-field / 4096-byte scratch limits. Its error set includes `DynamicTableReference`, `InvalidRequiredInsertCount`, `InvalidBase`, `Blocked`, `ScratchOverflow`, and `TooManyFields`; those do not all mean the wire field section is malformed. In particular, #255 already documents that the production dynamic decoder is not composed yet, while QPACK legitimately allows encoded field sections to reference the dynamic table. A valid PUSH_PROMISE that uses negotiated dynamic QPACK (or exceeds this diagnostic scratch capacity) can therefore still be falsely logged as `tardigrade_malformed_payload`.

Please separate **malformed** from **not decoded by this diagnostic helper**. A clean shape is a Tardigrade-only `qpack_decode_failed` / `undecoded` diagnostic that preserves the known frame type, decoded `push_id` when available, raw length, and bounded reason/error. Reserve `.malformed` for errors that prove malformed wire independently of the implementation's static-only/capacity limitations. Alternatively, decode through the real dynamic decoder once that production composition exists.

Add a regression test where the PUSH_PROMISE field section is syntactically valid but cannot be decoded by this static helper (for example a dynamic-table field section under non-zero QPACK capacity, or a valid field section exceeding the observer's local field/scratch capacity) and assert it is **not** reported as malformed.

The non-code acceptance gate is still outstanding: I found no PR or #255 issue record showing that a generated `.sqlog` was manually loaded in qvis. That Slice 1 compatibility check still needs to be performed and recorded.

CI run #1366 is currently in progress. Every completed job I checked is green; unit/build/appliance/TLS jobs are still finishing and there are no failures at this snapshot.

**After this PUSH_PROMISE diagnostic distinction and the qvis validation are addressed, the exhaustive checklist is satisfied.** This finding is not a newly opened review category; it is a regression in the implementation added specifically to fix the previous PUSH_PROMISE blocker. GitHub cannot accept REQUEST_CHANGES from the PR author, so this remains a blocking COMMENT review.

PR #606, opened 2026-08-11
Refresh qlog serializers for current drafts

Jaruso:
## Summary
- update QUIC qlog to draft-qualified `quic:*` names, current JSON-SEQ `QlogFileSeq` header fields, `.sqlog` constants, packet `header`/`raw` payload shapes, `decryption_failure` drops, and `quic:recovery_metrics_updated`
- update HTTP/3 qlog to `http3:*` names, current SETTINGS field names, `stream_type_set`, representative frame payloads, and a `tardigrade:*` QPACK blocking extension instead of the removed standard QPACK namespace
- refresh qlog docs and the composition smoke test for draft-14/draft-13 schema URIs and `.sqlog` artifacts

Refs #255 (slice 1 only)

## Validation
- `zig build test-quic --summary all --error-style verbose` (755/755 tests passed)

## Notes
- qvis is not installed in this environment (`command -v qvis` returned no path), so I could not perform the requested manual qvis load check locally.

### DIFF
--- a/docs/QUIC_QLOG.md
+++ b/docs/QUIC_QLOG.md
@@ -112,19 +112,19 @@ Names below are `namespace:event`. Transport events (`src/quic/qlog.zig`):
 
 | Requirement (#255)        | qlog event                          | Key data |
 |---------------------------|-------------------------------------|----------|
-| handshake                 | `quic:connection_started`           | odcid/scid/dcid lengths |
+| handshake                 | `quic:connection_started`           | required `local` / `remote` endpoint objects, plus Tardigrade CID-length diagnostics |
 |                           | `tardigrade:quic_handshake_progressed` | `stage` (started -> confirmed / failed) |
 |                           | `quic:connection_closed`            | `reason`, optional `error_code` |
-|                           | `quic:key_updated`                  | `key_phase` |
-| packet sent               | `quic:packet_sent`                  | type, number, length, ack-eliciting |
-| packet received           | `quic:packet_received`              | type, number, length |
-| packet lost               | `quic:packet_lost`                  | type, number |
+|                           | `quic:key_updated`                  | required `key_type`, optional full `key_phase` and `trigger` |
+| packet sent               | `quic:packet_sent`                  | `header`, `raw`, Tardigrade ack-eliciting diagnostic |
+| packet received           | `quic:packet_received`              | `header`, `raw` |
+| packet lost               | `quic:packet_lost`                  | `header` |
 | recovery metrics          | `quic:recovery_metrics_updated`     | RTT/PTO/cwnd/bytes-in-flight updates |
 | **deprotection failure**  | `quic:packet_dropped`               | `trigger:"decryption_failure"` |
 | PATH_CHALLENGE/RESPONSE   | `tardigrade:quic_path_validation`   | `phase` (challenge/response sent/received, validated, failed) |
-| migration                 | `quic:migration_state_updated`      | `trigger` (nat_rebinding/active), `state` (accepted/blocked) |
+| migration                 | `quic:migration_state_updated`      | required `new` migration state, optional `old` |
 | stream reset              | `tardigrade:quic_stream_reset`      | `direction` (reset/stop-sending, sent/received), stream id, error code |
-| flow-control blocked      | `quic:connection_data_blocked_updated` / `quic:stream_data_blocked_updated` | optional stream id, limit |
+| flow-control blocked      | `quic:connection_data_blocked_updated` / `quic:stream_data_blocked_updated` | required `new` blocked state, optional `old` / `reason`; stream variant requires `stream_id` |
 
 Application events (`src/http3/qlog.zig`):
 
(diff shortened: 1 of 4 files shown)


Jaruso (commented):
I reviewed this against the Slice 1 requirements recorded on #255 and the July 2026 qlog drafts (`main-schema-14`, `quic-events-13`, `h3-events-13`). The namespace/schema-URI/`.sqlog` direction is correct, and moving QPACK blocking to an explicit `tardigrade:*` extension is the right call. CI is also green.

**Blocking findings before merge:** several events now use standardized `quic:*` / `http3:*` names while retaining legacy payload shapes that do not satisfy the corresponding draft-13 CDDL. In particular: `connection_started`, `key_updated`, the data-blocked state events, `migration_state_updated`, and representative H3 HEADERS/SETTINGS/GOAWAY frames need their required fields/shapes updated. There is also a clock-contract mismatch: `Record.time_us` is documented as monotonic, while the trace header declares system time relative to the Unix epoch.

The Slice 1 test criterion also calls for golden/parse coverage of representative events. The current substring assertions can (and currently do) pass schema-invalid payloads, so please strengthen these tests to validate complete record shapes. Finally, the issue explicitly lists a manual qvis load as the compatibility gate; the PR notes that this has not yet been performed. That check still needs to happen before this slice is considered complete.

I left detailed inline comments with concrete expected shapes. After those are addressed, please update the corresponding event catalogue in `docs/QUIC_QLOG.md`, rerun `test-quic`, and validate a produced `.sqlog` in qvis.

Note: I would mark this **Request changes**, but GitHub does not allow the PR author to request changes on their own PR, so this is submitted as a COMMENT review.

Jaruso (commented):
Re-review against the previous blocking threads and the July 2026 qlog drafts.

The prior structural findings are substantially fixed: `connection_started` now carries the required endpoint objects, `key_updated` has the required key type/full phase shape, migration and blocked-state events use `old`/`new`, the header clock is correctly monotonic/unknown-epoch, H3 frames are variant-specific, and the smoke test now parses records and locks full golden JSON. CI is green on this head.

I still do not think this slice is merge-ready. The remaining inline findings are schema/serialization correctness issues:

1. `$BlockedReason` values do not match quic-events-13, so any event that sets the current reason values is schema-invalid.
2. The public `KeyType` model omits both 0-RTT values from the draft even though this project supports 0-RTT and this slice is intended to finish the serializer model before live wiring.
3. H3 SETTINGS names are free-form/IANA-style strings; draft-13 defines lowercase qlog enum names and requires `name_bytes` for unknown settings. The current golden test therefore locks a schema-invalid standard SETTINGS frame.
4. Dynamic HTTP field/setting text is interpolated directly into JSON without escaping. Valid field values containing quotes, backslashes, or control characters can corrupt the JSON-SEQ record. The trace header has the same issue for customizable title/description.

Please address these before merging. The manual qvis load requested by Slice 1 also remains unverified per the PR description; after the serializer fixes, that should still be done as the final compatibility check.

Jaruso (commented):
Re-review at `3c6ea667f39252e628eb8b392119feeb9a720518`.

The four blockers from my previous pass are addressed: the full draft-13 `KeyType` set is present, `$BlockedReason` now matches the standard values, SETTINGS names are typed/lowercase with `unknown` validation, and dynamic JSON strings are escaped. The earlier connection/migration/blocked-state/H3-frame/clock/golden-test fixes also remain intact. CI is green.

I found two remaining serializer-contract issues below. Both should be fixed before this Slice 1 schema becomes the production wiring target:

1. SETTINGS currently allows `name_bytes` on a known setting name, even though draft-13 defines `name_bytes` only when `name == "unknown"`.
2. H3 HEADERS/SETTINGS payloads are now variable-sized, but the serializer contract/docs still say a 512-byte buffer covers every event. That is no longer true and will become a predictable qlog-drop path once real header sections are wired.

There is also still one acceptance/validation item outside the code: the PR description says the required manual qvis load check has not been performed. The issue's Slice 1 plan explicitly calls for loading one produced trace in qvis as the final compatibility check.

Because this is the author's own PR, GitHub does not permit this account to submit a formal REQUEST_CHANGES review; treat the two inline comments and qvis validation gap as blocking.

Jaruso (commented):
Re-review of af599fcd. The two blockers from the prior pass are fixed: HTTP/3 SETTINGS now enforces the `unknown`/`name_bytes` invariant in both directions, and the variable-size H3 record-buffer contract is documented and covered with a >512-byte test. CI #1356 is green.

I found one remaining schema blocker in the existing QUIC surface: `quic:connection_closed` still allows `error_code` without either `connection_error:"unknown"` or `application_error:"unknown"`, which draft-13 explicitly requires for that field. I left the concrete fix options inline.

Separately, Slice 1 acceptance criterion #6 is still outstanding: the PR description says qvis was unavailable, so one produced `.sqlog` has not yet been manually loaded in qvis. Once the `connection_closed` model is corrected and that manual compatibility check is recorded, I do not see another blocker in this slice.

Jaruso (commented):
Re-review at `f5135389c` is clean from a code/schema perspective. The `connection_closed` fix correctly ties raw `error_code` values to either `connection_error:"unknown"` or `application_error:"unknown"`, matching quic-events-13, and tests cover both variants. I also rechecked the previously reported serializer/model issues; those fixes remain intact, all prior review threads are now resolved, and CI #1358 is green.

Against #255 Slice 1, implementation criteria 1–5 are satisfied: current `QlogFileSeq`/`.sqlog` output, `quic:*`/`http3:*` draft-13 shapes, draft-qualified schema URIs, explicit `tardigrade:*` extensions, and parse/golden coverage for the representative QUIC/H3 records.

The only remaining acceptance item is Slice 1 criterion #6: manually load one produced trace in qvis. The PR description still states that qvis is unavailable in the current environment and that this check has not been performed. I have no further code changes to request, but I would complete/document that qvis validation before merging this slice.

Issue #256, opened 2026-07-02
quic: add UDP buffer tuning, DPLPMTUD, ECN, and packet pacing performance work


Jaruso:
## Summary

Track the operational/performance hardening needed for pure Zig QUIC once the correctness path exists.

## Motivation

HTTP/3/QUIC is not automatically faster than h2. UDP buffer sizing, packet pacing, batching/GSO, ECN, and path MTU behavior can dominate performance and reliability at high throughput. These should be planned explicitly rather than discovered late.

## Scope

- UDP send/receive buffer guidance and optional socket tuning hooks.
- Max UDP payload size plumbing.
- DPLPMTUD / PMTU discovery strategy.
- Black-hole detection / fallback to safe datagram size.
- ECN receive/send plumbing where supported.
- Packet pacing implementation or integration with #244 pacing hooks.
- Future batching/GSO/GRO design notes.
- Benchmarks for:
  - small object
  - large object
  - streaming proxy
  - loss/reordering
  - high-bandwidth local loopback / Beelink-style runs
- Operator docs for sysctl/socket buffer recommendations.

## Non-goals

- BBR.
- Multipath.
- Kernel eBPF/reuseport routing.
- Replacing correctness tests with perf tests.

## Acceptance criteria

- Tardigrade documents safe UDP buffer defaults and operator tuning guidance.
- QUIC respects configured max UDP payload size.
- DPLPMTUD or a clearly documented conservative MTU policy exists.
- Pacing hooks from #244 are exercised by a real send path.
- Benchmarks report throughput, p95/p99, CPU/request, loss/PTO, packet counts, and UDP/socket settings.
- No perf feature bypasses flow control, congestion control, or anti-amplification rules.

## Testing strategy

- Unit tests for max-payload-size selection and PMTU state.
- Integration benchmarks with fixed environment metadata.
- Manual/CI perf job with artifacts.
- Regression tests for black-hole fallback if DPLPMTUD is implemented.

## References

- RFC 8899 — Packetization Layer PMTU Discovery
- RFC 9000 — QUIC max UDP payload and ECN considerations
- RFC 9002 — pacing/recovery relationship
- quic-go UDP buffer guidance
- performance research on QUIC pacing and high-rate links
- #240 Pure Zig QUIC and HTTP/3 Foundation
- #244 recovery/pacing hooks
- #149 competitive benchmark suite

Jaruso:
[...]

PR #607, opened 2026-08-11
perf(quic): make the outbound datagram size a real transport invariant (#256-A)

Jaruso:
## Summary

Slice **256-A** only: make the outbound QUIC datagram size a real transport invariant and reconcile every send/receive/recovery surface that depends on it. Pacing (256-C), socket buffer tuning (256-D), ECN (256-E), and the benchmark matrix (256-G) are not in this PR. DPLPMTUD (256-B) is not implemented; this PR provides the sender/path-size seam it will own.

### Receive capacity is not send size

RFC 9000 §18.2 `max_udp_payload_size` is receive capacity, not path MTU. The two directions are now separate:

- **`Config.max_udp_payload_size`** is local receive capacity only. It defaults to `datagram.max_size`; validation refuses to advertise more than the receive/deprotection buffers can process.
- **`Config.max_send_udp_payload_size`** is the sender's assumed current path size, defaulting to the RFC 9000 §14 floor of 1200. `TARDIGRADE_HTTP3_MAX_DATAGRAM_SIZE` maps here.

### Effective datagram size

`src/quic/datagram.zig` owns the bounds:

```zig
current_path_max: u64 = base_size,
send_ceiling:     u64 = max_size,
peer_max:        ?u64 = null,
// ordinary: min(current_path_max, send_ceiling, peer_max)
// PMTU probe ceiling: min(send_ceiling, peer_max)
```

`Connection.effectiveMaxDatagramSize()` derives the ordinary cap on demand, so authenticated peer parameters take effect immediately. Before peer parameters authenticate, sends stay at 1200. The independent probe ceiling gives #256-B headroom without requiring the operator to preconfigure a larger current path size.

### Recovery and congestion control

Making sender MDS dynamic required the recovery path to follow it consistently:

- **NewReno uses the sender's current MDS** for minimum/initial-window behavior, additive increase, and migration reset state.
- **Ordinary in-flight packets are bounded by remaining cwnd including final packet overhead and mandatory padding.** Raising the datagram cap cannot widen congestion-window overshoot.
- **`ack_eliciting` and `in_flight` are separate.** A packet made in-flight by PADDING is tracked and charged even when it is not ack-eliciting.
- **Candidate-path PATH_CHALLENGE/PATH_RESPONSE traffic is congestion-gated before path-control state is dequeued.**
- **Ordinary recovery tracking remains bounded.** Ordinary STREAM/CRYPTO traffic stops before the fixed recovery reserve instead of failing open.
- **Required recovery traffic cannot deadlock on that fixed bound.** PTO probes and mandatory-padding recovery packets use the fixed reserve first, then allocator-backed *recovery-only* overflow. PTO expiration never fabricates loss or retires an outstanding original merely to reclaim a bookkeeping slot; originals remain tracked and count in `bytes_in_flight` until normal ACK, loss detection, or key discard retires them.
- Recovery overflow participates in ACK processing, loss detection, packet-space discard, in-flight/PTO queries, and loss-deadline calculation just like the fixed tier.
- The parallel semantic `sent_records` store grows only when recovery overflow needs it. Because records can contain copied stateless-reset tokens, growth uses an explicit **copy → secure-wipe old backing → free** sequence rather than `ArrayList`'s normal unwiped growth path.

### Initial padding / coalescing

RFC 9000 §14.1's 1200-byte obligation stays with the Initial packet itself. Merely having later Handshake/Application write keys does **not** transfer responsibility to a later packet, because that later packet may still be blocked by tracker, congestion, or anti-amplification gates. The Initial therefore makes the datagram legal before any optional later packet is considered; later packets can still coalesce when the effective datagram cap has remaining room.

### Docs

`docs/HTTP3_ROLLOUT.md` documents the receive-vs-send distinction, the conservative 1200 default, peer limiting, and the current `>1200` operator assertion. Without #256-B black-hole fallback, an incorrectly raised path assertion can remain stalled because Tardigrade's own recovery retries the same oversized path assumption.

## Test plan

- [x] `zig build test-quic` — QUIC/H3 suite green on the final source.
- [x] `zig build test` — full unit suite green on the final source.
- [x] `zig fmt` on the touched QUIC files.
- [x] Datagram units cover floor/ceiling clamping, peer lowering, send ceiling, probe headroom, and out-of-range clamping.
- [x] Connection sizing regressions assert actual emitted datagram sizes, conservative pre-auth behavior, receive/send separation, peer limits, and raised path sizes.
- [x] Congestion regressions cover near-full cwnd, spent cwnd, mandatory PATH_RESPONSE padding, candidate-path gating, PTO cwnd exemption, and exact packet/BIF accounting.
- [x] Recovery/NewReno units cover 1200/1452/2048 MDS, minimum window, CA growth, loss/persistent congestion, migration reset, and padded non-ack-eliciting packets.
- [x] Ordinary tracker saturation backpressures rather than sending retransmittable data untracked.
- [x] A real PTO at the ordinary threshold is emitted with a tracker/sent-record entry and its bytes drain on ACK.
- [x] **Sustained total loss crosses the old 128-packet fixed limit and multiple PTO rounds beyond the fixed reserve; every required probe remains tracked and charged using recovery overflow.**
- [x] **Exact last-ordinary-slot Initial+Handshake regression:** Initial consumes the last ordinary tracker slot, Handshake remains pending, and the returned Initial-bearing UDP datagram is still exactly 1200 bytes with correct congestion accounting.
- [x] Runtime/config regressions pin the operator knob to the send side and keep receive-capacity advertisement aligned with actual ingress buffers.

Refs #256

### DIFF
--- a/docs/HTTP3_ROLLOUT.md
+++ b/docs/HTTP3_ROLLOUT.md
@@ -64,8 +64,10 @@ Consequences worth knowing before tuning it:
 - The 2048 ceiling is the transport's work-buffer size, not a path property.
 
 Nothing about this setting relaxes congestion control, flow control, or the
-server's anti-amplification budget; those gates run after the size cap and
-still bound every byte sent.
+server's anti-amplification budget. In particular, an ordinary packet's
+in-flight content is bounded by the *remaining congestion window*, not by this
+cap, so raising the cap cannot widen how far in-flight bytes cross the window.
+Pure ACKs (not in flight) and PTO probes keep their RFC 9002 exemptions.
 
 ## Reload
 
--- a/docs/HTTP3_ROLLOUT.md
+++ b/docs/HTTP3_ROLLOUT.md
@@ -31,6 +31,42 @@ During drain or rollback withdrawal, eligible TCP responses emit
 upstream `Alt-Svc` response headers and emits at most one gateway-owned
 `Alt-Svc` value.
 
+## Datagram size
+
+`TARDIGRADE_HTTP3_MAX_DATAGRAM_SIZE` is the **local** bound on the UDP payload
+of every datagram this process sends, and the `max_udp_payload_size` transport
+parameter it advertises to peers. It defaults to **1200** — the size RFC 9000
+§14 requires every QUIC path to carry — and is clamped into `[1200, 2048]`.
+
+It is only one of the inputs to the size actually emitted. The transport
+resolves one *effective* cap per connection as the smallest of:
+
+1. this local configured maximum;
+2. the peer's advertised `max_udp_payload_size`, once its transport parameters
+   are authenticated;
+3. the validated path size.
+
+Consequences worth knowing before tuning it:
+
+- **The cap sits at 1200 for the whole handshake.** A raised local value only
+  takes effect once the peer has authenticated and committed to accepting
+  larger datagrams. Datagrams carrying Initial packets are always padded to
+  1200 regardless.
+- **A peer always wins when it asks for less.** Raising this value can never
+  push a datagram past what the peer advertised.
+- **Raising it is an assertion about the path**, not a measurement. Tardigrade
+  does not yet run DPLPMTUD (RFC 8899), so a value above 1200 says "I know this
+  path carries this much". If it does not, those datagrams are dropped and the
+  connection stalls until the peer's PTO retransmits. Raise it only for paths
+  whose MTU you control end to end (a dedicated link, a loopback or same-rack
+  benchmark host, a known-jumbo-frame fabric). Leave it at the default on the
+  open internet.
+- The 2048 ceiling is the transport's work-buffer size, not a path property.
+
+Nothing about this setting relaxes congestion control, flow control, or the
+server's anti-amplification budget; those gates run after the size cap and
+still bound every byte sent.
+
 ## Reload
 
 Advertisement-only knobs can hot reload:
--- a/docs/HTTP3_ROLLOUT.md
+++ b/docs/HTTP3_ROLLOUT.md
@@ -33,22 +33,29 @@ upstream `Alt-Svc` response headers and emits at most one gateway-owned
 
 ## Datagram size
 
-`TARDIGRADE_HTTP3_MAX_DATAGRAM_SIZE` is the **local** bound on the UDP payload
-of every datagram this process sends, and the `max_udp_payload_size` transport
-parameter it advertises to peers. It defaults to **1200** — the size RFC 9000
-§14 requires every QUIC path to carry — and is clamped into `[1200, 2048]`.
+Two different quantities share the name "max UDP payload size" in QUIC, and
+Tardigrade keeps them separate.
 
-It is only one of the inputs to the size actually emitted. The transport
-resolves one *effective* cap per connection as the smallest of:
+**Receive capacity** is what this endpoint tells peers it is willing to receive,
+via the `max_udp_payload_size` transport parameter. It is a property of the
+receive buffers the implementation allocates — **2048 bytes** — not something an
+operator tunes, and it never changes for the life of a connection. The config
+layer refuses to advertise more than the transport can actually deprotect.
 
-1. this local configured maximum;
-2. the peer's advertised `max_udp_payload_size`, once its transport parameters
-   are authenticated;
-3. the validated path size.
+**Send size** is what this process puts on the wire, and that is what
+`TARDIGRADE_HTTP3_MAX_DATAGRAM_SIZE` controls. It defaults to **1200** — the
+size RFC 9000 §14 requires every QUIC path to carry — and is clamped into
+`[1200, 2048]`. The transport resolves the size actually emitted as the smallest
+of:
+
+1. the current path size (this knob, until DPLPMTUD replaces it);
+2. this endpoint's send ceiling (2048);
+3. the peer's advertised receive capacity, once its transport parameters are
+   authenticated.
 
 Consequences worth knowing before tuning it:
 
-- **The cap sits at 1200 for the whole handshake.** A raised local value only
+- **The send size sits at 1200 for the whole handshake.** A raised value only
   takes effect once the peer has authenticated and committed to accepting
   larger datagrams. Datagrams carrying Initial packets are always padded to
   1200 regardless.
@@ -56,18 +63,33 @@ Consequences worth knowing before tuning it:
   push a datagram past what the peer advertised.
 - **Raising it is an assertion about the path**, not a measurement. Tardigrade
   does not yet run DPLPMTUD (RFC 8899), so a value above 1200 says "I know this
-  path carries this much". If it does not, those datagrams are dropped and the
-  connection stalls until the peer's PTO retransmits. Raise it only for paths
-  whose MTU you control end to end (a dedicated link, a loopback or same-rack
-  benchmark host, a known-jumbo-frame fabric). Leave it at the default on the
-  open internet.
-- The 2048 ceiling is the transport's work-buffer size, not a path property.
+  path carries this much". If it does not, those datagrams are dropped in the
+  network. Tardigrade's own loss recovery will notice and retransmit — but with
+  no PMTU black-hole fallback yet, the retransmissions are the same oversized
+  datagrams, so the connection can stall indefinitely rather than recovering at
+  a smaller size. Raise it only for paths whose MTU you control end to end (a
+  dedicated link, a loopback or same-rack benchmark host, a known-jumbo-frame
+  fabric). Leave it at the default on the open internet.
 
 Nothing about this setting relaxes congestion control, flow control, or the
-server's anti-amplification budget. In particular, an ordinary packet's
-in-flight content is bounded by the *remaining congestion window*, not by this
-cap, so raising the cap cannot widen how far in-flight bytes cross the window.
-Pure ACKs (not in flight) and PTO probes keep their RFC 9002 exemptions.
+server's anti-amplification budget:
+
+- An in-flight packet's content is bounded by the *remaining congestion
+  window*, not by the datagram size, so raising the size cannot widen how far
+  in-flight bytes cross the window. The budget covers the packet's **final**
+  size, including the padding RFC 9000 mandates for Initial-bearing and
+  path-validation datagrams.
+- Path validation is delayed when the window cannot cover its padded datagram
+  (RFC 9000 §8.2 permits this) rather than being sent uncharged.
+- Recovery's packet tracker is bounded; when it is full the transport
+  backpressures instead of emitting in-flight packets it cannot track.
+- Pure ACKs (never in flight) and PTO probes keep their RFC 9002 exemptions.
+  A packet carrying PADDING is in flight even when nothing in it is
+  ack-eliciting, and is charged accordingly.
+
+RFC 9002's NewReno windows are defined in terms of the sender's current maximum
+datagram size, so the initial window, minimum window, and congestion-avoidance
+growth all scale with the value above rather than with a fixed 1200.
 
 ## Reload
 
--- a/src/edge_config.zig
+++ b/src/edge_config.zig
@@ -935,9 +935,11 @@ pub fn loadFromEnv(allocator: std.mem.Allocator) !EdgeConfig {
     const http3_retry_policy_str = envOrDefault(allocator, "TARDIGRADE_HTTP3_RETRY_POLICY", "off") catch unreachable;
     defer allocator.free(http3_retry_policy_str);
     const http3_retry_policy = try parseHttp3RetryPolicyConfig(http3_retry_policy_str);
-    // #256-A: one authoritative default, owned by `quic.datagram`. The
-    // transport clamps this into `[base_size, max_size]` and lowers it further
-    // whenever the peer advertises a smaller `max_udp_payload_size`.
+    // #256-A: the **send** size, owned by `quic.datagram`. This is not the
+    // `max_udp_payload_size` this endpoint advertises — that is receive
+    // capacity, fixed by the transport's buffers. The transport clamps this
+    // into `[base_size, max_size]` and lowers it further whenever the peer
+    // advertises smaller receive capacity.
     const http3_max_datagram_size = parseIntEnv(usize, allocator, "TARDIGRADE_HTTP3_MAX_DATAGRAM_SIZE", http.quic.datagram.base_size);
     const proxy_protocol_mode_str = envOrDefault(allocator, "TARDIGRADE_PROXY_PROTOCOL", "off") catch unreachable;
     defer allocator.free(proxy_protocol_mode_str);
--- a/src/edge_config.zig
+++ b/src/edge_config.zig
@@ -935,7 +935,10 @@ pub fn loadFromEnv(allocator: std.mem.Allocator) !EdgeConfig {
     const http3_retry_policy_str = envOrDefault(allocator, "TARDIGRADE_HTTP3_RETRY_POLICY", "off") catch unreachable;
     defer allocator.free(http3_retry_policy_str);
     const http3_retry_policy = try parseHttp3RetryPolicyConfig(http3_retry_policy_str);
-    const http3_max_datagram_size = parseIntEnv(usize, allocator, "TARDIGRADE_HTTP3_MAX_DATAGRAM_SIZE", 1350);
+    // #256-A: one authoritative default, owned by `quic.datagram`. The
+    // transport clamps this into `[base_size, max_size]` and lowers it further
+    // whenever the peer advertises a smaller `max_udp_payload_size`.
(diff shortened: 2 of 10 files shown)


Jaruso (commented):
**Blocking finding — changes needed before merge.** GitHub does not allow the connected identity to submit `REQUEST_CHANGES` on its own PR, so this is recorded as a comment review instead.

I reviewed all 7 changed files against the current #256 handoff and the stated 256-A scope.

The sizing work itself is in good shape: there is now one default/bounds module, authenticated peer transport parameters lower the cap, the handshake stays at 1200, emitted datagrams are covered by driver-level tests, candidate-path control traffic is capped, the runtime/config defaults no longer drift, and the existing restart-required reload guard already includes `http3_max_datagram_size`. The PR is also correctly scoped: DPLPMTUD, pacing, socket buffers, ECN, and the benchmark closeout remain follow-up slices rather than being half-implemented here.

The blocker is congestion coupling. Raising the effective datagram cap to 1452–2048 does not raise or otherwise couple the congestion send gate, which remains based on `recovery.max_datagram_size == 1200`. A connection with only 600 bytes of cwnd room can therefore still admit application data and then build a packet up to the newly raised effective cap. The recovery controller accounts the full emitted size afterward, so the new knob can materially increase how far `bytes_in_flight` crosses `congestion_window`. That conflicts with #256's explicit requirement that performance work not bypass congestion control.

Please couple the data-send budget to the effective datagram size/cwnd room (or otherwise prove an equivalent hard gate) and add a deterministic regression with the cwnd nearly full. The existing `<= congestion_window + seen.largest` assertion is too weak because it permits exactly the one-full-raised-datagram overshoot that needs to be caught.

A useful regression shape would be: establish with `max_udp_payload_size = 2048`, queue stream data, arrange `congestion_window - bytes_in_flight` to be between the old 600-byte threshold and the new packet size, and prove an ordinary non-PTO data packet cannot increase in-flight bytes beyond the allowed cwnd budget. Pure ACKs and PTO probes should retain their existing exemptions.

I did not find another blocking issue in this slice.

Jaruso (commented):
## Blocking — exhaustive second-pass review

I re-reviewed the current head (`6722d852`) end to end: all seven changed files, the #256-A acceptance slice, the prior #248 config contract, the path-validation/recovery code this diff interacts with, every production `Connection` transmit/receive path I could trace, and the completed CI matrix.

The **two findings from my previous review are fixed**. The ordinary application-data path now has a real near-full-cwnd regression and the previous threads are resolved.

This pass found the following remaining blockers. These are the complete actionable issues I found on the current head; please address them together before another review.

### 1. Separate advertised receive capability from sender/path maximum datagram size

The new model currently treats `Config.max_udp_payload_size` as both:

- the local endpoint's advertised QUIC `max_udp_payload_size` transport parameter, **and**
- the local sender/path cap used by `Limits.local_max`.

Those are different quantities. RFC 9000 defines `max_udp_payload_size` as the largest UDP payload **this endpoint is willing to receive**; it is endpoint receive capacity, not the current path's PMTU/MDS. The sender's current maximum datagram size is path-dependent and is separately constrained by the peer's advertised receive limit.

This conflation causes several concrete problems:

- the default `1200` advertises to the peer that it may never send us more than 1200 bytes, even though the runtime/connection allocates 2048-byte receive work buffers;
- `Limits.probeCeiling()` also defaults to 1200, so the seam intended for #256-B cannot actually probe above 1200 unless the operator pre-configures a larger value — DPLPMTUD cannot discover anything by default;
- `quic.config.Config.validate()` still accepts/advertises up to 65,527 while `Connection.ingestPacket()` has 2048-byte `work`/plaintext buffers, so direct users of the transport can advertise a receive capacity the implementation cannot process;
- `Connection.ingestPacket()` still hardcodes `[2048]` instead of using the new authoritative datagram symbol, so the claimed single work-buffer ceiling can drift again.

Please model these separately. Exact naming is up to you, but the semantics should be roughly:

```zig
// Endpoint capability advertised to the peer. Must not exceed actual receive
// capacity allocated by this implementation.
local_receive_max: u64 = datagram.max_size,

// Sender/path state. Safe current value starts at 1200; #256-B may raise it
// after validation. A separate configured/probe ceiling may be >1200.
current_path_max: u64 = datagram.base_size,
send_ceiling: u64 = datagram.max_size,

// Outbound ordinary datagram limit:
min(current_path_max, send_ceiling, peer.max_udp_payload_size)
```

The names/config UX can differ, but an immutable receive transport parameter must not double as mutable PMTU state. Add regressions proving (a) default ordinary sends stay conservative, (b) the advertised TP matches real receive capacity, (c) the default PMTU probe ceiling has headroom above 1200, and (d) all receive scratch buffers derive from the authoritative capacity.

### 2. Propagate the sender's current MDS into recovery/NewReno

Before this PR, the packet builder and `recovery.max_datagram_size` were both 1200. This PR makes the actual sender MDS dynamic up to 2048, but `recovery.zig` still hardcodes 1200 for:

- initial congestion window;
- minimum congestion window;
- congestion-avoidance additive increase;
- congestion reset after migration.

RFC 9002's `max_datagram_size` is the **sender's current maximum payload size**, and the NewReno calculations are defined in terms of that current value. Once #256-A makes MDS dynamic, recovery cannot remain a compile-time 1200 constant.

Please make MDS controller/connection state and update it when the active/current path MDS changes. Use the current validated/send MDS, not a future PMTU probe ceiling. Tests should cover 1452/2048 MDS, initial/minimum cwnd, congestion-avoidance growth, and migration reset behavior.

### 3. Congestion budgeting must include the final packet size, especially mandatory PADDING

The new `data_budget` fixes the first bug I reported for ordinary STREAM/CRYPTO content, but it caps content **before** the later mandatory padding step. The final in-flight packet can therefore still exceed the available congestion window.

Concrete case: queue an active-path `PATH_RESPONSE`, leave 700 bytes of cwnd room, and use a raised datagram cap. The frame fits the new `data_budget`; later the packet is expanded to 1200 bytes as required for path validation; then recovery charges the final ~1200 bytes. This is not a PTO exception, so the final `bytes_in_flight` exceeds cwnd.

Also, `ack_eliciting` and `in_flight` are not synonymous. A packet containing PADDING is in flight even if it contains no ack-eliciting frame. Today recovery tracking is guarded by `if (record.ack_eliciting)`, so a padded non-ack-eliciting packet can escape bytes-in-flight accounting entirely.

Please budget the **final in-flight packet size** (headers + AEAD overhead + required padding) before committing/dequeuing ordinary in-flight content, and track `in_flight` separately from `ack_eliciting`. Preserve the PTO exception, but PTO probes must still be counted in flight.

Required regressions:

```text
active PATH_RESPONSE + 700 bytes cwnd room -> deferred, still pending
active PATH_RESPONSE + >= required final size -> sent and BIF <= cwnd
padded non-ack-eliciting packet -> tracked as in-flight
PTO probe -> may exceed cwnd, but is still charged to BIF
```

### 4. Candidate-path validation packets currently have no congestion-window gate

`buildCandidatePacket()` now respects the new datagram cap and the candidate path's anti-amplification budget, but it still sends a 1200-byte PATH_CHALLENGE/PATH_RESPONSE regardless of congestion-window availability and only calls `recovery.onPacketSent()` afterward.

A path-validation packet is not automatically an RFC 9002 PTO probe. RFC 9000 explicitly permits PATH_RESPONSE transmission to be delayed when constrained by congestion control. Please gate candidate-path in-flight traffic before it is committed/sent. Ideally that uses path-specific congestion state; if the current architecture still shares recovery during validation, use a conservative applicable window rather than no gate at all. Do not dequeue a PATH_RESPONSE or clear `needs_send` until the packet is actually allowed to transmit.

Add tests for a spent congestion window leaving candidate control pending, then transmission/accounting once sufficient window becomes available.

### 5. Recovery tracker exhaustion must backpressure instead of sending untracked packets

This is especially important for #256's high-bandwidth target. `recovery.max_tracked_packets` is 128. `RecoveryController.onPacketSent()` adds the packet to the tracker **before** charging congestion. When the tracker is full, `Connection` catches that error and deliberately sends the already-sealed packet as "untracked best-effort".

That means after 128 outstanding packets, additional STREAM/CRYPTO/control packets can be sent without being added to `bytes_in_flight` and without recovery metadata. This is both a congestion-control bypass and a reliability hole at exactly the high-BDP workloads this performance ticket is preparing for.

Do not send retransmittable/in-flight content unless recovery can track it. Prefer preflighting/reserving tracker capacity before dequeuing frames and sealing the packet; if the bounded tracker is full, backpressure until ACK/loss frees a slot. Pure ACK-only packets that truly are not in flight can remain sendable.

Please add a saturation regression: fill the tracker, prove queued STREAM/CRYPTO data remains queued and no uncharged datagram leaves, free one slot via ACK/loss, then prove the next packet sends and is tracked/charged.

### Documentation correction

The rollout doc says a bad >1200 path assertion causes the connection to stall "until the peer's PTO retransmits." The peer cannot retransmit a Tardigrade-originated datagram it never received. **Tardigrade's** loss recovery/PTO will retry it; and because #256-B has not added PMTU black-hole fallback yet, those retries can remain oversized and the connection can remain stalled indefinitely. Please correct that failure description while touching this section.

---

### What I checked and am not asking to change

- The previous ordinary-data cwnd overshoot is fixed.
- Peer lowering of outbound size is correctly applied after authenticated transport parameters.
- The 1200-byte pre-authentication send behavior remains conservative.
- The explicit emitted-datagram cap tests are meaningful.
- The anti-amplification regression is sound for the path it exercises.
- HTTP/3 datagram-size config remains restart-required through the existing #257 reload guard; no hot-reload divergence found.
- The runtime/config/root wiring itself is clean once the receive-vs-send semantics above are corrected.
- Current CI on `6722d852` is fully green across the jobs that ran; these remaining issues are protocol/state-machine gaps not exercised by the current suite.

GitHub cannot let this account select `REQUEST_CHANGES` on its own PR, so this is submitted as a COMMENT review, but **the five items above are blocking before merge**.

Jaruso (commented):
**One blocking recovery issue remains on `c43d102f`; everything else from the prior review is fixed.** GitHub will not let the PR author submit `REQUEST_CHANGES` on their own PR, so this is a comment review, but I would not merge until the item below is fixed.

I re-reviewed all 10 changed files against the 256-A scope and #256's explicit requirement that this performance work not bypass congestion control/recovery. I also re-checked the current RFC 9000/9002 invariants and the current CI run. The receive-capability/send-MDS split is now correct, the DPLPMTUD probe ceiling has headroom, NewReno uses the live sender MDS, final mandatory padding is congestion-accounted, `in_flight` is separated from `ack_eliciting`, candidate-path validation is congestion-gated before dequeue, and the black-hole documentation is corrected. I resolved those prior threads. CI run 31517478650 is green across the unit/build/platform/appliance/H3 interop/perf-smoke checks.

### Remaining blocker: tracker saturation still fails open for PTO/in-flight control paths

`can_track` currently participates only in `can_send_data`. Later, the PTO branches explicitly use `probe` as an alternate admission condition (`can_send_data or probe`), and `want_ack` can also carry the builder forward independently. Therefore, once `RecoveryController` reaches `max_tracked_packets`, a PTO with queued CRYPTO/STREAM content (or an ACK that becomes in-flight due required PADDING) can still be sealed and returned.

At that point `recovery.onPacketSent()` necessarily returns `TooManyTrackedPackets`. The catch path adds the packet size directly to the congestion controller but creates **no tracker entry and no published sent record**. That is not safe degradation: an ACK/loss can never remove those bytes from `congestion.bytes_in_flight`, and retransmittable content carried by the packet has no recovery record to requeue. Repeated occurrences can permanently inflate bytes-in-flight and stall recovery.

RFC 9002 makes the fix slightly subtler than simply refusing probes: §6.2.4 requires at least one ack-eliciting PTO probe, and §7.5 says PTO probes are exempt from the congestion-window admission gate **but still count as bytes in flight**. So the implementation needs to guarantee tracking capacity for recovery traffic—for example, reserve recovery-only tracker capacity and stop ordinary traffic before consuming it, or use another bounded representation that can always record required PTO probes. The same guarantee should be made at the final `recordIsInFlight(record)` decision so a nominally ACK-only packet that becomes in-flight through mandatory PADDING cannot fall into the catch path either.

Please add regressions that (1) drive ordinary tracking to its backpressure threshold, fire a real PTO, and prove the probe is emitted **with** a tracker/sent-record entry and its bytes are later removed by ACK/loss; (2) exercise repeated PTOs near capacity so the reserve is not a one-probe illusion; and (3) prove an ACK-only packet that becomes in-flight through required PADDING is never sealed without a trackable slot.

The current PR description also needs a refresh before merge. It still describes `max_udp_payload_size` as the send knob, the old `Limits{local_max, validated_path_max}` model, and says congestion-control math is untouched. `c43d102f` intentionally changed all three. Update the summary/test-plan terminology to the new receive-capacity vs `max_send_udp_payload_size` model and the recovery/padding/tracker changes so the PR record matches what will actually merge.

Aside from this tracker/PTO root cause and the stale PR description, I did not find another remaining issue in the 256-A slice.

Jaruso (commented):
**Blocking — two correctness issues remain on current head `4ab6b34a`.** GitHub will not let this connected identity request changes on its own PR, so this is recorded as a comment review.

I re-reviewed the full 256-A slice and then diffed the only new commit from `c43d102f`. The previous findings are otherwise fixed: receive capacity is separate from sender/path MDS; the effective cap reaches the emitted packet path; NewReno uses live MDS for its formulas/reset; ordinary traffic cannot straddle cwnd; mandatory Initial/PATH_RESPONSE padding is congestion-accounted; candidate-path validation is gated; padded packets distinguish `in_flight` from `ack_eliciting`; and the prior immediate `probe`-bypasses-tracker bug is gone. I resolved that old thread.

Two edge cases still violate transport/recovery invariants:

1. **A fixed eight-entry recovery reserve only postpones tracker exhaustion; it does not make PTO progress safe.** RFC 9002 §6.2.4 requires at least one ack-eliciting probe on every PTO expiration. PTO expiration itself does *not* establish loss, so under a total-loss episode with no newer ACK, the existing tracked packets do not free slots. Starting at the ordinary threshold (120), four PTO expirations can consume the eight-slot reserve if two probes are emitted per PTO. The next PTO is suppressed because `canTrackRecoveryPacket()` is false. That turns a bounded bookkeeping implementation into a recovery deadlock until idle timeout. The reserve needs to support recovery for the whole connection lifetime, not just the first few PTOs—e.g. dynamically grow/compact recovery metadata while ordinary traffic remains backpressured, or use a rigorously derived bound that covers every PTO possible before the negotiated idle timeout. A larger magic constant is not sufficient. Add a total-loss regression that fires more PTOs than the reserve can hold, delivers no ACKs, and proves every PTO before idle timeout still emits at least one *tracked* ack-eliciting probe.

2. **Initial padding is deferred based on later write-key availability, not on whether a later packet will actually be emitted.** The new tracker threshold makes this deterministic: with Initial + Handshake write keys and pending flights, start one slot below the ordinary threshold. Initial sees a later key, so `is_last_level=false` and does not take responsibility for 1200-byte expansion. It sends and consumes the last ordinary slot. Handshake then sees `canTrackPacket()==false` and can return `null`. `pollTransmitOnPath()` returns the already-committed Initial packet alone, below 1200. RFC 9000 §14.1 requires every client Initial-bearing datagram, and every server datagram carrying an ack-eliciting Initial, to be at least 1200 bytes. Presence of a later key is not a guarantee of later contribution. Plan the coalesced datagram before committing the Initial, reserve the later packet's required resources, or make the Initial pad itself unless a later packet is guaranteed to be emitted. Add the one-slot-before-threshold coalesced Initial+Handshake regression and assert the returned Initial-bearing UDP payload is always >=1200 with correct tracker/cwnd accounting.

Conceptually:

```zig
// Before committing the first packet of an Initial-bearing datagram:
const plan = planDatagram(levels, tracker_capacity, cwnd, amp_budget);
// Only defer Initial padding when `plan` guarantees a later packet is emitted.
// Recovery capacity for PTO must remain available for every PTO before idle,
// not merely `reserved_tracked_packets` probe records.
```

Current CI run #1370 is green. These two states need focused regressions because the existing saturation tests intentionally accept waiting after the reserve fills and start fully saturated rather than crossing the coalescing threshold mid-datagram.

Jaruso (commented):
Final self-review after implementing the remaining blockers directly.

I re-audited the complete #256-A surface on `b1302537` after the fixes. The fixed tracker remains bounded for ordinary traffic; required PTO/mandatory-padding recovery remains fully trackable past the fixed reserve without inventing loss; recovery overflow participates in ACK/loss/key-discard/timer state; sent-record overflow preserves stateless-reset-token wiping; and Initial-bearing datagrams no longer depend on speculative later-level emission for the 1200-byte invariant.

Validation on this final source:
- `zig fmt` green
- `zig build test-quic` green
- `zig build test` green
- CI #1384 green across the build/test/interop/security/packaging/perf matrix
- both previously outstanding regressions covered and their threads resolved

I also rechecked the earlier #256-A blockers: receive-vs-send semantics, current NewReno MDS, final padding/in-flight accounting, candidate-path congestion gating, ordinary tracker backpressure, PTO congestion behavior, anti-amplification, and operator documentation. I found no regression and no additional blocker.

No remaining blockers found in #256-A. GitHub does not allow this identity to approve its own PR, so this COMMENT is the final no-blocker review.

PR #609, opened 2026-08-12
Wire QUIC qlog and shared TLS keylog

Jaruso:
## Summary
- Wire QUIC transport events through a connection-scoped runtime qlog observer, including ACK summaries, recovery metrics, stream reset/STOP_SENDING, and flow-control blocked events.
- Move keylog formatting to a TLS-owned helper and emit debug-only NSS keylog entries from the shared TLS transport event sink while preserving the QUIC compatibility import.
- Update qlog/keylog documentation for the implemented Slice 2 and Slice 4 seams.

## Validation

Part of #255: slices 2 and 4.


Jaruso (commented):
**Blocking review — changes are required before merge.** GitHub does not allow this account to submit `REQUEST_CHANGES` on its own PR, so this is submitted as a COMMENT review, but I do not consider the PR ready to merge.

I reviewed the current head (`7d5ae98`) end-to-end against the August 10 reconciliation on #255, specifically the Slice 2 and Slice 4 acceptance criteria this PR claims. The newest ACK-for-unsent-packet commit is also included in this pass, and the current CI run is green.

The overall layering is moving in the right direction: QUIC remains destination-agnostic, the HTTP/3 runtime owns qlog fan-out, and the NSS formatter moved into TLS. Both claimed slices are still materially incomplete, however, and several of the new paths produce incorrect observability data in real connections.

## Slice 4 — shared TLS keylog is not yet operational end-to-end

The main issues are the ClientHello correlation value and record-mode KeyUpdate handling.

The keylog context should own the ClientHello random for the lifetime of the handshake and derive the endpoint role from the TLS driver, rather than accepting a borrowed random pointer plus a caller-controlled role. A shape along these lines avoids both the server-side bootstrap problem and dangling pointers:

```zig
pub const Context = struct {
    enabled: bool = false,
    role: state.Role,
    client_random: [client_random_len]u8 = [_]u8{0} ** client_random_len,
    has_client_random: bool = false,
    sink: Sink = .{},

    pub fn setClientRandom(self: *Context, random: []const u8) error{InvalidClientRandom}!void {
        if (random.len != client_random_len) return error.InvalidClientRandom;
        @memcpy(&self.client_random, random);
        self.has_client_random = true;
    }
};
```

The client path needs to install `entropy.hello_random` before the 0-RTT secret can be emitted; the server path needs to copy the random from the parsed ClientHello before emitting any traffic secret, retaining the ClientHello1 value across HRR. `Driver.init` / the role-specific wrapper should set `.role` so a client cannot accidentally emit SERVER labels by leaving the current default in place.

For KeyUpdate, the generation number and the newly derived traffic secret need to come from the same state transition. Today `EventSink.emitKeyUpdate()` increments only a label counter, while `record_epoch_bridge` derives the next secret later. A safer shape is for the bridge's successful key update to be the source of truth:

```zig
const next = try deriveNextTrafficSecret(...);
self.write_key_generation += 1; // or read, by direction
keylog_ctx.emitApplicationSecret(
    direction,
    self.write_key_generation,
    next[0..digest_len],
);
// install/retain the secret, then wipe scratch
```

Do not fix this by merely consulting the context's current mutable generation from the bridge: multiple KeyUpdate events can be accumulated in one driver event batch before the bridge consumes the first one. Carry an explicit generation or use the bridge's generation counter at the point the new secret actually exists.

Also thread keylog configuration through a supported record-mode API (`PureZigRecordStream` constructor/options or a pre-handshake setter). The current record constructors leave the shared event sink's keylog context at its default; requiring callers to reach into `handshake_driver.sink` is not a usable shared TLS callback contract.

Please add integration tests that drive real client/server handshakes and real record-mode KeyUpdate processing. The current synthetic `emitKeyUpdate(); emitSecret(...)` test exercises a sequence the production backend does not produce. Cover generation 0 and 1+, client/server direction, 0-RTT, SHA-256/SHA-384 secret lengths, HRR correlation, missing correlation, and disabled/no-callback behavior.

## Slice 2 — required transition surface is incomplete

The issue explicitly calls for congestion/persistent-congestion transitions, stream opened/closed/reset/STOP_SENDING transitions, and connection/stream flow-control blocked **and unblocked** transitions. This PR adds ACK/recovery snapshots, reset/STOP_SENDING, and `flow_control_blocked`, but not the rest. It also does not emit recovery metrics at every material RTT/CWND/bytes-in-flight change (for example packet send, key discard, and path-controller reset/migration can change the snapshot).

Please make these actual state transitions rather than poll-time observations. A transport-neutral event shape could be along these lines:

```zig
flow_control_state_changed: struct {
    scope: enum { connection, stream },
    stream_id: ?StreamId = null,
    local: bool,
    old: BlockedState,
    new: BlockedState,
},
stream_state_changed: struct { id: StreamId, old: StreamState, new: StreamState },
congestion_state_changed: struct { old: CongestionState, new: CongestionState },
persistent_congestion,
```

Emit only when state actually changes, and add focused tests for blocked→unblocked, open→closed, congestion recovery/persistent congestion, plus invalid stream control frames producing no successful transition event.

## qlog correctness / lifecycle

The new runtime bridge currently has several fidelity problems called out inline:

- application packet-number space cannot distinguish 0-RTT from 1-RTT, so real 0-RTT packets are mislabeled as `1RTT`;
- the runtime never emits `quic:connection_started`, while `.state = established/closed` creates duplicate or semantically incorrect confirmation/close records alongside the explicit handshake/close events;
- a plural aggregate `packets_lost` event is converted into one standard singular `quic:packet_lost`, which under-represents multi-packet loss episodes;
- qlog-off still performs event mapping and a timestamp read for every mapped transport event, including every packet RX/TX, because the QUIC observer must remain installed for metrics.

Carry actual packet type/encryption level through the normalized event seam, make connection lifecycle events exactly-once, and either emit one standard `packet_lost` per packet or use an explicitly namespaced aggregate event with a count/bytes payload.

For the disabled fast path, short-circuit before conversion/timestamping, e.g.:

```zig
if (observer.qlog_sink.isEnabled()) {
    if (quicEventToQlog(event)) |qlog_event| {
        observer.qlog_sink.log(nowUs(), qlog_event);
    }
}
```

Finally, `src/quic/keylog.zig` is described as a compatibility re-export, but it changes the public `Label`, `Entry`/`Sink`, and `labelFor` API. Either keep a generation-0 legacy wrapper for the existing QUIC surface while exposing the new TLS API, or explicitly treat/document this as a breaking public API change rather than claiming compatibility.

I left inline comments on each concrete defect. After these are addressed, the important regression tests are sequence/integration tests rather than only one-event adapter tests: real 0-RTT packet → qlog, real connection lifecycle → exactly one start/confirm/close, blocked→unblocked state, rejected STOP_SENDING → no transition event, real TLS handshake → keylog, and real record KeyUpdate → `_1`/later secret output.

Jaruso (commented):
**Blocking review — changes are still required before merge.**

I re-reviewed the full PR at `f7bba6f08108f5154499008f478093831721cff8`, all 12 changed files, all four commits, the Aug. 10 #255 Slice 2 / Slice 4 acceptance criteria, the existing review threads, and the unchanged producer/consumer code that these new seams depend on (`stream.zig`, `recovery.zig`, `record_epoch_bridge.zig`, and `encrypted_stream.zig`). I also checked the current CI run: it is green across unit tests, TLS/H3 interop, appliance profiles, builds, lint/format, security, and perf smoke.

The latest fix commit did resolve several previous findings: ClientHello random ownership/capture is now safe, QUIC keylog role is set correctly, invalid STOP_SENDING is no longer logged before validation, 0-RTT packet typing is preserved, aggregate loss is no longer mislabeled as a singular standard qlog event, the disabled-qlog mapper/timestamp path is guarded, and the legacy `quic.keylog` API is restored.

The remaining blockers are below and are all grounded in live producer/consumer paths rather than helper-only tests:

1. **Real record-mode KeyUpdate still cannot emit `*_TRAFFIC_SECRET_1+`.** The next secret is derived later in `record_epoch_bridge`, outside `EventSink.emitSecret`; no real path feeds that value into keylog.
2. **The record-mode stream has no keylog-context composition path.** `PureZigRecordStream` still constructs `RecordHandshakeDriver.init(role, backend)` with the default context, so this shared callback is not actually usable by the production record carrier and client role would remain the context default if somebody only wired a sink later.
3. **Slice 2 remains incomplete:** no stream-open/stream-close event, no congestion-state/persistent-congestion event, and recovery snapshots still omit several material state changes (send-side bytes-in-flight, key discard, path migration reset).
4. **PTO observability is still stale for a valid ACK that newly acknowledges zero tracked packets.** The transport correctly resets the PTO backoff, but the snapshot is only emitted inside `acked_count > 0`.
5. **Peer flow-control reporting is not a state machine.** Every DATA_BLOCKED/STREAM_DATA_BLOCKED is emitted as `unblocked -> blocked`, with no dedupe and no peer-unblocked transition; then the qlog adapter discards the `local`/peer distinction entirely.
6. **The automatic RESET_STREAM generated in response to STOP_SENDING is not emitted as a local reset transition.** Explicit `resetStream()` is observable; the RFC-mandated automatic path is not.
7. **Close observability is still not one semantic transition.** Every retransmitted close datagram re-emits `close_sent` and therefore another `quic:connection_closed`; additionally, a pre-handshake application close converted to a transport `INTERNAL_ERROR` reports the original application error code instead of the wire code.
8. **Application closes are mapped with the wrong qlog trigger.** The qlog model explicitly has `.application` (and already tests it), but both sent and received application closes are serialized with trigger `"error"`.

Please add end-to-end/transition tests for these paths rather than only mapper/helper tests. In particular, Slice 4 needs a real record-mode client/server KeyUpdate test (including repeated updates and SHA-256/SHA-384), and Slice 2 needs stream lifecycle, congestion/persistent-congestion, peer/local blocked→unblocked, zero-newly-acked PTO reset, automatic RESET_STREAM, and close-retransmission tests.

GitHub does not permit this account to submit `REQUEST_CHANGES` on its own PR, so this is submitted as a COMMENT review, but it is intentionally blocking.

Jaruso (commented):
Blocking re-review on `01e6391` — **not ready to merge yet**.

I rechecked the full PR against the Aug. 10 #255 Slice 2 / Slice 4 reconciliation, all prior review findings, and the latest CI (green). The latest two commits fixed a substantial amount correctly: ClientHello-random ownership/role handling, real record-mode KeyUpdate secret emission, record-stream keylog composition, PTO reset reporting, automatic RESET_STREAM observability, peer/local flow-control separation, 0-RTT packet typing, aggregate loss semantics, default-off qlog conversion cost, application close mapping, and QUIC keylog API compatibility. I resolved the corresponding threads after verifying the actual production paths/tests.

What remains:

1. **The existing Slice 2 blocker is still open.** `Connection.Event` still has no stream-open/stream-close lifecycle event and no congestion/persistent-congestion transition. `emitRecoveryMetrics()` is still limited to ACK/loss/PTO, so packet sends, key discard, and path-migration recovery reset can change bytes-in-flight / RTT / CWND without a corresponding snapshot. This is explicitly required by the #255 Slice 2 acceptance comment.
2. **The new per-stream blocked-state tracker survives terminal send resets.** A blocked stream that is reset stays in `local_stream_flow_blocked`; a reordered/later `MAX_STREAM_DATA` can then manufacture an `unblocked` event after `reset_sent`, and the map retains terminal stream IDs unnecessarily. Inline comment has the concrete cleanup expected.
3. **`close_sent` no longer means “sent.”** It is now emitted from `startClose()` before a close packet is constructed, and it carries the semantic application code even when the eventual pre-1RTT wire frame is converted to a transport `INTERNAL_ERROR`. That makes the public transport event and `h3_interop_tool` log factually wrong. Use a semantic `local_close_started`/`close_started` transition for qlog, or keep `close_sent` at the actual first wire emission with a latch and the actual wire type/code.
4. **Slice 4 still lacks the acceptance-level integration matrix.** The real KeyUpdate keylog test only exercises the default SHA-256 record profile. #255 explicitly asks for 0-RTT and SHA-384 coverage too; those need to prove the callback sees the real 48-byte SHA-384 secret and a real `CLIENT_EARLY_TRAFFIC_SECRET` with the handshake-populated ClientHello random.
5. The keylog docs still describe `EventSink.emitSecret` as the single record-mode choke point, but generation-N KeyUpdate secrets are now emitted directly from `record_epoch_bridge`. Update the architecture/safety description to match the implementation.

CI is green, but these are behavioral/acceptance issues rather than build failures. Once the remaining Slice 2 surface and the inline items are fixed, I would do one final full event-sequence/keylog pass before merge.

Jaruso (commented):
Blocking re-review — changes are still required before merge. GitHub does not allow this account to submit `REQUEST_CHANGES` on its own PR, so this is a COMMENT review, but I do not consider the PR ready yet.

I re-reviewed the full current head (`2f5ac12`) against the Aug. 10 #255 Slice 2 / Slice 4 acceptance criteria and against every blocker from the previous pass. The latest commit genuinely fixes the stale blocked-state cleanup, the close-sent wire/semantic split, the keylog architecture documentation, and the missing real SHA-384 + accepted-0RTT keylog coverage; I resolved those prior threads. CI is green.

The remaining blockers are all in Slice 2 transport/qlog semantics:

1. Stream lifecycle events are still reconstructed incompletely in `Connection`: the first received STREAM event is hard-coded to `open` after the StreamManager has already applied the frame, so a first FIN-only STREAM is reported with the wrong state, and peer-side close/reset transitions are not comprehensively surfaced.
2. Congestion-state inference currently treats `cwnd == minWindow` as persistent congestion even though recovery does not actually detect an RFC persistent-congestion episode, and ACK-driven exit from recovery is not observed.
3. The newly-added stream/congestion/persistent events are explicitly discarded by the runtime qlog adapter, so they do not satisfy the acceptance requirement to wire those transitions into qlog.
4. Recovery snapshots still miss two real bytes-in-flight mutations: candidate-path in-flight sends and Retry’s direct Initial-space discard.
5. qlog `connection_closed` can still be emitted twice when this endpoint starts a local close and then receives the peer’s CONNECTION_CLOSE while in closing state.

Please address the inline comments below and add transition-sequence tests around each path. Once those are fixed, I’ll re-run the complete Slice 2/4 acceptance pass rather than only checking the local patches.

Jaruso (commented):
Blocking re-review on `68f2856` — this is much closer, but I still would not merge it yet.

The concrete five findings from the prior pass were addressed well: stream receive/reset lifecycle transitions now report their actual post-mutation state, candidate-path/Retry recovery snapshots are present, stream/congestion/persistent events are actually routed to qlog, and semantic close logging is latched exactly once. CI is green. I resolved those fixed threads.

I found three remaining acceptance/correctness issues in the new implementation:

1. The new persistent-congestion calculation is not RFC 9002 §7.6.2 compliant. It only examines losses from one `detectLost()` call, includes non-ack-eliciting ordinary packets as boundaries, cannot detect an acknowledged packet inside the candidate interval, does not prove an RTT sample existed before the boundary packets were sent, and uses `ptoDuration(space)` even though persistent-congestion duration must include `max_ack_delay` regardless of packet-number space. Requiring `time_threshold_losses > 0` also misses a qualifying interval whose boundary packets are declared lost by packet threshold. This can both falsely report and fail to report persistent congestion.
2. The newly emitted standard `quic:stream_state_updated` records are missing the draft-13-required `stream_side` field. The current aggregate `Stream.state()` is not enough to represent the separate QUIC sending/receiving state machines, especially FIN/RESET transitions and simultaneous close. The normalized transport event needs to preserve which side changed (and ideally local/remote trigger), then qlog can serialize the standard shape.
3. `packets_acked` is currently emitted as a Tardigrade extension because the code says current qlog has no standardized event. Draft-13 does define `quic:packets_acked`, with `packet_number_space` and `packet_numbers`. The connection currently discards the actual newly-acked packet numbers after reducing them to `{largest,count}`, so the standard event cannot be emitted.

I left the broad Slice 2 / persistent-congestion thread open and added targeted inline comments with implementation directions and regression cases. Slice 4/keylog remains clean from the previous review.

Jaruso (commented):
Blocking re-review on `59cc46c`. The three issues from the previous pass were substantially addressed: ACKs now preserve exact newly-acked packet numbers, stream events are side-aware, and persistent congestion has explicit recovery state plus prior-RTT/max-ack-delay handling. CI is green. I found three remaining correctness/schema issues in those new implementations, detailed inline below. Slice 4/keylog remains complete from the prior review. I am leaving the broader Slice 2 thread open until these are fixed.

Jaruso (commented):
Blocking: the ACK qlog schema and stream-side fixes from the previous pass are correct, and CI is green on this head. One Slice 2 correctness issue remains in the new persistent-congestion bookkeeping: after the fixed ACK-history capacity is reached, the implementation can no longer detect a later valid persistent-congestion interval, and per-space persistent state is not reset when that recovery space is discarded/restarted. Details and regression cases are in the inline comment. I found no additional Slice 4/keylog blocker in this pass.

Jaruso (commented):
One blocking recovery correctness issue remains; see inline comment. The previous lifetime ACK and Retry reset issues are fixed, CI is green, and I found no new blocker in the unchanged qlog or keylog paths.

Jaruso (commented):
The latest commit fixes the prior min/max overflow-gap bug: ACK barriers now preserve disjoint gaps, and the new regression exercises that exact case. I found one remaining blocker in the replacement representation.

`acked_send_time_ranges` is still a fixed `[max_tracked_packets]` array, but the number of relevant disjoint ACK ranges is not bounded by the number of concurrently tracked packets. A persistent-congestion candidate can remain active after one lost packet while the connection subsequently sends and ACKs packets one at a time. If those send timestamps are separated by >1us, each ACK remains a distinct range because the candidate's `start_us` keeps the evidence floor pinned. After 128 such ACKs, the 129th reaches the `std.debug.assert` here (and the following insertion would exceed the array in builds where the assertion is absent).

So the previous false-negative is fixed, but this can now fail a long-lived connection outright. Please make the storage correctness-preserving without relying on `max_tracked_packets` as a historical ACK-range bound. Acceptable approaches include exact spill storage, or more aggressive candidate/barrier finalization that *proves* old ranges can no longer affect any future candidate before discarding them. Do not merge disjoint ranges just to make room, since that recreates the false-negative fixed by this commit.

Please add a regression shaped like:
```zig
tracker.observePersistentCandidate(.application, .{
    .space = .application,
    .packet_number = 1,
    .time_sent_us = 0,
    .size = 100,
    .rtt_sample_available_at_send = true,
});

var pn: u64 = 2;
while (pn <= max_tracked_packets + 2) : (pn += 1) {
    const sent_at = pn * 10; // deliberately disjoint ranges
    try tracker.onPacketSent(.{
        .space = .application,
        .packet_number = pn,
        .time_sent_us = sent_at,
        .size = 100,
        .rtt_sample_available_at_send = true,
    });
    _ = tracker.onAcked(.application, pn, sent_at + 1);
}
```
This must not assert, overflow, discard a real ACK barrier, or synthesize gaps. Existing split-candidate, >128 lifetime ACK, Retry-reset, and gap-preservation tests should remain.

Everything else in this latest commit looks correct. I did not find a new qlog/keylog regression; this is the only blocking finding in the new delta.

Jaruso (commented):
Re-reviewed the latest head. The previous 128-range assertion/capacity blocker is fixed: exact ACK-barrier ranges now spill beyond the fixed array, the regression exercises the real send→ACK path, Retry cleanup still clears the state, and CI #1418 is green. I did not find a regression in the qlog/keylog/stream work.

One blocking recovery/resource-safety issue remains in the new spill path. ACK-barrier overflow is now allocator-backed from peer ACK processing, but OOM is converted to `@panic`, and the candidate/barrier lifetime is still unbounded. A single active candidate can remain pinned while arbitrarily many later disjoint ACK barriers accumulate, so one connection can grow this list until process memory exhaustion and then crash the process. The inline comment has the expected failure semantics and regression.

I cannot submit REQUEST_CHANGES on my own PR, so this is a blocking COMMENT review. Once the spill path is bounded/pruned and resource exhaustion is connection-local/fallible rather than a process panic, I do not currently see another Slice 2 / Slice 4 blocker.

Jaruso (commented):
Final verification pass on `9e2ff0a`. The prior blocker is genuinely fixed: ACK-barrier history is no longer allocator-unbounded, the peer ACK path no longer panics the process, capacity failure is propagated into `Connection.processAck()`, and CI #1421 passes.

I found one remaining blocker in the *lifetime* of that bounded history. The 256-range limit is reachable during ordinary healthy traffic, not just under memory pressure or an adversarial pattern. A single lost ack-eliciting packet seeds `persistent_candidates[space]`. Later ACKs whose send times are greater than `candidate.end_us` do not reset that one-packet candidate, and `ackEvidenceFloor()` stays pinned at `candidate.start_us`, so each disjoint later ACK remains retained. With packets sent/ACKed sequentially >1us apart and no further loss, there is at most one live packet at a time, but ACK #257 after that one loss deterministically returns `OutOfMemory`; `processAck()` converts it to `INTERNAL_ERROR` and closes an otherwise healthy connection.

This needs relevance-based candidate finalization, not a larger constant. When an ACK lies after the current candidate end, the old candidate can be retired once there is no still-outstanding candidate-eligible packet with a send time between `candidate.end_us` and that ACK's send time. At that point no future lost boundary can extend the old candidate without crossing the ACK, so resetting it is RFC-safe and lets existing pruning discard the obsolete barriers. Be sure to scan both fixed tracker entries and recovery overflow and restrict the bridge check to packets that could actually become persistent-congestion boundaries (ack-eliciting, in-flight, non-PMTU).

Please add two regressions together: (1) one lost candidate followed by well over the total 256-barrier budget of sequential send+ACK traffic must remain healthy and keep barrier storage bounded; (2) keep an outstanding candidate-eligible packet between the old candidate and a later ACK, then declare that packet lost and prove the old boundary was not finalized prematurely. That second test protects against fixing the capacity issue by creating a persistent-congestion false negative.

I rechecked the previously reviewed Slice 2/Slice 4 areas and the new commit only changes `connection.zig`/`recovery.zig`; no new qlog, stream-state, flow-control, or TLS/keylog regression surfaced. This is the only remaining blocker I found.

PR #610, opened 2026-08-12
perf(quic): discover the path MTU instead of assuming one (#256-B)

Jaruso:
## Summary

Slice B of #256. #256-A made the outbound datagram size one authoritative value; this makes that value **measured** rather than asserted.

- **`src/quic/pmtu.zig`** — a DPLPMTUD state machine (RFC 8899, RFC 9000 §14.3/§14.4). It lives in `path.Path`, so discovery is per network path *by construction*: a migration, or a tuple that has to be re-validated, starts over at 1200 rather than inheriting a size only the old path was shown to carry.
- **Probes are standalone datagrams** of exactly the size being validated, carrying PING + PADDING and nothing else — a probe dropped by a path too small for it costs no application progress. The first probe reaches straight for the ceiling (RFC 8899 §5.3 optimistic search); after that the window bisects to within 16 bytes. A size is only ruled out after three consecutive losses, so ordinary congestion cannot narrow the search.
- **Probe loss is not congestion** (RFC 9000 §14.4) — its bytes still leave the in-flight ledger, but no window reduction and no recovery period. A probe is otherwise fully congestion controlled and anti-amplification limited, with **none** of the RFC 9002 PTO exemptions.
- **Black-hole fallback** returns the size to the RFC 9000 §14 floor after three pieces of evidence, measured against the size actually in question: datagrams at or above the current send size lost while *smaller* ones are delivered, or consecutive PTOs with nothing acknowledged when every in-flight datagram is already oversized. The RFC 8899 raise timer (10 min) later re-enters the search, **discarding the previous failure bound** — it described a path condition that may no longer hold — so a path that regains MTU is rediscovered instead of being stuck at the size that once failed.
- **Feedback is scoped to the path incarnation that produced it.** Recovery deliberately keeps old-path packets in flight across a migration, so `SentRecord` stamps a `PathRef{key, generation}` and every ACK/loss/probe outcome resolves through it. A tuple that is re-validated gets a new generation, so outcomes still owed by its previous incarnation are dropped rather than applied to a fresh controller.

### Socket precondition

Discovery above 1200 requires the listener to establish a **no-IP-fragmentation contract** (RFC 8899 §3, RFC 9000 §14). Without it an acknowledged large probe would prove the peer *reassembled* it, not that the path carries it. Support is per-OS, because these constants are **not** shared across the BSDs — Darwin's IPv4 `IP_DONTFRAG` is 28, FreeBSD's is 67, and OpenBSD uses 28 for something else entirely:

| Platform | Mechanism |
| --- | --- |
| Linux | `IP_PMTUDISC_PROBE` / `IPV6_PMTUDISC_PROBE` — DF plus RFC 8899 §4.5's requirement that DPLPMTUD not be clamped by the kernel's own cached PMTU |
| Darwin (macOS/iOS/tvOS/watchOS) | `IP_DONTFRAG` (28) / `IPV6_DONTFRAG` (62) — DF only |
| FreeBSD | `IP_DONTFRAG` (67) / `IPV6_DONTFRAG` (62) — DF only |
| NetBSD, OpenBSD, DragonFly, others | none verified → conservative 1200 |

A wrong constant that a kernel happens to *accept* would report success without ever setting DF, so unverified platforms report failure rather than guessing. Where the contract cannot be established the listener warns and holds the send size at 1200, which is #256's documented conservative-policy alternative.

The QUIC transport's own `max_send_udp_payload_size` defaults to 1200 for the same reason: `src/quic/` owns no socket and cannot know whether a probe would be fragmented, so **discovery is opt-in by the composition root that created the socket**. An embedder using `quic.connection` directly stays at the conservative policy.

### Operator-visible change

`TARDIGRADE_HTTP3_MAX_DATAGRAM_SIZE` changes meaning and **defaults to 2048** (was 1200). It was an assertion — "this path carries this much" — and is now a **ceiling on what discovery may find**. The wider default is not a more aggressive one: the size actually emitted still starts at 1200 on every path and rises only where a probe was acknowledged *and* the socket contract holds. Lowering it is now the only reason to set it. See `docs/HTTP3_ROLLOUT.md`.

### Drive-by fix

`processAck` adopted the peer's `largest_acknowledged` without checking it against anything this endpoint had sent, so a peer acknowledging an unsent packet number (RFC 9000 §13.1) tripped `packetNumberLength`'s assertion — a remote panic in a safety-checked build. It now closes with `PROTOCOL_VIOLATION` before mutating any ACK reference state; declining the encoding reference remains as a backstop.

## Test plan

- [x] `zig build test` — full suite green (Zig 0.16).
- [x] `zig build build-tls-interop build-h3-interop` — the interop tools are **not** built by `zig build test`, which is how an earlier CI break got through; now part of the pre-push routine along with `zig fmt --check`.
- [x] **`pmtu.zig`** — optimistic first probe, bisection, retry-before-ruling-out, convergence *and* termination, ceiling clamping mid-flight, per-path reset, both black-hole signatures measured against the current PLPMTU, immediate corroboration, "congestion loss alone is not a black hole", "a loss below the send size is not MTU evidence", "PTO at the floor never falls back", raise-timer rediscovery upward, and a converged-at-ceiling search that does not re-arm.
- [x] **`recovery.zig`** — probe loss returns bytes without a congestion event; a probe lost alongside ordinary traffic charges only the ordinary bytes and takes the congestion timestamp from them.
- [x] **`connection.zig` driver** — over a harness that drops datagrams above a simulated path MTU: a 1200-only path is discovered as such with no black hole and no window cut; a 1452 path converges inside its real capacity and the discovered size reaches the wire; the probe datagram is exactly the probed size; a black hole falls back and arms the raise timer; peer capacity and configured ceiling each bound discovery; a probe waits for congestion window rather than taking an exemption; a migrated path revalidates; old-path loss and probe ACK after promotion leave the new path untouched; a re-validated tuple does not inherit its previous incarnation's feedback; a fallback completed by a smaller ACK publishes metric, event, and recovery size; a PTO driven by old-path packets is not charged to the new path while one with the active path's own traffic still is; ACKing an unsent packet closes with `PROTOCOL_VIOLATION` without panicking.
- [x] **`http3_runtime.zig`** — the config ceiling collapses to 1200 without the socket contract, and the no-fragmentation option is asserted to be accepted on Linux and macOS.

Refs #256

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,10 +25,14 @@ All notable user-facing changes to Tardigrade are documented here.
   than applied to the new one.
 
   Discovery above 1200 additionally requires the listener socket to establish a
-  no-IP-fragmentation contract (RFC 8899 §3: `IP_PMTUDISC_PROBE` on Linux,
-  `IP_DONTFRAG` on macOS/BSD). Without it an acknowledged large probe would be
-  measuring reassembly rather than the path, so on platforms where the kernel
-  refuses it the send size stays at 1200 regardless of configuration.
+  no-IP-fragmentation contract (RFC 8899 §3): `IP_PMTUDISC_PROBE` on Linux,
+  `IP_DONTFRAG` on Darwin and FreeBSD — whose IPv4 constants differ, so the
+  BSDs are handled per-OS rather than as a family. Without it an acknowledged
+  large probe would be measuring reassembly rather than the path, so where the
+  contract cannot be established the send size stays at 1200 regardless of
+  configuration. The QUIC transport's own default is 1200 for the same reason:
+  it owns no socket, so discovery is opt-in by whichever composition root
+  created one and can vouch for it.
 
   `TARDIGRADE_HTTP3_MAX_DATAGRAM_SIZE` changes meaning accordingly and now
   defaults to **2048**. It was an assertion — "this path carries this much" —
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,28 @@ All notable user-facing changes to Tardigrade are documented here.
 ## [Unreleased]
 
 ### Added
+- **QUIC discovers the path MTU instead of assuming one (#256-B)** — Tardigrade
+  now runs DPLPMTUD (RFC 8899, RFC 9000 §14.3/§14.4) per network path. Every
+  path starts at the 1200-byte size RFC 9000 §14 guarantees and rises only as
+  far as a padded probe datagram is actually acknowledged; the first probe
+  reaches straight for the ceiling, so a path that carries it is discovered in
+  one round trip. Probe loss is not treated as congestion (§14.4), but probes
+  are still congestion controlled and anti-amplification limited, with none of
+  the RFC 9002 PTO exemptions. When a path stops carrying the discovered size —
+  oversized datagrams lost while smaller ones arrive, or consecutive PTOs with
+  nothing acknowledged — the send size falls back to 1200 rather than
+  retransmitting the same oversized datagram until the idle timeout, and the
+  RFC 8899 raise timer later re-searches below the size that failed. Discovery
+  is per path and never inherited: a migration or a re-validated tuple starts
+  over at 1200.
+
+  `TARDIGRADE_HTTP3_MAX_DATAGRAM_SIZE` changes meaning accordingly and now
+  defaults to **2048**. It was an assertion — "this path carries this much" —
+  and is now a *ceiling on what discovery may find*. Raising it can no longer
+  put a size on the wire that nothing has measured, so the wider default is not
+  a more aggressive one: the size actually emitted still starts at 1200.
+  Lowering it is now the only reason to set it. See `docs/HTTP3_ROLLOUT.md`.
+
 - **Native TLS negotiates `record_size_limit` and can pad records (#359)** —
   the record transport now offers and honors RFC 8449's `record_size_limit`
   extension in both roles. Each side advertises the largest inner plaintext it
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,9 +16,19 @@ All notable user-facing changes to Tardigrade are documented here.
   oversized datagrams lost while smaller ones arrive, or consecutive PTOs with
   nothing acknowledged — the send size falls back to 1200 rather than
   retransmitting the same oversized datagram until the idle timeout, and the
-  RFC 8899 raise timer later re-searches below the size that failed. Discovery
-  is per path and never inherited: a migration or a re-validated tuple starts
-  over at 1200.
+  RFC 8899 raise timer (10 minutes) later re-enters the search, discarding the
+  previous failure bound — it described a path condition that may no longer
+  hold — so a path that regains MTU is rediscovered rather than being stuck at
+  the size that once failed. Discovery is per path *incarnation* and never
+  inherited: a migration, or a tuple that has to be re-validated, starts over
+  at 1200, and feedback still owed by a previous incarnation is dropped rather
+  than applied to the new one.
+
+  Discovery above 1200 additionally requires the listener socket to establish a
+  no-IP-fragmentation contract (RFC 8899 §3: `IP_PMTUDISC_PROBE` on Linux,
+  `IP_DONTFRAG` on macOS/BSD). Without it an acknowledged large probe would be
+  measuring reassembly rather than the path, so on platforms where the kernel
+  refuses it the send size stays at 1200 regardless of configuration.
 
   `TARDIGRADE_HTTP3_MAX_DATAGRAM_SIZE` changes meaning accordingly and now
   defaults to **2048**. It was an assertion — "this path carries this much" —
--- a/docs/HTTP3_ROLLOUT.md
+++ b/docs/HTTP3_ROLLOUT.md
@@ -42,34 +42,75 @@ receive buffers the implementation allocates — **2048 bytes** — not somethin
 operator tunes, and it never changes for the life of a connection. The config
 layer refuses to advertise more than the transport can actually deprotect.
 
-**Send size** is what this process puts on the wire, and that is what
-`TARDIGRADE_HTTP3_MAX_DATAGRAM_SIZE` controls. It defaults to **1200** — the
-size RFC 9000 §14 requires every QUIC path to carry — and is clamped into
-`[1200, 2048]`. The transport resolves the size actually emitted as the smallest
-of:
+**Send size** is what this process puts on the wire. It is *measured*, not
+configured: every path starts at **1200** — the size RFC 9000 §14 requires every
+QUIC path to carry — and only rises as far as DPLPMTUD has proven the path
+carries (see below). The transport resolves the size actually emitted as the
+smallest of:
 
-1. the current path size (this knob, until DPLPMTUD replaces it);
-2. this endpoint's send ceiling (2048);
+1. the current path size, as discovered by DPLPMTUD;
+2. this endpoint's send ceiling — `TARDIGRADE_HTTP3_MAX_DATAGRAM_SIZE`, clamped
+   into `[1200, 2048]` and defaulting to 2048;
 3. the peer's advertised receive capacity, once its transport parameters are
    authenticated.
 
-Consequences worth knowing before tuning it:
+`TARDIGRADE_HTTP3_MAX_DATAGRAM_SIZE` is therefore a **ceiling on what discovery
+may find**, not a size taken on trust. Consequences worth knowing:
 
-- **The send size sits at 1200 for the whole handshake.** A raised value only
-  takes effect once the peer has authenticated and committed to accepting
-  larger datagrams. Datagrams carrying Initial packets are always padded to
-  1200 regardless.
-- **A peer always wins when it asks for less.** Raising this value can never
-  push a datagram past what the peer advertised.
-- **Raising it is an assertion about the path**, not a measurement. Tardigrade
-  does not yet run DPLPMTUD (RFC 8899), so a value above 1200 says "I know this
-  path carries this much". If it does not, those datagrams are dropped in the
-  network. Tardigrade's own loss recovery will notice and retransmit — but with
-  no PMTU black-hole fallback yet, the retransmissions are the same oversized
-  datagrams, so the connection can stall indefinitely rather than recovering at
-  a smaller size. Raise it only for paths whose MTU you control end to end (a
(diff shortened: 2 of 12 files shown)


Jaruso (commented):
**Blocking review findings — changes required before merge.**

I reviewed all 11 changed files against the current #256 reconciliation/acceptance criteria, including the path-manager lifecycle, recovery migration behavior, packet tracking, probe construction, black-hole state machine, operator docs, and the full CI run. CI is green and the overall shape is good: the controller is owned by `Path`, probes are standalone/exact-sized and hard-gated by cwnd + anti-amplification, and probe loss is correctly removed from bytes-in-flight without causing a NewReno congestion event.

There are four protocol/correctness issues that still break the guarantees this slice claims:

1. PMTU ACK/loss feedback is routed to the *currently active* path, not the path the packet was sent on. `resetForPathMigration()` intentionally keeps old-path packets in recovery, so delayed ACK/loss after promotion can contaminate the new path's PMTU controller. This defeats the core per-path invariant.
2. Black-hole detection treats any ACK larger than 1200 as proof the current PLPMTU still works. If 1452-byte datagrams are black-holed while 1300-byte datagrams succeed, the 1300-byte ACK clears the evidence instead of corroborating it, and it also resets the stalled-PTO signal.
3. A normally converged search never arms `PMTU_RAISE_TIMER`, and `smallest_failed` is never relaxed. A path that later gains MTU on the same tuple can therefore never be rediscovered upward. RFC 8899's Search Complete behavior is specifically meant to periodically re-enter Search Phase for this case.
4. The unsent-packet ACK drive-by fix prevents an assertion but still accepts the invalid ACK into `largest_peer_acked`. Since this implementation can detect `largest_acknowledged >= next_pn`, handle the protocol violation before mutating ACK reference state instead of letting an attacker poison packet-number encoding state.

Please add deterministic regressions for each case: old-path ACK/loss after path promotion must not mutate the new controller; 1452 losses + successful 1300 traffic must fall back while a successful 1452 ACK clears evidence; a converged 1452 path that later supports 2048 must rediscover upward after the raise timer; and ACKing an unsent packet must close with `PROTOCOL_VIOLATION` without panicking.

I would not merge this revision until these four are addressed.

Jaruso (commented):
Re-review status: no new commits have landed since my prior review. The PR is still at `8b9a1f5d5cb980bb7fa478a24dd3d23ee607f936`, and all four blocking inline threads remain unresolved on the same code.

The outstanding blockers are therefore unchanged:

1. PMTU ACK/loss/probe feedback is attributed to the current active path instead of the path the packet was sent on, allowing old-path recovery feedback to mutate a newly promoted path.
2. `onOrdinaryAck()` clears black-hole evidence for any ACK above 1200 rather than comparing against the suspected PLPMTU, so successful smaller traffic such as 1300 bytes can mask failure of a 1452-byte PLPMTU.
3. A normal Search Complete below the ceiling never arms `PMTU_RAISE_TIMER`, and the retained failure bound prevents later same-path MTU increases from being rediscovered.
4. The unsent-packet ACK change avoids the packet-number assertion but still accepts and persists an invalid `largest_acknowledged` instead of closing with `PROTOCOL_VIOLATION`.

CI is still green on this same SHA, but these are protocol correctness issues not covered by the current tests. No additional findings can be meaningfully verified until a new commit addresses the existing threads. Do not merge this SHA.

Jaruso (commented):
Full re-review of current head `469cdf689b8dc5a79c363a4cdbfda4874e1a299b` against #256-B and RFC 8899 / RFC 9000.

The four findings from my previous review were meaningfully addressed: the unsent-ACK path now closes with PROTOCOL_VIOLATION; black-hole evidence compares against the current PLPMTU; Search Complete can rediscover a later MTU increase; and ACK/loss feedback now carries the egress PathKey instead of blindly using `activePlpmtu()`.

I then re-read the entire 11-file PR and the surrounding path/recovery/runtime code from scratch. I found four remaining blocking correctness gaps:

1. `PathKey` is not a sufficient identity for delayed PMTU feedback after a tuple is reset/revalidated. The same tuple can represent a fresh PMTU epoch while old packets for its previous incarnation are still in recovery.
2. A black-hole transition triggered by the newly-correct smaller-ACK corroboration path bypasses the connection-level transition publication, so metrics/events (and the immediate congestion-controller update) are missed.
3. Application PTO is still connection-wide, but every application PTO is charged as black-hole evidence to the current active path. Old-path packets deliberately survive migration, so they can drive false fallback on the new path.
4. The production UDP socket does not establish the no-IP-fragmentation contract DPLPMTUD requires. The runtime currently only sets `SO_REUSEADDR`; with a 2048-byte probe ceiling an ACK can otherwise validate fragmentation rather than the unfragmented path MTU.

I also left a documentation correction inline. In addition, please clean up adjacent stale contract text before merge: `src/quic/datagram.zig` still says 1200 is the operator-facing default and says an operator may assert `current_path_max`; both are obsolete under this PR. The PR body is also stale: it says the raise timer re-searches strictly below the failed size and describes the unsent-ACK fix as only declining the packet-number reference, while the current implementation correctly re-tests old failures and closes on the protocol violation. `docs/HTTP3_ROLLOUT.md` calls the new connection fields “Operator counters”, but the production `http3_runtime` event/metrics bridge does not currently surface `pmtu_updated` or those counters; either wire them through or describe them as connection-level diagnostics until #255 exposes them.

Current CI is green across Linux/macOS/ARM unit tests, appliance integration, H3 interop, builds, security and perf smoke. These findings are state-lifecycle and socket-contract cases not covered by that matrix.

**Blocking review: do not merge this head yet.**

Jaruso (commented):
Blocking review findings — changes still required before merge.

I re-reviewed the current head (`1c6ba7d568686f815b3f4d372e15b943292b0887`) from scratch: all 12 changed files, the latest fix commit, #256-B's reconciled acceptance criteria, path lifecycle/revalidation, connection-wide recovery across migration, PMTU ACK/loss/PTO attribution, the production UDP socket contract, direct `Connection` composition sites, docs, and the current CI run. CI 31595979403 is fully green across Linux/macOS/ARM unit/build/appliance jobs, H3 interop, security, packaging, and perf smoke.

The previous blockers are substantially fixed: generation-aware `PathRef` now prevents ordinary A-vs-B and validated A-v1-vs-A-v2 feedback leaks; ordinary ACK/loss/PTO fallback transitions share connection-level publication; the active-path PTO gate no longer charges old-path-only PTOs to the new path; the raise timer and unsent-ACK handling are correct; and the production HTTP/3 runtime now has a conservative 1200-byte fallback when its socket cannot establish a no-fragmentation policy.

I still found three correctness gaps:

1. A failed validation attempt can be retried with the *same generation*. Because recovery can retain the old attempt's PATH_CHALLENGE record, a delayed ACK from that failed attempt can later mutate the fresh controller after the retry is promoted and DPLPMTUD starts. A generation needs to identify every fresh validation attempt, not only a previously-validated tuple.
2. The no-fragmentation precondition is enforced only by `http3_runtime`, while the public QUIC `Config` now defaults `max_send_udp_payload_size` to 2048. Direct `Connection.init` callers therefore opt into >1200 discovery without proving their socket cannot fragment; the in-tree H3 interop tool and real-UDP smoke test are concrete examples. Keep the transport default conservative or make socket capability an explicit opt-in that defaults false.
3. The macOS/BSD socket branch assumes one numeric `IP_DONTFRAG` ABI across all BSDs. That is false: current FreeBSD defines IPv4 `IP_DONTFRAG` as 67 (IPv6 is 62), while OpenBSD option 28 is `IP_IPSEC_REMOTE_AUTH` and does not expose an IPv4 `IP_DONTFRAG` at all. Since `true` from this helper unlocks DPLPMTUD above 1200, an accepted-but-wrong option is a soundness failure, not merely a portability issue. Use verified per-OS constants/semantics and fall back to 1200 on OSes without a proven contract.

I also left one non-blocking stale-comment correction for the raise timer wording.

I would not merge this head yet. Because this is the authenticated author's own PR, I am submitting the review as a blocking COMMENT rather than a formal REQUEST_CHANGES review.

Jaruso (commented):
Re-review complete on `ce87a321c190548d87a23634903f3ab33533b36c` — **no blocking findings remain.**

I reviewed the one-commit delta from `1c6ba7d568686f815b3f4d372e15b943292b0887`, then re-read the full 12-file PR and the unchanged PMTU/recovery interactions rather than limiting this pass to the requested fixes. I also rechecked the current CI run and every prior review thread.

The last three blockers are correctly fixed:

- **Path incarnation after failed validation:** `.failed`, `.unvalidated`, and `.validated` now all receive a new generation when a fresh validation attempt begins; `.validating` / `.validated_pending_promotion` keep the current attempt's generation. The new driver regression covers delayed feedback from an expired attempt and proves it cannot mutate the replacement controller.
- **Socket capability ownership:** `quic.config.Config.max_send_udp_payload_size` is conservative at 1200 again, while the HTTP/3 runtime opts into its 2048 discovery ceiling only after its socket successfully establishes the no-fragmentation contract. Direct `Connection` embedders therefore cannot accidentally run unsound >1200 discovery.
- **Per-OS no-fragmentation options:** Linux uses PMTUDISC_PROBE, Darwin uses IPv4 28 / IPv6 62, FreeBSD uses IPv4 67 / IPv6 62, and unverified BSD families deliberately fall back to 1200. The docs and platform-gated tests match that support matrix.

The earlier blockers are also still fixed: per-path/per-incarnation ACK and loss attribution, black-hole evidence against the actual PLPMTU, immediate smaller-ACK fallback publication, PMTU_RAISE_TIMER rediscovery, active-path PTO attribution, probe loss exclusion from congestion reaction while retaining cwnd/anti-amplification accounting, and rejection of ACKs for unsent packet numbers before ACK reference state is mutated.

Against issue #256 Slice B, the implementation now satisfies the intended scope: conservative 1200 base, operator/peer discovery ceilings, exact standalone PING+PADDING probes, deterministic search/retry behavior, per-path migration/reset semantics, black-hole fallback and upward rediscovery, congestion/anti-amplification enforcement, and the required regressions. Current CI run #1395 is green across unit tests, Linux/macOS/ARM builds, appliance integration, H3/TLS interop, security, packaging, and perf smoke.

All 13 prior inline review threads have been resolved after verifying their fixes.

Only non-blocking cleanup I noticed: the PR description's **Socket precondition** paragraph still says `IP_DONTFRAG` on “macOS/BSD”. The implementation and `HTTP3_ROLLOUT.md` are more precise now: Darwin + FreeBSD are supported explicitly, while the other BSDs take the conservative 1200 fallback. Updating that sentence would keep the PR description aligned with the code, but it is not a merge blocker.

**From the code and Slice B acceptance-criteria standpoint, this head is ready to merge.**

PR #611, opened 2026-08-12
perf(quic): wire real UDP socket buffer tuning (#256-D)

Jaruso:
## Summary

Slice D of #256 — the listener asks for the UDP socket buffers it wants, and reports what the kernel actually gave it.

A full UDP receive buffer drops datagrams the kernel has already taken off the wire. QUIC cannot distinguish that from network loss: congestion control backs off and packets are retransmitted for a reason nothing on the path caused. Until now the listener set `SO_REUSEADDR` and nothing else, so a host whose default buffer was small had no knob at all.

- **`TARDIGRADE_HTTP3_UDP_RECV_BUFFER_BYTES` / `TARDIGRADE_HTTP3_UDP_SEND_BUFFER_BYTES`** request `SO_RCVBUF`/`SO_SNDBUF` on the QUIC socket before `bind`. Both default to `0` — leave the kernel's own sizing alone — which stays the right answer unless drops have been measured. Requests are clamped into `[2048, 1 GiB]` so an operator typo cannot wrap the signed 32-bit `setsockopt` ABI into a negative size.
- **Advisory, never fatal.** A kernel that refuses or trims the request still yields a listener that serves HTTP/3 correctly; failing startup over a performance knob would be worse than running with the default.
- **But never silent.** Linux caps requests at `net.core.rmem_max` while `setsockopt` still returns success, so the size is **read back with `getsockopt`** and classified: `applied`, `clamped` (accepted, cut down by a host ceiling — warns, and names the sysctl to raise), `unsupported` (refused), `unverified` (accepted, unreadable — reported as such rather than claiming a number the process never saw), or `default`. Linux returning roughly double what was set is per-datagram bookkeeping overhead, not payload capacity, and counts as a grant.
- **Effective values are published** on `http3_runtime.Snapshot.udp_buffers` and logged at startup. The `EffectiveBufferTuning` classification lives in `src/quic/udp.zig` as a syscall-free, platform-neutral policy alongside the existing `BufferTuning` seam (which was test-only until now, and whose `tuneBuffers` hook now returns the outcome); `http3_runtime` owns the syscalls, next to `configureNoFragment`.
- **Benchmark metadata** records the host's UDP buffer ceilings (`net.core.rmem_max`/`wmem_max`, or `kern.ipc.maxsockbuf`) and the configured request, since a run whose buffer filled shows up as loss in the results and makes two otherwise-identical runs incomparable.
- Both variables are **restart-required** — buffer sizes are socket state applied to the live descriptor at bind time — and rejected by reload validation like every other listener-owned H3 knob.

Nothing here touches congestion control, flow control, path validation, or anti-amplification: it changes only how much the kernel will hold on this process's behalf.

## Test plan

- [x] `zig build test --summary all` — 3593 passed, 10 skipped (same 10 as on `main`).
- [x] `zig fmt --check build.zig src/ tests/`, `shellcheck benchmarks/competitive/run.sh`.
- [x] Deterministic classification tests in `src/quic/udp.zig`: Linux doubling counted as a grant, exact-match grant, kernel ceiling reported as `clamped` with both numbers, accepted-but-unreadable reported as `unverified` rather than `applied`, outright refusal reported with the default still read back, nothing-requested still read back, and the `c_int` ABI clamp.
- [x] Real-socket tests in `src/http/http3_runtime.zig`: untouched socket reports the kernel default; a satisfiable request is granted and read back; an oversized request is clamped to the ABI bound before reaching the kernel.
- [x] Runtime wiring tests: a listener asking for 1 GiB still binds and bootstraps, with the outcome published; an untuned listener still reports both effective sizes.
- [x] Reload test: changing either target requires a listener restart.
- [x] Live gateway run on macOS — `applied` (8 MiB requested, 8 MiB granted), `clamped` (1 GiB requested → 8388608, warning names `kern.ipc.maxsockbuf`), and the untuned default readback (Darwin's 9216-byte send buffer, which is exactly the kind of thing this makes visible).

Linux-specific behavior (the `rmem_max` clamp and the doubled readback) is covered by the deterministic classification tests rather than by a host that exhibits it; CI runs the socket tests on Linux.

Slice D of #256 — the umbrella issue stays open for the remaining slices (C pacing, E ECN, F GSO/GRO notes, G benchmarks).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,25 @@ All notable user-facing changes to Tardigrade are documented here.
 ## [Unreleased]
 
 ### Added
+- **Tunable UDP socket buffers for the HTTP/3 listener (#256-D)** —
+  `TARDIGRADE_HTTP3_UDP_RECV_BUFFER_BYTES` and
+  `TARDIGRADE_HTTP3_UDP_SEND_BUFFER_BYTES` request `SO_RCVBUF`/`SO_SNDBUF` on
+  the QUIC listener socket. Both default to `0`, leaving the kernel's own
+  sizing alone. A full receive buffer drops datagrams that QUIC then treats as
+  network loss — congestion control backs off for a reason nothing on the path
+  caused — so hosts that measure drops under bursts can now raise it without
+  patching.
+
+  The request is advisory in both directions: a kernel that refuses or trims it
+  still gets a listener that serves HTTP/3 correctly, since buffer sizing is a
+  performance setting and not worth failing startup over. It is not, however,
+  allowed to fail silently. The size is read back with `getsockopt` — Linux
+  caps requests at `net.core.rmem_max` while `setsockopt` still reports
+  success, and returns roughly double what was set — so the listener logs and
+  publishes what the kernel actually granted, distinguishing a grant from a
+  clamp, an outright refusal, and an accepted-but-unreadable result. Effective
+  values appear on the runtime status snapshot, and competitive benchmark runs
+  now record the host's UDP buffer ceilings alongside their results.
 - **QUIC discovers the path MTU instead of assuming one (#256-B)** — Tardigrade
   now runs DPLPMTUD (RFC 8899, RFC 9000 §14.3/§14.4) per network path. Every
   path starts at the 1200-byte size RFC 9000 §14 guarantees and rises only as
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,9 +19,13 @@ All notable user-facing changes to Tardigrade are documented here.
   performance setting and not worth failing startup over. It is not, however,
   allowed to fail silently. The size is read back with `getsockopt` — Linux
   caps requests at `net.core.rmem_max` while `setsockopt` still reports
-  success, and returns roughly double what was set — so the listener logs and
-  publishes what the kernel actually granted, distinguishing a grant from a
-  clamp, an outright refusal, and an accepted-but-unreadable result. Effective
+  success — so the listener logs and publishes what the kernel actually
+  granted, distinguishing a grant from a clamp, an outright refusal, and an
+  accepted-but-unreadable result. Linux stores and reports twice what was set
+  (the other half is bookkeeping, not payload capacity), so that judgement is
+  made on the reading restated in request units: a 256 KiB request capped at
+  208 KiB reads back as 416 KiB, which is larger than the request and still a
+  clamp. Effective
   values appear on the runtime status snapshot, and competitive benchmark runs
   now record the host's UDP buffer ceilings alongside their results.
 - **QUIC discovers the path MTU instead of assuming one (#256-B)** — Tardigrade
--- a/benchmarks/competitive/README.md
+++ b/benchmarks/competitive/README.md
@@ -142,3 +142,13 @@ The combined JSON shape is:
 Only compare rows from the same output directory. Cross-run comparisons are
 valid only when host load, tool versions, server versions, ulimits, loopback
 metadata, build flags, and run parameters match.
+
+`host.network.udp_buffers` records the host-wide ceilings on per-socket UDP
+buffers (`net.core.rmem_max`/`wmem_max` on Linux, `kern.ipc.maxsockbuf` on
+macOS and the BSDs) alongside whatever `TARDIGRADE_HTTP3_UDP_RECV_BUFFER_BYTES`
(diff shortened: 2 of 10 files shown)


Jaruso (commented):
Full review against #256-D complete. I checked all 10 changed files, the production listener/config/reload wiring, the UDP abstraction changes, the real-socket and policy tests, the operator docs, benchmark metadata, and the current CI matrix. CI is green on Linux/macOS and the overall shape is good.

I found **one blocking correctness issue**, with a corresponding test problem: Linux's `SO_RCVBUF` / `SO_SNDBUF` readback is doubled for bookkeeping, but the classifier compares that doubled kernel value directly to the user request. That creates a large false-positive window where a request was actually capped by `rmem_max` / `wmem_max` but Tardigrade reports `applied` instead of `clamped`.

Concrete example on a stock-ish Linux host with `net.core.rmem_max = 212992`:

```text
requested = 262144
kernel user-visible cap = 212992
getsockopt(SO_RCVBUF) = 425984   # doubled Linux accounting value
current check: 425984 >= 262144 => applied   // wrong
```

That undermines the main diagnostic promise of this PR: the warning can disappear exactly when the operator's request was trimmed. It also makes the 256 KiB real-socket test pass for the wrong reason on Linux.

Please keep the platform-neutral policy seam, but normalize Linux's readback semantics before deciding grant vs clamp (while preserving the raw `getsockopt` value for reporting if that is the desired snapshot contract). For example, the effective comparison should be equivalent to:

```zig
const grant_basis = if (builtin.os.tag == .linux)
    got / 2
else
    got;

const status: BufferTuningStatus =
    if (grant_basis >= want) .applied else .clamped;
```

A cleaner boundary would be for `http3_runtime` to translate the platform-specific readback into a `granted_request_bytes` value and pass that into the syscall-free classifier, so `src/quic/udp.zig` remains OS-neutral.

Regression coverage I expect in the same fix:

```zig
// Linux: a true grant. 4 MiB request -> 8 MiB raw readback.
// normalized grant basis == 4 MiB => applied.

// Linux: capped request in the false-positive window.
// e.g. 300 KiB request -> 416 KiB-ish raw readback from a 208 KiB cap.
// normalized grant basis ~= 208 KiB => clamped.
```

Also fix the real-socket test so it does not assume 256 KiB is below every Linux host ceiling; either choose/derive a genuinely satisfiable request from the socket/host state, or make the platform-specific expectation explicit.

I did not find another blocking gap in this pass. Once this classification + regression is fixed, the next review should be a focused verification rather than another broad discovery round.

Jaruso (commented):
Second-pass review complete. The original Linux doubled-readback blocker is fixed correctly in 7f582b4: raw `getsockopt` bytes are preserved for diagnostics, the grant decision uses request-comparable bytes, the old false-positive window has a deterministic regression, and the fixed-size 256 KiB real-socket assumption is gone. I resolved both prior threads. CI is green on Linux/ARM Linux/macOS, including unit tests, format, lint, appliance tests, H3 interop, and perf smoke.

I found one remaining boundary defect and one test assumption caused by the same boundary. Please fix these together; I do not see another functional gap in the config → listener → readback/snapshot → reload → docs/benchmark-metadata wiring after this pass.

Because this account owns the PR GitHub will not allow a formal REQUEST_CHANGES review, so this is submitted as COMMENT, but I would fix the two inline points before merging.

Jaruso (commented):
Round 3 verification complete. I re-checked the full #256-D implementation and the fixes from the prior two rounds against the issue's socket-buffer acceptance criteria.

The two remaining round-2 findings are correctly fixed in `4ed5a8fa`:

- `max_buffer_bytes` is now `INT_MAX / 2`, so the configured upper bound is actually representable under Linux's doubled SO_RCVBUF/SO_SNDBUF storage semantics. The exact-bound regression proves the maximum can classify as `applied` when the host ceiling permits it while preserving the `clamped` path when it does not.
- The real-socket maximum-request test no longer assumes every host must clamp. It now accepts either `applied` or `clamped` and verifies that the verdict follows from the request-comparable readback, so a performance-tuned host cannot fail the correctness suite simply because its ceiling is high.

I also re-checked the unchanged wiring from the earlier passes: env/config -> runtime tuning before bind -> getsockopt readback -> Linux normalization -> snapshot/log publication -> restart-required reload detection -> operator docs/benchmark host metadata. I do not see another correctness, portability, or acceptance-criteria gap in this PR.

No further code changes requested from review. At the time of this review the targeted Linux/macOS unit tests, builds, formatting, lint, security, H3 interop, dependency review, and perf smoke are green; a few broader packaging/appliance jobs are still in progress.

PR #612, opened 2026-08-13
perf(quic): finish ECN end to end (#256-E)

Jaruso:
## Summary

The ECN wire pieces already existed on `main` — `udp.Ecn` named the four IP
codepoints and `frame.zig` parsed/encoded ACK_ECN — but nothing decided
anything with them. No packet was ever marked, no received codepoint was ever
counted, no ACK_ECN was ever generated, and a peer's reported counters went
nowhere. This is the missing half.

- **`src/quic/ecn.zig`** (new) — received counters per packet number space
  (RFC 9000 §13.4.1) and a per-path controller that marks ECT(0), then
  validates the peer's feedback. Pure state machine, no I/O.
- **`src/quic/connection.zig`** — ACK_ECN generation once any marked packet
  arrives in a space; per-packet `ecn_marked` on sent records; validation on
  ACK receipt; CE as an RFC 9002 §7.1 congestion event; a testing-window
  timer; `Transmit.ecn` for the socket layer. `ingestOnPathWithEcn` is the new
  entry point; `ingestOnPath` stays as the honest "this socket cannot see
  codepoints" form, which is `.unavailable` and deliberately *not* `.not_ect`.
- **`src/quic/path.zig`** — one ECN controller per path incarnation, alongside
  the PMTU controller and for the same reason.
- **`src/quic/recovery.zig`** — `onCongestionEvent` split out of
  `onPacketsLost`, since a CE report is congestion about a packet that
  *arrived* and must not reclaim in-flight bytes a second time.
- **`src/http/http3_runtime.zig`** — the real socket half: `IP_RECVTOS` /
  `IPV6_RECVTCLASS` at bind, `recvmsg`/`sendmsg` with ancillary data per
  datagram. Not a socket-wide `IP_TOS`, because ECN is validated per path and
  there would be no way to stop marking on the one path that failed.

### Trust

The peer's ACK_ECN counters are an input to congestion control, reachable by
anything on the path and by the peer itself, so they are never taken at face
value. Marking stops on that path when counters regress, when more marked
arrivals are reported than were ever marked, when ECT(1) appears (never sent
here), when acknowledged marked packets show no matching counter growth, or
when the testing window closes on silence. CE drives congestion response only
on a *validated* path, and the counters this endpoint reports back come only
from packets that passed AEAD — an off-path spoofer cannot inflate them.

Every failure ends in ordinary non-ECN operation, never a connection error.
`ecn_paths_disabled` being non-zero on the open internet is expected.

### Platform gating

Verified per-OS constants only (Linux, Darwin, FreeBSD); anywhere else the
listener runs without ECN rather than guessing — same rule as the
no-fragmentation table from #256-B.

One finding worth flagging: **Darwin aligns control messages to 4 bytes**
(`__DARWIN_ALIGN32`), not the platform word. Assuming otherwise reads the ECN
payload four bytes past where the kernel wrote it, so every received codepoint
comes back `.unavailable` and ECN looks like a network problem on every macOS
host. The state-machine tests all pass with that bug present, which is why
there is a loopback test pinning it against a real kernel.

### Operator surface

`TARDIGRADE_HTTP3_ECN` (default `true`) disables marking outright — the escape
hatch for the one failure mode per-path validation cannot see, a middlebox that
*drops* marked traffic rather than clearing the marks. Restart-required, and
enforced in `http3ListenerConfigChanged`. The runtime snapshot publishes
whether marking is actually running plus marked-sent, validated-path,
disabled-path, and CE counts.

## Test plan

- [x] `zig build test` — 3630 pass, 10 skipped (same 10 as `main`; nothing new
      skips), `zig fmt --check src/` clean
- [x] `ecn.zig`: baseline adoption, validation, regression, over-claim,
      unsent-codepoint, cleared-marks, CE-counts-toward-validation, testing
      timeout, disabled-controller inertness — deterministic, platform-neutral
- [x] `frame.zig`: ACK_ECN encode/decode roundtrip; a plain ACK still reports
      *no* counters rather than zeros
- [x] `connection.zig`: end-to-end validation over a modelled network that
      preserves / clears / rewrites / CE-marks codepoints; congestion response
      on CE; forged over-claim rejected with the window untouched; counters
      only from authenticated packets; `.unavailable` counts nothing; testing
      window closes on silence; handshake-space ACK counters ignored
- [x] `path.zig`: ECN state is per path incarnation and not inherited across
      migration; stale refs resolve nowhere
- [x] `http3_runtime.zig`: platform table, ECN-field extraction, malformed and
      unrelated control messages, and **a real loopback socket** — a marked
      datagram arrives ECT(0), an unmarked one arrives `.not_ect`. Both ran
      (did not skip) on macOS in this run.

Not in scope: batching/GSO/GRO design notes (#256-F) and the H3 benchmark
matrix (#256-G). The snapshot counters this adds are what #256-G will record.

Part of #256

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,32 @@ All notable user-facing changes to Tardigrade are documented here.
 ## [Unreleased]
 
 ### Added
+- **Explicit Congestion Notification for HTTP/3 (#256-E)** — the QUIC listener
+  now marks outbound datagrams ECT(0), counts the codepoints on received
+  packets, reports them to peers in ACK_ECN, and validates the counters peers
+  report back. A congested router can mark a packet instead of dropping it, so
+  congestion control reacts a round trip earlier and without a retransmission.
+
+  ECN is validated per network path and only after the handshake is confirmed;
+  a migration re-validates rather than inheriting an answer measured elsewhere.
+  Because the peer's counters are an input to congestion control and reachable
+  by anything on the path, they are checked against what was actually sent —
+  counters that regress, exceed what was marked, report a codepoint this
+  endpoint never sends, or fail to grow when marked packets are acknowledged
+  all turn marking off for that path. CE reports drive congestion response only
+  on a validated path, and the counters reported back come only from packets
+  that passed AEAD authentication, so an off-path spoofer cannot inflate them.
+  Every failure ends in ordinary non-ECN operation, never a connection error —
+  which is the common outcome on the open internet, where a great deal of
+  deployed gear clears or rewrites the field.
+
+  Platform support is gated on verified per-OS socket constants (Linux, Darwin,
+  FreeBSD); anywhere else the listener simply runs without ECN.
+  `TARDIGRADE_HTTP3_ECN` (default `true`) disables it outright, for hosts where
+  a middlebox drops marked traffic rather than clearing the marks — the one
+  failure mode per-path validation cannot see. The runtime status snapshot
+  publishes whether marking is actually running along with marked-sent,
+  validated-path, disabled-path, and CE counts. See `docs/HTTP3_ROLLOUT.md`.
 - **Tunable UDP socket buffers for the HTTP/3 listener (#256-D)** —
   `TARDIGRADE_HTTP3_UDP_RECV_BUFFER_BYTES` and
   `TARDIGRADE_HTTP3_UDP_SEND_BUFFER_BYTES` request `SO_RCVBUF`/`SO_SNDBUF` on
--- a/docs/CONCURRENCY.md
+++ b/docs/CONCURRENCY.md
@@ -168,6 +168,7 @@ they need public control.
 | Retry policy | address validation / tokens | `off` | `TARDIGRADE_HTTP3_RETRY_POLICY=off\|address_validation`; `address_validation` sends stateless QUIC Retry before allocating connection state |
 | Migration policy | `disable_active_migration` | NAT rebinding only | `TARDIGRADE_HTTP3_CONNECTION_MIGRATION=false` permits same-IP NAT rebinding but not active migration; `true` enables full migration |
 | UDP socket buffers | `SO_RCVBUF` / `SO_SNDBUF` | kernel default | `TARDIGRADE_HTTP3_UDP_RECV_BUFFER_BYTES` / `..._SEND_BUFFER_BYTES`; advisory, read back with `getsockopt` and published on the runtime snapshot |
+| ECN | ECT(0) marking + ACK_ECN | on where the kernel supports it | `TARDIGRADE_HTTP3_ECN=false` disables; validated per path and turned off for that path on inconsistent peer feedback |
 | qlog / keylog | debug artifact emission | off | local-debug toggle |
 | QPACK | SETTINGS | static-only, zero dynamic capacity | internal until dynamic QPACK lands |
 
--- a/docs/HTTP3_ROLLOUT.md
+++ b/docs/HTTP3_ROLLOUT.md
@@ -278,6 +278,14 @@ path it has been reached on:
   accounted for. Adopting the first report that happens to arrive would write
   off whatever was already in flight, and the next honest report covering
   those packets would look like an over-claim.
+
+  The counters that baseline is taken from advance only on reports that
+  *survived* validation. A rejected report is not a synchronisation point: a
+  count that regressed to 8 would otherwise become the baseline a later path is
+  measured from, and the real cumulative 10 arriving afterwards would credit
+  that path with growth it never earned. Counters that regress or over-claim
+  fail ECN closed for the connection; a stripped or rewritten codepoint
+  condemns only the route, and leaves the counters usable for the next path.
 - **A migration waits for the previous path's marks to drain** before the new
   one starts testing. Otherwise the old path's intact marks would validate a
   new path that is stripping every codepoint, and the old path's CE reports
--- a/docs/HTTP3_ROLLOUT.md
+++ b/docs/HTTP3_ROLLOUT.md
@@ -284,6 +284,12 @@ path it has been reached on:
   would halve the new path's window for congestion that is no longer on the
   route. Promotion additionally requires an acknowledgement of a packet the
   new path itself marked.
+
+  Only an *acknowledgement* drains that wait. A loss declaration does not:
+  QUIC loss is an inference, and the packet may well have arrived and been
+  counted while the ACK reporting it was itself lost. Since a packet that is
+  never acknowledged is never settled, the wait is bounded — and running out
+  turns ECN off for the connection rather than releasing on an assumption.
 - **An ACK that does not advance the largest acknowledged packet number is
   ignored entirely** (RFC 9000 §13.4.2.1). Cumulative counters arrive
   legitimately stale on a reordered ACK, and validating against them would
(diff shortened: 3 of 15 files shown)


Jaruso (commented):
Blocking review findings — changes are needed before I would consider #256-E complete.

I reviewed the current head (`ce68902b`) end-to-end against #256-E, RFC 9000 §13.4 / migration semantics, RFC 9002 congestion response, every changed file, and the current CI run. The follow-up commit correctly fixed the exhaustive `connection.Event` consumer; the previously failing TLS/H3 interop jobs are now green.

I found five remaining ECN correctness gaps, all called out inline. The socket/ancillary-data side is in good shape; the blockers are in ACK_ECN validation bookkeeping and path attribution.

The implementation needs to preserve three invariants:

1. ACK_ECN counts are cumulative per packet-number space, so reordered ACKs must not move validation state backward or fail a path.
2. A new path’s validation must only be established by evidence attributable to that path; old-path ECT/CE growth cannot validate or congest the new path.
3. ECN send metadata needed for validation has to outlive recovery’s declaration of loss, because a packet can still be acknowledged later.

A reasonable shape would be to keep connection-level application-space ECN feedback state (`last_successful_counts`, `largest_ecn_acked`), snapshot that state when a path begins testing, and use a path/epoch barrier so old-path marked packets cannot contribute to a new path’s validation. For example, the reordered-ACK gate should look roughly like:

```zig
const advances_ecn_largest = self.largest_ecn_acked[space_idx] == null or
    ack.largest_acknowledged > self.largest_ecn_acked[space_idx].?;

if (advances_ecn_largest) {
    self.processAckEcn(space, ack, ecn_acks, largest_acked_sent_us, now_us);
    self.largest_ecn_acked[space_idx] = ack.largest_acknowledged;
}
```

The exact data model can differ, but please add deterministic regressions for each inline scenario (reordered ACK, batched first feedback, idle-before-first-mark, late ACK after loss, and old/new-path mixed ECN feedback).

Note: GitHub does not permit the authenticated author account to submit `REQUEST_CHANGES` on its own PR, so this is filed as a COMMENT review; the findings above are nevertheless blocking.

Jaruso (commented):
Second full review against the current head and #256-E acceptance criteria. The previous baseline capture, reordered-counter regression, first-mark timer, retained late-ACK metadata, and interop switch issues were materially addressed. CI is green on this head.

I still do **not** think this is ready to merge. The new epoch/history design has several cases where information is discarded or treated as resolved before the peer's cumulative ECN counters are actually safe to attribute. Because those counters can drive both path validation and congestion response, the conservative direction has to be to stop/refrain from ECN when attribution is no longer provable—not release the barrier and continue testing.

I found 6 blocking issues below: two P1 attribution/validation holes and four P2 revalidation/fallback problems. The most important regression is that recovery declaring an old-path ECT packet lost is being treated as proof it cannot affect the peer's counters, even though QUIC loss is an inference and the implementation itself supports late ACKs of such packets.

Jaruso (commented):
Third review of current head `c84f6ab8`. No commits landed since the prior review and CI is green, so the six unresolved findings from that review remain blocking and I have not duplicated them. I found one additional P1 ECN trust-boundary issue inline: rejected peer counters are committed as the baseline for future path validation before validation succeeds. The socket integration and surrounding wiring still look sound, but the ECN validation/accounting needs another coherent pass before merge. GitHub does not permit the PR author to request changes on their own PR, so this is a blocking COMMENT review.

Jaruso (commented):
Fourth full review of the current head (`32d321e`) against #256-E and the prior review threads. The seven previous blockers were materially addressed: loss/eviction no longer silently release ECN evidence, stale ACK bookkeeping is separated from validation, ECT(1) is baseline-relative, pure ACKs are unmarked, socket rejection reaches transport state, and rejected counter reports no longer automatically become future baselines. Current CI run 31732916255 is fully green, including Linux/macOS/ARM unit/build jobs and TLS/H3 interop.

I still found three blocking edge cases in the new bookkeeping: two P1 cross-epoch synchronization/attribution holes and one P2 runtime capability-propagation hole. The common transport invariant is that a new path cannot start an ECN epoch until every counter obligation from the old epoch has either been checked against a trustworthy current ACK_ECN report or caused ECN to fail closed. Retiring packet metadata is not the same thing as obtaining that synchronization point.

The socket ancillary-data implementation, authenticated receive counting, ACK_ECN generation, CE congestion integration, config/operator wiring, and the conservative fail-closed direction otherwise look good.

As before, GitHub does not allow the authenticated author account to submit `REQUEST_CHANGES` on its own PR, so this is a COMMENT review; the three findings below are blocking.

Jaruso (commented):
Full re-review of `f72623d` against the remaining #256-E acceptance criteria and the three findings from the previous pass.

The ECN epoch synchronization work is now materially correct: plain-ACK and non-advancing-ACK obligations keep the migration barrier closed, carried evidence stays tied to its originating generation, and a later current ACK_ECN is required before the next path can baseline. The listener capability-loss state is also now durable across live connections rather than a one-pass edge flag. CI is fully green on this head across Linux/macOS/ARM tests/builds, interop, format, security, and perf smoke.

I found **one remaining functional blocker** in the unsupported-platform fallback, plus one documentation inconsistency. The functional issue is narrower than the prior runtime finding but real: the transport increments `ecn_marked_sent` before the socket result is known. The send that *discovers* the cmsg rejection is retried Not-ECT, and a connection that had not yet observed the global capability loss can likewise build one marked transmit that the runtime silently downgrades. Disabling the connection after `sendDatagram()` stops future marks, but it cannot undo the already-recorded metric. Because the PR defines `ecn_marked_sent`/the listener snapshot as packets/datagrams that actually went out carrying ECT(0), the fallback still leaves observable state that claims wire marking which never happened.

Once that is reconciled (and the stale non-advancing-ACK wording is corrected), I do not see another #256-E correctness blocker on this head.

Jaruso (commented):
Final full re-review of `3585f7c` against the #256-E acceptance criteria. **No blocking findings remain.**

I rechecked the full ECN path, not just the last delta:

- real recvmsg ancillary metadata feeds authenticated received ECT(0)/ECT(1)/CE accounting;
- sendmsg applies per-datagram ECT marking with platform gating and clean Not-ECT fallback;
- received counters drive ACK_ECN generation;
- peer feedback is validated against trusted cumulative state before it can advance the baseline;
- migration uses generation-tagged history plus `ecn_sync_owed`, so old-path growth cannot validate or congest a new path;
- loss declarations do not release ECN attribution evidence, and bounded evidence loss fails ECN closed;
- reordered/non-advancing ACKs retire packet metadata without validating stale cumulative counters;
- ECT(1) rejection is relative to the path baseline;
- pure ACK-only traffic is not used as marked validation evidence;
- CE reaches congestion control only after the active path has validated;
- socket capability loss disables future accepts and live transport ECN without failing QUIC;
- deterministic/state-machine and platform-gated socket tests cover the required valid/invalid/missing/CE/unsupported cases.

During this pass I committed `3585f7c` to correct the stale rollout wording around non-advancing ACKs and connection-wide fail-closed reasons. I also re-evaluated my prior `ecn_marked_sent` concern and withdrew it as a blocker: at the sans-I/O transport boundary it is a marking-instruction counter, while end-to-end preservation is proven by validated ACK_ECN feedback. Treating successful `sendmsg()` as proof of an on-wire ECN bit would provide false precision; canonical exported metric naming belongs with #255.

CI run `31750566904` is fully green across Linux/macOS/ARM unit tests, Debug/ReleaseSafe/ReleaseFast builds, format/lint/security/dependency audit, TLS/H3 interop, packaging, PKI differential, and perf smoke.

**Merge-ready from this review.**

PR #613, opened 2026-08-13
docs: document reload, drain, and shutdown behavior

### LINKED ISSUE #165: ux: document reload, drain, and shutdown behavior
## Summary
Document how Tardigrade handles config reloads, draining, graceful shutdown, active connections, worker jobs, upstream pools, and error cases.

## Motivation
Reload and shutdown behavior is operationally critical. Operators need to know what happens to in-flight requests and long-lived connections before they trust Tardigrade in production.

## Scope
- Explain supported reload mechanism(s).
- Document what happens to active client connections.
- Document what happens to upstream keepalive pools.
- Document drain timeout behavior and defaults.
- Document what happens when a new config fails validation during reload.
- Include example commands and expected logs.

## Acceptance criteria
- Docs clearly distinguish reload, graceful shutdown, and hard termination.
- Docs specify whether in-flight requests are allowed to complete.
- Docs specify timeout/default behavior.
- Docs link to config reference fields related to reload/drain.

## Notes
This is documentation-first, but may uncover runtime behavior that should be clarified in code or tests.

Jaruso:
## Summary
- add a dedicated reload, drain, and shutdown lifecycle doc
- document reload failure behavior, in-flight request handling, worker drain semantics, upstream pools, commands, logs, and metrics
- link the new doc from the README and graceful reload example, and correct the timeout matrix reload wording

## Validation
- test -f doc/link targets
- git diff --check

Closes #165

### DIFF
--- a/README.md
+++ b/README.md
@@ -217,6 +217,7 @@ operate without guessing which config file or pid file is active.
 | Concurrency & hot-path audit | [docs/CONCURRENCY.md](docs/CONCURRENCY.md) |
 | Event-loop backend evaluation (epoll/kqueue vs libxev/std.Io/io_uring) | [docs/EVENT_LOOP_BACKENDS.md](docs/EVENT_LOOP_BACKENDS.md) |
 | Observability | [docs/OBSERVABILITY.md](docs/OBSERVABILITY.md) |
+| Reload, drain, and shutdown | [docs/RELOAD_SHUTDOWN.md](docs/RELOAD_SHUTDOWN.md) |
 | Proxy security | [docs/PROXY_SECURITY.md](docs/PROXY_SECURITY.md) |
 | Security test plan | [docs/SECURITY_TEST_PLAN.md](docs/SECURITY_TEST_PLAN.md) |
 | TLS interop & conformance matrix | [docs/TLS_INTEROP_MATRIX.md](docs/TLS_INTEROP_MATRIX.md) |
--- a/docs/OBSERVABILITY.md
+++ b/docs/OBSERVABILITY.md
@@ -473,33 +473,33 @@ dropping connections:
 1. **Load** the new configuration from the environment / config file.
 2. **Validate** it. Invalid configuration is rejected here.
 3. **Install** the new config version through a lease-counted store
-   (`ReloadableConfigStore`): in-flight requests keep the config version they
-   started with, and the old version is retired only after its last lease is
-   released. New requests pick up the new version once installation completes.
+   (`ReloadableConfigStore`): in-flight requests keep their request-scoped config
+   lease, and the old version is retired only after its last lease is released.
+   New requests pick up the new version once installation completes.
 
 Guarantees:
 
-- **A failed reload never replaces the active config.** If load or validation
-  fails, the previous configuration stays active and continues serving traffic;
-  the failure is recorded and logged.
-- **In-flight requests are not disrupted** by a reload; they finish on the
-  config they began with.
+- **A failed reload before config publication keeps the previous published
+  config active** for request routing and config leases; the failure is recorded
+  and logged. Reload-owned runtime resources are not fully transactional.
+- **In-flight requests are not dropped** by a reload; they keep their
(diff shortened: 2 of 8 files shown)


Jaruso (commented):
Changes are required before merge. I traced the new operator contract against the current reload, worker-pool, parked-connection, HTTP/3, upstream-pool, CLI, and systemd behavior and found several correctness/acceptance gaps.

The largest issues are: the document promises fully atomic reload failure behavior even though some runtime state is mutated before config publication; graceful shutdown is described as a worker-pool-only soft deadline even though native HTTP/3 has a separate deadline that closes remaining H3 connections; most upstream-pool policy knobs listed here are not actually applied by hot reload; and the runnable example still contains contradictory/stale drain semantics.

Issue #165's required topic coverage is otherwise present, and CI for `875f628` is green. The inline comments below include the exact behavior I expect the docs to describe (or, where appropriate, the runtime behavior that would need to change if the stronger documented contract is intended).

Jaruso (commented):
Second full review of `564e8f4` against issue #165 and the current runtime. The previous eight review findings are substantially addressed, and CI is green, but I found additional lifecycle-contract gaps in the fresh pass that should be fixed before merge.

The remaining blockers are not stylistic: (1) a successful reload does not give in-flight requests a completely isolated old-config snapshot because process-shared runtime state is updated while old config leases can still be active; (2) the zero-timeout worker-pool wording incorrectly says all queued work is closed even though only queue-owned accepted sockets are closed; (3) the lifecycle docs still conflict on TCP's soft drain versus native H3's hard connection deadline; (4) the runnable config example still says the `tardi stop` command waits even though it only signals and returns; and (5) #165 explicitly requires a link to the related config reference, while the lifecycle page currently only enumerates the knobs.

One additional consistency fix: the graceful-reload README still says `TARDIGRADE_PID_FILE` is *required* for `tardi reload` / `tardi stop`, while this PR's canonical doc correctly documents `--pid <pid>` as an alternative. Please change that row (and the same absolute wording in `tardigrade.env.example`) to `required unless --pid is supplied` or equivalent.

I rechecked the prior findings for reload failure atomicity, parked keep-alive teardown, H3 draining, request-deadline timing, upstream pool reload policy, stop-command semantics, env drain wording, and systemd timeout ordering; those fixes look correct apart from the cross-document H3/soft-cap issue called out inline.

Jaruso (commented):
The five blockers from my previous review are addressed at this head. I re-traced the shutdown path and observability plumbing against the updated docs and found two remaining correctness gaps before #165 is complete.

The important one is that the worker-pool/H3 drain deadline still comes from the original startup `cfg`, while reload can publish a different `shutdown_drain_timeout_ms` and request lifecycle code reads that reloaded value. Worker thread/queue sizing is similarly startup-owned but not rejected on reload. That means a successful reload can expose config values that the actual shutdown path does not use consistently.

The second gap is observability: `drain_timed_out` / `tardigrade_drain_timeouts_total` report only the TCP worker-pool `DrainResult`; they do not report native H3 reaching its hard deadline, and forced-close counting also includes the zero-timeout immediate-close path. The new H3 distinction makes the current generic metric descriptions misleading.

Everything else from the prior rounds now matches the runtime, the config-reference links are present, and the issue's requested topic coverage is otherwise complete. Inline comments below contain the exact distinctions that still need to be documented or enforced in code.

Jaruso (commented):
Final re-review of the current head: no remaining merge blockers against issue #165. The two findings from the previous round are addressed: the docs now explicitly distinguish the startup-owned final drain/H3 deadline from reloadable request-scoped configuration, and drain metrics/logs are correctly scoped to the TCP worker-pool outcome rather than presented as global H3+TCP results. I also rechecked the earlier reload atomicity caveat, parked keepalive behavior, HTTP/3 drain semantics, request deadline timing, upstream-pool restart requirements, CLI stop behavior, zero-timeout queue ownership, systemd headroom, config-reference links, and example consistency. CI for 057ed31 is green. Ready to merge from a review standpoint.

PR #614, opened 2026-08-14
quic: expose transport and H3 Prometheus metrics

Jaruso:
## Summary

Implements slice 5 of #255 by adding the low-cardinality Prometheus surface for native QUIC/H3 observability.

- adds QUIC transport counters/gauges for active connections, handshake failures, Retry, anti-amplification blocking, PTO, loss, datagram payload bytes, stream resets, flow-control blocking, and deprotection failures
- adds native H3 request totals and latency histogram plus renderer coverage
- wires HTTP/3 runtime delta folding into GatewayState metrics callbacks so repeated polling does not re-add cumulative per-connection counters

## Tests

- `zig build test -- --test-filter "QUIC and H3 transport metrics"`
- `zig build test`

Refs #255

### DIFF
--- a/docs/QUIC_QLOG.md
+++ b/docs/QUIC_QLOG.md
@@ -261,27 +261,27 @@ destinations (qlog directory, keylog path) are supplied by the composition root
 that owns the writers, not by the transport config, keeping file I/O out of the
 transport core.
 
-## Metrics (Prometheus) — planned surface
+## Metrics (Prometheus)
 
 qlog answers "what happened on this one connection"; Prometheus answers "what is
 happening across all connections". The existing per-module `Metrics` counters
 are the source. The gateway `/status/metrics` endpoint (see
-`docs/OBSERVABILITY.md`) will export, when the pure-Zig backend is active:
+`docs/OBSERVABILITY.md`) exports, when the pure-Zig backend is active:
 
 - `tardigrade_quic_connections_active` (gauge)
-- `tardigrade_quic_handshake_failures_total{stage}`
+- `tardigrade_quic_handshake_failures_total{stage}` (`initial|handshake`)
 - `tardigrade_quic_retry_total`, `tardigrade_quic_amplification_blocked_total`
 - `tardigrade_quic_pto_total`, `tardigrade_quic_packets_lost_total`
 - `tardigrade_quic_bytes_sent_total`, `tardigrade_quic_bytes_received_total`
 - `tardigrade_quic_stream_resets_total`
-- `tardigrade_quic_flow_control_blocked_total`
+- `tardigrade_quic_flow_control_blocked_total{scope}` (`connection|stream`)
 - `tardigrade_quic_deprotection_failures_total`
 - `tardigrade_h3_qpack_blocked_streams` (gauge)
 - `tardigrade_h3_requests_total` and an h3 latency histogram
 
 Labels are kept low-cardinality (e.g. `stage`, not per-client) per the issue's
-non-goals. Wiring these into the gateway registry is follow-up work; the
-counters they read from already exist.
+non-goals. QUIC byte counters are accounted at the UDP datagram payload
+boundary so coalesced packets are not double-counted.
 
 Production caveat: the current native H3 connection path still uses the static
 decoder and does not compose `DynamicDecoder.decodeOrBlock()` into request
--- a/src/edge_gateway.zig
+++ b/src/edge_gateway.zig
@@ -525,6 +525,14 @@ pub fn run(cfg: *const edge_config.EdgeConfig) !void {
             .quic_early_data_decision_metrics_cb = recordQuicEarlyDataDecisionFromRuntime,
             .quic_zero_rtt_packet_metrics_ctx = &state,
             .quic_zero_rtt_packet_metrics_cb = recordQuicZeroRttPacketFromRuntime,
+            .quic_transport_metrics_ctx = &state,
+            .quic_transport_metrics_cb = recordQuicTransportDeltaFromRuntime,
+            .quic_connections_active_metrics_ctx = &state,
+            .quic_connections_active_metrics_cb = setQuicConnectionsActiveFromRuntime,
+            .quic_handshake_failure_metrics_ctx = &state,
+            .quic_handshake_failure_metrics_cb = recordQuicHandshakeFailureFromRuntime,
+            .h3_request_latency_metrics_ctx = &state,
+            .h3_request_latency_metrics_cb = recordH3RequestLatencyFromRuntime,
         }) catch |err| blk: {
             state.logger.warn(null, "HTTP/3 listener failed to initialize: {s}", .{@errorName(err)});
             break :blk null;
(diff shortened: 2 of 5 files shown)


Jaruso (commented):
**Blocking review — changes required before merge.** GitHub does not allow this account to submit `REQUEST_CHANGES` on its own PR, so this is posted as a comment review; the findings below are still merge blockers.

I reviewed all four changed files against the Slice 5 contract in #255, including the latest-main reconciliation/closeout criteria, and checked the current CI run (green).

The Prometheus schema/renderer is appropriately low-cardinality and the GatewayState/edge wiring is straightforward, but there are four runtime correctness issues that need to be fixed before this slice meets its acceptance criteria:

1. Runtime shutdown/loop exit destroys live connections without going through `removeConnection()`, leaving the active gauge stale and dropping final cumulative transport/stream/TLS deltas.
2. Parked 0-RTT request latency is restarted at resume time (after the continuation runs), so the histogram excludes the parked interval and resumed handler work.
3. Non-parked handler failures that terminate with a 500 are not counted/timed at all, while other error terminal paths are, so `tardigrade_h3_requests_total` is not an exact request-lifecycle count.
4. `streams_blocked_events` (MAX_STREAMS / stream-count exhaustion) is being folded into the `{scope="stream"}` byte flow-control series, conflating two distinct QUIC limits.

The current added test only exercises `Metrics` recording/rendering directly. #255 Slice 5 explicitly calls for runtime-level tests proving teardown returns the active gauge to zero, final deltas are flushed, repeated folds do not double-count, the actual loss/PTO/Retry/amplification/reset/block/deprotection paths hit the intended series, and H3 requests are counted/timed exactly once. Please add those regression tests along with the fixes below; they would have caught the lifecycle issues in this review.

CI is currently green across the build/unit/platform/security/interop jobs, so these are semantic/runtime gaps rather than compile/test failures.

Jaruso (commented):
Blocking follow-up review on `8ab2307`.

The four findings from my previous pass are addressed: bulk teardown now flushes through the removal path, non-parked handler failures are counted, parked retries preserve the original start timestamp, and `MAX_STREAMS` is no longer folded into stream byte-flow-control pressure. CI is green.

I found three remaining semantic bugs plus one explicit Slice 5 acceptance-test gap in the current implementation; see the inline comments. The important distinction is that these metrics have to describe what actually happened, not merely expose internally consistent counters:

1. handshake failures are currently classified from `handshake_error` presence rather than actual handshake progress, and administrative teardown can manufacture failures;
2. response-send/fallback-send failures terminally finish/reset requests but still bypass H3 request/latency accounting;
3. Retry is incremented even when the UDP send failed, despite the series being documented as packets sent;
4. the #255 Slice 5 test checklist explicitly requires `/metrics` coverage for repeated-fold/no-double-counting and the real loss/PTO/Retry/amplification/reset/block/deprotection source paths; the current renderer test injects a synthetic `QuicTransportDelta` and therefore cannot catch wiring mistakes in those paths.

One small closeout item as part of this slice: `docs/QUIC_QLOG.md` still calls this the **“planned surface”** and says wiring the counters into the gateway registry is follow-up work. That becomes false as soon as this PR merges; update that paragraph to describe the now-live Prometheus surface.

I would not merge until the three semantic issues and the Slice 5 runtime/served-metrics coverage are fixed. GitHub does not permit this connected account to submit `REQUEST_CHANGES` on its own PR, so this is recorded as a blocking COMMENT review.

Jaruso (commented):
**Blocking follow-up review — changes still required before merge.** I re-reviewed the full current PR against #255 Slice 5, not just the latest fix commit. The prior-round fixes for Retry send accounting, response-send failure accounting, delta folding/no-double-counting coverage, bounded flow-control labels, teardown flushing, and the documentation update are now in place, and the current CI run is fully green.

Two runtime semantic blockers remain, both in `http3_runtime.zig`, plus a regression-test hole that currently lets the handshake bug pass:

1. Handshake failure accounting still infers failure from the connection's *current* state. `Connection.isEstablished()` is only `state_ == .established`; a connection that successfully completed the handshake later becomes `.closing`/`.draining`/`.closed`, so normal post-handshake teardown can be counted as an Initial/Handshake failure. The same inference also turns drain-deadline closes into handshake failures. Track historical handshake success and administrative-close provenance instead.
2. A decoded `IncomingRequest` rejected by the graceful-drain boundary is reset and `finishRequest()`ed but never reaches `noteRequestCompletedWithLatency()`, so `tardigrade_h3_requests_total` / the histogram still violate the Slice 5 “counted/timed once” lifecycle requirement.
3. The new handshake regression does not assert stage values and reads an uninitialized `quic_observer` for the Initial case, so it cannot protect the metric semantics.

After these are corrected, I do not see another outstanding Slice 5 gap in the current five-file diff. The newly added source-path/delta-rendering test substantially addresses the runtime acceptance matrix from my previous review.

Jaruso (commented):
**Blocking final-pass review — two handshake-failure provenance issues remain.** The prior H3 request accounting, parked latency, Retry send accounting, teardown/final-delta flush, active gauge, flow-control scope, and repeated-fold acceptance coverage all look addressed on this head. The remaining problems are both in the new historical handshake/removal bookkeeping: `handshake_succeeded` is marked before QUIC establishment is actually accepted, and the drain-deadline path can overwrite a pre-existing genuine failure as administrative. Both can make `tardigrade_quic_handshake_failures_total{stage}` silently undercount real failures. Latest CI is still running on this head.

Jaruso (commented):
**Blocking follow-up review — one handshake lifecycle edge remains before merge.** GitHub will not allow this account to submit `REQUEST_CHANGES` on its own PR, so this is a comment review.

I re-reviewed the full current PR against #255 Slice 5 and the latest head (`ec532e2`), including the two findings from the previous pass. Both direct fixes are correct: QUIC establishment is now tracked from the authoritative `.state = .established` transition, and an already-closing protocol failure is no longer relabeled administrative by the drain helper. The previous request-counting, Retry, delta-folding, flow-control, teardown, histogram, bounded-label, and documentation findings remain addressed.

There is one remaining ordering bug inline: on a timer pass where a half-open connection has already crossed `handshake_timeout_us` **and** the graceful-drain deadline is expired, `closeForDrainDeadline()` runs before the runtime computes `stalled`. It marks the connection administrative, then the same pass observes it as timed out, but removal gives `administrative_close` precedence and suppresses the genuine handshake-timeout metric. Preserve whether the handshake deadline had already expired before initiating an administrative close, and add the corresponding regression.

The current CI run is still in progress, but every completed build/unit/security/interop/platform job I checked is green. I found no other Slice 5 correctness gap in this pass.

Jaruso (commented):
**Final follow-up review — no further changes requested.** I re-reviewed the full current PR at `f20432baaf7e9ea4936041e4b2f54eb473cb0060` against #255 Slice 5 after the latest-main merge/conflict resolution, rather than checking only the last patch.

The remaining timeout-vs-drain blocker is fixed correctly: the runtime now computes handshake-timeout provenance before an expired graceful drain can mutate the connection, carries the `RemovalReason` with the reap entry, and preserves an earlier administrative close without later reclassifying it as a timeout. The added regression covers the same-pass timeout/drain case.

I also rechecked the earlier review areas and they remain fixed: historical handshake success is tied to the QUIC `.established` transition; an already-failing connection is not relabeled administrative by the drain deadline; Initial/Handshake stage attribution is bounded and tested; teardown flushes final cumulative deltas and returns the active gauge to zero; Retry is counted only after a successful datagram send; stream flow-control excludes MAX_STREAMS pressure; repeated folds do not double-count; the transport source-path matrix is covered; terminal H3 response failures and drain-boundary rejections are counted exactly once; parked-request latency retains its original start time; byte accounting is at the UDP datagram boundary; and the docs describe the live surface and QPACK caveat accurately.

The merge from current `main` also looks clean in the conflict area: #615's `connectionWakeUs`/fine-grained `SocketWaiter` pacing integration remains intact alongside this PR's reap/removal logic.

I found no additional correctness or acceptance-criteria gaps. All CI jobs that have completed on this head are green; the workflow still has a small number of jobs finishing at the time of this review. From the code-review side, this is ready to merge once CI completes successfully.

PR #615, opened 2026-08-14
perf(quic): pace packets on an RFC 9002 schedule (#256-C)

Jaruso:
## Summary

A congestion window bounds how much data may be *outstanding*, not how fast it may leave. Until now the send path wrote until the window was full, so a connection with plenty of nominal capacity still handed its whole window to the first router in the path as one burst — which is how queue-overflow loss shows up on a link that had the capacity.

- **`recovery.Pacer`** — an RFC 9002 §7.7 leaky bucket replacing the coarse `pacingHint()`. Credit accrues at `N × congestion_window / smoothed_rtt`, with `N` of 2 in slow start and 1.25 in congestion avoidance. The ceiling is ten maximum-size datagrams, which bounds the opening flight *and* the flight a connection returning from idle may send. BBR and other rate estimators stay out of scope.
- **Generation and scheduling stay separable.** Nothing sleeps inside packet construction. `pollTransmitOnPath` declines to build paced data while the bucket is short; `Connection.nextSendTimeUs` reports when the schedule next releases something.
- **`http3_runtime` folds that release into its existing `wake_us`**, via a new `connectionWakeUs` seam split out of `serve` so the selection is testable without a socket. A paced connection neither spins nor sleeps past its own deadline, and one `drainConnectionTransmits` pass can no longer flush a cwnd-sized burst.
- **Responsiveness preserved.** ACKs, PTO probes, and Initial/Handshake flights are charged to the bucket but never delayed by it — an ACK is not congestion-controlled traffic, a probe is how a stalled connection recovers, and the handshake is already bounded by the initial window and anti-amplification.
- **Congestion control and anti-amplification stay the hard gates.** Pacing only ever delays; a connection blocked by the window reports no pacing deadline at all, so the loss/PTO timers keep owning that wakeup.
- **One pacing API.** `udp.SendDatagram.send_at_us` is documented as the same contract rather than a parallel one — it is null on the production path because the transport releases datagrams before they reach an endpoint.
- Docs: a `Packet pacing` section in `docs/HTTP3_ROLLOUT.md`, plus CHANGELOG and `docs/CONCURRENCY.md` seam updates.

## Test plan

All new tests drive a fake clock — no wall-clock sleeps.

- [x] `recovery.zig`: burst bound (exactly ten datagrams leave at one frozen instant, then the bucket is empty); spacing (release is exact at 8000 µs, refused at 7999, and steady across successive sends); idle restart (ten seconds of accrued credit collapses to one burst, not a backlog); cwnd as the harder gate (full bucket over a full window sends nothing); slow-start vs congestion-avoidance gain; only in-flight packets metered; a shrinking window halves the rate; migration discards the schedule with the window.
- [x] `connection.zig`: a frozen-clock drain is bounded by the burst and then returns nothing (no spin); `nextSendTimeUs` is strictly future, produces nothing at `release - 1`, and produces at `release`; steady spacing past the opening burst; idle restart bounded; a closed window adds no in-flight bytes and reports no pacing deadline; an ACK is not delayed by a spent bucket.
- [x] `http3_runtime.zig`: the loop's sleep deadline equals the pacing release when that beats every timer, and a cwnd-blocked connection contributes no wakeup of its own.
- [x] `zig build test` — full suite green, `zig fmt --check src/` clean.

One pre-existing note: `gateway_proxy.zig`'s Unix-socket test aborts under a socket-blocking sandbox (`setsockopt(SO_RCVTIMEO)` → `EINVAL`). Unrelated to this change and passes when the sandbox is off; it only became visible here because touching `http3_runtime.zig` invalidated that step's build cache.

Refs #256

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,20 +19,28 @@ All notable user-facing changes to Tardigrade are documented here.
   datagrams, which also caps what a connection returning from idle may send at
   once. BBR and other rate estimators remain out of scope.
 
-  Only application data waits on the bucket. Acknowledgements, PTO probes, and
-  Initial/Handshake flights are charged to it but never delayed by it: an ACK
-  is not congestion-controlled traffic, a probe is how a stalled connection
-  recovers, and the handshake is already bounded by the initial window and by
-  anti-amplification. Congestion control and anti-amplification stay the hard
-  gates — pacing only ever delays a datagram, never authorises one.
+  Only application data waits on the bucket. PTO probes, Initial/Handshake
+  flights, DPLPMTUD probes, and path-validation probes are charged to it but
+  never delayed by it — a probe is how a stalled connection recovers, and the
+  handshake is already bounded by the initial window and by anti-amplification.
+  Pure ACKs are exempt outright, neither delayed nor charged, since an ACK-only
+  packet is not congestion-controlled traffic and metering it would only add
+  latency to the peer's loss recovery. Congestion control and anti-amplification
+  stay the hard gates — pacing only ever delays a datagram, never authorises one.
 
   Packet construction never sleeps: `pollTransmitOnPath` declines to build
   paced data while the bucket is short, and the listener folds the connection's
-  next release time into the same poll timeout it already computes from timers,
-  so a paced connection neither spins nor sleeps past its own deadline. There
-  is no operator knob — a pacing rate adjustable independently of the window
-  would be a way to defeat congestion control rather than to tune it. See
-  `docs/HTTP3_ROLLOUT.md`.
+  next release time into the same sleep it already computes from timers, so a
+  paced connection neither spins nor sleeps past its own deadline. Because
+  pacing intervals are routinely sub-millisecond — a 480 kB window over a 10 ms
+  RTT is one datagram every ~20 µs — the QUIC listener now waits with `ppoll`
+  on Linux and a `kqueue` timeout on the BSDs, falling back to millisecond
+  `poll` elsewhere; without that, rounding each release up to 1 ms would have
+  capped the listener at one burst per millisecond regardless of window and RTT.
+
+  There is no operator knob — a pacing rate adjustable independently of the
+  window would be a way to defeat congestion control rather than to tune it.
+  See `docs/HTTP3_ROLLOUT.md`.
 - **Explicit Congestion Notification for HTTP/3 (#256-E)** — the QUIC listener
   now marks outbound datagrams ECT(0), counts the codepoints on received
   packets, reports them to peers in ACK_ECN, and validates the counters peers
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,34 @@ All notable user-facing changes to Tardigrade are documented here.
 ## [Unreleased]
 
 ### Added
+- **RFC 9002 packet pacing for HTTP/3 (#256-C)** — QUIC now spreads outbound
+  application data across the round trip instead of writing until the
+  congestion window is full. A window bounds how much data may be outstanding,
+  not how fast it may leave, so a sender with plenty of nominal capacity could
+  still hand its whole window to the first router in the path as one burst and
+  take queue-overflow loss for it.
+
+  The schedule is the leaky bucket RFC 9002 §7.7 describes: credit accrues at
+  `N × congestion_window / smoothed_rtt`, with `N` of 2 in slow start and 1.25
+  in congestion avoidance. Two things follow — datagrams leave one interval
+  apart once the bucket is empty, and a burst is capped at ten maximum-size
+  datagrams, which also caps what a connection returning from idle may send at
+  once. BBR and other rate estimators remain out of scope.
+
+  Only application data waits on the bucket. Acknowledgements, PTO probes, and
+  Initial/Handshake flights are charged to it but never delayed by it: an ACK
+  is not congestion-controlled traffic, a probe is how a stalled connection
+  recovers, and the handshake is already bounded by the initial window and by
+  anti-amplification. Congestion control and anti-amplification stay the hard
+  gates — pacing only ever delays a datagram, never authorises one.
+
+  Packet construction never sleeps: `pollTransmitOnPath` declines to build
+  paced data while the bucket is short, and the listener folds the connection's
+  next release time into the same poll timeout it already computes from timers,
+  so a paced connection neither spins nor sleeps past its own deadline. There
+  is no operator knob — a pacing rate adjustable independently of the window
+  would be a way to defeat congestion control rather than to tune it. See
+  `docs/HTTP3_ROLLOUT.md`.
 - **Explicit Congestion Notification for HTTP/3 (#256-E)** — the QUIC listener
   now marks outbound datagrams ECT(0), counts the codepoints on received
   packets, reports them to peers in ACK_ECN, and validates the counters peers
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,21 +12,27 @@ All notable user-facing changes to Tardigrade are documented here.
   still hand its whole window to the first router in the path as one burst and
   take queue-overflow loss for it.
 
-  The schedule is the leaky bucket RFC 9002 §7.7 describes: credit accrues at
-  `N × congestion_window / smoothed_rtt`, with `N` of 2 in slow start and 1.25
-  in congestion avoidance. Two things follow — datagrams leave one interval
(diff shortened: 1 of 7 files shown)


Jaruso (commented):
Changes required before merge.

I reviewed #615 against the current #256-C scope and current `main`, including the recovery scheduler, connection send gates, runtime wake integration, docs, and the existing post-handshake CRYPTO path.

There are two functional blockers in the pacing implementation and one production scheduling issue that prevents the claimed rate from being honored:

1. **The 1 ms recovery-timer floor is being used as a pacing floor.** For low-RTT/high-bandwidth paths this materially underpaces the connection. The computed ceil-div delay is already >= 1 µs whenever there is a non-zero deficit, so loss-recovery timer granularity should not be imposed on the pacer. The runtime then compounds this by converting the pacing deadline back to an integer-millisecond `poll()` timeout. With a ten-datagram bucket, a timer-driven sender can collapse to roughly ten datagrams per millisecond instead of the RFC-9002 rate. This especially conflicts with #256's high-bandwidth/loopback performance goal. Please preserve microsecond/sub-millisecond pacing end-to-end (for example, a sub-ms-capable wait primitive such as `ppoll`/timerfd on supported platforms, with a safe fallback), and add a low-RTT regression that would fail if a 50–500 µs computed release gets rounded to 1 ms.

2. **`nextSendTimeUs()` does not recognize application-level CRYPTO.** `buildPacket()` paces `.application` CRYPTO via `crypto_tx[2]`, but `nextSendTimeUs()` returns immediately unless `hasAppContent()` is true. `emitNewSessionTicket()` queues exactly this kind of post-handshake CRYPTO, so a spent bucket can cause `pollTransmitOnPath()` to refuse the ticket while the runtime never receives the pacing release deadline. A minimal fix is to make the wake predicate mirror the packet builder, e.g. `const has_paced_application = self.hasAppContent() or !self.crypto_tx[2].pending.isEmpty();`, and add a fake-clock NewSessionTicket regression proving refusal at `now`, a non-null future release, and emission at that release.

3. **The docs currently contradict the implementation for pure ACKs.** The implementation/tests intentionally meter only in-flight packets, so a pure ACK is neither delayed *nor charged* to the bucket. Please make the rollout doc/CHANGELOG/PR wording agree with that behavior, and explicitly document any other intentional pacing exemptions such as PMTU/path-validation probes rather than saying there are exactly three if the send path has more.

There is also a required integration step before this can merge: the PR head is one commit behind current `main` and GitHub reports it non-mergeable. Its merge base is `d97a128`, while current `main` is `981cc7c` (#609). When resolving the recovery conflict, preserve #609's `rtt_sample_available_at_send` bookkeeping and persistent-congestion semantics. In particular, the rebased send path should look conceptually like:

```zig
var tracked = packet;
tracked.rtt_sample_available_at_send = self.rtt.smoothed_rtt_us != null;
self.tracker.onPacketSentAssumeRecoveryCapacity(tracked);
if (tracked.in_flight) self.chargeSend(tracked);
```

(and the fallible `onPacketSent` equivalent), rather than taking the pre-#609 side of the conflict.

The overall separation of packet generation from scheduling is the right shape, and the cwnd/anti-amplification gates remain structurally independent. Once the wake predicate, sub-ms scheduling path, documentation, and rebase integration are corrected, the acceptance criteria should be close.

Jaruso (commented):
Changes still required before merge.

I re-reviewed the current rebased head (`b7c0c65`) end-to-end against #256-C, including every changed file, the previous four threads, the #609 recovery integration, the high-resolution runtime waiter, pacing gates/accounting, PMTU/path-validation/PTO exemptions, docs, and the current CI run.

The previous review findings are fixed and I resolved those four threads: the 1 ms pacing floor is gone, Linux/BSD/macOS now have a sub-millisecond wait path, application CRYPTO is included in `nextSendTimeUs()`, the ACK accounting docs now match the code, and the rebase preserved #609's `rtt_sample_available_at_send` bookkeeping. CI run 1434 is green across the Linux/macOS test/build matrix, perf smoke, H3/TLS interop, packaging, lint/format/security, and the appliance-profile jobs.

I found one remaining functional pacing-accounting bug plus two burst/specification issues that should be corrected before merge:

1. **Exempt packets lose pacing debt when the bucket is empty.** `onPacingSent()` uses saturating subtraction. That is fine for packets that were admitted by the bucket, but this PR intentionally lets PTO, Initial/Handshake, DPLPMTUD, and candidate-path validation packets bypass the pacing gate while still calling `chargeSend()`. If one of those packets leaves with fewer tokens than its size, the excess charge disappears. With the existing 1200-byte / 8 ms fixture, drain to zero, send one exempt 1200-byte packet at `now`, and ordinary application data is still released at `now + 8 ms`; if the exempt packet was actually charged, the sender owes that packet first and the next 1200-byte application packet should not be eligible until `now + 16 ms`. Please represent pacing debt explicitly (signed balance, `debt_bytes`, or a virtual-send-time schedule are all reasonable) so exempt sends may leave immediately but still push the later paced schedule forward. Add a deterministic regression covering the empty-bucket case, ideally through a real PTO/PMTU path as well as the unit-level pacer.

2. **`10 * max_datagram_size` is not always the initial congestion window the code/docs say it is.** Tardigrade supports an MDS up to 2048. At 2048, this bucket holds 20,480 bytes, while `CongestionController.initialWindow(2048)` is 14,720 bytes. After DPLPMTUD raises the path MDS, an idle connection can therefore restart with a burst substantially larger than the NewReno initial window even though the comments/docs explicitly say the ceiling is the initial window. Either cap the default bucket at `initialWindow(max_datagram_size)` (while preserving any intentionally separate, evidence-backed override), or change the contract/docs and justify why the implementation intentionally uses a 10-packet rather than initial-cwnd burst. Add a regression at an MDS above 1472 so this cannot silently drift again.

3. **The exact pacing gains are misattributed to final RFC 9002 §7.7.** The final RFC gives the `N * cwnd / srtt` form and says `N` should be a small value >= 1 (with 1.25 as an example); it does not prescribe `N=2` in slow start and `N=1.25` in congestion avoidance. That phase split appeared in earlier QUIC recovery drafts. The split itself can remain as an implementation choice if desired, but the code, CHANGELOG, and rollout docs should not say it is “exactly as RFC 9002 §7.7 describes.” Document it as Tardigrade's chosen gain policy (and optionally cite the historical draft) instead.

The generation/scheduling separation, congestion/anti-amplification hard gates, wake integration, session-ticket liveness fix, and fine-grained runtime waiter all look good on this pass. I did not find another issue in those previously reviewed paths.

Jaruso (commented):
Re-review complete on `d328b33604f83aac06a07e448930fd407dc30d57`.

I rechecked the full #256-C pacing slice, not just the last three fixes: recovery/pacer arithmetic, packet admission and charging, post-handshake CRYPTO, PTO/Initial/Handshake/PMTU/path-validation exemptions, congestion/anti-amplification gates, path-migration reset, production wake scheduling, UDP pacing seam, docs, and the deterministic regressions.

The three findings from the previous round are fixed correctly:

- Exempt in-flight packets now carry pacing debt through a signed balance. The regression proves an exempt 1200-byte send from an empty bucket leaves `-1200`, is only back to zero after one interval, and does not release ordinary application traffic until the second interval.
- The burst ceiling now derives from the NewReno initial window as well as the packet cap. At MDS 2048 it correctly caps credit at 14,720 bytes and permits only seven full-size packets at one instant rather than ten/20,480 bytes.
- The 2x slow-start / 1.25x avoidance split is now explicitly documented as Tardigrade policy inherited from earlier QUIC recovery drafts, not as a requirement of final RFC 9002 §7.7.

I also revalidated the prior fixes: sub-millisecond releases remain preserved end-to-end through `ppoll`/`kqueue`, application CRYPTO participates in `nextSendTimeUs`, ACK-only packets remain neither paced nor charged, #609's recovery bookkeeping is preserved, and the production sender still leaves congestion control and anti-amplification as the hard gates.

This satisfies the implementation requirements called out for #256-C: real cwnd/RTT pacing, bounded bursts, production wake integration without sleeping inside packet construction, no spin/cwnd-sized drain, responsive recovery traffic, normalized use of the existing pacing seam, and deterministic spacing/burst/idle/cwnd/deadline tests.

Current `main` is still `981cc7c`, the PR is mergeable, and CI run 1437 is green.

**No blocking code findings remain.** I would merge this version.

One non-blocking metadata cleanup: the PR description itself is now stale in three places — it still says pure ACKs are charged, still describes the ceiling as ten maximum-size datagrams, and presents the 2/1.25 split without the policy qualification. The code, CHANGELOG, and rollout documentation are correct; syncing the PR body would make the review record match the final implementation.

PR #616, opened 2026-08-14
docs: add configuration reference

Jaruso:
## Summary
- Add docs/CONFIGURATION.md as the complete operator-facing configuration reference.
- Document nginx-style config-file syntax, server/location blocks, all current EdgeConfig TARDIGRADE_* fields, defaults, enum values, and validation notes.
- Link the reference from README and examples index, and link the reference back to runnable examples.

## Validation
- git diff --check
- local Markdown link check for README.md, examples/README.md, docs/CONFIGURATION.md
- EdgeConfig TARDIGRADE_* coverage check against docs/CONFIGURATION.md

Closes #158

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,6 +7,7 @@
 <p align="center">
   <a href="https://github.com/Bare-Systems/Tardigrade/releases">Releases</a> |
   <a href="docs/SUPPORT_MATRIX.md">Support Matrix</a> |
+  <a href="docs/CONFIGURATION.md">Configuration</a> |
   <a href="docs/OBSERVABILITY.md">Observability</a> |
   <a href="SECURITY.md">Security</a> |
   <a href="CONTRIBUTING.md">Contributing</a>
@@ -213,6 +214,7 @@ operate without guessing which config file or pid file is active.
 | Topic | Location |
 | --- | --- |
 | Core v1 support matrix | [docs/SUPPORT_MATRIX.md](docs/SUPPORT_MATRIX.md) |
+| Configuration reference | [docs/CONFIGURATION.md](docs/CONFIGURATION.md) |
 | HTTP/3 rollout and lifecycle | [docs/HTTP3_ROLLOUT.md](docs/HTTP3_ROLLOUT.md) |
 | Concurrency & hot-path audit | [docs/CONCURRENCY.md](docs/CONCURRENCY.md) |
 | Event-loop backend evaluation (epoll/kqueue vs libxev/std.Io/io_uring) | [docs/EVENT_LOOP_BACKENDS.md](docs/EVENT_LOOP_BACKENDS.md) |
--- a/docs/CONFIGURATION.md
+++ b/docs/CONFIGURATION.md
@@ -3,12 +3,25 @@
 This reference documents the public Tardigrade configuration surface loaded by
 `EdgeConfig` and the nginx-style config-file adapter.
 
-Configuration precedence is:
+Runtime config-file selection, highest priority first:
 
-1. Built-in defaults.
-2. `TARDIGRADE_CONFIG_PATH`, when set, parsed as an nginx-style config file.
-3. `TARDIGRADE_SECRETS_PATH` and `TARDIGRADE_SECRET_KEYS`, when set.
-4. Environment variables, which override file-config values.
+1. `-c/--config` on commands that accept it.
+2. `TARDIGRADE_CONFIG_PATH`.
+3. `./tardigrade.conf`.
+4. `./config/tardigrade.conf`.
+5. `/etc/tardigrade/tardigrade.conf`.
+6. `$HOME/.config/tardigrade/tardigrade.conf`.
+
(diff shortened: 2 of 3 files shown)


Jaruso (commented):
**Review status: changes needed before merge.** GitHub does not permit the connected account to submit `REQUEST_CHANGES` on its own PR, so this is submitted as a comment review; the findings below are blocking.

Issue #158 asks for a source-of-truth operator reference: every public field, exact defaults/behavior, a per-field example, and at least one annotated full config. The new reference has strong breadth and the README/examples cross-links satisfy the navigation pieces, but there are still several blocking accuracy/acceptance gaps called out inline.

The most important correctness issue is that the documented source ordering does not match `EdgeConfig.loadFromEnv()`: process environment wins, then config-file overrides, then decrypted secret overrides, then built-in defaults. The current text reverses file and secret precedence and also omits runtime config-file discovery. Separately, `secrets_file`/`secret_key` are advertised as working config directives even though the secrets loader runs before the config file is parsed, so those directives cannot currently cause a secret file to be loaded; `secret_key` is also decryption key material, not an inline secret override.

The remaining changes are acceptance-criteria completeness: add an example for each field/directive, make derived/symbolic defaults concrete where the code has exact values, and actually annotate the full example. I included concrete expected documentation shapes/values in the inline comments so these should be addressable in one pass.

CI is green on the reviewed head. README and examples-index linking looks good, and the broad `EdgeConfig` key coverage is otherwise solid.

Jaruso (commented):
Second-pass review against #158 and the current implementation. The six blockers from my previous pass are materially addressed, and CI is green. I still found the blocking documentation/source-of-truth gaps called out inline below, so I would not merge this yet.

The remaining work is documentation of existing behavior, not feature expansion: fix the invalid encodings/examples, document the loader's source-specific alias behavior and validation constraints, cover the config-file directives the parser already supports, and make reload/profile/startup-owned behavior explicit. Once these are corrected I expect the acceptance criteria for #158 to be substantially satisfied.

Jaruso (commented):
**Review status: changes still needed before merge.** The previous 12 findings are addressed on this head, and CI #1435 is green. I re-audited the resulting reference against issue #158 and the runtime/parser rather than only checking thread resolution.

There are still a few source-of-truth gaps called out inline. The most important is reload semantics: the new section currently says all restart-owned changes are rejected, but the runtime has two distinct behaviors—some changes are hard-rejected before publication, while others are accepted/published even though startup-owned live resources keep their old state until restart. The new canonical example also enables master-process mode, which exposes an existing control-path limitation: the master does not consume/forward the SIGHUP reload flag, and workers share/overwrite the same PID file, so the example cannot currently be presented alongside the generic `tardi reload` contract without qualification.

The other remaining items are field-level acceptance gaps: exact TLS version values, `TLS_SERVER_NAME` profile scope, appliance PROXY-protocol restriction, missing restart warnings on four H3 listener-owned knobs, the CLI-only public env fields not receiving the same per-field schema/examples, and two effective-value edge cases (`RATE_LIMIT_RPS` and `APPROVAL_TTL_MS`).

Because the connected account is the PR author, GitHub will not allow a formal `REQUEST_CHANGES`; this COMMENT review is blocking. Once these are corrected, I do not currently have another unposted finding from this pass.

Jaruso (commented):
Fourth-pass review against #158 and the current implementation. The nine findings from my previous pass are materially addressed, and CI #1439 is green. I still found blocking source-of-truth gaps in the reference: the mixed top-level/server-block canonical example does not route as documented; several compact location examples do not parse; the reloadability table still omits startup-owned state; the OTEL rows claim an exporter that is not wired; the trust rows describe response verification the runtime does not perform; several protocol-adapter rows overstate current live wiring; and master-process control limitations are not documented. I also left one smaller correction for retry coercion.

These are documentation-of-existing-behavior fixes, not requests to add new features. I would not merge #616 yet. Once these are corrected, the reference should be much closer to satisfying #158 as an actual operator source of truth rather than a catalog of parsed fields.

Jaruso (commented):
Documentation-focused re-review against #158 on head `6f3c6906d5c63af65853ab2f7b121ff38b706020`. I rechecked this as a documentation change rather than using it as a vehicle to audit or redesign pre-existing runtime behavior.

The previous documentation inaccuracies are fixed: the location examples are parser-valid, server-block selection and the annotated example match current behavior, OTEL/trust/protocol-adapter rows accurately distinguish what is currently wired from what is only parsed/reserved, reload behavior is categorized without conflating hard rejection with startup-owned state, the master-mode caveat is documented, and retry-attempt semantics now match the loader.

Against #158 specifically, the reference now satisfies the acceptance criteria: the current public configuration surface is covered with types, defaults, valid/effective behavior, examples, and relevant warnings; a full annotated configuration is included; README and the examples index link to the reference; the reference links back to the curated runnable examples; and this PR remains documentation-only.

I have no remaining documentation blockers. CI #1444 is still in progress at the time of this review; assuming it passes, this is ready to merge.

PR #618, opened 2026-08-14
docs(quic): batching/GSO/GRO design notes (#256-F)

Jaruso:
## Summary

`#256`'s A/B/C/D/E slices landed the effective datagram cap, DPLPMTUD,
socket buffer tuning, ECN, and real RFC 9002 pacing. What's left of the
original scope is slice F: **design/follow-up notes for batching, GSO,
and GRO** — the issue is explicit that this should stay documentation
unless benchmark evidence justifies implementation, not a second
transport architecture bolted onto this PR.

Adds a "Batching, GSO, and GRO" section to `docs/HTTP3_ROLLOUT.md`
covering:

- **Where the runtime stands today** — `drainConnectionTransmits` and
  the receive loop in `src/http/http3_runtime.zig` move one datagram
  per `sendmsg`/`recvmsg` today; neither `sendmmsg`/`recvmmsg` nor UDP
  GSO/GRO (`UDP_SEGMENT`/`UDP_GRO`) is implemented.
- **Where it would plug in** — production batching can remain
  runtime-local in `src/http/http3_runtime.zig`; the production
  listener does not use the scalar `quic.udp.Endpoint` contract today.
  A send batch needs stable batch-owned backing for each produced
  datagram and explicit partial-send/fallback handling, since
  `Transmit.bytes` aliases the caller's scratch buffer and
  already-accounted packets must not be discarded or re-polled.
- **What batching must not break** — pacing's burst ceiling bounds
  what may be released into a batch; `sendmmsg` preserves per-message
  path/ECN metadata while a single GSO super-buffer requires
  compatible path/ECN/segment size; PMTU probes retain their
  packet-number identity (not size); GRO receive must preserve
  per-datagram routing/ECN semantics before one `ingest()` call per
  datagram.
- **Evidence bar before implementing** — reuse the existing
  competitive benchmark framework (#149); the trigger is a profile
  showing the syscall boundary actually dominates at realistic
  throughput, not a benchmark run merely existing.
- **Platform support** — `UDP_SEGMENT`/`UDP_GRO` are Linux-only with
  no macOS/BSD equivalent; any future implementation gates on the same
  per-platform tables the ECN and no-fragmentation sections already
  use, and falls back to the per-datagram path on failure.

No code changes — this is documentation only, matching the issue's
scope for this slice.

## Test plan

- [x] Docs-only change; no build/test surface affected.
- [x] Cross-checked the new section's claims against the current
      `src/quic/udp.zig` and `src/http/http3_runtime.zig` send/receive
      loops (`drainConnectionTransmits`, `serve`, `ingest`).
- [x] Matched terminology and section style to the existing "Socket
      buffers", "Explicit Congestion Notification", and "Packet
      pacing" sections in the same doc.

Part of #256

### DIFF
--- a/docs/HTTP3_ROLLOUT.md
+++ b/docs/HTTP3_ROLLOUT.md
@@ -532,6 +532,112 @@ operator could raise independently of the window would be a way to defeat
 congestion control rather than to tune it. A path whose measured capacity
 grows paces faster on its own.
 
+## Batching, GSO, and GRO
+
+This is a design note, not an implementation (#256-F). #256's non-goals rule
+out building this speculatively; what follows is where it would plug into the
+current contract, what it must not be allowed to break, and what evidence
+would justify picking it up.
+
+### Where the runtime stands today
+
+Every datagram crosses the syscall boundary alone. `drainConnectionTransmits`
+calls `sendDatagram` — one `sendmsg`/`sendto` — for each datagram
+`pollTransmitOnPath` releases, and the receive loop calls `receiveDatagram` —
+one `recvmsg` — repeatedly until the socket reports `EAGAIN`. At the packet
+sizes DPLPMTUD settles on (1200–2048 bytes) and the rates pacing now permits,
+a saturated loopback or datacenter link spends a growing share of its time
+crossing that boundary rather than building or parsing the datagrams that
+cross it. `sendmmsg`/`recvmmsg` (batched syscalls) and UDP GSO/GRO (kernel-side
+segmentation and coalescing via `UDP_SEGMENT` and `UDP_GRO`) are the standard
+answers to that cost, and neither is implemented.
+
+### Where it would plug in
+
+The seam is already there on both sides, and neither needs a new abstraction
+in `src/quic/udp.zig`: `SendDatagram` and `ReceivedDatagram` are already
+per-datagram values carrying their own `ecn` and address, which is what a
+batch is a list of.
+
+- **Send.** `Connection.pollTransmitOnPath` already returns one datagram at a
+  time by contract; `drainConnectionTransmits`'s `while` loop is the only
+  place that would change, assembling either a `sendmmsg` vector of
+  independent datagrams or a single GSO buffer of same-sized segments plus a
+  `UDP_SEGMENT` control message. Packet *construction* would not change —
+  pacing already keeps `pollTransmitOnPath` from blocking, and that
+  separation between "build a datagram" and "decide when it may leave" is the
+  same seam a batching send path needs, noted where pacing reaches the wire
+  above.
+- **Receive.** The `while (true) { receiveDatagram(...) }` loop in `serve`
+  already drains everything waiting before sleeping again; `recvmmsg` would
+  replace repeated `recvmsg` calls with one call returning several datagrams,
+  and `UDP_GRO` would let the kernel coalesce datagrams from one peer before
+  the syscall even happens. `ingest` already takes one datagram, one peer
+  address, and one ECN mark per call, so a batched receive path would still
+  call it once per datagram inside the batch — no change to DCID routing or
+  connection lookup.
+
+### What batching must not break
+
+A batch is a kernel-level optimization; it must stay invisible to the
+transport guarantees layered above it.
+
+- **Pacing** decides *whether* a datagram may leave, not how it is
+  transmitted once released. A GSO buffer could only ever hold datagrams the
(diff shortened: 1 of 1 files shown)


Jaruso (commented):
**Treat this as a blocking review before merge.** GitHub does not allow the connected account to formally request changes on its own pull request, so this is submitted as a COMMENT review instead.

The overall section is aimed at the right #256-F scope and covers the required areas (integration point, pacing/ECN/PMTU invariants, evidence bar, and platform fallback), but several parts of the current text would give the follow-up implementer the wrong contract. Please address all inline comments in this round; **do not implement batching/GSO/GRO in this PR**.

1. **Describe the actual runtime/UDP contract and batch ownership semantics.** The production H3 runtime currently bypasses `quic.udp.Endpoint`, and the endpoint itself is scalar. More importantly, the current transmit bytes alias caller-owned scratch storage, so a future batch needs stable backing for every already-produced datagram and a defined partial-send/fallback policy. The doc should explicitly choose whether batching remains runtime-local or requires a batch-capable endpoint contract; it should not claim no abstraction/API work is needed.
2. **Separate `sendmmsg` batching from GSO restrictions.** `sendmmsg` has one `msghdr` per datagram and can preserve per-message destination/ECN metadata. The same-path/same-ECN restriction belongs to a single `UDP_SEGMENT` GSO super-buffer, not to every batch.
3. **Correct PMTU probe identity.** `pmtu.Controller` deliberately identifies an outstanding probe by packet number; size is the probe target and may be retried. Preserve that distinction in the design note.
4. **State the actual receive-GRO ECN invariant.** A shared source tuple alone does not imply a shared ECN codepoint. The Linux GRO behavior that makes a shared TOS/TCLASS cmsg safe must be part of the stated platform contract before expanding that value to each reconstructed datagram.
5. **Attribute the docs-only scope correctly.** The original #256 scope asks for future batching/GSO/GRO design notes; the docs-only restriction comes from the #256-F reconciliation, not from the issue’s `Non-goals` list.

I also checked the rest of the added section against the current #256 A–E implementation: the 1200–2048 datagram range, bounded pacer burst, existing per-datagram runtime loop, benchmark-framework reuse, and Linux GSO/GRO version statements are consistent with the current tree. No additional code or tests should be necessary for this docs-only slice after these corrections.

Jaruso (commented):
Second-pass review on `41e90f51d2399ceaa04e105bb7b60834a8834e50`.

The **file-level changes are now good**. I rechecked the full revised `Batching, GSO, and GRO` section against the current production H3 runtime, scalar `quic.udp.Endpoint` contract, transmit scratch-buffer lifetime, DPLPMTUD probe tracking, ECN semantics, pacing/burst behavior, and the #256-F scope. All five findings from the prior review have been addressed, and I did not find another substantive correctness/design issue in the documentation itself. CI on this head is also green.

There is **one remaining pre-merge cleanup, in the PR description rather than the diff**: the Summary still contains two claims from the original version that the revised document correctly removed:

1. Under **“Where it would plug in”**, the PR body still says the send/receive loops are the only code that would change and that `SendDatagram`/`ReceivedDatagram` already make the seam batch-ready. The revised doc correctly explains that production bypasses `quic.udp.Endpoint`, the endpoint is scalar, and transmit batching needs stable batch-owned backing plus partial-send ownership semantics.
2. Under **“What batching must not break”**, the PR body still says PMTU probes are “identified by exact size.” The current implementation and revised doc correctly say the probe target has a size, but the outstanding probe is identified by **packet number** (`pmtu.Controller.outstanding_pn`).

Please sync those two bullets with the final document before merge. Suggested replacement text:

```md
- **Where it would plug in** — production batching can remain runtime-local in
  `src/http/http3_runtime.zig`; the production listener does not use the
  scalar `quic.udp.Endpoint` contract today. A send batch needs stable
  batch-owned backing for each produced datagram and explicit partial-send /
  fallback handling so already-accounted packets are not discarded or
  re-polled.
- **What batching must not break** — pacing's burst ceiling bounds what may be
  released into a batch; `sendmmsg` preserves per-message path/ECN metadata
  while a single GSO super-buffer requires compatible path/ECN/segment size;
  PMTU probes retain their packet-number identity; GRO receive must preserve
  per-datagram routing/ECN semantics before one `ingest()` call per datagram.
```

**No further code or documentation changes are requested.** Once the PR description is updated to stop contradicting the final document, I consider #618 ready to merge; I do not think another review round is necessary for that metadata-only edit.

PR #619, opened 2026-08-14
quic: wire HTTP/3 qlog and keylog artifacts

Jaruso:
## Summary

- Add explicit HTTP/3 qlog/keylog artifact configuration via `TARDIGRADE_HTTP3_QLOG_DIR` and `TARDIGRADE_HTTP3_KEYLOG_PATH`.
- Wire a gateway-owned artifact writer that creates per-connection `.sqlog` files, appends shared TLS keylog lines, and keeps file errors out of protocol state.
- Document enablement, restart behavior, sensitivity, and cleanup in QUIC qlog and configuration docs.

## Validation

- `zig build test --summary all` — 58/58 steps succeeded; 3723/3733 tests passed (10 skipped)

Refs #255 slice 6.


Jaruso (commented):
**Blocking review — changes required before merge.** GitHub does not allow the connected account to formally request changes on its own PR, so this is submitted as a COMMENT review, but I would treat it as REQUEST_CHANGES.

This is close structurally—the config/reload wiring, shared QUIC+H3 file format, default-off behavior, docs, and shared TLS keylog callback are headed in the right direction—but I found five issues that need to be fixed before this satisfies #255 slice 6.

The most important problems are lifecycle and hot-path safety: completed connections never release their qlog trace FDs, the gateway artifact-only path misses `quic:connection_started`, and qlog disk I/O is performed synchronously on the single HTTP/3 runtime thread even though the issue explicitly requires the writer to be bounded/non-blocking from the transport's perspective. There is also a keylog confidentiality issue when the configured file already exists with permissive mode bits, and the retained write-error counters are never surfaced to an operator, so truncated traces can still be silent.

Please also add a lifecycle-level test that exercises the runtime wiring rather than only invoking `ObservabilityArtifacts.write*Record` directly. At minimum it should prove that a real accepted trace contains `quic:connection_started`, QUIC/H3 records share the same `.sqlog`, connection teardown removes/closes the trace, and repeated connections do not grow the trace map/FD count. Add a keylog test that pre-creates the destination with permissive permissions and verifies enablement tightens or rejects it.

CI status at review time: every completed job I checked is green; the Ubuntu ARM appliance-profile job is still in progress. These findings are behavioral/acceptance gaps rather than build failures.

Jaruso (commented):
**Second blocking review — changes still required before merge.**

The prior round made substantial progress. I rechecked each of the five previous findings against head `527fccb`:

- Per-connection trace teardown: implemented for the normal queued path, but the new saturation policy can discard older `close_trace` commands and recreate the FD/resource leak under load (inline finding below).
- `quic:connection_started`: fixed correctly via `emitQuicQlog()` after successful admission.
- Blocking filesystem I/O: fixed architecturally by moving file creation/writes to a dedicated writer thread, but the producer side now drops records on *any transient mutex contention*, not just bounded-queue exhaustion. This is already causing the new lifecycle integration test to fail on Ubuntu CI.
- Existing keylog permissions: fixed; `fchmod(fd, 0600)` is applied before secrets are written and the permissive-file test is present.
- Artifact failure visibility: write failures are now logged and counted, but queue/serialization drops still have no operator-visible signal until final process teardown. A long-running debug session can therefore use a truncated trace without knowing it is incomplete.

CI is currently red on run 1451: `Test (ubuntu-latest)` fails `observability artifacts follow accepted connection lifecycle` because the expected `http3:stream_type_set` record is absent. macOS and ARM happen to pass, which is consistent with the scheduling-dependent `tryLock()` drop path rather than a deterministic serializer problem. The failed assertion also leaves the accepted connection allocated, producing five leak reports; please make the test cleanup unconditional so future assertion failures don't obscure the actual failure.

I would keep this PR blocked until the queue guarantees normal low-volume records are accepted when capacity exists, lifecycle/control commands cannot be discarded under saturation, and dropped-record diagnostics are visible before shutdown. After those changes, rerun the full CI matrix and specifically keep the lifecycle test deterministic across repeated runs.

Jaruso (commented):
**Blocking review — changes still required before merge.** GitHub does not allow the connected account to formally request changes on its own PR, so this is a COMMENT review, but I would still treat it as REQUEST_CHANGES.

The previous round is materially improved: the Ubuntu lifecycle failure is fixed, `connection_started` is present, keylog permissions are hardened, diagnostics are surfaced, and the writer owns normal file I/O. CI run 1456 is fully green across Linux/macOS/ARM and the appliance/interop jobs I checked.

I re-read the current queue implementation against #255 slice 6's explicit requirement that the qlog writer be **bounded/non-blocking from the transport's perspective**. Three queue/lifecycle problems remain:

1. `tryEnqueue()` now uses a blocking mutex. That fixes the earlier `tryLock()` scheduling-loss bug, but it regresses the acceptance criterion: the QUIC/H3/TLS producer can wait indefinitely if the writer thread is descheduled while holding the queue mutex. Disk I/O being outside the lock does not make the producer wait bounded.
2. The worker always consumes data before close commands. Under sustained tracing where producers keep the 256-entry data queue non-empty, `close_trace` commands can starve indefinitely, so closed-connection traces/FDs accumulate and the 1024-entry close queue can eventually fill.
3. The full-close-queue fallback is unsafe. `dropAllDataAssumeLocked()` cannot create space in `close_queue`, so the subsequent `close_queue_len == close_queue.len` check is necessarily still true and the runtime thread calls `closeTraceOnWriter()` itself. The writer processes a popped record outside the mutex, so this can free a `Trace`/file while the writer is using it, or access the writer-owned `traces` map concurrently. It also puts file close/free work back on the transport thread.

Please keep all trace-map/file ownership on the writer and use an ordered, producer-nonblocking queue design. A good shape is a lock-free/SPSC data queue plus close barriers that capture the producer sequence (`close { handle, after_seq }`), allowing the writer to close a trace as soon as all records that preceded that close have been consumed even if newer connections continue producing data. A unified FIFO with guaranteed non-droppable control admission can also work; the important properties are (a) producer operations have a hard bound, (b) close requests cannot starve/drop, (c) close ordering is after prior records for that handle, and (d) the producer never touches `traces` or file descriptors.

Please add tests that actually exercise the missing cases. The current saturation test fills the **data** queue and enqueues only 16 closes, so it never reaches `close_queue_capacity` or the unsafe fallback. Add a test that drives close-control saturation and another that keeps data arriving while connections close, proving trace reclamation still makes progress without waiting for the global data queue to become empty. Also retain a zero-drop low-volume test for the normal path.

Jaruso (commented):
**Blocking review — changes still required before merge.** GitHub does not allow the connected account to formally request changes on its own PR, so this is a COMMENT review, but I would treat it as REQUEST_CHANGES.

The previous three findings are directionally fixed: normal qlog/keylog enqueue is now producer-bounded, close barriers carry a sequence so they do not starve behind newer data, and overflow cleanup no longer mutates writer-owned trace state directly from the runtime thread. I re-read the replacement queue state machine and found three remaining concurrency/lifecycle bugs.

Most importantly, CI run 1460 is already reproducing one of them on `ubuntu-24.04-arm` appliance tests: `observability artifacts follow accepted connection lifecycle` returns from `waitIdle()` with `traceCount() == 0`, and `keylog initialization tightens permissive existing files` returns before the `cc` secret line is appended. The worker removes a data item from the ring before publishing `worker_active=true`, so the completion barrier has a window where it sees both “queue empty” and “worker inactive.”

There is also a classic lost-condition-wakeup window because producers now mutate the atomic predicate and call `Condition.signal()` without synchronizing with the mutex used by the waiter. A final record/close can therefore sit queued until some later event, and the same race during `deinit()` can leave `thread.join()` waiting forever. Please use a persistent wake primitive (counting semaphore/event counter/futex generation, etc.) so producer wakeups cannot be lost without reintroducing an unbounded producer mutex.

Finally, close-control overflow is still not semantically safe. `close_all_after_seq` eventually calls `closeAllTracesOnWriter()`, which closes traces for unrelated **active** connections too. Those connections can then emit another record and cause `traceFor()` to create a suffixed second `.sqlog`, splitting one live connection across files. Preserve the exact overflowed close intent, or if the only bounded fallback is to abandon capture, atomically disable further qlog writes and surface that truncation before closing all traces. Do not silently close active traces and then allow them to reopen.

Please add deterministic coverage for all three cases: (1) an in-flight/dequeued command cannot satisfy `waitIdle()` until processing finishes, (2) a single enqueue/close and shutdown cannot miss the writer wakeup, and (3) close-queue overflow with an unrelated active trace leaves that trace open and in the same file. Current CI is red on the first case already.

Jaruso (commented):
**No remaining blockers on this pass.** GitHub does not allow the connected account to approve its own PR, so I am submitting this as a COMMENT review; I would treat this as approval for #255 slice 6.

I re-reviewed the current head (`62f7e376`) against the full slice-6 acceptance criteria and specifically rechecked every blocker from the prior rounds. The latest race fix closes the remaining concurrency holes:

- the data/close producer path remains bounded and does not wait on the writer-owned mutex;
- `worker_active` is published before a dequeued command can disappear from `waitIdle()`'s completion predicate, fixing the ARM race from CI 1460;
- the semaphore + `wake_pending` protocol makes wakeups persistent across the check-to-sleep window and also gives `deinit()` a reliable wake path, so the old lost-condition-signal/join-hang case is gone;
- close barriers still preserve all records accepted before connection teardown and are serviced ahead of newer data once their sequence is ready;
- close-queue overflow first preserves the exact close intent in the bounded data queue; only if both bounded queues are saturated does it fail closed by disabling future qlog capture, surfacing a qlog drop, draining through the accepted-record barrier, and then closing writer-owned traces. Because qlog stays abandoned afterward, active connections cannot silently reopen a suffixed second trace;
- keylog append remains writer-serialized/O_APPEND, existing destinations are tightened to `0600`, and key material stays out of ordinary logs/metrics/qlog;
- qlog/keylog remain disabled by default, per-connection filenames are deterministic/collision-safe and non-secret, enablement warnings and cleanup/qvis/Wireshark guidance are documented, and artifact failures/drops remain isolated from protocol state and surfaced diagnostically.

The new regression coverage now exercises the worker-active handoff, repeated single-record semaphore wake/deinit, exact close-intent overflow with an unrelated active trace, and the fully saturated fail-closed path.

**CI:** run 1464 completed successfully. The general and appliance test matrices are green on x86 Linux, ARM Linux, and macOS; build variants, TLS dependency audit, TLS interop, H3 resumption/0-RTT interop, security, packaging, lint/format, and perf smoke are also green.

I do not see another code change required before merge for this slice.

Jaruso (commented):
Post-merge re-review: no new findings. PR #619 was squash-merged after the prior clean review with the same reviewed head (`62f7e37645437a73edb738cd9aaa499ad3195fce`). I compared the resulting main commit (`5cc70d5a2549f1c5dc2a6f98cc4f52ea6453702c`) against its parent: it contains the same eight changed files and the same +1095/-26 change totals as the reviewed PR, with no merge-resolution delta. The last PR CI run (#1464) was fully green, including x86 Linux, ARM Linux, macOS, appliance tests, TLS/H3 interop, security, packaging, and perf. The #255 slice-6 acceptance criteria remain satisfied. No blockers or follow-up changes identified.

PR #620, opened 2026-08-14
feat(cli): add profile-aware config init (tardi init <profile>)

Jaruso:
## Summary

- Extends the existing `tardi config init` starter generator into a profile-aware first-run experience per #162, without replacing it or introducing a second config system.
- Adds the concise primary command `tardi init <profile>` (stdout-only, config bytes only on success — no banner, no status text) for the five required profiles: `static`, `proxy`, `tls`, `metrics`, `prod`, each with a descriptive alias (`static-site`, `reverse-proxy`, `tls-termination`, `metrics-enabled`, `production-baseline`) that resolves to byte-identical output.
- Extends the verbose `tardi config init` with `--profile <profile>`, while fully preserving existing behavior (positional output path, `--stdout`, `--force`, overwrite refusal, parent-directory creation, and the legacy no-profile generic starter).
- New `src/config_profiles.zig` owns the canonical profile enum, exact-match alias table, and immutable template bytes — `src/main.zig` only owns CLI dispatch.
- All templates use only currently-supported nginx-style config syntax (no TOML, no new runtime/config directives). TLS/prod profiles point at operator-owned credential paths (`/etc/tardigrade/tls/...`) rather than embedding any secret. The `metrics` profile documents the existing `/status/metrics` endpoint and warns about `TARDIGRADE_METRICS_REQUIRE_AUTH` instead of inventing a new metrics directive.
- README Quick start now leads with `tardi init static > tardigrade.conf`; `examples/README.md` cross-references each profile to its corresponding full example.

## Test plan

- [x] `zig build test` (default/general profile) and `zig build test -Dtls-profile=appliance`
- [x] Unit coverage in `src/config_profiles.zig` and `src/main.zig`: every canonical profile/alias parses, aliases are byte-identical to canonical output, missing/unknown profile rejected (with discoverable valid names), too-many-arguments rejected, `-h`/`--help` prints usage without generating a config, `config init --profile` composes with positional path/`--force`, no-profile `config init` keeps the legacy starter.
- [x] New real-validator integration tests in `tests/integration.zig` (`zig build test-integration-init`, also covered by the existing `test-integration`/`test-appliance` CI jobs) spawn the actual built binary: `static`/`proxy`/`metrics` pass the real `tardi check` as generated; `tls`/`prod` pass after substituting the repository's Ed25519 TLS fixtures for the generated placeholder cert/key paths — verified under both the general (openssl-adapter) and appliance TLS profiles.
- [x] `./scripts/validate-examples.sh` — all 11 existing example configs still pass (unchanged).
- [x] Manual CLI smoke test of `tardi init <profile>` / `--help` / missing / unknown / too-many-args / `config init --profile ... --stdout`.

Closes #162


Jaruso (commented):
I found three changes I would make before merging this.

### 1. `--profile` consumes another option as its value instead of reporting a missing value

`parseConfigInitCommand` only checks `idx + 1 >= args.len` before accepting the token after `--profile`. That means malformed invocations such as:

```bash
tardi config init --profile --stdout
tardi config init --profile --force
tardi config init --profile --help
```

consume the following flag as `profile_name` and later report e.g. `unknown config profile '--stdout'` instead of `missing option value`. The new unit test only covers `--profile` at end-of-argv, so it misses this case.

Please reject a would-be profile value that is itself an option (at minimum starts with `-`) and add coverage for `--profile --stdout` / `--profile --force` so the CLI reports the actual mistake. This is especially worth fixing here because profile parsing is new operator-facing behavior and the issue explicitly calls for clean missing-value/error handling.

### 2. The new real-validator integration step is not enforced for the default/general profile on PRs

`build.zig` adds the useful `test-integration-init` step, but `.github/workflows/ci.yml` never invokes it. The normal `Integration tests` job explicitly runs only on pushes to `main`, not pull requests. PR CI currently gets these new tests only indirectly through the Linux `test-appliance` full integration suite, which runs with `-Dtls-profile=appliance`.

That leaves the default/general binary's real `tardi init ... -> tardi check` path untested before merge, even though TLS credential validation differs by build profile and the PR description says the generated profiles were verified under both general and appliance profiles.

Please wire the focused `zig build test-integration-init` step into PR CI for the normal/general profile (one Linux x64 job is enough; no need to multiply the full suite). The appliance suite can continue providing the native-profile coverage.

### 3. Add the operator-visible change to `CHANGELOG.md`

The repository's own `docs/CODE_REVIEW_CHECKLIST.md` requires `CHANGELOG.md` to be updated for operator-visible behavior. This PR adds a new top-level CLI command, five public profiles, aliases, and a new `config init --profile` option, but the changelog is untouched.

Please add a concise entry under `[Unreleased] -> Added` covering `tardi init <profile>` and the compatible `tardi config init --profile <profile>` form.

Everything else I checked looks aligned with #162: the nginx-style templates are sensible, aliases resolve through one canonical profile type, existing no-profile `config init` behavior is preserved, concise init stays stdout-clean, TLS/prod use external credentials rather than weakening validation, metrics documents the existing endpoint instead of inventing config syntax, and the generated-config tests exercise the production `check` path.

PR #621, opened 2026-08-14
docs(deployment): add production deployment docs for Docker and systemd

### LINKED ISSUE #166: ux: add production deployment docs for Docker and systemd
## Summary
Add practical production deployment documentation for running Tardigrade under Docker and systemd.

## Motivation
A server project feels much more adoptable when operators can see how to run it outside a development shell.

## Scope
- Document Docker usage if image/build support exists.
- Provide a sample Dockerfile or compose file if appropriate.
- Provide a sample systemd unit.
- Document recommended file layout for config, certs, static roots, and logs.
- Document restart/reload commands.
- Document basic production hardening checklist.

## Acceptance criteria
- Operators can run Tardigrade using Docker or systemd from the docs.
- Sample configs are valid and consistent with examples.
- Docs cover permissions, ports, log locations, and reload behavior.
- README links to deployment docs.

## Notes
This should avoid overpromising production readiness. Be explicit about known limitations.

---

## Maintainer implementation guidance — 2026-08-14

This issue is still valid, but the repository has changed substantially since it was created. Treat this as a **deployment integration + documentation** ticket, not a broad packaging effort from scratch.

### Current state / work already completed elsewhere

The following pieces already exist and should be reused rather than reimplemented:

- Published Linux release archives and DEB/RPM packages.
- `packaging/systemd/tardigrade.service` as the shared Linux service unit used by package builds.
- `packaging/tardigrade.conf` as the package starter config.
- `packaging/README.md` covering current package/install status.
- `examples/production-baseline/` covering a hardened config baseline.
- `docs/RELOAD_SHUTDOWN.md` covering reload, graceful drain, hard termination, in-flight work, worker drain behavior, upstream pools, commands, logs, and metrics.
- `docs/OBSERVABILITY.md` for logging/metrics behavior.
- `docs/SUPPORT_MATRIX.md` for the stable/experimental feature contract.
- Canonical CLI naming has already been resolved: `tardi` is the public command; `tardigrade` is a compatibility alias.

Do not duplicate those docs. Link to them from the deployment guide and only summarize the minimum needed to complete a deployment workflow.

### Dependency / implementation order

Issue #158 is being implemented by PR #616 (`docs/CONFIGURATION.md`). This issue should be implemented after #616 lands, or rebased on top of it before completion, so deployment docs can link to the canonical configuration reference rather than duplicating defaults or config semantics.

Expected dependency order:

1. #158 / PR #616 lands.
2. Implement this issue.
3. Deployment docs link to `docs/CONFIGURATION.md`, `docs/RELOAD_SHUTDOWN.md`, `docs/OBSERVABILITY.md`, `docs/SUPPORT_MATRIX.md`, `packaging/README.md`, and `examples/production-baseline/`.

### 1. Add `docs/DEPLOYMENT.md`

Create one canonical operator-facing deployment guide covering both host-native Linux/systemd and locally built Docker/OCI deployment.

The opening section must explicitly state the maturity boundary:

- Tardigrade supports practical operator-managed deployments.
- Do **not** claim generic "production-ready", "battle-tested", or "drop-in nginx replacement" status.
- The supported/stable runtime surface is defined by `docs/SUPPORT_MATRIX.md`.
- Experimental protocol surfaces remain experimental regardless of deployment method.

The guide should have two primary paths:

1. Host-native Linux + systemd.
2. Docker/OCI built locally from this repository.

Do not document an official published container image unless this issue actually adds and publishes one. The intended scope here is local Docker build support, not registry distribution.

### 2. Document the recommended host filesystem layout

Use a concrete layout like:

```text
/usr/bin/tardi                         binary
/etc/tardigrade/
    tardigrade.conf                    routing/runtime configuration
    tardigrade.env                     environment-only settings / secret references
    tls/
        fullchain.pem
        privkey.pem
/run/tardigrade/
    tardigrade.pid                     ephemeral process PID
/var/lib/tardigrade/                   service working/state directory
/var/lib/tardigrade/public/            optional static root
/var/log/tardigrade/                   optional file logs
```

Document expected permissions/ownership using the package behavior as the source of truth:

- `/etc/tardigrade/tardigrade.conf`: `root:root`, normally `0644`.
- `/etc/tardigrade/tardigrade.env`: `root:tardigrade`, `0640`.
- TLS private key: readable by the `tardigrade` service account/group only; never world-readable.
- `/run/tardigrade`: writable by `tardigrade`, ephemeral.
- `/var/lib/tardigrade`: `tardigrade:tardigrade` when writable runtime state is needed.
- Static roots should normally be read-only to the service when writes are unnecessary.

Also explain that arbitrary configured roots/Unix sockets may require additional filesystem access, so operators should adjust sandboxing deliberately rather than weakening it globally.

### 3. Fix the current systemd PID/control-path inconsistency

This is a required runtime/package fix discovered while researching this issue.

The current unit uses:

```ini
ExecStart=/usr/bin/tardi run -c /etc/tardigrade/tardigrade.conf
ExecReload=/usr/bin/tardi reload -c /etc/tardigrade/tardigrade.conf
ExecStop=/usr/bin/tardi stop -c /etc/tardigrade/tardigrade.conf
```

However, the packaged starter config and generated DEB/RPM environment templates do not currently configure a PID file. The CLI requires one of:

- `--pid`
- `--pid-file`
- a configured `pid` / `TARDIGRADE_PID_FILE`

for `reload`, `stop`, and process-aware `status` operations. Without one, the CLI reaches `MissingPidTarget`.

Therefore a fresh package install does not currently have a self-contained `systemctl reload tardigrade` / `systemctl stop tardigrade` control contract.

Fix the service so the runtime and control commands share a deterministic PID path.

Recommended shape:

```ini
[Service]
Type=simple
User=tardigrade
Group=tardigrade
WorkingDirectory=/var/lib/tardigrade

EnvironmentFile=-/etc/tardigrade/tardigrade.env
Environment=TARDIGRADE_PID_FILE=/run/tardigrade/tardigrade.pid

RuntimeDirectory=tardigrade
RuntimeDirectoryMode=0750

ExecStartPre=/usr/bin/tardi check /etc/tardigrade/tardigrade.conf
ExecStart=/usr/bin/tardi run -c /etc/tardigrade/tardigrade.conf
ExecReload=/usr/bin/tardi reload --pid-file /run/tardigrade/tardigrade.pid
ExecStop=/usr/bin/tardi stop --pid-file /run/tardigrade/tardigrade.pid

Restart=on-failure
RestartSec=2
TimeoutStopSec=35s
LimitNOFILE=65536

NoNewPrivileges=true
PrivateTmp=true
ProtectHome=true
ProtectSystem=full
UMask=0027

[Install]
WantedBy=multi-user.target
```

The exact hardening directives may be adjusted if testing proves one breaks a supported deployment, but the following properties are non-negotiable:

- service runs as `tardigrade`, not root;
- PID path is deterministic;
- systemd creates/manages the `/run/tardigrade` runtime directory;
- `ExecReload` and `ExecStop` target the same PID file the running process writes;
- `ExecStartPre` validates config before binding listeners;
- `TimeoutStopSec` is greater than the default graceful drain timeout (`TARDIGRADE_SHUTDOWN_DRAIN_TIMEOUT_MS`, currently 30000 ms).

Do **not** blindly add aggressive directives such as `ProtectSystem=strict`, restrictive syscall filters, `PrivateDevices`, or broad capability removal unless they are tested against supported static roots, TLS files, Unix sockets, persisted state, and other documented deployment surfaces.

### 4. Preserve shared package ownership of the systemd unit

Do not create separate DEB and RPM unit files.

Both package paths should continue consuming:

```text
packaging/systemd/tardigrade.service
```

Update package smoke assertions as needed so both formats enforce the corrected unit contract.

### 5. Strengthen DEB/RPM smoke tests

Update `scripts/test-deb-package.sh` and `scripts/test-rpm-package.sh` to assert at minimum:

- `RuntimeDirectory=tardigrade` exists in the installed unit.
- `Environment=TARDIGRADE_PID_FILE=/run/tardigrade/tardigrade.pid` exists.
- `ExecReload` uses `--pid-file /run/tardigrade/tardigrade.pid`.
- `ExecStop` uses the same PID file.
- `ExecStartPre` validates `/etc/tardigrade/tardigrade.conf`.
- `TimeoutStopSec` exceeds the documented 30-second default drain timeout.
- Existing config/env ownership assertions remain green.

Where CI can run real systemd, add an actual lifecycle smoke:

1. install package;
2. start service;
3. assert PID file appears;
4. run status against the PID file;
5. reload;
6. assert the same service process remains alive;
7. stop;
8. assert clean exit / PID cleanup.

If nested systemd is not practical in the existing package test containers, keep the textual unit assertions there and add any runtime service test in an environment that can honestly execute it. Do not claim grep-only assertions prove lifecycle behavior.

### 6. Add a real root-level `Dockerfile`

There is currently a `.dockerignore` but no Dockerfile. Add one.

Requirements:

- multi-stage build;
- build with the repository-supported Zig toolchain version;
- release-mode build (`zig build -Doptimize=ReleaseFast` or equivalent established release build invocation);
- final runtime image contains only runtime dependencies + `tardi`, not the Zig compiler/source tree;
- include the OpenSSL runtime dependency required by the default/general TLS profile and CA certificates;
- create a dedicated non-root `tardigrade` user;
- create writable `/var/lib/tardigrade` and `/run/tardigrade` paths owned by that user;
- do not bake production configuration, TLS keys, or secrets into the image;
- expose an unprivileged default port (8069 is acceptable);
- set `TARDIGRADE_PID_FILE=/run/tardigrade/tardigrade.pid`;
- use exec-form `ENTRYPOINT` so `tardi` is PID 1 and receives SIGTERM/SIGHUP directly;
- default command should run `/etc/tardigrade/tardigrade.conf`.

Conceptual final stage:

```dockerfile
COPY --from=build /src/zig-out/bin/tardi /usr/local/bin/tardi

RUN mkdir -p /etc/tardigrade /var/lib/tardigrade /run/tardigrade \
    && chown -R tardigrade:tardigrade /var/lib/tardigrade /run/tardigrade

USER tardigrade
WORKDIR /var/lib/tardigrade

ENV TARDIGRADE_PID_FILE=/run/tardigrade/tardigrade.pid

EXPOSE 8069/tcp

ENTRYPOINT ["/usr/local/bin/tardi"]
CMD ["run", "-c", "/etc/tardigrade/tardigrade.conf"]
```

Prefer a pinned base version/digest consistent with existing CI reproducibility practices.

### 7. Add `compose.yaml`

Provide a copy/paste local production-style Docker example rather than forcing operators to reconstruct a long `docker run` invocation.

It should demonstrate:

- local image build;
- config mounted read-only;
- optional static root mounted read-only;
- TLS directory mounted read-only in the TLS example/path;
- `/run/tardigrade` supplied as tmpfs;
- non-root runtime inherited from the image;
- port mapping;
- restart policy;
- process health/status check;
- graceful stop window longer than the Tardigrade drain timeout.

Example shape:

```yaml
services:
  tardigrade:
    build: .
    ports:
      - "8069:8069"
    volumes:
      - ./deploy/tardigrade.conf:/etc/tardigrade/tardigrade.conf:ro
      - ./public:/var/lib/tardigrade/public:ro
    tmpfs:
      - /run/tardigrade
    environment:
      TARDIGRADE_PID_FILE: /run/tardigrade/tardigrade.pid
      TARDIGRADE_REQUIRE_UNPRIVILEGED_USER: "true"
    restart: unless-stopped
    stop_grace_period: 35s
    healthcheck:
      test:
        [
          "CMD",
          "/usr/local/bin/tardi",
          "status",
          "--pid-file",
          "/run/tardigrade/tardigrade.pid"
        ]
      interval: 30s
      timeout: 5s
      retries: 3
```

Adjust paths only where needed to match the final Dockerfile/example layout.

### 8. Document exact Docker workflows

`docs/DEPLOYMENT.md` should include copy/paste commands for:

Build:

```bash
docker compose build
```

Validate config before startup:

```bash
docker compose run --rm tardigrade \
  check /etc/tardigrade/tardigrade.conf
```

Start:

```bash
docker compose up -d
```

Inspect:

```bash
docker compose ps
docker compose logs -f tardigrade
```

Validate + reload:

```bash
docker compose exec tardigrade \
  tardi check /etc/tardigrade/tardigrade.conf

docker compose exec tardigrade \
  tardi reload --pid-file /run/tardigrade/tardigrade.pid
```

Graceful stop:

```bash
docker compose stop
```

Explain that because `tardi` is PID 1 in the container, Docker's SIGTERM reaches Tardigrade's graceful shutdown path directly.

### 9. Document systemd workflows

Include exact operator commands:

```bash
sudo -u tardigrade tardi check /etc/tardigrade/tardigrade.conf

sudo systemctl start tardigrade
sudo systemctl status tardigrade
sudo systemctl reload tardigrade
sudo systemctl restart tardigrade
sudo systemctl stop tardigrade

sudo journalctl -u tardigrade
sudo journalctl -u tardigrade -f
```

Clearly distinguish:

- reloadable config change -> validate then `systemctl reload tardigrade`;
- restart-required/process-owned change -> `systemctl restart tardigrade`;
- exact reload/restart boundaries live in `docs/RELOAD_SHUTDOWN.md`.

Do not duplicate the entire reload lifecycle documentation here.

### 10. Document ports and privilege model

The deployment guide must cover:

- default/example listener port 8069;
- use of an unprivileged port such as 8069/8443 when a frontend LB/NAT/reverse proxy owns 80/443;
- if Tardigrade binds 80/443 directly under systemd, prefer narrowly granting `CAP_NET_BIND_SERVICE` rather than running the process as root;
- Docker examples should remain non-root and use host port publishing rather than elevating the container solely to bind 80/443.

### 11. Document logs accurately

Cover both logging modes:

- service stdout/stderr via `journalctl -u tardigrade`;
- configured file/error logs where `error_log` / `TARDIGRADE_ERROR_LOG_PATH` is used.

The package already installs logrotate coverage for `/var/log/tardigrade/*.log` and signals USR1 after rotation; Tardigrade has a real SIGUSR1 log-reopen path. Document this rather than inventing a second rotation mechanism.

Cross-link `docs/OBSERVABILITY.md` for full logging/metrics behavior.

### 12. Add a production hardening checklist

At minimum cover:

- run as the dedicated non-root `tardigrade` user;
- validate config before start/reload;
- TLS private-key permissions;
- mount static/config/TLS content read-only where possible;
- expose only required ports;
- restrict `/status/metrics` to trusted scrapers/network paths or enable its supported auth mechanism;
- tune rate limits and connection/request limits for the workload;
- configure health checks for upstreams where appropriate;
- ensure `TimeoutStopSec` / container `stop_grace_period` exceeds the Tardigrade drain timeout;
- retain/review logs and reload-failure metrics;
- use firewall/LB controls around operator-only paths;
- avoid putting live secrets into source control or baking them into container layers;
- back up any operator-managed persistent state if deployment-specific features require it.

### 13. Add Docker smoke coverage

Add a script such as `scripts/test-docker-image.sh` and wire it into an existing Linux CI/packaging surface.

The test should prove actual behavior, not just Dockerfile syntax:

1. build the image from a clean checkout;
2. assert runtime UID is not 0;
3. run `tardi version`;
4. mount a minimal valid config;
5. start the container;
6. wait for `/health`;
7. request `/health` successfully;
8. verify the PID file exists in the container;
9. run `tardi check`;
10. send reload using the PID file;
11. assert the service remains alive;
12. stop with Docker's normal SIGTERM path;
13. assert graceful exit;
14. ensure the final runtime image does not contain Zig/compiler build tooling.

If Docker Compose is available in CI, also validate:

```bash
docker compose -f compose.yaml config --quiet
```

### 14. Update packaging status docs

`packaging/README.md` currently says Docker/OCI is not implemented.

After this issue lands, change that status to reflect reality, for example:

```text
Docker / OCI image | Local build supported, not published
```

Link the root `Dockerfile`, `compose.yaml`, and `docs/DEPLOYMENT.md`.

Do not imply an official registry image exists unless one is actually published.

### 15. Update README/documentation navigation

Update `README.md` so operators can discover the new guide easily:

- add `Production deployment` -> `docs/DEPLOYMENT.md` in the Full Documentation table;
- add a short Install-section pointer for production/service/container setup;
- optionally add Deployment to the top link bar if it remains readable.

Also update `examples/production-baseline/README.md` to clarify that it is the hardened **configuration** example and the full host/container service setup is documented in `docs/DEPLOYMENT.md`.

### 16. Docker limitations must be explicit

The deployment guide should state:

- Docker support is local-build support unless/until a Bare Systems image is published;
- Linux container deployment is the target of this ticket;
- operators supply config, certs, static roots, and secrets externally;
- `/status/metrics` should not automatically be exposed publicly;
- DEB/RPM remain the more deeply integrated host-native Linux install path;
- experimental protocol/features remain governed by `docs/SUPPORT_MATRIX.md`.

## Explicit non-goals

Do **not** expand this issue into:

- GHCR/Docker Hub publishing;
- multi-architecture OCI manifest publishing;
- registry authentication or retention policy;
- OCI signing/provenance release work;
- Kubernetes manifests;
- Helm charts;
- Podman-specific integration;
- rootless-networking redesign;
- new server/config features;
- ACME/certificate-management redesign;
- generic production-readiness claims.

Those should be separate distribution/operator tickets if desired.

## Updated definition of done

This issue is complete only when all of the following are true:

- [ ] `docs/DEPLOYMENT.md` exists and covers both systemd and Docker deployment.
- [ ] Deployment docs link to the canonical configuration, reload/shutdown, observability, support matrix, packaging, and production-baseline docs rather than duplicating them.
- [ ] Host filesystem layout is documented.
- [ ] Config/env/TLS/state permissions and ownership are documented.
- [ ] Port/privilege guidance explicitly avoids running Tardigrade as root solely to bind 80/443.
- [ ] The shared systemd unit has a deterministic PID-file contract.
- [ ] A fresh package install can use a consistent start/status/reload/stop path.
- [ ] `ExecStartPre` validates config before service start.
- [ ] systemd stop timeout exceeds the default graceful drain timeout.
- [ ] DEB and RPM smoke tests assert the corrected service-unit contract.
- [ ] Root-level multi-stage `Dockerfile` exists.
- [ ] Final Docker image runs as non-root.
- [ ] Final Docker image does not contain build tooling/source unnecessarily.
- [ ] Docker config/certs/static roots/secrets are externalized rather than baked into the image.
- [ ] `compose.yaml` provides a copy/paste runnable local deployment.
- [ ] Docker PID file, reload, health, and graceful-stop behavior are smoke-tested.
- [ ] `packaging/README.md` accurately reports Docker as local-build-supported / not published.
- [ ] README links to the deployment guide.
- [ ] `examples/production-baseline/README.md` links to the deployment guide.
- [ ] Production hardening checklist exists.
- [ ] Maturity/known-limitations language is explicit and does not overclaim production readiness.
- [ ] Markdown links are validated.
- [ ] New shell scripts pass shellcheck / syntax checks.
- [ ] `git diff --check` passes.
- [ ] Package smoke tests pass.
- [ ] Docker smoke test passes.
- [ ] Normal repository CI remains green.

Jaruso:
## Summary

- Adds `docs/DEPLOYMENT.md`, the canonical operator guide covering both host-native Linux/systemd and locally built Docker deployment: filesystem layout (with explicit package-install vs. service-start vs. operator-created distinctions), permissions (including numeric-GID guidance for bind-mounted TLS material), port/privilege guidance, logs (including Docker file-log ownership and rotation), and a production hardening checklist. Links out to `docs/CONFIGURATION.md`, `docs/RELOAD_SHUTDOWN.md`, `docs/OBSERVABILITY.md`, `docs/SUPPORT_MATRIX.md`, `packaging/README.md`, and `examples/production-baseline/` rather than duplicating them.
- Fixes the systemd PID/control-path gap called out in the issue: `packaging/systemd/tardigrade.service` had no *deterministic* PID-file contract. `EnvironmentFile=` values override unit-level `Environment=` values, so a `TARDIGRADE_PID_FILE` set in `tardigrade.env` (as the shipped production-baseline example originally did) could desync the running process's actual PID file from what `ExecReload`/`ExecStop` target. `ExecStart` now forces the path via `/usr/bin/env TARDIGRADE_PID_FILE=... tardi run ...` at exec time, uses `RuntimeDirectory=tardigrade` for `/run/tardigrade`, validates config in `ExecStartPre`, and raises `TimeoutStopSec` (35s) above the default 30s graceful-drain window. Covered by a regression test in `scripts/test-docker-image.sh` that reproduces the exact conflicting-env-file scenario.
- RPM packaging now has parity with DEB: installs the same SIGUSR1-reopen logrotate policy at `/etc/logrotate.d/tardigrade` and chowns `/var/log/tardigrade` to the `tardigrade` user in `%post` — previously neither existed, so `error_log` couldn't be used on a fresh RPM install.
- Adds a root `Dockerfile` (multi-stage: Zig toolchain only in the build stage, fixed non-root `tardigrade` UID/GID `10001:10001` — not a distro-allocated ID — no build tooling in the final image) and `compose.yaml` as a local-build-only Docker workflow — no image is published to a registry.
- Adds `scripts/test-docker-image.sh`, wired into the `packaging-smoke` CI job: builds the image from a clean checkout, asserts the exact pinned UID/GID, hits `/health`, checks the PID file, validates config, reloads, runs the conflicting-env-file PID regression, and confirms a graceful `SIGTERM` exit.
- Strengthens `scripts/test-rpm-package.sh`'s build isolation: it previously copied the host/CI runner's pre-built `zig-out/` into the Rocky Linux container before rebuilding, so a failing in-container build could still "succeed" by packaging the stale host binary. Added `rm -rf zig-out .zig-cache` + `test -x zig-out/bin/tardi` to make it honest — which immediately surfaced two real, previously-masked build bugs, both fixed in `build.zig`:
  - `configureSystemLibrarySearchPaths` only knew Debian's multiarch library layout (`/usr/lib/x86_64-linux-gnu`), never RHEL's flat `lib64` layout, so it never pointed the linker at Rocky's real OpenSSL location.
  - Rocky 9's glibc 2.34 (which merged `libpthread` into `libc` and re-versioned `pthread_create`/`pthread_join` under the new `GLIBC_2.34` symbol version) needs an explicit `-Dtarget=<arch>-linux-gnu.2.34` floor — Zig's default native-glibc assumption predates that and can't resolve those symbol versions otherwise, even though they're genuinely present at runtime. Also fixed a hardcoded `--arch x86_64` in the same script that only breaks on non-x86_64 dev hosts.
- Updates `packaging/README.md`: Docker status row from "Not implemented" to "Local build supported, not published"; the systemd-unit status row now accurately says "structurally validated" rather than "exercised end-to-end" (the smoke tests assert unit-file text/layout, not a booted systemd lifecycle); the `## Upgrading` section now recommends `systemctl restart` after a package/binary upgrade (`reload` republishes config for the *already-running* process — it doesn't exec a newly installed binary), reserving `reload` for reloadable `tardigrade.conf`-only edits. Links `README.md` / `examples/production-baseline/README.md` to the new guide.
- `.dockerignore`: excludes repo-local `.zig/`/`dist/` build artifacts (previously ~400MB of avoidable context on every packaging-smoke CI run) and the now-unnecessary `zig-out/` re-inclusion (the Dockerfile always does its own clean-source `zig build`), and narrows the `tests/vectors` re-inclusion to exactly the fixture files `build.zig` embeds into the production binary (`tests/vectors/pki/malformed-truncated.der`, `tests/vectors/pki/reduced/{manifest.zig,*.der}`) instead of the whole tree.

## Test plan

- [x] `docker build .` — succeeds; confirmed the fixed runtime UID:GID (`10001:10001`, not a distro-allocated ID) and that `zig` is absent from the final image.
- [x] `docker compose -f compose.yaml config --quiet` — valid.
- [x] Full container lifecycle exercised locally end-to-end via `scripts/test-docker-image.sh` (run directly, not just read for plausibility): build, exact UID/GID assertion, `/health` returns 200 (matching `Host` header — documented as an operator gotcha), PID file present, `tardi check`, `tardi reload --pid-file` (hot reload applied, same PID stays alive), the conflicting-env-file PID-path regression (a stray `TARDIGRADE_PID_FILE` injected the way `EnvironmentFile=` would still resolves to the deterministic `/run/tardigrade/tardigrade.pid`), and `docker stop` → graceful-shutdown log line, clean exit code 0.
- [x] `scripts/test-rpm-package.sh` run end-to-end locally against a real (non-cached) Rocky Linux 9 container: genuine `zig build` from clean source with the `-Dtarget=aarch64-linux-gnu.2.34` fix, RPM build, `dnf install`, every corrected systemd-unit-contract assertion (`RuntimeDirectory`, forced `ExecStart` PID env, `ExecReload`/`ExecStop` `--pid-file`, `ExecStartPre`, `TimeoutStopSec`), the new logrotate/log-ownership assertions, and `dnf remove` — passed.
- [x] Docker build context measured directly (via a throwaway `COPY . /ctx; find /ctx` Dockerfile) to confirm `.zig/`, `dist/`, and `zig-out/` are fully excluded and only the intended `tests/vectors/pki/*` fixture files are present.
- [x] `shellcheck` on all new/modified shell scripts — clean.
- [x] `zig fmt --check build.zig` and `git diff --check` — clean.
- [x] CI (`packaging-smoke` and the rest of the matrix) is green on the current head.
- [x] Verified every DEB/RPM smoke-test `grep -F` assertion byte-for-byte against the actual unit file content, and all relative Markdown links added/edited in this PR resolve to real files/anchors.

Closes #166

### DIFF
--- a/docs/DEPLOYMENT.md
+++ b/docs/DEPLOYMENT.md
@@ -1,0 +1,323 @@
+# Production Deployment
+
+This is the canonical operator guide for running Tardigrade outside a
+development shell: host-native Linux with systemd, or a locally built Docker
+image. It covers filesystem layout, permissions, process control, logs, and a
+hardening checklist for both paths.
+
+## Maturity boundary
+
+Tardigrade supports practical operator-managed deployments of its stable
+core. It is **not** claimed to be generic "production-ready", "battle-tested",
+or a drop-in nginx/Envoy replacement. The supported/stable runtime surface is
+defined by the [Core v1 support matrix](SUPPORT_MATRIX.md) — anything listed
+there as `experimental` stays experimental regardless of how you deploy it.
+
+Read this guide alongside:
+
+- [Configuration reference](CONFIGURATION.md) — every config directive and
+  environment variable.
+- [Reload, drain, and shutdown](RELOAD_SHUTDOWN.md) — the full hot-reload and
+  graceful-shutdown contract; this guide only covers the operator commands
+  that trigger it.
+- [Observability](OBSERVABILITY.md) — logging and metrics fields in detail.
+- [Support matrix](SUPPORT_MATRIX.md) — the stable/experimental contract.
+- [packaging/README.md](../packaging/README.md) — what's actually built and
+  published (DEB/RPM/archives) versus local-build-only (Docker).
+- [examples/production-baseline/](../examples/production-baseline/README.md)
+  — a hardened *configuration* example (TLS, rate limiting, headers). This
+  guide covers the *service/container* layer around that config.
+
+Two supported deployment paths:
+
+1. [Host-native Linux + systemd](#host-native-linux--systemd) — the more
+   deeply integrated path, via published DEB/RPM packages.
+2. [Docker, built locally from this repository](#docker) — a local-build-only
+   workflow; no image is published to a registry today.
+
+## Host filesystem layout
+
+Both deployment paths converge on the same layout. The DEB/RPM packages
+create this automatically; the Docker image creates the container-local
+equivalent under `/etc/tardigrade`, `/var/lib/tardigrade`, and
+`/run/tardigrade`.
+
+```text
+/usr/bin/tardi                         binary
+/etc/tardigrade/
+    tardigrade.conf                    routing/runtime configuration
+    tardigrade.env                     environment-only settings / secret references
+    tls/
+        fullchain.pem
+        privkey.pem
+/run/tardigrade/
+    tardigrade.pid                     ephemeral process PID
+/var/lib/tardigrade/                   service working/state directory
+/var/lib/tardigrade/public/            optional static root
+/var/log/tardigrade/                   optional file logs
+```
+
+### Permissions and ownership
+
+The package `postinst`/`%post` scripts are the source of truth for this on a
+host-native install; the same properties apply inside the Docker image.
+
+| Path | Owner | Mode | Notes |
+| --- | --- | --- | --- |
+| `/etc/tardigrade/tardigrade.conf` | `root:root` | `0644` | Not secret; readable by the service via world-read. |
+| `/etc/tardigrade/tardigrade.env` | `root:tardigrade` | `0640` | May hold secret references; not world-readable. |
+| `/etc/tardigrade/tls/privkey.pem` | `root:tardigrade` (or narrower) | `0640` or tighter | Readable only by the service account/group. Never world-readable. |
+| `/run/tardigrade/` | `tardigrade:tardigrade` | `0750` | Ephemeral; recreated at start by systemd's `RuntimeDirectory=` or the container's tmpfs mount. |
+| `/var/lib/tardigrade/` | `tardigrade:tardigrade` | `0755` | Writable service state directory. |
+| `/var/lib/tardigrade/public/` | read-only to the service where possible | — | Static roots and mounted config don't need write access; mount `:ro`. |
+
+The service always runs as the dedicated non-root `tardigrade` user, both
+under systemd and in the container.
+
+If your config references additional roots, Unix sockets, or upstream
+sockets outside this layout, they need their own filesystem access —
+extend the systemd unit's directory allowlist or the container's mounts
+deliberately for that specific path rather than weakening sandboxing
+globally (for example, don't drop `ProtectSystem=full` or `ProtectHome=true`
+just to reach one extra path).
+
+## Host-native Linux + systemd
+
+Install via the published DEB or RPM package — see
+[packaging/README.md](../packaging/README.md) for exact install commands and
+current publish status. The package installs the binary, a starter config,
+the systemd unit, and creates the `tardigrade` system user and
+`/var/lib/tardigrade`.
+
+### The systemd unit's PID/control-path contract
+
+The shared unit at
+[`packaging/systemd/tardigrade.service`](../packaging/systemd/tardigrade.service)
+gives every fresh install a deterministic, self-contained control path:
+
+```ini
+Environment=TARDIGRADE_PID_FILE=/run/tardigrade/tardigrade.pid
+RuntimeDirectory=tardigrade
+RuntimeDirectoryMode=0750
+
+ExecStartPre=/usr/bin/tardi check /etc/tardigrade/tardigrade.conf
+ExecStart=/usr/bin/tardi run -c /etc/tardigrade/tardigrade.conf
+ExecReload=/usr/bin/tardi reload --pid-file /run/tardigrade/tardigrade.pid
+ExecStop=/usr/bin/tardi stop --pid-file /run/tardigrade/tardigrade.pid
+
+TimeoutStopSec=35s
+```
+
+- `RuntimeDirectory=tardigrade` makes systemd create and own
+  `/run/tardigrade` (mode `0750`, owned by the `tardigrade` service user) on
+  every start — no manual directory setup needed.
+- `Environment=TARDIGRADE_PID_FILE=...` and the `--pid-file` flags on
+  `ExecReload`/`ExecStop` all point at the same path the running process
+  writes, so `systemctl reload`/`stop` can always find it.
+- `ExecStartPre` runs `tardi check` against the config before the listener
+  binds, so a bad edit fails the start instead of taking down a working
+  process.
+- `TimeoutStopSec=35s` is set above the default 30s graceful-drain window
+  (`TARDIGRADE_SHUTDOWN_DRAIN_TIMEOUT_MS`, see
+  [RELOAD_SHUTDOWN.md](RELOAD_SHUTDOWN.md)) so systemd doesn't escalate to
+  `SIGKILL` before the drain finishes. If you raise the drain timeout,
+  raise `TimeoutStopSec` to match.
+
+Adjust the built-in hardening directives (`NoNewPrivileges`, `PrivateTmp`,
+`ProtectHome`, `ProtectSystem=full`, `UMask=0027`) if testing shows one
+breaks a deployment you rely on (extra static roots, Unix sockets, etc.), but
+keep the service non-root, the PID path deterministic, and `ExecStartPre`
+config validation.
+
+### Commands
+
+```bash
+# Validate a config edit before publishing it
+sudo -u tardigrade tardi check /etc/tardigrade/tardigrade.conf
+
+sudo systemctl start tardigrade
+sudo systemctl status tardigrade
+sudo systemctl reload tardigrade    # hot reload: SIGHUP, no dropped connections
+sudo systemctl restart tardigrade   # full restart: needed for process-owned settings
+sudo systemctl stop tardigrade
+
+sudo journalctl -u tardigrade
+sudo journalctl -u tardigrade -f
+```
+
+Reload vs. restart:
+
+- A config change that only affects request routing, TLS credentials, or
+  other reload-owned settings: validate, then `systemctl reload tardigrade`.
+- A change to listener shard topology, HTTP/3 listener-owned settings, or
+  other process-owned settings: `systemctl restart tardigrade` (reload
+  rejects these; see [RELOAD_SHUTDOWN.md](RELOAD_SHUTDOWN.md) for the exact
+  boundary).
+
+## Docker
+
+Docker support here is **local-build support**: `docker build`/`docker
+compose build` from the [`Dockerfile`](../Dockerfile) in this repository.
+There is no published Bare Systems image on any registry today — see
+[packaging/README.md](../packaging/README.md#current-status).
+
+The image is a multi-stage build: a build stage compiles `tardi` with the
+pinned Zig toolchain (matching `.github/workflows/ci.yml`) and the default
+"general" TLS profile (OpenSSL adapter — see
+[TLS_DEPENDENCY_POLICY.md](TLS_DEPENDENCY_POLICY.md)); the runtime stage
+contains only `tardi`, its OpenSSL/CA-certificate runtime dependencies, and a
+non-root `tardigrade` user. No Zig toolchain or source tree ships in the
+final image.
+
+### Build
+
+```bash
+docker compose build
+```
+
+### Validate config before startup
+
+```bash
+docker compose run --rm tardigrade \
+  check /etc/tardigrade/tardigrade.conf
+```
+
+### Start
+
+```bash
+docker compose up -d
+```
+
+### Inspect
+
+```bash
+docker compose ps
+docker compose logs -f tardigrade
+```
+
+### Validate + reload
+
+```bash
+docker compose exec tardigrade \
+  tardi check /etc/tardigrade/tardigrade.conf
+
+docker compose exec tardigrade \
+  tardi reload --pid-file /run/tardigrade/tardigrade.pid
+```
+
+### Graceful stop
+
+```bash
+docker compose stop
+```
+
+`tardi` runs as PID 1 in the container (exec-form `ENTRYPOINT`), so Docker's
+`SIGTERM` reaches Tardigrade's graceful shutdown path directly — no init
+shim swallowing or relaying the signal.
+
+### compose.yaml
+
+[`compose.yaml`](../compose.yaml) at the repo root is a copy/paste-runnable
+local example: builds the image, mounts a read-only config (and optional
+static root / TLS directory), supplies `/run/tardigrade` as a `tmpfs` mount
+owned by the container's non-root user, maps the port, and sets a graceful
+`stop_grace_period` longer than the drain timeout. The `tmpfs` entry needs
+explicit `uid`/`gid` mount options matching the image's `tardigrade` user —
+a plain tmpfs mount defaults to `root:root` and the process can't write its
+PID file into it.
+
+The bundled healthcheck runs `tardi status --pid-file ...` inside the
+container. Note its limitation: `tardi status` always exits `0` once the
+config loads, whether or not the process is actually reachable — it reports
+`status: stopped` in text rather than failing the check. It's still useful
+signal (it proves the entrypoint, working directory, and PID path are wired
+correctly), but it will not, by itself, flip a container unhealthy if the
+listener stops responding while the process is still alive. If you need a
+liveness check on the actual HTTP path, add `curl` or `wget` to the runtime
+image and point it at `/health` instead.
+
+## Ports and privilege model
+
+- The example/default listener port is `8069` (unprivileged, works for both
+  systemd and Docker without extra capabilities).
+- Put a frontend load balancer, NAT, or reverse proxy in front to expose
+  `80`/`443` rather than binding them directly, where that fits your
+  environment.
+- If Tardigrade must bind `80`/`443` directly under systemd, grant
+  `AmbientCapabilities=CAP_NET_BIND_SERVICE` (plus
+  `CapabilityBoundingSet=CAP_NET_BIND_SERVICE`) on the service unit instead
+  of running the process as root.
+- Keep Docker deployments non-root and use host port publishing
+  (`ports: ["80:8069"]`, etc.) rather than elevating the container just to
+  bind a low port.
+
+## Logs
+
+Two logging surfaces, both covered in full by
+[OBSERVABILITY.md](OBSERVABILITY.md):
+
+- **Container/service stdout** — under systemd, `journalctl -u tardigrade`;
+  under Docker, `docker compose logs -f tardigrade`. This is where runtime
+  and JSON access logs land unless `error_log`/`TARDIGRADE_ERROR_LOG_PATH`
+  is configured.
+- **Configured file logs** — set `error_log` (or
+  `TARDIGRADE_ERROR_LOG_PATH`) to write to `/var/log/tardigrade/*.log`
+  instead. The DEB/RPM packages already install a logrotate config at
+  `/etc/logrotate.d/tardigrade` that signals `SIGUSR1` after rotation;
+  Tardigrade has a real SIGUSR1 log-reopen handler, so log files reopen
+  cleanly on the existing rotation cadence. Don't invent a second rotation
+  mechanism.
+
+> **Operator note discovered while writing this guide**: the starter config
+> matches on `server_name localhost;`. A bare `curl http://<host>:8069/health`
+> against an IP or a mismatched `Host` header returns `404`, not `200` —
+> match the `Host` header (or your config's `server_name`) when
+> smoke-testing, or use `server_name _;`/a real hostname in your own config.
+
+## Production hardening checklist
+
+- [ ] Run as the dedicated non-root `tardigrade` user (default for both
+      paths — don't override it).
+- [ ] Validate config before every start/reload (`tardi check`; the systemd
+      unit already does this via `ExecStartPre`).
+- [ ] TLS private key permissions: never world-readable; restrict to the
+      service account/group.
+- [ ] Mount static roots, config, and TLS material read-only where the
+      service doesn't need to write them.
+- [ ] Expose only the ports you intend to serve.
+- [ ] Restrict `/status/metrics` to trusted scrapers/network paths, or set
+      `TARDIGRADE_METRICS_REQUIRE_AUTH=true` (see
+      [CONFIGURATION.md](CONFIGURATION.md)).
+- [ ] Tune rate limits and connection/request limits for your workload
+      (`TARDIGRADE_RATE_LIMIT_RPS`, `TARDIGRADE_MAX_ACTIVE_CONNECTIONS`,
+      `TARDIGRADE_MAX_IN_FLIGHT_REQUESTS`).
+- [ ] Configure upstream health checks where you proxy to origins.
+- [ ] Set `TimeoutStopSec` (systemd) / `stop_grace_period` (Docker) above
+      `TARDIGRADE_SHUTDOWN_DRAIN_TIMEOUT_MS`.
+- [ ] Retain and review logs and reload-failure metrics
+      (`tardigrade_reload_failure_total`, `/tardigrade/reload/status`).
+- [ ] Use firewall/load-balancer controls around operator-only paths.
+- [ ] Never commit live secrets or bake them into container image layers —
+      externalize config, certs, and secrets.
+- [ ] Back up any operator-managed persistent state your specific deployment
+      relies on.
+
+See [examples/production-baseline/](../examples/production-baseline/README.md)
+for a config-level pre-production checklist covering TLS, HSTS, and rate
+limiting values.
+
+## Known limitations
+
+- Docker support here is local-build only. No image is published to any
+  registry unless and until a separate distribution ticket adds that.
+- This guide targets Linux containers; other container runtimes (Podman,
+  etc.) aren't tested against it.
+- Operators supply their own config, certificates, static roots, and secrets
+  — none of these are baked into the image or package.
+- `/status/metrics` is not automatically restricted; apply the checklist
+  item above before exposing a deployment publicly.
+- DEB/RPM remain the more deeply integrated host-native Linux install path
+  (user creation, logrotate, systemd wiring all handled by the package).
+- Every experimental protocol/feature in the [support matrix](SUPPORT_MATRIX.md)
+  stays experimental regardless of deployment method — deploying via
+  Docker or systemd does not change a feature's maturity level.
--- a/docs/DEPLOYMENT.md
+++ b/docs/DEPLOYMENT.md
@@ -37,30 +37,40 @@ Two supported deployment paths:
 
 ## Host filesystem layout
 
-The two paths use the same *directory* layout, but they're populated
-differently: the DEB/RPM package creates every path below automatically
-(including `tardigrade.env`); the Docker image only creates the directories
-themselves (`/etc/tardigrade`, `/var/lib/tardigrade`, `/run/tardigrade`,
-`/var/log/tardigrade`) — you supply `tardigrade.conf`, `tardigrade.env`
-(if you use one), and any TLS material as bind mounts. See
-[Docker](#docker) below for the exact mount shape and the one path that
-differs: the packages install the binary at `/usr/bin/tardi`, the image
-installs it at `/usr/local/bin/tardi`.
+The two paths converge on the same *directory* layout, but different pieces
+of it come from different places and at different times — not everything
+below is created by the package install step. Three categories:
+
+- **Created by the package install** (DEB/RPM `postinst`/`%post`; the Docker
+  image creates the equivalent directories at build time): the binary,
+  `/etc/tardigrade/tardigrade.conf`, `/etc/tardigrade/tardigrade.env`,
+  `/var/lib/tardigrade`, `/var/log/tardigrade`.
+- **Created at service start, not install**: `/run/tardigrade` is created by
+  systemd's `RuntimeDirectory=` when the unit starts (see
+  [below](#the-systemd-units-pidcontrol-path-contract)), not by the package.
+  Under Docker it's a `tmpfs` mount you supply in `compose.yaml`.
+- **Operator-created, only if you use them**: neither package nor image
+  creates `/etc/tardigrade/tls/` or `/var/lib/tardigrade/public/` — you
+  create these (or bind-mount them, under Docker) only if your config
+  actually references TLS material or a static root.
 
 ```text
-/usr/bin/tardi                         binary (packages) — /usr/local/bin/tardi in the Docker image
-/etc/tardigrade/
-    tardigrade.conf                    routing/runtime configuration
-    tardigrade.env                     environment-only settings / secret references
-    tls/
-        fullchain.pem
-        privkey.pem
-/run/tardigrade/
-    tardigrade.pid                     ephemeral process PID
-/var/lib/tardigrade/                   service working/state directory
-/var/lib/tardigrade/public/            optional static root
-/var/log/tardigrade/                   optional file logs
+Package install:              /usr/bin/tardi                     binary (/usr/local/bin/tardi in the Docker image)
+                               /etc/tardigrade/tardigrade.conf    routing/runtime configuration
+                               /etc/tardigrade/tardigrade.env     environment-only settings / secret references
+                               /var/lib/tardigrade/                service working/state directory
+                               /var/log/tardigrade/                 file-log directory
+Service start (systemd RuntimeDirectory= / Docker tmpfs mount):
+                               /run/tardigrade/tardigrade.pid      ephemeral process PID
+Operator-created when used:   /etc/tardigrade/tls/fullchain.pem
+                               /etc/tardigrade/tls/privkey.pem
+                               /var/lib/tardigrade/public/          optional static root
 ```
+
+Under Docker, the image only creates the directories themselves — you
+supply `tardigrade.conf`, `tardigrade.env` (if you use one), and any TLS
+material as bind mounts. See [Docker](#docker) below for the exact mount
+shape.
 
 ### Permissions and ownership
 
@@ -92,9 +102,28 @@ owns the file — group and user *names* don't cross the container boundary,
 only the numeric IDs do. The image's `tardigrade` group is a fixed
 `10001`; if your host's `tardigrade` group (created by the DEB/RPM package
 via `useradd --system`, so its GID is whatever the distro allocated) isn't
-also `10001`, either `chgrp`/`chmod` the mounted file for the container's
-GID or mount it world-readable-within-the-container (`0644`) if that fits
-your threat model — never make it world-readable on the host.
+also `10001`, you cannot fix this with `chmod 0644` — a bind mount exposes
+the *same host inode* inside the container, so `chmod` changes the mode on
+the host file too and makes the key world-readable there, not just "within
+the container." That fails the never-world-readable requirement outright.
+
+Instead, give the container access to the correct numeric group without
+loosening the host file's mode. Create (or reuse) a host group whose GID
+owns the TLS key at `0640`, and add the container to that group with
+Compose's `group_add`:
+
+```yaml
+services:
+  tardigrade:
+    group_add:
+      - "${TARDIGRADE_TLS_GID}"   # numeric GID of the host group that owns the key
+```
+
+The key stays `root:<that-group> 0640` on the host; the container gains
+read access via the matching numeric supplementary group rather than
+through a mode change. An alternative is staging a copy of the key into a
+volume you initialize with owner/group `10001` directly, if you'd rather
+not add a supplementary group.
 
 If your config references additional roots, Unix sockets, or upstream
 sockets outside this layout, they need their own filesystem access —
@@ -327,12 +356,33 @@ Two logging surfaces, both covered in full by
   `/etc/logrotate.d/tardigrade`, which signals `SIGUSR1` after rotation;
   Tardigrade has a real SIGUSR1 log-reopen handler, so log files reopen
   cleanly on the existing rotation cadence. Don't invent a second rotation
-  mechanism. The Docker image creates the same `tardigrade`-owned
-  `/var/log/tardigrade` directory, but it isn't persistent or rotated by
-  anything inside the container — mount a host path or named volume over
-  it if you use `error_log` under Docker (`- ./logs:/var/log/tardigrade`
-  in `compose.yaml`), and handle rotation on the host side. Without that
-  mount, stdout/stderr (`docker compose logs`) is the durable log surface.
+  mechanism.
+
+  The Docker image creates the same `tardigrade`-owned `/var/log/tardigrade`
+  directory, but nothing inside the container persists or rotates it. Using
+  `error_log` under Docker needs two things, not just a mount:
+
+  1. **A pre-owned host path or named volume.** A bind mount such as
+     `- ./logs:/var/log/tardigrade` *replaces* the image's `10001:10001`
+     directory with whatever is at `./logs` on the host. A freshly created
+     (often root-owned) host directory will make `error_log` fail with
+     `EACCES`. Prepare it first:
+     ```bash
+     install -d -o 10001 -g 10001 -m 0755 ./logs
+     ```
+     or use a named volume and let the container's own directory-creation
+     step own it on first start.
+  2. **Rotation that reopens the file, not just renames it.** The DEB/RPM
+     logrotate policy sends `SIGUSR1` for exactly this reason — Tardigrade
+     keeps writing to the old (now-renamed) inode otherwise. A host-side
+     `logrotate` `postrotate` needs the equivalent signal delivered to the
+     container, e.g. `docker kill --signal=USR1 <container>` (or the
+     `docker compose` equivalent), not just a `copytruncate`/rename step.
+
+  Without a persistent mount, `docker compose logs` on stdout/stderr is the
+  default log surface — but don't treat it as inherently durable either;
+  retention depends on the configured Docker logging driver and container
+  lifecycle, the same as any other container's stdout.
 
 > **Operator note discovered while writing this guide**: the starter config
 > matches on `server_name localhost;`. A bare `curl http://<host>:8069/health`
@@ -344,8 +394,11 @@ Two logging surfaces, both covered in full by
 
 - [ ] Run as the dedicated non-root `tardigrade` user (default for both
       paths — don't override it).
-- [ ] Validate config before every start/reload (`tardi check`; the systemd
-      unit already does this via `ExecStartPre`).
+- [ ] Validate `tardigrade.conf` edits with `tardi check` before reload when
+      practical — `ExecStartPre` does not run on `systemctl reload`, only on
+      `start`/`restart`; hot reload runs its own validation and leaves the
+      previously published config active if the candidate is rejected (see
+      [RELOAD_SHUTDOWN.md](RELOAD_SHUTDOWN.md)).
 - [ ] TLS private key permissions: never world-readable; restrict to the
       service account/group.
 - [ ] Mount static roots, config, and TLS material read-only where the
--- a/docs/DEPLOYMENT.md
+++ b/docs/DEPLOYMENT.md
@@ -41,10 +41,14 @@ The two paths converge on the same *directory* layout, but different pieces
 of it come from different places and at different times — not everything
 below is created by the package install step. Three categories:
 
-- **Created by the package install** (DEB/RPM `postinst`/`%post`; the Docker
-  image creates the equivalent directories at build time): the binary,
-  `/etc/tardigrade/tardigrade.conf`, `/etc/tardigrade/tardigrade.env`,
-  `/var/lib/tardigrade`, `/var/log/tardigrade`.
+- **Created by the DEB/RPM package install** (`postinst`/`%post`): the
+  binary, a starter `/etc/tardigrade/tardigrade.conf`, an
+  `/etc/tardigrade/tardigrade.env` template, `/var/lib/tardigrade`,
+  `/var/log/tardigrade`. The Docker image creates only the *directories*
+  (binary, `/var/lib/tardigrade`, `/var/log/tardigrade`) at build time — it
+  does **not** create a `tardigrade.conf` or `tardigrade.env`; see
+  [Docker](#docker) below for how config and environment values reach the
+  container instead.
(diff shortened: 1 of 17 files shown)


Jaruso (commented):
**Changes needed before merge.** GitHub will not let this account submit `REQUEST_CHANGES` on its own PR, so this is a COMMENT review with blocking findings.

This is a strong implementation overall and it covers most of #166's expanded scope: canonical deployment docs, README wiring, a real multi-stage Docker image, Compose, non-root execution, Docker lifecycle smoke coverage, systemd hardening, deterministic-looking control commands, package assertions, and a clear maturity disclaimer. The current CI run is green.

I do not think this should merge yet because several operator-facing contracts are still incorrect or incomplete, and these are exactly the areas #166 is meant to make trustworthy.

### Blocking acceptance gaps

1. **The systemd PID path is not actually deterministic yet.** `EnvironmentFile=` values override unit-level `Environment=` values. The repository's own production-baseline env currently sets `TARDIGRADE_PID_FILE=/tmp/tardigrade.pid`, so following the linked example makes the running process write `/tmp/tardigrade.pid` while `ExecReload`/`ExecStop` target `/run/tardigrade/tardigrade.pid`. Force the runtime PID value at `ExecStart` command level (or add an explicit run-time CLI PID override), update the baseline example, and cover the conflicting-env case in tests.

2. **The documented systemd validation command does not validate the same effective configuration that the service uses.** `sudo -u tardigrade tardi check ...` does not load `/etc/tardigrade/tardigrade.env`, while the service does and process environment has highest precedence. The guide also needs to state that edits to `tardigrade.env` require a restart; a SIGHUP cannot mutate the running process environment.

3. **RPM file logging is currently broken relative to the new docs.** The RPM does not install `/etc/logrotate.d/tardigrade`, and `/var/log/tardigrade` is packaged root-owned while the service runs as `tardigrade`. A normal RPM install therefore cannot create the documented `/var/log/tardigrade/*.log` files. Prefer making RPM behavior match DEB: install the logrotate policy, make the log directory writable by `tardigrade`, and assert both in `test-rpm-package.sh`.

4. **The Docker filesystem/permission story is not yet internally consistent.** The image uses `/usr/local/bin/tardi`, does not create `tardigrade.env`, and does not create/chown `/var/log/tardigrade`, while the guide says both paths converge on the same layout. Either create the writable container log path and document how to persist it, or explicitly make the container path stdout/stderr-first and document what an operator must mount for file logging.

5. **The container's numeric UID/GID is an implicit implementation detail but Compose treats `999:999` as a stable API.** Pin the image UID/GID explicitly and use the same values in Compose/tests. Also document that bind-mounted TLS key permissions are numeric: a host file that is `root:tardigrade 0640` is not automatically readable by a container group merely because it has the same name.

6. **The deployment guide omits the existing master/worker reload limitation.** `docs/CONFIGURATION.md` already says master/worker mode does not provide coherent PID-file/SIGHUP reload control across all workers. The systemd section/known-limitations section should say that the packaged `systemctl reload` workflow is for single-process mode today.

### Test reliability issue found while checking CI

The successful packaging job contains a concerning false-positive path in the existing RPM smoke test: inside Rocky Linux, `zig build -Doptimize=ReleaseFast` prints a compile failure (`unable to open library directory '/usr/lib/x86_64-linux-gnu'` / `'/lib/x86_64-linux-gnu'`), but the script continues and packages a binary anyway. Because the test does `cp -a /repo /tmp/tardigrade`, it also copies the CI runner's pre-existing `zig-out/`, so a stale host-built binary can satisfy `/tmp/tardigrade/zig-out/bin/tardi` even when the intended Rocky build did not produce one.

While this PR is strengthening `test-rpm-package.sh`, please make that build isolation real too. At minimum:

```bash
cp -a /repo /tmp/tardigrade
cd /tmp/tardigrade
rm -rf zig-out .zig-cache
test -x zig-out/bin/tardi
```

If that exposes a real Rocky build problem, fix the build environment rather than allowing the smoke test to package the runner artifact. Otherwise the RPM smoke result does not prove what its header says it proves.

### What already looks good

- README and cross-document wiring are good.
- The maturity language appropriately avoids claiming generic production readiness.
- Docker is correctly described as local-build-only rather than a published image.
- The Docker lifecycle smoke genuinely exercises start, HTTP health, PID creation, reload, and graceful SIGTERM shutdown.
- `RuntimeDirectory=`, non-root systemd execution, `ExecStartPre`, conservative sandboxing, and the 35-second stop window are directionally correct.
- The DEB package assertions cover the new unit shape correctly once the PID override semantics above are fixed.

After the inline items and RPM smoke isolation are fixed, #166 should be in good shape for another focused verification pass.

Jaruso (commented):
Second-pass review — **changes still needed before merge**.

The prior implementation blockers are substantially addressed. I verified the new head against #166 and the previous review, and I checked the successful CI run, including the Packaging smoke logs. In particular: the systemd PID path is now forced at exec time; the Rocky RPM smoke now deletes stale artifacts and genuinely rebuilds against glibc 2.34; RPM logrotate and `/var/log/tardigrade` ownership are present; Docker UID/GID is pinned and tested; the master/worker reload limitation and env-file restart semantics are documented; and the Docker build context dropped from ~399 MB to ~39 MB.

I still found four operator-facing defects in `docs/DEPLOYMENT.md`, called out inline below. Please fix all four.

There are also two consistency items in `packaging/README.md` that need correction before #166 should close:

1. The status table still says the systemd unit is “exercised end-to-end by both the DEB and RPM smoke tests.” Those tests do not boot systemd or exercise start/status/reload/stop; they install the package and assert unit text/layout. The issue guidance explicitly says not to present grep-only assertions as lifecycle proof. Either change this to something like “installed and structurally validated by both package smoke tests” or add a real systemd-capable lifecycle test.

2. The `## Upgrading` workflow is now inconsistent with the deployment guide. It still shows standalone `tardi check` followed by `systemctl reload`. That check does not load `/etc/tardigrade/tardigrade.env`, env-file changes require restart, and—most importantly for an *upgrade* section—`systemctl reload` does not exec a newly installed `tardi` binary. After a package/binary upgrade, `systemctl restart tardigrade` should be the default. Reserve reload for reloadable `tardigrade.conf` edits where neither the binary nor environment changed, and link to `docs/DEPLOYMENT.md` / `docs/RELOAD_SHUTDOWN.md` for the boundary.

One non-blocking cleanup is also worth doing while here: `.dockerignore` now says repo-local build artifacts should not enter the build context, but it still explicitly re-includes `zig-out/bin/tardi`. The Dockerfile always performs its own `zig build`, so the `zig-out/bin` exceptions are unnecessary and mean the CI Docker build still receives the host-built binary from the earlier Packaging-smoke step. Ignoring all of `zig-out/` would make the clean-source-build contract stronger and shrink the context further.

Finally, the PR body is stale after the fixes: its test plan still says runtime UID `999` even though the image contract is now `10001:10001`, and it still says the DEB/RPM tests are waiting for CI even though the current CI run completed successfully. Please update that metadata so the merged PR accurately records what was tested.

Once the inline issues plus the two packaging-doc inconsistencies are fixed, I do not currently see another implementation blocker against #166.

docs/DEPLOYMENT.md
```diff
@@ -0,0 +1,446 @@
...
+                               /var/log/tardigrade/                 file-log directory
+Service start (systemd RuntimeDirectory= / Docker tmpfs mount):
+                               /run/tardigrade/tardigrade.pid      ephemeral process PID
+Operator-created when used:   /etc/tardigrade/tls/fullchain.pem
+                               /etc/tardigrade/tls/privkey.pem
+                               /var/lib/tardigrade/public/          optional static root
+```
+
+Under Docker, the image only creates the directories themselves — you
+supply `tardigrade.conf`, `tardigrade.env` (if you use one), and any TLS
```
Jaruso: **Blocking: `tardigrade.env` is not an application-readable bind-mounted config file under Docker.** Tardigrade's loader consumes `TARDIGRADE_*` values from the **process environment**; it does not automatically parse `/etc/tardigrade/tardigrade.env`. The existing production-baseline instructions prove the intended model by `source`ing the file before launch. As written, an operator who follows this sentence and bind-mounts `tardigrade.env` will silently get none of those overrides/secrets.

Please separate the native-package and container contracts:

- DEB/RPM install `/etc/tardigrade/tardigrade.env`, and systemd consumes it through `EnvironmentFile=`.
- The Docker image creates `/etc/tardigrade/` but does **not** create/read a `tardigrade.env` file itself.
- Docker operators should inject the file through Compose `env_file:` (or explicit `environment:`), not bind-mount it as application config.

For example, `compose.yaml` can demonstrate:

```yaml
services:
  tardigrade:
    # Optional operator environment values:
    # env_file:
    #   - ./deploy/tardigrade.env
    environment:
      # Keep these explicit: Compose `environment:` overrides `env_file:`, so
      # even a baseline env file containing another PID path cannot desync the
      # runtime/control contract.
      TARDIGRADE_PID_FILE: /run/tardigrade/tardigrade.pid
      TARDIGRADE_REQUIRE_UNPRIVILEGED_USER: "true"
```

Also fix the earlier bullet saying the Docker image creates the package's “equivalent” `tardigrade.conf`/`tardigrade.env`; the Dockerfile creates the directories and binary, while the config is mounted and environment values are injected.

docs/DEPLOYMENT.md
```diff
@@ -0,0 +1,446 @@
...
+  registry unless and until a separate distribution ticket adds that.
+- This guide targets Linux containers; other container runtimes (Podman,
+  etc.) aren't tested against it.
+- Operators supply their own config, certificates, static roots, and secrets
+  — none of these are baked into the image or package.
+- `/status/metrics` is not automatically restricted; apply the checklist
+  item above before exposing a deployment publicly.
+- DEB/RPM remain the more deeply integrated host-native Linux install path
+  (user creation, logrotate, systemd wiring all handled by the package).
+- The packaged `systemctl reload`/PID-file control workflow targets
```
Jaruso: **Blocking: the master-process limitation is broader than `systemctl reload`.** The current text scopes only the packaged reload workflow, but the same implementation creates three deployment-visible problems:

1. **Docker reload is affected too.** The Docker section runs `tardi reload --pid-file ...`, i.e. the same PID-file/SIGHUP mechanism already documented as incoherent when `master_process true;`.
2. **USR1 log rotation is not coherent across workers.** `runPlan` calls `configureErrorLog` before the master/worker branch, so each worker owns its own stderr/error-log descriptor. The package logrotate policy sends USR1 with `--kill-who=main`, and Docker's `docker kill --signal=USR1 <container>` reaches PID 1. `runMaster` consumes that flag and calls `reopenErrorLog(cfg)` only in the master; it never forwards USR1 to `children`, while workers' gateway loops have their own reopen handler. Those worker FDs therefore keep writing the rotated inode.
3. **The 35s stop window is not generally sufficient in multi-worker mode.** `runMaster` currently loops over children and calls `kill(...)` + `wait(...)` for one child before proceeding to the next. With a 30s worker drain window, multiple workers can serialize shutdown beyond the documented 35s supervisor/container grace period.

For this deployment/docs ticket, the minimal safe fix is to explicitly scope the PID/reload, file-log reopen, and 35s graceful-stop examples to `master_process false` and list master mode as a broader known limitation. If you want master mode supported by these examples, fix signal fan-out instead: signal **all** worker PIDs first, then wait for all of them, and propagate USR1 to every worker (or otherwise reopen every worker's descriptor), with integration coverage. For example, the shutdown shape should be conceptually:

```zig
for (children) |child| {
    if (child.id) |pid| std.posix.kill(pid, std.posix.SIG.TERM) catch {};
}
for (&children) |*child| {
    _ = child.wait(compat.io()) catch {};
}
```

and USR1 needs analogous fan-out. Please also put the single-process warning immediately beside the Docker reload command if the runtime remains unchanged.

docs/DEPLOYMENT.md
```diff
@@ -0,0 +1,446 @@
...
+  directory, but nothing inside the container persists or rotates it. Using
+  `error_log` under Docker needs two things, not just a mount:
+
+  1. **A pre-owned host path or named volume.** A bind mount such as
+     `- ./logs:/var/log/tardigrade` *replaces* the image's `10001:10001`
+     directory with whatever is at `./logs` on the host. A freshly created
+     (often root-owned) host directory will make `error_log` fail with
+     `EACCES`. Prepare it first:
+     ```bash
+     install -d -o 10001 -g 10001 -m 0755 ./logs
```
Jaruso: This copy/paste command normally needs elevated privilege: an unprivileged operator cannot `chown` a host directory to arbitrary UID/GID `10001:10001`. Please make the setup example runnable as written, e.g.

```bash
sudo install -d -o 10001 -g 10001 -m 0755 ./logs
```

or show `mkdir -p` followed by `sudo chown 10001:10001 ./logs`. The ownership guidance itself is now correct; this is just the final operational step needed for the documented bind mount to work.

docs/DEPLOYMENT.md
```diff
@@ -0,0 +1,446 @@
...
+limiting values.
+
+## Known limitations
+
+- Docker support here is local-build only. No image is published to any
+  registry unless and until a separate distribution ticket adds that.
+- This guide targets Linux containers; other container runtimes (Podman,
+  etc.) aren't tested against it.
+- Operators supply their own config, certificates, static roots, and secrets
+  — none of these are baked into the image or package.
```
Jaruso: This is still factually inconsistent with the native package path. The Docker image intentionally bakes in no operator config, but both DEB and RPM **do** install a starter `/etc/tardigrade/tardigrade.conf` and `/etc/tardigrade/tardigrade.env` template. Please distinguish those cases, e.g.:

> The Docker image contains no config/certificates/secrets; operators mount/inject them. Native packages install a non-secret starter config and environment template, which operators replace/edit for their deployment; certificates, static content, and live secrets remain operator-supplied.

That also lines up with the filesystem section instead of contradicting it here.

Jaruso (commented):
This is very close now. I verified the fixes from both prior rounds against the current head and resolved the stale threads: deterministic systemd PID handling, fixed Docker UID/GID, accurate package-created paths, safe TLS bind-mount guidance, RPM logrotate/ownership parity, config-validation wording, systemd master-mode reload caveat, Docker context size, upgrade/restart guidance, and the `ExecStartPre` wording are all fixed. The current CI run is also green, including DEB, clean Rocky RPM, and Docker lifecycle smoke.

I still recommend holding merge because #166 is specifically an operator-contract ticket and the remaining issues would make documented production workflows behave differently from what the guide says:

1. Docker `tardigrade.env` handling is still wrong. Tardigrade reads `TARDIGRADE_*` values from the process environment; it does not parse a bind-mounted `/etc/tardigrade/tardigrade.env`. The production baseline correctly `source`s its env file, while the Compose example currently has no `env_file:`. Please distinguish package-installed EnvironmentFile from Docker environment injection and demonstrate `env_file:`/`environment:` in Compose. Keep the explicit `environment:` PID value so it overrides any conflicting value from the env file.

2. The master-process limitation is broader than the current `systemctl reload` warning. The Docker reload command uses the same PID-file/SIGHUP mechanism; USR1 file-log reopening reaches only the master while workers own their own stderr descriptors; and `runMaster` currently terminates/waits workers one at a time, so the generic 35s stop window is not sufficient to guarantee a 30s drain across multiple workers. Either explicitly scope these deployment/control/log-rotation examples to single-process mode, or fix the master to fan out lifecycle signals to all workers before waiting and add coverage.

3. The Docker bind-mounted log-directory setup command needs privilege to assign UID/GID 10001 on a normal host. Make the copy/paste command actually runnable (`sudo install ...` or equivalent).

4. The Known Limitations sentence saying config is not baked into the package contradicts the actual DEB/RPM behavior: both packages install a starter `tardigrade.conf` and env template. Distinguish the Docker image from native packages there.

Once those are corrected, I do not see another issue-#166 acceptance-criteria gap in the current implementation.

Jaruso:
Pushed fixes for the third-pass review (1be83b53):

1. **`tardigrade.env` bind-mount misconception, fixed everywhere it appeared.** Confirmed `tardi` reads `TARDIGRADE_*` from the process environment only — it never parses a mounted file. Fixed the filesystem-layout section, the Docker walkthrough, and `compose.yaml` itself to demonstrate `env_file:` (which genuinely injects a host file's contents as container env vars) with an explicit `environment:` block layered on top so the deterministic PID path can't be desynced by whatever the env file sets.

2. **Master-process scope was too narrow — verified against source before writing anything.** Read `src/main.zig`'s `runMaster`: confirmed the shutdown loop kills and waits each worker one at a time rather than fanning out signals and waiting once (so the 35s window doesn't generally bound multi-worker shutdown), and confirmed `reopenErrorLog` on `SIGUSR1` only reopens the master's own fd, never forwarding to workers (which each open their own error-log fd independently in `configureErrorLog`). Took the "minimal safe fix" path you offered: expanded the known-limitations text to cover reload, USR1 rotation, and the stop window together, added a matching warning directly beside the Docker reload command, and scoped every affected example to `master_process false`. Didn't attempt the multi-worker signal fan-out fix itself — that's a genuine runtime concurrency change (with its own test coverage needs) that belongs in a separate PR, not folded into a docs/packaging ticket.

3. **Log-dir setup command needed `sudo`** — added; `chown` to an arbitrary UID isn't something an unprivileged operator can run as originally written.

4. **Known Limitations contradicted the filesystem-layout section** — fixed: Docker bakes in nothing, DEB/RPM install a non-secret starter config/env template that operators are expected to edit; certs and live secrets stay operator-supplied either way.

Re-verified after these changes: `compose.yaml` still parses cleanly, and the full Docker container lifecycle (build, UID/GID, health, PID file, reload, the conflicting-env-file regression, graceful stop) still passes end-to-end against the updated state.

packaging/systemd/tardigrade.service
```diff
@@ -8,13 +8,32 @@ Type=simple
...
 EnvironmentFile=-/etc/tardigrade/tardigrade.env
-ExecStart=/usr/bin/tardi run -c /etc/tardigrade/tardigrade.conf
-ExecReload=/usr/bin/tardi reload -c /etc/tardigrade/tardigrade.conf
-ExecStop=/usr/bin/tardi stop -c /etc/tardigrade/tardigrade.conf
+Environment=TARDIGRADE_PID_FILE=/run/tardigrade/tardigrade.pid
+
+RuntimeDirectory=tardigrade
+RuntimeDirectoryMode=0750
+
+ExecStartPre=/usr/bin/tardi check /etc/tardigrade/tardigrade.conf
```
Jaruso: **Blocking contract mismatch:** `ExecStart` now correctly forces the deterministic PID variable *after* `EnvironmentFile=` merging, but `ExecStartPre` does not. With the repository's own production-baseline env (`TARDIGRADE_PID_FILE=/tmp/tardigrade.pid`), this check loads `/tmp/tardigrade.pid` while the process it is about to start is forced to `/run/tardigrade/tardigrade.pid`. That contradicts `docs/DEPLOYMENT.md`'s claim that the unit preflight sees the same effective environment as `ExecStart`.

Please force the same runtime-owned value for both commands:

```ini
ExecStartPre=/usr/bin/env TARDIGRADE_PID_FILE=/run/tardigrade/tardigrade.pid /usr/bin/tardi check /etc/tardigrade/tardigrade.conf
ExecStart=/usr/bin/env TARDIGRADE_PID_FILE=/run/tardigrade/tardigrade.pid /usr/bin/tardi run -c /etc/tardigrade/tardigrade.conf
```

Then update both package smoke greps and the deployment-guide unit snippet/explanation. The PID field may not currently have path-specific validation, but the whole point of `ExecStartPre` here is to validate the effective configuration the service will actually execute; leaving one environment value deliberately different makes that contract brittle and the docs inaccurate.

docs/DEPLOYMENT.md
```diff
@@ -0,0 +1,517 @@
...
+  reload control across all workers — use restart semantics instead
+  (`systemctl restart tardigrade`, or recreate the container) rather than
+  `reload`.
+- **USR1 log-reopen after rotation.** Each worker opens its own error-log
+  file descriptor independently at startup. `SIGUSR1` delivered to the
+  master (via `systemctl kill --signal=USR1`, `docker kill --signal=USR1`,
+  or the packaged logrotate policy) only reopens the *master's own*
+  descriptor today — it is not forwarded to worker processes, so workers
+  keep writing to the rotated-away inode.
+- **The 35s stop window.** The current shutdown path stops workers one at a
```
Jaruso: **This stop limitation is not equivalent under the current systemd unit and Docker.** The unit does not set `KillMode=`, so systemd uses the default `control-group`. Per `systemd.kill(5)`/`systemd.service(5)`, after this unit's `ExecStop` helper sends SIGTERM and exits, systemd applies the stop signal to *all remaining processes in the service cgroup*. The workers therefore receive SIGTERM concurrently from systemd instead of necessarily waiting for the master's serial `kill` + `wait` loop, so `TimeoutStopSec=35s` is not inherently an `N × 30s` problem for the packaged systemd path.

There is a related distinction in the preceding USR1 bullet: `systemctl kill --signal=USR1 tardigrade.service` defaults to targeting **all** unit processes, so that manual command can signal the workers too. The packaged logrotate policy is the thing that explicitly uses `--kill-who=main`, while Docker's `docker kill` targets container PID 1.

Please split the warning by supervisor instead of saying "None of this is specific to Docker or systemd":

- SIGHUP/PID-file reload limitation: both paths.
- packaged systemd logrotate (`--kill-who=main`) and Docker USR1: main-only today; plain `systemctl kill --signal=USR1 ...` is not main-only.
- serial master shutdown / 35s grace limitation: Docker (and other master-only signal paths), not the current default-`KillMode=control-group` systemd stop behavior.

I would **not** change the unit to `KillMode=process` just to make the warning true; systemd explicitly discourages that because child processes can escape service lifecycle management. Correct the docs to match the supervisor behavior instead.

Jaruso (commented):
**Changes still needed before merge.** GitHub does not allow this account to request changes on its own PR, so this is a COMMENT review with the remaining blocking contract issues.

This pass is otherwise very clean. I re-checked all 17 changed files against #166 and every finding from the prior three rounds. The four previous outstanding doc issues are fixed and I resolved those outdated threads. The current CI run is green, including DEB packaging, the now-clean Rocky RPM rebuild, and the Docker lifecycle smoke. I also checked the new `build.zig` RHEL library-path/glibc-target work against the native x86_64/aarch64 release workflow and do not see another issue there.

I found two remaining systemd-specific accuracy problems:

1. **`ExecStartPre` still validates a different PID setting from `ExecStart`.** The unit correctly forces `TARDIGRADE_PID_FILE=/run/tardigrade/tardigrade.pid` at `ExecStart` command level so an `EnvironmentFile=` value cannot override the runtime PID contract. But `ExecStartPre` is not wrapped the same way, so it still sees a conflicting `TARDIGRADE_PID_FILE` from `tardigrade.env` (the repository's own production-baseline env intentionally contains `/tmp/tardigrade.pid` for standalone/manual runs). That makes the guide's statement that `ExecStartPre` sees the same effective environment as `ExecStart` false. Force the PID value for the preflight too, and update the DEB/RPM assertions plus the unit snippet in `docs/DEPLOYMENT.md`.

2. **The new master/worker warning treats systemd and Docker as if they have identical signal fan-out, but they do not.** The packaged unit leaves `KillMode=` at systemd's default `control-group`; after the asynchronous `ExecStop` command exits, systemd sends its stop signal to all remaining processes in the service cgroup. That means worker SIGTERM/drain begins across the cgroup rather than being limited to Tardigrade's serial master `kill`/`wait` loop, so the documented `N × drain`/35-second limitation is not generally true for this systemd unit. Similarly, `systemctl kill --signal=USR1 tardigrade.service` defaults to signaling all unit processes, whereas the packaged logrotate policy explicitly uses `--kill-who=main` and Docker's `docker kill` reaches PID 1. Please scope the limitations precisely instead of saying they are identical under both supervisors.

A good final wording split would be:

- PID-file/SIGHUP **reload** limitation: applies to both systemd and Docker while `master_process true`.
- packaged logrotate (`--kill-who=main`) and Docker `kill --signal=USR1`: worker log descriptors are not reopened; a manual systemd `systemctl kill --signal=USR1 tardigrade.service` without a main-only selector is different because systemctl's default target is all processes.
- serial master shutdown / 35s grace limitation: applies to Docker or other supervisors that signal only the master; do not claim the same serial timing for the current systemd unit with default `KillMode=control-group`.

The relevant systemd semantics are documented in `systemd.kill(5)` (`KillMode=control-group` is the default) and `systemd.service(5)` (`ExecStop` commands that merely send a signal and return are followed immediately by the unit's normal KillMode termination step).

Once these two small contract corrections are made, I do not currently see another #166 acceptance-criteria gap or implementation blocker.

Jaruso:
Pushed fixes for the fourth-pass review (34ae9984):

1. **`ExecStartPre` now forces the same PID value as `ExecStart`.** Wrapped it with the same `/usr/bin/env TARDIGRADE_PID_FILE=/run/tardigrade/tardigrade.pid ...` override, so the preflight check validates against the same effective environment the service actually starts with — the production-baseline env's own `/tmp/tardigrade.pid` (intentional there, for its standalone-run instructions) can no longer make `ExecStartPre` see a different PID setting than `ExecStart`. Updated both `scripts/test-deb-package.sh` and `scripts/test-rpm-package.sh` assertions and the unit snippet/explanation in `docs/DEPLOYMENT.md` to match, and cross-checked the new grep patterns byte-for-byte against the actual unit file.

2. **Split the master/worker warning by supervisor**, per your `systemd.kill(5)`/`systemd.service(5)` breakdown — this was a real correction, not just wording: the packaged unit leaves `KillMode=` at the default `control-group`, so systemd delivers the stop signal to the whole cgroup concurrently once `ExecStop` returns, meaning the 35s/N-worker drain concern is genuinely Docker-specific (only PID 1 gets signaled there), not a systemd problem. Similarly noted that a manual `systemctl kill --signal=USR1` (no `--kill-who=main`) targets the whole cgroup by default, unlike the packaged logrotate policy's explicit `--kill-who=main` or Docker's PID-1-only `docker kill`. Rewrote both the `### Commands` section warning and the Known Limitations entry to reflect this per-supervisor split exactly as you laid it out, and did not touch `KillMode=` to force uniformity — left it at the safer default per your explicit guidance against `KillMode=process`.

This round's changes are docs/unit-file/smoke-test-script only (no Dockerfile/compose.yaml changes), so I didn't re-run the Docker lifecycle smoke test — cross-checked the corrected `ExecStartPre`/`ExecStart` grep assertions directly against the unit file instead, along with a full markdown link/anchor pass and `shellcheck`/`git diff --check`, all clean.

Jaruso (commented):
Fifth-pass review: **no further code/doc changes requested for #166.** I re-checked the full current diff against the issue's expanded acceptance criteria and all findings from the prior rounds.

Both fourth-pass blockers are fixed correctly:

- `ExecStartPre` now gets the same command-level `TARDIGRADE_PID_FILE=/run/tardigrade/tardigrade.pid` override as `ExecStart`, so the preflight and runtime no longer diverge when `tardigrade.env` contains a conflicting PID path. Both DEB and RPM smoke tests assert the corrected line.
- The master/worker documentation now correctly distinguishes supervisor behavior: PID-file/SIGHUP reload remains a limitation under both systemd and Docker; main-only USR1 paths are called out specifically; and the serialized-worker/35s stop concern is scoped to Docker/master-only signal paths rather than incorrectly applying it to the packaged systemd unit's default control-group kill behavior.

I also re-scanned the rest of the PR: Docker multi-stage/non-root behavior, fixed numeric UID/GID and tmpfs ownership, Compose env injection, TLS bind-mount permissions, filesystem/logging guidance, RPM logrotate and log ownership parity, clean Rocky RPM build isolation/glibc handling, `.dockerignore`, upgrade/restart guidance, README wiring, maturity/known-limitations language, and Docker lifecycle smoke coverage all still line up with #166.

One CI caveat, but **not a #621 implementation finding**: the latest workflow run on `34ae998` is marked `cancelled` because `Test appliance profile (ubuntu-24.04-arm)` reached the job's time limit while its full integration-suite step was still running. The changes in this final commit are limited to the systemd/deployment docs and package-smoke assertions, and the **Packaging smoke job passed completely on this exact SHA**, including DEB, clean Rocky RPM, and Docker lifecycle smoke. The immediately preceding head had a fully green matrix. I would rerun the cancelled ARM appliance job before merging so branch CI is green on the final SHA, but I do not see a code change to request from this PR for that timeout.

Because this account is the PR author GitHub does not allow a normal approval review, so this COMMENT is the equivalent review result: **#166 is implementation-complete and merge-ready once the cancelled CI job is rerun successfully.**

Jaruso:
Thanks for the fifth-pass review — glad #166 is implementation-complete.

Reran the one flagged CI job (\`Test appliance profile (ubuntu-24.04-arm)\`, which had hit the 30-minute job timeout mid integration-suite — an infra timeout, not a real failure). It passed on rerun in 6m55s. Full branch CI is now green on the current head (34ae9984), including Packaging smoke (DEB, clean Rocky RPM, Docker lifecycle) and every other job.

PR #624, opened 2026-08-14
quic: retain opt-in H3 interop observability artifacts

Jaruso:
## Summary

- add h3_interop_tool --qlog-dir / --keylog-path flags for opt-in .sqlog and NSS keylog capture
- extend scripts/interop/run-interop.sh with INTEROP_QLOG, INTEROP_KEYLOG, and INTEROP_ARTIFACT_DIR wiring plus sensitive keylog markers
- document the local-debug artifact workflow and public-CI keylog restrictions

Refs #255

## Validation

- zig build build-h3-interop --summary all --error-style verbose
- bash -n scripts/interop/run-interop.sh
- INTEROP_WORKDIR=<tmp> scripts/interop/run-interop.sh (no peers: 8 skipped, no artifacts)
- h3_interop_tool client smoke with --qlog-dir/--keylog-path: .sqlog header written, .keys created 0600, warning emitted
- fake external-server failure smoke: script exited nonzero, printed artifact root once, retained per-case .sqlog/.keys and SENSITIVE-KEYLOG.txt

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -840,6 +840,7 @@ pub fn build(b: *std.Build) void {
     h3_interop_mod.addImport("http3", http3_mod);
     h3_interop_mod.addImport("stream_transport", stream_transport_mod);
     h3_interop_mod.addImport("tls_interop_matrix", tls_interop_matrix_mod);
+    h3_interop_mod.addImport("zig_compat", compat_mod);
     const h3_interop_tool = b.addExecutable(.{
         .name = "h3_interop_tool",
         .root_module = h3_interop_mod,
--- a/build.zig
+++ b/build.zig
@@ -848,6 +848,9 @@ pub fn build(b: *std.Build) void {
     const h3_interop_install = b.addInstallArtifact(h3_interop_tool, .{});
     const h3_interop_step = b.step("build-h3-interop", "Build the native HTTP/3 interop client/server tool");
     h3_interop_step.dependOn(&h3_interop_install.step);
+    const h3_interop_tests = b.addTest(.{ .root_module = h3_interop_mod });
+    const run_h3_interop_tests = b.addRunArtifact(h3_interop_tests);
+    test_step.dependOn(&run_h3_interop_tests.step);
 
     // Out-of-process record-transport TLS conformance driver (#338). Runs
     // the shared engine as client or server against external peers
--- a/docs/QUIC_QLOG.md
+++ b/docs/QUIC_QLOG.md
@@ -14,8 +14,9 @@ This document is the design of record. The scaffolding it describes lives in:
   gateway HTTP/3 listener.
 
 The event models, JSON-SEQ serializers, HTTP/3 qlog bridge, QUIC qlog bridge,
-shared TLS keylog emission seam, and gateway qlog/keylog file destinations are
-implemented. #247 artifact retention remains opt-in harness work.
+shared TLS keylog emission seam, gateway qlog/keylog file destinations, and
+#247 external H3 interop artifact retention are implemented. All artifact
+capture remains opt-in.
 
 ## Goals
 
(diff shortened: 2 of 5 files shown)


Jaruso (commented):
Changes are needed before merge. I reviewed this against the reconciled #255 acceptance checklist, especially Slice 7 (extending the existing #247 H3 interop harness with opt-in qlog/keylog failure artifacts), and walked all five changed files plus the current CI run.

The overall shape is right: default runs remain artifact-free; `INTEROP_QLOG` and `INTEROP_KEYLOG` are independently opt-in; native client/server directions receive the new flags; keylogs get explicit sensitivity warnings/markers; the failure artifact root is printed once; and ordinary CI does not enable/upload keylogs. The existing build/test/security matrix is otherwise healthy.

Three artifact-correctness gaps need to be fixed, called out inline:

1. `--keylog-path` is opened with truncation for every `ArtifactSink.init`. The server deliberately supports multiple sequential connections for `--requests N`, so a later connection can erase the earlier connection's NSS secrets from the same per-role `.keys` file. This needs append semantics (preferably the same `O_APPEND` + 0600 pattern already used by the production observability sink).
2. The interop `.sqlog` never emits `quic:connection_started`. `Connection.Event` has no start event; the production HTTP/3 runtime synthesizes it at connection admission. Because this harness only maps `Connection.Event`, its new traces omit a required #255 event class unless the composition root emits it explicitly for both client and server.
3. Artifact drop diagnostics are not reliably surfaced. The server destroys each per-connection sink without ever reporting its counters, while the client only reports them on the success tail; explicit `std.process.exit(...)` failure paths bypass defers. A truncated failure artifact can therefore look complete, contrary to the qlog/keylog sink contract.

There is also one mechanical blocker: CI run 1468 is red because `zig fmt --check build.zig src/ tests/` reports `tests/h3_interop_tool.zig`. All other completed build/test/security jobs passed, so run `zig fmt tests/h3_interop_tool.zig` (or the repo-wide formatter) and re-run CI.

Please add focused regression coverage while fixing these. At minimum, cover a server artifact path reused across more than one connection/request so prior keylog lines are retained, and verify an interop qlog contains its header plus `quic:connection_started` and at least one representative transport/H3 event. The current CI exercises the default harness path but does not prove the new opt-in artifact behavior.

Once these are fixed and formatting is green, the implementation should satisfy the remaining Slice 7 closeout work for #255.

Jaruso (commented):
Second-pass review on `feac8d0d`:

The three blocking findings from my prior review are fixed correctly:

- keylog files now use the same append-only `0600` pattern as the production HTTP/3 observability sink, with regression coverage proving two connection sinks preserve both sets of secrets;
- client and server traces explicitly synthesize `quic:connection_started`, and the new qlog artifact test asserts the header plus representative QUIC/H3 records;
- artifact drop diagnostics are now reported per server connection and on the client's explicit `process.exit` failure paths, so truncated failure evidence is no longer silent.

I re-read the full PR diff after those fixes. I do not see another code defect in the harness/artifact implementation. CI for this head is fully green, including Format, Lint, Linux/macOS/ARM tests, security, and H3 interop.

There is **one remaining #255 closeout/validation item** before I would let this PR actually close the issue. The 2026-08-10 reconciliation comment says to keep #255 open until a produced `.sqlog` successfully loads in qvis. PR #606 explicitly documented that qvis was unavailable and that manual load check was not performed, and I cannot find a later #255 PR recording that validation.

So one of these should happen before merge:

1. Produce a real `.sqlog` from the now-wired path, load it in qvis, and record the successful validation here/on #255; **or**
2. Change `Closes #255` to `Refs #255`, merge this code, then close #255 only after the qvis compatibility check is completed.

No additional code changes are requested from this review unless the qvis check exposes a schema/compatibility problem.

Jaruso (commented):
Re-review on current head `feac8d0d`:

There have been **no code changes since my previous review**; this is the same commit I already re-read end-to-end. CI remains green on this exact head.

I also re-checked the #255 issue discussion for the one remaining closeout item. There is still no recorded successful qvis load of a produced `.sqlog`. The 2026-08-10 reconciliation explicitly lists that manual qvis compatibility check as the final validation for Slice 1 / issue closeout, while PR #606 records that the check could not be performed at the time.

So my disposition is unchanged:

- **No additional code defects found.** The implementation itself is ready.
- **Do not let this PR close #255 yet unless the qvis validation has actually been performed and recorded.** Either document a successful qvis load and keep `Closes #255`, or change this PR to `Refs #255` and leave the issue open for that final validation.

There is nothing else I would ask the implementation agent to change on this head.

PR #633, opened 2026-08-19
Wire circuit breaker into proxy request paths

### LINKED ISSUE #627: proxy: reconcile configured circuit-breaker with the live upstream request path
## Problem

`GatewayState` constructs a process-wide circuit breaker from
`TARDIGRADE_CB_THRESHOLD` / `TARDIGRADE_CB_TIMEOUT_MS`
(`http.circuit_breaker.CircuitBreaker`), and exposes
`circuitTryAcquire()` / `circuitRecordFailure()` / `circuitRecordSuccess()` /
`circuitStateName()` with isolated unit test coverage
(`src/gateway_state.zig`, e.g. "gateway circuit breaker opens under
upstream failure pressure", "gateway circuit breaker recovers through a
half-open probe").

However, as of the current `main` (commit `fde7bc1` at the time of
writing), none of the live proxy request paths appear to call these
methods:

- `src/gateway_proxy.zig` — no call site; the breaker is only mentioned in
  a comment
- `src/gateway_proxy_runtime.zig` — no call site
- `src/gateway_control_plane_proxy.zig` — no call site
- `src/gateway_handlers.zig` — no call site

So a configured `TARDIGRADE_CB_THRESHOLD` currently has no effect on
request handling: failures are never recorded against the breaker, and no
request is ever gated by `circuitTryAcquire()`.

Despite this, [`docs/OBSERVABILITY.md`'s operator troubleshooting
table](docs/OBSERVABILITY.md#operator-troubleshooting) describes a
circuit-breaker-open symptom ("All requests to one backend failing fast
with no upstream contact... the breaker recovers via a half-open probe
once `upstream_fail_timeout` elapses") as if it's live behavior, and
[`docs/SUPPORT_MATRIX.md`](docs/SUPPORT_MATRIX.md) lists "Upstream health
checks and basic load balancing" (representative surface
`health_checker`, `circuit_breaker`) as `stable`, "covered by integration
scenarios."

This was surfaced while writing the canonical troubleshooting runbook in
#167 / PR #626 — the guide originally repeated the circuit-breaker-as-503
claim from `OBSERVABILITY.md`, and independent source inspection during
review couldn't confirm the wiring, so the runbook now deliberately avoids
documenting it as live behavior and points here instead.

Note: #188 is closed as completed and its own text assumes the breaker
already participates in upstream overload handling, so it is not an
accurate current owner for this gap.

## Acceptance criteria

Decide whether the circuit breaker is a supported proxy feature:

- **If yes:** wire `circuitTryAcquire()` as a gate before proxy attempts
  and `circuitRecordFailure()`/`circuitRecordSuccess()` on the relevant
  outcomes across the data-plane and control-plane proxy paths, define the
  client-visible status/error category for an open breaker, and add live
  integration coverage for the open → half-open → closed state transitions
  (not just the existing isolated unit tests).
- **If no:** remove or clearly deprecate the unused
  `TARDIGRADE_CB_THRESHOLD`/`TARDIGRADE_CB_TIMEOUT_MS` configuration and
  runtime state, and reconcile `OBSERVABILITY.md`'s troubleshooting table
  and `SUPPORT_MATRIX.md`'s `stable` listing with the actual behavior.

Either way, add regression coverage (a test asserting the breaker either is
or is not consulted on the live path) so runtime and documentation cannot
drift apart again silently.

## Related

- #167 / PR #626 — canonical troubleshooting runbook, where this gap was
  found during review.
- #188 (closed) — assumes the breaker already participates in overload
  handling; not an accurate owner for the current state.

Jaruso:
## Summary
- gate live data-plane, control-plane, and HTTP/3 proxy attempts on the configured circuit breaker
- feed breaker success/failure from live upstream outcomes, including absolute proxy_pass targets
- document the open-breaker response as 503 upstream_circuit_open and add live integration coverage for open -> half-open -> closed recovery

Fixes #627

## Validation
- zig build --summary all --error-style verbose
- zig build test-integration --summary all --error-style verbose --test-timeout 120s

Note: broad `zig build test --summary all --error-style verbose --test-timeout 120s` was interrupted after it remained silent beyond the configured timeout; no Zig test process was left running.

### DIFF
--- a/docs/OBSERVABILITY.md
+++ b/docs/OBSERVABILITY.md
@@ -399,9 +399,12 @@ Notes on the two pool-style resources called out in the issue:
   least-connections balancer sheds a saturated backend to the least-loaded
   healthy one and returns no candidate when every backend is unhealthy (the
   request is shed deterministically rather than queued without bound); the
-  circuit breaker fast-fails once a backend trips its failure threshold and
-  recovers through a single half-open probe. Both paths have fault-injection
-  coverage in `src/gateway_state.zig`.
+  process-wide circuit breaker fast-fails proxied requests with HTTP `503` and
+  JSON `"code":"upstream_circuit_open"` once upstream failures trip its
+  configured threshold, then recovers through a half-open probe after
+  `TARDIGRADE_CB_TIMEOUT_MS`. The state machine has focused coverage in
+  `src/gateway_state.zig`, and live proxy gating is covered in
+  `src/gateway_proxy_runtime.zig`.
 - **Log / metrics sink slow or unavailable** — access logging is best-effort and
   never blocks the request that emitted it. In buffered mode the buffer flushes
   at its configured threshold and is cleared regardless of the write outcome, so
@@ -445,7 +448,7 @@ regression test in `src/gateway_accept.zig`.
 | `413` / `431` / `414` in access logs | Oversized body, headers, or URI | Confirm the configured request limits match legitimate client traffic before raising them |
 | Tail latency spikes under many idle keepalive clients | Parked-connection backlog or too-long idle timeout | Watch parked `timeouts_total`; tune the idle-park timeout and `max_requests_per_connection` |
 | Gaps in access logs, throughput otherwise normal | Log sink (stderr pipe / syslog) slow or unavailable | Logging is best-effort: dropped lines are counted internally and the request path is never blocked; check the downstream log collector / pipe rather than the gateway |
-| All requests to one backend failing fast with no upstream contact | Circuit breaker open for that backend | Expected protection after repeated upstream failures; confirm the upstream is healthy — the breaker recovers via a half-open probe once `upstream_fail_timeout` elapses |
+| Proxied requests fail fast with `503` / `upstream_circuit_open` and no upstream contact | Circuit breaker open | Expected protection after repeated upstream failures; confirm the upstream is healthy — the breaker recovers via a half-open probe once `TARDIGRADE_CB_TIMEOUT_MS` elapses |
 
 Tuning principle: raise a limit only after confirming the gateway has headroom
 (CPU, memory, fds) to honor it. The limits exist so that exhaustion produces a
--- a/docs/OBSERVABILITY.md
+++ b/docs/OBSERVABILITY.md
@@ -403,8 +403,10 @@ Notes on the two pool-style resources called out in the issue:
   JSON `"code":"upstream_circuit_open"` once upstream failures trip its
   configured threshold, then recovers through a half-open probe after
   `TARDIGRADE_CB_TIMEOUT_MS`. The state machine has focused coverage in
-  `src/gateway_state.zig`, and live proxy gating is covered in
-  `src/gateway_proxy_runtime.zig`.
+  `src/gateway_state.zig` and `src/gateway_proxy_runtime.zig`; live
+  open → half-open → closed behavior is covered by
+  `proxy.circuit_breaker_opens_fast_fails_and_recovers` in
+  `tests/integration.zig`.
 - **Log / metrics sink slow or unavailable** — access logging is best-effort and
   never blocks the request that emitted it. In buffered mode the buffer flushes
   at its configured threshold and is cleared regardless of the write outcome, so
--- a/src/edge_gateway.zig
+++ b/src/edge_gateway.zig
@@ -3319,16 +3319,16 @@ fn h2UnsupportedProxyOrchestration(cfg: *const edge_config.EdgeConfig, matched: 
         cfg.proxy_streaming_mode.responseStreamingEnabled(),
     )) return true;
 
-    // `upstream_max_fails`/`upstream_fail_timeout_ms` drive passive
-    // upstream health accounting that `handleLocationProxyPass` records
-    // around the low-level executor this path calls directly; skipping it
-    // means a configured circuit breaker silently never opens (or closes)
+    // `upstream_max_fails`/`upstream_fail_timeout_ms` and the process-wide
+    // circuit breaker depend on health/accounting that `handleLocationProxyPass`
+    // records around the low-level executor this path calls directly; skipping
+    // it means configured upstream protection silently never opens (or closes)
     // for traffic negotiated over H2. Deliberately NOT included:
     // `upstream_pool_enabled` -- it defaults to true, and connection
     // pooling itself is already correctly threaded through via
     // `state.upstream_pool`/`state.h2_pool` below, so treating it as
     // unsupported would fail closed on every H2 proxy_pass by default.
-    if (cfg.upstream_max_fails > 0) return true;
+    if (cfg.upstream_max_fails > 0 or cfg.cb_threshold > 0) return true;
 
     const pool = gs.upstreamPoolForScope(cfg, .global);
     return cfg.upstream_retry_attempts > 1 or
--- a/src/gateway_control_plane_proxy.zig
+++ b/src/gateway_control_plane_proxy.zig
@@ -8,6 +8,7 @@
 //! work can replace the current bounded-buffer compatibility executor.
 
 const compat = @import("zig_compat");
+const builtin = @import("builtin");
 const std = @import("std");
 const http = @import("http.zig");
 const edge_config = @import("edge_config.zig");
@@ -48,6 +49,7 @@ pub fn controlPlaneBufferedResponseLimit(cfg: *const edge_config.EdgeConfig) usi
 
 fn controlPlaneAttemptErrorCountsAsUpstreamFailure(err: anyerror) bool {
     return switch (err) {
+        error.ControlPlaneResponseMaterializationFailed,
         error.OutOfMemory,
         error.UpstreamUntrusted,
         error.UpstreamAtCapacity,
@@ -73,6 +75,21 @@ fn recordControlPlaneProxyFailure(
         state.circuitRecordFailurePermit(permit);
     } else {
         state.recordProxyUpstreamFailure(cfg, upstream_base_url, permit);
+    }
+}
+
+fn recordControlPlaneProxyStatus(
+    state: *GatewayState,
+    cfg: *const edge_config.EdgeConfig,
+    upstream_base_url: []const u8,
+    absolute_target: bool,
+    permit: http.circuit_breaker.Permit,
+    status: u16,
+) void {
+    if (status >= 500) {
+        recordControlPlaneProxyFailure(state, cfg, upstream_base_url, absolute_target, permit);
+    } else {
+        recordControlPlaneProxySuccess(state, cfg, upstream_base_url, absolute_target, permit);
     }
 }
 
@@ -154,6 +171,7 @@ pub const ControlPlaneProxyResult = struct {
     set_cookie: ?[]u8,
     cacheable: bool,
     upstream_addr: []const u8,
+    accounted: bool = false,
 };
 
 pub const ControlPlaneProxyExecution = union(enum) {
@@ -387,7 +405,7 @@ pub fn executeBoundedControlPlaneJsonProxy(
                 absolute_proxy_target,
                 circuit_permit,
             ) catch |err| {
-                if (err == error.DownstreamWriteFailed) return err;
+                if (err == error.DownstreamWriteFailed or err == error.ControlPlaneResponseMaterializationFailed) return err;
                 if (controlPlaneAttemptErrorCountsAsUpstreamFailure(err)) {
                     recordControlPlaneProxyFailure(state, cfg, upstream_base_url, absolute_proxy_target, circuit_permit);
                 } else {
@@ -413,8 +431,10 @@ pub fn executeBoundedControlPlaneJsonProxy(
                 return exec;
             },
             .buffered => |res| {
+                if (!res.accounted) {
+                    recordControlPlaneProxyStatus(state, cfg, upstream_base_url, absolute_proxy_target, circuit_permit, res.status);
+                }
                 if (res.status >= 500) {
-                    recordControlPlaneProxyFailure(state, cfg, upstream_base_url, absolute_proxy_target, circuit_permit);
                     if (attempt + 1 < max_attempts) {
                         allocator.free(res.body);
                         allocator.free(res.content_type);
@@ -423,8 +443,6 @@ pub fn executeBoundedControlPlaneJsonProxy(
                         if (res.set_cookie) |cookie| allocator.free(cookie);
                         continue;
                     }
-                } else {
-                    recordControlPlaneProxySuccess(state, cfg, upstream_base_url, absolute_proxy_target, circuit_permit);
                 }
                 return exec;
             },
@@ -581,19 +599,16 @@ fn executeBoundedControlPlaneJsonProxyAttempt(
         const upstream_reason = buffered_resp.reason;
         const upstream_content_type = buffered_resp.headerValue("Content-Type") orelse JSON_CONTENT_TYPE;
         const upstream_content_disposition = buffered_resp.headerValue("Content-Disposition");
+        recordControlPlaneProxyStatus(state, cfg, upstream_base_url, absolute_proxy_target, circuit_permit, status_code);
+        try maybeFailControlPlaneMaterializationAfterStatus();
         const upstream_location = if (buffered_resp.headerValue("Location")) |location|
-            try allocator.dupe(u8, location)
+            allocator.dupe(u8, location) catch return error.ControlPlaneResponseMaterializationFailed
         else
             null;
         errdefer if (upstream_location) |location| allocator.free(location);
         const cacheable = !bufferedUpstreamResponseHasNoStore(&buffered_resp);
         const stream_status = enable_streaming_success and (status_code == 200 or cfg.proxy_stream_all_statuses);
         if (stream_status) {
-            if (status_code >= 500) {
-                recordControlPlaneProxyFailure(state, cfg, upstream_base_url, absolute_proxy_target, circuit_permit);
-            } else {
-                recordControlPlaneProxySuccess(state, cfg, upstream_base_url, absolute_proxy_target, circuit_permit);
-            }
             writeStreamedUpstreamResponse(
                 downstream_writer,
                 status_code,
@@ -613,15 +628,15 @@ fn executeBoundedControlPlaneJsonProxyAttempt(
         }
 
         if (status_code != 200) {
-            const buffered_content_type = try allocator.dupe(u8, upstream_content_type);
+            const buffered_content_type = allocator.dupe(u8, upstream_content_type) catch return error.ControlPlaneResponseMaterializationFailed;
             errdefer allocator.free(buffered_content_type);
             const buffered_content_disposition = if (upstream_content_disposition) |cd|
-                try allocator.dupe(u8, cd)
+                allocator.dupe(u8, cd) catch return error.ControlPlaneResponseMaterializationFailed
             else
                 null;
             errdefer if (buffered_content_disposition) |cd| allocator.free(cd);
             const buffered_set_cookie = if (sticky_set_cookie) |cookie|
-                try allocator.dupe(u8, cookie)
+                allocator.dupe(u8, cookie) catch return error.ControlPlaneResponseMaterializationFailed
             else
                 null;
             errdefer if (buffered_set_cookie) |cookie| allocator.free(cookie);
@@ -649,21 +664,22 @@ fn executeBoundedControlPlaneJsonProxyAttempt(
                     .set_cookie = buffered_set_cookie,
                     .cacheable = false,
                     .upstream_addr = upstream_host,
+                    .accounted = true,
                 },
             };
         }
 
-        const body = try allocator.dupe(u8, buffered_resp.body);
+        const body = allocator.dupe(u8, buffered_resp.body) catch return error.ControlPlaneResponseMaterializationFailed;
         errdefer allocator.free(body);
-        const buffered_content_type = try allocator.dupe(u8, upstream_content_type);
+        const buffered_content_type = allocator.dupe(u8, upstream_content_type) catch return error.ControlPlaneResponseMaterializationFailed;
         errdefer allocator.free(buffered_content_type);
         const buffered_content_disposition = if (upstream_content_disposition) |cd|
-            try allocator.dupe(u8, cd)
+            allocator.dupe(u8, cd) catch return error.ControlPlaneResponseMaterializationFailed
         else
             null;
         errdefer if (buffered_content_disposition) |cd| allocator.free(cd);
         const buffered_set_cookie = if (sticky_set_cookie) |cookie|
-            try allocator.dupe(u8, cookie)
+            allocator.dupe(u8, cookie) catch return error.ControlPlaneResponseMaterializationFailed
         else
             null;
         errdefer if (buffered_set_cookie) |cookie| allocator.free(cookie);
@@ -690,9 +706,16 @@ fn executeBoundedControlPlaneJsonProxyAttempt(
                 .set_cookie = buffered_set_cookie,
                 .cacheable = cacheable,
                 .upstream_addr = upstream_host,
+                .accounted = true,
             },
         };
     }
+}
+
+var test_fail_control_plane_materialization_after_status: bool = false;
+
+fn maybeFailControlPlaneMaterializationAfterStatus() !void {
+    if (builtin.is_test and test_fail_control_plane_materialization_after_status) return error.ControlPlaneResponseMaterializationFailed;
 }
 
 test "control-plane buffered response limit is explicit and bounded" {
@@ -942,6 +965,103 @@ test "control-plane downstream streaming write failure records known 500 and nev
         FailingControlPlaneDownstreamWriter{},
         &state,
         true,
+        null,
+    ));
+
+    try std.testing.expectEqual(@as(usize, 1), origin.requestCount());
+    try std.testing.expectEqual(@as(usize, 0), state.upstreamUnhealthyCount());
+    try std.testing.expectEqualStrings("open", state.circuitStateName());
+    try std.testing.expect(state.circuitTryAcquirePermit() != null);
+}
+
+test "control-plane post-status materialization failure records known 200 and never retries" {
+    const allocator = std.testing.allocator;
+    var origin = try ControlPlaneTestOrigin.start(allocator, &.{ 200, 200 });
+    defer origin.stop();
+    try origin.run();
+
+    var state: GatewayState = undefined;
+    initControlPlaneProxyTestState(&state, allocator);
+    defer deinitControlPlaneProxyTestState(&state);
+    var cfg = initControlPlaneProxyTestConfig("http://configured.example");
+    cfg.upstream_retry_idempotent_only = false;
+    cfg.upstream_retry_attempts = 3;
+    const target = try std.fmt.allocPrint(allocator, "http://127.0.0.1:{d}/api", .{origin.port()});
+    defer allocator.free(target);
+    state.circuit_breaker = http.circuit_breaker.CircuitBreaker.init(.{ .threshold = 1, .timeout_ms = 0, .half_open_successes = 1 });
+    state.circuitRecordFailurePermit(state.circuitTryAcquirePermit().?);
+
+    test_fail_control_plane_materialization_after_status = true;
+    defer test_fail_control_plane_materialization_after_status = false;
+    var body_buf: [1024]u8 = undefined;
+    var body_writer = compat.fixedBufferStream(&body_buf);
+    try std.testing.expectError(error.ControlPlaneResponseMaterializationFailed, executeBoundedControlPlaneJsonProxy(
+        allocator,
+        &cfg,
+        .global,
+        target,
+        null,
+        "{}",
+        "test-correlation",
+        "127.0.0.1",
+        null,
+        null,
+        null,
+        null,
+        null,
+        "localhost",
+        null,
+        body_writer.writer(),
+        &state,
+        false,
+        null,
+    ));
+
+    try std.testing.expectEqual(@as(usize, 1), origin.requestCount());
+    try std.testing.expectEqual(@as(usize, 0), state.upstreamUnhealthyCount());
+    try std.testing.expectEqualStrings("closed", state.circuitStateName());
+}
+
+test "control-plane post-status materialization failure records known 500 and never retries" {
+    const allocator = std.testing.allocator;
+    var origin = try ControlPlaneTestOrigin.start(allocator, &.{ 500, 500 });
+    defer origin.stop();
+    try origin.run();
+
+    var state: GatewayState = undefined;
+    initControlPlaneProxyTestState(&state, allocator);
+    defer deinitControlPlaneProxyTestState(&state);
+    var cfg = initControlPlaneProxyTestConfig("http://configured.example");
+    cfg.upstream_retry_idempotent_only = false;
+    cfg.upstream_retry_attempts = 3;
+    const target = try std.fmt.allocPrint(allocator, "http://127.0.0.1:{d}/api", .{origin.port()});
+    defer allocator.free(target);
+    state.circuit_breaker = http.circuit_breaker.CircuitBreaker.init(.{ .threshold = 1, .timeout_ms = 0, .half_open_successes = 1 });
+    state.circuitRecordFailurePermit(state.circuitTryAcquirePermit().?);
+
+    test_fail_control_plane_materialization_after_status = true;
+    defer test_fail_control_plane_materialization_after_status = false;
+    var body_buf: [1024]u8 = undefined;
+    var body_writer = compat.fixedBufferStream(&body_buf);
+    try std.testing.expectError(error.ControlPlaneResponseMaterializationFailed, executeBoundedControlPlaneJsonProxy(
+        allocator,
+        &cfg,
+        .global,
+        target,
+        null,
+        "{}",
+        "test-correlation",
+        "127.0.0.1",
+        null,
+        null,
+        null,
+        null,
+        null,
+        "localhost",
+        null,
+        body_writer.writer(),
+        &state,
+        false,
         null,
     ));
 
--- a/src/gateway_control_plane_proxy.zig
+++ b/src/gateway_control_plane_proxy.zig
@@ -326,7 +326,7 @@ pub fn executeBoundedControlPlaneJsonProxy(
         else
             null;
         defer if (sticky_set_cookie) |cookie| allocator.free(cookie);
-        if (!state.circuitTryAcquire()) return error.CircuitOpen;
+        const circuit_permit = state.circuitTryAcquirePermit() orelse return error.CircuitOpen;
         state.recordUpstreamAttemptStart(upstream_base_url);
         const exec = blk: {
             defer state.recordUpstreamAttemptEnd(upstream_base_url);
@@ -354,9 +354,9 @@ pub fn executeBoundedControlPlaneJsonProxy(
                 sticky_set_cookie,
             ) catch |err| {
                 if (controlPlaneAttemptErrorCountsAsUpstreamFailure(err)) {
-                    state.recordProxyUpstreamFailure(cfg, upstream_base_url);
+                    state.recordProxyUpstreamFailure(cfg, upstream_base_url, circuit_permit);
                 } else {
-                    state.circuitReleaseProbe();
+                    state.circuitReleasePermit(circuit_permit);
                 }
                 last_err = err;
                 if (attempt + 1 < max_attempts) {
@@ -369,15 +369,15 @@ pub fn executeBoundedControlPlaneJsonProxy(
         switch (exec) {
             .streamed_status => |streamed| {
                 if (streamed.status >= 500) {
-                    state.recordProxyUpstreamFailure(cfg, upstream_base_url);
+                    state.recordProxyUpstreamFailure(cfg, upstream_base_url, circuit_permit);
                 } else {
-                    state.recordProxyUpstreamSuccess(cfg, upstream_base_url);
+                    state.recordProxyUpstreamSuccess(cfg, upstream_base_url, circuit_permit);
                 }
                 return exec;
             },
             .buffered => |res| {
                 if (res.status >= 500) {
-                    state.recordProxyUpstreamFailure(cfg, upstream_base_url);
+                    state.recordProxyUpstreamFailure(cfg, upstream_base_url, circuit_permit);
                     if (attempt + 1 < max_attempts) {
                         allocator.free(res.body);
                         allocator.free(res.content_type);
@@ -387,7 +387,7 @@ pub fn executeBoundedControlPlaneJsonProxy(
                         continue;
                     }
                 } else {
-                    state.recordProxyUpstreamSuccess(cfg, upstream_base_url);
+                    state.recordProxyUpstreamSuccess(cfg, upstream_base_url, circuit_permit);
                 }
                 return exec;
             },
--- a/src/gateway_control_plane_proxy.zig
+++ b/src/gateway_control_plane_proxy.zig
@@ -160,6 +160,7 @@ pub const ControlPlaneProxyExecution = union(enum) {
     streamed_status: struct {
         status: u16,
         upstream_addr: []const u8,
+        accounted: bool = false,
     },
     buffered: ControlPlaneProxyResult,
 };
@@ -383,7 +384,10 @@ pub fn executeBoundedControlPlaneJsonProxy(
                 state,
                 enable_streaming_success,
                 sticky_set_cookie,
+                absolute_proxy_target,
+                circuit_permit,
             ) catch |err| {
+                if (err == error.DownstreamWriteFailed) return err;
                 if (controlPlaneAttemptErrorCountsAsUpstreamFailure(err)) {
                     recordControlPlaneProxyFailure(state, cfg, upstream_base_url, absolute_proxy_target, circuit_permit);
                 } else {
@@ -399,10 +403,12 @@ pub fn executeBoundedControlPlaneJsonProxy(
         };
         switch (exec) {
             .streamed_status => |streamed| {
-                if (streamed.status >= 500) {
-                    recordControlPlaneProxyFailure(state, cfg, upstream_base_url, absolute_proxy_target, circuit_permit);
-                } else {
-                    recordControlPlaneProxySuccess(state, cfg, upstream_base_url, absolute_proxy_target, circuit_permit);
+                if (!streamed.accounted) {
+                    if (streamed.status >= 500) {
+                        recordControlPlaneProxyFailure(state, cfg, upstream_base_url, absolute_proxy_target, circuit_permit);
+                    } else {
+                        recordControlPlaneProxySuccess(state, cfg, upstream_base_url, absolute_proxy_target, circuit_permit);
+                    }
                 }
                 return exec;
             },
@@ -450,6 +456,8 @@ fn executeBoundedControlPlaneJsonProxyAttempt(
     state: *GatewayState,
     enable_streaming_success: bool,
     sticky_set_cookie: ?[]const u8,
+    absolute_proxy_target: bool,
+    circuit_permit: http.circuit_breaker.Permit,
 ) !ControlPlaneProxyExecution {
     const proxy_json_extra_header_slack = 12;
     const proxy_json_owned_header_value_slack = 3;
@@ -581,6 +589,11 @@ fn executeBoundedControlPlaneJsonProxyAttempt(
         const cacheable = !bufferedUpstreamResponseHasNoStore(&buffered_resp);
         const stream_status = enable_streaming_success and (status_code == 200 or cfg.proxy_stream_all_statuses);
         if (stream_status) {
+            if (status_code >= 500) {
+                recordControlPlaneProxyFailure(state, cfg, upstream_base_url, absolute_proxy_target, circuit_permit);
+            } else {
+                recordControlPlaneProxySuccess(state, cfg, upstream_base_url, absolute_proxy_target, circuit_permit);
+            }
             writeStreamedUpstreamResponse(
                 downstream_writer,
                 status_code,
@@ -596,7 +609,7 @@ fn executeBoundedControlPlaneJsonProxyAttempt(
                 writeChunk(downstream_writer, buffered_resp.body) catch return error.DownstreamWriteFailed;
             }
             downstream_writer.writeAll("0\r\n\r\n") catch return error.DownstreamWriteFailed;
-            return .{ .streamed_status = .{ .status = status_code, .upstream_addr = upstream_host } };
+            return .{ .streamed_status = .{ .status = status_code, .upstream_addr = upstream_host, .accounted = true } };
         }
 
         if (status_code != 200) {
@@ -848,7 +861,7 @@ fn deinitControlPlaneProxyTestState(state: *GatewayState) void {
     state.h2_pool.deinit();
 }
 
-test "control-plane downstream streaming write failure stays neutral for health and circuit" {
+test "control-plane downstream streaming write failure records known 200 and never retries" {
     const allocator = std.testing.allocator;
     var origin = try ControlPlaneTestOrigin.start(allocator, &.{200});
     defer origin.stop();
@@ -858,8 +871,12 @@ test "control-plane downstream streaming write failure stays neutral for health 
     initControlPlaneProxyTestState(&state, allocator);
     defer deinitControlPlaneProxyTestState(&state);
     var cfg = initControlPlaneProxyTestConfig("http://configured.example");
+    cfg.upstream_retry_idempotent_only = false;
+    cfg.upstream_retry_attempts = 3;
     const target = try std.fmt.allocPrint(allocator, "http://127.0.0.1:{d}/api", .{origin.port()});
     defer allocator.free(target);
+    state.circuit_breaker = http.circuit_breaker.CircuitBreaker.init(.{ .threshold = 1, .timeout_ms = 0, .half_open_successes = 1 });
+    state.circuitRecordFailurePermit(state.circuitTryAcquirePermit().?);
 
     try std.testing.expectError(error.DownstreamWriteFailed, executeBoundedControlPlaneJsonProxy(
         allocator,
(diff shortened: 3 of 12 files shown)


src/gateway_state.zig
```diff
@@ -2394,6 +2394,7 @@ pub const GatewayState = struct {
     }
 
     pub fn recordUpstreamFailure(self: *GatewayState, cfg: *const edge_config.EdgeConfig, upstream_base_url: []const u8) void {
+        self.circuitRecordFailure();
```
Jaruso: **[P1] Do not treat every `recordUpstreamFailure()` call as a circuit-breaker failure.** This helper is not currently an origin-only signal. For example, `executeBoundedControlPlaneJsonProxy()` catches *any* error from `executeBoundedControlPlaneJsonProxyAttempt()` and calls this method, but that attempt can return `error.UpstreamUntrusted`, URL/config errors, or `error.OutOfMemory` before any upstream contact. The buffered/streaming data-plane paths also have local errors that can reach the failure hook. With this unconditional call, a threshold of 1 can therefore open the process-wide breaker because of a local configuration/allocation failure, and subsequent clients receive `503 upstream_circuit_open` even though the origin never failed.

Please keep circuit accounting tied to classified upstream outcomes rather than piggy-backing blindly on passive-health accounting. One safe shape would be to leave `recordUpstreamFailure()` responsible for passive per-origin health and explicitly call `circuitRecordFailure()` only after confirmed upstream connect/timeout/protocol/5xx outcomes, while local capacity, policy, allocation, downstream-write, and pre-contact validation errors remain neutral. For example:

```zig
const class = classifyProxyAttemptError(err);
switch (class) {
    .upstream_failure => {
        state.recordUpstreamFailure(cfg, upstream_base_url);
        state.circuitRecordFailure();
    },
    .local_or_client_failure => {},
}
```

Please add a regression test that proves a pre-contact/local error does **not** advance/open the circuit, while a real upstream transport/5xx failure does.

tests/integration.zig
```diff
@@ -4500,6 +4500,63 @@ test "interop.openssl.h2.tls_resume" {
...
+
+    var open = try sendRequest(allocator, tardigrade.port, .{ .method = "GET", .path = "/cb", .body = null, .headers = &.{} });
+    defer open.deinit();
+    try std.testing.expectEqual(@as(u16, 503), open.status_code);
+    try assertContains(open.body, "\"code\":\"upstream_circuit_open\"");
+    compat.sleepNs(50 * std.time.ns_per_ms);
+    try std.testing.expectEqual(@as(u32, 2), upstream.requestCount());
+
+    compat.sleepNs(300 * std.time.ns_per_ms);
+    var recovered = try sendRequest(allocator, tardigrade.port, .{ .method = "GET", .path = "/cb", .body = null, .headers = &.{} });
```
Jaruso: **[P1] This recovery test does not catch the current half-open concurrency bug.** `CircuitBreaker.tryAcquire()` currently implements `.half_open => self.success_count < self.config.half_open_successes`. With the default `half_open_successes = 1`, every concurrent caller that arrives before the first probe records its result sees `success_count == 0` and is allowed through. `GatewayState.circuitTryAcquire()` holds the mutex only for the call itself, so the mutex does not reserve a probe while it is in flight. Once this PR makes the breaker live, a recovering origin can therefore receive a thundering herd instead of one probe.

Please make half-open acquisition reserve an in-flight probe and reject additional callers until that probe resolves (including a neutral/release path if the request is rejected locally before upstream contact). A minimal state-machine shape is:

```zig
.half_open => {
    if (self.half_open_probe_in_flight) return false;
    self.half_open_probe_in_flight = true;
    return true;
}
```

and clear that reservation in success/failure (or an explicit neutral release). Then extend this live test so that after the open timeout two requests overlap while the first recovery probe is deliberately held at the upstream; assert exactly one reaches the upstream and the other gets `503 upstream_circuit_open`. The existing serial `recovered` request cannot distinguish correct single-probe behavior from the current unlimited-half-open behavior.

src/gateway_handlers.zig
```diff
@@ -2264,6 +2279,11 @@ fn handleHttp3LocationProxyPass(
     };
     defer upstream_response.deinit(allocator);
     defer ctx.state.metricsReleaseProxyBufferedBytes(upstream_response.bodyLen());
+    if (upstream_response.statusCode() >= 500) {
+        ctx.state.recordUpstreamFailure(ctx.cfg, ctx.cfg.upstream_base_url);
```
Jaruso: **[P2] Absolute H3 `proxy_pass` targets are being charged to the wrong passive-health key here.** `handleHttp3LocationProxyPass()` resolves `target`, but `selection_base_url` is always `ctx.cfg.upstream_base_url`. If the location uses an absolute target such as `proxy_pass http://other-origin:9000/foo`, a 5xx here calls `recordUpstreamFailure(..., ctx.cfg.upstream_base_url)`, which can mark an unrelated configured base/pool member unhealthy. The H1 path added in this PR already recognizes this distinction and uses direct `circuitRecordFailure/Success()` for absolute targets specifically instead of mutating passive health.

Please mirror that behavior in H3. For example:

```zig
const absolute_target = gs.isAbsoluteHttpUrl(std.mem.trim(u8, target, " \t\r\n"));
if (absolute_target) {
    if (upstream_response.statusCode() >= 500) {
        ctx.state.circuitRecordFailure();
    } else if (upstream_response.statusCode() != @intFromEnum(http.Status.too_early)) {
        ctx.state.circuitRecordSuccess();
    }
} else {
    // existing recordUpstreamFailure/Success using the selected base
}
```

Please cover an H3 absolute target with a configured, different `upstream_base_url` so a failure cannot poison the wrong passive-health entry.

src/gateway_handlers.zig
```diff
@@ -2187,6 +2193,11 @@ const Http3Early425ProxyContinuation = struct {
                 self.state.metricsRecordEarlyDataRetry(.success);
             }
             try applyHttp3ProxyResponse(allocator, response, self.state, &upstream_response, self.correlation_id);
+            if (upstream_response.statusCode() >= 500) {
```
Jaruso: **[P2] Record the upstream outcome before downstream response materialization.** The preceding `try applyHttp3ProxyResponse(...)` can fail locally (notably while duplicating the response body/headers). If the upstream just supplied the half-open recovery `200` and `applyHttp3ProxyResponse` fails, these new lines never run, so the successful upstream probe is not recorded and the breaker remains in half-open/open recovery state for a downstream-side allocation failure. The main H3 path below already has the safer order: record the upstream status first, then apply the response.

Please reorder this continuation to match that path:

```zig
if (upstream_response.statusCode() >= 500) {
    self.state.recordUpstreamFailure(&self.cfg_snapshot, self.selection_base_url);
} else if (upstream_response.statusCode() != @intFromEnum(http.Status.too_early)) {
    self.state.recordUpstreamSuccess(&self.cfg_snapshot, self.selection_base_url);
}
try applyHttp3ProxyResponse(...);
```

That keeps circuit state based on the upstream outcome regardless of a later local/downstream error.

src/gateway_proxy_runtime.zig
```diff
@@ -913,6 +937,7 @@ pub fn runBufferedProxyAttempts(
                 try attempt_executor.onStaleConnectionRetry(stale_conn_retries, max_stale_conn_retries);
                 continue;
             }
+            if (err == error.CircuitOpen) return .circuit_open;
```
Jaruso: **[P2] Preserve the early-425 retry result when the retry is blocked by the circuit.** If an early-data request receives 425, waits/drives the downstream handshake, and another concurrent request opens the breaker before this request's retry, `early_425_retry_result_pending` is true here. Returning `.circuit_open` before `recordEarly425RetryResultOnce(..., .failure)` means the retry attempt is never finalized in the early-data metrics/context, unlike `ProxyBudgetExhausted` and `UpstreamAtCapacity` below.

The no-pending case is already a no-op, so simply record first:

```zig
retry_state.recordEarly425RetryResultOnce(attempt_executor, .failure);
if (err == error.CircuitOpen) return .circuit_open;
```

Please add a harness test where the second (post-425) execute returns `error.CircuitOpen` and assert the retry result is recorded as failure exactly once.

src/edge_gateway.zig
```diff
@@ -3319,16 +3319,16 @@ fn h2UnsupportedProxyOrchestration(cfg: *const edge_config.EdgeConfig, matched:
...
+    // circuit breaker depend on health/accounting that `handleLocationProxyPass`
+    // records around the low-level executor this path calls directly; skipping
+    // it means configured upstream protection silently never opens (or closes)
     // for traffic negotiated over H2. Deliberately NOT included:
     // `upstream_pool_enabled` -- it defaults to true, and connection
     // pooling itself is already correctly threaded through via
     // `state.upstream_pool`/`state.h2_pool` below, so treating it as
     // unsupported would fail closed on every H2 proxy_pass by default.
-    if (cfg.upstream_max_fails > 0) return true;
+    if (cfg.upstream_max_fails > 0 or cfg.cb_threshold > 0) return true;
```
Jaruso: **[P2] The modified H2 integration test does not exercise the new `cb_threshold` side of this condition.** `interop.h2.proxy_passive_health_tracking_fails_closed` still sets `TARDIGRADE_UPSTREAM_MAX_FAILS=3`, so `cfg.upstream_max_fails > 0` is already true. That test would pass unchanged if `or cfg.cb_threshold > 0` were deleted.

Please add/split a breaker-only H2 case with passive health disabled, e.g.:

```zig
.{ .name = "TARDIGRADE_UPSTREAM_MAX_FAILS", .value = "0" },
.{ .name = "TARDIGRADE_CB_THRESHOLD", .value = "3" },
```

and assert `h2_proxy_policy_unsupported` plus zero upstream contacts. That pins the intentional H2 fail-closed behavior specifically to the breaker configuration.

docs/OBSERVABILITY.md
```diff
@@ -399,9 +399,12 @@ Notes on the two pool-style resources called out in the issue:
...
   healthy one and returns no candidate when every backend is unhealthy (the
   request is shed deterministically rather than queued without bound); the
-  circuit breaker fast-fails once a backend trips its failure threshold and
-  recovers through a single half-open probe. Both paths have fault-injection
-  coverage in `src/gateway_state.zig`.
+  process-wide circuit breaker fast-fails proxied requests with HTTP `503` and
+  JSON `"code":"upstream_circuit_open"` once upstream failures trip its
+  configured threshold, then recovers through a half-open probe after
+  `TARDIGRADE_CB_TIMEOUT_MS`. The state machine has focused coverage in
+  `src/gateway_state.zig`, and live proxy gating is covered in
```
Jaruso: **[P2] Reconcile the canonical runbook and point "live" coverage at the actual live test.** `docs/TROUBLESHOOTING.md` still contains an explicit block saying `circuitTryAcquire` / `circuitRecord*` are *not called from the live proxy request path*, that the guide therefore does not document a breaker-open 503, and that operators should treat breaker-shaped symptoms as other causes until #627 is resolved. Merging this PR leaves the canonical troubleshooting guide directly contradicting this new text.

Please update that #627 block into the real `503 upstream_circuit_open` troubleshooting guidance as part of this PR. Also, this sentence says "live proxy gating is covered in `src/gateway_proxy_runtime.zig`", but the new real-process coverage is `tests/integration.zig`; the runtime test added here is a focused unit/harness check. Suggested wording:

```md
The state machine has focused coverage in `src/gateway_state.zig` and
`src/gateway_proxy_runtime.zig`; live open → half-open → closed behavior is
covered by `proxy.circuit_breaker_opens_fast_fails_and_recovers` in
`tests/integration.zig`.
```

That keeps the operator docs and the acceptance-criteria evidence aligned.

Jaruso (commented):
I reviewed all 9 changed files against #627's acceptance criteria and traced the breaker through H1 streaming/buffered, control-plane, H3/425 continuation, H2 policy fallback, shared outcome accounting, docs, and the new integration coverage.

The overall direction is right: the PR adds gating on the primary live attempt paths, defines `503 upstream_circuit_open`, feeds successful/failed responses back into the state machine, and adds a real-process H1 open/recovery scenario. **I would treat this as changes-required before merge** because there are still correctness holes that can make the live breaker behave differently from its contract:

1. `recordUpstreamFailure()` is not an origin-only signal today, so unconditionally coupling it to `circuitRecordFailure()` lets local/configuration failures (including pre-contact `UpstreamUntrusted` and `OutOfMemory`) trip the breaker.
2. The existing circuit-breaker state machine does not actually enforce the documented single half-open probe under concurrency; once this PR makes the breaker live, multiple requests can stampede the recovering origin.
3. H3 final-response accounting can attribute an absolute `proxy_pass` target's health result to `cfg.upstream_base_url`, unlike the H1 absolute-target handling added here.
4. The parked H3 425 continuation records the upstream result only after downstream response materialization, so a local `applyHttp3ProxyResponse` failure can strand/mis-state the breaker.
5. A circuit-open race during an early-425 retry bypasses the pending retry-result metric.
6. The new H2 `cb_threshold` branch is not actually isolated by the modified integration test because `upstream_max_fails > 0` already forces the same unsupported path.
7. The canonical `docs/TROUBLESHOOTING.md` still explicitly says the circuit breaker is not wired into live proxy traffic and points operators to #627; this PR would make that statement stale immediately. The new OBSERVABILITY text also points to a unit file for "live" coverage instead of the real-process integration test.

Please also add focused regression coverage for the control-plane `CircuitOpen` path/error mapping while addressing the failure-classification issue. CI on this head is not complete: build/format/security/etc. jobs passed, but the unit-test jobs were cancelled and the integration job was skipped, so the new live integration test has not been demonstrated by this Actions run.

GitHub does not permit the PR author's own account to submit a formal `REQUEST_CHANGES` review, so this is submitted as `COMMENT`; the P1/P2 inline findings below are still intended as pre-merge changes.

src/http/circuit_breaker.zig
```diff
@@ -58,26 +60,40 @@ pub const CircuitBreaker = struct {
...
+                break :blk true;
+            },
         };
     }
 
+    /// Release an admitted half-open probe when the request did not reach the
+    /// upstream and therefore has no success/failure outcome to record.
+    pub fn releaseProbe(self: *CircuitBreaker) void {
+        if (self.config.threshold == 0) return;
+        if (self.state == .half_open) self.half_open_probe_in_flight = false;
```
Jaruso: **[P1] The half-open permit needs request ownership; this global boolean can be released or completed by the wrong in-flight request.** Consider this legal ordering with a short CB timeout: request A is admitted while `closed` and is slow; request B fails and opens the breaker; after the timeout request C transitions the breaker to `half_open` and owns the recovery probe; then A finishes. If A succeeds, `recordSuccess()` sees the *current* state as `half_open` and closes the breaker/clears C's reservation even though A was never the recovery probe. If A instead hits any neutral/local path that calls `releaseProbe()`, this line clears C's reservation and a second concurrent probe can enter.

There is a second failure mode from the same missing ownership: a legitimate half-open probe that gets `HttpConnectionClosing` (the stale pooled-connection retry path) or an RFC8470 `425` keeps `half_open_probe_in_flight = true`, then the same logical request loops and calls `tryAcquire()` again. It rejects itself as `CircuitOpen`; the original reservation is never consumed, so recovery can wedge.

Please make acquisition return a request-scoped permit/lease (an enum/token with a half-open generation/epoch is enough) and require that permit on success/failure/release. A permit acquired while `closed` must not be able to complete/release a later half-open probe, and a half-open permit should be retained across neutral stale/425 retries rather than reacquired. For example, conceptually:

```zig
const Permit = union(enum) { ordinary, half_open: u64 };

pub fn tryAcquire(self: *CircuitBreaker) ?Permit { ... }
pub fn recordSuccess(self: *CircuitBreaker, permit: Permit) void { ... }
pub fn recordFailure(self: *CircuitBreaker, permit: Permit) void { ... }
pub fn release(self: *CircuitBreaker, permit: Permit) void { ... }
```

Then carry the permit in the production H1/H3 attempt executor/parked continuation until a definitive upstream outcome or true neutral release. Please add regression coverage for (a) an old closed-state request completing while another request owns the half-open probe, and (b) a half-open probe surviving a neutral stale-connection or 425 retry without self-blocking.

src/gateway_proxy_runtime.zig
```diff
@@ -1098,11 +1168,19 @@ const ProductionBufferedProxyAttemptExecutor = struct {
...
+    }
+
     fn onTerminalAttemptError(
         self: *ProductionBufferedProxyAttemptExecutor,
-        _: anyerror,
+        err: anyerror,
     ) !void {
-        self.state.recordUpstreamFailure(self.cfg, self.selection_base_url);
+        if (proxyAttemptErrorCountsAsUpstreamFailure(err)) {
+            self.state.recordProxyUpstreamFailure(self.cfg, self.selection_base_url);
```
Jaruso: **[P2] Absolute H1 targets still poison passive health on transport errors and retryable 5xxs.** The final-response path in this PR correctly checks `isAbsoluteHttpUrl(target)` and uses `circuitRecord*()` only, because an absolute `proxy_pass` bypasses `selection.base_url`. But this buffered error hook always calls `recordProxyUpstreamFailure(self.cfg, self.selection_base_url)`, and `onConfigured5xxRetry()` below does the same. The streaming catch path has the same mismatch for `RequestCancelled` and other upstream errors.

Example: `upstream_base_url/pool = http://origin-a` and `location /x { proxy_pass http://origin-b/x; }`. `resolveProxyTarget()` sends the request to origin B, but a connection failure here increments passive health for origin A. H3 now carries `absolute_target` through its executor and correctly does circuit-only accounting; H1 should mirror that.

Please compute `absolute_target` once before entering the streaming/buffered branches, carry it on `ProductionBufferedProxyAttemptExecutor`, and branch all definitive upstream failures/retry-5xx outcomes the same way as the final response:

```zig
if (self.absolute_target) {
    self.state.circuitRecordFailure();
} else {
    self.state.recordProxyUpstreamFailure(self.cfg, self.selection_base_url);
}
```

Do the equivalent in the streaming catch path, and add a regression with a deliberately different configured base/pool and absolute target proving a target failure cannot mark the configured base unhealthy.

tests/integration.zig
```diff
@@ -4500,6 +4500,97 @@ test "interop.openssl.h2.tls_resume" {
...
+    try std.testing.expect(probe_a.body_matched);
+    const ok_a: u8 = if (probe_a.status == 200) 1 else 0;
+    const ok_b: u8 = if (probe_b.status_code == 200) 1 else 0;
+    const open_a: u8 = if (probe_a.status == 503) 1 else 0;
+    const open_b: u8 = if (probe_b.status_code == 503) 1 else 0;
+    const ok_count: u8 = ok_a + ok_b;
+    const open_count: u8 = open_a + open_b;
+    try std.testing.expectEqual(@as(u8, 1), ok_count);
+    try std.testing.expectEqual(@as(u8, 1), open_count);
+    try std.testing.expectEqual(@as(u32, 3), upstream.requestCount());
```
Jaruso: **[P2] This still does not prove the acceptance criterion's final `half-open -> closed` transition on the live path.** The race now proves that exactly one recovery probe reaches the upstream and the concurrent request gets 503, which is good. But the test ends immediately after the probe's 200. An implementation that returns that 200 while accidentally leaving the breaker half-open (or leaving its reservation set) would still pass.

Please give the upstream one more `200` response and, after `thread_a.join()`, immediately send another `/cb` request without waiting another breaker timeout. Assert it gets 200 and `upstream.requestCount()` becomes 4. That directly proves the successful half-open probe closed the breaker and normal traffic resumes, completing the issue's required live `open -> half-open -> closed` coverage.

Jaruso (commented):
Second full review against #627. The original seven findings have been addressed in the current head, but I found three remaining correctness/acceptance gaps that I would treat as blocking before merge:

1. **P1: the half-open reservation is not request-owned.** A global `half_open_probe_in_flight` boolean can be released/completed by an older request that was admitted while the breaker was still closed, and the actual half-open probe can also block itself on neutral stale/425 retries. The permit needs request/attempt ownership (ideally a token/epoch/lease) that survives neutral retries and is consumed only by that request's definitive outcome.
2. **P2: H1 absolute `proxy_pass` failures/retry-5xx outcomes still mutate passive health for `selection.base_url`.** The final-response path correctly uses circuit-only accounting for absolute targets, but the streaming catch and buffered executor error/retry hooks do not. That can mark an unrelated configured pool member unhealthy when the absolute target fails.
3. **P2: the live integration test still does not prove the final `closed` transition.** It proves open -> one half-open probe + concurrent fast-fail, but stops immediately after the probe returns 200. Add a post-probe request that succeeds immediately and reaches the upstream to pin half-open -> closed.

CI on this head is also still incomplete: build/format/lint/security jobs pass, but the unit-test jobs were cancelled and the integration job was skipped. After the code fixes, please get a complete test run (or at minimum the focused unit/integration suites for these paths) before merging.

GitHub will not let this account submit `REQUEST_CHANGES` on its own PR, so this is filed as a COMMENT review, but the P1/P2 items above are change requests.

src/gateway_handlers.zig
```diff
@@ -2062,7 +2126,9 @@ const Http3Early425ProxyContinuation = struct {
             .forwarded_proto = forwarded_proto,
             .incoming_host = incoming_host,
             .selection_base_url = executor.selection_base_url,
+            .absolute_target = executor.absolute_target,
             .budget_start_ms = executor.budget_start_ms,
+            .circuit_permit = executor.circuit_permit,
```
Jaruso: **[P1] A parked half-open permit can be abandoned here and wedge recovery permanently.** This transfers `executor.circuit_permit` into the parked continuation, but `Http3Early425ProxyContinuation.deinit()` never releases an outstanding permit. There are at least two live exits where that matters: (1) the client/QUIC stream disappears while the retry is parked, so the registered `deinit_fn` destroys the continuation without `run()` ever consuming the permit; and (2) `run()` reaches `perAttemptTimeoutMs()` after the request sat parked long enough to exhaust the total retry budget—the catch returns a 504 before `execute()` and never releases `self.circuit_permit`. If this is the half-open permit, `half_open_probe_in_flight` remains true forever and every later request is rejected; no request can become the next recovery probe.

Please make the continuation the explicit owner and give destruction a release backstop. For example:

```zig
fn deinit(self: *Http3Early425ProxyContinuation, allocator: std.mem.Allocator) void {
    if (self.circuit_permit) |permit| {
        self.state.circuitReleasePermit(permit);
        self.circuit_permit = null;
    }
    self.config_lease.release();
    // existing frees...
}
```

Also make the transfer one-way after parking succeeds so the executor does not conceptually retain duplicate ownership:

```zig
try self.request.parkEarly425Retry(...);
self.circuit_permit = null; // continuation now owns it
```

and either explicitly release/null in the pre-attempt budget-error path or rely on the destructor only if that continuation is guaranteed to be destroyed immediately afterward. Please add two regressions: park a **half-open** 425 retry and destroy/cancel it before resume, then assert another acquire is admitted as the probe; and park it until the retry budget expires before `execute`, then assert the permit is released and recovery is not wedged.

src/http/circuit_breaker.zig
```diff
@@ -45,62 +54,111 @@ pub const CircuitBreaker = struct {
...
+            .half_open => blk: {
+                if (self.half_open_probe_in_flight) break :blk null;
+                self.half_open_probe_in_flight = true;
+                break :blk Permit{ .half_open = self.half_open_generation };
             },
-            .half_open => self.success_count < self.config.half_open_successes,
         };
     }
 
+    pub fn tryAcquire(self: *CircuitBreaker) bool {
```
Jaruso: **[P2] The legacy boolean API is now broken because it discards the permit needed to resolve half-open state.** `tryAcquire()` can transition `open -> half_open`, set `half_open_probe_in_flight = true`, and then throw away the returned `.half_open(generation)` token. The existing no-argument `recordSuccess()` / `recordFailure()` wrappers now pass `.ordinary`, which are intentionally ignored while the breaker is half-open. So this formerly valid public pattern now wedges forever:

```zig
cb.recordFailure(); // opens
if (cb.tryAcquire()) { // transitions to half-open and discards the permit
    cb.recordSuccess(); // .ordinary => no-op in half-open
}
// cb stays half-open with its probe reserved
```

`releaseProbe()` has the inverse ownership problem: it fabricates a permit from the *current* generation, so a caller with no token can release somebody else's live probe. `GatewayState.circuitTryAcquire()` / `circuitRecordSuccess()` / `circuitRecordFailure()` / `circuitReleaseProbe()` retain the same problematic compatibility surface (and `circuitReleaseProbe()` currently passes `.ordinary`, so it cannot release a half-open probe at all).

The permit design is the right concurrency fix, but the old API cannot safely emulate it. Please remove/private/deprecate the no-permit request-lifecycle methods and migrate callers/tests to the permit API, rather than silently preserving names with incompatible semantics. If these wrappers must remain public for compatibility, they need a design that actually preserves the acquired token through completion—which a bare `bool` cannot do safely under concurrency. Add a regression for whatever public contract remains so `open -> acquire -> success` cannot silently become a permanent half-open state.

Jaruso (commented):
Re-reviewed the full current head (`570a79f`) against #627 and all prior review threads. The three blockers from the previous round are addressed: half-open permits are now request-owned across normal retries, H1 absolute `proxy_pass` failures no longer poison an unrelated passive-health entry, and the live integration test now proves traffic resumes after the successful half-open probe. The current CI run is also green across the unit/appliance matrices.

I found two remaining permit-ownership problems that should be fixed before merge:

1. **P1:** a parked H3 early-425 continuation can be destroyed or time out before its resumed upstream attempt without releasing the half-open permit it owns, leaving the breaker permanently wedged in `half_open` with `half_open_probe_in_flight=true`.
2. **P2:** the retained public boolean/no-permit compatibility API (`tryAcquire` + `recordSuccess`/`recordFailure`/`releaseProbe`, and the corresponding `GatewayState` wrappers) is no longer semantically valid. `tryAcquire()` can reserve a half-open permit and discard its token, while `recordSuccess()`/`recordFailure()` use `.ordinary` and therefore cannot resolve that probe.

Once those ownership holes are closed, I do not see another acceptance-criteria gap in #627 from this pass. As before, GitHub does not permit the PR author account to submit a formal REQUEST_CHANGES review, so this is submitted as COMMENT with the blockers called out inline.

src/http/circuit_breaker.zig
```diff
@@ -45,59 +54,92 @@ pub const CircuitBreaker = struct {
...
 
-        self.last_failure_ns = compat.nanoTimestamp();
-        switch (self.state) {
-            .closed => {
-                self.failure_count += 1;
-                if (self.failure_count >= self.config.threshold) {
-                    self.state = .open;
+        switch (permit) {
+            .ordinary => {
+                if (self.state == .closed) {
```
Jaruso: **[P1] Tag ordinary permits with the breaker generation too.** The half-open token now has request ownership, but `.ordinary` is still generationless. A request admitted in the old closed state can finish after a later half-open probe has already succeeded and closed the breaker; at that point this `if (self.state == .closed)` accepts the stale outcome as if it belonged to the recovered generation.

With threshold 1, this is deterministic:

```zig
const stale = cb.tryAcquirePermit().?; // old .ordinary request starts
cb.recordFailurePermit(.ordinary);     // another request opens the breaker
const probe = cb.tryAcquirePermit().?; // half-open generation 1
cb.recordSuccessPermit(probe);         // half-open -> closed
cb.recordFailurePermit(stale);         // re-opens the recovered breaker today
```

The symmetric success path has the same cross-generation problem: an old pre-open request can finish successfully after recovery and reset failures accumulated by newer post-recovery requests.

Please give closed-state permits an epoch/generation too and apply success/failure only when that epoch is still current, e.g. conceptually:

```zig
const Permit = union(enum) {
    closed: u64,
    half_open: u64,
};
```

Increment/invalidate the closed generation when a breaker cycle starts so stale pre-open outcomes cannot mutate a later recovered generation. Please add regressions for both a stale failure after successful recovery (must not re-open) and a stale success after recovery (must not clear newer failures).

src/gateway_control_plane_proxy.zig
```diff
@@ -337,7 +353,11 @@ pub fn executeBoundedControlPlaneJsonProxy(
                 enable_streaming_success,
                 sticky_set_cookie,
             ) catch |err| {
-                state.recordUpstreamFailure(cfg, upstream_base_url);
+                if (controlPlaneAttemptErrorCountsAsUpstreamFailure(err)) {
```
Jaruso: **[P1] This type-based catch still charges downstream write failures to the upstream circuit.** `executeBoundedControlPlaneJsonProxyAttempt()` performs the upstream request first, but on the streamed-response path it then calls `writeStreamedUpstreamResponse`, `writeChunk`, and `downstream_writer.writeAll`. Those helpers can return downstream/client writer errors. Any such error not in this short deny-list falls through to `true` here, so a client disconnect/downstream write failure calls `recordProxyUpstreamFailure()` and can open the process-wide breaker even though the origin already returned successfully.

That directly contradicts the new operator contract in `TIMEOUTS.md` / `TROUBLESHOOTING.md` that downstream write failures do not trip the breaker.

Please classify by provenance/stage rather than treating unknown `anyerror` values as origin failures. One safe shape is to normalize downstream serialization failures to a dedicated neutral error before they escape the attempt:

```zig
writeStreamedUpstreamResponse(...) catch return error.DownstreamWriteFailed;
// same for body/chunk writes
```

and explicitly release the permit / avoid passive-health accounting for `DownstreamWriteFailed`. Alternatively, record the already-known upstream outcome before downstream materialization and structure the return so the outer switch cannot double-count it.

Please add a regression with an upstream `200` and a downstream writer that fails while streaming; assert passive health is unchanged and the breaker remains closed.

src/gateway_control_plane_proxy.zig
```diff
@@ -349,15 +369,15 @@ pub fn executeBoundedControlPlaneJsonProxy(
         switch (exec) {
             .streamed_status => |streamed| {
                 if (streamed.status >= 500) {
-                    state.recordUpstreamFailure(cfg, upstream_base_url);
+                    state.recordProxyUpstreamFailure(cfg, upstream_base_url, circuit_permit);
```
Jaruso: **[P2] Absolute control-plane `proxy_pass` targets still mutate passive health for the selected configured base.** `resolveProxyTarget()` explicitly allows an absolute `proxy_pass_target` to bypass `upstream_base_url`, but every success/failure path here records against `upstream_base_url`. If configured/pool origin A is selected while the absolute target actually sends the request to B, B's 5xx/transport failure can mark A unhealthy, and B's success can incorrectly rehabilitate A. H1 and H3 now correctly use circuit-only accounting for absolute targets; the control-plane path should match them.

Compute the absolute-target distinction once and branch **all** definitive outcomes (streamed, buffered, and classified terminal errors), for example:

```zig
if (absolute_target) {
    state.circuitRecordFailurePermit(circuit_permit);
} else {
    state.recordProxyUpstreamFailure(cfg, upstream_base_url, circuit_permit);
}
```

and likewise for success. Please add a regression with configured origin A and absolute target B proving B cannot poison or clear A's passive-health entry while B's real upstream outcome still affects the process-wide circuit.

Jaruso (commented):
Re-reviewed current head `a61c99c0` against #627 and all prior review threads.

The previous blockers are substantially addressed: half-open ownership is now permit-scoped, the unsafe legacy boolean lifecycle APIs are gone, parked H3 425 permits have destruction/budget-release backstops, H1/H3 absolute targets no longer poison passive health, early-425 retry metrics are finalized correctly, the breaker-only H2 fail-closed branch is pinned, the operator docs are reconciled, and the live integration now proves open -> one in-flight half-open probe -> closed with normal traffic resuming afterward. Current CI run `32476280622` is green, including unit tests and the appliance-profile full integration suites.

I found three remaining correctness gaps that should be fixed before merge:

1. **[P1]** Ordinary permits are still generationless, so a request admitted before an open/recovery cycle can later mutate the newly recovered closed generation.
2. **[P1]** Control-plane streamed downstream write failures still flow through the broad attempt-error classifier and can trip passive health / the process-wide breaker even after the upstream succeeded.
3. **[P2]** Control-plane absolute `proxy_pass` targets still attribute passive health to the selected configured base rather than using circuit-only accounting for the actual absolute target.

I left concrete reproductions and fix shapes inline. These are the remaining places I found where outcome ownership/provenance can cross breaker generations or be attributed to the wrong side/origin.

src/gateway_control_plane_proxy.zig
```diff
@@ -337,7 +384,11 @@ pub fn executeBoundedControlPlaneJsonProxy(
                 enable_streaming_success,
                 sticky_set_cookie,
             ) catch |err| {
-                state.recordUpstreamFailure(cfg, upstream_base_url);
+                if (controlPlaneAttemptErrorCountsAsUpstreamFailure(err)) {
+                    recordControlPlaneProxyFailure(state, cfg, upstream_base_url, absolute_proxy_target, circuit_permit);
+                } else {
+                    state.circuitReleasePermit(circuit_permit);
```
Jaruso: **[P1] Do not release-and-retry after `DownstreamWriteFailed`; the upstream outcome is already definitive here.** The new error normalization correctly prevents a downstream writer failure from being *counted as* an upstream failure, but it goes one step too far: `executeBoundedControlPlaneJsonProxyAttempt()` fully buffers `buffered_resp` and knows `status_code` **before** it calls `writeStreamedUpstreamResponse` / `writeChunk` / the terminal chunk write. So when one of those downstream writes returns `error.DownstreamWriteFailed`, this branch throws away a real upstream outcome by releasing the permit.

That breaks recovery in both directions. If this request owns the half-open permit and the origin returned `200`, `releasePermit()` leaves the breaker half-open instead of closing on the successful probe. If `proxy_stream_all_statuses` is enabled and the origin returned a `5xx`, the same path releases the half-open permit instead of re-opening the breaker. Relative targets also fail to apply the already-known success/failure to passive health.

There is also a more serious retry consequence immediately below: when `upstream_retry_idempotent_only=false` and `upstream_retry_attempts > 1`, `DownstreamWriteFailed` reaches `continue`, so the gateway can send the **same control-plane POST to the origin again after the first POST already completed successfully and only the client write failed**. That can duplicate side effects, and there is no useful downstream response left to retry anyway.

Please separate the upstream outcome boundary from downstream delivery. Once `buffered_resp.status_code` is known, consume the permit exactly once from that status, then treat a downstream serialization/write failure as terminal and never re-run the upstream attempt. For example, structurally:

```zig
// upstream request completes and status is known
const upstream_result = try execute...;
recordControlPlaneOutcome(state, cfg, upstream_base_url, absolute_target,
    circuit_permit, upstream_result.status);

// only after breaker/passive-health accounting is finalized
writeControlPlaneResult(downstream_writer, upstream_result) catch
    return error.DownstreamWriteFailed; // terminal: no upstream retry
```

If keeping the current function split, return enough state to tell the outer layer that the permit/outcome was already consumed, and special-case `DownstreamWriteFailed` before the retry `continue`.

Please extend the regression coverage with a **half-open** circuit whose origin returns `200` while the downstream writer fails: assert the breaker becomes `closed` despite the client-side failure. Also set `upstream_retry_idempotent_only=false` and `upstream_retry_attempts=2+` and assert the test origin sees exactly **one** POST. A `proxy_stream_all_statuses=true` + upstream `500` variant should prove the half-open probe re-opens even if downstream delivery fails.

Jaruso (commented):
Re-reviewed current head `fca15fc7` against #627 and the previous review rounds.

The three findings from the last pass are fixed: closed-state permits are now generation-scoped with stale-success/stale-failure regressions; absolute control-plane targets use circuit-only accounting instead of mutating the configured origin's passive health; and downstream writer errors are explicitly separated from upstream transport errors. The earlier H1/H3/H2/docs/live-recovery fixes also remain intact. Current CI run `32483112105` is green, including the full unit suite (`3854/3855`, one intentional skip) and the appliance-profile full integration suites on Linux x64/ARM.

I found one remaining P1 in the new downstream-write handling. The control-plane attempt has already fully received the upstream response before it starts writing the streamed response to the downstream client. If that downstream write fails, the outer catch currently releases the permit and—when configured retries are enabled—continues the upstream retry loop. That loses a definitive upstream outcome (so a successful half-open probe does not close, and a known 5xx does not re-open) and can replay a completed control-plane POST after the origin already processed it.

I left the concrete state transitions, retry consequence, and a fix/test shape inline. Other than this outcome-boundary issue, I did not find another remaining blocker in the current diff.

src/gateway_state.zig
```diff
@@ -1381,22 +1381,28 @@ pub const GatewayState = struct {
         };
     }
 
-    pub fn circuitTryAcquire(self: *GatewayState) bool {
+    pub fn circuitTryAcquirePermit(self: *GatewayState) ?http.circuit_breaker.Permit {
         self.circuit_mutex.lock();
         defer self.circuit_mutex.unlock();
-        return self.circuit_breaker.tryAcquire();
+        return self.circuit_breaker.tryAcquirePermit();
```
Jaruso: **[P1] Keep the process-owned breaker configuration coherent across hot reloads.** `GatewayState.circuit_breaker` is constructed once in `edge_gateway.run()` from the startup `cfg.cb_threshold` / `cfg.cb_timeout_ms`, and every live gate here consults that stored object. But `hotReloadConfig()` installs a new `EdgeConfig` and `applyReloadedRuntimeConfig()` never reconfigures this breaker or rejects changes to those two fields.

That creates exactly the runtime/config drift #627 is intended to eliminate. For example:

```text
startup: TARDIGRADE_CB_THRESHOLD=0   -> state breaker disabled
reload:  TARDIGRADE_CB_THRESHOLD=1   -> active EdgeConfig says enabled
```

After the reload, H1/H3/control-plane calls through this method still admit everything because the process-owned breaker retains threshold 0. The inverse is worse operationally: start with threshold 1, reload to 0, and the published config says the breaker is disabled while the old process state can continue opening/fast-failing requests. `CB_TIMEOUT_MS` changes are silently ignored too. The changed H2 policy makes the divergence visible across protocols because it reads `cfg.cb_threshold` directly, while these request paths read the stale process object.

Please either support breaker reconfiguration under `circuit_mutex` with explicit generation/state semantics, or make these fields restart-only and reject a reload that changes them (like the existing replay/listener restart-only gates). If reconfiguring live, do **not** simply assign a fresh breaker whose generation restarts at 1 while old permits are still in flight; old permits must be invalidated safely.

Please add reload regressions covering at least disabled→enabled and enabled→disabled (or assert that both changes are rejected if restart-only), plus timeout changes.

src/gateway_proxy_runtime.zig
```diff
@@ -632,16 +671,25 @@ pub fn handleLocationProxyPass(
                 "",
                 transcript_redactions,
             );
-            if (!isAbsoluteHttpUrl(std.mem.trim(u8, target, " \t\r\n"))) {
+            if (absolute_proxy_target) {
+                if (streamed.status_code >= 500 or streamed.upstream_aborted) {
```
Jaruso: **[P1] Check `local_capacity_aborted` before `upstream_aborted` for absolute streaming targets.** In the H2 streaming relay, a local response-buffer capacity failure sets **both** `upstream_aborted = true` and `local_capacity_aborted = true`. The relative-target branch below correctly gives `local_capacity_aborted` precedence and releases the permit neutrally, but this absolute-target branch tests `streamed.upstream_aborted` first, so local memory pressure is recorded as a process-wide circuit failure.

A healthy absolute origin returning (for example) 200 can therefore trip the global breaker solely because Tardigrade ran out of local relay-buffer capacity; with a low threshold, unrelated proxy traffic then receives `503 upstream_circuit_open`. That contradicts this PR's documented/operator contract that local capacity failures do not trip the breaker.

Please classify the relay outcome once before branching on passive-health attribution, e.g.:

```zig
if (streamed.local_capacity_aborted) {
    state.circuitReleasePermit(circuit_permit);
} else if (streamed.status_code >= 500 or streamed.upstream_aborted) {
    if (absolute_proxy_target) {
        state.circuitRecordFailurePermit(circuit_permit);
    } else {
        state.recordProxyUpstreamFailure(cfg, selection.base_url, circuit_permit);
    }
} else {
    // success, circuit-only for absolute; combined for relative
}
```

Please add a regression for an absolute target whose streaming result has `local_capacity_aborted=true` (and therefore `upstream_aborted=true`) and assert the breaker remains closed / the permit is released rather than failed.

src/gateway_control_plane_proxy.zig
```diff
@@ -45,6 +46,50 @@ pub fn controlPlaneBufferedResponseLimit(cfg: *const edge_config.EdgeConfig) usi
         DEFAULT_CONTROL_PLANE_BUFFERED_RESPONSE_LIMIT;
 }
 
+fn controlPlaneAttemptErrorCountsAsUpstreamFailure(err: anyerror) bool {
+    return switch (err) {
```
Jaruso: **[P1] `OutOfMemory` can also occur after the upstream POST has completed, so treating every OOM as neutral/retryable can duplicate the request.** This is the same phase-boundary problem as the existing `DownstreamWriteFailed` finding, but a different error class that the requested fix needs to cover.

Inside `executeBoundedControlPlaneJsonProxyAttempt()`, `buffered_resp.status_code` is known before several response-materialization allocations: duplicating `Location`, content type/disposition/cookie, the 200 body, etc. If any of those allocations returns OOM, this classifier returns false, the outer catch releases the permit, and—when retries are configured—continues the loop. The origin may therefore receive the same control-plane POST again even though the first POST already completed and its status was known. A half-open 200/500 is also discarded instead of closing/re-opening the breaker from the real origin result.

Please make the fix structural rather than adding more error names to this deny-list: once `buffered_resp.status_code` exists, record that upstream outcome exactly once and mark subsequent local response materialization/delivery errors as terminal with **no upstream retry**. Pre-contact/pre-status OOM can remain neutral.

A regression using a failing allocator after a test origin returns 200 should assert (1) the origin sees exactly one POST and (2) a half-open probe closes despite the local materialization failure; a 500 variant should re-open.

src/gateway_proxy_runtime.zig
```diff
@@ -552,6 +574,7 @@ pub fn handleLocationProxyPass(
             ) catch |err| {
                 state.recordUpstreamAttemptEnd(selection.base_url);
                 if (err == error.ClientAborted) {
+                    state.circuitReleasePermit(circuit_permit);
```
Jaruso: **[P2] `ClientAborted` is phase-ambiguous here and can erase a known upstream 5xx.** `executeStreamingHttpProxyRequest()` uses `error.ClientAborted` both for genuinely client-side request/upload failures (neutral is correct) **and** for downstream response-head/body write failures after the upstream response status has already been parsed. This unconditional release therefore loses a definitive origin failure in the latter case.

Concrete half-open case: the origin returns a 500 response head, Tardigrade starts serializing that response, the downstream socket closes, and the helper surfaces `ClientAborted`. This branch releases the half-open permit instead of recording the known 500, leaving the breaker half-open and allowing another probe rather than re-opening immediately.

Please preserve provenance from the streaming executor rather than using one error tag for both phases. One workable shape is to return/propagate whether an upstream status was observed (and its code) alongside the downstream delivery failure, so a known 5xx consumes the permit as a failure while a pre-response client upload/read abort remains neutral. Add regressions for both sides: pre-response client abort does not affect the breaker; upstream 500 + downstream write abort does.

Jaruso (commented):
Re-reviewed the full PR against #627 on the same head (`fca15fc7`; there have been no code changes since review #4993774281). The existing P1 about `DownstreamWriteFailed` remains valid. A fresh pass found four additional correctness gaps, all around keeping circuit state tied to the actual upstream outcome rather than local runtime/config state.

The live open → half-open → closed integration coverage, single-probe ownership/generation work, H1/H3 absolute-target passive-health attribution, H2 breaker-only fail-closed coverage, docs, and current CI all still look good. CI is green, including the full appliance-profile integration suites, but none of the cases below are currently pinned by tests.

The most important common theme for the remaining fixes is to make the breaker outcome boundary explicit: once a request has a definitive upstream status, consume its permit from that status exactly once; local response delivery/materialization failures must neither erase that outcome nor cause a second upstream delivery. Likewise, the process-owned breaker configuration must stay coherent with the active reloaded `EdgeConfig`.

src/gateway_control_plane_proxy.zig
```diff
@@ -336,8 +384,15 @@ pub fn executeBoundedControlPlaneJsonProxy(
                 state,
                 enable_streaming_success,
                 sticky_set_cookie,
+                absolute_proxy_target,
+                circuit_permit,
             ) catch |err| {
-                state.recordUpstreamFailure(cfg, upstream_base_url);
+                if (err == error.DownstreamWriteFailed) return err;
```
Jaruso: **[P1] The downstream-write case is fixed here, but the post-response OOM finding is still unresolved.** `executeBoundedControlPlaneJsonProxyAttempt()` can return `OutOfMemory` *after* `buffered_resp.status_code` is known while duplicating `Location`, content type/disposition/cookie, or the buffered 200 body. `controlPlaneAttemptErrorCountsAsUpstreamFailure(error.OutOfMemory)` is false, so this catch still releases the permit and the retry block below can resend the completed control-plane POST. The structural boundary needs to cover every local error after upstream status is known, not only `DownstreamWriteFailed`: record the known status once, then make all response-materialization/delivery failures terminal. Please add a failing-allocator regression proving one POST and correct half-open close/re-open behavior.

src/gateway_state.zig
Jaruso: **[P1] Still outstanding on `9355c0ea`: hot reload can publish different `cb_threshold` / `cb_timeout_ms` while this process-owned breaker retains the startup configuration.** The latest commit does not touch the reload/state path. Please either safely reconfigure under `circuit_mutex` with generation invalidation, or reject breaker-setting changes as restart-only, with reload regressions for enable/disable and timeout changes.

src/gateway_proxy_runtime.zig
Jaruso: **[P1] Still outstanding on `9355c0ea`: local streaming capacity can be counted as an upstream failure for absolute targets.** The H2 relay can return both `local_capacity_aborted=true` and `upstream_aborted=true`; this absolute-target branch still tests the latter first. Give `local_capacity_aborted` precedence before target attribution so local memory pressure releases the permit neutrally instead of tripping the process-wide breaker.

src/gateway_proxy_runtime.zig
Jaruso: **[P2] Still outstanding on `9355c0ea`: `ClientAborted` remains phase-ambiguous.** Streaming uses this error both before any upstream status exists and for downstream response-write failure after a status has been parsed. This unconditional release can therefore erase a known upstream 5xx/half-open failure. Preserve whether/status of an observed upstream response so a post-status downstream abort records that real outcome, while a pre-response client abort remains neutral.

Jaruso (commented):
Re-reviewed current head `9355c0ea` against the five outstanding findings from the prior rounds and the #627 acceptance criteria.

The latest commit **does fix the original control-plane downstream-write finding**: once the upstream status is known on the streamed path, the code records that outcome before downstream delivery, marks the returned streamed result as already accounted, and treats `DownstreamWriteFailed` as terminal instead of retrying. The new half-open 200 and streamed-500 regressions both configure multiple retry attempts and assert the origin receives exactly one POST, which covers the failure mode I requested.

However, the four findings from review #4994286134 are still outstanding:

1. **P1 hot-reload breaker drift** — `cb_threshold` / `cb_timeout_ms` are still copied into the process-owned breaker only at startup; the latest commit does not touch reload/state configuration.
2. **P1 absolute streaming local-capacity attribution** — the absolute-target streaming branch still checks `upstream_aborted` before `local_capacity_aborted`, so a local relay-capacity abort can trip the global breaker.
3. **P1 post-response OOM retry** — the new terminal guard special-cases only `DownstreamWriteFailed`. OOM can still happen after `buffered_resp.status_code` is known while duplicating response metadata/body; that path still releases the permit and can retry the already-completed control-plane POST.
4. **P2 streaming `ClientAborted` provenance** — the data-plane streaming helper still uses `ClientAborted` for both pre-response client failures and downstream response-write failures after a status is known, and the caller still releases the permit unconditionally.

I did not find a new regression introduced by `9355c0ea` itself. CI run #1552 is green on this head, but these four logic gaps remain blockers for calling the implementation complete.

src/gateway_proxy_runtime.zig
```diff
@@ -632,18 +703,17 @@ pub fn handleLocationProxyPass(
...
-                // for health purposes (#140).
-                if (streamed.local_capacity_aborted) {
-                    state.logger.warn(correlation_id, "streaming relay truncated: local proxy buffer capacity exhausted", .{});
-                } else if (streamed.status_code >= 500 or streamed.upstream_aborted) {
-                    state.recordUpstreamFailure(cfg, selection.base_url);
-                } else {
-                    state.recordUpstreamSuccess(cfg, selection.base_url);
-                }
-            }
+            recordStreamingProxyOutcome(
```
Jaruso: **[P2] Preserve the downstream abort after recording the known upstream outcome.** The new transport result correctly fixes the earlier circuit-attribution bug by returning `status_code` plus `downstream_aborted_after_status=true` instead of throwing `ClientAborted` after response headers exist. But this caller never checks that flag. After this accounting call it continues as if delivery succeeded: `metricsRecordProxyClientAbort()` is never called, normal transcript/status/streaming metrics are emitted, and `handleLocationProxyPass` returns the upstream status instead of propagating the broken downstream connection.

Please keep this status accounting first, then restore the client-abort lifecycle, e.g. conceptually:

```zig
recordStreamingProxyOutcome(...);

if (streamed.downstream_aborted_after_status) {
    state.metricsRecordProxyClientAbort();
    return error.ClientAborted;
}

// normal transcript/status/streaming completion accounting
```

If you intentionally want the transcript to retain the observed upstream status, that can happen before the return, but the downstream write failure still needs to reach the connection layer and should not be counted as a normally delivered response. Add a regression where an upstream 500 is known, the downstream writer fails, and assert both: (1) the half-open breaker re-opens from the 500, and (2) the request still surfaces `ClientAborted` / increments the client-abort metric exactly once. Keep the existing pre-response client-abort regression neutral for the breaker.

Jaruso (commented):
Re-reviewed current head `dc74094c` against issue #627 and every outstanding finding from the previous rounds.

The four findings still open on `9355c0ea` are now substantively fixed:
- breaker threshold/timeout reload drift is prevented by rejecting those changes as restart-only, with comparator coverage for both threshold directions and timeout changes;
- streaming local-capacity aborts are classified before `upstream_aborted`, so absolute targets no longer trip the process-wide breaker from local memory pressure;
- control-plane status is now accounted before post-status response materialization, those failures are terminal/no-retry, and 200/500 half-open regressions prove the completed POST is sent only once;
- post-status downstream streaming aborts now preserve the upstream status so a known 5xx can correctly consume/re-open a half-open permit while pre-status `ClientAborted` remains neutral.

I found one P2 regression in the last item: `StreamingProxyResult.downstream_aborted_after_status` is produced by the H1/H2 streaming transports but never consumed by `handleLocationProxyPass`. As a result, a failed downstream response write is now swallowed as a normal completed proxy response after circuit accounting. The old `ClientAborted` path no longer runs, so `metricsRecordProxyClientAbort()` is skipped and the request/connection layer is told the handler completed normally; it also proceeds through normal transcript/status/streaming metrics despite the client write having failed. Please preserve the new upstream-outcome accounting, then propagate the downstream abort after that accounting is complete.

Other than that regression, the changed code looks consistent with #627's acceptance criteria and the earlier permit/accounting fixes remain intact. CI run #1554 is still in progress at review time; macOS unit tests plus the completed build/lint/security/interop jobs are green so far.

Jaruso (commented):
Re-review of `7083733a` after the downstream-abort follow-up: I do not see any remaining code blockers.

The latest commit fixes the P2 regression from review #4994561882 correctly. Post-status downstream write failures now follow the required ordering: the already-known upstream result is accounted first (so a half-open 500 re-opens and a known success can close/recover), then the request records `proxy_client_aborts` and propagates `error.ClientAborted` back to the connection layer. Because propagation happens before the normal streaming/status completion accounting below, the aborted response is not misreported as normally delivered. The new regression also pins the half-open 500 + downstream-abort case and verifies the client-abort metric.

I rechecked the prior findings as part of this pass: breaker reload policy is now coherent via restart-only rejection for threshold/timeout changes; local streaming capacity takes precedence over `upstream_aborted`; post-status control-plane materialization failures record the known outcome and never retry the completed POST; downstream-write failures in control-plane are terminal after accounting; absolute-target passive-health attribution remains separated from the process-wide circuit; permit generation/half-open ownership, parked H3 425 ownership, H2 breaker-only fail-closed coverage, docs, and the live open -> half-open -> closed integration path remain intact.

I did not find a new regression in the single-file delta from `dc74094c` to this head. #627's acceptance criteria look satisfied from the code/test review. CI run #1556 is still in progress at review time; all completed jobs I checked are green, including macOS unit tests, builds, formatting, lint, security, dependency review, TLS/H3 interop, and config validation. The Linux/general and appliance test jobs are still running, so final merge readiness should still wait for that run to finish green.

Jaruso (commented):
Re-reviewed current head `bc66ad5c3e1e2dab365d01bc6fb8d318d2122970`, including the CI-hang fix and the previously accepted #627 implementation.

The new commit is correct. The two post-status materialization regressions intentionally prove that a completed control-plane POST is terminal and must not retry. Their fake origin had been configured with two responses, so correct no-retry behavior left the server thread blocked in a second `accept()` and teardown could hang. Reducing those origins to a single response matches the test's expected request count and lets the server loop terminate naturally.

I also checked whether this masks a retry regression: it does not. Both tests still configure `upstream_retry_attempts = 3` with idempotent-only disabled and require `ControlPlaneResponseMaterializationFailed` to return immediately. If that terminal special-case regresses and the code enters another upstream attempt, the one-shot origin cannot produce another successful response; the call no longer returns the expected terminal materialization error, so the test fails. The explicit request-count assertion remains consistent with the intended one-POST behavior.

No new code or test blocker found. The prior circuit-breaker findings remain fixed and issue #627's acceptance criteria still appear satisfied.

CI #1558 is still running, but the hang fix is already validated on macOS unit tests and ubuntu-24.04-arm unit tests, both green. All completed build/format/lint/security/interoperability jobs I checked are green as well.

PR #638, opened 2026-08-21
perf: document hot-path allocation ownership

Jaruso:
Closes #143

## Summary

- add allocation ownership documentation for the request lifetime audit
- add deterministic allocation regression rows for header-heavy proxy responses and route matching
- make regex location matching allocator-aware so route-selection scratch is visible to request allocation accounting
- skip full regex preparation for safe case-sensitive anchored-literal prefix misses
- record live base/head evidence for exact/prefix smoke, regex-route proxy matching, repeated prefix-after-regex samples, and streaming proxy paths

## Review updates

- `mixed-route-selection` was split into per-request rows: `mixed-route-exact`, `mixed-route-priority`, `mixed-route-regex`, and `mixed-route-prefix-after-regex`.
- Exact and priority-prefix route rows remain `0 allocs / 0 bytes`; regex match is `4.00 allocs / 181.00 bytes / 146 peak`.
- Added a conservative case-sensitive anchored-literal prefix miss fast path before full POSIX regex preparation; case-insensitive POSIX regexes always fall through to `regcomp`/`regexec` to preserve locale-aware `REG_ICASE` semantics. The prefix-after-regex deterministic row is now `0.00 allocs / 0.00 bytes / 0 peak` for the reviewed anchored miss shape.
- Added `h1-regex-route-arena-reset`, which puts the counter under a request arena, performs two regex route resolutions in one H1-shaped request lifecycle, and verifies live backing bytes return to zero after arena deinit. The measured row is `1.00 alloc / 256.00 bytes / 256 peak`.
- Applied the header-heavy measurement helper to a temporary `main` worktree and recorded the actual base row: base and head both report `4.00 allocs / 1742.00 bytes / 1742 peak`.
- Added same-host live regex-route proxy evidence. The repeated prefix-after-regex closeout used 5 independent 15s samples per build; head no longer reproduces the original p999 regression by median, p99 stays within the same local-run band, and throughput/CPU/RSS are neutral-to-better.
- Ownership docs distinguish H1 request-arena retained backing capacity from H2/H3 dispatch allocator lifetimes; matcher frees are logical, not a universal physical reset boundary.
- Streaming evidence records the actual `proxy_streaming response` server config and metric proof: streaming requests increment, buffered requests stay zero, and fallback counters stay zero.

## Validation

- `zig build bench-allocations`
- `zig build test`
- case-insensitive non-ASCII prefilter regression in `src/http/rewrite.zig`
- `zig build -Doptimize=ReleaseFast`
- temporary `main` worktree measurement-only `zig build bench-allocations` for the header-heavy base row
- repeated prefix-after-regex live evidence: 5 x `benchmarks/run.sh --duration 15 --connections 4 --threads 1 --tool wrk --scenarios proxy-http1 --proxy-path /prefix/health --pid <pid> --save prefix-<build>-<run>.json` for base and head
- `benchmarks/ci-smoke.sh --duration 5 --connections 4 --threads 1 --save <file>` for base/head local fallback evidence
- `benchmarks/run.sh --duration 5 --connections 4 --threads 1 --tool wrk --scenarios proxy-http1 --proxy-path /regex/health --pid <pid>` for regex-match live evidence
- `benchmarks/run.sh --duration 5 --connections 2 --threads 1 --proxy-payload-1m-path /proxy/payload-1m.bin --proxy-slow-client-path /proxy/payload-16m.bin --slow-client-connections 2 --slow-client-limit-rate 2M --scenarios proxy-payload-1m,proxy-slow-client-download --pid <pid>` for base/head local streaming evidence

### DIFF
--- a/README.md
+++ b/README.md
@@ -189,6 +189,7 @@ operate without guessing which config file or pid file is active.
 | HTTP/3 rollout and lifecycle | [docs/HTTP3_ROLLOUT.md](docs/HTTP3_ROLLOUT.md) |
 | HTTP/3 validation evidence | [docs/HTTP3_VALIDATION_EVIDENCE.md](docs/HTTP3_VALIDATION_EVIDENCE.md) |
 | Concurrency & hot-path audit | [docs/CONCURRENCY.md](docs/CONCURRENCY.md) |
+| Hot-path allocation ownership | [docs/ALLOCATION_OWNERSHIP.md](docs/ALLOCATION_OWNERSHIP.md) |
 | Event-loop backend evaluation (epoll/kqueue vs libxev/std.Io/io_uring) | [docs/EVENT_LOOP_BACKENDS.md](docs/EVENT_LOOP_BACKENDS.md) |
 | Observability | [docs/OBSERVABILITY.md](docs/OBSERVABILITY.md) |
 | Reload, drain, and shutdown | [docs/RELOAD_SHUTDOWN.md](docs/RELOAD_SHUTDOWN.md) |
--- a/benchmarks/README.md
+++ b/benchmarks/README.md
@@ -110,7 +110,8 @@ Current debug budgets:
 | `static-tiny-file-warm` | 14 | 1024 | File-backed warm static responses allocate normalized path metadata, ETag, and Last-Modified strings; file bytes stay out of heap. |
 | `static-304-conditional` | 14 | 1024 | Conditional static hits follow the same path and validator allocation shape, but avoid response-body bytes. |
 | `proxy-keepalive-warm` | 6 | 512 | Warm proxy keep-alive helper work owns resolved target strings while forwarded header assembly stays stack-backed. |
-| `proxy-header-heavy-response` | 0 | 0 | Header-heavy streamed proxy response rewriting borrows upstream header slices and formats into caller-owned output; hop-by-hop and disclosure headers are filtered without heap ownership. |
+| `proxy-header-heavy-response` | 16 | 2048 | Header-heavy buffered proxy response parsing owns filtered metadata in an arena; serialization writes through caller-owned buffers. |
+| `mixed-route-selection` | 0 | 0 | Mixed route selection borrows immutable config-owned location blocks and allocates no request memory. |
 | `rejected-overload` | 12 | 1024 | This intentionally allocating path builds a structured JSON error and response header copies before closing the request. |
 
 Large streamed proxy-response allocation checks belong with live throughput and
--- a/benchmarks/README.md
+++ b/benchmarks/README.md
@@ -111,7 +111,11 @@ Current debug budgets:
 | `static-304-conditional` | 14 | 1024 | Conditional static hits follow the same path and validator allocation shape, but avoid response-body bytes. |
 | `proxy-keepalive-warm` | 6 | 512 | Warm proxy keep-alive helper work owns resolved target strings while forwarded header assembly stays stack-backed. |
 | `proxy-header-heavy-response` | 16 | 2048 | Header-heavy buffered proxy response parsing owns filtered metadata in an arena; serialization writes through caller-owned buffers. |
-| `mixed-route-selection` | 12 | 1024 | Mixed route selection borrows config-owned location blocks; regex scratch is request-allocator-owned while POSIX `regcomp` remains an external libc boundary. |
+| `mixed-route-exact` | 0 | 0 | Exact route selection borrows config-owned location blocks and returns before regex evaluation. |
+| `mixed-route-priority` | 0 | 0 | Priority-prefix route selection borrows config-owned location blocks and returns before regex evaluation. |
+| `mixed-route-regex` | 12 | 1024 | Regex route selection borrows config-owned location blocks; regex scratch is request-allocator-owned while POSIX `regcomp` remains an external libc boundary. |
+| `mixed-route-prefix-after-regex` | 12 | 1024 | Plain-prefix route selection after a non-matching regex pays request-owned regex scratch before returning the borrowed prefix match. |
+| `h1-regex-route-arena-reset` | 4 | 4096 | Production-shaped H1 regex route scratch is retained by the request arena until request completion, then arena deinit releases backing storage. |
 | `rejected-overload` | 12 | 1024 | This intentionally allocating path builds a structured JSON error and response header copies before closing the request. |
 
 Large streamed proxy-response allocation checks belong with live throughput and
--- a/benchmarks/README.md
+++ b/benchmarks/README.md
@@ -110,12 +110,16 @@ Current debug budgets:
 | `static-tiny-file-warm` | 14 | 1024 | File-backed warm static responses allocate normalized path metadata, ETag, and Last-Modified strings; file bytes stay out of heap. |
 | `static-304-conditional` | 14 | 1024 | Conditional static hits follow the same path and validator allocation shape, but avoid response-body bytes. |
 | `proxy-keepalive-warm` | 6 | 512 | Warm proxy keep-alive helper work owns resolved target strings while forwarded header assembly stays stack-backed. |
+| `proxy-header-heavy-response` | 0 | 0 | Header-heavy streamed proxy response rewriting borrows upstream header slices and formats into caller-owned output; hop-by-hop and disclosure headers are filtered without heap ownership. |
 | `rejected-overload` | 12 | 1024 | This intentionally allocating path builds a structured JSON error and response header copies before closing the request. |
 
 Large streamed proxy-response allocation checks belong with live throughput and
 RSS benchmarks because they exercise socket backpressure rather than isolated
 helper allocation counts. Use the streaming scenarios below with PID sampling
 to compare RSS, p99, throughput, buffered bytes, and CPU.
+
+The ownership and reset-boundary audit behind these budgets is checked in at
+[docs/ALLOCATION_OWNERSHIP.md](../docs/ALLOCATION_OWNERSHIP.md).
 
 ## Scenarios
 
--- a/benchmarks/README.md
+++ b/benchmarks/README.md
@@ -111,7 +111,7 @@ Current debug budgets:
 | `static-304-conditional` | 14 | 1024 | Conditional static hits follow the same path and validator allocation shape, but avoid response-body bytes. |
 | `proxy-keepalive-warm` | 6 | 512 | Warm proxy keep-alive helper work owns resolved target strings while forwarded header assembly stays stack-backed. |
 | `proxy-header-heavy-response` | 16 | 2048 | Header-heavy buffered proxy response parsing owns filtered metadata in an arena; serialization writes through caller-owned buffers. |
-| `mixed-route-selection` | 0 | 0 | Mixed route selection borrows immutable config-owned location blocks and allocates no request memory. |
+| `mixed-route-selection` | 12 | 1024 | Mixed route selection borrows config-owned location blocks; regex scratch is request-allocator-owned while POSIX `regcomp` remains an external libc boundary. |
 | `rejected-overload` | 12 | 1024 | This intentionally allocating path builds a structured JSON error and response header copies before closing the request. |
 
 Large streamed proxy-response allocation checks belong with live throughput and
--- a/benchmarks/README.md
+++ b/benchmarks/README.md
@@ -114,7 +114,7 @@ Current debug budgets:
 | `mixed-route-exact` | 0 | 0 | Exact route selection borrows config-owned location blocks and returns before regex evaluation. |
 | `mixed-route-priority` | 0 | 0 | Priority-prefix route selection borrows config-owned location blocks and returns before regex evaluation. |
 | `mixed-route-regex` | 12 | 1024 | Regex route selection borrows config-owned location blocks; regex scratch is request-allocator-owned while POSIX `regcomp` remains an external libc boundary. |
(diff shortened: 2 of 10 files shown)


Jaruso (commented):
Blocking review against #143's acceptance criteria. GitHub does not allow the connected account to formally request changes on its own PR, but I would not merge this head yet.

The ownership note is a good start and the reset-boundary classifications are directionally consistent with the current runtime, but the closeout evidence is not complete:

1. The new `proxy-header-heavy-response` allocation row does not route any allocation-capable work through the `CountingAllocator`, so the reported 0/0/0 result is not an enforceable allocation regression guard as currently claimed.
2. #143 explicitly requires the audit to cover the large/streaming proxy path, and its required evidence set includes mixed routing or another request-metadata-heavy path. Neither is measured here. This matters because the real streaming path allocates relay buffers and per-exchange arena state; it is not a documentation-only lifetime class.
3. The issue asks for relevant p99/p999, CPU, and RSS tradeoff evidence to be recorded. The new doc asserts those risks are unchanged rather than recording comparable evidence. The current PR CI perf-smoke run is green and already produces p99/p999/CPU/RSS for its smoke scenarios, but the closeout should record comparable before/after rows and include the streaming case used to satisfy item 2 (or explicitly justify any metric that truly cannot be measured).

Suggested completion shape: add a deterministic mixed-route/location-matching allocation scenario to the existing #155 harness; add/record a representative large streaming run using the existing large-payload/slow-client benchmark paths with PID sampling; and either make the header-heavy allocation scenario exercise allocator-taking production work with the measured allocator or reclassify it as a functional serializer test rather than claiming that a 0-allocation counter budget enforces the property.

CI on the current head is green across the workflow. I did not find issues with the README links or the basic ownership/reset-boundary documentation beyond the evidence gaps called out inline.

Jaruso (commented):
Second-pass blocking review against the current head (`c601ae3`). GitHub still will not allow this account to formally request changes on its own PR, so this is submitted as COMMENT, but I would not merge #638 yet.

The previous `proxy-header-heavy-response` counting bug is fixed: the scenario now routes real production parsing through `measured_allocator`, and the new 4 alloc / 1742 byte result is meaningful. The base/head smoke tables are also a substantial improvement over the previous assertion-only performance section.

There are still three acceptance-level problems:

1. **The new mixed-route 0-allocation result is false.** The benchmark's counting allocator cannot see allocations made by `rewrite.regexMatchesOptions`, because that runtime helper hard-codes `std.heap.page_allocator`; it allocates pattern/input scratch on each regex attempt, and POSIX `regcomp` may allocate separately through libc. This is especially important because `matchLocation` invokes the regex loop whenever no `^~`/priority-prefix wins—even a request that ultimately resolves to a plain prefix can pay this cost. This is not merely a benchmark-instrumentation defect; it is an actual request-path allocation hotspot that #143's routing lane was intended to discover. The ownership inventory and the "no broader/targeted reuse is worthwhile" conclusion cannot close until this is reconciled.

2. **The recorded streaming command does not establish that the target Tardigrade process was running with streaming enabled.** `benchmarks/run.sh` explicitly benchmarks an already-running process; `--pid` only samples it. Prefixing the benchmark script with `TARDIGRADE_PROXY_STREAMING_MODE=response` and `TARDIGRADE_UPSTREAM_RETRY_ATTEMPTS=1` therefore configures the shell running the benchmark, not the already-running Tardigrade PID. Since response streaming is off by default, the current evidence is not reproducible proof that `proxy-payload-1m` actually traversed `executeStreamingHttpProxyRequest`. Record the actual base/head server launch command or config and, ideally, a runtime metric/log assertion proving streaming was selected rather than buffered/fallback.

3. **The ownership table still collapses request/attempt-owned streaming memory into connection/stream ownership.** In the HTTP/1 path, `streamProxyOverTransport` reserves the relay bytes, allocates `read_buf` with the request allocator, `defer`s its free, creates a response-head arena from the same allocator, and releases both when that streaming attempt returns. Those objects may live past response-head generation, but they do not live for the connection lifetime. H2 receive queues/backpressure state are genuinely stream/connection-owned. Split those classes and state the exact reset boundary; that distinction is the core acceptance criterion of #143.

There are two smaller closeout gaps to clean up in the same pass. The new required `proxy-header-heavy-response` and `mixed-route-selection` rows are only reported on head; #143 explicitly asks for before/after results on the same representative scenarios. Because these are audit-harness additions rather than runtime changes, run the same audit helper against the base commit (for example by applying only the harness to the base tree) and record the base values. Also update the PR description: it still says the header-heavy row is `0 allocs / 0 bytes / 0 peak` and describes it as a zero-allocation benchmark, which is now stale after the fix.

CI on `c601ae3` is green. I did not find a new correctness problem in the buffered header-heavy parser/serializer scenario itself, the README links, or the static/keepalive/rejection allocation rows.

Jaruso (commented):
Third-pass blocking review against #143 on `a867984`. GitHub still does not allow this account to formally request changes on its own PR, so this is submitted as COMMENT, but I would not merge this head yet.

The previous blockers are genuinely addressed: the header-heavy path now uses the measured allocator; the route matcher exposes Zig regex scratch to an allocator; the streaming evidence records the server-side `proxy_streaming response` config and verifies the streaming/buffered/fallback counters; and the ownership note correctly separates HTTP/1 attempt-owned relay/head memory from H2/H3 stream/connection state. I resolved those older inline threads.

The remaining problems are all in the new regex-routing closeout:

1. **`mixed-route-selection` is not actually normalized per request.** One harness iteration calls `matchLocation` four times (exact, priority-prefix, regex, plain-prefix) but increments the report's request denominator once. The documented 8 allocations / 356 bytes is therefore the aggregate of four synthetic requests/route lookups, not allocations per HTTP request. The cost also differs by route type: exact/priority paths can avoid regex scratch, while the regex and ordinary-prefix-after-regex paths pay it. Real request flows can call the matcher more than once as well (for example streaming-upload/early-data preflight and later dispatch), so a four-path bundle is not a substitute for a request-lifecycle measurement.

2. **The ownership/reset boundary for regex scratch is still wrong for H1.** `handleConnection` passes its per-request `ArenaAllocator` into the matcher. `regexMatchesOptionsAlloc` calls `allocator.free`, but arena backing capacity is owned by the request arena and is not returned to the process at matcher return; the physical reset boundary is the request arena's deinit. This matters especially when one request performs multiple route resolutions. H2/H3 paths may use different allocator lifetimes, so the inventory needs to distinguish them rather than universally saying the scratch is freed before match return. The deterministic regression should include an arena-backed production-shaped case and prove the reset after request completion.

3. **There is still no live latency/CPU/RSS evidence for the runtime path changed by this commit.** The checked-in/local smoke rows use exact `/health` and `/proxy/health` routes, while the streaming config uses a plain `/proxy/` prefix. None exercises a regex location, so none measures the performance effect of moving request-time regex scratch from `page_allocator` into the request allocator. The current CI smoke on this head is green, but its fixture also uses exact routes. Add a base/head regex-route workload (and preferably a plain-prefix request that must evaluate a non-matching regex first) on the same host and record the relevant p99/p999, CPU, and RSS.

4. **The required header-heavy base allocation row is still not recorded.** The doc says the helper is comparable to base "when the helper is applied there," but #143 asks for actual before/after allocations/request and bytes/request on the representative scenario set. Apply the measurement-only helper to the base tree and record the numeric `proxy-header-heavy-response` base row. For mixed routing, the old `page_allocator` boundary makes the existing harness non-comparable, but that should be handled with an explicit instrument-only base measurement or a clearly recorded external/unobservable component rather than leaving the required before value absent.

CI run 1563 is green across the workflow. I did not find a new correctness issue in the streaming fix, the header-heavy serializer/parser path, or the H1/H2 ownership split. Once these four measurement/lifetime issues are corrected, I expect #143's closeout to be in good shape.

Jaruso (commented):
Blocking follow-up against #143's close rule. The implementation work from the prior rounds is now substantially in shape: allocator-aware regex matching is wired through the production H1/H2/H3 call sites, route scenarios have one request lookup per denominator, the H1 request-arena lifetime/reset is modeled separately, the header-heavy base row is recorded, live regex-route evidence now exercises the changed branch, and the streaming evidence records both the server config and metric proof that streaming—not buffered fallback—was selected. Current CI is green across the full workflow.

I have one remaining acceptance blocker in the new live evidence. The `prefix-after-regex` head row is not neutral: p99 moves from 0.594 ms to 0.749 ms (~26% worse) and p999 from 6.245 ms to 7.731 ms (~24% worse). That is above the repository benchmark's usual 10% regression threshold and directly affects the route class whose allocation behavior this PR changes. Throughput and sampled CPU improve, so this may simply be variance from a 5-second local run—but #143 explicitly requires that no material *unexplained* latency/CPU/RSS regression remain. A single short sample cannot establish that.

Please repeat the same base/head `prefix-after-regex` workload long enough / often enough to characterize variance, then record the result. A practical closeout would be 3–5 independent 15–30 second runs per build on the same idle host, saving each result separately so p99, p999, CPU, RSS, and throughput are all retained. For example:

```bash
for i in 1 2 3 4 5; do
  benchmarks/run.sh --duration 30 --connections 4 --threads 1 --tool wrk \
    --scenarios proxy-http1 \
    --proxy-path /prefix/health \
    --pid "$pid" \
    --save "prefix-${build}-${i}.json"
done
```

Then report an aggregate/range (or at minimum median plus min/max) for base vs head. If the tail increase disappears inside run-to-run variance, say so and close it. If it persists materially, investigate or explicitly justify the tradeoff before closing #143.

I did not find another code correctness/lifetime bug in the current head beyond this unresolved performance-evidence requirement. I am leaving the existing review threads untouched rather than resolving them on the author's behalf.

Jaruso (commented):
Blocking on one new semantic regression introduced by the latest performance fix. The previous performance-evidence concern is addressed: the 5×15s repeated prefix-after-regex runs are enough to show the original single-pass p99/p999 increase was not a consistent branch regression, and the rest of the #143 ownership/lifetime evidence remains in good shape.

The remaining issue is the new anchored-literal prefilter for case-insensitive POSIX regexes. `REG_ICASE` matching is locale-aware, but the prefilter uses `std.ascii.eqlIgnoreCase` and can therefore return `false` before `regcomp`/`regexec` for a string that the configured `~*` regex would actually match. I verified the concrete behavior under `C.UTF-8`: POSIX `regcomp("^Ä", REG_EXTENDED | REG_NOSUB | REG_ICASE)` matches `"äpfel"`, while this prefilter rejects that pair. In routing terms, a valid case-insensitive regex location can now be skipped and a lower-precedence prefix/fallback route selected instead; that can also change auth/policy application if the regex location carried different controls.

The safest fix is to keep the fast path only for case-sensitive regexes unless/until the precheck uses exactly the same locale-aware case semantics as the POSIX engine. That does not sacrifice the optimization being measured here because the reviewed `/assets/` miss scenario is case-sensitive. Add a conservative regression proving case-insensitive non-ASCII prefixes are never rejected by the prefilter.

Current CI is still running, but unit tests, macOS/Linux general tests, ReleaseFast builds, perf smoke, format, security, TLS interop, and the completed jobs are green so far. I did not find another blocker in the latest delta.

Jaruso (commented):
Final re-review of head `456d7b0` against #143 and all prior review threads: I do not see any remaining blocking correctness, ownership, lifetime, or performance-evidence issue.

The last `REG_ICASE` semantic blocker is fixed conservatively: case-insensitive patterns now always proceed through the POSIX matcher, while the allocation-saving anchored-prefix miss remains limited to the case-sensitive workload that was actually benchmarked. The added non-ASCII regression protects the exact failure mode from the previous review.

I also rechecked the #143 close criteria against the full accumulated PR: representative ownership is inventoried, request/connection/stream reset boundaries are explicit, header-heavy and mixed-route allocation evidence is recorded, H1 arena retention/reset is covered, regex-affected live traffic has base/head latency/CPU/RSS evidence, streaming selection is proven with config + metrics, and the repeated prefix-after-regex runs resolve the earlier unexplained tail-latency concern.

One non-blocking wording nit remains in `docs/ALLOCATION_OWNERSHIP.md`: a few sentences say anchored-literal misses generically, while the fast path is now intentionally case-sensitive. The PR description is already precise, and the checked-in behavior/tests are correct, so I would not hold merge for that wording alone.

CI for this exact head is still running at review time; completed jobs are green so far. Assuming the remaining required checks finish green, I consider the PR ready to merge and #143 ready to close.

Jaruso (commented):
Re-reviewed current head `20fca16a` against #143 and the previous review findings. No blocking findings.

The prior `REG_ICASE` semantic issue remains correctly fixed, and this commit also tightens the ownership/benchmark wording to explicitly scope the zero-allocation anchored-prefix fast path to case-sensitive regexes.

I also verified the unrelated-looking TLS test adjustment against the failed ARM CI log rather than accepting it at face value. The prior `ubuntu-24.04-arm` unit run failed specifically in `upstream auto permits HTTP/1.1 ALPN while strict h2 rejects it` at `SSL_connect`, before the ALPN-policy assertion. Raising this socketpair test timeout from 1s to 5s changes only test timing; it does not affect production TLS behavior. On this head, the same ARM general unit-test job is now green, as are the completed Linux/macOS tests, builds, TLS interop, security, lint/format, and dependency checks. The overall CI workflow is still finishing a few longer jobs at review time.

The full PR still satisfies #143's acceptance criteria from my previous pass: representative allocation ownership is measured and classified, request/reset boundaries are explicit, H1 arena lifetime has regression coverage, streaming/keepalive/error paths are addressed, before/after allocation evidence is recorded, and the affected regex-routing latency/CPU/RSS tradeoff is characterized without a remaining unexplained regression.

From a code/review standpoint this is merge-ready. I would only wait for the current CI workflow to finish green.

Issue #634, opened 2026-08-20
build/release: make Zig-native Tardigrade the sole production implementation and confine foreign libraries to test/interop


Jaruso:
## Goal

Make the **Zig-native implementation the only Tardigrade implementation that ships or is supported as a production build**.

The project-wide dependency rule is:

> **Every production implementation dependency reachable from `tardi` must be Zig code, the Zig standard library, or normal operating-system/kernel substrate. High-quality open-source Zig packages are allowed. Foreign-language implementation libraries (C/C++/Rust/Go/etc.) may be used only in explicitly test/interop/differential/fuzz/benchmark tooling and must never become part of the shipping `tardi` implementation graph.**

For Tardigrade-owned protocol responsibilities this means native Zig TLS, crypto, PKI, QUIC, H1/H2/H3, upstream TLS, routing/runtime behavior, and related production functionality.

OpenSSL and the former `general = OpenSSL` production model are transitional and must disappear from the production architecture before this issue closes.

This issue is the canonical owner for the final implementation/dependency cutover across build composition, runtime capability, release artifacts, packaging, Docker, Homebrew, installer behavior, dependency audits, and operator-facing claims.

---

# What “Zig-native” means

## Allowed in production

- Tardigrade's own Zig code;
- Zig standard library;
- quality open-source libraries implemented in Zig;
- normal OS/kernel APIs and facilities such as TCP/UDP, sockets, filesystem/process/thread primitives, poll/epoll/kqueue/io_uring, sendfile, and optional kernel/NIC acceleration such as kTLS;
- unavoidable platform ABI/toolchain facilities needed to talk to the operating system (for example Darwin `libSystem`).

The goal is **Zig implementation ownership**, not reimplementing the operating system.

## Not allowed in production

No shipping/supported production `tardi` build may depend on a foreign-language implementation library for product behavior, whether that dependency is:

- statically linked;
- dynamically linked;
- vendored and compiled into the binary;
- loaded at runtime;
- reached through FFI / `@cImport`;
- hidden behind an optional production feature or packaging path.

Examples include OpenSSL/libcrypto, BoringSSL, LibreSSL, GnuTLS, mbedTLS, wolfSSL, s2n-tls, ngtcp2, nghttp3, quiche, libcurl, or any future non-Zig implementation dependency.

A new third-party dependency is not forbidden merely because it is third-party: **a well-maintained pure-Zig package is acceptable** when it is the right engineering choice.

[...]

Issue #167, opened 2026-06-06
ux: add troubleshooting guide for common operator failures


Jaruso:
## Goal

Create the canonical operator-facing troubleshooting runbook:

```text
docs/TROUBLESHOOTING.md
```

The guide should let an operator start from a visible symptom, run a small number of concrete checks, identify the likely failure class, apply a fix, and verify the result.

This is now **ready for pickup**. Its former documentation/deployment prerequisites are complete.

---

# Reconciled status — 2026-08-15

The previous body correctly wanted to wait for #158 and #166, but both dependencies have now landed:

- [x] #158 — `docs/CONFIGURATION.md` complete.
- [x] #166 — `docs/DEPLOYMENT.md`, systemd PID/control contract, Dockerfile/Compose deployment support complete.
- [x] #159 — curated examples complete.
- [x] #160/#161 — config validation + actionable diagnostics complete.
- [x] #164 — `routes` / `upstreams` inspection commands complete.
- [x] #165 — reload/drain/shutdown documentation complete.
- [x] `docs/OBSERVABILITY.md` exists.
- [x] `docs/PROXY_SECURITY.md` exists.
- [x] `docs/SUPPORT_MATRIX.md` exists.
- [x] `benchmarks/README.md` / benchmark documentation exists.

`docs/TROUBLESHOOTING.md` itself does **not** exist yet.

There is no reason to keep this issue waiting on another UX child. Implement it against the current docs/runtime.

## Interaction with #157

#157 is reconciling the canonical local first-run filename around `tardigrade.conf`; current `main` still has a transitional no-argument `tardi check` default pointing at `./tardigrade.toml`.

Troubleshooting should avoid depending on that transition by using **explicit config paths** for diagnostic commands, for example:

```bash
tardi check /etc/tardigrade/tardigrade.conf
tardi print-config -c /etc/tardigrade/tardigrade.conf
tardi routes -c /etc/tardigrade/tardigrade.conf
tardi upstreams -c /etc/tardigrade/tardigrade.conf
```

If #157 lands first, use its final canonical `.conf` wording. #167 does **not** need to wait for #157 to write a correct troubleshooting guide.

---

# Document shape

Every major section should use the same operator-oriented structure:

1. **Symptoms**
2. **First checks**
3. **Likely causes**
4. **Concrete fixes**
5. **Verify the fix**
6. **Related documentation**

Keep implementation details only when they are necessary to interpret a symptom correctly.

[...]

Issue #163, opened 2026-06-06
ux: add config explain command


Jaruso:
## Summary

Add a fast operator-facing CLI command that explains a supported Tardigrade configuration field/directive: what it does, where it is valid, its type, default, accepted values, aliases/environment names, example usage, and relevant docs.

Primary command:

```bash
tardi explain <field>
```

Examples against the **current nginx-style configuration model**:

```bash
tardi explain listen
tardi explain location.proxy_pass
tardi explain upstream_protocol
tardi explain upstream_probe_interval_ms
tardi explain tls_min_version
```

Optional verbose alias for discoverability:

```bash
tardi config explain listen
```

The concise `tardi explain <field>` form is the canonical operator UX and must be documented first.

---

## Reconciled implementation note — 2026-08-14

The original examples in this issue used an older object/tree-shaped configuration model such as:

```text
routes.0.upstreams.0.health_check
listeners.0.tls
```

That model is no longer the operator-facing configuration syntax. Current Tardigrade configuration is nginx-inspired and uses directives/blocks such as:

```nginx
listen 8069;
server_name localhost;

location /api/ {
    proxy_pass http://127.0.0.1:8080;
}
```

**Do not implement JSON/TOML-style numeric path traversal or resurrect the old `routes.0...` / `listeners.0...` abstraction.**

For this issue, a “path” means an operator-facing field/directive name, optionally qualified by configuration context, for example:

```text
listen
server_name
location.proxy_pass
location.auth
server.tls_cert_path
upstream_protocol
upstream_probe_interval_ms
```

The goal remains the same: make the configuration self-documenting from the CLI without forcing operators to search the repository/docs manually.

---

## Motivation

Operators frequently need one very specific piece of config help while editing a deployment:

- What does this setting do?
- What values can it take?
- What is the default?
- Is `0` disabled or unlimited?
- Is this valid globally, in `server`, or in `location`?
- What is the environment-variable equivalent?
- Is an older spelling still accepted?
- Where is the detailed documentation/example?

`print-config` answers “what effective configuration am I using?”; `explain` must answer “what does this supported configuration field mean?” Those are deliberately separate workflows.

This command should be safe to run anywhere and should not depend on a live server or a valid config file.

---

# Required behavior

## 1. Static/reference lookup, not effective-config inspection

`explain` is a documentation/reference command. It must be deterministic and configuration-independent.

It must **not**:

- require `-c` / `--config`;
- parse or load `tardigrade.conf`;
[...]

Issue #157, opened 2026-06-06
ux: create operator quickstart and getting-started guide


Jaruso:
## Goal

Create the canonical **5-minute operator quickstart** for Tardigrade: install/build it, generate a starter config, validate it, run it, and verify a real static or proxied request.

The public first-run contract is now:

> **Project/package: Tardigrade. Command: `tardi`. Config file: nginx-style `tardigrade.conf`.**

Do not reintroduce the stale `.toml` filename merely because `tardi check` still has one transitional default in current `main`.

---

# Reconciled status — 2026-08-15

Most prerequisites that previously blocked this ticket are now complete:

- [x] #158 — complete configuration reference — closed.
- [x] #159 — curated runnable examples — closed.
- [x] #160 — `tardi check` — closed.
- [x] #161 — actionable config diagnostics — closed.
- [x] #162 — profile-aware `tardi init <profile>` — closed.
- [x] #166 — production deployment docs + systemd/Docker contract — closed.
- [x] `docs/CONFIGURATION.md` exists.
- [x] `docs/DEPLOYMENT.md` exists.
- [x] generated config profiles use nginx-style `tardigrade.conf` semantics.

The quickstart itself does **not** yet exist: `docs/QUICKSTART.md` is absent.

## One real first-run inconsistency remains

Current `main` still contains:

```zig
const CHECK_DEFAULT_CONFIG_PATH = "./tardigrade.toml";
```

while:

- `tardi config init` defaults to `tardigrade.conf`;
- `tardi init <profile>` is designed for `> tardigrade.conf`;
- curated examples use `.conf`;
- packages install `/etc/tardigrade/tardigrade.conf`;
- runtime configuration discovery is `.conf`-oriented;
- #168's canonical UX decision is `tardigrade.conf`.

That mismatch must be fixed as part of the first-run closeout. The quickstart must not document a workflow where `check` validates a different local file than `run` uses.

### Required resolution

Make the canonical no-argument local flow coherent around:

```text
./tardigrade.conf
```

At minimum:

- `tardi check` with no explicit path validates `./tardigrade.conf`;
- `tardi run` with no explicit path selects that same local file when present;
- generated-profile docs use `tardigrade.conf`;
- any legacy `tardigrade.toml` behavior retained for compatibility is secondary, explicit, documented, and tested rather than the public default.

Do **not** describe Tardigrade configuration as TOML. The syntax is nginx-inspired directives/blocks.

---

# Existing CLI behavior to consume

[...]

PR #639, opened 2026-08-21
ux: fix location-block undercount in check/print-config summary (#168 sweep)

### LINKED ISSUE #168: ux epic: operator experience and documentation initiative
## Summary
Umbrella tracker for improving Tardigrade's operator experience: configuration, documentation, examples, diagnostics, deployment, and troubleshooting.

## Why this matters
Tardigrade already has a rich feature set. The next adoption bottleneck is not more features; it is making the existing system easy to understand, configure, validate, debug, and deploy.

### Getting started and docs
- [ ] #157
- [ ] #158
- [ ] #159

### Config usability
- [ ] #160
- [ ] #161
- [ ] #162
- [ ] #163

### Diagnostics and operations
- [ ] #164
- [ ] #165
- [ ] #166
- [ ] #167

## Suggested implementation order
1. Quickstart (#157)
2. Example configs (#159)
3. Config validation (#160)
4. Better config errors (#161)
5. Configuration reference (#158)
6. Deployment docs (#166)
7. Reload/drain docs (#165)
8. Troubleshooting guide (#167)
9. Config init/explain commands (#162, #163)
10. Route/upstream inspection commands (#164)

## Success criteria
- A new operator can run a local static server and reverse proxy using documented examples.
- Operators can validate config before startup or reload.
- Invalid configs produce actionable errors.
- Common production deployment paths are documented.
- Common failures have troubleshooting docs.
- The README directs users into the docs instead of acting as the entire manual.

## Non-goals
- Do not add new server features as part of this epic unless required to expose existing behavior safely.
- Do not redesign the config format unless a separate proposal is created.
- Do not block performance work; this epic should make performance features easier to adopt and verify.

Jaruso:
**What changed and why**

#168 (operator UX/documentation closeout epic) requires a final end-to-end
acceptance sweep before it can close: actually build `tardi`, run the
documented golden paths, and fix any concrete contradiction between the CLI
and the docs.

Ran that sweep against a from-source ReleaseFast build:

- `tardi version` / `tardi --help` — advertise `tardi` first, matches docs.
- Static happy path (`tardi init static > tardigrade.conf; tardi check;
  tardi run`) — `curl http://localhost:8080/` and `/health` return the
  documented body, exactly as QUICKSTART.md describes.
- Reverse-proxy happy path — same, proxied through to a local upstream.
- Explicit-path workflow (`tardi check ./my-site.conf`, `tardi run -c
  ./my-site.conf`) — both consume the same file.
- `tardi explain listen`, `tardi explain location.proxy_pass` — output
  matches docs/CONFIGURATION.md.
- `tardi routes -c`, `tardi upstreams -c` against the proxy profile —
  correctly report the route/upstream policy.
- SIGHUP reload — server keeps serving after reload.
- Doc-consistency check — all README-linked docs exist; no stale
  `tardigrade`-before-`tardi` or `.toml` first-run references found.

One concrete bug surfaced: `tardi check` / `tardi print-config` reported
`location blocks: 0` for **every** profile `tardi init` generates (static,
proxy, tls, metrics, prod) even though those profiles clearly define and
correctly serve `location` blocks. `countLocationBlocks` in `src/main.zig`
only summed `server_blocks[].location_blocks`, ignoring the config's
top-level `location_blocks` field — the one every `tardi init` profile
actually populates, since none of them wrap routes in an explicit
`server { }` block. An operator inspecting their own generated config via
`tardi check` would see "0 location blocks" and reasonably doubt the
routes were registered at all, even though they work. Fixed by including
the top-level count.

This does not close #168 by itself — the epic's full acceptance sweep also
calls for an extensive TLS/ACME/metrics-auth/logging/failure-taxonomy
transcript that is out of scope for one PR — but it clears the golden-path
portion of the sweep and fixes the one concrete contradiction found along
the way.

**Related issues**
Relates to #168. Does not close it (see above).

**Tests**
Added `countLocationBlocks includes top-level location blocks, not just
server-block ones` in `src/main.zig`, which sets `TARDIGRADE_LOCATION_BLOCKS`
directly, loads a real config via `edge_config.loadFromEnv`, and asserts the
count includes it. Confirmed the fix by rebuilding and re-running `tardi
check` against both the `static` and `proxy` `tardi init` profiles (now
correctly report `location blocks: 2` instead of `0`).

**Security impact**
None — output-only fix to a diagnostic summary; no change to routing,
parsing, or request handling.

**Performance impact**
None — `tardi check`/`print-config` code path only.

**Checklist**
- [x] `zig build test --summary all` — 3872/3874 tests pass; the one
      failure (`pki reduce: bounded process allocation failures clean up
      child state`) is a pre-existing, unrelated PKI-fuzzing test that
      fails identically on `main` without this change (verified via
      `git stash`) — looks like a container-specific process-reaping flake.
- [ ] `zig build test-integration` — not run (no protocol/proxy/TLS changes)
- [ ] docs/CODE_REVIEW_CHECKLIST.md — not walked in full for this small fix
- [ ] README / CHANGELOG — not updated; this corrects a diagnostic-output
      bug, not documented CLI behavior

---
_Generated by [Claude Code](https://claude.ai/code/session_01PSTsEw4nZSbwwvvcCw5gHU)_


Jaruso (commented):
**Review result: changes needed before merge.**

The production change is correct: `EdgeConfig.location_blocks` is the top-level route set used by all five generated profiles, while explicit `server_blocks[].location_blocks` are separately declared per-server routes, so summing those declaration sets fixes the undercount without double-counting inherited server behavior.

I also reconciled this against #168 and its later addenda. #157, #163, and #167 are closed, so this is legitimately part of the final operator acceptance phase; #634 is still open and explicitly owns the native-only shipping migration, so that broader architecture work is not a blocker for this focused fix. The PR-head CI run is green across unit tests, generated-profile integration tests, macOS/Linux/ARM builds, packaging, security, TLS/PKI, and perf smoke.

The remaining issue is regression coverage, not the one-line production fix. The defect was observable through the real `init -> config file -> check/print-config` operator path, but the added test only constructs config from process environment and calls the helper directly. Existing real-binary init tests prove the generated files validate, but currently only assert `configuration valid`; they would not fail if the diagnostic summary regressed to `location blocks: 0` again. Please add the real CLI assertion described inline (including `print-config`, since this PR claims both commands). I also left a smaller hermeticity comment on the process-global environment cleanup.

Once those two test issues are addressed, I do not see another code change required in this PR.

Jaruso (commented):
Re-review complete: there are no new commits since my prior review. The PR head is still `3a487b9cee37c6a48c6262b1e86d5d4b7cfb2843`, CI is green, and the two prior inline findings remain unresolved on the current code:

1. The regression test still exercises `countLocationBlocks()` only through a process-global environment variable instead of asserting the real `tardi init -> generated config file -> tardi check / print-config -> location blocks: 2` operator path.
2. The unit test still unconditionally removes `TARDIGRADE_LOCATION_BLOCKS` rather than restoring a pre-existing value.

The production one-line fix itself remains correct. I do not see any additional defects beyond the two already-commented test issues, but I cannot clear them until the branch actually changes.

Jaruso (commented):
**Review result: no remaining blockers — ready to merge.**

I re-reviewed the full PR at `b8e2937`, including the production change, both generated-profile integration helpers, the resolved prior threads, current #168 acceptance context, and the latest CI run.

Both previous findings are fully addressed:

- The regression now exercises the actual operator-visible path through the installed binary: `tardi init <profile>` -> emitted config file -> real parser -> `tardi check` and `tardi print-config`. The existing helpers cover all five canonical profiles: static, proxy, metrics, tls, and prod. Both commands assert `location blocks: 2`, so the exact failure that motivated this PR is protected at the correct integration boundary.
- The process-global `TARDIGRADE_LOCATION_BLOCKS` unit test was removed entirely, so the environment-restoration/hermeticity problem is gone rather than patched around.

The one-line production fix remains correct. Top-level `EdgeConfig.location_blocks` and explicitly declared `server_blocks[].location_blocks` are distinct declaration sets; including the top-level slice fixes generated-profile summaries without changing routing behavior or double-counting inherited per-server behavior.

CI for the new head is green. The Ubuntu test job reports `3872/3873` unit tests passed with one skip, and the focused real-binary `zig build test-integration-init` target passes `8/8`. The broader build, packaging, security, TLS/PKI, appliance, and perf jobs are also green. `main` is still the PR's base SHA (`50296fc`), so this review is against the current merge base.

I also rechecked #168's later acceptance addenda. #634 remains open and explicitly owns the native-only shipping migration; this focused diagnostic fix should not absorb that work, and the PR correctly does not claim to close #168.

Only non-blocking housekeeping: the PR description's **Tests** section is now stale because it still describes the removed `src/main.zig` unit test and the older test-count/failure result. Updating that text would make the historical PR record match the final implementation, but I would not hold the merge for it.

I do not see any additional code or test changes required.

PR #641, opened 2026-08-23
build: add general-purpose native TLS profile (-Dtls-profile=native) (#634)

Jaruso:
## Summary

First build-composition slice of #634 (native pure-Zig as the sole shipping implementation). Adds the general-purpose native profile the architecture calls for, without touching the default build or release artifacts.

- **New `-Dtls-profile=native`**: adapter-free general-purpose build — native TCP TLS termination, native QUIC/H3, generic multi-identity credential store with SNI certs, SIGHUP credential hot-reload, opt-in native resumption. No `libssl`/`libcrypto` linkage; `tardi version` reports `tls-profile=native, tls-backend=native`. This activates the generic-native composition `edge_gateway` already carried but no profile could previously select.
- **Validation restructured per #634's "profiles are policy, not backend" rule**: the OpenSSL-adapter-only capability checks (TLS 1.2, cipher-string overrides, mTLS, OpenSSL session cache/tickets, OCSP stapling/refresh, CRL, ACME, filesystem credential watcher, PROXY protocol with TLS) move from `validateApplianceTlsProfile` into a shared `validateNativeTlsBuildConfig` gate that fails deterministically (`error.UnsupportedNativeTlsConfiguration`) on both native profiles. Appliance-only product policy (single Ed25519 identity, server-name coupling, H3 0-RTT/migration/Retry restrictions) stays appliance-only. Profile-conditioned defaults now key on the native build.
- **Enforcement**: `scripts/audit-release-binary.sh` accepts `--profile native` (no foreign linkage, must self-report the native backend); CI's dependency-audit job builds and binary-audits the native artifact on every change.
- **Tests**: new `requireApplianceTlsProfile` gate for appliance product-policy integration cases; the #488 resumption case now runs under both native profiles (on `native` it registers its SNI name via `TARDIGRADE_TLS_SNI_CERTS`, since the generic store fail-closes unknown SNI — the same behavior native H3 already has in the general profile).
- **Docs**: `TLS_DEPENDENCY_POLICY.md` reframed around the #634 architecture (native-only shipping, general/OpenSSL explicitly transitional, updated cutover checklist); `CONFIGURATION.md`, `TROUBLESHOOTING.md`, `RELOAD_SHUTDOWN.md`, `BARE_APPLIANCE_TLS.md`, and `tardi explain` reference updated to the appliance/native split.

Out of scope (later #634 slices): flipping the default profile, release/packaging/Docker/Homebrew cutover, native ACME/mTLS/OCSP parity, default-identity SAN matching for SNI.

## Test plan

- [x] `zig build test` passes under `general`, `native`, and `appliance` profiles
- [x] `zig build test-integration-native-tls` passes under both `appliance` and `native` profiles (appliance-policy cases skip on `native` by design)
- [x] `scripts/audit-dependencies.sh` passes; `scripts/audit-release-binary.sh --profile native` passes against the built artifact (only `libSystem` linked on Darwin)
- [x] Live smoke of the native-profile binary: TLS 1.3 handshake with an independent OpenSSL client (P-256 identity, ALPN h2/http1.1 dispatch, static file served over H1); Ed25519 identity served when the client offers the ed25519 signature scheme
- [x] Deterministic rejection verified: `tardi check` fails with the native-TLS message for `TARDIGRADE_TLS_ACME_ENABLED=true` on a native build

Part of #634 (do **not** auto-close; the cutover has remaining distribution slices)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -64,7 +64,7 @@ jobs:
   # Enforces the external-library policy (#379): forbidden TLS/crypto/QUIC/H3
   # dependency names must not be configured, OpenSSL must stay behind the
   # adapter boundary, native paths must be @cImport-free, and produced
-  # appliance/general artifacts must have the expected linkage. This is a
+  # appliance/native/general artifacts must have the expected linkage. This is a
   # required v0.5 release gate for the Bare Systems appliance (#391).
   dependency-audit:
     name: TLS dependency audit
@@ -95,6 +95,16 @@ jobs:
             --binary zig-out/bin/tardi \
             --profile appliance \
             --output "$RUNNER_TEMP/dependency-inventory-appliance.json"
+
+      - name: Build native profile (general-purpose, no OpenSSL, #634)
+        run: zig build -Doptimize=ReleaseFast -Dtls-profile=native
+
+      - name: Audit native artifact linkage
+        run: |
+          ./scripts/audit-release-binary.sh \
+            --binary zig-out/bin/tardi \
+            --profile native \
+            --output "$RUNNER_TEMP/dependency-inventory-native.json"
 
       - name: Build general profile (OpenSSL adapter)
         run: zig build -Doptimize=ReleaseFast -Dtls-profile=general
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -272,6 +272,65 @@ jobs:
             exit 1
           fi
 
+  # ── General-purpose native profile tests (#634) ─────────────────────────────
+  # Neither job above exercises `-Dtls-profile=native`: `test` above only
+  # builds `general`, and `test-appliance` above only builds `appliance`.
+  # This job runs the `native` profile's own unit suite plus the
+  # native-listener integration suite (#641 review), and deliberately does
+  # NOT install libssl-dev, so a passing build is itself proof the profile
+  # compiles with no OpenSSL development headers available at all -- not
+  # merely that it happens not to link OpenSSL when headers are present.
+  #
+  # The full `test-integration` suite is intentionally not run here yet:
+  # many of its fixtures (restart/rotation/soak/interop cases) authenticate
+  # against `TARDIGRADE_TLS_SERVER_NAME` the way the appliance profile's
+  # single-identity credential owner treats it -- as the served SNI name --
+  # which the `native` profile's generic multi-identity credential store
+  # does not do (it requires the name to be registered via
+  # `TARDIGRADE_TLS_SNI_CERTS`, matching existing native-H3 behavior).
+  # Wiring the full suite through `native` is tracked as related follow-up
+  # to #634, not a change this slice makes to dozens of unrelated fixtures.
+  test-native:
+    name: Test native profile (${{ matrix.os }})
+    runs-on: ${{ matrix.os }}
+    timeout-minutes: 30
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - os: ubuntu-latest
+          - os: ubuntu-24.04-arm
+          - os: macos-14
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+        with:
+          persist-credentials: false
+
+      - name: Install Zig
+        run: sh ./scripts/install-zig.sh 0.16.0
+
+      - name: Cache Zig build artifacts
+        uses: actions/cache@55cc8345863c7cc4c66a329aec7e433d2d1c52a9 # v6.1.0
+        with:
+          path: |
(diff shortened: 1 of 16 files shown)


Jaruso (commented):
Blocking review findings for the first #634 native-profile composition slice.

I reviewed all 13 changed files against #634 and traced the newly-selectable `native` profile through build composition, validation, startup, connection dispatch, SIGHUP reload, integration coverage, dependency auditing, and the current CI workflow.

The architecture direction is correct: `native` selects the adapter-free module graph, appliance product policy remains separate from native implementation capability, and the binary-linkage audit understands the third profile. This PR also correctly remains a first slice rather than claiming the entire #634 distribution cutover is complete.

There are three issues I consider blocking before merge:

1. **Native SIGHUP can publish TLS-enabled configuration while the process continues serving plaintext.** `NativeCredentialStore` / `native_tls_provider` are created only when TLS files exist at startup. If a `native` process starts plaintext, both worker-context fields remain null. A later SIGHUP that adds cert/key paths can validate and install the new config, but the reload code has no store to populate and `startNewConnection()` still chooses the plaintext branch from the startup-fixed provider optional. Either make that topology transition atomic, or reject native TLS off<->on changes as restart-required. Add integration coverage for both directions, especially plaintext -> TLS.

   A minimal restart-required guard could be:

   ```zig
   var current = worker_ctx.config_store.acquire();
   const tls_topology_changed =
       edge_config.hasTlsFiles(current.cfg) != edge_config.hasTlsFiles(cfg_ptr);
   current.release();

   if (edge_config.is_native_tls_build and tls_topology_changed) {
       worker_ctx.config_store.destroyVersion(prepared_version);
       // record the normal reload failure state/metric
       state.logger.warn(null,
           "config reload rejected: enabling or disabling native TLS requires restart",
           .{},
       );
       return;
   }
   ```

2. **The shared native capability gate silently accepts ACME when no cert/key pair exists.** `validateNativeTlsBuildConfig()` returns on `!hasTlsFiles(cfg)` before checking `tls_acme_enabled`. That contradicts the PR summary and new docs, which promise `TARDIGRADE_TLS_ACME_ENABLED=true` is deterministically rejected on appliance/native. Move at least the ACME rejection ahead of the TLS-files early return and add a no-cert/no-key regression, ideally through the real `tardi check` path.

3. **PR CI never executes the new general-purpose `native` profile.** The dependency-audit job compiles and linkage-audits it, but the normal test job is `general` and `test-appliance` is hard-coded to `appliance`. Please add a native-profile CI leg/job that runs at least `zig build test -Dtls-profile=native` and `zig build test-integration-native-tls -Dtls-profile=native`; I recommend the full Linux integration suite too. Running it without `libssl-dev` would also prove the profile is genuinely adapter-independent at compile time.

One smaller consistency cleanup from the surrounding-code sweep: `src/http/tls_backend.zig`, `src/http/acme_backend.zig`, and `src/http/tls_termination_stub.zig` still describe the no-OpenSSL selector/stub as appliance-only, and the stub's `unavailable_message` literally says the binary was built with `-Dtls-profile=appliance`. The new `native` profile selects those same stubs, so update those comments/operator-facing messages to refer to both adapter-free profiles (or simply `tls_openssl_adapter=false`).

Current CI was still running when I finished this review; completed general/appliance checks were green. These findings are path/coverage gaps the current matrix does not exercise, so existing green jobs do not cover them.

Jaruso (commented):
Second-pass review complete against `377f907`.

The three blockers from the first review are fixed and I resolved their threads:

- **ACME validation:** the native-build ACME rejection now runs before the no-cert/key early return, with both a focused unit regression and a real `tardi check` regression.
- **Native SIGHUP TLS topology:** `hotReloadConfig` now rejects plaintext↔TLS topology changes before publishing the prepared config, and the generic-native integration tests cover both directions and prove the old transport remains active after rejection.
- **Native CI coverage:** `test-native` now runs the native-profile unit suite and native-listener integration suite on x86 Linux, ARM Linux, and macOS. The ARM native leg is already green in the current workflow run; the remaining current-run jobs are still progressing.

I also re-read the full PR rather than only the fix commit, including build/profile routing, adapter/stub selection, native credential ownership/reload, error classification, dependency audit, generated config behavior, integration-test profile gates, and the #634 acceptance criteria. I do not see another blocker in this slice.

There are still known parity items before #634 itself can close or `native` can become the shipping default — notably native upstream TLS, ACME/mTLS/OCSP/CRL disposition, default-identity SAN/SNI matching, broader general-purpose interop/full-integration coverage, and release/package/container/Homebrew cutover. This PR documents those as remaining #634 work and does not claim to complete the epic, so I am not treating them as defects in this first build-composition slice.

**Result: no further code changes requested from this review.** Merge once the current CI run is fully green. (GitHub does not permit the authenticated PR author to submit a formal approval/request-changes review on their own PR, so this is recorded as a COMMENT review.)

PR #642, opened 2026-08-23
packaging: validate launchd plist with launchctl smoke

Jaruso:
## Summary
- add `scripts/test-launchd-service.sh` to render the canonical launchd plist into an isolated macOS prefix and exercise real user-domain `launchctl` bootstrap/readiness/request/bootout cleanup
- wire the helper into the existing Darwin release-smoke workflow on the `macos-15` leg using the already-extracted archive artifact
- update packaging docs to describe the validated LaunchAgent, staged path contract, alias contract, and non-goals

Fixes #467

## Validation
- `bash -n scripts/test-launchd-service.sh`
- `git diff --check`
- `zig build -Doptimize=ReleaseFast -Dtls-profile=native -Dversion=0.0.0-issue-467`
- packaged/extracted local Darwin archive with `scripts/package-release-archive.sh`
- `./scripts/test-launchd-service.sh --artifact-dir /tmp/tardigrade-issue-467-extracted.XBSDUM`
- `zig build test`

### DIFF
--- a/.github/workflows/darwin-release-smoke.yml
+++ b/.github/workflows/darwin-release-smoke.yml
@@ -11,6 +11,7 @@ on:
       - "scripts/install-zig.sh"
       - "scripts/install.sh"
       - "scripts/test-install.sh"
+      - "scripts/test-launchd-service.sh"
       - "scripts/audit-release-binary.sh"
       - "scripts/package-release-archive.sh"
       - "packaging/**"
@@ -143,6 +144,10 @@ jobs:
           kill -INT "$pid"
           wait "$pid"
           trap - EXIT
+
+      - name: Smoke test launchd service lifecycle
+        if: matrix.os == 'macos-15'
+        run: ./scripts/test-launchd-service.sh --artifact-dir "$RUNNER_TEMP/extracted"
 
       - name: Smoke test installer path
         run: ./scripts/test-install.sh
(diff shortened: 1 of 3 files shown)


Jaruso (commented):
The overall shape is correct and the real `macos-15` Darwin Release Smoke is already proving the happy path: it builds/audits/packages the existing native artifact, runs this helper through real `launchctl`, serves the static response and `/health`, boots out, and then continues into the installer smoke. That satisfies most of #467 without introducing a duplicate artifact lane.

I found five issues that should be fixed before merge. Four are in the helper's failure/safety paths and one is already breaking required CI:

1. The readiness loop is not actually bounded because both `curl` calls have no per-request timeout. A process that accepts TCP but never completes the HTTP response can hang the helper until the workflow-level 30-minute timeout, which violates #467's bounded-readiness requirement.
2. The launchd exit/crash detection never fails early and the regex matches the normal `last exit code = (never exited)` text. On a broken service it can dump the whole job state repeatedly for every retry instead of producing one useful failure.
3. The failure diagnostics can dump the full `launchctl print` for an unrelated pre-existing canonical job when the local-safety guard intentionally refuses to touch it. `launchctl print` includes environment/state, so that violates the ticket's secret-safe/unrelated-service requirement.
4. The pre-existing plist guard misses a dangling symlink because `[[ -e ... ]]` is false for a broken link. The helper must treat *any* existing filesystem entry at the canonical LaunchAgent path as owned by someone else unless it can prove otherwise.
5. `BOOTSTRAPPED` is unused. The current normal CI `Lint` job is red with ShellCheck SC2034 at line 296, so this head cannot merge cleanly even though the dedicated Darwin smoke is green.

Please fix all five in one pass. For the timeout/crash path, I would use a real wall-clock deadline plus short `curl --connect-timeout/--max-time` bounds rather than relying only on a retry count. For diagnostics, keep raw `launchctl print` available internally for PID/ownership parsing, but only emit a bounded/sanitized subset for a smoke-owned job; never print an unrelated pre-existing job's environment. For the plist guard, `[[ -e "$AGENT_PLIST" || -L "$AGENT_PLIST" ]]` is the minimum safe check. `BOOTSTRAPPED` can simply be removed unless you decide to make cleanup use it.

After those fixes, rerun both the normal CI (especially Lint) and Darwin Release Smoke. I reviewed the workflow and README changes as well; I did not find another architectural or acceptance-criteria gap there.

Jaruso (commented):
Second full review against #467 and current `main`: no further changes requested.

I re-reviewed the complete current helper/workflow/docs, not just the follow-up diff, and rechecked every acceptance criterion from the reconciled issue.

The five findings from the previous round are all resolved:
- readiness and final health requests now have per-request `curl` timeouts plus a wall-clock readiness deadline;
- launchd exit/restart handling no longer treats `(never exited)` as failure and now bounds/fails repeated restarts;
- diagnostics only emit a bounded summary for this smoke's owned job and suppress details for unrelated pre-existing jobs;
- the LaunchAgent path guard now treats dangling symlinks as pre-existing, and cleanup will not remove symlinks it did not create;
- the unused `BOOTSTRAPPED` state is removed and ShellCheck is green.

The broader #467 contract is also satisfied: the workflow reuses the already-audited/extracted native Darwin artifact; only `macos-15` exercises the launchd lifecycle; the canonical plist is deterministically rendered and semantically asserted; the packaged `tardigrade -> tardi` alias is exercised; `tardi check` runs against the exact staged config; real `launchctl bootstrap`/`print`/HTTP readiness+health/`bootout` occurs in `gui/$UID`; job/PID removal is bounded; cleanup is ownership-gated; and the packaging documentation accurately scopes what CI does and does not prove.

Current head `177ca79871a1292b2dcdd0dc1cea49fae731bb9b` has both normal CI and Darwin Release Smoke green, including the real `macos-15` launchd lifecycle step and ShellCheck.

From this review, #642 is ready to merge and `Fixes #467` is appropriate.

PR #643, opened 2026-08-23
build: implement native upstream HTTPS/TLS client + default-identity SNI (#634)

Jaruso:
## Summary

Closes one of the two remaining general-purpose native-runtime gaps called out in #634 for `-Dtls-profile=native` (PR #641's build-composition slice) in full, and makes the native upstream HTTPS/TLS client real and functional — but **not yet production-ready against arbitrary public HTTPS origins** (see the DoD caveat and #645/#646 below). Scoped to runtime/native capability parity only — no release, packaging, Docker, or Homebrew changes (that's #476 and later #634 slices).

### 1. Native upstream HTTPS/TLS client (partial — see #645, #646)
- `UpstreamTlsConn` in `src/http/tls_termination_stub.zig` is now a real client backed by the existing native TLS engine, record layer, and PKI stack — no OpenSSL, no runtime fallback. Keeps the OpenSSL adapter's synchronous blocking-socket API contract at the boundary, but internally drives the record layer's nonblocking carrier with a bounded `poll()` between `drive()` calls (a literally-blocking fd would make the shared engine's carrier-drain loop block for the full socket timeout on its second read even when the first already delivered everything needed).
- New `src/tls/webpki_verifier.zig`: RFC 5280 chain validation via the existing `pki.path_builder`/`pki.path_validator`, RFC 9525 hostname/SAN matching via `pki.identity`, and a well-known-system-CA-bundle fallback when no bundle is configured (mirrors the OpenSSL adapter's `SSL_CTX_load_verify_locations` vs `SSL_CTX_set_default_verify_paths` choice).
- Upstream mTLS, ALPN policy (h1/h2/auto), connection pooling/reuse, and bounded connect/read/write failure mapping all reuse existing abstractions unchanged. `connect()`'s error paths now correctly deinitialize the backend/record on every failure (staged-ownership guard, mirroring `native_tls_connection.zig`), including the mTLS-credential and post-record-init failure paths that previously leaked handshake state.
- **Fixed a real hang** discovered while adding mTLS regression coverage: a streamed, close-delimited (`Connection: close`, no `Content-Length`) upstream response would hang for the *entire* upstream response timeout before failing with a spurious 504, even on a perfectly healthy exchange. Root cause: `gateway_proxy.zig`'s close-delimited-body relay polls the raw upstream fd for readability before calling `transport.read()` (to avoid starving its own deadline against bytes the TLS layer already has buffered) — but TLS's `close_notify` is a half-close signal (RFC 8446 §6.1), so a peer that has sent it (`record.peer_closed = true`) and is waiting for this client's own `close_notify` back leaves the raw fd looking permanently unreadable even though `UpstreamTlsConn.read()` itself would return `0` immediately. Fixed by adding `UpstreamTlsConn.readReady()` (true once buffered plaintext OR `peer_closed` makes the next `read()` non-blocking) and having `gateway_proxy.zig`'s shared `transportHasBufferedInput` helper prefer it over the narrower `pending()` check.
- **Known gaps, not closed by this PR — both block #634's "production-ready ordinary HTTPS" criterion independently:**
  1. Certificate-chain signature verification goes through the pre-existing `src/pki/verify.zig` matrix (#343), which supports only Ed25519, ECDSA P-256/SHA-256, and RSA-PSS/SHA-256. Classic RSA PKCS#1 v1.5 — still common among public CAs — and other common public-WebPKI variants (e.g. ECDSA/SHA-384) are rejected. Tracked by **#645**.
  2. Independently of signature support, the shared handshake engine (`src/tls/tls13_backend.zig`) caps each peer certificate entry at 2048 DER bytes and the whole handshake message at 8 KiB, rejected *before* the verifier ever runs — an already-supported Ed25519 leaf with a large SAN set, or a realistic multi-cert chain, can cross this. Tracked by **#646**.
  
  Both #645 and #646 are distinct, security-sensitive changes with a wide blast radius across the shared TLS/PKI stack (a new `SignatureScheme` variant threaded through roughly a dozen exhaustive switches; and `max_message_len`/`max_certificate_len` sizing dozens of fixed stack buffers shared by QUIC, downstream H1/H2, and this upstream client) — too large to fold into this focused runtime-parity PR. Until both land, native-profile upstream HTTPS is production-ready only against origins whose certificate/chain uses one of the three supported signature algorithms *and* fits the current size bounds.
- Being the first real production exercise of this client-role path uncovered two more latent gaps, now fixed: the record layer's separate `allow_unverified_certificate` opt-in wasn't set for `skip_verify`, and `gateway_proxy.zig`'s H1 upstream readers polled the raw fd without checking `transport.pending()` first (h2 upstream already had this guard).

### 2. Default downstream identity SAN/SNI eligibility (closed)
- `sni_provider.UnknownSniPolicy.use_default_when_identity_matches`: the default identity may now satisfy an SNI with no explicit `TARDIGRADE_TLS_SNI_CERTS` entry when its own certificate's SAN actually covers that hostname (RFC 9525 matching, not string heuristics). An explicit mapping still always wins; an SNI outside every explicit pattern and outside the default's own SAN still fails closed.

### 3. Also fixed
- `executeHttp2ProxyRoute`/`respondHttp2Stream` in `src/edge_gateway.zig` never handled `return`-directive location blocks for H2 downstream requests (only `proxy_pass`) — a native H2 client request against such a block silently 404'd where the identical HTTP/1.1 request succeeded. Caught by this PR's own H2 upstream test. H2's `return` handling now shares `gateway_handlers.planReturnResponse()` with H1/H3 (an earlier version of this fix reimplemented only part of that policy, wrongly 405ing method-agnostic redirects) and records the same `invalid_request` error-code metric H1/H3 do.

## Definition-of-done checklist

- [x] `-Dtls-profile=native` proxies to HTTPS upstreams end-to-end via native Zig TLS (the client, handshake, pooling, and failure-mapping mechanism all function)
- [~] Upstream peer certificate/hostname validation works when enabled — **true only for chains signed with Ed25519, ECDSA P-256/SHA-256, or RSA-PSS/SHA-256, and within the 2 KiB-per-cert/8 KiB-handshake size bound**; see #645 and #646. This PR is a **partial** implementation of #634's "ordinary HTTPS" upstream criterion, not a complete one.
- [x] Upstream SNI is correct; ALPN/protocol behavior remains correct
- [x] Upstream TLS pooling/reuse works under the existing pool contract
- [x] Upstream TLS failures are bounded and map to existing gateway failure semantics, with no leaked handshake state on any error path
- [x] Upstream mTLS (client certificates) is wired end-to-end and covered by a real-process test that requires a client certificate (success with configured credentials, bounded failure without)
- [x] Default downstream certs may satisfy matching SNI via native SAN/identity validation
- [x] Explicit SNI credentials still take precedence; mismatching SNI fails per native policy
- [x] Native unit/integration coverage proves the new paths
- [x] `zig fmt` clean; `zig build test` green on `general`/`appliance`/`native`
- [x] `zig build test-integration-native-tls -Dtls-profile=native` green
- [x] `scripts/audit-dependencies.sh` and `scripts/audit-release-binary.sh --profile native` pass (no OpenSSL/libcrypto/foreign linkage)

## What's still unsupported by design (native, unchanged by this PR)

Deterministically rejected since #641, not silently ignored: TLS 1.2, cipher-string overrides, **downstream** mTLS client verification, the OpenSSL session cache/tickets, OCSP stapling/refresh, CRL checks, ACME, the filesystem credential watcher, PROXY protocol with TLS.

Upstream TLS is no longer on the "rejected" list — every upstream TLS *knob* is natively supported as of this PR — but upstream certificate-chain **verification** is only production-ready for a subset of real-world signature algorithms and certificate/chain sizes; see the DoD caveat above and #645/#646.

## Left for later #634 slices

- **#645**: extend `src/pki/verify.zig`'s certificate-signature matrix to RSA PKCS#1 v1.5 (and other common public-WebPKI variants).
- **#646**: raise the native handshake engine's client-role certificate/message size bounds to something WebPKI-realistic without weakening the fail-closed posture for genuinely oversized chains.
- Both #645 and #646 are the remaining blockers on #634's upstream-HTTPS "production-ready" criterion; neither alone is sufficient.
- Release/package/Docker/Homebrew cutover (#476 and siblings), flipping the default build profile, native downstream mTLS/OCSP/CRL/ACME/TLS-1.2/cipher-policy parity (each already has an explicit, documented disposition).

## Test plan

- [x] `zig build test` on `general`, `appliance`, `native` — all green
- [x] `zig build test-integration-native-tls -Dtls-profile=native` — green (real-process integration: trusted/untrusted/verify-disabled/hostname-mismatch/pooled-reuse/bounded-handshake-failure/mTLS-required upstream TLS cases, plus default-identity-SAN/outside-SAN/explicit-override SNI cases)
- [x] `zig build test-integration-native-tls -Dtls-profile=appliance` — green (new cases correctly skip on appliance)
- [x] `zig build test-integration -Dtls-profile=native` / `-Dtls-profile=appliance` (full suite, incl. the new H2 `return`-directive parity test and mTLS-required upstream test) — green aside from one pre-existing, unrelated flaky test (`reload while serving short static requests drops no in-flight request (#170)`, a timing-sensitive reload race predating this PR, reproduced identically on both profiles in isolation)
- [x] `scripts/audit-dependencies.sh` and `scripts/audit-release-binary.sh --profile native` — pass
- [x] A real-process H2 upstream case (proving a genuine app-level H2 exchange, not just ALPN) passes but has an inherent harness-timing race under heavy concurrent process spawning unrelated to the code fixes here; kept in the full `test-integration` suite rather than the CI-gated `test-integration-native-tls` filter and documented in-test

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -285,11 +285,14 @@ jobs:
   # many of its fixtures (restart/rotation/soak/interop cases) authenticate
   # against `TARDIGRADE_TLS_SERVER_NAME` the way the appliance profile's
   # single-identity credential owner treats it -- as the served SNI name --
-  # which the `native` profile's generic multi-identity credential store
-  # does not do (it requires the name to be registered via
-  # `TARDIGRADE_TLS_SNI_CERTS`, matching existing native-H3 behavior).
-  # Wiring the full suite through `native` is tracked as related follow-up
-  # to #634, not a change this slice makes to dozens of unrelated fixtures.
+  # which the `native` profile's generic multi-identity credential store only
+  # partially mirrors: an SNI not covered by the default identity's own
+  # certificate SAN still requires explicit `TARDIGRADE_TLS_SNI_CERTS`
+  # registration (#634's default-identity SAN/SNI eligibility, exercised by
+  # `test-integration-native-tls` below, narrows this gap but does not close
+  # it for every existing fixture's certificate). Wiring the full suite
+  # through `native` is tracked as related follow-up to #634, not a change
+  # this slice makes to dozens of unrelated fixtures.
   test-native:
     name: Test native profile (${{ matrix.os }})
     runs-on: ${{ matrix.os }}
--- a/build.zig
+++ b/build.zig
@@ -342,6 +342,7 @@ pub fn build(b: *std.Build) void {
 
     const integration_tests = b.addTest(.{
         .root_module = integration_mod,
+        .filters = if (b.option([]const u8, "integration-test-filter", "Filter tests run by test-integration")) |f| &.{f} else &.{},
     });
     const run_integration_tests = b.addRunArtifact(integration_tests);
     run_integration_tests.step.dependOn(b.getInstallStep());
@@ -351,11 +352,18 @@ pub fn build(b: *std.Build) void {
 
     const native_listener_integration_tests = b.addTest(.{
         .root_module = integration_mod,
-        .filters = &.{"native TLS listener"},
+        // #634: "native upstream https" covers the native upstream HTTPS/TLS
+        // client (`UpstreamTlsConn`) regressions alongside the pre-existing
+        // downstream "native TLS listener" cases (which also include the
+        // default-identity SAN/SNI eligibility tests) -- one step for every
+        // native-profile TLS-specific real-process integration case, so CI
+        // exercises both without needing the full (slower) `test-integration`
+        // suite.
+        .filters = &.{ "native TLS listener", "native upstream https" },
     });
     const run_native_listener_integration_tests = b.addRunArtifact(native_listener_integration_tests);
     run_native_listener_integration_tests.step.dependOn(b.getInstallStep());
-    const native_listener_integration_step = b.step("test-integration-native-tls", "Run native TLS listener HTTP integration tests");
+    const native_listener_integration_step = b.step("test-integration-native-tls", "Run native TLS listener and upstream HTTPS integration tests");
     native_listener_integration_step.dependOn(&run_native_listener_integration_tests.step);
 
     // #162: `tardi init <profile>` / `config init --profile` real
@@ -950,6 +958,14 @@ pub fn build(b: *std.Build) void {
     // The appliance credential loader (#392) reuses the PKI PEM/X.509
     // machinery; pki does not import tls_core, so this stays acyclic.
     tls_core_mod.addImport("pki", pki_mod);
+    // #634: `sni_provider.zig`'s default-identity SAN eligibility check
+    // (`defaultIdentitySatisfiesSni`) reuses `pki.x509`/`pki.identity`.
+    // `session_cache*_mod` build `session_cache.zig`/
+    // `session_cache_persistence.zig` as standalone test roots (not through
+    // `tls_core_mod`) but still reach `sni_provider.zig` transitively, so
+    // they need the same "pki" module wired directly.
+    session_cache_mod.addImport("pki", pki_mod);
+    session_cache_persistence_mod.addImport("pki", pki_mod);
     const pki_tests = b.addTest(.{ .root_module = pki_mod });
     const run_pki_tests = b.addRunArtifact(pki_tests);
     const pki_step = b.step("test-pki", "Run pure-Zig PKI DER unit tests");
--- a/docs/CONFIGURATION.md
+++ b/docs/CONFIGURATION.md
@@ -324,6 +324,13 @@ file (`/etc/ssl/certs/ca-certificates.crt`, `/etc/pki/tls/certs/ca-bundle.crt`,
 `native`; if neither an explicit bundle nor a well-known location is usable,
 verification fails deterministically rather than silently trusting nothing.
 
+`appliance`/`native` verification currently authenticates certificate chain
+signatures made with Ed25519, ECDSA P-256/SHA-256, or RSA-PSS/SHA-256 only
+(the pre-existing #343 native signature matrix); an origin signed with
+classic RSA PKCS#1 v1.5 or another unsupported algorithm fails verification
+even with a correct CA bundle and hostname (see `docs/TROUBLESHOOTING.md`).
+`general` has no such restriction.
+
 ### Health Checks And Circuit Breaking
 
 For active-probe aliases, `TARDIGRADE_UPSTREAM_PROBE_*` names are preferred only
--- a/docs/CONFIGURATION.md
+++ b/docs/CONFIGURATION.md
@@ -313,6 +313,17 @@ the feature or let runtime code choose a fallback. Boolean parsing accepts
 | `TARDIGRADE_UPSTREAM_TLS_CLIENT_CERT` | path | `""` | PEM client certificate for upstream mTLS. | `TARDIGRADE_UPSTREAM_TLS_CLIENT_CERT=/etc/tardigrade/upstream.crt` |
 | `TARDIGRADE_UPSTREAM_TLS_CLIENT_KEY` | path | `""` | PEM client private key for upstream mTLS. | `TARDIGRADE_UPSTREAM_TLS_CLIENT_KEY=/etc/tardigrade/upstream.key` |
 
+Behavior is identical across all three `-Dtls-profile` builds (#634): `general`
+uses the OpenSSL adapter's upstream client, `appliance`/`native` use the
+native upstream TLS client (`src/http/tls_termination_stub.zig`'s
+`UpstreamTlsConn`) — same knobs, same semantics, no OpenSSL. "System defaults"
+for an empty `TARDIGRADE_UPSTREAM_TLS_CA_BUNDLE` means the OpenSSL default
+verify paths on `general`, or the first readable well-known system CA bundle
+file (`/etc/ssl/certs/ca-certificates.crt`, `/etc/pki/tls/certs/ca-bundle.crt`,
+`/etc/ssl/cert.pem`, and similar OS-standard locations) on `appliance`/
+`native`; if neither an explicit bundle nor a well-known location is usable,
+verification fails deterministically rather than silently trusting nothing.
+
 ### Health Checks And Circuit Breaking
 
 For active-probe aliases, `TARDIGRADE_UPSTREAM_PROBE_*` names are preferred only
@@ -358,7 +369,7 @@ settings.
 | `TARDIGRADE_TLS_CIPHER_LIST` | string | `""` | OpenSSL TLS <=1.2 cipher list. Appliance/native profiles require empty. | `TARDIGRADE_TLS_CIPHER_LIST=ECDHE+AESGCM` |
 | `TARDIGRADE_TLS_CIPHER_SUITES` | string | `""` | TLS 1.3 cipher suites. Appliance/native profiles require empty. | `TARDIGRADE_TLS_CIPHER_SUITES=TLS_AES_256_GCM_SHA384` |
 | `TARDIGRADE_TLS_SERVER_NAME` | DNS name | `""` | Unused by the general/OpenSSL terminator. Appliance requires one non-wildcard DNS name when TLS is enabled; appliance credential/name changes require restart. | `TARDIGRADE_TLS_SERVER_NAME=edge.example.com` |
-| `TARDIGRADE_TLS_SNI_CERTS` | encoded list | `""` | Additional SNI certs as <code>name:cert:key&#124;name2:cert2:key2</code>. Appliance profile requires empty. | `TARDIGRADE_TLS_SNI_CERTS=api.example.com:/a.crt:/a.key` |
+| `TARDIGRADE_TLS_SNI_CERTS` | encoded list | `""` | Additional SNI certs as <code>name:cert:key&#124;name2:cert2:key2</code>. Appliance profile requires empty. On `native`, an explicit entry always wins for its hostname; for any hostname with no entry, the default (`tls_cert_path`/`tls_key_path`) identity is served instead if its own certificate's SAN already covers that hostname (#634), otherwise the handshake fails closed. | `TARDIGRADE_TLS_SNI_CERTS=api.example.com:/a.crt:/a.key` |
 | `TARDIGRADE_TLS_SESSION_CACHE` | bool | `true` general, `false` appliance/native | OpenSSL session cache. Appliance/native profiles reject true (see `TARDIGRADE_TLS_NATIVE_RESUMPTION_MODE`). | `TARDIGRADE_TLS_SESSION_CACHE=true` |
 | `TARDIGRADE_TLS_SESSION_CACHE_SIZE` | u32 | `20480` | OpenSSL session cache size. | `TARDIGRADE_TLS_SESSION_CACHE_SIZE=40960` |
 | `TARDIGRADE_TLS_SESSION_TIMEOUT_SECONDS` | u32 | `300` | OpenSSL session timeout. | `TARDIGRADE_TLS_SESSION_TIMEOUT_SECONDS=600` |
--- a/docs/CONFIGURATION.md
+++ b/docs/CONFIGURATION.md
@@ -329,7 +329,10 @@ signatures made with Ed25519, ECDSA P-256/SHA-256, or RSA-PSS/SHA-256 only
 (the pre-existing #343 native signature matrix); an origin signed with
 classic RSA PKCS#1 v1.5 or another unsupported algorithm fails verification
 even with a correct CA bundle and hostname (see `docs/TROUBLESHOOTING.md`).
-`general` has no such restriction.
+Independently, `appliance`/`native` also caps each peer certificate entry at
+2048 DER bytes and the whole handshake message at 8 KiB; a certificate or
+chain exceeding either bound fails even when its signature algorithm is
+otherwise supported. `general` has neither restriction.
 
 ### Health Checks And Circuit Breaking
 
--- a/docs/TLS_DEPENDENCY_POLICY.md
+++ b/docs/TLS_DEPENDENCY_POLICY.md
@@ -213,17 +213,26 @@ alongside the release assets and SBOM.
       foreign linkage, generic multi-identity credentials, deterministic
       rejection of OpenSSL-adapter-only settings, built and binary-audited in
       CI.
-- [x] Native upstream HTTPS/TLS client: `proxy_pass https://…` upstreams work
-      under `-Dtls-profile=native` with native PKI verification, hostname/SAN
-      checking, SNI, ALPN/protocol selection, connection pooling/reuse, and
-      bounded failure mapping — no OpenSSL, no runtime fallback. Verification
-      is currently limited to chains signed with Ed25519, ECDSA P-256/SHA-256,
-      or RSA-PSS/SHA-256 (the pre-existing #343 signature matrix); see the
-      caveat above.
+- [x] Native upstream HTTPS/TLS client exists and is wired into the data
+      plane: `proxy_pass https://…` upstreams work under `-Dtls-profile=native`
+      through the native TLS engine, record layer, and PKI stack — hostname/
+      SAN checking, SNI, ALPN/protocol selection, connection pooling/reuse,
+      and bounded failure mapping all function — no OpenSSL, no runtime
+      fallback.
+- [ ] Native upstream HTTPS/TLS client is **production-ready against ordinary
+      public HTTPS origins** (#634's actual upstream criterion) — **not yet
+      true, and this is a partial implementation until this box is checked.**
+      Verification currently authenticates only Ed25519, ECDSA P-256/SHA-256,
+      and RSA-PSS/SHA-256 certificate-chain signatures (the pre-existing #343
+      matrix, see the caveat above); a public origin signed with classic RSA
+      PKCS#1 v1.5 — still common among public CAs — or another unsupported
+      algorithm fails closed with a native-profile 502 even with a correct CA
+      bundle and hostname. Tracked by #645.
 - [ ] Native certificate-signature matrix (`src/pki/verify.zig`) extended to
       cover classic RSA PKCS#1 v1.5 (and other common public-WebPKI signature
       variants) so native-profile upstream verification is not restricted to
-      a subset of real-world certificate authorities.
+      a subset of real-world certificate authorities. Closing this item also
+      closes the one directly above (#645).
 - [x] Default downstream identity SAN/SNI eligibility: a default certificate
       may satisfy a requested SNI its own SAN actually covers, without
       requiring an explicit `TARDIGRADE_TLS_SNI_CERTS` entry; explicit
--- a/docs/TLS_DEPENDENCY_POLICY.md
+++ b/docs/TLS_DEPENDENCY_POLICY.md
@@ -59,21 +59,40 @@ the binary. There is no runtime switch and no fallback between profiles.
   multi-identity credential loading (default cert/key plus
   `TARDIGRADE_TLS_SNI_CERTS`), native TCP TLS termination, native QUIC/H3,
   SIGHUP-driven credential reload, opt-in native resumption
-  (`TARDIGRADE_TLS_NATIVE_RESUMPTION_MODE`). The credential store is
-  fail-closed on SNI, matching the existing native HTTP/3 behavior: the
-  default identity serves clients that send no SNI, and every host name
-  clients will request must be registered via `TARDIGRADE_TLS_SNI_CERTS`
-  (default-identity SAN matching is part of the remaining #634 parity
-  work).
+  (`TARDIGRADE_TLS_NATIVE_RESUMPTION_MODE`), and a **native upstream HTTPS/TLS
+  client** (`src/http/tls_termination_stub.zig`'s `UpstreamTlsConn`, #634) —
+  `proxy_pass https://…` upstreams are served entirely by the native TLS
+  engine, record layer, and PKI stack, never by OpenSSL.
+- Downstream SNI selection: an explicit `TARDIGRADE_TLS_SNI_CERTS` mapping
+  always wins first; failing that, the default identity is served if its own
+  certificate's SAN actually covers the requested hostname (RFC 9525 matching
+  via `src/pki/identity.zig`, not string heuristics — see
+  `sni_provider.UnknownSniPolicy.use_default_when_identity_matches`); failing
+  that, the handshake fails closed. Absent SNI still serves the default
+  identity unconditionally, unchanged from before.
+- Upstream TLS: certificate verification (`TARDIGRADE_UPSTREAM_TLS_VERIFY`,
+  `TARDIGRADE_UPSTREAM_TLS_CA_BUNDLE`) uses the same native PKI path-building/
+  validation as downstream mTLS would (`src/tls/webpki_verifier.zig`), falling
+  back to a small set of well-known system CA bundle file locations when no
+  bundle is configured; SNI/hostname derivation
+  (`TARDIGRADE_UPSTREAM_TLS_SERVER_NAME`), upstream ALPN/protocol selection
+  (`TARDIGRADE_UPSTREAM_PROTOCOL`), upstream client certificates (mTLS to the
+  origin), and existing connect/response timeouts all continue to work
+  unchanged; disabling verification still encrypts the connection, it just
+  skips identity checking. See `docs/CONFIGURATION.md` for the full knob
+  list and `docs/TROUBLESHOOTING.md` for native-profile-specific upstream TLS
+  failure modes.
 - Must not link `libssl`, `libcrypto`, or any other foreign TLS, crypto, QUIC,
   HTTP/3, or certificate library — audited identically to the appliance
   profile.
-- OpenSSL-adapter-only settings (TLS 1.2, cipher-string overrides, mTLS client
-  verification, the OpenSSL session cache/tickets, OCSP stapling, CRL checks,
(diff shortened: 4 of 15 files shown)


Jaruso (commented):
## Blocking review findings

I reviewed all 13 changed files against the current #634 reconciliation and, specifically, the two runtime acceptance criteria this PR is intended to close: production-ready native `https://` upstream support and default-identity SAN/SNI fallback.

I found **three blockers that should be treated as changes required before merge**:

1. **The WebPKI verifier is not yet compatible with ordinary public WebPKI chains.** The new verifier delegates certificate-path signatures to `pki.verify`, whose current matrix intentionally rejects classic RSA PKCS#1 v1.5 certificate signatures (and other common public-CA variants such as ECDSA/SHA-384). The integration fixture explicitly stays all-Ed25519 to avoid that limitation. That means many otherwise-valid system-CA HTTPS origins will fail closed with 502, so #634's “production-ready” upstream criterion and this PR's “ordinary HTTPS upstreams” DoD are not met yet.

2. **The new H2 `return` handling changes established H1/H3 semantics.** It bypasses the shared `ghandlers.planReturnResponse()` policy: redirects are supposed to be method-agnostic and a non-empty 3xx return body becomes `Location`, but the H2 path currently returns 405 for non-GET/HEAD and sends the redirect target as response content. This needs to reuse the shared plan and get protocol-parity tests.

3. **Current CI is still red on the SNI precedence test.** The second commit replaced one OpenSSL-version-dependent presentation string with another. On current head `2ba4e919`, the macOS native-profile job still fails looking for `Peer signature type: ecdsa_secp256r1_sha256`. The test should compare stable certificate material (served leaf DER/PEM/fingerprint) rather than `s_client`'s human-readable diagnostic text.

The rest of the native adapter looks structurally sound from this pass: the native client-role TLS wiring, SNI propagation, `skip_verify` behavior, ALPN selection, phase timeout handoff, connection ownership/pooling, mTLS credential seam, pending-plaintext handling, and default-identity SAN fallback all line up with the intended architecture.

GitHub does not permit the connected account to submit a formal `REQUEST_CHANGES` review on its own PR, so this is submitted as a comment review; **the three inline findings below are nevertheless merge-blocking findings**. Please address them and add the noted regression coverage before merging this runtime-cutover slice.

Jaruso (commented):
Second pass against current head `7900546b` and #634's current acceptance criteria.

The two prior mechanical issues are fixed: the SNI precedence test now compares the actual served DER and current CI is green on Linux x86_64, Linux ARM64, and macOS; the H2 `return` path now routes through `planReturnResponse()` and the redirect/method tests cover the original semantic regressions. I resolved those two old threads.

There are still **3 merge blockers** on the current head:

1. **#634's native-upstream production-readiness criterion is still not met; it has only been deferred to #645.** #634 still explicitly requires native `https://` upstream TLS/client support to be *production-ready* with verification, and its runtime-gap text is about ordinary HTTPS origins. #645 now correctly documents that valid, still-common RSA PKCS#1 v1.5 public-WebPKI chains produce a native-profile 502. Filing #645 is good follow-up hygiene, but it cannot simultaneously make this PR's `[x] ordinary HTTPS` / “shipping runtime suitable” claims true. Either land the practical WebPKI signature coverage before claiming this slice closes that #634 criterion, or re-scope this PR/documentation/description as a partial native-upstream implementation and leave the #634 production-ready item explicitly open until #645 lands.

2. **`UpstreamTlsConn.connect()` leaks/retains initialized TLS state on several error exits.** The only broad `errdefer` destroys the raw `UpstreamTlsState`; after `state.backend` is initialized, client-credential errors return without `backend.deinit()`, and after `state.record` is initialized a failure in the final ALPN `select()` destroys the state without `record.deinit()`. The downstream native constructor already demonstrates the correct staged-ownership pattern. Please add an ownership-aware cleanup guard so every post-backend failure deinitializes either the backend directly or the record (which owns backend teardown) before the state allocation is destroyed, and add focused failure-path coverage.

3. **The H2 `return` repair still misses error-code metrics parity.** `.return_response.method_not_allowed` builds the same `invalid_request` 405 payload as H1, but unlike H1's `recordReturnResponseMetrics()` and H2's adjacent `.local_rejection`, it never increments `err_invalid_request`. Add the error-code metric and assert it in the H2 regression test.

Two small truthfulness cleanups should also be folded into this round so they don't survive the cutover: `edge_gateway.zig` still logs that upstream mTLS/SNI overrides “apply to OpenSSL-backed connections” even though this PR makes them native too, and `tls_termination_stub.zig`'s `UpstreamTlsState` comment says the fd stays blocking/never gets `O_NONBLOCK` even though `connect()` immediately calls `setNonBlocking(fd)`.

Because the connected GitHub account owns this PR, GitHub does not permit a formal `REQUEST_CHANGES` review; this is submitted as `COMMENT`, but the three items above are blocking.

Jaruso (commented):
## Re-review: two remaining blockers

The fixes from the previous round are good. I rechecked the staged `UpstreamTlsConn` ownership, H2 return/metrics parity, the revised #634 scope wording, and the current CI state. The old WebPKI-scope, lifecycle, and H2-metrics threads are resolved; current CI and Darwin Release Smoke are green.

I found two remaining gaps that should be addressed before merging:

1. **#645 is not the only blocker on #634's “ordinary public HTTPS” criterion.** The native TLS engine independently rejects every peer `CertificateEntry` over 2048 DER bytes (`tls13_backend.max_certificate_len`) and caps a main-handshake message at 8 KiB. Those checks happen before `WebPkiVerifier`, so even after #645 adds RSA PKCS#1/ECDSA signature variants, otherwise-valid public chains can still fail solely on certificate/chain size. This is not merely theoretical: a valid Ed25519 CA-issued certificate (already in the supported signature matrix) can exceed 2 KiB simply from a larger SAN set. The docs/PR body therefore must not say #645 alone closes ordinary-HTTPS readiness. Either make the client-side certificate/handshake bounds practically WebPKI-compatible here, or track that separately and describe #645 as one known blocker rather than the remaining/sole blocker.

2. **Native upstream mTLS is claimed as supported but is not exercised through the new adapter.** I traced the engine path and confirmed handshake-time client auth exists (`CertificateRequest` -> `beginClientAuthFlight` -> Certificate/CertificateVerify), so the feature is plausible; however every new `native upstream https:` integration case omits `TARDIGRADE_UPSTREAM_TLS_CLIENT_CERT/KEY`. This PR is the first code that wires those existing engine capabilities into `UpstreamTlsConn`, so engine-level coverage alone cannot prove the adapter/configuration path. Add a real-process case that requires a client certificate and proves configured credentials succeed while missing credentials fail closed.

Once those are handled, I don't currently see another code blocker in this PR.

Jaruso (commented):
## Follow-up review — current head `5300a981`

The substantive findings from the earlier rounds are addressed: the SNI precedence test now compares served DER instead of OpenSSL presentation text; H2 uses the shared `planReturnResponse()` policy and records the 405 error metric; `UpstreamTlsConn.connect()` has staged backend/record teardown; and the #634 status is now truthful, with #645 and #646 separately tracking the two known public-WebPKI blockers. I also verified the new mTLS wiring and the `readReady()`/`close_notify` transport change structurally.

There is **one remaining merge blocker on this head**: the required macOS native-profile job is red in the new mTLS test. CI run `32641707034`, job `97199681405` fails `native upstream https: client certificate (mTLS) required by the origin` in `decodeChunkedHttpBody()` with `InvalidCharacter`; the same native job is green on Linux x86_64 and ARM64. The PR body currently checks `test-integration-native-tls -Dtls-profile=native` as green, which is not true for the current matrix. Please diagnose whether the platform difference is `openssl s_server -WWW` response behavior or a real streamed/chunked framing defect and fix the root cause; do not skip macOS or weaken the mTLS assertion.

I left one additional **non-blocking assurance note** on the H2 parity test: H2 now matches H1's zero-body redirect behavior, but the existing H3 renderer still puts the redirect target in the response body, so the test's claim of H1/H3 parity is stronger than reality. Either make H3 parity real or narrow the wording/file a follow-up.

Once the macOS native test is green on the exact head, I do not see another #643-specific runtime blocker in the latest changes. #634's broader production-ready upstream criterion correctly remains open for #645 + #646.

Jaruso (commented):
## Blocking re-review findings

I re-reviewed the two commits since `5300a981` and traced the second macOS failure through the native record layer and the streaming response error path. The previous macOS blocker is **not resolved yet**.

Two pieces of evidence matter here:

1. `14c1de63` did **not** fix the reproduced mTLS failure. CI run `32642579699`, macOS native job `97201842404`, still fails the exact same test with the exact same `decodeChunkedHttpBody -> InvalidCharacter` signature. The extra pooled-reuse timeout in that run may indeed be environmental pressure, but it is a different failure class (`ReadTimeout`) and does not explain a response that has already reached chunk decoding.
2. The proposed `peer_closed` explanation conflicts with the record-layer implementation. Once `handleAlert(close_notify)` sets `peer_closed`, `canProcessCarrierInput()` becomes false (`... && !self.peer_closed`), and `feedCiphertextInternal()` also immediately returns `bytes.len` when `peer_closed` is true. Therefore an extra `record.drive()` after `peer_closed` cannot drain the hypothetical trailing ciphertext the new comment says it is draining. TLS records are also processed in wire order, so application-data records preceding `close_notify` must be consumed before the alert can be reached.

More importantly, there is a concrete malformed-downstream-response path that matches the CI symptom. `streamProxyOverTransport()` writes the downstream `200`/`Transfer-Encoding: chunked` response head and sets its local `wrote_downstream = true`. If the native TLS/body relay then returns a non-client error, `executeStreamingHttpProxyRequest()` tears down the upstream and returns the error, but that commitment state is not exposed in its public result/error contract. `gateway_proxy_runtime.zig` catches the error and unconditionally calls `sendApiError(...)`. That can append a second `HTTP/1.1 502 ...` response after an already-started chunked `200`; the test decoder then sees `HTTP/1.1 ...` where it expects the next hexadecimal chunk-size line, which produces exactly `InvalidCharacter`.

Please fix the post-commit streaming-error contract rather than increasing the test timeout. After downstream response commitment, an upstream read/TLS failure must never serialize a second HTTP response. It should terminate/truncate the already-committed response appropriately (and record the upstream-abort outcome), while pre-commit failures may still synthesize 502/504. Also capture/assert the raw response in a regression so a second status line after a chunked head can never recur.

The earlier H3 redirect-body parity note remains non-blocking and unresolved; I am not adding a duplicate comment for it here.

Current latest-head CI (`ecd97951`) is queued, so the PR body's `[x] test-integration-native-tls ... green` claim is also not yet demonstrated on this head.

Jaruso (commented):
The previous double-response diagnosis is addressed for the direct HTTP/1 streaming path, and reverting the speculative `UpstreamTlsConn.read()` change plus fixing the H3 redirect body are both correct. I still found three blocking gaps in the follow-up, all around making the commitment fix complete and making the required mTLS proof trustworthy.

1. `downstream_committed` is only populated by the direct H1 loop. The `streamViaH2Pool()` route—including its ALPN-to-H1 fallback—can still throw after writing the downstream head without propagating commitment, recreating the exact second-response corruption.
2. Even when direct H1 correctly sets `downstream_committed`, the earlier `RequestCancelled` catch branch still unconditionally writes a 504. `relayUpstreamBody()` can raise `RequestCancelled` after the response head is committed, so the new guard has to apply before *every* response-writing error branch, not only the final catch-all.
3. The temporary 10-attempt mTLS retry from `0747dfac` must be removed now that `9251c4e1` established that its `InvalidHttpResponse` symptom came from a real product bug. Otherwise the required #634 mTLS regression can pass while swallowing recurrences of the very failure this PR is supposed to prove fixed.

Please make downstream commitment a complete streaming-exchange contract across H1, ALPN fallback, and H2; make all post-commit error classifications non-writing/connection-closing; and restore the mTLS success case to a single-shot assertion. Then the newest native CI matrix needs to pass without that retry.

Jaruso (commented):
Re-review of head `9faa2fa3`.

This round does not resolve the outstanding blockers. The only new commit changes `tests/integration.zig`: it renames the mTLS test so `test-integration-native-tls` no longer runs it. The previous exact head (`93e7a492`, CI run 32645345993, macOS native job 97208684138) failed this same test after exhausting all 10 retries with `error.MtlsUpstreamResponseNeverCompleted`. On the new head, the macOS native job is green because the filtered suite dropped from 22 tests to 21 (16 pass + 5 skip); that is removal of coverage, not evidence that the failing native mTLS exchange is fixed.

The two code-level blockers from my previous review are also still present because this commit does not touch those files:

1. `executeStreamingHttpProxyRequest()` still returns through `streamViaH2Pool()` without threading `downstream_committed`. Its ALPN-H1 fallback can propagate a post-head `streamProxyOverTransport()` error with the outer flag still false, and the real H2 body loop can return `RequestCancelled` after its response head is committed.
2. `gateway_proxy_runtime.zig` still handles `RequestCancelled` before the generic `downstream_committed` guard and unconditionally serializes a 504. Direct H1 therefore still has a post-commit cancellation path that can append a second HTTP response.

Those existing review threads remain blocking and need actual code changes plus the requested regressions.

The new test change adds a third blocking concern: do not make a required native feature proof 'best effort' simply because one required platform reliably catches a failure. This PR explicitly claims upstream mTLS is wired end-to-end and native-profile integration coverage proves the new paths. The macOS runner is part of the required native matrix; a 10/10 failure on that platform is evidence to diagnose, not a reason to remove the case from the gate. Please restore this test to the `native upstream https:` CI filter, remove the retry swallowing malformed/truncated responses, and make one clean mTLS exchange pass on macOS. If `openssl s_server -WWW` is genuinely the unstable component, replace only the HTTP-serving portion with a deterministic fixture while keeping an independent TLS/mTLS peer; do not remove the platform from the proof.

I would not merge this head.

Jaruso (commented):
Concrete implementation plan for the remaining blockers. Please treat this as the closeout recipe.

### A. Make `downstream_committed` a complete error-return contract

The invariant should be: **every error return from `executeStreamingHttpProxyRequest()` sets `downstream_committed=true` iff a downstream response head has already been successfully written.** Right now only the direct-H1 loop satisfies that.

1. Add `downstream_committed: *bool` to `streamViaH2Pool(...)`.
2. Pass the existing pointer from `executeStreamingHttpProxyRequest(...)` into the `streamViaH2Pool(...)` call.
3. In `streamViaH2Pool`'s `.h1 => |tls_ptr|` ALPN fallback, do not use bare `try streamProxyOverTransport(...)`. Preserve its local `wrote_downstream` on error:

```zig
var wrote_downstream = false;
const res = streamProxyOverTransport(
    allocator,
    tls_ptr,
    tls_ptr.fd,
    requested_relay_bytes,
    uri,
    method,
    extra_headers,
    buffered_body,
    streaming_body,
    downstream_conn,
    downstream_writer,
    security,
    alt_svc,
    sticky_set_cookie,
    correlation_id,
    connect_timeout_ms,
    read_deadline_ms,
    cancel_token,
    &wrote_downstream,
    proxy_buffer_limits,
    proxy_buffer_observer,
    .{ .origin = h1_origin_account, .global = proxy_buffer_global },
) catch |err| {
    downstream_committed.* = wrote_downstream;
    return err;
};
return res.result;
```

4. In the real `.h2` branch, set `downstream_committed.* = true` **immediately after** `gpres.writeStreamedUpstreamResponseHeadFromHeaders(...)` succeeds and before the body loop. Every failure before that successful write must leave it false.

### B. Restructure the runtime catch so post-commit paths can never call `sendApiError`

In `gateway_proxy_runtime.zig`, commitment must dominate every response-writing error branch. Recommended ordering:

```zig
if (err == error.ClientAborted) {
    // existing client-abort behavior
    ...
}

if (err == error.OutOfMemory) {
    // preserve existing OOM propagation; it does not synthesize a response
    ...
}

if (downstream_committed) {
    downstream_broken.* = true;

    if (err == error.RequestCancelled) {
        // preserve current circuit/upstream-failure accounting
        if (absolute_proxy_target) {
            state.circuitRecordFailurePermit(circuit_permit);
        } else {
            state.recordProxyUpstreamFailure(cfg, selection.base_url, circuit_permit);
        }
        if (ctx.lifecycle) |lc| lc.logTimeout("upstream_connect");
        ctx.setUpstreamResult(
            resolved.upstream_host,
            @intFromEnum(http.Status.gateway_timeout),
            0,
        );
        // IMPORTANT: no sendApiError here.
        return @intFromEnum(http.Status.gateway_timeout);
    }

    // Generic post-commit upstream failure. Preserve current health/circuit
    // accounting and 502-vs-504 classification, but never serialize again.
    if (absolute_proxy_target) {
        state.circuitRecordFailurePermit(circuit_permit);
    } else {
        state.recordProxyUpstreamFailure(cfg, selection.base_url, circuit_permit);
    }
    const status: http.Status = switch (err) {
        error.Timeout, error.WouldBlock => .gateway_timeout,
        else => .bad_gateway,
    };
    ctx.setUpstreamResult(resolved.upstream_host, @intFromEnum(status), 0);
    // IMPORTANT: no sendApiError here.
    return @intFromEnum(status);
}

// Only below this point are we provably pre-commit. Existing 413/503/504/502
// branches may call sendApiError here.
```

You can factor the duplicated classification/accounting into a helper. The important part is structural: **there must be no `sendApiError` reachable when `downstream_committed` is true.**

### C. Add the two missing commitment regressions

Keep the current direct-H1 RST regression and add both of these:

1. **ALPN-H1 fallback:** HTTPS + `TARDIGRADE_UPSTREAM_PROTOCOL=auto`, H1-only TLS origin, valid response head + partial body followed by a reset/failure. The request must enter `streamViaH2Pool(...).h1`. Read raw downstream bytes and assert exactly one `HTTP/1.1 ` status line. Assert the connection truncates/closes rather than receiving an appended 502/504.

2. **Cancellation after commitment:** stream a response that writes its head, then stalls the body until the existing cancellation/deadline fires. Read raw downstream bytes and assert exactly one status line and no appended `HTTP/1.1 504`.

### D. Restore a real CI-gated native mTLS proof

The current rename is not a fix. Restore the `native upstream https:` prefix so `test-integration-native-tls` runs the case on macOS, Linux x86_64, and Linux ARM64.

Remove the 10-attempt retry. One bounded exchange must succeed cleanly. Keeping the 20s timeout is fine; swallowing `InvalidHttpResponse`, `ReadTimeout`, and `ConnectionResetByPeer` is not.

Simplify the fixture instead of removing macOS from the proof. Keep an implementation-independent TLS/mTLS server that **requires and validates the client certificate**, but do not use `openssl s_server -WWW`'s demo HTTP framing as the application layer. Make the peer emit deterministic HTTP/1.1 bytes with explicit length, e.g.:

```text
HTTP/1.1 200 OK\r\n
Content-Length: <exact byte length>\r\n
Connection: close\r\n
\r\n
native-upstream-mtls-body
```

If the cleanest solution is a tiny test-only OpenSSL-backed helper peer rather than `s_server -WWW`, do that. The success case must perform one request and assert 200 + body; the control without `TARDIGRADE_UPSTREAM_TLS_CLIENT_CERT/KEY` must still fail closed with 502.

After these changes, rerun the exact native matrix. I want the same single-attempt mTLS success/control proof green on all required platforms.

Jaruso (commented):
Line-level implementation note for the commitment contract.

Jaruso (commented):
Line-level implementation note for post-commit error handling.

Jaruso (commented):
Line-level implementation note for restoring the required mTLS proof.

Jaruso (commented):
Re-review of `a5702de9`.

The production fixes from the last round are now implemented correctly: `downstream_committed` is threaded through `streamViaH2Pool`, the ALPN→H1 catch propagates `wrote_downstream`, the real H2 branch marks the commitment boundary immediately after the response head write, and `RequestCancelled` no longer unconditionally writes a second 504 after commitment. I also re-read the rebuilt mTLS proof: it is back under the required `native upstream https:` filter, the retry loop is gone, and `OpensslRawPeer` gives it deterministic `Content-Length` framing while OpenSSL still independently enforces `-Verify 1 -CAfile`. That is the right shape.

**One blocker remains: do not defer item C.** This commit explicitly says the two path-specific regressions were left for a follow-up. Those tests were part of the requested closeout because the first implementation of this same bug fixed one path while silently leaving two others. The code now looks right, but the exact new branches have no regression that would fail if either commitment assignment disappears later.

Please add both tests in this PR:

1. **ALPN→H1 fallback, post-head failure:** exercise the `.h1` arm returned by `streamViaH2Pool`, make `streamProxyOverTransport` fail after it has written the downstream 200 head, and assert the propagated commitment is true / no replacement response is serialized. If producing a deterministic TLS RST is awkward, extract the small negotiated-H1 relay/catch into a helper that takes the transport generically, call that helper from the `.h1` arm, and unit-test it with a fake transport that returns a valid HTTP head + partial body and then `error.ConnectionResetByPeer`. The test must prove `downstream_committed == true` on the error path, not merely test a boolean helper in isolation.

2. **H2 mid-relay cancellation:** reuse the existing `H2SlowResponseOrigin` test scaffold. Have it send `:status 200` HEADERS and hold the body. Use a test writer that cancels a `CancellationToken` only after it has captured the complete downstream response head (`\r\n\r\n`). Pass that token to `streamViaH2Pool`; the next body-loop `cancelStopped()` must return `error.RequestCancelled`. Assert `downstream_committed == true`. Together with the existing raw-response RST regression in `tests/integration.zig`, this pins both halves of the contract: the transport layer reports commitment, and the runtime does not append a second response after commitment.

The current exact-head CI run is still in progress; macOS native has not reached `test-integration-native-tls` yet, so the restored single-shot mTLS case still needs to go green there before merge. #645/#646 remain correctly documented as separate blockers to #634's ordinary-public-HTTPS criterion and are not blockers to merging this focused partial slice.

Jaruso (commented):
**Update after the exact-head CI completed:** macOS is still blocking merge. Run `32647621863`, job `97214197223` (`Test native profile (macos-14)`) ran the restored 22-test native TLS suite and failed the single-shot mTLS success case with `ReadTimeout` at `tests/integration.zig:17731`. So the old malformed-chunk symptom is gone, but the replacement `OpensslRawPeer` still does not reliably deliver the response on macOS.

The failure points directly at the remaining harness assumption: this test writes the HTTP response to `s_server` stdin **before any TLS client has connected**, and assumes OpenSSL will buffer those bytes and later flush them into the first established TLS session. That may happen with the contributor's OpenSSL build, but the macOS runner demonstrates it is not a portable contract we can base required CI on.

Please make response emission request-synchronized, not preloaded:

1. Change `OpensslRawPeer.start()` to spawn with both `.stdin = .pipe` **and `.stdout = .pipe`**. Keep `-quiet`, `-Verify 1`, and `-CAfile`; OpenSSL remains the independent mTLS implementation.
2. Add a helper that blocks reading `child.stdout` until it has received the proxied HTTP request head through `\r\n\r\n` (bounded to e.g. 64 KiB and a 5–10s deadline). Only after that request is observed should it write the fixed `Content-Length` response to `child.stdin`.
3. Because the downstream request must be in flight while the helper waits for the upstream request, run the responder on a small test thread. The deterministic sequence should be:
   - start `OpensslRawPeer`;
   - start Tardigrade;
   - spawn responder thread: `waitForHttpRequestHead(); sendToClient(response_bytes);`
   - perform exactly one `sendRequestWithTimeout(...)` to Tardigrade;
   - join responder and assert it observed a request + wrote the response;
   - assert downstream 200/body.
4. Alternatively, and even more robustly, add a tiny **test-only OpenSSL server helper executable** that does `SSL_accept -> SSL_read until \r\n\r\n -> SSL_write fixed Content-Length response -> SSL_shutdown`. #634 explicitly permits foreign implementations in test/interop tooling, and this avoids depending on `s_server`'s interactive stdin semantics entirely.
5. Keep the negative control with no client cert/key and expect 502. Do not reintroduce retries, rename the test out of the filter, or special-case macOS.
6. Required proof before merge: this exact `native upstream https: client certificate (mTLS) required by the origin` case passes single-shot in `test-integration-native-tls` on macOS, Linux x86_64, and Linux ARM64.

The two path-specific commitment regressions requested in review `5002701248` are still required as well. Production commitment logic looks correct now; remaining work is deterministic assurance plus getting the exact-head native matrix green.

Jaruso (commented):
Review result: **merge-ready** on `3fbe355f`.

I re-reviewed the two commits since `a5702de9` against the outstanding #643/#634 acceptance items and the previous change requests.

The previously blocking items are now satisfied:

- `streamViaH2Pool`'s negotiated-H1 path now goes through `runNegotiatedH1Exchange`, which owns the real `streamProxyOverTransport` commitment contract and propagates `wrote_downstream` on error.
- The real H2 path marks `downstream_committed` immediately after the downstream head succeeds, before its body-loop cancellation/error boundary.
- The `RequestCancelled` runtime path honors `downstream_committed` and closes the broken downstream instead of serializing a second 504.
- Both previously requested regressions are present and deterministic: negotiated-H1 post-head timeout => committed, and real H2 mid-relay cancellation => `RequestCancelled` + committed.
- The mTLS proof is back under the required `native upstream https:` filter, has no retry loop, and now synchronizes the fixed-length response to an actually observed request on `openssl s_server` stdout rather than relying on pre-connect stdin buffering.
- Exact current-head CI is green, including the macOS native job's full filtered suite (`17 pass, 5 skip`, 22 total), plus the Linux native legs and the Darwin release smoke.

I also rechecked the scope against #634. The PR continues to state correctly that #645 (public-WebPKI signature coverage) and #646 (production-sized certificate/handshake bounds) remain independent blockers to the broader ordinary-public-HTTPS production-readiness criterion. Those follow-ups do not block merging this scoped mechanism/SNI slice.

I found no remaining production-code blocker. The inline note below is only failure-path cleanup in the test harness and does **not** change this merge-ready conclusion.

PR #644, opened 2026-08-23
Seed release-backed Homebrew formula tooling

Jaruso:
## Summary
- keep the checked-in Homebrew formula preparatory until a native #634 release exists; no longer points at the old v0.5.0 artifacts
- harden formula rendering so production updates use --tag, derive checksums/assets from one release, and reject releases without native dependency inventories
- add synthetic and release-backed Homebrew smoke scripts; both exercise brew test, installed binary linkage, and startup/request behavior when a valid formula is available
- add Linux x86_64 and aarch64 Homebrew smoke CI coverage and document the release/tap publication gates

## Validation
- ruby -c packaging/homebrew/tardigrade.rb
- bash -n scripts/update-homebrew-formula.sh scripts/test-homebrew-formula.sh scripts/test-homebrew-release-formula.sh
- shellcheck scripts/update-homebrew-formula.sh scripts/test-homebrew-formula.sh scripts/test-homebrew-release-formula.sh
- actionlint .github/workflows/ci.yml
- renderer fixture render to non-canonical formula path
- fixture mode rejects canonical formula overwrite and --tap-dir writes
- ./scripts/update-homebrew-formula.sh --tag v0.5.0 correctly rejects the old non-native release because it lacks native dependency inventories
- zig build -Doptimize=ReleaseFast -Dtls-profile=native -Dversion=0.0.0-homebrew-smoke
- ./scripts/audit-release-binary.sh --binary zig-out/bin/tardi --profile native --output /tmp/tardigrade-native-homebrew-smoke-inventory.json
- ./scripts/test-install.sh
- ./scripts/test-homebrew-release-formula.sh correctly exits while the checked-in formula is preparatory

## Notes
- I did not modify Bare-Systems/homebrew-tap because another agent is working there. #466 still requires tap main to be reconciled before the issue can close.
- Local ./scripts/test-homebrew-formula.sh reaches Homebrew download/checksum verification but is blocked by this Mac's outdated Apple Command Line Tools before install. CI now runs the full synthetic Homebrew smoke on Linux x86_64 and aarch64.

Refs #466


Jaruso (commented):
**Blocking review — do not merge this head yet.**

I reviewed all 9 changed files against #466, the current `v0.5.0` tag/release implementation, the current `main` packaging contract, the live `Bare-Systems/homebrew-tap` contents, and this PR's CI run. GitHub will not let the PR author formally submit `REQUEST_CHANGES` on their own PR, so this is a `COMMENT` review, but the findings below are change-request level.

The overall direction is good: one generated formula source, a real Homebrew install path, and reuse of `audit-release-binary.sh` are the right building blocks. However, this head has several correctness/acceptance gaps.

### 1. The checked-in formula points at the wrong artifact family

This is the biggest blocker. `v0.5.0` predates the native shipping cutover required by #466/#634:

- At tag `v0.5.0`, `.github/workflows/release.yml` installs `libssl-dev` and builds with plain `zig build -Doptimize=ReleaseFast "-Dversion=$VERSION"` (no `-Dtls-profile=native`).
- The current build defaults the TLS profile to `general`, i.e. the transitional OpenSSL adapter.
- At tag `v0.5.0`, `package-release-archive.sh` packages the real executable as `tardigrade` and makes `tardi -> tardigrade`; current `main` has the opposite/canonical layout (`tardi` real, `tardigrade -> tardi`).

So the formula references real checksums, but they are checksums for a release that does **not** satisfy the native-only shipping contract and does not have the same archive layout exercised by this PR's local smoke. The smoke creates a new archive from today's native `zig-out/bin/tardi`, so it can pass while the actual URL in the checked-in formula is unusable or violates #466.

Expected fix: do not publish/advertise a release-backed canonical formula against `v0.5.0`. Either wait for/use the first published release containing the #634 native cutover and current archive layout, or keep this PR explicitly preparatory until such a release exists. In addition to the synthetic local-archive smoke, add a release-backed gate that installs the exact checked-in formula URL/checksum and audits the installed bytes, conceptually:

```bash
brew install --formula packaging/homebrew/tardigrade.rb
installed="$(brew --prefix)/bin/tardi"
"$REPO_ROOT/scripts/audit-release-binary.sh" \
  --binary "$installed" --profile native \
  --output "$TMPDIR/dependency-inventory.json"
"$installed" version
```

That is the test which proves the public release bytes referenced by the formula are the same native artifact family required by #634.

### 2. The new runtime smoke currently fails in CI

The `Packaging smoke` job on this head failed in `Smoke test Homebrew formula`. The generated config uses a nested `server { ... }` block; Tardigrade validates it but starts on its default `0.0.0.0:8069`, while the test curls `127.0.0.1:18089`. The log shows ten connection failures followed by `Tardigrade edge listening on 0.0.0.0:8069`.

Use the actual top-level Tardigrade config grammar already used by the project, for example a top-level `listen 18089;` plus a deterministic `location = /health { return 200 ...; }`, then curl that endpoint.

### 3. The Homebrew formula's own `test do` is broken and is never exercised

`tardi version` prints the embedded version/profile, e.g. `0.5.0 (tls-profile=native, tls-backend=native)` (or `dev (...)` for this CI build). It does not contain the literal `Tardigrade`, so `assert_match "Tardigrade", ...` is wrong. `brew install` does not run `test do`, and the smoke never invokes `brew test`, so this escaped the new test path.

Please make the formula test assert the real contract and execute it in CI. For a release-backed formula this can be:

```ruby
output = shell_output("#{bin}/tardi version")
assert_match version.to_s, output
assert_match "tls-profile=native", output
assert_equal (bin/"tardi").realpath, (bin/"tardigrade").realpath
```

Then run `brew test` against the tapped formula. If the synthetic smoke keeps the fake formula version, build the smoke binary with the same `-Dversion=...` so the formula test is meaningful.

### 4. The renderer does not guarantee version/tag/checksums come from one release

`--version` and the checksum source are independent inputs. This currently succeeds:

```bash
./scripts/update-homebrew-formula.sh \
  --version 0.5.1 \
  --checksums-url https://github.com/Bare-Systems/Tardigrade/releases/download/v0.5.0/tardigrade-checksums.txt
```

and renders `v0.5.1` URLs with `v0.5.0` hashes. That directly violates #466's requirement that version/tag/checksums all come from one release and that assets exist before formula references are updated.

Please make a release/tag the authoritative input and derive the manifest + asset URLs from that release metadata, or at minimum validate that a `--checksums-url` is for `v${VERSION}` and verify every emitted archive is present in that same release. The live tap agent's Ruby updater is already using the stronger pattern of resolving `/releases/tags/<tag>`, enumerating its assets, and then downloading that release's checksum manifest; whichever implementation becomes canonical should preserve that coupling.

### 5. The CI determinism gate is permanently pinned to `0.5.0`

The lint job always re-renders from `v0.5.0`. A legitimate future formula update to `0.5.1` will fail `git diff --exit-code` until someone manually edits CI too. That is contrary to the maintainable release-driven update path this issue is trying to establish.

Please derive the current formula release from one checked-in source of truth, or test rendering against a stable fixture separately from validating the currently published formula.

### 6. Advertised Linux ARM64 coverage is not formula-smoked

The formula emits both Linux x86_64 and aarch64 branches, but the Homebrew smoke runs only on `ubuntu-latest` x86_64. Homebrew currently treats both Intel x86_64 and ARM64/AArch64 Linux as supported architectures, and this repository already has `ubuntu-24.04-arm` jobs. #466 explicitly requires advertised Linux branches to be formula-tested where Homebrew-on-Linux can exercise them honestly.

Please add a Homebrew smoke matrix for `ubuntu-latest` and `ubuntu-24.04-arm` (it can be a smaller dedicated job rather than duplicating every packaging smoke).

### 7. Make the runtime assertions use the installed formula binary explicitly

After install, the script audits `$(brew --prefix)/bin/tardi`, but `tardi version`, `tardi check`, and `tardi run` are invoked through ambient `PATH`. CI happens to get the Homebrew path first after `brew shellenv`, but locally another `tardi` can make the startup/request part of the smoke exercise the wrong binary. Resolve the installed path once and use it for every assertion/run.

### 8. Reconcile the tap ownership model before claiming #466 completion

I also checked `Bare-Systems/homebrew-tap`. Its current `codex/466-seed-homebrew-tap` branch has a separate `scripts/update-tardigrade-formula.rb`, and its README describes that tap-local updater as the release update flow. This PR says `Tardigrade/packaging/homebrew/tardigrade.rb` + `scripts/update-homebrew-formula.sh` is canonical and the tap is only a generated copy. Those are two competing ownership models right now.

The companion tap work needs to be reconciled to one model (remove the independent updater or make the tap canonical and remove the duplicate renderer here). #466 explicitly requires a non-divergent source of truth.

Also, `packaging/homebrew/tap-README.md` currently shows the generic `brew tap` / `brew install` commands without a Linux-only/current-release warning. Until a native release that actually satisfies the formula exists, do not publish that README as though the public path is generally usable; and while Darwin assets are absent, state explicitly that macOS Homebrew is not yet supported.

### Acceptance-criteria status at this head

- Canonical ownership: **not yet reconciled with the tap repo**.
- Real URLs/checksums: **real v0.5.0 hashes, but wrong/non-native artifact family for #466**.
- Linux x86_64 formula smoke: **implemented but currently failing**.
- Linux aarch64 formula smoke: **missing**.
- Darwin formula branches: correctly absent until real assets exist.
- No formula OpenSSL declaration: yes, but the referenced v0.5.0 binary itself was built through the old OpenSSL-backed shipping path.
- Native installed-binary audit: present for the synthetic current-main archive, **not the actual release bytes referenced by the formula**.
- Compatibility alias: synthetic current-main layout is tested; **v0.5.0's opposite alias layout is not**.
- Release-to-tap flow: documented, but version/checksum coupling and cross-repo ownership still need hardening.

Once these are corrected, this should be a strong implementation of the Homebrew lane.

Jaruso (commented):
Re-review complete on current head `88f473d60ce70f4563f423f7d5813d641a8f53a6`.

There is no new commit since my previous review, so none of the blocking findings have been addressed yet. I verified the branch directly rather than relying only on PR metadata: `scripts/test-homebrew-formula.sh` still contains the nested `server { listen 127.0.0.1:18089; ... }` config that caused the packaging-smoke failure, and the checked-in formula still targets `v0.5.0`.

Current status remains **do not merge**:

1. The formula still references `v0.5.0`, whose published artifacts predate the native #634 shipping cutover and use the old executable/archive layout.
2. The Homebrew runtime smoke is still broken; CI is red because Tardigrade starts on its default `8069` while the smoke curls `18089`.
3. The formula `test do` still asserts `"Tardigrade"` against output that actually contains version/profile metadata, and the smoke still does not run `brew test`.
4. Renderer inputs still allow version/tag/checksum manifest mismatch.
5. The determinism CI gate remains hard-coded to `0.5.0`.
6. Advertised Linux aarch64 coverage is still not Homebrew-smoked.
7. Runtime smoke commands still use ambient `PATH` rather than consistently invoking the installed formula binary.
8. The source-of-truth conflict with the independent updater in `Bare-Systems/homebrew-tap` remains unresolved.

The existing inline comments from the previous review are still current and contain the concrete expected fixes/code shapes. I did not add duplicate inline comments because the relevant lines and findings are unchanged. Please push the remediation commit(s), then this is ready for another substantive review.

Jaruso (commented):
**Second-round review — still change-request level, but the first-round blockers were substantially addressed.**

I re-reviewed the full 10-file diff at `1926346d`, re-checked #466's acceptance criteria, re-ran the earlier findings against the new implementation, inspected the current CI failures, and—per maintainer direction—checked the **actual `Bare-Systems/homebrew-tap` `main` branch**, not an agent branch.

The previous six inline findings are fixed in this head, and I resolved those threads: v0.5.0 is no longer emitted as a supposedly native formula, the renderer's production path is tag-coupled to one release and checks its inventories/assets, the formula test now asserts the real native version contract and `brew test` runs, CI covers Linux x86_64+aarch64, the runtime config uses the real top-level grammar, and the tap README is now explicitly preparatory.

There are five remaining changes I would make before merging this head:

1. **CI is currently red in ShellCheck.** The cleanup trap suppression uses `SC2329`, but this runner reports `SC2317` for the trap-only cleanup body. The same pattern exists in both Homebrew smoke scripts. Use the code ShellCheck is actually emitting (or restructure the cleanup so no suppression is needed).

2. **The x86_64 Homebrew smoke is also red, for a different reason.** It now binds the intended `18089` port, but every `/health` request returns 404. The config declares `server_name localhost` while curl sends `Host: 127.0.0.1:18089`. Tardigrade's own troubleshooting contract explicitly says unmatched `Host` returns 404. Use `curl -H 'Host: localhost' http://127.0.0.1:18089/health` or curl `http://localhost:18089/health`. This should be applied to the shared runtime probe used by both synthetic and release-backed coverage.

3. **The release-backed smoke still stops short of #466's runtime acceptance criterion.** It installs the exact release URL/checksum bytes, runs the formula test, prints version, and audits linkage—but never starts those exact installed bytes and serves a real request. The synthetic smoke does that against locally repackaged bytes. Please run the same `check -> run -> curl -> terminate` probe in `test-homebrew-release-formula.sh` (ideally factor it into one helper used by both scripts) so a published archive cannot pass purely on metadata/linkage while failing at startup.

4. **Fixture mode is described as local-only but is not mechanically contained.** `--version/--checksums` can still use the default canonical `packaging/homebrew/tardigrade.rb` output and can still be combined with `--tap-dir`, bypassing the `--tag` release-asset/inventory validation and copying fixture data into a tap checkout. Production safety should be enforced in code, not by the usage text. For example, reject `--tap-dir` unless `--tag` is set, and require fixture mode to receive an explicit non-canonical `--formula` path.

5. **The synthetic smoke's hard-coded formula version does not match the documented local invocation.** The script renders `version "0.0.0-homebrew-smoke"` and now runs `brew test`, which asserts that version appears in the binary. CI builds with `-Dversion=0.0.0-homebrew-smoke`, but the docs still show `zig build -Doptimize=ReleaseFast -Dtls-profile=native` before running the smoke; that binary reports `dev`, so the documented manual workflow fails. Either derive/parameterize the fixture version from the input binary, or make the required `-Dversion=0.0.0-homebrew-smoke` build explicit everywhere this smoke is documented.

### Cross-repo `homebrew-tap/main` state

I checked `Bare-Systems/homebrew-tap` **main** at commit `2cabb7139b2127eac204950e3096f54c4edfa4b5`. It currently contains all three things this remediation correctly says must not be the final state:

- `Formula/tardigrade.rb` still publishes the **v0.5.0** Linux archives;
- `README.md` currently advertises `brew tap Bare-Systems/tap && brew install tardigrade` as a usable path;
- `scripts/update-tardigrade-formula.rb` is a separate tap-local release updater, and the README describes it as authoritative.

So this PR's new ownership documentation is directionally correct, but **#466 cannot be considered complete until actual tap `main` is reconciled**: remove/disable the stale v0.5.0 public formula/install advertisement, choose exactly one updater/source-of-truth, and later sync the first native release-backed formula through that path. If this PR is intentionally a preparatory Tardigrade-side slice, that cross-repo reconciliation can follow separately; it just must remain an explicit close blocker for #466.

Finally, please update the PR description itself. It still says this PR seeds a release-backed `v0.5.0` formula and that the current v0.5.0 formula is supported, which is now the opposite of the corrected implementation.

GitHub does not permit this account to submit `REQUEST_CHANGES` on its own PR, so this is submitted as `COMMENT`, but I would hold merge until the four code/test issues above and the current red CI are fixed.

Jaruso (commented):
Post-merge re-review of `b26b412d14a75711b21dedc38526f116f1f93c72` together with the actual `Bare-Systems/homebrew-tap/main` (`2cabb7139b2127eac204950e3096f54c4edfa4b5`).

The Tardigrade-side remediation is substantially complete: all previous review findings are fixed, those threads are now resolved, the final CI run is green, and both Linux x86_64 and aarch64 Homebrew synthetic smokes pass. The tag/checksum/native-inventory coupling, release-backed runtime smoke, explicit installed-binary usage, and fixture/production boundary are all in good shape.

This merged PR should still be treated as **preparatory work, not completion of #466**, because the public tap's `main` branch has not been reconciled:

1. `homebrew-tap/main` still publishes `v0.5.0`, which predates the #634 native cutover and therefore violates the native shipping contract.
2. Its README still advertises `brew tap Bare-Systems/tap` / `brew install tardigrade` against that invalid release.
3. It still carries `scripts/update-tardigrade-formula.rb`, creating a second independent formula generator despite this PR documenting Tardigrade as canonical.
4. Its formula's runtime test declares `server_name localhost` but curls `127.0.0.1` without a matching `Host`, so the same virtual-host 404 defect caught earlier is present on tap main.
5. Tap CI runs only `ubuntu-latest`; once a valid formula advertises Linux aarch64, the public formula should be exercised on ARM too, and when Darwin branches are published, supported macOS architectures need clean Homebrew install/runtime coverage before #466 closes.
6. The cross-repo release-to-tap write path is still manual; #466 still needs the chosen PR/bot flow plus narrow, secret-safe credentials and failure-visible/idempotent behavior.

I found one additional Tardigrade-side future-proofing issue inline: `--tap-dir` currently overwrites the public README with a static pre-release warning on every future formula update.

Net: the merged Tardigrade tooling is a solid canonical foundation, but **do not close #466 or advertise the public tap until tap main is replaced/reconciled and the first native release-backed formula passes the exact public-path smokes.**

PR #647, opened 2026-08-23
pki: extend native certificate-signature matrix to RSA PKCS#1 v1.5 (#645)

Jaruso:
## Summary

Closes #645. Extends the native X.509 certificate-chain signature matrix
(`src/pki/verify.zig`, originally #343) to also verify RSA PKCS#1 v1.5
signatures — `sha256WithRSAEncryption` (required) and
`sha384WithRSAEncryption` (small additional surface area over the same
primitive) — the classic RSA signature scheme still common among public CAs
and intermediates, which was blocking the native upstream HTTPS/TLS client
(#634/#643) from authenticating ordinary public WebPKI chains.

**This is certificate-chain-signature verification only.** RFC 8446 §4.2.3
forbids `rsa_pkcs1` schemes in TLS 1.3 `CertificateVerify`, so the new
`provider.SignatureScheme` variants are never advertised or selected for the
handshake's own proof-of-possession step — `src/tls/crypto_profile.zig`'s
`supportsSignatureScheme` continues to unconditionally refuse them, and a new
regression (`adding the RSA PKCS#1 v1.5 provider primitive does not advertise
it for TLS CertificateVerify`) proves this holds even after the provider
gains the primitive.

- **`src/crypto/rsa.zig`**: `verifyPkcs1v15Sha256`/`verifyPkcs1v15Sha384`,
  reusing the existing RSA public-key parser, accepted modulus sizes, and
  `std.crypto.ff.Modulus` public-exponent exponentiation (no second
  bignum/modexp stack). Validates EMSA-PKCS1-v1_5 strictly by constructing
  the exact expected encoded message for the modulus width and
  constant-time-comparing it against the recovered block — no permissive,
  partial, or prefix-only DigestInfo parsing.
- **`src/crypto/provider.zig`/`pure_zig.zig`**: two new `SignatureScheme`
  variants, wired into `Provider.verify` and pure-Zig capability discovery.
- **`src/pki/verify.zig`**: `resolveScheme` recognizes the new schemes and
  requires an RSA issuer key; `validateIssuerKeyParameters`/
  `issuerPublicKey`/`signatureBytes` give PKCS#1 the same strict RSA-key and
  modulus-sized-signature discipline as RSA-PSS. `sha256/384WithRSAEncryption`
  `AlgorithmIdentifier` parameters accept explicit NULL or absent (common
  interop variance), reject anything else.
- **`src/tls/sni_provider.zig`**: exhaustive-switch fix — the new schemes
  deliberately have no TLS credential-signing counterpart.
- **`src/crypto/profile.zig`**: two new capability-matrix rows (`pki`-only
  integration, general-purpose-only product enablement, mirroring the
  RSA-PSS row — native-appliance product enablement is a separate #391
  policy decision this PR doesn't make).
- **`src/pki/verify_tests.zig`**: the existing `rsa_ca.crt`/`ecdsa_leaf.crt`
  fixtures (RSA CA self-signature; P-256 leaf cross-signed by the RSA CA)
  now verify instead of returning `UnsupportedSignatureAlgorithm`, plus
  tampered-TBS, tampered-signature, wrong-issuer-key, malformed-signature,
  and `AlgorithmIdentifier`-parameter coverage.
- **`tests/integration.zig`**: a real `native upstream https:` integration
  test drives a full downstream request → gateway proxy → `UpstreamTlsConn`
  → native TLS client → `WebPkiVerifier` → path validation →
  `verifyCertificateSignature` → pure-Zig provider → RSA PKCS#1 verifier,
  using the existing `sha256WithRSAEncryption` `tests/fixtures/tls/{ca,server}.crt`
  fixtures (747/804 DER bytes, safely below #646's 2 KiB/8 KiB bounds, so
  this test isolates #645 from that separate open blocker). `ca.crt` needed
  reissuing (same key/subject/validity) to add the `basicConstraints
  critical,CA:TRUE` extension RFC 5280 requires of a CA cert and the native
  trust store enforces — `server.crt` re-verifies against the reissued cert
  unchanged, and the two existing tests that hand this pair to real `openssl
  s_server -CAfile` are unaffected.
- **`tests/crypto_vectors.zig`**: new OpenSSL-sourced positive/negative
  vector coverage (`tests/vectors/rsa_pkcs1/`) for both schemes, required by
  the checked-in capability-matrix/vector-coverage contract.
- **Docs**: `TLS_DEPENDENCY_POLICY.md`, `CONFIGURATION.md`,
  `TROUBLESHOOTING.md` updated to remove the RSA PKCS#1 v1.5 caveat while
  explicitly retaining #646 (certificate/handshake size bounds) as an
  independent, still-open blocker, and truthfully noting that other
  algorithms (e.g. ECDSA/SHA-384) remain unsupported.

## Out of scope (intentionally)

- #646 (2 KiB cert / 8 KiB handshake bounds) — untouched, still open.
- OpenSSL production profile — untouched.
- Default build profile — untouched.
- ECDSA/SHA-384 or any other new curve/provider architecture.
- #634's broader "arbitrary public HTTPS" criterion is **not** marked complete.

## Test plan

- [x] `zig build test --summary all` (3879/3889 passed, 10 pre-existing skips)
- [x] `zig build test -Dtls-profile=native --summary all` (3845/3855 passed)
- [x] `zig build test -Dtls-profile=appliance --summary all` (3845/3855 passed)
- [x] `zig build test-pki-differential --summary all`
- [x] `zig build test-integration-native-tls -Dtls-profile=native --summary all` (18/23 passed, 5 pre-existing platform skips) — includes the new end-to-end PKCS#1 upstream HTTPS test
- [x] `scripts/audit-dependencies.sh` (no forbidden dependencies, OpenSSL confined to adapter boundary, no `@cImport` in native paths)

References #634.


Jaruso (commented):
**Blocking review finding**

The RSA primitive/PKI verification work itself looks strong: the implementation reuses the existing strict RSA parser/modexp path, constructs the exact EMSA-PKCS1-v1_5 block and constant-time compares it, covers SHA-256/SHA-384 positive and negative vectors, preserves the RSA-PKCS1 exclusion from TLS 1.3 CertificateVerify, updates the WebPKI path tests/docs, and CI/dependency audits are green.

There is one merge-blocking protocol gap, though, and it is directly called out by #645's scope. RFC 8446 §4.2.3 separates `signature_algorithms` (CertificateVerify) from `signature_algorithms_cert` (certificate signatures), and says that when `signature_algorithms_cert` is absent, the `signature_algorithms` extension also applies to certificate signatures. Tardigrade currently has no `signature_algorithms_cert` extension at all (`ExtensionType` has no value 50 and repo-wide search finds no implementation). This PR correctly keeps `rsa_pkcs1` out of `signature_algorithms`, so the native client still signals to a conforming server that PKCS#1-signed certificate chains are unsupported even though the verifier can now authenticate them.

The OpenSSL end-to-end fixture does not catch that mismatch: TLS 1.3 permits a server that cannot construct a chain matching the client's advertised certificate-signature algorithms to continue with a chain containing unsupported algorithms, so a single-certificate `s_server` can send this PKCS#1 chain anyway. The test therefore proves the new verifier path works once such a chain arrives, but not that the client advertises the capability needed for ordinary WebPKI chain selection.

Please add the certificate-signature negotiation surface without widening CertificateVerify. A concrete shape I would expect is roughly:

```zig
pub const ExtensionType = enum(u16) {
    // ...
    signature_algorithms = 13,
    signature_algorithms_cert = 50,
    // ...
};

pub const SignatureScheme = enum(u16) {
    rsa_pkcs1_sha256 = 0x0401,
    ecdsa_secp256r1_sha256 = 0x0403,
    rsa_pkcs1_sha384 = 0x0501,
    rsa_pss_rsae_sha256 = 0x0804,
    ed25519 = 0x0807,
};
```

Then model a *separate* certificate-signature offer list (derived from what the WebPKI verifier/provider can actually verify) and emit it as ClientHello extension 50. Keep the existing `signature_schemes` / extension 13 list unchanged for CertificateVerify, so RSA PKCS#1 remains impossible there. The second ClientHello after HRR should preserve the certificate-signature offer as required by the normal ClientHello retry rules.

Please also add a wire-level regression that decodes/observes the generated native ClientHello and proves both sides of the boundary: extension 13 excludes `rsa_pkcs1_*`, while extension 50 includes at least `rsa_pkcs1_sha256` (and `rsa_pkcs1_sha384` now that this PR claims SHA-384 certificate support). That makes this invariant testable without relying on OpenSSL's allowed fallback behavior.

Once that signaling gap is fixed, I do not see another blocker in the current diff.

Jaruso (commented):
Re-review complete on 4bdaa80a. No remaining blocking findings.

The previous P1 is correctly fixed. `signature_algorithms_cert` (extension 50) is now a distinct ClientHello capability path, emitted in both ClientHello1 and the HRR ClientHello2. The certificate-signature list contains Ed25519, ECDSA-P256/SHA-256, RSA-PSS/SHA-256, and RSA PKCS#1 v1.5 SHA-256/SHA-384, while the ordinary `signature_algorithms` / CertificateVerify policy continues to reject both PKCS#1 schemes. The new wire-level regression verifies that separation directly for CH1 and CH2, so the earlier OpenSSL fallback hole in the E2E evidence is closed.

I also checked the shared-wire-enum side effect from adding `rsa_pkcs1_sha384 = 0x0501`: `validateNativePolicy` still rejects PKCS#1 from the CertificateVerify policy before handshake start, and `checkProofOfPossession` therefore cannot route an RSA-PKCS#1 CertificateVerify into the provider. The production native upstream path passes the same `CryptoProvider` to the TLS backend and `WebPkiVerifier`, so the advertised PKCS#1 certificate capability matches the verifier that will actually validate the chain.

The original #645 acceptance criteria remain satisfied: strict EMSA-PKCS1-v1_5 verification (required SHA-256 plus SHA-384), provider/PKI wiring, real native upstream HTTPS coverage through WebPkiVerifier, updated caveat docs, and no foreign crypto dependency. RFC 8446 §4.2.3 separation is now correctly implemented rather than merely documented. Current head CI and Darwin Release Smoke are green, including native/appliance tests, PKI differential, and the TLS dependency/linkage audit.

I resolved both prior review threads. From this review, PR #647 is clear to merge.

PR #648, opened 2026-08-24
packaging: keep Homebrew tap README tap-owned

Jaruso:
## Summary
- stop `scripts/update-homebrew-formula.sh --tap-dir` from rewriting `homebrew-tap/README.md`
- add deterministic regression coverage that proves a tap-owned README sentinel survives `--tag --tap-dir` synchronization while the generated formula is still copied
- update Homebrew packaging and release docs to make the ownership boundary explicit

## Validation
- `bash -n scripts/update-homebrew-formula.sh`
- `bash -n scripts/test-homebrew-tap-sync.sh`
- `shellcheck scripts/update-homebrew-formula.sh scripts/test-homebrew-tap-sync.sh`
- `find scripts/ packaging/ tests/vectors/pki -name "*.sh" -exec shellcheck {} +`
- existing Homebrew renderer validation from CI, including fixture canonical-output guard and `ruby -c` checks
- `./scripts/test-homebrew-tap-sync.sh`
- `git diff --check`

This resolves the remaining README overwrite finding from #466, but does not close #466 because the issue still requires release-to-tap automation and verification against the first qualifying native release.

Refs #466


Jaruso (commented):
The implementation itself is correct and stays within the intended #466 scope: `--tap-dir` now copies only the generated formula, fixture-mode protections are unchanged, the new regression test exercises the production `--tag --tap-dir` path with deterministic mocked release metadata, and the relevant Homebrew/lint jobs pass.

I found one remaining ownership/documentation inconsistency that should be fixed before merge: this PR declares `homebrew-tap/README.md` tap-owned but leaves `packaging/homebrew/tap-README.md` in Tardigrade. That file is now dead and contains an objectively stale statement that `scripts/update-homebrew-formula.sh` copies it into the tap, plus release-specific preparatory policy text. Please delete that file in this PR rather than leaving a second tap README/template behind.

CI note: the overall CI run is currently red only because the unchanged `Test native profile (macos-14)` integration lane timed out in `native upstream https: verified request to a trusted upstream succeeds`. The PR-specific `Lint` / renderer / `test-homebrew-tap-sync.sh`, Linux x86_64 + ARM64 Homebrew smoke jobs, and Darwin Release Smoke all passed. I do not see evidence that failure is caused by this patch, but I would rerun the failed lane before merging.

Once the stale `packaging/homebrew/tap-README.md` is removed and CI is green on rerun, I don't see another blocker for this narrow PR. This should remain `Refs #466`, not close #466.

Jaruso (commented):
Re-reviewed the full PR at `9756043cdaf148f5e7f78196e4dd114cd952eb7b` against #466 and the narrow tap-README ownership follow-up.

The previous finding is fully addressed: `packaging/homebrew/tap-README.md` is now removed, so Tardigrade no longer retains a stale second copy of tap policy/documentation. The resulting ownership model is consistent: Tardigrade owns deterministic formula generation/synchronization, while `Bare-Systems/homebrew-tap` owns its public README.

I also rechecked the complete diff, not just the follow-up commit:
- `--tap-dir` writes only `Formula/tardigrade.rb`;
- fixture mode still cannot use `--tap-dir` or overwrite the canonical formula;
- the regression test exercises the production `--tag --tap-dir` path with mocked release metadata and proves a pre-existing tap README remains unchanged;
- release asset/checksum/native-inventory validation is unchanged;
- packaging/release documentation matches the new ownership boundary;
- scope remains limited to this #466 follow-up.

No further code changes requested. The latest Lint job, including `test-homebrew-tap-sync.sh`, is green. Full CI and Darwin Release Smoke are still running on this head, so merge once required checks are green. This PR resolves the README-sync follow-up but does not close #466; release-to-tap automation and first qualifying release verification remain separate work.

PR #653, opened 2026-08-24
tls: raise native handshake engine's cert/message size bounds (#646)

Jaruso:
## Summary
- Raises `src/tls/tls13_backend.zig`'s `max_certificate_len` (2 KiB → 8 KiB per `CertificateEntry`) and `max_message_len` (8 KiB → 16 KiB per handshake message) so ordinary public WebPKI certificates/chains — a large-SAN leaf, or a multi-certificate chain — no longer get rejected by the engine's own resource bounds before the peer verifier ever runs. This was a second, independent blocker (alongside #645) on native-profile upstream HTTPS being production-ready against arbitrary public origins.
- `max_peer_chain_bytes` is now derived from `max_message_len` instead of carrying its own literal, and `transcript.max_pending_len` (which can't import `tls13_backend.zig` to reference the constant directly, to avoid a circular import) is kept in sync via a `comptime` assertion.
- Per-connection buffers sized to `max_message_len` grew accordingly (two handshake reassemblers, the HelloRetryRequest ClientHello1 capture, and the transcript's pre-selection buffer), so the "does not embed maximum ticket storage" backend-size budget tests were raised and, where the whole-struct comparison had stopped being a meaningful proxy for that claim, rewritten to assert the precise inline-vs-heap post-handshake reassembly boundary instead.
- Adds a large-SAN Ed25519 fixture (`native_ed25519_bigsan_ca.{crt,key}` + `native_ed25519_bigsan_chain.crt`: 61 SAN DNS names, ~3.3 KiB DER — past the old 2 KiB per-entry cap, within the new 8 KiB one, and within `pki/x509`'s separate, pre-existing 64-entry SAN-count parse limit) and a real end-to-end native upstream HTTPS integration test proving the fix against it.
- Updates `docs/TLS_DEPENDENCY_POLICY.md`, `docs/TROUBLESHOOTING.md`, `docs/CONFIGURATION.md`, and `docs/BARE_APPLIANCE_TLS.md` to drop the resolved size caveat and reflect the new bounds; #634's cutover checklist marks this item done (#645's signature-matrix gap remains separately open).

## Test plan
- [x] `zig build test` — 3871/3881 tests pass (10 pre-existing skips), all three `-Dtls-profile` builds compile.
- [x] `zig build test-tls` / `zig build test-quic` — full pass after adjusting size-boundary tests tuned to the old constants.
- [x] `zig build test-integration-native-tls -Dtls-profile=native` — new large-SAN upstream HTTPS test passes (18/23, 5 pre-existing skips unrelated to this change); also verified clean under `-Dtls-profile=appliance` and the default `general` profile (all native-only cases correctly skip).
- [x] `scripts/audit-dependencies.sh` passes.

Closes #646


Jaruso (commented):
Technical disposition: **APPROVE — no blocking findings.** GitHub does not permit this account to submit an `APPROVE` review on its own PR, so this is recorded as a comment review.

Reviewed all 13 changed files against #646's acceptance criteria and the surrounding #634/#645 cutover constraints.

Acceptance check:
- The native-client proof is real: the new leaf is 3,295-byte DER with 61 DNS SANs, so it is genuinely above the old 2 KiB `CertificateEntry` cap; it is signed by the new Ed25519 CA and its public key matches the existing server fixture key used by `startNativeUpstreamTardigrade`. The end-to-end test therefore exercises the intended native upstream verification path rather than a synthetic unit-only path.
- Resource protection remains fail-closed. `max_certificate_len` is still finite at 8 KiB, `max_message_len` at 16 KiB, `max_peer_chain_bytes` is correctly tied to the message ceiling, per-entry and aggregate writer preflights remain in place, and existing oversize/boundary tests were retuned to the new constants rather than removed.
- The shared-engine memory impact is accounted for: both main-handshake reassemblers, HRR/PSK capture, and transcript pre-selection buffering track the raised bound; the transcript duplicate is guarded by a comptime equality assertion; the backend/QUIC footprint guards were updated while retaining the post-handshake inline-vs-heap invariant.
- Documentation removes only the #646 size blocker and correctly leaves #645's certificate-signature matrix as the remaining native-upstream WebPKI caveat.

CI is still running overall, but the completed relevant lanes I checked are green so far, including format, security/lint, TLS interop, native macOS, and native ARM64 unit/listener integration coverage. I found no functional, security, or acceptance-criteria blocker. One non-blocking stale-comment nit is inline.

Issue #629, opened 2026-08-16
tls: prevent TCP/H3 credential split on mixed OpenSSL builds while preserving native reload semantics


Jaruso:
## Goal

Prevent a running Tardigrade process from serving different certificate identities on TCP H1/H2 and QUIC/H3 after reload while the repository still contains the transitional mixed OpenSSL/native composition.

This issue is now explicitly **transitional compatibility work under #634**, not a long-term architecture investment.

---

# Architecture reconciliation — 2026-08-20

#634 establishes the final project architecture:

> **Every deployed `tardi` artifact uses Tardigrade's native Zig TLS/crypto/PKI/QUIC/H3 implementation. OpenSSL is retained only as independent test/interop/benchmark infrastructure, not as a shipping production backend.**

That changes how #629 should be solved.

The long-term fix is to eliminate the mixed production composition entirely. Do **not** build substantial new OpenSSL reload/lifecycle machinery merely to keep the transitional backend feature-equivalent with native Tardigrade.

If #634 removes the OpenSSL-backed production path before #629 receives an independent fix, and the certificate-split state is therefore no longer reachable in any shipping composition, close #629 as **not planned/superseded by #634** after proving the native shipping path is coherent.

If there is still a real shipping window in which a mixed OpenSSL TCP + native H3 artifact remains published, implement only the minimum safety behavior required to prevent divergent live identities until cutover.

---

# Current transitional defect

Today the mixed build can have two credential owners with different reload behavior:

```text
TCP H1/H2 -> OpenSSL-owned certificate context
QUIC/H3   -> NativeCredentialStore
```

A reload changing certificate/key/SNI identity can allow native H3 to adopt identity B while OpenSSL/TCP remains on identity A until restart.

That split is invalid even for the transitional artifact.

---

# Preferred resolution order

## Preferred: remove the mixed shipping composition via #634

If #634 lands first:

1. shipping TCP H1/H2 uses the native Zig credential/TLS path;
2. H3 continues using the native Zig credential path;
3. prove both protocol families select the same configured identity at startup/reload according to the native contract;
4. prove the old OpenSSL adapter is not present in shipping artifacts;
5. close #629 without adding new OpenSSL lifecycle architecture.

This is the expected long-term outcome.

## Temporary fallback only if mixed builds still ship

If the repository must continue publishing a mixed artifact before #634 is complete, enforce this safety invariant:

```text
before reload:
TCP = identity A
H3  = identity A

config changes identity to B
[...]

Issue #649, opened 2026-08-24
build: retire the OpenSSL production backend and make native Tardigrade the sole default implementation


Jaruso:
## Goal

Complete the central production-architecture cutover from #634: make the Zig-native implementation the **only supported production implementation** selected by normal Tardigrade builds, and retire the transitional `general = openssl-adapter` production backend.

This issue is intentionally about **build/runtime composition and production-backend retirement**. It is not the place to finish public-WebPKI breadth (#645/#646), Homebrew (#466), Docker/package distribution cleanup (separate #634 child), or final published-release proof (separate #634 child).

---

## Context

#634 is an umbrella epic. Current `main` already has:

- a general-purpose native production-shaped profile (`-Dtls-profile=native`, #641);
- native downstream H1/H2 TLS;
- native QUIC/H3;
- native generic credential/SNI handling;
- native upstream HTTPS/TLS client and default-identity SAN/SNI fallback (#643);
- official release archives already built/audited as native (#476).

The remaining public-WebPKI blockers are tracked independently:

- #645 — common certificate-chain signature algorithms, especially RSA PKCS#1 v1.5;
- #646 — realistic certificate/handshake-message bounds.

Once #645 and #646 land, there is no architectural reason to keep OpenSSL as a supported production backend merely to provide the normal general-purpose runtime.

Current transitional composition still includes:

```text
TlsProfile = enum { general, appliance, native }
```

with default:

```text
tls-profile=general
```

and `general` linking the OpenSSL adapter through `configureSsl()`.

`tardi version` still recognizes the OpenSSL-backed production identity, and production-oriented source/build surfaces still contain the adapter boundary (`src/http/tls_termination.zig`, `src/http/acme_client.zig`, related build wiring).

That state violates #634's final architecture even though release.yml explicitly selects `native` today.

---

## Dependency / start rule

This issue should be implemented **after #645 and #646 are complete**, so the generic native client can honestly replace the old OpenSSL-backed general-purpose production path for ordinary public HTTPS origins.

Do not hide unfinished #645/#646 behavior by keeping an OpenSSL production fallback.

---

# Required work

## 1. Freeze the final operator feature disposition

Before deleting the OpenSSL production backend, enumerate every operator-visible feature/config knob that historically depended on it and give it exactly one final disposition:

```text
native + supported
explicitly unsupported with deterministic config failure
non-production test/interop only
```

[...]

PR #654, opened 2026-08-24
build: retire the OpenSSL production TLS backend and make native the sole default

Jaruso:
## Summary

Completes the central build/source cutover #634 called for: **every supported production Tardigrade build is now native-only**. Depends on #645/#646 (both merged), which closed the public-WebPKI gaps that let the native implementation honestly replace the OpenSSL general-purpose production path.

- Collapsed `TlsProfile` to `{ general, appliance }` — both pure-Zig native. `general` (the default, no flag needed) now covers what was previously the separate `-Dtls-profile=native` value; `-Dtls-profile=general` no longer means "link OpenSSL." `-Dtls-profile=native` is no longer a valid flag.
- Removed `configureSsl` and every production OpenSSL linkage path from `build.zig`. The `evp_oracle`/`crypto_openssl_diff` differential-test tooling now builds unconditionally, decoupled from TLS profile — it was always test/interop-only and never part of the production module graph.
- Deleted the OpenSSL-backed `src/http/tls_termination.zig` and `src/http/acme_client.zig` implementations and the `tls_backend.zig`/`acme_backend.zig` selector layer that used to switch between them and their native counterparts. The native implementations (previously the `_stub.zig` twins) are promoted into their place — the final architecture, not a fallback. A later commit on this PR removed the resulting inert downstream `TlsTerminator`/`TlsConnection` scaffolding entirely and renamed the file to `src/http/upstream_tls.zig`, since it now holds only the native upstream HTTPS/TLS client.
- `edge_gateway.zig`/`edge_config.zig`/`gateway_shutdown.zig`: removed the now-always-true/always-false branches this retirement leaves behind, including the #629 mixed OpenSSL-TCP/native-HTTP3 credential-owner special case, which can no longer occur — there is exactly one credential owner per profile now, shared by TCP and HTTP/3 alike.
- `tardi version` always reports `tls-backend=native`; the `tls-backend=openssl-adapter` string is gone.
- Every former OpenSSL-only operator capability (TLS 1.2, cipher overrides, downstream mTLS, session cache/tickets, OCSP, CRL, ACME, the credential watcher, PROXY+TLS) fails config validation deterministically in *every* profile now (`validateNativeTlsBuildConfig` is unconditional), not just the previously-adapter-free ones. Full disposition table in `docs/TLS_DEPENDENCY_POLICY.md`.
- CI: both profiles are built/audited (no more three-profile matrix); dropped now-unnecessary `libssl-dev` installs from production-only build jobs (`build`, `examples`, `packaging-smoke`, `perf-smoke`, `integration-test`) so they double as proof the default build needs no OpenSSL headers at all; the retired `-Dtls-profile=native` job's unique `test-integration-native-tls` coverage is preserved under the general profile.
- Rewrote `docs/TLS_DEPENDENCY_POLICY.md` for the final architecture, and corrected stale "OpenSSL still links in production" claims across README, CONFIGURATION, TROUBLESHOOTING, RELOAD_SHUTDOWN, DEPLOYMENT, UPSTREAM_POOLING, CRYPTO_PROVIDER, EVENT_LOOP_BACKENDS, OBSERVABILITY, TIMEOUTS, BARE_APPLIANCE_TLS, QUICKSTART, and the packaging READMEs.

**Non-goals** (per the issue): Docker/package distribution internals, Homebrew (#466), and final published-release proof remain separate #634 children and are untouched here (only literal command-line prose that would otherwise fail was fixed).

**Review follow-up** (addressed on this PR, not spun off): `src/crypto/profile.zig`'s `ProductProfile` is now exactly `enum { general, appliance }` (was `general_purpose_openssl`/`native_appliance`); the downstream `TlsTerminator`/`TlsConnection` model and `WorkerContext.tls` are removed entirely rather than kept as inert scaffolding; `src/http/tls_termination.zig` is renamed to `src/http/upstream_tls.zig` (with `NegotiatedProtocol` relocated to the neutral `negotiated_dispatch.zig` and its dead OpenSSL-only error variants dropped); the QUIC header-protection rows in the crypto capability matrix now correctly claim both product profiles, since `general` drives the same in-process QUIC/H3 stack as `appliance` since this PR; and the ACME/config-validation diagnostics no longer describe retired settings as if an OpenSSL fallback still exists.

## Test plan

- [x] `zig build` (default, no flags) succeeds; `otool -L zig-out/bin/tardi` shows only `libSystem.B.dylib`
- [x] `zig build -Dtls-profile=appliance` succeeds; same linkage check
- [x] `./scripts/audit-dependencies.sh` passes
- [x] `./scripts/audit-release-binary.sh --binary zig-out/bin/tardi --profile general` and `--profile appliance` both pass with `links_openssl: false`
- [x] `zig build test` passes in full (general/default profile), including the now-unconditional `test-crypto-openssl` differential step
- [x] `zig build -Dtls-profile=appliance test` passes in full
- [x] `zig build test-integration-native-tls` passes (both profiles)
- [x] `zig build test-integration-resumption-interop` passes (43/49, 6 skipped — missing local `gtlsclient`/ngtcp2 binary in this sandbox, unrelated to this change)
- [x] `tardi version` reports `tls-profile=general, tls-backend=native` and `tls-profile=appliance, tls-backend=native`
- [ ] Full `zig build test-integration` run (in progress locally at time of PR open; CI will also run it on merge to main)


Jaruso (commented):
**Review result: changes required before merge.** GitHub does not permit the PR author to submit a formal `REQUEST_CHANGES` review on their own PR, so this is posted as a comment review, but the two P1 findings below are merge-blocking against #649's acceptance criteria.

I reviewed the full #649 cutover against the issue's acceptance criteria, including the build graph, runtime ownership/reload paths, native upstream HTTPS client, integration-test rewiring, dependency/binary audits, release workflows, and docs.

The production behavior is largely in the right shape: default/general and appliance both select the native implementation, `configureSsl` and the shipping OpenSSL linkage are gone, retired operator features fail closed, native credentials are shared across TCP/H3 in the general profile, upstream HTTPS remains a real native client, and CI is structurally wired to exercise the final native configurations on Linux x86_64, Linux ARM64, and macOS.

The remaining blockers are architectural-completion gaps. #649's close rule is stronger than "the old backend is unreachable": it explicitly requires transitional adapter/stub/module naming and indirection to be removed/reconciled, and all runtime/profile vocabulary to reflect final native ownership.

1. **Retired downstream terminator plumbing is still embedded in the runtime model.** `edge_gateway.zig` declares `tls_terminator` but never assigns it, keeps the #629 mixed-owner branch and maintenance/deinit plumbing, passes `.tls` into `WorkerContext`, `gateway_shutdown.zig` still updates it, `DownstreamTransport` still has `OpenSslTransport/.openssl`, and the promoted `tls_termination.zig` carries an inert `TlsTerminator`/`TlsConnection` plus an `.openssl` encrypted-stream backend solely to keep those dead branches compiling. That is exactly the transitional scaffolding #649 says to retire, and it contradicts the PR summary's statement that the #629 special case was removed.

   The end state should remove the dead downstream terminator model rather than keep a permanent stub. For example:
   ```zig
   pub const WorkerContext = struct {
       config_store: *ReloadableConfigStore,
       state: *GatewayState,
       native_credentials: ?*http.native_tls_connection.NativeCredentialStore,
       native_tls_provider: ?tls_core.credentials.CredentialProvider,
       // ... no `tls` field
   };

   pub const DownstreamTransport = union(enum) {
       plaintext: std.posix.fd_t,
       native: *native_tls_connection.NativeTlsConnection,
   };
   ```
   Then remove `tls_terminator` from `edge_gateway.run`, the mixed-owner block, `runMaintenance`/deinit calls, the `worker_ctx.tls` reload branch, and the inert downstream `TlsTerminator`/`TlsConnection` compatibility API. The real native `UpstreamTlsConn` can remain as the upstream HTTPS surface.

2. **The source product-profile model still encodes the retired OpenSSL product.** `src/crypto/profile.zig` still defines `.general_purpose_openssl`, and the updated documentation explicitly calls that enum/its rows a pre-#649 holdover that has not been updated. The issue's close rule specifically requires *all runtime/profile vocabulary* to describe the final native architecture. The product capability matrix is source-level policy metadata and its own docs call it the source of truth, so documenting the inconsistency does not satisfy the cutover.

   Please update the source model itself, e.g.:
   ```zig
   pub const ProductProfile = enum {
       general,
       appliance,
   };
   ```
   and update each `enabled_product_profiles` row and the `src/tls/crypto_profile.zig` tests to express actual native general/appliance policy. If OpenSSL capability metadata is still useful for differential/interop testing, keep that as explicitly non-production oracle/reference metadata rather than a product profile selectable by the source capability model.

There is also one small stale ACME comment called out inline: the promoted stub is unavailable in every shipping profile now, not just appliance.

I did not find a separate native TLS functional regression in the changed H1/H2/H3/upstream paths after tracing the new ownership and test rewiring. I also verified the workflow definitions contain the required platform/profile coverage and audits, but GitHub's combined-status endpoint exposed no check results for the head SHA, so I am not representing the current Actions run state as green.

src/http.zig
```diff
@@ -54,16 +54,13 @@ pub const negotiated_dispatch = @import("http/negotiated_dispatch.zig");
...
-pub const tls_termination = @import("http/tls_backend.zig");
-/// ACME client backend selected by `-Dtls-profile` (#379): OpenSSL-backed in
-/// the general profile, a no-OpenSSL stub in the appliance profile. Selected
-/// at the module graph so the appliance never links OpenSSL through ACME.
-pub const acme_client = @import("http/acme_backend.zig");
+/// Native TLS termination surface, identical across every `-Dtls-profile`
+/// (#379, #649): no `@cImport("openssl/...")` and no foreign TLS/crypto
+/// linkage in any shipping build.
+pub const tls_termination = @import("http/tls_termination.zig");
+/// Native ACME client surface, identical across every `-Dtls-profile`
```
Jaruso: [P1] After the follow-up cleanup in `234275a9`, this is no longer a “TLS termination surface”: `http/tls_termination.zig` contains only the native **upstream HTTPS client**. #649 explicitly requires transitional module naming/indirection to be reconciled so names describe the final architecture. Please rename/rehome this as e.g. `upstream_tls.zig`/`http.upstream_tls`, update `upstream_pool.zig`, `upstream_h2.zig`, `gateway_proxy.zig`, etc., and move `NegotiatedProtocol` into a neutral module (probably `negotiated_dispatch.zig`) so downstream protocol dispatch does not import an upstream-client module merely for an enum. While touching the final surface, also remove unreachable legacy error tags from its `TlsError` (`OpenSslInitFailed`, `CipherConfigFailed`, `CrlLoadFailed`, `OcspLoadFailed`, etc.); those names are leftover adapter vocabulary in a native-only API.

src/crypto/profile.zig
```diff
@@ -142,29 +148,29 @@ pub const Row = struct {
...
+    row(.{ .aead = .chacha20_poly1305 }, "ChaCha20-Poly1305", .zig_std_crypto, .supported, .openssl_provider, .provider_deferred, "seal/open round-trip, tamper rejection, AD mismatch", .{ .{ .tls_record, .live }, .{ .quic_packet_protection, .live } }, .{ .appliance, .general }, "The native appliance's engine now negotiates TLS_CHACHA20_POLY1305_SHA256 (#564), through the same generic, CryptoProvider-routed record-protection path as the other two suites. QUIC packet protection now selects ChaCha20-Poly1305 from explicit negotiated-suite metadata (#566)."),
+    row(.{ .quic_header_protection = .aes_128 }, "QUIC AES-128 header protection", .zig_std_crypto, .supported, .openssl_provider, .provider_deferred, "RFC 9001 header-protection sample and provider mask parity", .{.{ .quic_packet_protection, .live }}, .{.appliance}, "src/quic/tls_adapter.zig applies and removes header protection through CryptoProvider on every send/receive path (#490); the direct Aes128.initEnc form remains only as differential test-vector fixtures. QUIC-specific: the general-purpose backend is TLS-over-TCP, not QUIC."),
```
Jaruso: [P1] This still encodes the retired `general = OpenSSL/TCP-only` product model. After #649, `.general` is native and this PR's own general-profile H3/resumption/0-RTT coverage drives `src/quic/tls_adapter.zig`, so QUIC header protection is a live capability of `.general` too. The AES-256 and ChaCha rows immediately below repeat the same mismatch, and the test near the bottom explicitly pins `!header_protection_profiles.contains(.general)`. Please make all three rows `. { .appliance, .general }` (normal Zig spacing), remove the “general-purpose backend is TLS-over-TCP, not QUIC” claim, flip the assertion, and update `docs/CRYPTO_PROVIDER.md`. Otherwise the source-of-truth product matrix is still describing the architecture #649 is deleting.

src/edge_config.zig
```diff
@@ -2992,17 +2981,15 @@ fn validateTlsServerNamePolicy(cfg: *const EdgeConfig) !void {
...
-/// `validateApplianceTlsProfile`/`validateTlsServerNamePolicy` instead.
+/// #634/#649: capability gate shared by every TLS profile — `general` and
+/// `appliance` are both native-only builds. These are implementation
+/// limits of the native TCP TLS path — settings only the retired OpenSSL
+/// adapter implemented must fail deterministically at validation time
+/// rather than being silently ignored. Appliance-only *product policy*
+/// (single identity, required server name, HTTP/3 feature restrictions)
+/// lives in `validateApplianceTlsProfile`/`validateTlsServerNamePolicy`
+/// instead.
 fn validateNativeTlsBuildConfig(cfg: *const EdgeConfig) !void {
```
Jaruso: [P2] Making this capability gate unconditional is correct, but the ACME rejection immediately below still says `the ACME client is OpenSSL-adapter-only until its native implementation lands`. That is now an operator-facing reference to a production option #649 removes: there is no adapter profile to choose. Keep the deterministic failure, but describe the actual disposition, e.g. `ACME automation is not implemented by native Tardigrade; provision certificate/key material externally (no production fallback)`. I would also rename the nearby `native-TLS builds reject OpenSSL-adapter-only TLS settings...` test to `retired/unsupported legacy TLS settings` and clean the `EdgeConfig` comments that still speak about a current “general OpenSSL terminator”/“OpenSSL-facing” settings boundary.

Jaruso (commented):
Changes are still required before merge.

The three blockers from my previous review were fixed and I resolved those threads. I re-reviewed the resulting architecture against #649 rather than only checking the replies, and found three remaining acceptance-criteria gaps:

1. **[P1] Final module ownership/name is still transitional.** After removing `TlsTerminator`/`TlsConnection`, `http/tls_termination.zig` is exclusively an upstream TLS client, but the canonical export/module is still named `tls_termination`, and downstream `negotiated_dispatch.zig` imports that upstream-client module just to get `NegotiatedProtocol`. #649 explicitly requires module names/indirection to describe the final architecture rather than migration history. Rehome it as something like `upstream_tls.zig`, put the shared negotiated-protocol enum in a neutral module, and update the imports/exports. While doing that, shrink the remaining native client error surface so dead `OpenSslInitFailed`/CRL/OCSP/cipher-adapter errors are not part of the final native API.

2. **[P1] `ProductProfile.general` still carries one semantic assumption from the retired OpenSSL product.** The QUIC header-protection rows claim `.appliance` only and literally say the general backend is TLS-over-TCP, while this PR makes `general` native and runs H3/0-RTT through that same QUIC packet-protection implementation. Fix all three header-protection rows (AES-128/AES-256/ChaCha), the assertion below, and the generated/docs matrix so `.general` reflects the actual live product.

3. **[P2] One operator-facing unsupported-feature diagnostic still points at the retired adapter.** ACME validation says the client is “OpenSSL-adapter-only until its native implementation lands.” After #649 there is no supported production adapter to choose. The deterministic rejection is correct; the message needs to say ACME automation is currently unsupported by native Tardigrade/no production fallback exists. Please sweep the adjacent test/comment vocabulary as well where it still describes current settings as “OpenSSL-adapter-only” instead of retired/legacy settings.

The runtime cleanup added in `234275a9` otherwise looks correct: native TCP TLS owns its connections through `ActiveRegistry/ManagedConnection`, so removing the OpenSSL-only parked TLS state does not drop native keepalive behavior; new native handshakes take a current config lease and derive listener protocol policy from it, so deleting `TlsTerminator.updateProtocolPolicy()` does not lose reload behavior; one native credential owner is shared by TCP and H3 in each profile; and I did not find a new H1/H2/H3 or upstream-HTTPS data-path regression.

CI also is not green on the current head: Actions run `32718135441` has overall conclusion `failure`. Every one of the 30 jobs exposed by the connector's first page is green (including general/appliance unit suites, appliance full integration on both Linux architectures, general native-listener tests on x86_64/ARM64, TLS dependency audit, TLS interop, and H3 resumption/0-RTT), so the failed job is outside that first page. Please get the current head to a green required run or identify/resolve the remaining failed job before merge.

Finally, update the PR description before merging: its “Follow-up spun off” paragraph still says `ProductProfile.general_purpose_openssl` was intentionally left for later, but this PR now fixes it in `3b95b4c3`.

src/http.zig
Jaruso: Fixed in 5202263f: renamed to `src/http/upstream_tls.zig`. Moved `NegotiatedProtocol` into the neutral `negotiated_dispatch.zig` (both this upstream client and the downstream native listener need it, so neither should be a dependency of the other just for the enum) and updated every import site (`http.zig`'s export, `edge_gateway.zig`, `gateway_protocol_policy.zig`, `gateway_proxy.zig`, `gateway_control_plane_proxy.zig`, `upstream_h2.zig`, `upstream_pool.zig`, `webpki_verifier.zig` comments). Also shrank `TlsError` to the variants this client actually returns, dropping `OpenSslInitFailed`/`CertificateKeyMismatch`/`ProtocolConfigFailed`/`CipherConfigFailed`/`CrlLoadFailed`/`OcspLoadFailed` (none were reachable — confirmed via grep, nothing outside the definition referenced them).

src/crypto/profile.zig
Jaruso: Fixed in 5202263f: the three QUIC header-protection rows (AES-128/AES-256/ChaCha20) now claim `.{ .appliance, .general }` instead of `.{.appliance}`. Confirmed `src/quic/tls_adapter.zig` has no profile branching and CI's general-profile `test-integration-resumption-interop` job already exercises this exact header-protection path via H3/0-RTT, so `.general` genuinely drives it, not just `.appliance`. Updated the corresponding assertion in `profile.zig`'s test and the `docs/CRYPTO_PROVIDER.md` matrix row.

src/edge_config.zig
Jaruso: Fixed in 5202263f: the ACME diagnostic now reads "ACME issuance/renewal is not implemented by native Tardigrade and #649 retired the only production build that ever supported it" instead of implying a still-supported OpenSSL fallback. Also renamed the adjacent test from "...reject OpenSSL-adapter-only TLS settings..." to "...reject legacy OpenSSL-only TLS settings..." (and its one comment reference in tests/integration.zig) — swept the rest of the codebase for the same pattern and found nothing else describing a currently-supported setting as OpenSSL-adapter-only.

Jaruso:
Addressed all three findings from the second review round in 5202263f:

1. **[P1] Final module ownership/name.** `src/http/tls_termination.zig` → `src/http/upstream_tls.zig`. `NegotiatedProtocol` moved to the neutral `negotiated_dispatch.zig` (both the upstream client and downstream native listener depend on it now, symmetrically) instead of downstream code importing an upstream-only module just for the enum. `TlsError` shrunk to the 9 variants this client actually returns — dropped the 6 dead OpenSSL-adapter-only ones.
2. **[P1] `ProductProfile.general`'s stale QUIC-header-protection claim.** The three header-protection rows now claim both `.general` and `.appliance` — `general` is native since #649 and runs the same in-process QUIC/H3 stack, which CI's general-profile H3 resumption/0-RTT job already exercises. Fixed the row review text, the pinned test assertion, and the docs matrix.
3. **[P2] Stale ACME/adapter-only vocabulary.** The ACME-disabled diagnostic no longer implies a supported OpenSSL fallback exists. Renamed the adjacent "OpenSSL-adapter-only TLS settings" test to "legacy OpenSSL-only" and swept the rest of the tree for the same pattern — nothing else needed it.

Also fixed a stale `docs/CRYPTO_PROVIDER.md` paragraph left over from before 234275a9 (still described the since-deleted inert `TlsTerminator`/`TlsConnection` types), and updated the PR description's "Follow-up spun off" paragraph, which was inaccurate after this round.

`zig build test` passes in full on both profiles. Pushing now — the prior CI run's only failure (`Test general profile native TLS listener (macos-14)`, a `ReadTimeout` in the same upstream-pool test category that was flaky on a different platform last run, with a clean local re-run confirming it) should get a fresh shot on this push.

docs/UPSTREAM_POOLING.md
```diff
@@ -40,7 +40,8 @@ In scope (Phases 1 / 1b / 1c / 2 / 3):
   keep-alive still spares the origin per-request accept/fd churn and brings
   unix upstreams under the same idle/lifetime/active-cap policy and metrics
   as TCP.
-- For TLS the pooled entry owns the OpenSSL connection, so the handshake is
+- For TLS the pooled entry owns the native upstream `UpstreamTlsConn`
+  (`src/http/tls_termination.zig`, #634/#649), so the handshake is
```
Jaruso: [P2] The prose now correctly says this pooled entry owns the **native** `UpstreamTlsConn`, but the path immediately contradicts the rename completed in `5202263f`: `src/http/tls_termination.zig` no longer exists as the final module; this should be `src/http/upstream_tls.zig`. #649 explicitly makes final module naming part of the close rule, so please sweep docs for the retired path after the rename rather than leaving a dangling migration-era reference.

src/http/upstream_pool.zig
```diff
@@ -17,7 +17,7 @@
 
 const std = @import("std");
 const compat = @import("zig_compat");
-const tls_termination = @import("tls_backend.zig");
+const upstream_tls = @import("upstream_tls.zig");
```
Jaruso: [P2] The import/type rename is correct, but the final source-vocabulary sweep is still incomplete around it. This file's module/PooledConn comments still say TLS entries own “the OpenSSL connection”; `upstream_h2.zig` still explains the live native transport's `pending()` as `SSL_pending`/bytes buffered in OpenSSL; `edge_config.zig` still says `tls_server_name` is unused by “the general OpenSSL terminator” and describes native resumption relative to “the OpenSSL terminator's own session config”; `edge_gateway.zig` still documents/tests impossible “OpenSSL-only TCP” shipping compositions; and `negotiated_dispatch.zig` still calls its generic test fixture `TestOpenSslConn`. These are not useful historical notes—they describe current fields/helpers in present tense. #649's close rule says production runtime/profile vocabulary must reflect native ownership. Please do one final mechanical sweep, e.g. call the pool entry a native `UpstreamTlsConn`, describe `pending()` as already-decrypted native TLS plaintext, describe the old `tls_session_*` knobs as retired/unsupported legacy knobs, and rename generic test fixtures to transport-neutral names. Historical “used to be OpenSSL” comments can stay where clearly past tense.

Jaruso (commented):
Third-pass review result: **one bounded cleanup remains, and CI is not yet merge-green.**

The three findings from my previous review are substantively fixed in `5202263f`: `tls_termination.zig` is now `upstream_tls.zig`, `NegotiatedProtocol` has a neutral home and dead upstream error tags are gone; the QUIC header-protection product rows now correctly include both `.general` and `.appliance`; and the ACME rejection no longer suggests a selectable OpenSSL production fallback. I verified those changes and resolved the previous threads.

I did not find a new H1/H2/H3, credential ownership, reload, resumption, keepalive, or upstream-HTTPS data-path bug introduced by the follow-up commit. The module/type rewiring itself looks coherent.

There is still one #649 close-rule cleanup to finish: present-tense source/docs vocabulary continues to describe **live native objects or currently evaluated configurations** as OpenSSL-owned. I left two inline comments with concrete examples. This should be a mechanical sweep, not another architecture change. Historical notes such as “this used to be implemented by OpenSSL” are fine when clearly past tense; the problem is statements such as a live `PooledConn.tls` owning “the OpenSSL connection,” a native transport's `pending()` being described as `SSL_pending`, `EdgeConfig` saying a current field is unused by “the general OpenSSL terminator,” and `edge_gateway` tests/comments describing an `OpenSSL-only TCP` production composition that no longer exists. Suggested final terminology is along the lines of:

```zig
/// Appliance-only fixed server-name policy; the general native credential
/// store derives SNI eligibility from its configured/default identities.
tls_server_name: []const u8,

/// Legacy `tls_session_*` knobs are retained only for deterministic rejection;
/// supported native resumption is configured through `tls_native_*`.
tls_native_resumption_mode: []const u8,
```

Also, the current required CI run `32731119003` is not green: **`Test general profile native TLS listener (macos-14)` failed** in `integration.test.native upstream https: two proxied requests reuse the pooled TLS connection` with `error.ReadTimeout` while reading the proxied response. Linux x86_64 and Linux ARM64 general-profile listener jobs are green, appliance Linux/ARM/macOS jobs are green, the TLS dependency audit is green, and Darwin Release Smoke is green. This failure pattern looks plausibly timing-sensitive rather than caused by the rename/metadata-only follow-up commit, but #649 explicitly requires macOS native coverage, so it still needs a clean rerun or a demonstrated fix before merge.

Once the vocabulary sweep is done and that macOS job is green, I do not currently see another #649 acceptance-criteria blocker.

Jaruso:
CI update: the previously-flagged run (\`32718135441\`, overall \`failure\`) and the follow-up push both failed the same single job — \`Test general profile native TLS listener (macos-14)\` — 3 times in a row, always the same 1-2 tests (\`verification disabled succeeds against the same untrusted fixture\` / \`two proxied requests reuse the pooled TLS connection\`), always \`error.ReadTimeout\`, never on Linux runners, never locally.

Root cause: those two are the only tests in that group that need a full *successful* native TLS handshake through two stacked listeners (this test's own proxying \`tardigrade\` plus a second full \`startNativeUpstreamTardigrade\` process) to actually complete within the file's usual 10s budget — sibling tests using the same origin setup but expecting a fast-fail rejected handshake passed reliably every time. Widened both to 20s in \`74bf279b\`, matching this file's existing precedent for its heaviest upstream-https test. Local `zig build test-integration-native-tls` still passes 19/24 (5 skipped, 0 failed).

Fresh CI run in progress on \`74bf279b\`.

Jaruso:
CI is fully green on `74bf279b`. The macOS timeout widening held (that job passed cleanly). One new job — `Test general profile native TLS listener (ubuntu-latest)`, which had passed cleanly on both prior runs — hit 3 simultaneous `ReadTimeout` failures on tests I hadn't touched; re-running just that job passed clean, consistent with one-off runner contention rather than a code or test-timing issue. All review threads from both rounds are addressed and replied to; PR description is up to date.

Jaruso (commented):
Re-reviewed current head `74bf279b` against #649 and the prior review findings.

The new commit is limited to widening two integration-test client read deadlines from 10s to 20s. I do not see a new product/runtime regression from that change: it does not alter Tardigrade's production upstream/TLS timeout configuration or handshake logic, only the outer test harness deadline. The previously failing macOS general-profile native-TLS coverage is now green, as are the overall CI run and Darwin Release Smoke.

No new functional H1/H2/H3, credential/SNI, reload/resumption, keepalive/pooling, dependency-boundary, or upstream-HTTPS defect found in this pass.

Changes are still required before merge only because the two P2 threads from review `5008271181` remain untouched on this head:

1. `docs/UPSTREAM_POOLING.md` still points the native `UpstreamTlsConn` at the retired `src/http/tls_termination.zig` path instead of `src/http/upstream_tls.zig`.
2. The final source-vocabulary sweep is still incomplete: `upstream_pool.zig` describes live native pooled TLS state as an OpenSSL connection, `upstream_h2.zig` describes `pending()` in terms of `SSL_pending`/OpenSSL buffering, `edge_config.zig` still describes current fields relative to a "general OpenSSL terminator", `edge_gateway.zig` still documents/tests impossible "OpenSSL-only" shipping compositions, and `negotiated_dispatch.zig` still names its generic fixture `TestOpenSslConn`.

Those are the only remaining #649 close-rule gaps I can substantiate. Once those two existing threads are fixed, I currently see no remaining blocker to merging #654.

Jaruso (commented):
I fixed the two remaining #649 review findings directly on the branch and resolved both threads.

The final cleanup is intentionally mechanical only:
- `docs/UPSTREAM_POOLING.md` now points at `src/http/upstream_tls.zig`.
- `upstream_pool.zig` describes the native `UpstreamTlsConn`, not an OpenSSL connection.
- `upstream_h2.zig` describes `pending()` as already-decrypted transport plaintext rather than `SSL_pending`/OpenSSL buffering.
- `edge_config.zig` describes the old `tls_session_*` knobs as retired/unsupported legacy fields and no longer references a current general OpenSSL terminator.
- `edge_gateway.zig` no longer describes impossible OpenSSL-only production compositions or a current OpenSSL terminator.
- `negotiated_dispatch.zig` uses transport-neutral `TestGenericConn`/`generic_conn` vocabulary.

I verified the temporary one-shot helper is absent from the final tree; compared with the previously green functional head `74bf279b`, the net source change is exactly those six docs/comment/test-vocabulary files. No runtime behavior was changed.

Because the cleanup commit was produced by GitHub Actions (whose token suppresses downstream workflow triggers), I added an empty validation commit on the same tree so the normal PR gates run against the exact final content. CI 1631 and Darwin Release Smoke 48 are currently pending. There are no remaining review findings from my side; merge is gated only on those final checks completing green.

PR #656, opened 2026-08-24
packaging: cut Docker and standalone DEB/RPM tooling over to native-only

### LINKED ISSUE #650: packaging: cut Docker and standalone DEB/RPM tooling over to native-only Tardigrade
## Goal

Complete the remaining non-Homebrew distribution cutover required by #634 so every production-facing local/package/container path ships the Zig-native Tardigrade implementation and carries **no OpenSSL/libcrypto dependency for product behavior**.

This issue owns:

- the root Docker image path;
- standalone DEB packaging tooling;
- standalone RPM packaging tooling;
- production-facing package/container smoke and dependency checks.

Homebrew is **not duplicated here**: #466 remains the canonical owner for the public `Bare-Systems/homebrew-tap` formula and release-to-tap automation.

---

## Context

The official GitHub release pipeline already builds Linux/Darwin archives with the native profile and packages audited native binaries into DEB/RPM artifacts (#476). That means the **official release package path** is already native-shaped.

However current local/standalone production surfaces remain transitional:

### Docker

The root `Dockerfile` currently:

- installs `libssl-dev` in the builder;
- runs default `zig build -Doptimize=ReleaseFast`;
- installs `libssl3` in the runtime image;
- documents the historical OpenSSL-backed `general` profile as the default Docker composition.

### Standalone DEB/RPM builders

The standalone package builders still accept/infer:

```text
--tls-backend native
--tls-backend openssl-adapter
```

and can emit OpenSSL runtime dependencies for `openssl-adapter` artifacts.

That compatibility mode is transitional and must disappear once #649 retires the OpenSSL production backend.

---

## Dependency / start rule

Implement after, or in tight coordination with:

- #645 and #646 — native upstream public-WebPKI closeout;
- #649 — native becomes the sole supported production implementation.

Do not preserve a package/container OpenSSL escape hatch after #649 removes it from the product architecture.

#466 remains independent but complementary: #634 requires both this issue and #466 before all advertised distribution paths are native-only.

---

# Required work

## 1. Make the Docker build native-only

The production-facing Docker build must compile the same native implementation the project supports elsewhere.

After #649, prefer the final normal build command rather than pinning a now-obsolete migration flag. If the final build still requires an explicit product-policy selector, choose the general-purpose native policy intentionally.

The Docker builder must not install OpenSSL development packages merely to build Tardigrade.

Remove:

```text
libssl-dev
```

and any equivalent foreign implementation build dependency that exists solely for the old adapter.

## 2. Remove OpenSSL from the Docker runtime image

The runtime image must not install:

```text
libssl3
```

or another foreign TLS/crypto implementation solely for Tardigrade.

Normal CA trust material such as `ca-certificates` remains valid OS/package substrate and must not be confused with a foreign product implementation.

Retain the established non-root UID/GID, filesystem ownership, signal behavior, and local/operator Docker workflow unless a change is required by the native cutover.

## 3. Audit the exact Docker binary/image

Do not infer native-only composition from the Dockerfile text alone.

Add/extend smoke so the binary copied into the runtime stage:

- self-reports the final native production identity;
- passes the native/production binary dependency audit;
- does not dynamically link OpenSSL/libcrypto or another forbidden foreign implementation;
- starts successfully inside the final runtime image;
- handles a minimal real HTTP request.

Where practical, inspect the final runtime image/package inventory too, so Tardigrade-specific foreign implementation dependencies cannot be reintroduced accidentally.

## 4. Remove `openssl-adapter` compatibility from standalone DEB tooling

The standalone Debian builder must package the only supported production Tardigrade implementation.

Remove behavior that:

- accepts `--tls-backend openssl-adapter`;
- infers `tls-backend=openssl-adapter` as a valid package case;
- emits `Depends: libssl3 | libssl1.1` for Tardigrade;
- instructs users to choose between native and OpenSSL production binaries.

The builder should fail clearly if handed a binary that does not satisfy the final native production identity/dependency contract.

Do not rely solely on a filename or caller-provided boolean to establish the backend.

## 5. Remove `openssl-adapter` compatibility from standalone RPM tooling

Apply the same contract to RPM packaging:

- no OpenSSL-backed production mode;
- no Tardigrade OpenSSL runtime dependency metadata;
- final native binary identity/audit required;
- no transitional backend vocabulary in help text or documentation.

## 6. Preserve architecture/cross-package usability

Standalone package generation may still need to package a prebuilt cross-architecture binary that cannot execute on the packaging host.

Do not solve that by reintroducing a caller-controlled `--tls-backend openssl-adapter` trust switch.

Use a trustworthy native-only mechanism, for example:

- require the artifact to come from an already-audited release inventory;
- accept an explicit native dependency-inventory file generated by the canonical audit;
- or otherwise validate the artifact without creating a second backend selector.

The key invariant is that cross-packaging remains possible while the final architecture has only one production implementation.

## 7. Update package/container tests

Add/adjust tests that fail if:

- Docker again installs/links OpenSSL for Tardigrade;
- DEB metadata gains a Tardigrade OpenSSL dependency;
- RPM metadata gains a Tardigrade OpenSSL dependency;
- standalone builders accept `openssl-adapter` as production;
- the packaged/container binary does not identify as the final native implementation;
- startup/request smoke fails.

Reuse `scripts/audit-release-binary.sh` or its post-#649 successor rather than implementing unrelated duplicate linkage logic.

## 8. Reconcile distribution documentation

Update relevant packaging/deployment docs so they no longer describe:

- Docker as OpenSSL-backed;
- DEB/RPM as backend-selectable;
- `libssl` as a Tardigrade runtime requirement.

Do not claim Homebrew is complete here; #466 controls when the public tap can be advertised as real/release-backed.

---

# Explicit non-goals

Do not use this issue to:

- implement #645 or #646;
- perform the core build/backend retirement owned by #649;
- duplicate/replace Homebrew #466;
- publish an official Docker registry image unless separately requested;
- add signing/notarization/DMG/pkg work;
- ban CA certificate packages or ordinary OS/runtime substrate;
- perform the final public-release evidence sweep owned by the separate #634 closeout issue.

---

# Acceptance criteria

## Docker

- [ ] Docker builds the final native production `tardi` implementation.
- [ ] Docker builder installs no OpenSSL/libcrypto development package for Tardigrade.
- [ ] Docker runtime installs no OpenSSL/libcrypto runtime package for Tardigrade.
- [ ] Exact runtime binary passes native-only dependency/linkage audit.
- [ ] Final image startup + real HTTP request smoke passes as the existing non-root user.

## DEB

- [ ] Standalone DEB tooling no longer accepts or emits `openssl-adapter` as a production mode.
- [ ] DEB metadata contains no OpenSSL runtime dependency for Tardigrade.
- [ ] Cross-architecture packaging remains supportable without trusting an obsolete backend selector.
- [ ] Packaged artifact is tied to native-only audit/identity evidence.

## RPM

- [ ] Standalone RPM tooling no longer accepts or emits `openssl-adapter` as a production mode.
- [ ] RPM metadata contains no OpenSSL runtime dependency for Tardigrade.
- [ ] Cross-architecture packaging remains supportable without trusting an obsolete backend selector.
- [ ] Packaged artifact is tied to native-only audit/identity evidence.

## Docs/tests

- [ ] CI/regression coverage detects reintroduction of Tardigrade OpenSSL dependencies in Docker/DEB/RPM.
- [ ] Packaging/deployment docs describe only the native production implementation.
- [ ] #466 remains the explicit Homebrew owner; no contradictory formula ownership is introduced here.

---

# Close rule

Close this issue when **Docker and standalone DEB/RPM tooling can only produce/package the final native Tardigrade production implementation, no Tardigrade OpenSSL dependency remains in their build/runtime metadata, and exact packaged/container artifacts pass native linkage plus startup/request smoke.**

Together with #466, this satisfies #634's remaining distribution-surface cutover.

## Related

- #634 — parent epic
- #649 — remove OpenSSL production backend / default native
- #466 — Homebrew formula and release-to-tap automation
- #476 — native official release/archive/package foundation, merged
- #645 — public-WebPKI signature breadth
- #646 — public-WebPKI size bounds

Jaruso:
## Summary
- Root `Dockerfile`: removed `libssl-dev` from the build stage and `libssl3` from the runtime stage. The default `general` TLS profile has been pure-Zig native since #649, so neither is needed to build or run `tardi`.
- `scripts/test-docker-image.sh`: extracts the exact binary copied into the runtime stage and audits it with `scripts/audit-release-binary.sh` (linkage + self-report); also builds the `build` target and asserts neither stage *explicitly* installs an OpenSSL package (`apt-mark showmanual`, not a raw `dpkg -l` presence check — `ca-certificates` has a genuine transitive `Depends: openssl` on this base image, which must stay legitimate OS substrate).
- `packaging/deb/build.sh` / `packaging/rpm/build.sh`: removed `--tls-backend native|openssl-adapter` and the `tardi version`-based backend inference. The packaged binary's native identity is now always proven with `scripts/audit-release-binary.sh` — self-audited when the binary is host-executable, or via a new `--audit-inventory` flag (a pre-generated inventory JSON matched to the binary by SHA-256) for cross-architecture packaging where the artifact can't run on the packaging host. Neither builder can emit an `openssl`/`libssl` runtime dependency any more.
- `packaging/rpm/tardigrade.spec`: removed the conditional `Requires: openssl-libs` block outright — there is no longer a backend that needs it.
- `scripts/audit-release-binary.sh`: inventory JSON gains a `binary_sha256` field so a caller-supplied `--audit-inventory` can be cryptographically bound to the exact artifact rather than trusted blindly.
- Regression coverage: `scripts/test-deb-package.sh` / `scripts/test-rpm-package.sh` now assert `--tls-backend openssl-adapter` is rejected, exercise the `--audit-inventory` cross-architecture path (including a mismatched-SHA rejection), audit the exact binary the package installs, and start it against its packaged config for a real `/health` request — not just `tardi version`. Both also stop pre-installing `libssl3`/`openssl-devel`/`openssl-libs` into their smoke-test containers, since installing them ahead of time would mask a real regression.
- `scripts/test-rpm-package.sh`: also drops the Rocky/RHEL 9 glibc-version floor (`-Dtarget=$arch-linux-gnu.2.34`). It existed because the old OpenSSL adapter's `libssl.so` pulled in glibc 2.34's reversioned pthread symbols before Zig's default floor knew about them; confirmed by testing that the native build (which links no OpenSSL) no longer needs it.
- Docs (`packaging/README.md`, `docs/DEPLOYMENT.md`, `docs/TLS_DEPENDENCY_POLICY.md`) and `CHANGELOG.md` updated to match; also fixed a stale `--profile native` reference in `packaging/README.md` (the profile name is `general`).
- Small, related fix folded in: `scripts/update-homebrew-formula.sh`'s `validate_native_inventory()` still compared a release inventory's `.profile`/`.reported_profile` against the retired `"native"` name instead of `"general"` (a #654 rename this one check missed), which would have rejected every real release inventory; fixed, along with the stale `test-homebrew-tap-sync.sh` fixture that was masking it. This doesn't take Homebrew publication ownership away from #466 — it's a one-field vocabulary fix so the existing #466-owned tooling reads `#649`'s actual profile name.

## Test plan
- [x] `./scripts/audit-dependencies.sh` passes
- [x] `zig build -Doptimize=ReleaseFast` succeeds; `tardi version` reports `tls-profile=general, tls-backend=native`
- [x] `./scripts/audit-release-binary.sh --binary zig-out/bin/tardi --profile general --output …` passes and emits `binary_sha256`
- [x] `bash -n` on every modified shell script
- [x] Manually exercised the full DEB/RPM build → `--audit-inventory` cross-arch → install → installed-binary audit → real HTTP request flow, and the Docker build/runtime `apt-mark` checks (including the regression case), in real containers since this sandbox can't run the actual CI jobs
- [x] `packaging-smoke` CI job (Docker/DEB/RPM smoke) green on this PR
- [x] Two rounds of review feedback addressed and confirmed by a follow-up review

Closes #650

### DIFF
--- a/packaging/README.md
+++ b/packaging/README.md
@@ -11,12 +11,12 @@ versus what is a local-build-only tool.
 | --- | --- | --- |
 | Linux release archives (`.tar.gz`, x86_64/aarch64) | **Supported, published** | Built and attached to every GitHub release by `.github/workflows/release.yml`, alongside `install.sh`, `tardigrade-checksums.txt`, per-arch SPDX SBOMs, dependency inventories, and provenance. Releases containing #476 build these official artifacts with the default `-Dtls-profile=general` (pure-Zig native since #649); older already-published releases may predate that cutover. |
 | macOS release archives (`.tar.gz`, darwin x86_64/arm64) | **Implemented in #476; awaiting first release** | #476 adds `macos-15-intel` and `macos-15` release rows for `tardigrade-darwin-x86_64.tar.gz` and `tardigrade-darwin-arm64.tar.gz`, using the same archive/SBOM/inventory/provenance pipeline as Linux. Both rows build with the default `-Dtls-profile=general` (pure-Zig native since #649) without Homebrew OpenSSL, audit out foreign TLS/crypto/QUIC/H3 linkage, assert the Mach-O architecture, package/extract the archive, verify native build identity, run a real static-site startup/request smoke from the extracted artifact, and exercise the checksum-verifying installer path. The currently published latest release still predates #476, so these assets are not public until the first intentional release containing it. |
-| DEB (`packaging/deb/build.sh`) | **Supported, published** | Built for `amd64`/`arm64` from the same release binaries as the `.tar.gz` archives and attached to every GitHub release; also usable as a local builder. Host-native builds infer dependency metadata from `tardi version`; cross-compiled binaries can declare `--tls-backend native|openssl-adapter` explicitly. Native binaries declare no OpenSSL runtime dependency, while a transitional local `general`/OpenSSL binary still gets the dependency it actually needs. |
-| RPM (`packaging/rpm/build.sh`) | **Supported, published** | Same treatment as DEB, for `x86_64`/`aarch64`. Host-native builds infer the backend and cross-compiled builds can pass it explicitly. The spec's OpenSSL runtime dependency is conditional on the packaged binary/backend; official native release packages do not declare it. Built from the same Ubuntu-runner binary as the archives — see the glibc compatibility note below if targeting an older RHEL-family release. |
+| DEB (`packaging/deb/build.sh`) | **Supported, published** | Built for `amd64`/`arm64` from the same release binaries as the `.tar.gz` archives and attached to every GitHub release; also usable as a local builder. The packaged binary's native identity is always proven with `scripts/audit-release-binary.sh` — self-audited when the binary is host-executable, or via a SHA-256-bound `--audit-inventory` file for cross-architecture packaging — never by a caller-supplied backend flag (#650). Every package declares no OpenSSL runtime dependency; there is no other production backend to package. |
+| RPM (`packaging/rpm/build.sh`) | **Supported, published** | Same treatment as DEB, for `x86_64`/`aarch64`, including the native-identity audit and no OpenSSL runtime dependency (#650). Built from the same Ubuntu-runner binary as the archives — see the glibc compatibility note below if targeting an older RHEL-family release. |
 | systemd unit (`packaging/systemd/tardigrade.service`) | **Supported** | Installed and structurally validated (unit file text/layout, permissions) by both the DEB and RPM smoke tests. Neither test boots systemd or exercises a real start/status/reload/stop lifecycle — see [docs/DEPLOYMENT.md](../docs/DEPLOYMENT.md#the-systemd-units-pidcontrol-path-contract) for the unit contract these assertions check. |
 | launchd plist (`packaging/launchd/io.baresystems.tardigrade.plist`) | **CI-validated user LaunchAgent template** | The Darwin release-smoke workflow renders the checked-in host-style `/usr/local/...` template into an isolated prefix on the `macos-15` Apple Silicon runner, stages the native Darwin archive's `tardi` binary plus `tardigrade -> tardi` compatibility alias, and proves a real `launchctl bootstrap` -> readiness/request -> `bootout` lifecycle in the current user's launchd domain. |
-| Homebrew (`packaging/homebrew/tardigrade.rb`) | **Preparatory; awaiting native release** | `scripts/update-homebrew-formula.sh` renders the formula from one release tag and verifies the referenced archives exist in that same release. The current public `v0.5.0` release predates the native #634 shipping cutover and old/new archive-layout switch, so it must not be used for the public tap formula. Generate and publish the tap formula only after a release publishes native audited archives with canonical `tardi` plus the `tardigrade` alias. |
-| Docker / OCI image | **Local build supported, not published** | Root [`Dockerfile`](../Dockerfile) and [`compose.yaml`](../compose.yaml) build a runtime image locally; smoke-tested via `scripts/test-docker-image.sh` in the `packaging-smoke` CI job. No registry-published image or container-publishing workflow exists. Docker's remaining OpenSSL cutover is tracked by #634 and is separate from the raw release/archive lane in #476. See [docs/DEPLOYMENT.md](../docs/DEPLOYMENT.md) for the full workflow. |
+| Homebrew (`packaging/homebrew/tardigrade.rb`) | **Preparatory; awaiting native release** | `scripts/update-homebrew-formula.sh` renders the formula from one release tag and verifies the referenced archives exist in that same release. The current public `v0.5.0` release predates the native #634 shipping cutover and old/new archive-layout switch, so it must not be used for the public tap formula. Generate and publish the tap formula only after a release publishes native audited archives with canonical `tardi` plus the `tardigrade` alias. #466, not this document, owns the public tap. |
+| Docker / OCI image | **Local build supported, not published** | Root [`Dockerfile`](../Dockerfile) and [`compose.yaml`](../compose.yaml) build a runtime image locally; smoke-tested via `scripts/test-docker-image.sh` in the `packaging-smoke` CI job, which extracts the exact runtime binary and audits it with `scripts/audit-release-binary.sh` rather than trusting the Dockerfile text. Neither build stage installs OpenSSL/libcrypto for Tardigrade (#650). No registry-published image or container-publishing workflow exists. See [docs/DEPLOYMENT.md](../docs/DEPLOYMENT.md) for the full workflow. |
 
 ## Official release implementation
 
@@ -28,17 +28,22 @@ zig build -Doptimize=ReleaseFast -Dversion=<version>
 ```
 
 The release workflow audits each produced binary with
-`scripts/audit-release-binary.sh --profile native`. That audit fails if the
+`scripts/audit-release-binary.sh --profile general`. That audit fails if the
 artifact links `libssl`, `libcrypto`, or another forbidden foreign
 TLS/crypto/QUIC/H3 implementation and verifies that `tardi version` reports
-`tls-profile=native, tls-backend=native`.
+`tls-profile=general, tls-backend=native`.
 
-Linux DEB/RPM packages are built from that same audited host-native binary. The
-package builders infer its backend from `tardi version`, and the release job then
+Linux DEB/RPM packages are built from that same audited host-native binary.
+`packaging/deb/build.sh`/`packaging/rpm/build.sh` self-audit the binary with
+`scripts/audit-release-binary.sh` (rather than inferring its backend from
+`tardi version` or trusting a caller-supplied flag), and the release job then
 inspects the resulting package metadata and fails if either native package
-declares an OpenSSL/libssl/libcrypto runtime dependency. Local cross-compiled
-packages can provide `--tls-backend native|openssl-adapter` explicitly when the
-target binary cannot execute on the packaging host.
+declares an OpenSSL/libssl/libcrypto runtime dependency. Local
+cross-architecture packages — where the target binary cannot execute on the
+packaging host — provide `--audit-inventory` with a
+`scripts/audit-release-binary.sh --output` inventory already generated for
+that exact binary (matched by SHA-256) instead of asserting a backend
+directly.
 
 ## Quick install (recommended)
 
@@ -104,12 +109,10 @@ the native profile for a shipping-equivalent package:
 # 1. Build the binary first (cross-compile for the target arch as needed)
 zig build -Doptimize=ReleaseFast
 
-# 2. Build the DEB. Passing the backend explicitly also works when the target
-#    binary is for a different architecture than the packaging host.
+# 2. Build the DEB.
 ./packaging/deb/build.sh \
   --version 0.50 \
-  --arch amd64 \
-  --tls-backend native
+  --arch amd64
 
 # Output: dist/tardigrade_0.50_amd64.deb
 ```
@@ -120,13 +123,29 @@ sudo apt install ./dist/tardigrade_0.50_amd64.deb
 sudo systemctl enable --now tardigrade
 ```
 
-When `--tls-backend` is omitted, the DEB builder infers it from an executable
-host-native binary's `tardi version` output. Native binaries have no OpenSSL
-package dependency; a transitional local binary reporting
-`tls-backend=openssl-adapter` gets `Depends: libssl3 | libssl1.1`. Unknown
-backends fail rather than generating ambiguous dependency metadata. For a
-foreign-architecture binary, pass `--tls-backend native` or
-`--tls-backend openssl-adapter` explicitly.
+The DEB builder always proves the packaged binary is Tardigrade's sole native
+production implementation with `scripts/audit-release-binary.sh` — it never
+trusts a caller-supplied backend flag. When `--binary` is executable on the
+packaging host, the builder self-audits it directly. For a foreign-architecture
+binary that cannot run on the packaging host, generate an inventory for it
+first (e.g. on the target architecture, or under emulation) and pass it in:
+
+```bash
+./scripts/audit-release-binary.sh \
+  --binary path/to/foreign-arch/tardi \
+  --profile general \
+  --output /tmp/tardi-inventory.json
+
+./packaging/deb/build.sh \
+  --version 0.50 \
+  --arch arm64 \
+  --binary path/to/foreign-arch/tardi \
+  --audit-inventory /tmp/tardi-inventory.json
+```
+
+The inventory is matched to `--binary` by SHA-256, so it must have been
+generated for that exact artifact. Every produced package has no OpenSSL
+package dependency; there is no other production backend to package.
 
 The DEB package:
 - Installs the binary to `/usr/bin/tardi`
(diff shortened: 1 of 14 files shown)


scripts/test-docker-image.sh
```diff
@@ -55,6 +56,22 @@ fi
...
+# self-reports the native production identity.
+docker create --name "$AUDIT_CONTAINER_NAME" "$IMAGE_TAG" >/dev/null
+docker cp "${AUDIT_CONTAINER_NAME}:/usr/local/bin/tardi" "$TMPDIR/tardi"
+docker rm -f "$AUDIT_CONTAINER_NAME" >/dev/null 2>&1
+chmod +x "$TMPDIR/tardi"
+"${REPO_ROOT}/scripts/audit-release-binary.sh" \
+    --binary "$TMPDIR/tardi" \
+    --profile general \
+    --output "$TMPDIR/dependency-inventory.json"
+echo "docker runtime binary: native-only linkage audit passed"
```
Jaruso: [P2] Keep this exact-binary audit, but it does not cover the other Docker regression #650 explicitly requires: reintroducing an OpenSSL package into either stage. If `libssl-dev` is added back to the builder, or `libssl3` is added back to the runtime image but remains unused, this audit still passes because the binary linkage is unchanged.

Please add a complementary stage/package-root assertion. Because `ca-certificates` is explicitly allowed and may pull crypto packages transitively, check packages that were *manually requested* rather than simply asserting no OpenSSL package exists at all. For example:

```bash
BUILD_IMAGE_TAG="${IMAGE_TAG}-build"
docker build --target build -t "$BUILD_IMAGE_TAG" "$REPO_ROOT"

for tag in "$BUILD_IMAGE_TAG" "$IMAGE_TAG"; do
    manual_tls="$(docker run --rm --entrypoint sh "$tag" -c \
        "apt-mark showmanual | grep -Ei '^(libssl|libcrypto|openssl)' || true")"
    if [[ -n "$manual_tls" ]]; then
        echo "FAIL: Docker stage explicitly installs foreign TLS/crypto packages: $manual_tls" >&2
        exit 1
    fi
done
```

(And clean up the build-stage image tag in `cleanup()`.) This catches the `libssl-dev`/`libssl3` regression while still allowing ordinary CA-store substrate.

scripts/test-deb-package.sh
```diff
@@ -45,6 +45,27 @@ fi
...
+if "${REPO_ROOT}/packaging/deb/build.sh" \
+    --version "$VERSION" \
+    --arch amd64 \
+    --binary "$BINARY" \
+    --tls-backend openssl-adapter \
+    --output "${OUTPUT_DIR}/rejected" 2>/dev/null; then
+    echo "FAIL: deb builder accepted --tls-backend openssl-adapter" >&2
+    exit 1
+fi
+echo "deb builder: --tls-backend openssl-adapter correctly rejected"
```
Jaruso: [P2] The new `--audit-inventory` branch is the mechanism that satisfies #650's cross-architecture-packaging acceptance criterion, but this CI test never enters it; it only packages the executable host-native binary. The PR test plan currently covers this branch manually only.

Please make the inventory path a regression test for **both** builders. You can exercise it without a real foreign architecture: generate a canonical inventory while the current binary is executable, copy the same bytes to a non-executable file (`chmod -x`) so the builder is forced into the inventory branch, and verify packaging succeeds. Then change the bytes (or use an inventory from a different binary) and assert the SHA mismatch is rejected.

For example:

```bash
inventory="$TMPDIR/native-inventory.json"
foreign_sim="$TMPDIR/tardi-cross-sim"
./scripts/audit-release-binary.sh --binary "$BINARY" --profile general --output "$inventory"
cp "$BINARY" "$foreign_sim"
chmod -x "$foreign_sim"

./packaging/deb/build.sh \
  --version "$VERSION" --arch amd64 \
  --binary "$foreign_sim" --audit-inventory "$inventory" \
  --output "$OUTPUT_DIR/cross"

printf '\0' >> "$foreign_sim"
if ./packaging/deb/build.sh ... --binary "$foreign_sim" --audit-inventory "$inventory" ...; then
    echo "FAIL: mismatched audit inventory was accepted" >&2
    exit 1
fi
```

Mirror the success + mismatch cases in `test-rpm-package.sh` (or factor the duplicated verification into a shared helper and test that helper once).

scripts/test-deb-package.sh
```diff
@@ -45,6 +45,27 @@ fi
...
+
+# ── Regression: package metadata must declare no OpenSSL runtime dependency ──
+DEB_DEPENDS="$(dpkg-deb -f "$DEB_PATH" Depends 2>/dev/null || true)"
+if printf '%s\n' "$DEB_DEPENDS" | grep -qiE 'libssl|libcrypto|openssl'; then
+    echo "FAIL: native DEB unexpectedly declares an OpenSSL dependency: $DEB_DEPENDS" >&2
+    exit 1
+fi
+echo "deb metadata: no OpenSSL dependency"
```
Jaruso: [P2] This covers package metadata, but #650's required-work/close rule also asks for the **exact packaged artifact** to prove native identity/linkage and pass startup + a real HTTP request. After installation this test currently only runs `/usr/bin/tardi version >/dev/null`, so it neither asserts the native identity string nor exercises the packaged server.

Please add two checks here and mirror them in the RPM smoke:

1. Extract/audit the binary that is actually inside the package (e.g. `dpkg-deb -x "$DEB_PATH" "$TMPDIR/deb-root"` followed by `scripts/audit-release-binary.sh --binary "$TMPDIR/deb-root/usr/bin/tardi" --profile general`). For RPM, audit the installed `/usr/bin/tardi` or extract the RPM and audit that copy.
2. After installing, start `/usr/bin/tardi` from `/etc/tardigrade/tardigrade.conf` as the `tardigrade` user, wait for readiness, and issue a real request such as `curl -fsS -H 'Host: localhost' http://127.0.0.1:8069/health`. Set the PID file to a path owned by `tardigrade` and terminate/wait for the process cleanly afterward.

That makes DEB/RPM satisfy the same exact-artifact + real-request close rule that the Docker lane already satisfies.

packaging/README.md
```diff
@@ -104,12 +109,10 @@ the native profile for a shipping-equivalent package:
 # 1. Build the binary first (cross-compile for the target arch as needed)
 zig build -Doptimize=ReleaseFast
 
-# 2. Build the DEB. Passing the backend explicitly also works when the target
-#    binary is for a different architecture than the packaging host.
+# 2. Build the DEB.
```
Jaruso: [P3] `packaging/deb/build.sh` now unconditionally invokes `jq`, but this DEB local-build walkthrough never installs it. The RPM walkthrough was updated to `dnf install rpm-build jq`; please do the equivalent here (for example, `sudo apt install jq` before invoking the builder, plus any non-default packaging prerequisite you want to document). Otherwise the documented local DEB path fails on a minimal Debian/Ubuntu host even though the builder's own header lists `jq` as required.

Jaruso (commented):
**Blocking changes before merge.** GitHub will not let this account submit `REQUEST_CHANGES` on its own PR, so this is filed as a comment review, but I would not merge #656 yet.

The implementation direction is good and the current happy paths are working: the default Docker build is native, the standalone builders no longer expose `--tls-backend`, package metadata no longer emits OpenSSL dependencies, the official release workflow does not depend on the removed flag, and CI has already passed lint/dependency-audit plus the DEB/RPM smoke steps.

#650's acceptance/close rule is not fully enforced yet:

1. Docker CI proves the runtime binary is native-only, but it would not fail if `libssl-dev`/`libssl3` were re-added to the Docker stages without being linked. The issue explicitly requires regression coverage for Docker installation as well as linkage.
2. The new `--audit-inventory` mechanism is the feature that preserves cross-architecture packaging, but neither DEB nor RPM CI exercises that branch. Right now its only validation is manual.
3. The issue's close rule requires the exact packaged/container artifacts to pass native linkage plus startup/request smoke. Docker does this; DEB/RPM currently stop at metadata/layout and `tardi version >/dev/null`. Audit the binary actually extracted/installed from each package, assert its native identity, and run a minimal real HTTP request from the installed package.
4. The DEB local-build docs need to install `jq`, which is now an unconditional builder prerequisite (the RPM section already does this).

While touching `scripts/test-rpm-package.sh`, please also fix the Rocky/glibc comment that still explains the target override in terms of dynamically linking `/usr/lib64/libssl.so`; after #649/#650 that explanation is stale and contradicts the native-only production model. If the explicit glibc floor is still required for a native-only reason, document that reason instead.

Once these are covered, the actual cutover code looks aligned with #650 and the next review should be verification of these additions, not another architectural pass.

Jaruso:
Addressed all five points in 0e84b4a6:

1. **Docker install regression coverage** — `scripts/test-docker-image.sh` now builds the `build` target too and asserts neither stage *explicitly* installs an OpenSSL package. Note: a raw `dpkg -l` presence check doesn't work here — `ca-certificates` on `debian:bookworm-slim` has a genuine transitive `Depends: openssl` (→ `libssl3`) for its own tooling, so `libssl3` is unavoidably present in the runtime image regardless of anything Tardigrade does. That's exactly the "ordinary OS/package substrate" the issue carves out. I used `apt-mark showmanual` instead, which distinguishes packages the Dockerfile explicitly named on an `apt-get install` line from ones pulled in only as someone else's dependency — verified it passes on the current Dockerfile and fails on a synthetic one with `libssl3` re-added.

2. **`--audit-inventory` CI coverage** — both DEB and RPM smoke tests now exercise the cross-architecture path (a non-executable copy of the binary stands in for a foreign-arch artifact) plus the mismatched-SHA-256 rejection, instead of leaving it validated only by hand.

3. **Real artifact smoke** — both smoke tests now audit the *exact binary the package installs* with `scripts/audit-release-binary.sh`, then start it with its packaged config and make a real HTTP request against `/health`, not just `tardi version`.

4. **DEB docs** — `packaging/README.md`'s DEB "Build locally" section now lists `jq` as a prerequisite, matching RPM.

5. **Stale glibc comment** — turned out to be more than a wording fix: I tested it directly, and the `-Dtarget=$arch-linux-gnu.2.34` floor is no longer needed at all. It existed because linking the old OpenSSL adapter's `libssl.so` pulled in glibc 2.34's reversioned pthread symbols before Zig's default floor knew about them. Tardigrade's native build links no OpenSSL, so I confirmed (bare native target, no floor) the failure doesn't reproduce and removed the workaround rather than just re-explaining it.

This sandbox can't run the actual CI jobs, so I exercised all of the above manually end-to-end in real containers (Docker build+runtime `apt-mark` checks including the regression case, and the full DEB/RPM build → audit-inventory → install → installed-binary-audit → real HTTP request flow) before pushing.

Jaruso (commented):
Follow-up review of 7db5e31: the four findings from my previous review are addressed, and I do not have any further code changes to request.

Verified against #650:

- Docker now checks both sides of the requirement: the exact runtime binary is audited for native-only linkage/identity, and both build/runtime stages are checked for explicitly installed OpenSSL packages while still allowing `ca-certificates` to bring ordinary crypto substrate transitively.
- DEB and RPM now exercise the `--audit-inventory` cross-architecture path in CI, including a successful SHA-bound inventory case and a mismatched-SHA rejection case.
- Both installed package binaries are audited with the canonical release-binary audit and then used for a real `/health` HTTP request, satisfying the package exact-artifact/startup/request close rule.
- The DEB docs now include `jq`, and the stale Rocky/OpenSSL rationale was reconciled with the native-only build.
- The small Homebrew inventory-vocabulary update (`native` -> `general`) is consistent with #649's final profile identity and does not take ownership away from #466, though the PR description now inaccurately says Homebrew is completely untouched; that is housekeeping, not a merge blocker.

I also rechecked the surrounding original cutover: the Dockerfile itself no longer names OpenSSL packages, DEB/RPM cannot accept `--tls-backend`, package metadata cannot emit the old OpenSSL dependencies, the release workflow still calls the builders compatibly, and #466 remains the explicit public-tap owner.

All four prior review threads are resolved. Lint and the updated DEB smoke are already green on this head; the latest Packaging smoke job is still running the RPM/Docker steps at the time of this review. Assuming that run finishes green, this is ready to merge from code-review/acceptance-criteria perspective.

Issue #676, opened 2026-08-25
test(runtime): torture reload, certificate rotation, drain, and resource stability


Jaruso:
## Goal

Exercise Tardigrade's reload, credential rotation, graceful shutdown/drain, PID control, keepalive, queued-work, and resource-lifetime behavior as an adversarial state-machine campaign against the current v0.6.x runtime.

This is **correctness and hardening**, not benchmark characterization. The primary assertions are atomicity, bounded completion, no state divergence, and resource return/plateau. It can run on ordinary local/CI environments and should not wait on #593/#669.

## Reference contracts

- `docs/RELOAD_SHUTDOWN.md`
- `docs/TIMEOUTS.md`
- `docs/OBSERVABILITY.md`
- `tests/integration.zig`
- existing H3 soak/lifecycle tests

## Required lifecycle guarantees to validate

The campaign should prove the documented behavior rather than invent new semantics:

- successful reload validates/prepares before publishing a new request configuration;
- in-flight requests retain coherent request-scoped state;
- accepted connections can remain open across ordinary reload where supported;
- failed reload leaves the last published configuration active;
- native general-profile credentials can be replaced atomically when the configuration supports hot credential reload;
- restart-owned listener/topology settings are rejected rather than partially applied;
- graceful shutdown drains admitted work up to the configured deadline and then closes remaining queue/connection ownership deterministically;
- H2/H3 drain boundaries/GOAWAY prevent new work past the advertised boundary;
- parked keepalive connections do not prevent process termination forever.

## Harness philosophy

Prefer extending/reusing `tests/integration.zig` and existing runtime metrics/status surfaces over creating a second lifecycle framework.

Add a standalone torture script only if it remains a thin orchestrator around the real `tardi` process and reusable assertions.

## Reload matrix

Run continuous traffic while performing each operation.

### Valid reloads

- [ ] route change
- [ ] static root/content change where reload-owned
- [ ] proxy upstream change where supported
- [ ] timeout/limit change where reload-owned
- [ ] repeated same-config reload
- [ ] SIGHUP path
- [ ] `tardi reload` control path

Assert:

- new requests observe the new config after publication;
- in-flight requests are not corrupted by mixed snapshots;
- active connections remain usable where the contract permits;
- reload success counters increment exactly once per successful publication.

### Invalid reloads

- [ ] syntax error
- [ ] semantic/config validation error
- [ ] invalid path/resource referenced during prepare
[...]

PR #681, opened 2026-08-25
perf: execute #589 competitive benchmark suite on real hardware, fix harness bugs, file real Tardigrade defect

Jaruso:
## Summary

Real-hardware execution of the #589 competitive benchmark suite (Tardigrade
vs NGINX, HAProxy, Caddy) for #593, run on a Debian 13 / Intel N150 host.
This covers **only** the four-server competitive-suite lane of #593 — the
accept-batching/fairness suite (#146/#601), the epoll-vs-io_uring backend
comparison (#148/#602), and the HTTP/3/QUIC evidence lane were not executed
in this session and remain open under #593.

Also merges in the benchmark-harness evidence-integrity fixes originally
opened as #668 (closed in favor of this PR — same commits, carried
forward here).

**Full report:** [`benchmarks/results/2026-08-25-beelink-competitive/REPORT.md`](benchmarks/results/2026-08-25-beelink-competitive/REPORT.md)

### Fourth review round — addressed

All previously reported blockers were confirmed fixed. One remaining item:
the jq `//` false-coercion bug fixed in round 3 had no regression test —
Tests 24-28 covered multi-run aggregation and the remote runner, but none
exercised `write_combined_outputs` itself, so a future reintroduction of
`// true` would have passed every existing test. Added **Test 29**, which
runs the real `write_combined_outputs` (not a copied jq expression) against
a synthetic fixture with explicit `supported:false`/`covered:false` and
asserts both survive real CSV and Markdown rendering — using the fixture
and assertions from the reviewer's inline suggestion.

### Third review round — addressed

1. **`.gitignore` ignored the wrong tree.** `benchmarks/results/` is the
   repo's existing durable dated-results archive (precedent going back to
   2026-05-02) — #593 explicitly prefers a dated summary committed there.
   The actual transient, regenerate-anytime path is
   `benchmarks/competitive/run.sh`'s default `--out-dir`,
   `benchmarks/competitive/results/<timestamp>/`. Now ignoring that
   instead.
2. **The canonical JSON still said the invalid uneven-route scenario was
   `covered: true`.** Root cause: jq's `//` operator treats `false` the
   same as `null`, so every `X // Y // true`-style fallback rendering
   `covered`/`supported` booleans in `competitive/run.sh` silently coerced
   an explicit `false` back to `true`. This also affected the CSV export
   for HAProxy's `static-large-http1` and the H3 rows (their Markdown
   rendering happened to use a direct `== false` check and was already
   correct). Fixed every occurrence, marked `uneven-route-distribution` as
   `covered: false` with a reason in the generator, and added a "Reason"
   column to the Markdown table. Regenerated the retained CSV/Markdown from
   the corrected combined JSON using the real rendering code — the
   underlying benchmark numbers are unchanged, only the coverage
   metadata and the two previously-mis-rendered CSV rows.
3. **The new remote-execution behavior had no regression tests.** Added
   `_remote_wrk_error_count` unit coverage, a failed-SSH-execution test,
   and an unparseable-output test — all asserting `run_wrk_remote` fails
   loudly instead of reporting a fabricated `0`.
4. **Provenance wording mismatch.** REPORT.md's closing note still called
   the pre-rebase `efe0876` "this branch" after the earlier section had
   already explained it's no longer reachable from this branch's history.
   Fixed to match, and clarified that later review-response commits only
   touch rendering/reporting, not the benchmark execution path.

### Second review round — addressed

1. **`remote-run.sh` could silently report `0 req/s, 0 errors` for a
   failed remote run.** Now fails loudly on a failed SSH/wrk execution or
   unparseable output, and counts Non-2xx + Socket errors together like
   `benchmarks/run.sh`'s own `wrk_error_count`.
2. **`uneven-route-distribution` report claim corrected.** The underlying
   harness runs its three routes sequentially, not concurrently, so it
   doesn't test route-fairness under concurrent uneven traffic. Filed the
   harness-design gap as **#683**.
3. **RSS claim corrected** — scoped to the core competitive-suite
   scenarios, with the one exception (`upstream-tls-handshake-reuse`,
   8.89 MiB) called out explicitly.
4. **Local-vs-cross-worker reuse ratio surfaced** — #593 explicitly asks
   for this; the data was already computed, just never shown.
5. Two smaller `remote-run.sh` bugs: missing `--remote` parser arm, and a
   `head -n -1` GNU-only portability bug (caught by my own new test on
   macOS).

### First review round — addressed

1. Report/results restored (combined artifacts only, not the ~50
   intermediate per-pass raw JSON files).
2. Fixed a real shell-injection hole in `remote-run.sh` (unvalidated
   `--threads`/`--connections`/`--duration`).
3. Dropped an unrelated #676 commit via rebase onto current `main`.
4. Added regression coverage for the #668 multi-run aggregation fix.
5. Made the repetition/variance gap explicit rather than implying
   completeness #593 doesn't yet have.

### Harness bugs found and fixed (this PR)

- HAProxy reverse-proxy routes always 404'd since #589 (rule-ordering
  issue in `haproxy.cfg.in`).
- `idle-keepalive-active-traffic` silently reported 0 req/s for every
  server (k6's `K6_DURATION` env var collides with its own reserved
  override mechanism).
- `remote-run.sh` could fabricate a clean `0 req/s, 0 errors` result on a
  failed remote execution.
- jq's `//` operator silently coerced explicit `covered`/`supported: false`
  values back to `true` in every retained CSV/Markdown rendering.

### Real Tardigrade defect found and filed (not fixed here)

**#682**: `max_requests_per_connection` (default 100) closes the
connection without ever sending `Connection: close` on the response that
hits the limit, so a client that reuses the connection per the header it
just received gets no response on its next request — the ~1% wrk error
rate on Tardigrade's static-file scenarios that NGINX/HAProxy/Caddy don't
show at the same or higher throughput. Root cause and suggested fix
included; left for dedicated follow-up.

### Harness design gap found and filed (not fixed here)

**#683**: the upstream-pool matrix's `uneven-route-distribution` scenario
runs its three routes sequentially rather than as concurrent mixed
traffic, so it doesn't test the concurrent route-fairness question
#149/#593 actually ask about. Needs concurrent-process orchestration in
`upstream-pool-matrix.sh`, not a one-line fix — left for dedicated
follow-up. The retained artifact now correctly marks this scenario
`covered: false` with this explanation.

### Headline results (see REPORT.md for full tables/caveats)

- Tardigrade's RSS stayed under 7 MiB across every core competitive-suite
  scenario — dramatically lower than NGINX (~28 MiB), HAProxy (~28-30 MiB),
  Caddy (~55-58 MiB).
- HAProxy leads CPU-efficiency and connection-churn/keepalive-burst
  throughput on this host; NGINX and HAProxy both beat Tardigrade's raw
  tiny-static req/s (partly explained by #682's ~1% dropped-request rate).
- Upstream-pool contention sweep (1/2/4 workers) shows no lock-contention
  bottleneck in this host's worker-count range.
- Roughly 59-62% of reused upstream connections crossed worker boundaries
  in the higher-volume scenarios.
- **Explicitly not covered**: repeated/variance-checked evidence (single
  clean pass only), and concurrent uneven-route fairness (#683).

### Environment (see REPORT.md for full metadata)

Debian 13, Linux 6.17.13-2-pve, Intel N150 (4 cores), 10 GiB RAM, Zig
0.16.0, Tardigrade built `-Doptimize=ReleaseFast` with the default
`-Dtls-profile=general` — confirmed pure-Zig native (no OpenSSL linkage via
`ldd`). Not a dedicated idle host (concurrent RTSP transcoding, Docker,
Cloudflare tunnel, etc.) — see REPORT.md's non-idle-host caveat.

## Related issues

Refs #593
Refs #682 (real Tardigrade protocol bug filed from this work)
Refs #683 (harness concurrency gap filed from this work)
Supersedes #668 (same harness-audit commits, carried forward in this branch)

## Tests

- `bash benchmarks/test-report.sh` — 13/13 passed
- `bash benchmarks/test-h3-benchmark.sh` — 106/106 passed (23 new
  assertions added across all three review rounds)
- `shellcheck benchmarks/*.sh benchmarks/competitive/*.sh` — clean

## Security impact

Fixed a shell-injection hole in `benchmarks/remote-run.sh` (`--threads`/
`--connections`/`--duration` interpolated unquoted into a remote SSH
command). Local benchmark tool, not production code, but the fix is real.

## Performance impact

None to Tardigrade itself. The one real Tardigrade defect found (#682) is
filed separately, not fixed here.

## Checklist

- [x] `zig fmt --check build.zig src/ tests/` — not applicable, no `src/`/`tests/` changes
- [x] `zig build test --summary all` — not applicable, no source changes
- [x] `benchmarks/test-report.sh` and `benchmarks/test-h3-benchmark.sh` green
- [x] New harness behavior covered by regression tests (Tests 24-29)
- [ ] README / CHANGELOG — not updated; this is benchmark execution/results, not operator-visible behavior


.gitignore
```diff
@@ -48,6 +48,10 @@ RESEARCH*
 # Local benchmark baselines (laptop-fallback runs, not canonical)
 benchmarks/results/baseline-pre-change/
 
+# Benchmark run output — large raw artifacts, not meant to live in git.
+# Summarize findings in the PR/issue instead of committing this tree.
+benchmarks/results/
+
```
Jaruso: **Blocking: this ignores the wrong side of the benchmark evidence boundary and deletes the report #593 needs.** `benchmarks/results/` is already the repository's durable dated-results convention, and the current PR head no longer contains the linked `.../REPORT.md`. Meanwhile the competitive harness's *generated* default output is `benchmarks/competitive/results/...`, so this rule does not even suppress the default transient output directory.

Please keep durable summaries under `benchmarks/results/` available to git, restore the 2026-08-25 human-readable report (or an equivalent durable report), and ignore only transient generated output if that is the intent, e.g.

```gitignore
# Local/generated competitive harness output
benchmarks/competitive/results/
```

Large raw JSON/CSV can be attached elsewhere, but the report must link the durable archive/location so #593's artifact-retention acceptance criteria remain verifiable after this PR is merged/squashed.

benchmarks/remote-run.sh
```diff
@@ -124,18 +140,23 @@ run_wrk_remote() {
...
-        header_flags="-H 'Host: ${HOST_HEADER}'"
+        header_flags="-H $(sq "Host: ${HOST_HEADER}")"
     fi
 
     local raw
+    # shellcheck disable=SC2029 # sq() is meant to expand client-side, quoting each value for the remote shell
     raw=$(ssh "$REMOTE_HOST" \
-        "${WRK_PATH} -t${THREADS} -c${CONNECTIONS} -d${DURATION}s -L ${header_flags} '${url}'" \
+        "$(sq "$WRK_PATH") -t${THREADS} -c${CONNECTIONS} -d${DURATION}s -L ${header_flags} $(sq "$url")" \
         2>&1) || true
```
Jaruso: **Blocking: the SSH command is still injectable through the numeric CLI options.** `WRK_PATH`, the header, and URL are now shell-quoted, but `THREADS`, `CONNECTIONS`, and `DURATION` come directly from `--threads`, `--connections`, and `--duration` and are interpolated raw into remote shell text. For example, a value such as `--duration '1; touch /tmp/injected #'` terminates the `wrk` command and starts another remote command.

Because this patch specifically claims to fix unsafe interpolation, validate these arguments before constructing the command. For example:

```bash
validate_positive_int() {
    local name="$1" value="$2"
    [[ "$value" =~ ^[1-9][0-9]*$ ]] || {
        echo "$name must be a positive integer, got: $value" >&2
        exit 2
    }
}
validate_positive_int --duration "$DURATION"
validate_positive_int --connections "$CONNECTIONS"
validate_positive_int --threads "$THREADS"
```

Quoting complete arguments with `sq "-t${THREADS}"`, etc. would also close the shell boundary, but numeric validation is preferable because these options have numeric semantics.

tests/integration.zig
```diff
@@ -1466,10 +1466,12 @@ fn prepareBearClawFixture(
     defer allocator.free(config_text_cert);
     const config_text_key = try std.mem.replaceOwned(u8, allocator, config_text_cert, "/etc/tardigrade/tls/privkey.pem", server_key_abs);
     defer allocator.free(config_text_key);
+    const config_text_sn = try std.mem.replaceOwned(u8, allocator, config_text_key, "server_name api.example.com;", "server_name tardigrade.test;");
```
Jaruso: **Please drop this unrelated #676 test-harness commit from PR #681.** This Bearclaw/TLS rewrite comes from `ac240510` (`test: resolve Issue 676 test harness failures`), not the #593 benchmark work. #680 has now merged into `main` and closed #676 with its own integration-test changes, so this branch is also one merge commit behind current main and carries an alternate implementation of already-merged work.

Please rebase onto current `main` and remove `ac240510` from this PR. After that rebase, if any specific Bearclaw change is still demonstrably needed, put it in a focused follow-up with its own rationale/tests. Keeping it here makes the benchmark PR unnecessarily risky and also contradicts the PR checklist statement that there are no `tests/` changes.

benchmarks/test-h3-benchmark.sh
```diff
@@ -187,9 +187,16 @@ echo "==> Test 2c: --runs > 1 aggregates each run's scenario-local quic delta in
     RESULTS_JSON='{}'
     declare -A _run_quic_list=()
     declare -A _run_rps_list=()
+    declare -A _run_p50_list=()
```
Jaruso: These new declarations make the existing QUIC aggregation tests compatible with the expanded `flush_multirun_result`, but they do not actually test the evidence-integrity behavior added in `benchmarks/run.sh`. A regression could again drop p50/p95/p999/throughput/CPU/RSS/FD—or coerce a null sample incorrectly—and this test would still pass because it only asserts the QUIC block/runs counters.

Please add a focused multi-run fixture with values for every newly aggregated metric plus at least one `null`, then assert the resulting JSON. For example, exercise `p50='1 null 3'` and verify the aggregate is `2`, verify p95/p999/throughput/CPU are retained, and use RSS/FD samples such as `10 14 12` / `20 25 22` to prove the output peaks are `14` / `25`. That makes the #668 fix itself regression-protected rather than only keeping the old tests compiling.

Jaruso (commented):
**Review result: changes required before merge.** GitHub will not let this account submit a formal `REQUEST_CHANGES` review on its own pull request, so I am recording the blocking review as `COMMENT` instead.

I reviewed the full PR against the competitive-suite portion of #593, the carried-forward #668 harness audit, the benchmark evidence that existed before the final deletion commit, #682, current `main`, and the current CI run.

There is good work here: the HAProxy routing fix is correct, the k6 named-scenario failure is real and the failure propagation fix is appropriate, the #668 evidence-integrity changes are directionally sound, unsupported HAProxy static-large/H3 cases are called out rather than fabricated, and the Tardigrade static-file failure was correctly isolated into focused follow-up #682.

I found four changes required before merge:

1. **The durable benchmark evidence was deleted and the PR's report link is dead.** The current head returns 404 for `benchmarks/results/2026-08-25-beelink-competitive/REPORT.md`. Commit `9c0df293` removed the 317-line report plus the combined JSON/CSV, per-pass JSON, summary, and upstream-pool data, while claiming the full report lives in the PR description. It does not: the PR description contains only headlines/caveats. This fails #593's requirements to preserve canonical artifacts/environment metadata and provide a human-readable report in-repo or linked from the issue. If raw artifacts are too large for git, that is fine, but they need a durable attachment/link and the repository should retain the concise report/provenance.

2. **`remote-run.sh` still has a remote-shell injection hole in the exact code being hardened.** `sq()` now protects paths/URLs/headers, but `THREADS`, `CONNECTIONS`, and `DURATION` remain raw CLI strings interpolated into the SSH command. Validate them as positive integers before command construction (preferred), or quote each complete option before it crosses the SSH shell boundary.

3. **This PR contains an unrelated alternate #676 integration-test implementation.** Commit `ac240510` changes Bearclaw/TLS integration helpers and tests, while the PR description says there are no `tests/` changes. #680 has since merged to `main` and closed #676 with its own test-harness changes. Please rebase onto current `main` and drop the unrelated #676 commit from this benchmark PR; if anything from it is still necessary after the rebase, it should be a focused follow-up rather than bundled into #593 evidence work.

4. **The multi-run evidence-integrity fix is not regression-tested for the metrics it adds.** `test-h3-benchmark.sh` was updated only enough to declare the new accumulator arrays; its existing assertions still check QUIC aggregation, not p50/p95/p999/throughput/CPU/RSS/FD retention or null exclusion. Please add focused assertions proving (a) a null sample is not coerced to zero, (b) the latency/throughput/CPU values survive `--runs > 1`, and (c) RSS/FD use the intended maximum. Otherwise the central #668 fix can regress while the cited 83/83 test suite remains green.

Acceptance reconciliation for this PR's stated slice: the four-server non-smoke run appears to have executed and material defects were investigated/filed; unsupported rows are documented; the 1/2/4-worker pool sweep is explicitly scoped and correctly does **not** claim to satisfy #593's high-core requirement. The accept-batching, epoll-vs-io_uring, H3/QUIC, and true high-core lanes remain open, as the PR says. One additional #593 requirement also remains unproven in the deleted report: the issue asks for enough repetitions to identify run-to-run instability, but the documented canonical command is a single run and no variance/repetition analysis is preserved. If this PR is intended to represent completion of the competitive-server lane rather than a preliminary real-hardware slice, please add repeated clean runs/variance evidence; otherwise make that limitation explicit as remaining work under #593.

CI was still in progress at the end of this review. Format, lint/shellcheck, core tests, native TLS jobs, security, and several build jobs were green; I did not treat still-running jobs as failures.

benchmarks/remote-run.sh
```diff
@@ -124,18 +156,23 @@ run_wrk_remote() {
...
     if [[ -n "$HOST_HEADER" ]]; then
-        header_flags="-H 'Host: ${HOST_HEADER}'"
+        header_flags="-H $(sq "Host: ${HOST_HEADER}")"
     fi
 
     local raw
+    # shellcheck disable=SC2029 # sq() is meant to expand client-side, quoting each value for the remote shell
     raw=$(ssh "$REMOTE_HOST" \
-        "${WRK_PATH} -t${THREADS} -c${CONNECTIONS} -d${DURATION}s -L ${header_flags} '${url}'" \
+        "$(sq "$WRK_PATH") -t${THREADS} -c${CONNECTIONS} -d${DURATION}s -L ${header_flags} $(sq "$url")" \
```
Jaruso: **Blocking: this remote runner can still fabricate a clean result when the remote execution fails.** The command immediately below still ends with `|| true`, so a failed SSH session or non-zero `wrk` exit is discarded. Then the parser only counts `Non-2xx`; it does not count wrk's `Socket errors: connect/read/write/timeout`. A dropped SSH connection can therefore fall through as `rps=0`, `p50=0`, `p99=0`, `errors=0`, and a real `Socket errors: read N` run can be recorded with `errors=0`.

That is the same class of evidence-integrity failure this PR just fixed for k6 and `run_connection_churn`. Please fail closed here too and reuse the same socket-error semantics as the local runner. For example:

```bash
local ssh_status=0
raw=$(ssh "$REMOTE_HOST" \
    "$(sq "$WRK_PATH") -t${THREADS} -c${CONNECTIONS} -d${DURATION}s -L ${header_flags} $(sq "$url")" \
    2>&1) || ssh_status=$?

if (( ssh_status != 0 )); then
    echo "remote wrk failed for ${label} (ssh/wrk exit ${ssh_status})" >&2
    printf '%s\n' "$raw" >&2
    return 1
fi

if ! grep -qE '^Requests/sec:' <<<"$raw"; then
    echo "remote wrk produced no usable result for ${label}" >&2
    printf '%s\n' "$raw" >&2
    return 1
fi

errors=$(printf '%s\n' "$raw" | wrk_error_count)
if (( errors != 0 )); then
    echo "remote wrk reported ${errors} client/socket errors for ${label}" >&2
    printf '%s\n' "$raw" >&2
    return 1
fi
```

`wrk_error_count` can be shared/extracted or duplicated with the exact `Non-2xx` + `Socket errors:` parsing already used by `benchmarks/run.sh` / the upstream-pool matrix. Please add a regression covering both a non-zero SSH/wrk exit and a synthetic `Socket errors: read 3` output so this cannot silently regress again.

benchmarks/results/2026-08-25-beelink-competitive/REPORT.md
```diff
@@ -0,0 +1,346 @@
...
+  `static-large-http1` scenarios reproducibly show roughly a 1% wrk `read`
+  socket-error rate that NGINX, HAProxy, and Caddy do not show at the same
+  or higher throughput, on the same host, same harness, same run. This was
+  isolated to a real, independently-reproducible protocol bug — see below.
+
+### Upstream pool matrix (Tardigrade only)
+
+From `upstream-pool-matrix.json` / the table in `competitive-summary.md`:
+
+- **Uneven route distribution**: route-a-hot (8244 req/s), route-b-warm
```
Jaruso: **Blocking: this row is not actually an uneven concurrent-traffic test, so the starvation conclusion is unsupported.** The retained harness implementation creates these values with sequential independent runs:

```bash
hot_json="$(run_measured_wrk hot-origin hot "$DURATION" "$CONNECTIONS" "$THREADS" true)"
route_a_json="$hot_json"
route_b_json="$(run_measured_wrk route-b route-b 2 4 1 false)"
route_c_json="$(run_measured_wrk route-c route-c 1 2 1 false)"
```

and only afterward labels them with `requested_weight: 80/15/5`. Route A runs for ~15s, then B for ~2s, then C for ~1s; the routes never contend with each other. Their req/s values are isolated-route capacities, not an observed 80/15/5 traffic distribution, so they cannot tell us whether a hot route starves warm/cold routes. #149/#593 explicitly require **uneven traffic distribution across workers/routes**.

Please either:

1. fix `upstream-pool-matrix.sh` to generate the 80/15/5 routes concurrently and rerun this row, preserving per-route success/latency plus shared pool metrics; e.g. a k6 script with concurrent scenarios/executors for `/route-a`, `/route-b`, and `/route-c`, or three synchronized background load generators; **or**
2. mark this scenario `covered: false` / invalid in the retained artifact, remove the starvation conclusion, and explicitly carry the true uneven-route run forward as remaining #593 work.

Do not keep the current sequential measurements labeled as starvation/fairness evidence.

Jaruso (commented):
Second full review against current head and #593/#149 acceptance requirements.

The **four blockers from my first review are fixed** and I resolved those threads:
- durable report/canonical artifacts are restored;
- the numeric SSH interpolation hole is closed with positive-integer validation;
- the unrelated #676 integration-test work is gone after the rebase;
- the multi-run evidence-integrity fix now has meaningful regression coverage.

CI is also fully green on `b2fcb572`.

I still would **not merge this revision yet** because two published-evidence paths can produce or claim invalid results:

1. `remote-run.sh` still swallows a failed remote `ssh/wrk` execution and ignores wrk socket errors, so it can record `0 req/s, 0 errors` for a failed run. That directly conflicts with this PR's evidence-integrity goal.
2. The report presents the upstream `uneven-route-distribution` row as starvation/fairness evidence, but the underlying harness runs route A, B, and C sequentially with different durations/concurrency. They never compete, so this does not exercise uneven concurrent route traffic as required by #149/#593.

I also left two report-accuracy/acceptance comments: the `RSS < 7 MiB across every scenario` statement is contradicted by the retained upstream TLS row (`8.89 MiB`), and the report should explicitly surface the measured local-vs-cross-worker reuse ratio because #593 requires that analysis.

Two smaller cleanup items I found while doing the complete pass:
- `remote-run.sh` documents `--remote SSH_HOST`, but the argument parser has no `--remote)` arm, so the advertised override currently exits as an unknown option. Please add `--remote) REMOTE_HOST="$2"; shift 2 ;;` (ideally with a parsing smoke test).
- The PR body says `benchmarks/results/` is now gitignored and the five result files were force-added, but the current rebased `.gitignore` does **not** ignore that directory. Please correct the PR description so the retention policy matches the tree. Also, the report's final `efe0876 (this branch...)` wording should identify `efe0876` as the actual **pre-rebase benchmarked SHA** rather than implying it is still reachable from the rebased branch; recording the equivalent post-rebase commit mapping would make provenance unambiguous.

Because GitHub does not let the authenticated PR author formally request changes on their own PR, this is submitted as `COMMENT`, but the two evidence-integrity findings above are blocking in substance.

benchmarks/results/2026-08-25-beelink-competitive/REPORT.md
```diff
@@ -0,0 +1,346 @@
...
+- **Idle-keepalive + active traffic** (k6, 20 idle holders + 10 active-burst
+  VUs): HAProxy leads (22975 req/s), then NGINX (18588), then Tardigrade and
+  Caddy essentially tied (13550 / 13526).
+- **CPU usage**: Tardigrade's reported `cpu_pct_avg` on the tiny-static pass
+  (232.80%, i.e. ~2.3 cores of the host's 4) is comparable to NGINX
+  (214.58%) and lower than Caddy (269.71%). HAProxy used less CPU than all
+  three (201.00%) while serving ~3.4x Tardigrade's and ~1.8x NGINX's
+  req/s on this scenario — HAProxy is the clear CPU-efficiency leader on
+  this workload on this host.
+- **RSS**: Tardigrade's RSS stayed consistently under 7 MiB across every
```
Jaruso: Please correct this claim before publishing the report. It is true for the four-server comparison rows, but not **every scenario** in the retained evidence: `upstream-pool-matrix.json` records `rss_mb_peak: 8.89` for `upstream-tls-handshake-reuse`. The PR description repeats the same `<7 MiB across every scenario` wording.

A precise version would be something like: “Across the four-server competitive scenarios, Tardigrade stayed under 7 MiB RSS; the separate upstream-pool matrix peaked at 8.89 MiB in the TLS-origin scenario.” That preserves the genuinely strong memory result without overstating the artifact.

benchmarks/results/2026-08-25-beelink-competitive/REPORT.md
```diff
@@ -0,0 +1,346 @@
...
+### Upstream pool matrix (Tardigrade only)
+
+From `upstream-pool-matrix.json` / the table in `competitive-summary.md`:
+
+- **Uneven route distribution**: route-a-hot (8244 req/s), route-b-warm
+  (8078 req/s), route-c-cold (6885 req/s) — reuse ratio ≥ 0.9998 on every
+  route; no evidence of one route starving another under this traffic
+  shape.
+- **Many low-volume origins** (16 origins): even distribution, ~7.5 req/s
+  per origin as configured, reuse ratio 1.0, 0 errors.
```
Jaruso: #593 explicitly requires the report to cover **local vs cross-worker upstream reuse behavior**, but this bullet only gives the aggregate reuse ratio. The retained hot-origin row already has the evidence: `local_reuse=48,623`, `cross_worker_reuse=75,881`, `reused_connections=124,504`, so cross-worker reuse is about **60.95%** of reuse on this 4-worker run.

Please surface that ratio (and preferably the 1/2/4-worker contention-row ratios as well) and interpret it. This matters because it proves the shared pool is not merely achieving a high reuse rate by staying worker-local; cross-worker sharing is actually being exercised, which is one of #149/#593's explicit validation questions.

.gitignore
```diff
@@ -48,6 +48,12 @@ RESEARCH*
 # Local benchmark baselines (laptop-fallback runs, not canonical)
 benchmarks/results/baseline-pre-change/
 
+# Benchmark run output — large raw artifacts, not meant to live in git.
+# Summarize findings in the PR/issue instead of committing this tree.
+# (Individual report/combined-result files can still be force-added when a
+# run's findings are worth keeping — see benchmarks/results/2026-08-25-*/.)
+benchmarks/results/
```
Jaruso: **Blocking: this reintroduces the round-one evidence-boundary bug.** `benchmarks/results/` is not just disposable harness output; it is already the repo's durable dated-results convention, and #593 explicitly prefers dated human-readable conclusions there. Current `main` already tracks `2026-05-02`, `05-06`, `05-10`, `05-13`, and `08-10`, and this PR deliberately adds `2026-08-25-beelink-competitive` there. Ignoring the entire tree means the next durable benchmark report is silently hidden unless someone remembers `git add -f`.

The generated/transient competitive harness output is under `benchmarks/competitive/results/`, so keep the archive path trackable and ignore the generator path instead, e.g.

```gitignore
# Local/generated competitive harness output
benchmarks/competitive/results/
```

If there are specific large raw artifact subdirectories that should never be committed, ignore those narrowly. Please do not globally ignore `benchmarks/results/`.

Jaruso (commented):
Round 3 review result: **changes still required before merge**. GitHub will not allow this account to submit `REQUEST_CHANGES` on its own PR, so this is recorded as a blocking COMMENT review.

The second-round implementation work is materially better: `remote-run.sh` now preserves non-zero SSH/wrk failures, refuses unparseable throughput output, counts wrk socket errors, and correctly parses `--remote`; the RSS claim is corrected; the local-vs-cross-worker reuse analysis is now surfaced; and #683 is a good focused follow-up for the real concurrent uneven-route harness work. I resolved the prior RSS and cross-worker-reuse threads.

Four items remain:

1. **Blocking — `.gitignore` reintroduces the exact evidence-boundary problem from round one.** `benchmarks/results/` is the repository's existing durable dated-results archive (2026-05-02, 05-06, 05-10, 05-13, 08-10, and this PR's 08-25 report). #593 explicitly prefers a dated summary there. Ignoring that entire tree means future durable reports silently disappear unless every author remembers `git add -f`. The transient generated path is `benchmarks/competitive/results/`; ignore that instead.
2. **Blocking — the canonical machine-readable artifact still says the invalid uneven-route scenario is covered.** The prose now correctly says the sequential route passes are not fairness evidence, and #683 correctly tracks the fix, but `upstream-pool-matrix.json` still has `"covered": true` and `competitive-summary.md` renders `Covered = true`. Any automation consuming the canonical JSON will conclude #149/#593's uneven-route requirement passed. Mark this scenario invalid/uncovered in the retained artifact and summary while preserving the per-route numbers as informational measurements.
3. **Blocking test gap — the new remote execution behavior is not regression-tested.** The previous review explicitly requested coverage for a non-zero SSH/wrk exit and synthetic `Socket errors: read 3`. Test 25 only proves the unrelated `--remote` parser arm works. The PR checklist currently says new harness behavior is covered by Tests 24/25, but the evidence-integrity behavior added to `run_wrk_remote` has no automated assertion. Add focused tests for `_remote_wrk_error_count`, failed remote execution, and ideally the unparseable-output fail-closed path.
4. **Minor provenance correction — REPORT.md still ends by calling `efe0876` “this branch” after the report itself correctly explains that SHA is pre-rebase and no longer reachable from the branch.** Make the final reproducibility note match the earlier provenance explanation and, preferably, identify the reachable post-rebase equivalent commit alongside the exact benchmarked SHA.

Acceptance reconciliation remains intentionally partial: this PR can merge without completing all of #593 because it explicitly leaves repetitions/variance, accept-batching/fairness, epoll-vs-io_uring, H3, high-core contention, and now concurrent uneven-route fairness open. The requirement here is simply that the evidence this PR *does* retain accurately identifies what is and is not covered.

CI for head `213e1c08` is largely green at review time (format, lint/shellcheck, unit/integration profiles, perf smoke/H3 harness tests, security, Homebrew, TLS/PKI, builds); Packaging smoke was still finishing its Docker-image step when checked. No currently completed job was failing.

benchmarks/results/2026-08-25-beelink-competitive/upstream-pool-matrix.json
```diff
@@ -0,0 +1,453 @@
+{
+  "_meta": {
+    "suite": "upstream-pool-distribution",
+    "duration_s": 15,
+    "worker_threads": 4,
+    "note": "p99_ttfb_ms is measured from k6 http_req_waiting where present; p99_ms is wrk end-to-end latency."
+  },
+  "scenarios": {
+    "uneven-route-distribution": {
+      "covered": true,
```
Jaruso: **Blocking: the canonical artifact still claims this invalid scenario is covered.** The updated REPORT.md correctly explains that these three measurements were sequential isolated-route passes and therefore do **not** satisfy #149/#593's concurrent uneven-route/fairness requirement; #683 correctly tracks the harness rewrite. But the machine-readable source of truth still says `covered: true`, and `competitive-summary.md` therefore still renders `uneven-route-distribution | true`.

That leaves automation/downstream readers with the opposite conclusion from the prose. You do not need to implement #683 in this PR, but this retained artifact must describe the evidence honestly. For example:

```json
"uneven-route-distribution": {
  "covered": false,
  "invalid_reason": "Sequential per-route measurements; concurrent uneven-route fairness was not exercised. See #683.",
  "measurements_valid_for": "isolated per-route throughput/reuse only",
  "routes": { ... }
}
```

Then update/regenerate `competitive-summary.md` so its Covered column is also `false`. Preserve the route numbers; just stop marking them as satisfying the scenario.

benchmarks/test-h3-benchmark.sh
```diff
@@ -746,6 +761,78 @@ check "validation doc does not document nonexistent netem apply/clear subcommand
...
+check "rss_mb_peak uses the run-to-run maximum (12), not a mean or the last sample" \
+    '"rss_mb_peak": 12' "$MULTIRUN_METRICS_JSON"
+check "open_fds_peak uses the run-to-run maximum (45), not a mean or the last sample" \
+    '"open_fds_peak": 45' "$MULTIRUN_METRICS_JSON"
+check "errors sums across all runs (0+2+1=3), not just the last run" \
+    '"errors": 3' "$MULTIRUN_METRICS_JSON"
+rm -f /tmp/tardi-multirun-metrics-test.json
+
+echo ""
+echo "==> Test 25: remote-run.sh's documented --remote flag is actually parsed, not an unknown option (PR #681 review)"
```
Jaruso: **The requested regression coverage for the remote evidence-integrity fix is still missing.** This new Test 25 only proves `--remote` is accepted by argument parsing. The actual second-round blocker was that `run_wrk_remote` could turn failed SSH/wrk execution and socket errors into apparently clean evidence. The implementation now fixes that, but nothing here asserts it.

Please add focused regression coverage for at least both cases requested in the previous review:

```bash
# Parser semantics: socket errors must be counted.
eval "$(source_functions \"$REMOTE_RUN_SH\" _remote_wrk_error_count)"
REMOTE_ERRORS="$(printf '%s\n' \
  'Socket errors: connect 0, read 3, write 0, timeout 0' |
  _remote_wrk_error_count)"
check "remote wrk counts socket errors" "3" "$REMOTE_ERRORS"
```

and run `run_wrk_remote` in a subshell with a mocked `ssh()` that returns (say) 255, asserting the subshell exits non-zero. A third mock returning exit 0 but output without `Requests/sec:` should assert the unparseable-output path also fails closed. That directly protects the behavior this PR now claims to have fixed; a `--remote --help` smoke test does not.

benchmarks/results/2026-08-25-beelink-competitive/REPORT.md
```diff
@@ -0,0 +1,413 @@
...
+Exact commands used for the canonical run:
+
+```bash
+zig build -Doptimize=ReleaseFast   # Zig 0.16.0
+./benchmarks/competitive/run.sh \
+  --binary zig-out/bin/tardi \
+  --out-dir benchmarks/competitive/results/2026-08-25-beelink-competitive
+```
+
+Tardigrade commit: `efe0876` (this branch, `claude/pr-593-implementation-84d51b`).
```
Jaruso: Small provenance inconsistency: the environment table now correctly says `efe0876` is the **pre-rebase benchmarked SHA and is no longer reachable from this branch**, but this final line still calls it “this branch.” Please make the reproducibility footer match the earlier explanation, e.g. `Benchmarked SHA: efe0876 (pre-rebase); equivalent reachable post-rebase benchmark code: 3de7fd7f` (or whatever exact wording/provenance you want to preserve). Since #593 explicitly requires the exact Tardigrade SHA/commands, this should not contradict itself at the point readers will copy from.

Jaruso (commented):
Fourth-pass review on `b91631be`.

All previously reported blockers are now fixed and I resolved their threads:
- `.gitignore` now ignores only `benchmarks/competitive/results/`, leaving the durable `benchmarks/results/` archive trackable.
- `uneven-route-distribution` is now `covered: false` with an explicit reason in the generator and retained JSON/CSV/Markdown.
- `remote-run.sh` now has direct regression coverage for socket-error counting, failed SSH/wrk execution, and unparseable output.
- the REPORT.md provenance footer now matches the pre-rebase benchmarked-SHA explanation.

I rechecked the retained CSV/Markdown and the false values are correct now (HAProxy static-large and H3 unsupported rows are `false`; the uneven-route matrix row is also `false`). I did not find another implementation or artifact correctness bug in the new delta.

**One remaining requested change before merge:** the newly discovered jq `//` false-coercion bug itself is not regression-protected. Tests 24–28 cover multi-run aggregation and the remote runner, but none execute `write_combined_outputs` with explicit `supported: false` / `covered: false`. Because this exact bug silently made machine-readable benchmark evidence claim unsupported/uncovered scenarios were valid, please add a focused synthetic output test before merging. I left the concrete fixture/assertions inline.

CI run #1692 is currently on attempt 2; the perf-smoke job (including `benchmarks/test-h3-benchmark.sh`) and the other completed jobs are green, with one ARM unit-test job still running as of this review.

I am using COMMENT rather than REQUEST_CHANGES because this is the authenticated author's own PR. Once the false-preservation regression test is present and green, I expect this PR to be ready from my review perspective.

Jaruso (commented):
Re-reviewed current head `ff087f1f` against the full #593/#149 scope this PR claims, all prior review findings, and the latest delta.

**No further code or evidence changes requested.** Test 29 now exercises the actual `write_combined_outputs` path with explicit `supported:false` and `covered:false`, and asserts those values survive into both CSV and Markdown. That closes the last regression-protection gap from the previous review. The latest two commits only modify `benchmarks/test-h3-benchmark.sh`; I found no new behavior or evidence regressions.

I also re-verified the previously blocking areas: durable benchmark results remain trackable under `benchmarks/results/`; transient competitive output is ignored under `benchmarks/competitive/results/`; remote `wrk` execution fails closed and has direct regression coverage; the sequential uneven-route measurements are explicitly marked `covered:false` with #683 tracking the real concurrent-fairness work; RSS/provenance/local-vs-cross-worker reporting remains consistent; and the unrelated #676 integration changes remain absent.

The PR is still correctly scoped as a **partial #593 deliverable**, not closure of #593: repeated variance evidence, high-core testing, concurrent uneven-route fairness, accept batching/fairness, epoll-vs-io_uring, and the separate H3/QUIC evidence lane remain outside this run as documented.

Current CI run #1697 is still in progress. Every completed job I checked is green, including format, lint/shellcheck, macOS/Linux builds/tests, native TLS listener jobs, TLS interop, PKI differential, dependency review, and Homebrew smoke. The benchmark-harness Perf smoke job is still running at review time.

**Review verdict: ready to merge once the current CI run completes successfully.** I am submitting this as COMMENT rather than APPROVE because GitHub does not allow approving your own PR.

Issue #677, opened 2026-08-25
test(http): run H2/H3 real-client functional and malformed-input release sweep


Jaruso:
## Goal

Run a black-box **functional, interoperability, malformed-input, lifecycle, and resource-settle** sweep for HTTP/2 and HTTP/3 against the current native shipping `tardi` artifact before controlled-host performance work.

This issue is intentionally narrower than #389:

- #389 owns final H2/H3 **stable promotion**, performance baseline, support-matrix changes, and release-definition closeout.
- this issue owns a pre-performance **correctness hardening pass** that can run on ordinary local/CI environments now.

Any evidence produced here should be reusable by #389, but closing this issue does not by itself mark H2/H3 stable.

## Artifact rule

Where practical, exercise the same native release candidate / installed artifact used by the other v0.6.x hardening issues. Record:

- `tardi version`
- source/release SHA
- artifact/package identity
- OS/architecture
- client/tool versions

Do not infer release behavior solely from a debug/source-only binary when an exact release artifact is available.

## Existing owners to reuse

- `tests/integration.zig`
- `tests/quic_h3_smoke.zig`
- `tests/quic_h3_e2e.zig`
- `tests/quic_h3_udp_smoke.zig`
- `tests/http3_soak.zig`
- `tests/h3_interop_tool.zig`
- `scripts/interop/run-interop.sh`
- `docs/HTTP3_VALIDATION_EVIDENCE.md`
- `docs/HTTP3_ROLLOUT.md`
- `docs/RELOAD_SHUTDOWN.md`
- #389 stable-promotion checklist

Do not duplicate the native H3 parser/state-machine fuzzing tracked by #675.

# HTTP/2 sweep

## External client capability

Use at least one independent H2 client such as `nghttp`, a verified H2-capable curl build, and/or browser tooling. Record the exact version and capability output.

Example:

```bash
nghttp --version || true
h2load --version || true
nghttp -nv https://tardigrade.test:18443/
```

## Negotiation / entry

- [ ] TLS 1.3 handshake succeeds.
- [ ] ALPN selects `h2`.
- [ ] H2 is not entered when ALPN did not select `h2`.
- [ ] valid client connection preface accepted.
- [ ] initial SETTINGS exchange succeeds.

## Application behavior

- [ ] static GET 200
- [ ] HEAD
- [ ] 404/error response
- [ ] configured `return`/simple response path
- [ ] reverse proxy request/response
- [ ] upstream 5xx/error propagation
[...]

Issue #389, opened 2026-07-11
epic(http): graduate HTTP/2 and HTTP/3 from experimental to stable


Jaruso:
## Goal

Graduate **HTTP/2** and **HTTP/3 / QUIC** from `experimental` to `stable` in `docs/SUPPORT_MATRIX.md` when each protocol independently satisfies Tardigrade's stable-support contract against the native Zig shipping `tardi` artifact.

This is now a **final release-definition / evidence / documentation closeout tracker**, not an implementation epic.

HTTP/2 and HTTP/3 may graduate independently. Close #389 only when both rows are `stable` and the README, release notes/checklist, operator docs, and actual released artifact all agree with that claim.

---

# Reconciliation — 2026-08-31

Re-audited against current `main` at `7afc1b4920c83f48ce404bfbf9309796fa8f4583`, the completed installed-artifact sweep in #677, the completed controlled-host campaign in #593 / PR #699, the release-process reconciliation in #678, and the now-completed public Homebrew boundary in #670.

**Do not cancel this issue.** The stable labels have still not been flipped: `docs/SUPPORT_MATRIX.md` continues to list both H2 and H3 as `experimental`, and the README still presents the stable product primarily as HTTP/1.1.

However, the previous body materially overstated what remains. Almost all correctness, lifecycle, interop, H3 performance, native-artifact, packaging, and release-process evidence now exists. The remaining work is narrow.

## Completed owners to consume — do not reopen or recreate

| Owner | Current meaning for #389 |
| --- | --- |
| #247 | **complete** — integrated native H3 assurance/soak foundation is reusable. |
| #255 | **complete** — H3 observability/artifact work is reusable. |
| #256 | **complete** — QUIC/H3 transport hardening and benchmark machinery are reusable. |
| #625 / #640 | **complete / non-blocking** — qvis compatibility is not a promotion ceremony. |
| #677 / #694 | **complete** — installed `v0.6.4` H2/H3 release correctness, malformed-input, external-client/peer, lifecycle, Alt-Svc, cancellation/drain, and resource-settle evidence is reusable. |
| #593 / #699 | **complete** — controlled-host performance campaign is done; do not reopen a general Proxmox/competitive benchmark campaign here. |
| #678 | **complete** — release cadence, hardening-evidence provenance, security-plan, packaging, and release-checklist reconciliation are done. |
| #670 | **complete** — public Homebrew installed-artifact smoke machinery/evidence is done. |

[...]

PR #684, opened 2026-08-26
test: extend HTTP release sweep coverage

Jaruso:
## Summary
Refs #677. Reusable by #389.

This PR continues the post-#680 #677 release sweep work by adding repeatable harness coverage for the automatable local rows and fixing integration harness failures that still reproduced on current `main` in this environment.

## What changed
- Added `-Dtardigrade-bin-path=...` so integration tests can exercise a selected installed/release-candidate `tardi` binary instead of always using the freshly installed build output.
- Added `scripts/run-http-release-sweep.sh`, which records artifact identity/tool versions and runs the focused H2/H3 deterministic release-sweep rows against `TARDI_BIN`.
- Tightened downstream H2 malformed-frame handling to emit protocol-scoped GOAWAY/RST behavior for invalid SETTINGS, WINDOW_UPDATE, stream-zero, CONTINUATION sequencing, and HPACK failures.
- Added focused integration tests for the new H2 malformed failure-scope rows, including stream-level WINDOW_UPDATE zero preserving the connection for an unrelated later stream.
- Switched remaining Bearclaw HTTPS integration probes off curl and onto the SNI-aware pure-Zig TLS helper, preserving Host header behavior while avoiding the macOS LibreSSL/fixture SNI path.
- Extended the native upstream H2 best-effort retry path to absorb the observed `ConnectionRefused` startup/teardown race.
- Updated #677/#389 evidence docs to point at the new sweep path and to mark the early Bearclaw failures as historical, not current product defects.

## Validation
Passing locally on macOS arm64:
- `git diff --check`
- `zig build test-integration -Dintegration-test-filter='interop.h2.' --summary all --error-style verbose` -> 27/27 passed
- `TARDI_BIN="$PWD/zig-out/bin/tardi" scripts/run-http-release-sweep.sh` -> completed; H2 focused rows, native TLS/H2 rows, `test-quic`, and `build-h3-interop` passed; external H3 peers skipped because no peer env vars were configured
- `zig build test --summary all --error-style verbose` -> 3841/3851 passed, 10 skipped
- `zig build test-integration --summary all --error-style verbose` -> 169/188 passed, 19 skipped
- `zig build test-integration -Doptimize=ReleaseFast -Dintegration-test-filter='interop.h2.' --summary all --error-style verbose` -> 27/27 passed

## Artifact and client identity captured by the sweep script
- selected `TARDI_BIN`: `zig-out/bin/tardi`
- `tardi version`: `dev (tls-profile=general, tls-backend=native)`
- binary SHA-256: `78bf8e1bdd74a9d162f86b761ea4093bca414ce82d757faf7f823077dd94a8d5`
- OS: Darwin 25.3.0 arm64
- Zig: 0.16.0
- curl: 8.7.1 with nghttp2/1.68.0, no HTTP/3 support reported here
- nghttp/h2load: nghttp2/1.69.0
- OpenSSL: 3.6.3
- GnuTLS: 3.8.13

## Resource-settle evidence
The sweep reran `zig build test-quic`, including the PR-safe H3 resource-settle soaks. The observed after-settle samples returned tracked connections, active CID routes, native connections, and open file descriptors to baseline for repeated connections, resumed reconnects, and cancelled requests. Detailed samples are in the command output and summarized in `docs/HTTP2_HTTP3_RELEASE_SWEEP_677.md`.

## Skipped or still-open #677 rows
This PR does not claim #677 is fully closable yet. Remaining rows are explicitly documented:
- actual installed/Homebrew release-candidate artifact execution; the `TARDI_BIN=... scripts/run-http-release-sweep.sh` path now exists, but this environment used local `zig-out/bin/tardi`
- independent `nghttp` application exchange with correct loopback SNI; this `nghttp` build has no `--resolve`/`--connect-to`, so it cannot prove `tardigrade.test` SNI without host resolver changes
- browser protocol attempts for Safari/Chrome/Firefox
- quiche and aioquic external H3 peer directions; no peer paths configured here
- live Alt-Svc advertised-endpoint H3 proof and disabled/withdrawn endpoint proof
- malformed/truncated frame-header and declared-payload-shorter-than-frame rows where no complete H2 frame exists to answer with GOAWAY

No new focused GitHub issues were opened from this run.

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -323,7 +323,11 @@ pub fn build(b: *std.Build) void {
     allocation_regression_step.dependOn(&run_allocation_regression.step);
 
     const integration_options = b.addOptions();
-    integration_options.addOption([]const u8, "tardigrade_bin_path", b.getInstallPath(.bin, "tardi"));
+    integration_options.addOption(
+        []const u8,
+        "tardigrade_bin_path",
+        b.option([]const u8, "tardigrade-bin-path", "Path to the tardi executable used by integration tests") orelse b.getInstallPath(.bin, "tardi"),
+    );
 
     const integration_mod = b.createModule(.{
         .root_source_file = b.path("tests/integration.zig"),
--- a/docs/HTTP2_HTTP3_RELEASE_SWEEP_677.md
+++ b/docs/HTTP2_HTTP3_RELEASE_SWEEP_677.md
@@ -556,6 +556,9 @@ covered:
 - valid fragmented HEADERS/CONTINUATION blocks are buffered, decoded once
   `END_HEADERS` arrives, and dispatched only after the complete field section
   is available.
+- over-limit encoded HEADERS/CONTINUATION accumulation resets only the
+  offending stream and leaves a later unrelated stream on the same connection
+  usable.
 - malformed/truncated HPACK integer encoding now returns connection-scope
   `GOAWAY` with `COMPRESSION_ERROR`.
 
@@ -567,7 +570,7 @@ zig build test-integration -Dintegration-test-filter='interop.h2.' \
 ```
 
 Result on 2026-08-25: passed. Build summary reported `8/8 steps succeeded;
-29/29 tests passed`.
+30/30 tests passed`.
 
 ## Independent H2 Client Attempts
 
--- a/docs/HTTP2_HTTP3_RELEASE_SWEEP_677.md
+++ b/docs/HTTP2_HTTP3_RELEASE_SWEEP_677.md
@@ -556,9 +556,9 @@ covered:
 - valid fragmented HEADERS/CONTINUATION blocks are buffered, decoded once
   `END_HEADERS` arrives, and dispatched only after the complete field section
   is available.
-- over-limit encoded HEADERS/CONTINUATION accumulation resets only the
-  offending stream and leaves a later unrelated stream on the same connection
-  usable.
+- over-limit encoded HEADERS/CONTINUATION accumulation returns connection-scope
+  `GOAWAY` with `COMPRESSION_ERROR`, including when existing buffered request
+  bodies leave insufficient remaining connection-memory budget.
 - malformed/truncated HPACK integer encoding now returns connection-scope
   `GOAWAY` with `COMPRESSION_ERROR`.
 
@@ -570,7 +570,7 @@ zig build test-integration -Dintegration-test-filter='interop.h2.' \
 ```
 
 Result on 2026-08-25: passed. Build summary reported `8/8 steps succeeded;
-30/30 tests passed`.
+31/31 tests passed`.
 
 ## Independent H2 Client Attempts
 
--- a/docs/HTTP2_HTTP3_RELEASE_SWEEP_677.md
+++ b/docs/HTTP2_HTTP3_RELEASE_SWEEP_677.md
@@ -47,6 +47,32 @@ Observed identity:
 - file type: `Mach-O 64-bit executable arm64`
 - size: `4.6M`
 
+## Repeatable Release-Artifact Sweep Path
+
+this PR adds a reusable sweep entrypoint that can target an installed or
+release-candidate binary directly:
+
+```sh
+TARDI_BIN="$(command -v tardi)" \
+  scripts/run-http-release-sweep.sh
+```
+
+For local ReleaseFast fallback validation when no installed package is
+available:
+
+```sh
+zig build -Doptimize=ReleaseFast -Dversion=issue-677-release-sweep
+TARDI_BIN="$PWD/zig-out/bin/tardi" \
+  scripts/run-http-release-sweep.sh
+```
+
+The script records `tardi version`, the source SHA, executable path, SHA-256,
+OS/architecture, Zig, curl, nghttp, h2load, OpenSSL, GnuTLS, and configured H3
+peer paths into `.zig-cache/http-release-sweep-677/metadata.txt`, then runs
+the focused H2/H3 deterministic rows with `-Dtardigrade-bin-path` so
+integration tests exercise the selected artifact instead of silently using the
+freshly built debug binary.
+
 ## Passed Local Gates
 
 ```sh
@@ -455,14 +481,15 @@ the example targets. That produced:
 
 This workaround was local to `/tmp` and is not a Tardigrade product change.
 
-## Failed Local Gate
+## Historical Full Integration Gate Attempt
 
 ```sh
 zig build test-integration --summary all --error-style verbose
 ```
 
-Result: failed. Build summary reported `6/8 steps succeeded`; the integration
-test binary reported `160 pass, 19 skip, 3 fail (182 total)`.
+Historical result from the early #680 evidence pass: failed. Build summary
+reported `6/8 steps succeeded`; the integration test binary reported
+`160 pass, 19 skip, 3 fail (182 total)`.
 
 Failing tests:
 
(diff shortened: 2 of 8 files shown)


Jaruso (commented):
I found several blockers before this is safe to merge:

1. **WINDOW_UPDATE error handling is RFC-wrong and the new regression test codifies the wrong behavior.** In `src/edge_gateway.zig`, `parseWindowUpdateIncrement(...)` collapses both an increment of zero and an invalid payload length into the same catch path, then emits `FLOW_CONTROL_ERROR`. RFC 9113 §6.9 requires an increment of zero to use `PROTOCOL_ERROR` (RST_STREAM for a stream, connection error for stream 0), while a WINDOW_UPDATE whose length is not exactly 4 must always be a connection-level `FRAME_SIZE_ERROR`. The new `window_update_zero_increment_connection_goaway_and_stream_rst` test expects 0x3, so it currently locks in the non-conformant code. Split these conditions and assert the required codes/scopes.

2. **`scripts/run-http-release-sweep.sh` can return success after a failed gate.** The script uses `set -u`, but neither `set -e` nor an explicit failure accumulator. `run_step` returns the command status, but each top-level call is followed by more commands, and the final `say "release sweep completed"` returns 0. A failed H2/H3/test gate can therefore be followed by a successful later step/skip and the whole release sweep exits 0. This harness needs fail-fast behavior (`set -e`/careful conditionals) or explicit status aggregation with a final non-zero exit.

3. **The new "release-artifact H2 + H3 sweep" only targets `TARDI_BIN` for the integration/H2 portions.** `zig build test-quic`, `zig build build-h3-interop`, and `scripts/interop/run-interop.sh` exercise source-tree test code / `zig-out/bin/h3_interop_tool`; the existing interop script explicitly launches that tool as its native client/server. So `TARDI_BIN=/opt/homebrew/bin/tardi scripts/run-http-release-sweep.sh` does not prove H3 behavior of that installed artifact even though the PR/docs describe the sweep that way. #677's goal is black-box H2/H3 against the shipping artifact. Either add an H3 black-box row that actually launches the selected `tardi`, or make the evidence/claims explicit that those H3 rows are source-tree regression evidence rather than release-artifact proof.

4. **CONTINUATION sequencing is tracked without actually assembling the field block.** A HEADERS frame without END_HEADERS is still HPACK-decoded immediately, can mark the request ready on END_STREAM, and `h2DispatchReadyStreams` still runs at the end of that loop iteration. Later CONTINUATION payloads are discarded; the code only clears `continuation_stream_id` on END_HEADERS. That means a valid fragmented header block can fail decoding or be processed before the field section is complete. Buffer HEADERS/CONTINUATION fragments for the stream and decode/dispatch only after END_HEADERS.

There are also two current CI blockers on head `2192139`: **Lint** fails shellcheck with SC2015 on lines 83-84 of the new sweep script, and **Production dependency audit (Linux)** rejects the new script's direct reference to `scripts/interop/run-interop.sh` as an explicit non-production helper from production-scanned metadata. Those need to be resolved before merge as well.

The overall direction is good—artifact selection, focused malformed-frame coverage, and the Bearclaw SNI cleanup are useful—but these issues undermine the sweep's correctness/green signal.

Jaruso (commented):
Second pass on head `70451fa`: the prior review items are substantially addressed, but I still see two blockers in the new H2 handling.

1. **Stream-level zero WINDOW_UPDATE sends RST_STREAM but leaves the stream live internally.** The `raw_inc == 0` stream branch writes `RST_STREAM(PROTOCOL_ERROR)` and immediately `continue`s without removing `streams`, `pending`, `pending_responses`, or `ready_streams`. That differs from the existing stream-window-overflow path immediately below, which performs that cleanup after sending RST_STREAM. Once we send RST_STREAM, the stream is closed from our side; retaining its pending request/state means later frames on that stream can still be processed internally and the retained body/response allocations live until some later cleanup/connection close. Please factor the stream-reset cleanup into a helper and use it for both zero-increment and overflow errors. Add a regression that proves the offending stream cannot subsequently complete/dispatch (while an unrelated stream still can).

2. **The new CONTINUATION accumulator is unbounded and bypasses the configured request-header/connection-memory limits while END_HEADERS is absent.** `continuation_block.appendSlice(...)` grows on every valid CONTINUATION frame, but no check is made against `cfg.request_limits.effectiveMaxHeadersTotalSize()` (default 32 KiB) or `cfg.max_connection_memory_bytes`. Because RFC sequencing permits an arbitrary number of consecutive CONTINUATION frames until END_HEADERS, a client can keep sending <=16 KiB frames and grow this allocation for the lifetime of the connection. This turns the correctness fix into a straightforward memory-DoS path and also bypasses the repository's explicit aggregate-header limit before HPACK decode. Bound the encoded field-section accumulator (ideally accounting it in connection memory as well), fail the stream/connection according to the chosen H2 limit policy, and add a regression using repeated CONTINUATION frames without END_HEADERS.

Everything else from my previous review looks fixed on this head: WINDOW_UPDATE length/zero error codes are now split correctly, the sweep uses `set -eu`, the H3 source-tree-vs-artifact distinction is explicit, the direct interop helper reference was removed, shellcheck and the production dependency audit are green, and valid fragmented HEADERS/CONTINUATION are now decoded only after END_HEADERS.

CI is still running overall, but the previously failing Lint and Production dependency audit jobs are both green.

Jaruso (commented):
Third pass on head `df77116`: the two issues from the prior review were addressed structurally (centralized stream reset cleanup and a bounded CONTINUATION accumulator), and CI is now fully green. I still see one protocol blocker and one resource-accounting gap in the new limit behavior:

1. **The over-limit HEADERS/CONTINUATION path discards a field block without decompressing it, which violates RFC 9113 §4.3 and can desynchronize HPACK state for the entire connection.** When the encoded block exceeds `max_encoded_header_block_bytes`, the code sends `RST_STREAM(ENHANCE_YOUR_CALM)`, clears `continuation_block` / `continuation_stream_id`, and then allows later streams to proceed. But HTTP/2 field compression is connection-wide and stateful: RFC 9113 §4.3 explicitly requires receivers to reassemble and decompress HEADERS/CONTINUATION field blocks even when those frames are going to be discarded; if a receiver does *not* decompress a field block, it MUST terminate the connection with `COMPRESSION_ERROR`. See https://www.rfc-editor.org/rfc/rfc9113.html#section-4.3. The current regression does not expose this because its oversized continuation bytes are all zero and the later stream uses a fresh literal block, so it does not depend on any dynamic-table mutation that may have been discarded. Either (a) keep consuming/decompressing the rejected field block while suppressing application delivery, preserving the HPACK decoder context, or (b) if the implementation intentionally stops buffering/decompressing at the limit, make the condition connection-fatal with `GOAWAY(COMPRESSION_ERROR)` rather than resetting only the stream. A strong regression would put an incremental-indexing representation in the rejected block and have the next stream reference that dynamic-table entry; the connection-survival implementation must still decode that correctly.

2. **The new CONTINUATION buffer is capped *individually* by `max_connection_memory_bytes`, but is not accounted together with memory already buffered on the connection.** The existing DATA path enforces `buffered_request_bytes + frame.payload.len <= max_connection_memory_bytes`; the new header accumulator instead uses `min(max_headers_total_size, max_connection_memory_bytes)` without subtracting/accounting `buffered_request_bytes`, and `continuation_block` is never added to that accounting. So a connection that already has buffered request bodies can retain those bytes plus another full header-block allowance and exceed the operator-facing `max_connection_memory_bytes` contract. Please either include `continuation_block` in the same connection accounting or bound it against the *remaining* connection-memory budget. A focused test can hold body bytes on one stream, then start an oversized fragmented header block on another and assert the configured aggregate ceiling is respected.

The previous stream-level zero-WINDOW_UPDATE cleanup issue itself looks fixed: `h2ResetStreamState` now removes stream/pending response/request/ready state and is reused by both zero-increment and overflow reset paths. The prior sweep/CI issues also remain fixed, and both CI workflows on this head completed successfully.

Jaruso (commented):
Fourth pass on head `8521377`: I do not see a remaining code blocker from the issues raised in the prior reviews.

The two outstanding items from the third pass are addressed correctly:

1. **HPACK state / over-limit CONTINUATION behavior:** the server now sends connection-scope `GOAWAY(COMPRESSION_ERROR)` and terminates instead of resetting one stream and discarding the remainder of an undecoded field block. That is the correct RFC 9113 §4.3 escape hatch when the implementation chooses not to decompress the rejected field block, so there is no longer a connection-wide HPACK desynchronization risk.

2. **Aggregate request-memory accounting:** `h2EncodedHeaderBlockLimit(...)` now subtracts `buffered_request_bytes` from `max_connection_memory_bytes`, and the new regression covers the case where a partially buffered request body leaves insufficient room for the CONTINUATION field block.

The earlier WINDOW_UPDATE cleanup, CONTINUATION assembly, sweep fail-fast behavior, H3 artifact/source-tree scoping, shellcheck, and production dependency-audit issues also remain fixed.

One follow-up worth tracking separately (not introduced by this PR, so I am not blocking #684 on it): downstream H2 still does not apply the configured *decoded* request limits (`max_header_count`, `max_header_size`, decoded `max_headers_total_size`, and URI length) inside/after HPACK decompression. `h2EncodedHeaderBlockLimit` is a useful encoded-memory cap, but HPACK can expand/repeat indexed fields, and `Decoder.decode` currently allocates output without those request-limit checks. That deserves its own hardening issue, ideally with bounded HPACK output accounting, but it is pre-existing behavior rather than a regression in this PR.

At this point the code review is clean from my side. Current head CI has restarted and is still pending (`CI` queued, Darwin Release Smoke in progress), so I would merge once those head-specific checks finish green.

Jaruso (commented):
Fifth pass on head `941382a`: clean from my side.

The only change since the prior clean review is `tests/bounded_process.zig`, addressing the Ubuntu unit-test failure/race. The new timeout-path check is appropriately narrow: when `MultiReader.fill(...)` reaches the deadline, `reapExitedChild(...)` first uses nonblocking `waitpid` to determine whether the direct child has already exited. Only in that proven-exited case does the helper preserve the child's real exit/signal term via `resultForTerm(...)`; otherwise it falls through to the existing timeout kill/reap path. After reaping the direct child it still terminates the process group, so descendants that kept stdout/stderr pipes open are cleaned up rather than leaked.

I don't see a new double-wait, stale-child-id, or process-group cleanup regression in this delta. The shared result conversion also removes duplicated term/outcome/output ownership logic cleanly.

Head `941382a` is mergeable, and both the full CI workflow and Darwin Release Smoke completed successfully. The H2/H3 issues from the earlier review passes remain resolved. I have no blocking findings on this head.

Because this PR is authored by the authenticated account, I can't give it a meaningful self-approval state; treat this COMMENT as merge-ready from this review.

PR #685, opened 2026-08-26
fix(http): send Connection: close on the response that hits max_requests_per_connection

### LINKED ISSUE #682: fix(http): max_requests_per_connection closes socket without signaling Connection: close, dropping the client's next request
## Summary

`max_requests_per_connection` (default 100, `src/edge_config.zig:1367`) closes
a keep-alive connection server-side after the configured request count, but
the response that hits the limit is still sent with `Connection: keep-alive`.
The very next request a client sends on that (now-closed) connection gets no
response at all — `RemoteDisconnected`, not a clean `Connection: close` on
the last valid response the way HTTP/1.1 keep-alive is supposed to signal a
connection teardown.

Under any sustained keep-alive load this produces client-visible errors at a
rate of `1 / max_requests_per_connection` (~1% at the default 100) — real,
counted `wrk` "Socket errors: read" failures on every reused connection.

## Root cause

`src/edge_gateway.zig:1271-1285` (in the loop driving one physical
connection):

```zig
var keep_alive = false;
handleConnection(conn, session, cfg, ctx.state, &keep_alive, connection_ip, enable_proxy_protocol) catch |err| { ... };
served.* += 1;
const max_requests_per_connection = cfg.max_requests_per_connection;

if (!keep_alive or http.shutdown.isShutdownRequested()) return .close;
if (max_requests_per_connection > 0 and served.* >= max_requests_per_connection) return .close;
```

`handleConnection` (`src/edge_gateway.zig:3734`) decides `keep_alive` purely
from the request's own `Connection` header / HTTP version — it has no idea
how many requests have already been served on this connection, so it cannot
know this is the last one the server will allow. The response is written and
sent with `Connection: keep-alive` (or no header, defaulting to keep-alive
under HTTP/1.1). Only *after* `handleConnection` returns does the caller
check `served.* >= max_requests_per_connection` and unilaterally close the
socket — with the client already told, via the response headers it just
received, that it's free to reuse the connection.

## Reproduction

Built from current `main` (ReleaseFast, default `general` TLS profile — not
TLS-related, plain HTTP/1.1), config with `root` serving a static file,
default `max_requests_per_connection` (100):

```python
import http.client
conn = http.client.HTTPConnection("127.0.0.1", <port>, timeout=3)
for i in range(1, 105):
    conn.request("GET", "/tiny.txt")
    resp = conn.getresponse()
    resp.read()
    print(i, resp.status, resp.getheader("Connection"))
```

Output:

```
97 200 keep-alive
98 200 keep-alive
99 200 keep-alive
100 200 keep-alive
101 EXCEPTION RemoteDisconnected: Remote end closed connection without response
```

Also visible under `wrk` at any sustained keep-alive concurrency — e.g.
`wrk -t4 -c32 -d15s http://127.0.0.1:<port>/tiny.txt` against a plain static
route consistently reports `Socket errors: read <N>` where
`N ≈ total_requests / max_requests_per_connection`, and 0 connect/write/
timeout errors, isolating this to exactly this failure mode.

## Suggested fix direction

Either:
- Have `handleConnection` (or its caller) know the connection is about to
  hit the limit *before* writing the response, and send `Connection: close`
  on that final response, or
- Keep the socket open long enough after the limit-triggering response to
  cleanly read-and-discard (or reject with a real HTTP response, e.g.
  `HTTP/1.1 200` + `Connection: close`, or at minimum a graceful FIN) a
  request the client sends before it can process the close.

The first approach is more standard — it's exactly what nginx/HAProxy/Caddy
do for their own `keepalive_requests`-equivalent settings, which is why they
show zero errors at the same or higher throughput in side-by-side comparison
(see #593).

## Found via

Discovered executing #593's real-hardware competitive benchmark suite —
Tardigrade's plain static-file scenario reproducibly showed a ~1% wrk error
rate that NGINX/HAProxy/Caddy did not show at the same or higher throughput,
in an otherwise identical harness/config. Isolated and confirmed on an
otherwise-idle single-instance repro, independent of the benchmark harness.

Refs #593

Jaruso:
## Summary
- `handleConnection` decided `Connection: keep-alive` vs `close` purely from the request's own header/HTTP version, with no idea how many requests had already been served on the physical connection. When a response hit `max_requests_per_connection`, the caller (`serveOneRequestWithConfig`) closed the socket right after — but the response itself still went out promising `keep-alive`, so the client believed it could reuse the (now-closed) connection and got `RemoteDisconnected` on its next request instead of a clean `Connection: close`.
- `serveOneRequestWithConfig` now computes whether the in-flight request is the last one this connection will be allowed to serve (`served + 1 >= max_requests_per_connection`) and passes that down into `handleConnection`, which forces `keep_alive = false` for that response so the client sees a standard `Connection: close` — matching how nginx/HAProxy/Caddy handle their own `keepalive_requests`-equivalent settings.

## Test plan
- [x] Added `tests/integration.zig`: `max_requests_per_connection sends Connection: close on the response that hits the limit (#682)` — sets `TARDIGRADE_MAX_REQUESTS_PER_CONNECTION=3`, sends 3 keep-alive requests on one socket, asserts the first two say `keep-alive`, the third says `close`, and the socket then reads a clean EOF.
- [x] Verified the new test fails without the fix (`RemoteDisconnected`-style assertion failure on the `Connection: close` check) and passes with it.
- [x] `zig build` and `zig build test` pass.

Closes #682

### DIFF
--- a/src/edge_gateway.zig
+++ b/src/edge_gateway.zig
@@ -1269,8 +1269,15 @@ fn serveOneRequestWithConfig(
         header_timeout_ms;
     setConnTimeouts(conn, header_timeout_ms, write_timeout_ms);
 
+    const max_requests_per_connection = cfg.max_requests_per_connection;
+    // This is the last request the server will allow on this physical
+    // connection: handleConnection must send `Connection: close` on its
+    // response rather than `keep-alive`, or the client will reuse a
+    // connection the server is about to tear down and lose its next request.
+    const is_last_allowed_request = max_requests_per_connection > 0 and served.* + 1 >= max_requests_per_connection;
+
     var keep_alive = false;
-    handleConnection(conn, session, cfg, ctx.state, &keep_alive, connection_ip, enable_proxy_protocol) catch |err| {
+    handleConnection(conn, session, cfg, ctx.state, &keep_alive, connection_ip, enable_proxy_protocol, is_last_allowed_request) catch |err| {
         if (isBenignDisconnect(err)) {
             ctx.state.logger.debug(null, "keepalive connection closed by peer: {}", .{err});
         } else {
@@ -1279,7 +1286,6 @@ fn serveOneRequestWithConfig(
         return .close;
     };
     served.* += 1;
-    const max_requests_per_connection = cfg.max_requests_per_connection;
 
     if (!keep_alive or http.shutdown.isShutdownRequested()) return .close;
     if (max_requests_per_connection > 0 and served.* >= max_requests_per_connection) return .close;
@@ -3731,7 +3737,7 @@ fn setConnTimeouts(conn: anytype, read_timeout_ms: u32, write_timeout_ms: u32) v
     }
 }
 
-fn handleConnection(conn: anytype, session: *ConnectionSession, cfg: *const edge_config.EdgeConfig, state: *GatewayState, keep_alive_out: *bool, connection_ip: []const u8, enable_proxy_protocol: bool) !void {
+fn handleConnection(conn: anytype, session: *ConnectionSession, cfg: *const edge_config.EdgeConfig, state: *GatewayState, keep_alive_out: *bool, connection_ip: []const u8, enable_proxy_protocol: bool, is_last_allowed_request: bool) !void {
     var keep_alive = false;
     keep_alive_out.* = false;
     defer keep_alive_out.* = keep_alive;
@@ -3881,6 +3887,9 @@ fn handleConnection(conn: anytype, session: *ConnectionSession, cfg: *const edge
     keep_alive = request.keepAlive();
     if (streaming_request_body != null) keep_alive = false;
     if (http.shutdown.isShutdownRequested()) keep_alive = false;
+    // Tell the client this is the last response on this connection (RFC
+    // 7230 §6.6) instead of closing the socket after promising keep-alive.
+    if (is_last_allowed_request) keep_alive = false;
 
     // --- Correlation ID ---
     const correlation_id = try http.correlation.fromHeadersOrGenerate(allocator, &request.headers);
--- a/tests/integration.zig
+++ b/tests/integration.zig
@@ -16324,9 +16324,10 @@ test "max_requests_per_connection sends Connection: close on the response that h
 
     // The server closed after the limiting response — the client sees a clean
     // EOF on a further read rather than the connection silently swallowing
-    // the next request.
+    // the next request. Do not swallow read errors here: a stalled or reset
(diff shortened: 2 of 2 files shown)


Jaruso (commented):
The production change is correctly targeted at #682: `serveOneRequestWithConfig` decides *before* the response is written whether the current request reaches `max_requests_per_connection`, and `handleConnection` folds that into the existing `keep_alive` decision so normal response paths emit `Connection: close`. The existing post-response close path remains intact, `max_requests_per_connection == 0` still means unlimited, and the boundary behavior is correct for values such as 1 and 3.

I found one blocking problem in the new regression test. The test says it verifies that the server closes the socket cleanly after the limiting response, but `stream.read(&probe) catch 0` converts every read error into the same value as EOF. A socket that remains open until the 5-second timeout, or one that terminates with a reset/error, therefore passes the asserted `n == 0` just like a clean FIN. That weakens the exact regression coverage being added for #682 and makes the test comment/test plan stronger than what is actually asserted.

Please make the read error observable and require a genuine zero-length read. The simplest form should be along these lines:

```zig
var probe: [1]u8 = undefined;
const n = try stream.read(&probe);
try std.testing.expectEqual(@as(usize, 0), n);
```

Because the stream already has a 5-second timeout, a failure to close will then fail the test instead of being silently accepted. If `compat.NetStream.read` exposes platform-specific timeout/reset errors that need normalization, handle those explicitly as failures rather than mapping them to zero.

I would treat this as blocking before merge even though GitHub does not allow me to submit a formal REQUEST_CHANGES review on a PR opened by the connected account.

CI check: I inspected the current 38 check runs. No failures were present at review time; several long-running build/test jobs were still in progress, so the workflow was not fully complete yet.

Jaruso (commented):
Second-pass review complete on current head `c6995cf5`.

The prior blocking test issue is fixed exactly as requested: the EOF assertion now uses `try stream.read(&probe)`, so timeout/reset errors fail the regression test instead of being misclassified as a clean FIN. I re-reviewed both changed files against #682. The production logic still correctly determines the last allowed request before writing the response, propagates that decision into the existing HTTP/1 keep-alive path, emits `Connection: close` on the limiting response, preserves `0` as unlimited, and retains the post-response socket close.

The regression test now covers the required observable behavior: responses before the limit remain keep-alive, the limiting response says `Connection: close`, and the connection then reaches genuine EOF.

Current CI for `c6995cf5` is fully green. The main CI workflow and Darwin Release Smoke both completed successfully; the Ubuntu appliance job also ran the full integration suite successfully. I found no new defects or unmet acceptance criteria on this pass.

No blocking findings. This is ready to merge from my review perspective.

Issue #670, opened 2026-08-25
release: black-box public Homebrew install and installed-artifact smoke


Jaruso:
## Goal

Validate the **actual public Homebrew distribution path and the installed `tardi` binary** as a black-box consumer would use it.

The current repository has strong release-archive and generated-formula smokes, but those are not the same thing as installing from `Bare-Systems/homebrew-tap` and then exercising only the installed artifact. This issue closes that distribution-boundary gap for the current v0.6.x train.

## Why this matters

The first 0.6.x releases already exposed distribution-only defects that normal source/CI testing did not catch (CPU baseline and Linux glibc portability). The same principle applies to Homebrew: the exact bytes and formula path users install need their own evidence.

This is **not** a benchmark issue. It should run on ordinary CI/local environments and must not wait on #593/#669.

## Existing machinery to reuse

- `scripts/test-homebrew-formula.sh`
- `scripts/test-homebrew-release-formula.sh`
- `scripts/test-homebrew-tap-sync.sh`
- `scripts/update-homebrew-formula.sh`
- `scripts/audit-release-binary.sh`
- `.github/workflows/darwin-release-smoke.yml`
- public tap: `Bare-Systems/homebrew-tap`

Do not replace the generated-formula smoke. The public-tap smoke catches a different class of failure.

## Required platforms

- [ ] macOS ARM64
- [ ] macOS x86_64
- [ ] Linux x86_64 with Linuxbrew
- [ ] Linux ARM64 where runner availability is practical

### Fully-qualified one-shot install

```bash
brew install bare-systems/tap/tardigrade
```

### Explicit tap + short-name install

Current Homebrew requires trust for non-official taps. Validate the supported explicit path as well:

```bash
brew tap bare-systems/tap
brew trust --formula bare-systems/tap/tardigrade
brew install tardigrade
```

If Homebrew changes this UX again, update the test/documentation to match current Homebrew behavior rather than weakening the assertion.

## Installed-artifact assertions

After install, **do not use any checkout-local binary**.

- [ ] `command -v tardi` resolves inside Homebrew prefix.
- [ ] `command -v tardigrade` resolves to the packaged alias.
- [ ] `tardi version` reports the expected release version.
- [ ] version output reports the intended native/general TLS profile/backend.
- [ ] `tardi init static` works from the installed binary.
- [ ] `tardi check` accepts the generated config.
- [ ] `tardi run` starts successfully.
- [ ] static GET succeeds.
[...]

PR #686, opened 2026-08-26
Add public Homebrew tap smoke

Jaruso:
## Summary
- add a black-box `scripts/test-public-homebrew-tap.sh` that installs from `bare-systems/tap`, verifies `tardi` and `tardigrade` resolve under the Homebrew prefix, checks release identity and native TLS reporting, audits the installed binary, and runs init/check/run/static/health/SIGINT from a temp directory
- add a manual/scheduled `Public Homebrew Smoke` workflow covering macOS ARM64, macOS x86_64, Linux x86_64, and Linux ARM64 for both fully qualified and explicit tap/short-name install modes
- document the post-release public tap smoke in packaging docs and the release checklist

Refs #670. The issue should be closed after this PR merges and the scheduled/manual smoke has run successfully on the hosted platform matrix.

## Validation
- `bash -n scripts/test-public-homebrew-tap.sh scripts/audit-dependencies.sh`
- `shellcheck scripts/test-public-homebrew-tap.sh scripts/audit-dependencies.sh`
- `git diff --check`
- `./scripts/audit-dependencies.sh`
- `./scripts/test-public-homebrew-tap.sh --install-mode qualified` on local Darwin arm64: passed against public tap commit `b7bda62bda3da9127e52cedd06bb9ffed51aa635`, formula `0.6.2`, Homebrew `6.0.18-173-g14dedc3`
- `./scripts/test-public-homebrew-tap.sh --install-mode tap-short` on local Darwin arm64: passed against the same public tap commit/formula

Note: `actionlint` is not installed in this local environment, so workflow syntax validation is left to GitHub CI.

### DIFF
--- a/.github/workflows/public-homebrew-smoke.yml
+++ b/.github/workflows/public-homebrew-smoke.yml
@@ -2,6 +2,11 @@ name: Public Homebrew Smoke
 
 on:
   workflow_dispatch:
+    inputs:
+      expected_version:
+        description: "Expected public Tardigrade release version, with or without leading v. Defaults to the latest stable GitHub release."
+        required: false
+        type: string
   schedule:
     - cron: "37 11 * * 1"
 
@@ -60,6 +65,7 @@ jobs:
       - name: Smoke public Homebrew tap install
         env:
           INSTALL_MODE: ${{ matrix.install_mode }}
+          EXPECTED_VERSION: ${{ inputs.expected_version || '' }}
         run: |
           set -euo pipefail
           if [ "$RUNNER_OS" = Linux ]; then
--- a/.github/workflows/public-homebrew-smoke.yml
+++ b/.github/workflows/public-homebrew-smoke.yml
@@ -1,0 +1,68 @@
+name: Public Homebrew Smoke
+
+on:
+  workflow_dispatch:
+  schedule:
+    - cron: "37 11 * * 1"
+
+permissions:
+  contents: read
+
+concurrency:
+  group: public-homebrew-smoke-${{ github.ref }}
+  cancel-in-progress: false
+
+jobs:
+  public-homebrew-smoke:
+    name: Public Homebrew smoke (${{ matrix.os }}, ${{ matrix.install_mode }})
+    runs-on: ${{ matrix.os }}
+    timeout-minutes: 30
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - os: macos-15
+            install_mode: qualified
+          - os: macos-15
+            install_mode: tap-short
+          - os: macos-15-intel
+            install_mode: qualified
+          - os: macos-15-intel
+            install_mode: tap-short
+          - os: ubuntu-latest
+            install_mode: qualified
+          - os: ubuntu-latest
+            install_mode: tap-short
+          - os: ubuntu-24.04-arm
+            install_mode: qualified
+          - os: ubuntu-24.04-arm
+            install_mode: tap-short
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+        with:
+          persist-credentials: false
+
+      - name: Install Linux dependencies
+        if: runner.os == 'Linux'
+        run: sudo apt-get install -y build-essential curl file git procps ruby
+
+      - name: Install Linuxbrew
+        if: runner.os == 'Linux'
+        env:
+          NONINTERACTIVE: "1"
+        run: |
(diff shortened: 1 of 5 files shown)


Jaruso (commented):
**Blocking findings — I would request changes if GitHub allowed self-review change requests.**

I reviewed all five changed files against #670, the surrounding Homebrew/release machinery, the current public tap formula, and the completed CI/Darwin smoke runs. The workflow matrix and runner labels are valid, actionlint/shellcheck are green, and the install/alias/TLS/linkage/runtime coverage is otherwise well shaped.

Two acceptance-level gaps remain:

1. **Release identity is not independently verified against the release we expect to be public.** The current tap can be stale and this smoke will still pass, because `brew test` only proves the public formula and its binary agree with each other, while the harness itself only checks the TLS markers. #670 explicitly requires `tardi version` to report the expected release version. The expected version needs to come from outside the tap under test (for example a workflow input for manual release validation and/or the latest stable GitHub release for the scheduled run), then both the tap formula version and installed binary output should be asserted against it.
2. **`server.log` is discarded for several runtime failures.** Only readiness failure prints it. A wrong static body, `/health` failure, or non-clean SIGINT exit trips `set -e`, then the EXIT trap deletes the temp tree. #670 explicitly calls for retaining the startup log on failure. Please make failure cleanup emit/preserve that log for every post-start failure.

After these are fixed, the implementation should be in good shape. As the PR body already notes, #670 should remain open until the new manual/scheduled public-tap workflow has actually completed successfully across the hosted matrix; the PR-time CI does not execute this post-release workflow.

Jaruso (commented):
Re-review: there are no new commits since my previous review. The PR head is still `441269c2bea9e48b4de85703faa54c6d8bfb18f3`, so the implementation is byte-for-byte the same revision I reviewed previously.

Both existing blocking threads therefore still apply and remain unresolved:

1. **P1 — expected release identity is not independently asserted.** A stale public tap can still pass because the harness checks TLS markers but never compares the tap formula and installed `tardi version` against an expected release value sourced outside the tap under test.
2. **P2 — `server.log` is still discarded for post-readiness failures.** Static body, `/health`, `kill`, or non-zero `wait` failures can exit through `set -e` and then delete the temp directory before the runtime log is preserved/emitted.

CI and Darwin Release Smoke are green on this head, but those do not address these #670 acceptance-level gaps. No additional defects were introduced because there is no new code to inspect. Please push the fixes before the next re-review.

Jaruso (commented):
**One blocker remains after the second-pass fixes.**

I re-reviewed all five changed files against #670 and rechecked the two prior findings. Both previous blockers are fixed and I resolved those threads: the harness now compares the public formula and installed binary against an independently supplied/resolved expected version, and the EXIT cleanup emits `server.log` on any post-start failure before deleting the temporary tree.

The remaining blocker is in the new default latest-release resolver. GitHub-hosted runners have `gh` preinstalled, so the scheduled workflow will always take the `gh release view` branch when `expected_version` is empty. GitHub's documented Actions usage requires `GH_TOKEN` for `gh`, but this workflow never sets it. As written, the recurring scheduled smoke can fail at release-version lookup before exercising the identity assertions. The same default path also fails locally when `gh` is installed but unauthenticated instead of falling back to the public REST endpoint.

Please make the resolver robust to an unauthenticated `gh` and authenticate the workflow path (or simply use the public REST endpoint consistently). After that, I don't see another code-level gap against #670. The new CI run's lint/actionlint/shellcheck stage is green and Darwin Release Smoke is green; the full CI run is still finishing as of this review.

As before, #670 should remain open after merge until the actual manual/scheduled `Public Homebrew Smoke` has succeeded across the hosted matrix, since PR-time CI intentionally does not execute this post-release workflow.

Jaruso (commented):
Re-review complete on `1965b0ad0ed98aad1310fe4acaebe4ecfd6d2f48`.

The last blocking finding is fixed: the workflow now sets `GH_TOKEN: ${{ github.token }}` for hosted runs, and `latest_stable_release_version()` falls back to the public releases API if `gh` is unavailable or fails. I re-swept the full five-file change against #670 and did not find any additional correctness or acceptance-criteria gaps.

The prior review threads are now resolved. Actionlint/shellcheck and the relevant completed CI jobs are green, and Darwin Release Smoke is green. The current CI run still has a few long-running jobs in progress at review time, so merge should still wait for the required checks to finish successfully.

From a code-review standpoint this is ready to merge once CI is green. Keep #670 open after merge until the new manual/scheduled `Public Homebrew Smoke` has actually completed successfully across the hosted platform matrix, as the PR body already states.

PR #689, opened 2026-08-26
ci: verify v0.6.x Linux archives on Ubuntu 22.04 and Rocky 9

### LINKED ISSUE #671: release: verify v0.6.x Linux archives on Ubuntu 22.04 and Rocky 9
## Goal

Prove that the **exact published Linux release archive** for the current v0.6.x line starts and serves traffic on older still-relevant glibc userlands, specifically Ubuntu 22.04 and Rocky Linux 9.

This is a compatibility gate for the release artifact itself, not for a source rebuild.

## Why this matters

The v0.6.x release train has already demonstrated why release-binary compatibility must be tested directly:

- one release needed the CPU baseline pinned to avoid `Illegal instruction` on real hardware;
- a later release required a newer glibc than Ubuntu 22.04 / Rocky 9 provide;
- the current packaging model intentionally pins the Linux userland floor to avoid those failures.

The release invariant therefore needs an executable regression, not just linker/configuration intent.

## Scope

Required x86_64 userlands:

- [ ] `ubuntu:22.04`
- [ ] `rockylinux:9`

ARM64 may be added if a representative older-userland runner/container path is available, but x86_64 is the release-blocking minimum for this ticket.

## Artifact rule

Use the archive that would be or has been published, e.g.:

```text
tardigrade-linux-x86_64.tar.gz
```

Do **not** run `zig build` inside the target distro and call that compatibility evidence.

For pre-publication release validation, feed the exact release-workflow archive into the containers. For post-publication validation, download the exact release asset and record its SHA-256.

## Required assertions on each distro

- [ ] archive extracts cleanly;
- [ ] `tardi` is executable;
- [ ] `tardi version` succeeds;
- [ ] no `GLIBC_X.Y not found` or loader error;
- [ ] no CPU-feature `Illegal instruction` failure on representative hardware;
- [ ] `tardi check` succeeds on a minimal config;
- [ ] `tardi run` starts;
- [ ] an HTTP request receives the expected body;
- [ ] clean SIGINT shutdown succeeds;
- [ ] linkage/dependency audit output is retained.

Example target config:

```nginx
listen 18089;
server_name localhost;

location = /health {
    return 200 compat-ok;
}
```

Example container shape:

```bash
docker run --rm \
  -v "$PWD/artifact:/artifact:ro" \
  ubuntu:22.04 \
  /bin/sh -euxc '
    apt-get update
    apt-get install -y curl ca-certificates
    /artifact/tardi version
    /artifact/tardi check /tmp/tardigrade.conf
    /artifact/tardi run -c /tmp/tardigrade.conf >/tmp/tardi.log 2>&1 &
    pid=$!
    trap "kill \$pid 2>/dev/null || true" EXIT
    # poll and curl /health, assert compat-ok, then SIGINT + wait
  '
```

Repeat with `rockylinux:9` using `dnf`.

## CI/release integration

Add a release-artifact compatibility job that runs against the **archive output of the release workflow** before publication where practical.

A post-release/manual verification can additionally download the public asset, but it should not be the only gate if the pipeline can test the exact candidate archive earlier.

The job should fail loudly on loader/version/CPU incompatibility and print the full startup log.

## Evidence to retain

- release/tag or candidate identifier
- archive filename
- SHA-256
- source commit SHA
- target image digest/version
- `ldd --version` / glibc version where available
- architecture
- `tardi version`
- dependency/linkage audit
- startup log on failure

## Non-goals

- throughput/latency benchmarking
- epoll/io_uring comparison
- `perf`, `strace`, or `tc netem`
- #593/#669 controlled-host performance
- proving every Linux distribution is supported

## Acceptance criteria

- [ ] Exact v0.6.x Linux x86_64 release archive passes on Ubuntu 22.04.
- [ ] Exact archive passes on Rocky Linux 9.
- [ ] Test proves actual request handling, not only `--version`.
- [ ] Test proves clean process shutdown.
- [ ] Release/candidate SHA and userland versions are retained.
- [ ] A regression job is integrated into release CI or a clear documented alternative is committed.
- [ ] Any discovered ABI/CPU portability defect is fixed and the same artifact-path test passes before closure.

## Related

- #670 — public Homebrew installed-artifact smoke.
- #593 / #669 — controlled-host performance is separate and should not block this issue.

Jaruso:
### Description
This PR implements the requested CI verification for Linux release archives to ensure compatibility with older still-relevant glibc userlands (specifically `ubuntu:22.04` and `rockylinux:9`).

### Changes
- **`scripts/verify-linux-compatibility.sh`**: Extracts the `.tar.gz` artifact within an arbitrary Docker image, starts the process, verifies clean shutdown, and asserts `compat-ok` health status.
- **`linux-release-smoke.yml`**: Runs compatibility smoke tests on PRs using `x86_64-linux-gnu.2.28` to catch glibc regressions early.
- **`release.yml`**: Adds a `verify-linux-archive` job that gates `publish-release`, ensuring we test the exact release archive before publishing.

Closes #671

### DIFF
--- a/.github/workflows/linux-release-smoke.yml
+++ b/.github/workflows/linux-release-smoke.yml
@@ -1,0 +1,122 @@
+name: Linux Release Smoke
+
+on:
+  pull_request:
+    paths:
+      - ".github/workflows/linux-release-smoke.yml"
+      - ".github/workflows/release.yml"
+      - "build.zig"
+      - "build.zig.zon"
+      - "src/**"
+      - "scripts/install-zig.sh"
+      - "scripts/verify-linux-compatibility.sh"
+      - "scripts/audit-release-binary.sh"
+      - "scripts/package-release-archive.sh"
+      - "packaging/**"
+  workflow_dispatch:
+    inputs:
+      tag:
+        description: Published release tag to verify
+        required: true
+        default: v0.6.2
+
+permissions:
+  contents: read
+
+concurrency:
+  group: linux-release-smoke-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  linux-release-smoke:
+    name: Linux release smoke (${{ matrix.distro }})
+    runs-on: ubuntu-latest
+    timeout-minutes: 30
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - archive_name: tardigrade-linux-x86_64
+            zig_target: x86_64-linux-gnu.2.28
+            distro: ubuntu:22.04
+          - archive_name: tardigrade-linux-x86_64
+            zig_target: x86_64-linux-gnu.2.28
+            distro: rockylinux:9
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+        with:
+          persist-credentials: false
+
+      - name: Install Zig
+        if: github.event_name == 'pull_request'
+        run: sh ./scripts/install-zig.sh 0.16.0
+
+      - name: Print Zig version
+        if: github.event_name == 'pull_request'
+        run: zig version
+
+      - name: Build native release binary
+        if: github.event_name == 'pull_request'
+        env:
+          ZIG_TARGET: ${{ matrix.zig_target }}
+        run: |
+          zig build -Doptimize=ReleaseFast -Dcpu=baseline "-Dtarget=$ZIG_TARGET" -Dversion=0.0.0-ci-smoke
+
+      - name: Audit native release binary linkage
+        if: github.event_name == 'pull_request'
+        run: |
+          ./scripts/audit-release-binary.sh \
+            --binary zig-out/bin/tardi \
+            --profile general \
+            --output "$RUNNER_TEMP/dependency-inventory-${{ matrix.archive_name }}.json"
+
+      - name: Package artifact
+        if: github.event_name == 'pull_request'
+        env:
+          ARCHIVE_NAME: ${{ matrix.archive_name }}
+        run: |
+          ./scripts/package-release-archive.sh \
+            --archive-name "$ARCHIVE_NAME" \
+            --binary zig-out/bin/tardi \
+            --staging-dir dist \
+            --output-dir "$RUNNER_TEMP/release-artifacts"
+
+      - name: Setup PR expected version
+        if: github.event_name == 'pull_request'
+        run: echo "EXPECTED_VERSION=0.0.0-ci-smoke" >> "$GITHUB_ENV"
+
+      - name: Resolve published release identity
+        if: github.event_name == 'workflow_dispatch'
+        env:
+          TAG: ${{ inputs.tag }}
+        run: |
+          set -euo pipefail
+          git fetch --force origin "refs/tags/$TAG:refs/tags/$TAG"
+          release_sha="$(git rev-list -n 1 "$TAG")"
+          expected_version="${TAG#v}"
+          printf 'published_tag=%s\n' "$TAG"
+          printf 'expected_version=%s\n' "$expected_version"
+          printf 'source_commit_sha=%s\n' "$release_sha"
+          printf 'EXPECTED_VERSION=%s\n' "$expected_version" >> "$GITHUB_ENV"
+
+      - name: Download published release asset
+        if: github.event_name == 'workflow_dispatch'
+        env:
+          GH_TOKEN: ${{ github.token }}
+          TAG: ${{ inputs.tag }}
+          ARCHIVE_NAME: ${{ matrix.archive_name }}
+        run: |
+          mkdir -p "$RUNNER_TEMP/release-artifacts"
+          gh release download "$TAG" -p "${ARCHIVE_NAME}.tar.gz" -D "$RUNNER_TEMP/release-artifacts"
+
+      - name: Verify compatibility on ${{ matrix.distro }}
+        env:
+          ARCHIVE_NAME: ${{ matrix.archive_name }}
+          DISTRO: ${{ matrix.distro }}
+        run: |
+          ./scripts/verify-linux-compatibility.sh \
+            "$DISTRO" \
+            "$RUNNER_TEMP/release-artifacts/$ARCHIVE_NAME.tar.gz" \
+            "$EXPECTED_VERSION"
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -344,11 +344,59 @@ jobs:
           if-no-files-found: error
           retention-days: 1
 
+  verify-linux-archive:
+    name: Verify ${{ matrix.archive_name }} on ${{ matrix.distro }}
+    needs:
+      - prepare-release
+      - build-release
+    if: needs.prepare-release.result == 'success' && needs.prepare-release.outputs.should_publish == 'true'
+    runs-on: ubuntu-latest
+    timeout-minutes: 30
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - archive_name: tardigrade-linux-x86_64
+            distro: ubuntu:22.04
+          - archive_name: tardigrade-linux-x86_64
+            distro: rockylinux:9
+
+    steps:
+      - name: Checkout release commit
+        uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+        with:
+          ref: ${{ needs.prepare-release.outputs.release_sha }}
+          persist-credentials: false
+
+      - name: Download archive artifact
+        uses: actions/download-artifact@3e5f45b2cfb9172054b4087a40e8e0b5a5461e7c # v8.0.1
+        with:
+          name: ${{ matrix.archive_name }}
+          path: ${{ runner.temp }}/verify-artifacts
+
+      - name: Verify compatibility on ${{ matrix.distro }}
+        env:
+          ARCHIVE_NAME: ${{ matrix.archive_name }}
+          DISTRO: ${{ matrix.distro }}
+          RELEASE_TAG: ${{ needs.prepare-release.outputs.tag }}
+          EXPECTED_VERSION: ${{ needs.prepare-release.outputs.version }}
+          RELEASE_SHA: ${{ needs.prepare-release.outputs.release_sha }}
+        run: |
+          set -euo pipefail
+          printf 'candidate_tag=%s\n' "$RELEASE_TAG"
+          printf 'expected_version=%s\n' "$EXPECTED_VERSION"
+          printf 'source_commit_sha=%s\n' "$RELEASE_SHA"
+          ./scripts/verify-linux-compatibility.sh \
+            "$DISTRO" \
+            "$RUNNER_TEMP/verify-artifacts/$ARCHIVE_NAME.tar.gz" \
+            "$EXPECTED_VERSION"
+
   publish-release:
     name: Publish GitHub release assets
     needs:
       - prepare-release
       - build-release
+      - verify-linux-archive
     if: needs.prepare-release.result == 'success' && needs.prepare-release.outputs.should_publish == 'true'
     runs-on: ubuntu-latest
     permissions:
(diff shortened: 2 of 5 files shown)


Jaruso (commented):
**Review result: blocking changes required before merge.**

I reviewed all three changed files against #671 and its implementation-ready close rule. The overall release-gate placement is correct — `publish-release` now depends on verification of the archive emitted by `build-release`, and the smoke exercises config validation, a real HTTP response, and SIGINT — but this is not ready to merge yet.

Blocking changes:

1. **The verifier cannot consume the archive paths supplied by either workflow.** Both callers pass an absolute `$RUNNER_TEMP/...tar.gz`; the script prepends `$(pwd)/`, producing a nonexistent bind source. The current Linux Release Smoke run confirms this exact failure: Docker mounts a directory at `/artifact.tar.gz`, then `tar` reports `Cannot read: Is a directory`.
2. **#671's retained-evidence contract is incomplete.** Record the archive filename + SHA-256 before execution, candidate/release SHA and expected version, image tag + resolved digest, `/etc/os-release`, `uname -m`, glibc version, `tardi version`, and dependency/linkage output. Required audit commands should not be hidden behind `|| true`. Also force `--platform linux/amd64` so the x86_64 requirement is explicit rather than an assumption about the hosted runner.
3. **The PR does not prove the already-published current v0.6.x artifact.** The latest release is v0.6.2 and its published `tardigrade-linux-x86_64.tar.gz` has GitHub digest `sha256:2f98ed94aaf9caf652d4add7045bc446e6056b07845387dd0b3e0f3d919c2a03`. `workflow_dispatch` currently rebuilds source as `0.0.0-ci-smoke`; that is useful regression coverage but cannot satisfy the ticket's current-artifact close rule. Add a manual/tag input path (or a separate manual workflow) that downloads the published asset and sends those exact bytes through the same verifier on Ubuntu 22.04 and Rocky 9.
4. **Shutdown verification needs a bounded failure path with diagnostics.** `wait "$pid"` can block until the Actions job-level timeout (and the new release job currently has no explicit timeout). Poll for exit after SIGINT with a short deadline; if it remains alive or exits non-zero, print `tardi.log`, clean it up, and fail. Add a job timeout to `verify-linux-archive` as a final safety net.

Please also tighten the small interface/packaging checks while touching the script: validate `$#` before reading `$1/$2` (`set -u` currently makes the usage branch unreachable for missing args), `test -f` the archive before invoking Docker, verify the `tardigrade -> tardi` symlink after extraction, and capture/assert the expected version/profile/backend rather than merely requiring `tardi version` to exit 0.

Once those are fixed and the Linux smoke passes **both** distro legs, this should meet the intended shape of #671.

Jaruso (commented):
Second full review against #671 and the implementation-ready acceptance plan.

This is materially improved: the absolute-path mount bug is fixed; Ubuntu 22.04 now successfully extracts the archive, executes `tardi version`, validates config, serves `compat-ok`, and exits cleanly after SIGINT; image/archive hashes are now printed; and `publish-release` is correctly gated by the compatibility matrix. I resolved the old mount and shutdown threads.

I am **not ready to merge yet**. GitHub will not allow this account to submit `REQUEST_CHANGES` on its own PR, so this is formally a COMMENT, but the items below are blocking.

### Current blocking state

1. **Rocky Linux 9 still does not execute Tardigrade.** The current Linux Release Smoke run fails in package setup because Rocky 9 already contains `curl-minimal`; `dnf install curl` conflicts with it. This means the primary Rocky acceptance criterion is still unproven.
2. **Main CI is red because this PR broke the production dependency audit.** `scripts/audit-dependencies.sh` rejects the newly-added Ubuntu package declaration (`file` and `binutils` are not in the reviewed metadata dependency allowlist). This needs an intentional policy-compatible fix, not suppression.
3. **The archive symlink assertion is not actually an assertion.** If `tardigrade` is missing entirely, the current conditional simply skips the check and the verifier can pass.
4. **The test still does not bind the binary's reported version to the expected candidate/tag, and the release/source commit SHA is not explicitly retained per matrix leg.** A mislabeled/wrong-version archive could therefore pass `tardi version` as long as it says `tls-profile=general` and `tls-backend=native`.
5. **The post-release path exists now, but #671 cannot close until it is actually run against the published v0.6.2 bytes on both distros.** The expected published x86_64 archive SHA is `2f98ed94aaf9caf652d4add7045bc446e6056b07845387dd0b3e0f3d919c2a03`; the v0.6.2 annotated tag resolves to source commit `34a331021b9f132639c6624ab9576bac9ef6edbe`. The final manual evidence should show those identities, `0.6.2` from the binary, both distro/userland identities, a real `compat-ok` request, and clean SIGINT shutdown.

### Expected verification before the next review

After making the inline changes, please run/require all of these:

```text
bash -n scripts/verify-linux-compatibility.sh
shellcheck scripts/verify-linux-compatibility.sh
./scripts/audit-dependencies.sh
actionlint .github/workflows/linux-release-smoke.yml .github/workflows/release.yml
git diff --check
```

Then the PR-triggered **Linux Release Smoke must be green for both Ubuntu 22.04 and Rocky 9**, and normal CI must be green. Finally, manually dispatch `Linux Release Smoke` from this branch with `tag=v0.6.2` and retain/link the successful run. In that run I expect to see:

```text
published_tag=v0.6.2
expected_version=0.6.2
source_commit_sha=34a331021b9f132639c6624ab9576bac9ef6edbe
archive=tardigrade-linux-x86_64.tar.gz
archive_sha256=2f98ed94aaf9caf652d4add7045bc446e6056b07845387dd0b3e0f3d919c2a03
```

plus the resolved Ubuntu/Rocky image digests, `/etc/os-release`, `uname -m`, glibc version, strict linkage output, exact `tardi version`, successful HTTP body assertion, and clean shutdown for **both** distros.

If those items are fixed and the evidence run is green, I do not currently see another architectural gap in #671.

Jaruso (commented):
Third-pass review of the full current head against #671 and every prior thread.

The second-round fixes are substantially correct: Rocky preserves `curl-minimal`; the dependency metadata allowlist was narrowly extended; archive layout is now strict; the verifier requires the exact expected version/profile/backend; release candidate tag/version/source SHA are retained; the manual workflow resolves the published tag to its commit; the distro checks are independent matrix legs; and the scratch `pr-body.md` file is gone.

I found **one remaining code-level acceptance gap** before I would consider the implementation complete: the issue's implementation plan explicitly requires retained ELF symbol-version evidence (`readelf --version-info` or equivalent). `binutils` is now installed in both distros, but the verifier never invokes `readelf`, so that evidence is still missing. I left an exact inline fix below.

The required post-release v0.6.2 run has also actually been dispatched from this PR head. Ubuntu 22.04 has already passed the exact published archive with SHA-256 `2f98ed94aaf9caf652d4add7045bc446e6056b07845387dd0b3e0f3d919c2a03`, source commit `34a331021b9f132639c6624ab9576bac9ef6edbe`, and exact binary identity `0.6.2 (tls-profile=general, tls-backend=native)`. Rocky 9 was still executing when this review was submitted, so #671's close rule is not yet fully evidenced until that leg completes successfully. The fresh PR CI runs are also still queued/running.

Once `readelf` evidence is added and the final PR + manual Rocky legs are green, I do not currently see another implementation gap in #671.

Jaruso (commented):
**Review result: one blocking regression remains.**

The requested `readelf --version-info` evidence is now present and the rest of the PR still matches #671's implementation/close criteria. However, the optional normalized GLIBC summary introduced a shell-quoting regression in `scripts/verify-linux-compatibility.sh`. The container program is passed to `/bin/sh -c` as one outer single-quoted string; the new inner single quotes terminate/reopen that string and corrupt the command. This is confirmed by the manual v0.6.2 run on this exact head (`33004696189`): **both Ubuntu 22.04 and Rocky 9 fail before any compatibility assertions execute**, with Ubuntu reporting `./scripts/verify-linux-compatibility.sh: 72: Syntax error: "(" unexpected`.

Please fix the quoting as described inline, then rerun both (1) normal PR Linux Release Smoke and (2) `workflow_dispatch` with `tag=v0.6.2` from this head. The manual run must again show the exact published digest `2f98ed94aaf9caf652d4add7045bc446e6056b07845387dd0b3e0f3d919c2a03`, source commit `34a331021b9f132639c6624ab9576bac9ef6edbe`, exact `0.6.2` binary identity, **and now the final `readelf` evidence** on both Ubuntu and Rocky.

I did not find another code/design issue in this pass. Once that rerun is green and the normal CI is green, #671's close rule should be satisfied.

Jaruso (commented):
Full re-review of the current head against #671 is complete.

The implementation/runtime acceptance criteria are now satisfied. The final manual `workflow_dispatch` run `33010021019` passed the exact published v0.6.2 `tardigrade-linux-x86_64.tar.gz` on both Ubuntu 22.04 and Rocky Linux 9 using the final verifier. The retained evidence binds the test to:

- `published_tag=v0.6.2`
- `source_commit_sha=34a331021b9f132639c6624ab9576bac9ef6edbe`
- `archive_sha256=2f98ed94aaf9caf652d4add7045bc446e6056b07845387dd0b3e0f3d919c2a03`
- exact `0.6.2 (tls-profile=general, tls-backend=native)` identity
- Ubuntu 22.04.5 / glibc 2.35 and Rocky 9.3 / glibc 2.34
- strict archive layout + `tardigrade -> tardi`
- `file`, `ldd`, full `readelf --version-info`, normalized GLIBC references
- successful `tardi check`, real `compat-ok` HTTP response, and clean bounded SIGINT shutdown

I also verified that the current head's `scripts/verify-linux-compatibility.sh` blob (`53f81dd13acb2f8e35899ed2dd2f4ba0b22e701e`) and `.github/workflows/linux-release-smoke.yml` blob (`562ce9fc4c0c93e50c1b3a5d2776e07bccd5b5c9`) are byte-for-byte identical to the versions used by that successful final-evidence run, so the later commit-history rewrite/documentation addition does not invalidate the runtime evidence.

There is **one remaining requested change**, in the newly-added CHANGELOG entry: it claims these checks guarantee the manylinux2014 glibc-floor baseline. That is not true. manylinux2014 corresponds to the glibc 2.17 baseline, while the exact artifact's retained `readelf` evidence contains `GLIBC_2.28`; the build itself is intentionally targeted at `x86_64-linux-gnu.2.28`. Passing Rocky 9 (glibc 2.34) and Ubuntu 22.04 (glibc 2.35) therefore proves the project-required distro compatibility, not compatibility with glibc 2.17. I left an exact replacement below.

Fresh PR-triggered CI on the current head is still running/queued as of this review. Once the changelog claim is corrected and the current required checks are green, I have no other code, architecture, packaging, or #671 acceptance changes to request.

Jaruso (commented):
Full re-review complete on head `347fc9f24ab440c8c1b2b80c2ff16f7f32abc622`. I rechecked the entire five-file diff against #671 and its implementation-ready acceptance plan, including the release gate, PR/manual workflow paths, archive/layout assertions, exact version/profile/backend check, native-userland linkage evidence, `readelf --version-info`, HTTP runtime probe, bounded clean SIGINT shutdown, dependency-audit allowlist change, and changelog wording.

No blocking findings remain.

Evidence checked:
- Current-head `Linux Release Smoke` is green on both `ubuntu:22.04` and `rockylinux:9`.
- Current-head full `CI` is green, including actionlint/shellcheck, production dependency audit, packaging smoke, CPU/glibc-floor policy checks, and the broader test matrix.
- Current-head `Darwin Release Smoke` is green.
- Manual published-asset run `33010021019` passed both distros while downloading the existing `v0.6.2` release asset rather than rebuilding it.
- That run verified the exact published archive SHA-256 `2f98ed94aaf9caf652d4add7045bc446e6056b07845387dd0b3e0f3d919c2a03`, source commit `34a331021b9f132639c6624ab9576bac9ef6edbe`, exact `0.6.2 (tls-profile=general, tls-backend=native)` identity, linkage/GLIBC symbol-version evidence, real `compat-ok` HTTP handling, and clean SIGINT exit on both Ubuntu 22.04 and Rocky 9.
- The current `scripts/verify-linux-compatibility.sh` is unchanged from the verifier used by that successful manual evidence run.
- The final changelog wording now accurately describes the `x86_64-linux-gnu.2.28` target and the two required tested userlands without claiming a manylinux2014/GLIBC 2.17 guarantee.

All prior review threads are resolved. #671's close criteria are satisfied. This PR is ready to merge.

PR #690, opened 2026-08-26
Fix uneven route benchmark concurrency

### LINKED ISSUE #683: benchmarks: uneven-route-distribution runs routes sequentially, doesn't test concurrent fairness (#149/#593)
## Summary

`benchmarks/competitive/upstream-pool-matrix.sh`'s `uneven-route-distribution`
scenario is meant to validate that the shared upstream connection pool
treats an uneven mix of concurrent route traffic (one hot route, one warm,
one cold) fairly — this is part of the #147/#149 validation matrix that
#593 executes and reports on.

As currently implemented, it doesn't test that. It calls `run_measured_wrk`
three separate times, once per route, each a single blocking `wrk`
invocation that runs to completion before the next one starts:

```bash
route_a_json="$(run_measured_wrk route-a route-a 4 8 2 true)"   # runs to completion
route_b_json="$(run_measured_wrk route-b route-b 2 4 1 false)"  # then this
route_c_json="$(run_measured_wrk route-c route-c 1 2 1 false)"  # then this
```

(exact durations/connections/threads per call may drift; the point is
they're sequential calls, each blocking until its own `wrk` process exits.)

The three routes never compete for the shared worker pool or connection
pool at the same time — each gets the pool entirely to itself. This can't
demonstrate or refute route-level starvation/fairness under real uneven
*concurrent* traffic, which is what #149's acceptance criteria and #593's
"is accept distribution/route traffic reasonably fair" language actually
ask for.

## Impact

Every #593 competitive-suite report that includes this scenario (see the
2026-08-25 report under `benchmarks/results/`) has had to caveat the
`uneven-route-distribution` numbers as "informational, not fairness
evidence" rather than treating them as answering the #149 acceptance
question they're nominally there to answer.

## Suggested fix direction

Run the three routes as concurrent `wrk` processes (background each with
`&`, `wait` for all three) against the same live Tardigrade instance and
worker pool, then:

- Take one `current_metrics` snapshot before all three start and one after
  all three finish, so pool-level deltas (reuse counts, lock-wait,
  contention) reflect the full concurrent window rather than three
  disjoint windows.
- Keep each route's own `rps`/`p99`/errors attributed per-route from its
  own `wrk` process's output, same as today.
- Decide whether per-route worker-queue-depth/lock-wait attribution is
  possible at all with concurrent traffic (it may not be, if the metrics
  are process-wide, not per-route) — if not, document that the *combined*
  contention/reuse numbers are what's actually measurable under concurrent
  uneven traffic, not a per-route breakdown of contention.
- Update the weight labels (`requested_weight: 80/15/5` in the combined
  JSON) to reflect actual concurrent request-rate ratios achieved, not
  just wrk's configured `-c`/`-t` per route, since concurrent contention
  may shift the realized ratio away from the configured one.

## References

Refs #149
Refs #593

Jaruso:
## Summary

- replace the uneven-route-distribution alias/sequential route runs with one concurrent route-a/route-b/route-c wrk window
- record per-route wrk rps/p99/errors/config/achieved shares separately from combined process-wide pool/CPU/RSS/fd metrics
- add a deterministic benchmark harness test with a fake wrk barrier and wire it into CI shellcheck/test coverage

Fixes #683

## Validation

- bash -n benchmarks/competitive/upstream-pool-matrix.sh benchmarks/competitive/test-upstream-pool-matrix.sh
- shellcheck benchmarks/competitive/upstream-pool-matrix.sh benchmarks/competitive/test-upstream-pool-matrix.sh
- ./benchmarks/competitive/test-upstream-pool-matrix.sh

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -58,6 +58,8 @@ jobs:
           benchmarks/run.sh
           benchmarks/competitive/run.sh
           benchmarks/competitive/netem-impair.sh
+          benchmarks/competitive/upstream-pool-matrix.sh
+          benchmarks/competitive/test-upstream-pool-matrix.sh
           benchmarks/test-h3-benchmark.sh
 
       # A native `zig build` with no explicit -Dcpu embeds the exact CPU
@@ -681,6 +683,9 @@ jobs:
       # coverage.
       - name: Run H3 benchmark harness tests (#256-G)
         run: ./benchmarks/test-h3-benchmark.sh
+
+      - name: Run upstream pool matrix harness tests (#683)
+        run: ./benchmarks/competitive/test-upstream-pool-matrix.sh
 
       - name: Stage perf smoke artifact
         if: always()
--- a/benchmarks/competitive/run.sh
+++ b/benchmarks/competitive/run.sh
@@ -1123,10 +1123,11 @@ write_combined_outputs() {
              (.value.quic.udp_buffers.recv.status // null), (.value.quic.udp_buffers.send.status // null)]),
         (if .upstream_pool_matrix != null then
             (.upstream_pool_matrix.scenarios | to_entries[] |
+                (.value.combined // .value) as $v |
                 ["tardigrade", ("upstream-pool/" + .key), (if .value.covered != null then .value.covered elif .value.supported != null then .value.supported else true end), (.value.reason // .value.sharding_note // null),
-                 (.value.rps // null), null, null, (.value.p99_ms // null), null, (.value.p99_ttfb_ms // null), null,
-                 (.value.cpu_pct_avg // null), (.value.cpu_ms_per_request // null), (.value.rss_mb_peak // null), (.value.open_fds_peak // null), (.value.errors // null),
-                 (.value.pool_lock_wait_ns_per_request // null), (.value.pool_lock_wait_ns_per_acquire // null),
+                 ($v.rps // null), null, null, ($v.p99_ms // null), null, ($v.p99_ttfb_ms // null), null,
+                 ($v.cpu_pct_avg // null), ($v.cpu_ms_per_request // null), ($v.rss_mb_peak // null), ($v.open_fds_peak // null), (.value.errors // $v.errors // null),
+                 ($v.pool_lock_wait_ns_per_request // null), ($v.pool_lock_wait_ns_per_acquire // null),
                  null, null, null, null, null, null, null]),
             (.upstream_pool_matrix.scenarios["uneven-route-distribution"].routes // {} | to_entries[] |
                 ["tardigrade", ("upstream-pool/uneven/" + .key), (if .value.covered == null then true else .value.covered end), (.value.reason // null),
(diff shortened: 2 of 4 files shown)


Jaruso (commented):
**Review result: changes required before merge.** GitHub will not allow this connected account to submit a formal `REQUEST_CHANGES` review on its own PR, so I am recording the blockers as a comment review instead.

I reviewed the full PR against #683, its implementation-ready reconciliation, the upstream-pool requirements carried by #149/#593, the existing competitive result formatter, and current head CI.

The core design is good and matches most of #683: route A is no longer aliased to `/hot`; all three routes use the same duration and run concurrently; the intended 16/3/1 offered-load skew is recorded; route-local RPS/p99/errors are separated from one process-wide metrics/monitoring window; achieved shares come from measured RPS; malformed/non-zero route cases are covered; and the focused harness test is wired into CI. The perf-smoke job confirms the new deterministic harness currently passes.

There are four items that need to be fixed before merge:

1. **Current head CI is red.** The Lint job (`98262503595`) fails because the new trap-cleanup suppression uses `SC2329`, while the CI ShellCheck invocation reports `SC2317` on `rm -rf "$TEST_DIR"`.
2. **Child status publication is racy.** The parent treats status-file existence as completion, but `> "$status_file"` creates the file before `printf` writes the status. A poll can therefore read an empty file and interpret it as zero, violating #683's explicit requirement to preserve non-zero `wrk` exits. Publish status atomically (and preferably capture `wait` status explicitly rather than depending on Bash `errexit` call-context behavior).
3. **`achieved_request_count` is computed from the wrong elapsed interval.** Route RPS is based on each `wrk` process's own `summary.duration`, while the passed `elapsed` covers the whole shared monitor window from before launch until after the last route exits. Multiplying those values does not produce the route's achieved request count. The issue explicitly makes this field optional, so removing it is fine; otherwise emit/consume an exact route-local request count (for example `summary.requests` from `wrk-summary.lua`).
4. **The new nested `.combined` schema is not wired through the existing competitive CSV/Markdown formatter.** `benchmarks/competitive/run.sh` still reads aggregate upstream scenario fields at `.value.rps`, `.value.cpu_pct_avg`, `.value.new_connections_per_sec`, reuse/lock fields, etc. After this PR the uneven-route aggregate row will render those as `-` even though the evidence exists under `.value.combined`. Update the formatter to use `.combined // .` for aggregate fields (or another schema-compatible equivalent) and cover that path so the canonical #593 output does not silently lose the new combined-window evidence.

I left inline comments with concrete fixes for each point. Once these are addressed and CI is green, the implementation should be very close to satisfying #683 completely.

Jaruso (commented):
Re-review complete on b65bceeb5f0b0907917686a84e7a777c5d31d9c4. I re-checked the full #683 acceptance path and the changes made since my previous review.

All four prior findings are resolved:

- the ShellCheck suppression now covers the CI diagnostic and the Lint job is green;
- child exit status is captured explicitly and published atomically before the parent observes it;
- the misleading route-local `achieved_request_count` has been removed;
- `benchmarks/competitive/run.sh` now reads aggregate uneven-route metrics through `(.value.combined // .value)`, preserving CSV/Markdown reporting while keeping route-local detail separate.

I also re-checked the failure/cleanup path, concurrent-window semantics, route attribution, achieved-share calculation, and downstream reporting for new regressions. I did not find any additional blocking issues. The focused #683 harness passes in CI, Perf smoke is green, Lint is green, and the full head CI run completed successfully.

From a code-review standpoint this is ready to merge. GitHub does not permit the connected account to approve its own PR, so this is submitted as a COMMENT review rather than APPROVE.

